PR #7, opened 2019-02-19
Port exceptions submodule of moveit_core

vmayoral:


moveit_core/exceptions/CMakeLists.txt
```diff
@@ -1,11 +1,11 @@
 set(MOVEIT_LIB_NAME moveit_exceptions)
 
 add_library(${MOVEIT_LIB_NAME} src/exceptions.cpp)
-set_target_properties(${MOVEIT_LIB_NAME} PROPERTIES VERSION "${${PROJECT_NAME}_VERSION}")
+# set_target_properties(${MOVEIT_LIB_NAME} PROPERTIES VERSION ${${PROJECT_NAME}_VERSION})
```
mlautman: remove dead code

moveit_core/exceptions/src/exceptions.cpp
```diff
@@ -35,14 +35,20 @@
 /* Author: Acorn Pooley, Ioan Sucan */
 
 #include <moveit/exceptions/exceptions.h>
-#include <ros/console.h>
+// #include <ros/console.h>
```
mlautman: remove dead code

moveit_core/exceptions/src/exceptions.cpp
```diff
@@ -35,14 +35,20 @@
...
+#include <rcutils/logging_macros.h>
+#include "rclcpp/rclcpp.hpp"
+
+#define ROS_ERROR RCUTILS_LOG_ERROR
 
 moveit::ConstructException::ConstructException(const std::string& what_arg) : std::runtime_error(what_arg)
 {
-  ROS_ERROR_NAMED("exceptions", "Error during construction of object: %s\nException thrown.", what_arg.c_str());
+  // ROS_ERROR_NAMED("exceptions", "Error during construction of object: %s\nException thrown.", what_arg.c_str());
```
mlautman: remove dead code

moveit_core/exceptions/src/exceptions.cpp
```diff
@@ -35,14 +35,20 @@
...
+  ROS_ERROR("Error during construction of object: %s\nException thrown.", what_arg.c_str());
 }
 
 moveit::Exception::Exception(const std::string& what_arg) : std::runtime_error(what_arg)
 {
-  ROS_ERROR_NAMED("exceptions", "%s\nException thrown.", what_arg.c_str());
+  // ROS_ERROR_NAMED("exceptions", "%s\nException thrown.", what_arg.c_str());
```
mlautman: same

moveit_core/exceptions/src/exceptions.cpp
davetcoleman: +1 to keeping NAMED

moveit_core/exceptions/src/exceptions.cpp
```diff
@@ -35,14 +35,20 @@
 /* Author: Acorn Pooley, Ioan Sucan */
 
 #include <moveit/exceptions/exceptions.h>
-#include <ros/console.h>
+// #include <ros/console.h>
+#include <rcutils/logging_macros.h>
+#include "rclcpp/rclcpp.hpp"
+
+#define ROS_ERROR RCUTILS_LOG_ERROR
```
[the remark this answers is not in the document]
davetcoleman: this has already been created for moveit, just this week!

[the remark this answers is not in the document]
wjwwood: There is a "named" version in `RCLCPP_ERROR`, see:

http://docs.ros2.org/crystal/api/rclcpp/logging_8hpp.html#ad8f3687eda7312561768319e99cd4ca6

It requires that you provide a "Logger", but that's just a struct wrapping a string, so it's essentially a name. You can get the node's logger (which is just a logger named after the node and node namespace) which is recommended:

http://docs.ros2.org/crystal/api/rclcpp/classrclcpp_1_1Node.html#a1d6fda144e3748d52d3e10ea81e9558f

Once you have it, you can "extend" the logger name with `get_child(string)`:

http://docs.ros2.org/crystal/api/rclcpp/classrclcpp_1_1Logger.html#a68ab28459973c1072d3ad7a593ac994e

Or you can create a logger from scratch using your own invented name (less good, but ok if you don't have a node in the context, e.g. pre node construction) with `rclcpp::get_logger(string)`:

http://docs.ros2.org/crystal/api/rclcpp/namespacerclcpp.html#ae7295751947c08312aa69f45fd673171

Of you can use just an empty string when calling `rclcpp::get_logger("")`, which is the least good alternative because it doesn't give any context to the log consumer (user). This is essentially what unnamed macros in rcutils do like `RCUTILS_LOG_ERROR` (i.e. as opposed to `RCUTILS_LOG_ERROR_NAMED`), e.g.:

https://github.com/ros2/rcutils/blob/a7581a333444c2804f2b7c27dd0da1b2946bebff/resource/logging_macros.h.em#L183

In general though, you guys should be using `rclcpp`'s macros rather than the ones in `rcutils`, since you're depending on `rclcpp`. They are implemented in `rcutils` so we can reuse them in our C api and in python, as well as in c++.

See also:

- https://index.ros.org/doc/ros2/Concepts/Logging/
- https://github.com/ros2/demos/blob/master/logging_demo/src/logger_usage_component.cpp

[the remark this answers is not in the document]
wjwwood: Also, it's worth saying, that we don't provide an "unnamed" macro in rclcpp because unlike ROS 1's `ROS_ERROR`, in ROS 2 there's no singleton node who's name can be used with it. So having `RCLCPP_ERROR` with no logger/name as an argument would simply have no name or node attributed to it, unlike ROS 1. So it would be misleading/incorrect to use that as the alternative to ROS 1's `ROS_ERROR`. The correct equivalent in ROS 2 is `RCLCPP_ERROR(node->get_logger(), "my message");`.

moveit_core/exceptions/src/exceptions.cpp
wjwwood: You should use `RCLCPP_ERROR(some->way->to_get->node->get_logger(), "...");`, but if access to the node is impossible here, then something like `RCLCPP_ERROR(rclcpp::get_logger("moveit"), "...");`.

wjwwood: Note, if you're going for the latter you can store the logger object somewhere to avoid creating a `std::string` each time you get logger.

moveit_core/exceptions/src/exceptions.cpp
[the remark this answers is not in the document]
davetcoleman: ``moveit_core`` isn't suppose to have the notion of a ROS node, that all lives in the ``moveit_ros`` set of packages. Are you saying we'd need to track the node object throughout the moveit_core code? is there a logger that doesn't require this? or would, say, ``rclcpp::get_logger("moveit")`` suffice for a non-ROS part of the code?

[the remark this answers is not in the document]
wjwwood: > `moveit_core` isn't suppose to have the notion of a ROS node, that all lives in the `moveit_ros` set of packages.

You include `rclcpp/rclcpp.hpp` just above this, shouldn't not include ROS headers if moveit_core isn't supposed to depend on ROS concepts? I guess you mean just node, not ROS, in which case that's ok, I just assumed since you were including rclcpp you were making nodes too.

> Are you saying we'd need to track the node object throughout the moveit_core code?

Not necessarily, if moveit_ros creates a node, then from it a logger, then it could pass that logger into moveit_core's initialization and moveit_core could use that passed in logger as it's logger (or as the parent to a new child logger). So long as moveit_core depends on rclcpp that's fine.

If you want to avoid rclcpp, then you could still pass the logger name into moveit_core as a string, and then use the `RCUTILS_ERROR_NAMED(<passed in logger name, "msg");`.

> is there a logger that doesn't require this?

 I said above:

```
Or you can create a logger from scratch using your own invented name (less good, but ok if you don't have a node in the context, e.g. pre node construction) with `rclcpp::get_logger(string)`:

http://docs.ros2.org/crystal/api/rclcpp/namespacerclcpp.html#ae7295751947c08312aa69f45fd673171
```

> or would, say, `rclcpp::get_logger("moveit")` suffice for a non-ROS part of the code?

You could just do that, but it would probably be better to extend a node logger with `moveit`, so that if you have multiple nodes using moveit you can disambiguate two similar log messages.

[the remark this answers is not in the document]
davetcoleman: > I guess you mean just node, not ROS, in which case that's ok, I just assumed since you were including rclcpp you were making nodes too.

We decided in moveit1 to use the same unified logging approach throughout the whole code base for simplicity and better use experience at the command line. However the ideal of moveit_core is to not have any pub/sub components, which is nice compartmentalization. Not sure what the best approach is for this in ros2.

> then it could pass that logger into moveit_core's initialization and moveit_core could use that passed in logger as it's logger (or as the parent to a new child logger).

This feels a lot different than ros1... was the logger / node a global variable in ROS1 and now its not in ROS2? This will affect all our class constructors...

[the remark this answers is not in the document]
davetcoleman: @wjwwood would you be willing to prototype your suggested change for moveit in a ``WIP`` PR? Suggested files to look at:
https://github.com/ros-planning/moveit2/blob/master/moveit_core/robot_model/src/robot_model.cpp#L93

https://github.com/ros-planning/moveit2/blob/master/moveit_core/logging/include/moveit/logging/logging.h#L43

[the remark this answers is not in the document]
wjwwood: > This feels a lot different than ros1... was the logger / node a global variable in ROS1 and now its not in ROS2? This will affect all our class constructors...

I tried to explain why here:

> Also, it's worth saying, that we don't provide an "unnamed" macro in rclcpp because unlike ROS 1's `ROS_ERROR`, in ROS 2 there's no singleton node who's name can be used with it. So having `RCLCPP_ERROR` with no logger/name as an argument would simply have no name or node attributed to it, unlike ROS 1. So it would be misleading/incorrect to use that as the alternative to ROS 1's `ROS_ERROR`. The correct equivalent in ROS 2 is `RCLCPP_ERROR(node->get_logger(), "my message");`.

-- https://github.com/ros-planning/moveit2/pull/7#discussion_r258703338

As for you changing constructors, I assume there's a code data structure that everything else inherits from, like a `parent` in Qt? If so then you can simply add it as an option to the constructor of that, and all the children can get it from the parent.

This is how it's structured in the Node, all children can get the logger name by asking the node their associated with (which they have a reference to already).

As for negatively impacting the constructors, asking the parent (assuming they already have a reference to the parent) will eliminate the need to change the constructor of most classes. For the other cases, I changing to use an "options structure/class" so that you can add settings in the future without having to modify the constructor's signature. For example, we just did this with Node:

https://github.com/ros2/rclcpp/pull/622

> https://github.com/ros-planning/moveit2/blob/master/moveit_core/robot_model/src/robot_model.cpp#L93

This is going to be a kind of poor logging experience, since you're going to get log output like:

```
[INFO] [robot_model] Loading robot model 'Foo'...
```

Which is fine unless you have multiple instances running. You'd prefer something like:

```
[INFO] [simulated_robot1.my_moveit_node.robot_model] Loading robot model 'Foo'...
[INFO] [simulated_robot2.my_moveit_node.robot_model] Loading robot model 'Foo'...
```

Which would make it easier to filter and stuff like that.

At the very least I'd recommend changing the `LOGNAME` from `robot_model` to something like `moveit.robot_model` as `robot_model` is very generic.

----

Again, I don't know the internal structure of moveit well enough, but if moveit is designed to only ever be run once per process (and you wish to keep it that way) you can use a global to store the logger you want to use. That would also prevent you from needing to pass it down through all your classes that do logging.

----

In the end, this is all subjective, you guys should look at the log output and adjust it until it's what you want for your users. I think ROS 2 gives you all the tools you need to do that, but it might require you to do something like pass the logger name down or something like that.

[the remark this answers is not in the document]
davetcoleman: > I assume there's a code data structure that everything else inherits from, like a `parent` in Qt?

There is not :-/

> but if moveit is designed to only ever be run once per process (and you wish to keep it that way) you can use a global to store the logger you want to use.

+1 to this approach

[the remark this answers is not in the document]
wjwwood: My wording was kind of bad in that quote, what I meant was more like:

> I assume there's a _core_ data structure from which everything else is created and which owns all of those things, like `parent` in Qt.

I think the answer is still that this assumption was wrong, but I just wanted to be clear.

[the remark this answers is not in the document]
wjwwood: Which reading it one more time is also inaccurate... oh boy. One more time:

> I assume there's a _core_ data structure from which everything else is created and which owns all of those things, like `Node` in ROS 2, which is used to create things like `Publisher`'s, so the publisher has access to the node that created it.

> Or maybe like Qt, where you pass the parent in to each "sub-entity" but it does not own it.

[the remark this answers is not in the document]
davetcoleman: MoveGroup is the only thing i can think of that reminds me of what you maybe saying...

moveit_core/exceptions/CMakeLists.txt
vmayoral: https://github.com/ros-planning/moveit2/pull/7/commits/dcfdc76a25da65d7a5c6f5272546d0bc72e28ba3

moveit_core/exceptions/src/exceptions.cpp
vmayoral: https://github.com/ros-planning/moveit2/pull/7/commits/9f9e4cf134102474a5087cec71781f60020332b5

vmayoral:
Addressed some of the concerns expressed above regarding dead code. I'd suggest we follow the recommendations of @wjwwood in another PR since it'll require changes in several of the submodules of moveit_core.

Connected to https://github.com/ros-planning/moveit2/issues/21.

moveit_core/exceptions/src/exceptions.cpp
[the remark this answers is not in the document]
vmayoral: See https://github.com/ros-planning/moveit2/issues/21#issuecomment-466765096

moveit_core/exceptions/CMakeLists.txt
```diff
@@ -1,11 +1,10 @@
 set(MOVEIT_LIB_NAME moveit_exceptions)
 
 add_library(${MOVEIT_LIB_NAME} src/exceptions.cpp)
-set_target_properties(${MOVEIT_LIB_NAME} PROPERTIES VERSION "${${PROJECT_NAME}_VERSION}")
```
davetcoleman: please revert this change

moveit_core/exceptions/CMakeLists.txt
```diff
@@ -1,11 +1,10 @@
...
-target_link_libraries(${MOVEIT_LIB_NAME} ${catkin_LIBRARIES} ${urdfdom_LIBRARIES} ${urdfdom_headers_LIBRARIES} ${Boost_LIBRARIES})
+target_link_libraries(${MOVEIT_LIB_NAME} ${rclcpp_LIBRARIES} ${rmw_implementation_LIBRARIES} ${urdfdom_LIBRARIES} ${urdfdom_headers_LIBRARIES} ${Boost_LIBRARIES})
 
 install(TARGETS ${MOVEIT_LIB_NAME}
-        LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-        ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION})
-install(DIRECTORY include/ DESTINATION ${CATKIN_GLOBAL_INCLUDE_DESTINATION})
+        LIBRARY DESTINATION lib
+        ARCHIVE DESTINATION lib)
+install(DIRECTORY include/ DESTINATION include)
```
davetcoleman: this kind of changes would be ideal to do across the code base at once

moveit_core/exceptions/src/exceptions.cpp
```diff
@@ -35,14 +35,17 @@
...
-#include <ros/console.h>
+#include <rcutils/logging_macros.h>
+#include "rclcpp/rclcpp.hpp"
+
+#define ROS_ERROR RCUTILS_LOG_ERROR
 
 moveit::ConstructException::ConstructException(const std::string& what_arg) : std::runtime_error(what_arg)
 {
-  ROS_ERROR_NAMED("exceptions", "Error during construction of object: %s\nException thrown.", what_arg.c_str());
+  ROS_ERROR("Error during construction of object: %s\nException thrown.", what_arg.c_str());
```
davetcoleman: this is a regression from moveit1 - we want to keep the namespaces

moveit_core/exceptions/CMakeLists.txt
vmayoral: https://github.com/ros-planning/moveit2/pull/7/commits/06c07e054e551f759a81dbbe2c430be6c20a0100

moveit_core/exceptions/src/exceptions.cpp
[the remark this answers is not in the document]
vmayoral: https://github.com/ros-planning/moveit2/pull/7/commits/7138f303bd5900532f0fabd9d19933013d164a0e

moveit_core/exceptions/src/exceptions.cpp
vmayoral: Logging changed https://github.com/ros-planning/moveit2/pull/7/commits/7138f303bd5900532f0fabd9d19933013d164a0e

mlautman:
@henningkayser Can you give this a 2nd review?

moveit_core/exceptions/src/exceptions.cpp
```diff
@@ -35,14 +35,18 @@
...
 moveit::ConstructException::ConstructException(const std::string& what_arg) : std::runtime_error(what_arg)
 {
-  ROS_ERROR_NAMED("exceptions", "Error during construction of object: %s\nException thrown.", what_arg.c_str());
+  RCLCPP_ERROR(logger_exceptions, "Error during construction of object: %s\nException thrown.", what_arg.c_str());
 }
 
 moveit::Exception::Exception(const std::string& what_arg) : std::runtime_error(what_arg)
 {
-  ROS_ERROR_NAMED("exceptions", "%s\nException thrown.", what_arg.c_str());
+  RCLCPP_ERROR(logger_exceptions, "Exception thrown.", what_arg.c_str());
```
henningkayser: Why is "%s\n" removed from the error message? This way the exception is not printed to the log

moveit_core/exceptions/src/exceptions.cpp
```diff
@@ -35,14 +35,18 @@
 /* Author: Acorn Pooley, Ioan Sucan */
 
 #include <moveit/exceptions/exceptions.h>
-#include <ros/console.h>
+#include <rcutils/logging_macros.h>
+#include "rclcpp/rclcpp.hpp"
+
+// Logger
+rclcpp::Logger logger_exceptions = rclcpp::get_logger("exceptions");
```
henningkayser:
```suggestion
static const rclcpp::Logger LOGGER = rclcpp::get_logger("exceptions");
```
(see https://github.com/ros-planning/moveit2/pull/33)

moveit_core/exceptions/src/exceptions.cpp
vmayoral: That's right, thanks for the catch @henningkayser https://github.com/ros-planning/moveit2/pull/7/commits/b329df8f4a41b061a55a138f03e942bfb02bc1d3

moveit_core/exceptions/src/exceptions.cpp
vmayoral: I'm ok with this change but it'll require further changes in the file (those places where logger_exceptions is used will also need to be renamed).

vmayoral: Maybe, shall we leave this change aside for now and make a global change if we decide to use `LOGGER`?

vmayoral:
Except for https://github.com/ros-planning/moveit2/pull/7#discussion_r265255169m, are we good to go here?

moveit_core/exceptions/src/exceptions.cpp
henningkayser: I'm fine with this!

PR #8, opened 2019-02-19
Port kinematics_base submodule moveit_core

vmayoral:


moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -149,9 +151,9 @@ class KinematicsBase
   static const double DEFAULT_SEARCH_DISCRETIZATION; /* = 0.1 */
   static const double DEFAULT_TIMEOUT;               /* = 1.0 */
 
-  /** @brief Signature for a callback to validate an IK solution. Typically used for collision checking. */
-  typedef boost::function<void(const geometry_msgs::Pose& ik_pose, const std::vector<double>& ik_solution,
-                               moveit_msgs::MoveItErrorCodes& error_code)>
+  /** @brief The signature for a callback that can compute IK */
```
davetcoleman: documentation regression, please restore

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -612,33 +614,74 @@ class KinematicsBase
   template <typename T>
   inline bool lookupParam(const std::string& param, T& val, const T& default_val) const
   {
-    ros::NodeHandle pnh("~");
-    if (pnh.hasParam(group_name_ + "/" + param))
-    {
-      val = pnh.param(group_name_ + "/" + param, default_val);
-      return true;
+    // ros::NodeHandle pnh("~");
```
davetcoleman: remove

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -612,33 +614,74 @@ class KinematicsBase
...
+    //   return true;
+    // }
+    //
+    // if (nh.hasParam("robot_description_kinematics/" + param))
+    // {
+    //   val = nh.param("robot_description_kinematics/" + param, default_val);
+    //   return true;
+    // }
+    //
+    // val = default_val;
```
davetcoleman: remove or add TODO

moveit_core/kinematics_base/CMakeLists.txt
```diff
@@ -1,18 +1,38 @@
...
 
 install(TARGETS ${MOVEIT_LIB_NAME}
-        LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-        ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION})
+        LIBRARY DESTINATION lib
+        ARCHIVE DESTINATION lib)
 
-install(DIRECTORY include/ DESTINATION ${CATKIN_GLOBAL_INCLUDE_DESTINATION})
+install(DIRECTORY include/
+				DESTINATION include)
```
mlautman: Please use spaces instead of tabs

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
mlautman: remove dead code

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -612,33 +614,74 @@ class KinematicsBase
...
   inline bool lookupParam(const std::string& param, T& val, const T& default_val) const
   {
-    ros::NodeHandle pnh("~");
-    if (pnh.hasParam(group_name_ + "/" + param))
-    {
-      val = pnh.param(group_name_ + "/" + param, default_val);
-      return true;
+    // ros::NodeHandle pnh("~");
+
+    auto node = rclcpp::Node::make_shared("lookupparam");
```
mlautman: use underscores for node names to increase readability

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -612,33 +614,74 @@ class KinematicsBase
...
+
+    auto parameters_lookup = std::make_shared<rclcpp::SyncParametersClient>(node);
+
+    auto groupname_param = parameters_lookup->get_parameters({group_name_ + "/" + param});
+
+    for (auto & parameter : groupname_param) {
+        if (!parameter.get_name().compare(group_name_ + "/" + param)) {
```
mlautman: Please clang format these PR's so that we don't have a huge clang PR later. We use the following bash function to automatically run `clang-format-3.9` on the entire moveit repo.

```bash!
# Setup:
#   sudo apt install clang-format-3.9

function console_red {
    echo -n -e "\e[00;31m"
}

function console_nored {
    echo -n -e "\e[00m"
}

alias clang-format="clang-format-3.9"

function clang_this_directory_custom # recursive
{
    # Make sure no changes have occured in repo
    if ! git diff-index --quiet HEAD --; then
        # changes
        console_red
        read -p "You have uncommitted changes, are you sure you want to continue? (y/n)" resp
        console_nored

        if [ "$resp" = "y" ]; then
            echo "Formatting..."
        else
            return -1
        fi
    fi

    find . -name '*.h' -or -name '*.hpp' -or -name '*.cpp' -o -name '*.c' -o -name '*.cc' | xargs clang-format-3.9 -i -style=file $1
}
```

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -612,33 +614,74 @@ class KinematicsBase
...
+
+    for (auto & parameter : robot_description_kinematics_param) {
+        if (!parameter.get_name().compare("robot_description_kinematics/" + param)) {
+          val = parameter.value_to_string();
+          return true;
+        }
     }
 
-    val = default_val;
+    // if (pnh.hasParam(group_name_ + "/" + param))
```
mlautman: Remove dead code before merging

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -658,6 +701,6 @@ class KinematicsBase
 private:
   std::string removeSlash(const std::string& str) const;
 };
-};
+}
```
mlautman: Why is this no longer needed?

moveit_core/kinematics_base/src/kinematics_base.cpp
```diff
@@ -112,15 +112,15 @@ bool KinematicsBase::initialize(const std::string& robot_description, const std:
     return initialize(robot_description, group_name, base_frame, tip_frames[0], search_discretization);
   }
 
-  ROS_ERROR_NAMED(LOGNAME, "This solver does not support multiple tip frames");
+  ROS_ERROR_NAMED(LOGNAME.c_str(), "This solver does not support multiple tip frames");
```
mlautman: Perhaps it would be better to save LOGNAME as a C string to begin with or to change all of the logging to use the new ROS2 logging functionality.

davetcoleman: Total guess here, but this logging functionality is likely all C code...

moveit_core/kinematics_base/CMakeLists.txt
```diff
@@ -1,18 +1,38 @@
...
 
 # This line is needed to ensure that messages are done being built before this is built
-add_dependencies(${MOVEIT_LIB_NAME} ${catkin_EXPORTED_TARGETS})
+# add_dependencies(${MOVEIT_LIB_NAME} ${catkin_EXPORTED_TARGETS})
+
+
+target_link_libraries(${MOVEIT_LIB_NAME} ${rclcpp_LIBRARIES} ${rmw_implementation_LIBRARIES} ${urdfdom_LIBRARIES} ${urdfdom_headers_LIBRARIES} ${Boost_LIBRARIES} ${common_interfaces_LIBRARIES})
+ament_target_dependencies(${MOVEIT_LIB_NAME}
+  urdf
+  urdfdom_headers)
```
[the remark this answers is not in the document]
davetcoleman: @mlautman ?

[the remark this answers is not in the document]
mlautman: Ament/Colcon is A LOT more free-form than Catkin which is very frustrating from the perspective of new users. I am not very knowledgeable about cmake but this would be a meaningful conversation to have with @dirk-thomas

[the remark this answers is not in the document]
wjwwood: You don't need `target_link_libraries` _and_ `ament_target_dependencies`. For each dependency given to `ament_target_dependencies` it will (attempt) to extract the libraries and other settings and add them to the target given, see:

https://github.com/ament/ament_cmake/blob/409ee0403d8d1eba0dc5a58cf52dcec5eadc91d2/ament_cmake_target_dependencies/cmake/ament_target_dependencies.cmake#L15-L30

> Ament/Colcon is A LOT more free-form than Catkin

What does free-form mean exactly?

If you guys want to type less lines, at the expense of precision, you can try `ament_cmake_auto`:

https://github.com/ament/ament_cmake/tree/master/ament_cmake_auto

Unfortunately we haven't had time to add examples and details documentation, but it will do things like automatically add the include dirs and link libraries to your topics based on the dependencies in your `package.xml`.

I had a prototype of something I called `ros_cmake_auto` (`ament_cmake_auto` + ROS specific stuff) that used it (indirectly) here:

https://github.com/wjwwood/ros_cmake_auto_examples

Example:

https://github.com/wjwwood/ros_cmake_auto_examples/blob/master/cpp_examples/node_with_publisher/CMakeLists.txt

And using `ament_cmake_auto` here: 

- https://github.com/wjwwood/ros_cmake_auto_examples/blob/532f56417db60052cfb46174cae8cbfeae6105a2/ros_cmake_auto/cmake/ros_cmake_auto_add_node.cmake#L28-L31
- https://github.com/wjwwood/ros_cmake_auto_examples/blob/532f56417db60052cfb46174cae8cbfeae6105a2/ros_cmake_auto/cmake/ros_cmake_auto_end.cmake#L4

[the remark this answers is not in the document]
dirk-thomas: The [docblock of `ament_target_dependencies`](https://github.com/ament/ament_cmake/blob/207a07b72af1e5a780887d7c2268df57ef456af3/ament_cmake_target_dependencies/cmake/ament_target_dependencies.cmake#L15-L30) describes what it is doing. It simply provides a convenience function so that you don't have to call `target_include_directories`, `target_link_libraries`, etc. separately. It only works when the dependency defines standardized variable names (e.g. `xxx_INCLUDE_DIRS`, `xxx_LIBRARIES`).

So calling `target_link_libraries` and `ament_target_dependencies` for the same dependency is redundant.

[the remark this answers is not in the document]
wjwwood: I don't really recommend the simplified interfaces for a core library like moveit, however. I think it should be precise and careful about what it links against what and what it exports to other packages.

I'd recommend learning cmake and what all the parts do (even if you find it frustrating) because they have a purpose.

moveit_core/kinematics_base/src/kinematics_base.cpp
wjwwood: If you use the `RCLCPP_ERROR` version it will take a `std::string`. The `RCUTILS_ERROR` macro is pure C, so it cannot take a C++ type.

See: https://github.com/ros-planning/moveit2/pull/7#discussion_r258702364

moveit_core/kinematics_base/CMakeLists.txt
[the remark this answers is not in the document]
mlautman: @wjwwood @dirk-thomas I appreciate all of the additional information. I understand that these tools and documentation are still a work in progress. By "free-form" I only mean that as of yet best practices have not yet emerged and finding examples to use as boilerplate can be difficult. 

<quick_rant>
One of ROS1's strengths is that catkin makes it very simple to create a ROS1 application. The `catkin_create_pkg` tool provides users with a heavily commented package that works out of the box and establishes best practices that covers 99% of use cases. By contrast, `ros2 pkg create` produces a bare-bones package with minimal documentation that doesn't really do much. I fully understand that time is limited but speaking as a ROS1 developer who is migrating to ROS2, I can't overstate how important documentation and best practices are to driving adoption and community building.
</quick_rant>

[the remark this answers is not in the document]
dirk-thomas: As always this boils down to resources. Someone needs to spend a significant amount of time and resources to write good tutorials as we are used to from ROS 1. And since we are fairly resource constraint we can't do everything which would be nice to have/do at the same time. As a consequence we have been focusing our limited resources to develop the necessary features and functionality so that ROS 2 provides all the necessary foundations available in ROS 1 (pub/sub, services, actions, C++, Python, command line tool, rviz, rqt, components, etc.). At some point we will hopefully be able to spend more effort on documentation - it is just something nobody considers paying for :wink:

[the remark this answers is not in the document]
wjwwood: I think part of the thinking was also that "you can do so many things, that it would be cluttered to put them all in one template file", so there is also this repository which has full package examples for many different cases, but it too needs documentation and more examples:

https://github.com/ros2/examples/tree/master/rclcpp

The above is sort of like what other projects call a "cookbook".

There's also a repository for tutorials, which are intended to be more heavily documented (commented) and less varied, but geared more for teaching basics:

https://github.com/ros2/tutorials

But they are sadly out of date and we don't even test them regularly (we build the aforementioned examples on every new pr however).

As dirk said, we have plenty of ideas how to improve on-boarding, but it's always behind feature development in priorities.

moveit_core/kinematics_base/CMakeLists.txt
vmayoral: Thanks, 791fee916e5ba796962c3477483968ee05e9ef39

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
vmayoral: Cool, thanks a415ab5ef046bb2081f4a967a083ae651c2e618c

moveit_core/kinematics_base/src/kinematics_base.cpp
vmayoral: 43a77cf2ad36201607eeee8f780bc853ad42827a

vmayoral:
@mlautman and @davetcoleman please have a look again at this PR. 

I've opted for not using `LOGGER` for now since as is, it'll need to be defined in the header causing conflicts with other submodules. Alternatively, we can change a bit the distribution of the code and push the implementation of `searchPositionIK` and other functions (now implemented in the header file) to the cpp file. Let me know if this approach is acceptable and I'll make the changes.

moveit_core/kinematics_base/CMakeLists.txt
```diff
@@ -1,18 +1,19 @@
+cmake_minimum_required(VERSION 3.5)
 set(MOVEIT_LIB_NAME moveit_kinematics_base)
 
 add_library(${MOVEIT_LIB_NAME} src/kinematics_base.cpp)
-set_target_properties(${MOVEIT_LIB_NAME} PROPERTIES VERSION "${${PROJECT_NAME}_VERSION}")
```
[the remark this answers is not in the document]
davetcoleman: this should be kept

moveit_core/kinematics_base/CMakeLists.txt
```diff
@@ -1,18 +1,19 @@
...
+
 # Avoid warnings about deprecated members of KinematicsBase when building KinematicsBase itself.
 # TODO: remove when deperecated variables (tip_frame_, search_discretization_) are finally removed.
```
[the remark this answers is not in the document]
davetcoleman: +1

moveit_core/kinematics_base/CMakeLists.txt
```diff
@@ -1,18 +1,19 @@
...
-target_compile_options(${MOVEIT_LIB_NAME} PRIVATE -Wno-deprecated-declarations)
+# target_compile_options(${MOVEIT_LIB_NAME} PRIVATE -Wno-deprecated-declarations)
 
 # This line is needed to ensure that messages are done being built before this is built
-add_dependencies(${MOVEIT_LIB_NAME} ${catkin_EXPORTED_TARGETS})
-
-target_link_libraries(${MOVEIT_LIB_NAME} ${catkin_LIBRARIES} ${urdfdom_LIBRARIES} ${urdfdom_headers_LIBRARIES} ${Boost_LIBRARIES})
+# add_dependencies(${MOVEIT_LIB_NAME} ${catkin_EXPORTED_TARGETS})
```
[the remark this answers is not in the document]
davetcoleman: let's delete it entirely rather than comment out.

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -37,12 +37,11 @@
 #ifndef MOVEIT_KINEMATICS_BASE_KINEMATICS_BASE_
 #define MOVEIT_KINEMATICS_BASE_KINEMATICS_BASE_
 
-#include <geometry_msgs/PoseStamped.h>
-#include <moveit_msgs/MoveItErrorCodes.h>
+#include <geometry_msgs/msg/pose_stamped.hpp>
+#include <moveit_msgs/msg/move_it_error_codes.hpp>
```
davetcoleman: ``move_it`` is always treated as one word: ``moveit``. for example, the string right before it says ``moveit_msgs`` not ``move_it_msgs``

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -53,12 +52,15 @@ namespace core
...
-}
-}
+}  // namespace core
+}  // namespace moveit
 
 /** @brief API for forward and inverse kinematics */
 namespace kinematics
 {
+// Logger
+rclcpp::Logger logger_kinematics_base = rclcpp::get_logger("kinematics_base");
```
davetcoleman: please let's stick with ``logger``

(or ``static const rclcpp::Logger LOGGER``)

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
vmayoral: @davetcoleman, can't do that. This is imposed by the transformation from .msgs to header files and I doubt @wjwwood and other would accept a change in the core ros2 libraries only for this. Simplest way of doing what you suggest would require renaming `MoveItErrorCodes.msg` into `MoveitErrorCodes.msg`. I'm ok with doing so but that's outside of the scope of this PR.

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
vmayoral: When using the form you recommend:
```bash
/Users/victor/ros2_moveit_ws/src/moveit2/moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h:317:5: error: static_assert failed due to requirement '::std::is_same<typename std::remove_reference<decltype(kinematics::LOGGER)>::type, typename ::rclcpp::Logger>::value' "First argument to logging macros must be an rclcpp::Logger"
    RCLCPP_ERROR(kinematics::LOGGER, "This kinematic solver does not support searchPositionIK with multiple poses");
```

I didn't dive very deep into the problem and instead partially followed your comment at 8b4f35c57334d4d18025919e336b62e0daf80b45.

vmayoral:
@davetcoleman and @mlautman, consider again this PR

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
wjwwood: I was never really in favor of the implicit transformation myself, but that is what's there now and there was a long discussion about it at the time (so it wasn't picked arbitrarily). So I agree that it would be difficult to decide to change it now.

Perhaps a more likely proposal would be to generate both patterns (e.g. `move_it_error_codes.hpp` and `MoveItErrorCodes.hpp`), so you could use which ever you like.

I will say though, just in my opinion, that `MoveitErrorCodes.msg` looks more correct to me than `MoveItErrorCodes.msg`, though that's completely subjective.

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -612,34 +615,55 @@ class KinematicsBase
...
-    {
-      val = pnh.param(group_name_ + "/" + param, default_val);
-      return true;
-    }
+    auto node = rclcpp::Node::make_shared("lookup_param");
 
-    if (pnh.hasParam(param))
+    auto parameters_lookup = std::make_shared<rclcpp::SyncParametersClient>(node);
+
+    auto groupname_param = parameters_lookup->get_parameters({ group_name_ + "/" + param });
```
henningkayser: `auto` is nice where the type is obvious but in cases like this it's more confusing then helpful.
Especially since the following for loops are using `auto` as well this type should be made explicit.

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -612,34 +615,55 @@ class KinematicsBase
...
+      if (!parameter.get_name().compare(group_name_ + "/" + param))
+      {
+        val = parameter.value_to_string();
+        return true;
+      }
     }
 
-    ros::NodeHandle nh;
-    if (nh.hasParam("robot_description_kinematics/" + group_name_ + "/" + param))
+    auto only_param = parameters_lookup->get_parameters({ param });
```
henningkayser: same

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
henningkayser: Also, the `SyncParametersClient` class contains the functions `has_param()` and `get_parameter()` which can be used to replace `hasParam()` and `param()` directly. This way we don't have to loop and compare the results as well.

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -612,34 +615,55 @@ class KinematicsBase
...
-    ros::NodeHandle pnh("~");
-    if (pnh.hasParam(group_name_ + "/" + param))
-    {
-      val = pnh.param(group_name_ + "/" + param, default_val);
-      return true;
-    }
+    auto node = rclcpp::Node::make_shared("lookup_param");
 
-    if (pnh.hasParam(param))
+    auto parameters_lookup = std::make_shared<rclcpp::SyncParametersClient>(node);
```
henningkayser:
```suggestion
    rclcpp::SyncParametersClient parameters_lookup(node);
```
I don't see a reason for using a shared pointer here.

moveit_core/kinematics_base/src/kinematics_base.cpp
```diff
@@ -112,17 +110,18 @@ bool KinematicsBase::initialize(const std::string& robot_description, const std:
     return initialize(robot_description, group_name, base_frame, tip_frames[0], search_discretization);
   }
 
-  ROS_ERROR_NAMED(LOGNAME, "This solver does not support multiple tip frames");
+  RCLCPP_ERROR(kinematics::LOGGER, "This solver does not support multiple tip frames");
```
henningkayser:
```suggestion
  RCLCPP_ERROR(LOGGER, "This solver does not support multiple tip frames");
```
We're already inside the `kinematics` namespace

moveit_core/kinematics_base/src/kinematics_base.cpp
```diff
@@ -112,17 +110,18 @@ bool KinematicsBase::initialize(const std::string& robot_description, const std:
...
 }
 
 bool KinematicsBase::initialize(const moveit::core::RobotModel& robot_model, const std::string& group_name,
                                 const std::string& base_frame, const std::vector<std::string>& tip_frames,
                                 double search_discretization)
 {
-  ROS_WARN_NAMED(LOGNAME, "IK plugin for group '%s' relies on deprecated API. "
-                          "Please implement initialize(RobotModel, ...).",
-                 group_name.c_str());
+  RCLCPP_WARN(kinematics::LOGGER,
```
henningkayser: same

moveit_core/kinematics_base/src/kinematics_base.cpp
```diff
@@ -205,14 +204,14 @@ bool KinematicsBase::getPositionIK(const std::vector<geometry_msgs::Pose>& ik_po
 
   if (ik_poses.size() != 1)
   {
-    ROS_ERROR_NAMED(LOGNAME, "This kinematic solver does not support getPositionIK for multiple tips");
+    RCLCPP_ERROR(kinematics::LOGGER, "This kinematic solver does not support getPositionIK for multiple tips");
```
henningkayser: same

moveit_core/kinematics_base/src/kinematics_base.cpp
```diff
@@ -205,14 +204,14 @@ bool KinematicsBase::getPositionIK(const std::vector<geometry_msgs::Pose>& ik_po
...
     result.kinematic_error = KinematicErrors::MULTIPLE_TIPS_NOT_SUPPORTED;
     return false;
   }
 
   if (ik_poses.empty())
   {
-    ROS_ERROR_NAMED(LOGNAME, "Input ik_poses array is empty");
+    RCLCPP_ERROR(kinematics::LOGGER, "Input ik_poses array is empty");
```
henningkayser: same

moveit_core/kinematics_base/CMakeLists.txt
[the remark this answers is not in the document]
mlautman: Please delete this commented out line

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -311,7 +314,7 @@ class KinematicsBase
     }
 
     // Otherwise throw error because this function should have been implemented
-    ROS_ERROR_NAMED("kinematics_base", "This kinematic solver does not support searchPositionIK with multiple poses");
+    RCLCPP_ERROR(kinematics::LOGGER, "This kinematic solver does not support searchPositionIK with multiple poses");
```
mlautman: We are already in the namespace

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -435,8 +438,8 @@ class KinematicsBase
   virtual const std::string& getTipFrame() const
   {
     if (tip_frames_.size() > 1)
-      ROS_ERROR_NAMED("kinematics_base", "This kinematic solver has more than one tip frame, "
-                                         "do not call getTipFrame()");
+      RCLCPP_ERROR(kinematics::LOGGER, "This kinematic solver has more than one tip frame, "
```
mlautman: We are already in the namespace

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
vmayoral: I agree we should probably just rename the message to `MoveitErrorCodes.msg`.

I raised an issue in our fork at https://github.com/AcutronicRobotics/moveit_msgs/issues/4. I'm resolving this comment for now.

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
vmayoral: Thanks for the input @henningkayser. While I don't fully agree in this case (I think it's pretty obvious from the constructor) I've accepted your suggestion in a commit that required further changes 7f706b5cbc17b8f48f223beaf3ade7615733396f

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
vmayoral: 7f706b5cbc17b8f48f223beaf3ade7615733396f

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
vmayoral: Partially complied with your request @henningkayser at b2d436f024d1003ea57d9f8f6ff88ddde0f3c9a1.

Please make a suggestion if you require further changes.

vmayoral:
Please advice if we're good to go here

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
henningkayser: `get_parameters()` is meant to be used for multiple parameters but here it's always called with only a single entry. In this case it's much more straightforward to query every parameter like in the sample code below:
```C++
    bool param_success = node.get_parameter<T>(group_name + "/" + param, val);
```

No need to create a parameter vector or loop over the results and also we can use the node directly and don't need to cast it to `rclcpp::SyncParametersClient`.

Since `lookupParam()` does in fact query multiple parameters in different locations we can absolutely use `Node::get_parameters()` but in that case including all parameter names:
```C++
val = default_val;
std::vector<std::string> param_names = {
  group_name_ + "/" + param, param,
  "robot_description_kinematics/" + group_name_ + "/" + param,
  robot_description_kinematics/" + param
};
std::vector<rclcpp::parameter::ParameterVariant> param_results =
  node.getParameters(param_names);
bool param_success = !param_results.empty();
if (param_success)
    val = param_results[0].get_value<T>();
return param_success;
```

I actually prefer the second implementation.
Please also note that your current implementation calls `value_to_string()` on all parameters but `lookupParam()` is a template function that supports all parameter types. The `get_value()` function takes care of using the correct type.

anasarrak: I've already made some changes (on planning_scene_monitor) related to this:

https://github.com/AcutronicRobotics/moveit2/blob/5a8ad535e33548434b02e5e9e735ab614465f5a5/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp#L1472-L1516

henningkayser: @anasarrak this looks much better. However, each if/else block can be implemented in a single line using `node->get_parameter_or(param, val, default_val)`.

henningkayser:
@vmayoral I took the freedom and reimplemented the lookupParam() function as I requested and also resolved the merge conflicts. I think this is ready to be merged now.

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -612,34 +615,17 @@ class KinematicsBase
...
-    if (pnh.hasParam(param))
+    // TODO(henningkayser): reuse node and fix private namespace lookup from kinematics.yaml
+    auto node = rclcpp::Node::make_shared("kinematics_base.lookup_param");
+    std::vector<rclcpp::Parameter> param_results = node->get_parameters(
+        { group_name_ + "/" + param, param, "robot_description_kinematics/" + group_name_ + "/" + param,
+          "robot_description_kinematics/" + param });
+    if (!param_results.empty())
     {
-      val = pnh.param(param, default_val);
+      val = param_results.get_value<T>();
```
mlautman: This is a vector. I think you need to get the parameter at first index right? @henningkayser

mlautman: nit picky but I think this should be the order:
```cpp
 { 
    param,
    group_name_ + "/" + param,  
    "robot_description_kinematics/" + param,
    "robot_description_kinematics/" + group_name_ + "/" + param
}
```

henningkayser: > This is a vector. I think you need to get the parameter at first index right? @henningkayser

@mlautman You're absolutely right. Fixed!

> nit picky but I think this should be the order:
> 
> ```c++
>  { 
>     param,
>     group_name_ + "/" + param,  
>     "robot_description_kinematics/" + param,
>     "robot_description_kinematics/" + group_name_ + "/" + param
> }
> ```

I kept the old order and don't think we should change it.
After all the specification (with group name) should override the default (without group name), no?

mlautman: fair enough. My nit was aesthetic

mlautman:
@henningkayser lgtm

PR #14, opened 2019-02-19
Port robot_trajectory submodule of moveit_core to ROS 2

anasarrak:


anasarrak:
After further testing realized needs more work, will re-open the PR later.

PR #37, opened 2019-03-10
Update CI to ROS2

vmayoral:
Submit changes to enable CI. @LanderU, feel free to add anything in case I missed something.

### DIFF
--- a/.docker/ci/Dockerfile
+++ b/.docker/ci/Dockerfile
@@ -1,50 +1,30 @@
-# moveit/moveit:melodic-ci
-# Sets up a base image to use for running Continuous Integration on Travis
+FROM ros:crystal-ros-base-bionic
 
-FROM ros:melodic-ros-base
-MAINTAINER Dave Coleman dave@picknik.ai
+MAINTAINER Lander Usategui  <lander@erlerobotics.com>
 
-ENV TERM xterm
+ENV ROS_DISTRO crystal
+WORKDIR /ros2_ws
 
-# Setup catkin workspace
-ENV CATKIN_WS=/root/ws_moveit
-WORKDIR $CATKIN_WS
-
-# Commands are combined in single RUN statement with "apt/lists" folder removal to reduce image size
-# https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers
 RUN \
-    mkdir src && \
-    cd src && \
-    #
-    # Download moveit source, so that we can get necessary dependencies
-    wstool init --shallow . https://raw.githubusercontent.com/ros-planning/moveit/${ROS_DISTRO}-devel/moveit.rosinstall && \
-    #
-    # Update apt package list as previous containers clear the cache
-    apt-get -qq update && \
-    apt-get -qq dist-upgrade && \
-    #
-    # Install some base dependencies
-    apt-get -qq install -y \
-        # Some source builds require a package.xml be downloaded via wget from an external location
+    apt-get update && apt-get install -y \
+        python3-colcon-common-extensions \
+        libboost-all-dev \
+        libglew-dev \
+        freeglut3-dev \
+        pkg-config \
+        libfcl-dev \
+        libassimp-dev \
+        libqhull-dev \
+        ros-$ROS_DISTRO-action-msgs \
+        ros-$ROS_DISTRO-resource-retriever \
+        ros-$ROS_DISTRO-image-transport \
+        python3-vcstool \
+        libopencv-dev \
         wget \
-        # Required for rosdep command
-        sudo \
-        # Preferred build tools
-        python-catkin-tools \
-        clang clang-format-3.9 clang-tidy clang-tools \
-        ccache && \
-    #
-    # Download all dependencies of MoveIt!
-    rosdep update && \
-    rosdep install -y --from-paths . --ignore-src --rosdistro ${ROS_DISTRO} --as-root=apt:false && \
-    #
-    # Remove the source code from this container
-    # TODO: in the future we may want to keep this here for further optimization of later containers
-    cd .. && \
-    rm -rf src/ && \
-    #
-    # Clear apt-cache to reduce image size
-    rm -rf /var/lib/apt/lists/*
-
-# Continous Integration Setting
-ENV IN_DOCKER 1
+        && rm -rf /var/lib/apt/lists/* \
+    # Setup ROS2 workspace
+    && mkdir -p /ros2_ws/src && cd /ros2_ws \ 
+    && wget https://raw.githubusercontent.com/AcutronicRobotics/moveit2/master/external-repos.repos \
+    && vcs import src < external-repos.repos
+COPY ./docker-entrypoint.sh /
+ENTRYPOINT ["/docker-entrypoint.sh"]
--- a/.docker/ci/docker-entrypoint.sh
+++ b/.docker/ci/docker-entrypoint.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+set -e
+
+/bin/bash -c "find /opt/ros/crystal/ -name tf2_eigen | xargs rm -rf && source /opt/ros/crystal/setup.bash \
+	      && touch /ros2_ws/src/image_common/camera_calibration_parsers/COLCON_IGNORE \
+              && touch /ros2_ws/src/image_common/camera_info_manager/COLCON_IGNORE \
+              && colcon build --merge-install"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,43 +1,27 @@
-# This config file for Travis CI utilizes https://github.com/ros-planning/moveit_ci/ package.
-sudo: required
-dist: xenial  # distro used by Travis, moveit_ci uses the docker image's distro
-services:
-  - docker
-language: cpp
-cache: ccache
-compiler: gcc
-
 notifications:
   email:
     recipients:
-      - 130s@2000.jukuin.keio.ac.jp
-env:
-  global:
-    - MOVEIT_CI_TRAVIS_TIMEOUT=85  # Travis grants us 90 min, but we add a safety margin of 5 min
-    - ROS_DISTRO=melodic
-    - ROS_REPO=ros
-    - UPSTREAM_WORKSPACE=moveit.rosinstall
-    - TEST_BLACKLIST=moveit_ros_perception  # mesh_filter_test fails due to broken Mesa OpenGL
-    - WARNINGS_OK=false
-  matrix:
-    - TEST="clang-format catkin_lint"
-    - TEST=clang-tidy-fix
-    - ROS_DISTRO=melodic
-    - ROS_DISTRO=kinetic
+      - lander@erlerobotics.com
+      - anasarrak@erlerobotics.com
 
-matrix: # Add a separate config to the matrix, using clang as compiler
+matrix:
   include:
-    # add a config to the matrix using clang as compiler and also test ikfast plugin creation
-    - compiler: clang
-      env: ROS_REPO=ros-shadow-fixed
-           BEFORE_DOCKER_SCRIPT="source moveit_kinematics/test/test_ikfast_plugins.sh"
-
-  fast_finish: true  # finish, even if allow-failure-tests still running
-  allow_failures:
-    - env: TEST=clang-tidy-fix  # need to fix clang-tidy issues
+    - os: linux
+      dist: bionic
+      services:
+        - docker
+      sudo:
+        - required
+    - os: osx
+      osx_image: xcode10.1
 
 before_script:
-  - git clone -q --depth=1 https://github.com/ros-planning/moveit_ci.git .moveit_ci
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then bash ./osx-dependencies.sh; fi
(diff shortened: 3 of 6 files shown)


.docker/ci/Dockerfile
```diff
@@ -1,50 +1,30 @@
...
-    # Update apt package list as previous containers clear the cache
-    apt-get -qq update && \
-    apt-get -qq dist-upgrade && \
-    #
-    # Install some base dependencies
-    apt-get -qq install -y \
-        # Some source builds require a package.xml be downloaded via wget from an external location
+    apt-get update && apt-get install -y \
+        python3-colcon-common-extensions \
+        libboost-all-dev \
```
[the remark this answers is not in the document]
henningkayser: Also, are these really all packages that are required to build moveit_core?

external-repos.repos
```diff
@@ -0,0 +1,61 @@
+repositories:
```
henningkayser: Why is moveit2 not included here? This way travis just compiles the dependencies only.
Also we should rename this file to moveit2.repos for consistency.

.docker/ci/Dockerfile
```diff
@@ -1,50 +1,30 @@
-# moveit/moveit:melodic-ci
-# Sets up a base image to use for running Continuous Integration on Travis
+FROM ros:crystal-ros-base-bionic
 
-FROM ros:melodic-ros-base
-MAINTAINER Dave Coleman dave@picknik.ai
+MAINTAINER Lander Usategui  <lander@erlerobotics.com>
```
henningkayser: I'm not sure if this should change, @davetcoleman ?

external-repos.repos
henningkayser: I think it would be cleaner to separate temporary dependencies to `AcutronicRobotics` and those that are not released yet somehow. Either just with comments or in a separate .repos file.

.docker/ci/Dockerfile
```diff
@@ -1,50 +1,30 @@
...
-
-# Continous Integration Setting
-ENV IN_DOCKER 1
+        && rm -rf /var/lib/apt/lists/* \
+    # Setup ROS2 workspace
+    && mkdir -p /ros2_ws/src && cd /ros2_ws \ 
+    && wget https://raw.githubusercontent.com/AcutronicRobotics/moveit2/master/external-repos.repos \
+    && vcs import src < external-repos.repos
+COPY ./docker-entrypoint.sh /
+ENTRYPOINT ["/docker-entrypoint.sh"]
```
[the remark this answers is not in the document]
mlautman: Agreed, `moveit_ci` has some really nice features that have been developed over time that will really benefit MoveIt2

.docker/ci/Dockerfile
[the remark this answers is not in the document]
mlautman: Some of these should come with the ROS crystal image.

.travis.yml
```diff
@@ -1,43 +1,27 @@
...
+      services:
+        - docker
+      sudo:
+        - required
+    - os: osx
+      osx_image: xcode10.1
 
 before_script:
-  - git clone -q --depth=1 https://github.com/ros-planning/moveit_ci.git .moveit_ci
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then bash ./osx-dependencies.sh; fi
```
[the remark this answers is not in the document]
LanderU: Thank you for your review! AFAIK we cannot build an OS X image on Docker. What do you mean exactly here?

.docker/ci/Dockerfile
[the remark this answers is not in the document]
LanderU: Not really, we've extra dependencies for the rest of the modules.

external-repos.repos
LanderU: > Why is moveit2 not included here? This way travis just compiles the dependencies only.
Also we should rename this file to moveit2.repos for consistency.

+1 for rename the file. I use in this way to mount the repo as a volume on Docker when we build it. This way we can build the code of the current branch. But, I can set the corresponding branch in moveit2.repos for each branch.

.docker/ci/Dockerfile
LanderU: No problem with this, I'll keep Dave as maintainer.

.travis.yml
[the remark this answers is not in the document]
henningkayser: You're right about the docker image. Other than running a VM inside docker there is probably no way to do this. My point is that we should try to use the same setup and build methods accross different platforms. The question is if we want to implement osx support in `moveit_ci`

.docker/ci/Dockerfile
```diff
@@ -1,50 +1,30 @@
-# moveit/moveit:melodic-ci
-# Sets up a base image to use for running Continuous Integration on Travis
+FROM ros:crystal-ros-base-bionic
```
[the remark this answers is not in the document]
LanderU: https://github.com/osrf/docker_images/blob/eb35e27e8a0842c8fb493bbdef37b314f931aa59/ros/crystal/ubuntu/bionic/desktop/Dockerfile#L5-L8

external-repos.repos
henningkayser: I totally see the point in doing it this way. `moveit_ci` already handles the corresponding branch inside [travis.sh](https://github.com/ros-planning/moveit_ci/blob/master/travis.sh) anyway. I think the original reason why the parent repo was included in .rosinstall files is that it makes it easier to setup workspaces. @rhaschke any thoughts on this?

.docker/ci/Dockerfile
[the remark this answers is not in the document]
henningkayser: Yes, we could use the desktop docker, but this would probably install a lot of unnecessary packages as well. I think installing packages by hand is fine, just use rosdep where possible.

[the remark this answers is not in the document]
LanderU: +1

external-repos.repos
rhaschke: I'm not fully sure what you are referring to exactly.
But, yes MoveIt's .rosinstall file comprises all repositories required to build MoveIt from source, i.e. also including MoveIt itself. travis.sh [here](https://github.com/ros-planning/moveit_ci/blob/e05280b6f127a35f2b40f0634fd7d9039a0f7e10/travis.sh#L185-L186) takes care that the to-be-tested repo is removed from the .rosinstall file before initializing the workspace.

.docker/ci/Dockerfile
```diff
@@ -1,50 +1,30 @@
...
+    apt-get update && apt-get install -y \
+        python3-colcon-common-extensions \
+        libboost-all-dev \
+        libglew-dev \
+        freeglut3-dev \
+        pkg-config \
+        libfcl-dev \
+        libassimp-dev \
+        libqhull-dev \
+        ros-$ROS_DISTRO-action-msgs \
```
mlautman: These should be rosdeps.

external-repos.repos
mlautman: @LanderU Is there a reason to switch from `.rosinstall` to `.repos`? A good deal of the MoveIt infrastructure assumes the use of wstool with `.rosinstall` files. We can migrate over to vcs but we would probably want to have good reasons for doing so.

osx-compile.sh
```diff
@@ -0,0 +1,22 @@
...
+find ros2-osx/ -name tf2_eigen | xargs rm -rf
+
+source ros2-osx/setup.bash
+mkdir -p /tmp/ros2_ws/src
+cp -r /tmp/moveit2 /tmp/ros2_ws/src
+cd /tmp/ros2_ws && wget https://raw.githubusercontent.com/AcutronicRobotics/moveit2/master/external-repos.repos
+vcs import src < external-repos.repos
+export OPENSSL_ROOT_DIR="/usr/local/opt/openssl"
+#Ignore packages
+touch src/image_common/camera_calibration_parsers/COLCON_IGNORE
```
mlautman: Can you elaborate on why `camera_calibration_parsers` shouldn't be built in a comment? If this is an OSX issue then we will want to know that we can delete this line when the package is fully ported

.docker/ci/Dockerfile
ruffsl: FYI that field is deprecated:  
https://docs.docker.com/engine/reference/builder/#maintainer-deprecated

.docker/ci/docker-entrypoint.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -e
+
+/bin/bash -c "find /opt/ros/crystal/ -name tf2_eigen | xargs rm -rf && source /opt/ros/crystal/setup.bash \
+	      && touch /ros2_ws/src/image_common/camera_calibration_parsers/COLCON_IGNORE \
```
mlautman: `camera_calibration_parsers` and `camera_info_manager` can be skipped with `--packages-skip camera_info_manager camera_calibration_parsers`

.docker/ci/docker-entrypoint.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -e
+
+/bin/bash -c "find /opt/ros/crystal/ -name tf2_eigen | xargs rm -rf && source /opt/ros/crystal/setup.bash \
```
mlautman: why are we removing tf2_eigen from the source install? This needs good documentation

.travis.yml
```diff
@@ -1,43 +1,27 @@
...
-  - docker
-language: cpp
-cache: ccache
-compiler: gcc
-
 notifications:
   email:
     recipients:
-      - 130s@2000.jukuin.keio.ac.jp
-env:
```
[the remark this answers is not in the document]
mlautman: +1

mlautman:
@vmayoral @LanderU I am having a little trouble getting the AcutronicRobotics master branch to build on my machine. It seems that rosdep install doesn't work yet so I went through the dependencies and fixed all the issues that I found along the way:

https://github.com/AcutronicRobotics/urdf_parser_py/pull/1
https://github.com/ros-planning/srdfdom/pull/48
https://github.com/AcutronicRobotics/octomap/pull/1
https://github.com/AcutronicRobotics/octomap_msgs/pull/7
https://github.com/AcutronicRobotics/moveit_msgs/pull/2
https://github.com/AcutronicRobotics/object_recognition_msgs/pull/5

Now that all of that works on my machine, I am trying to get `colcon build --packages-up-to moveit_core` to run successfully. 

Are there any feature branches that I could be using to make this process go a little smoother?

external-repos.repos
LanderU: We're using .repos because ROS2 is using this kind of file: https://github.com/ros2/ros2/blob/master/ros2.repos

.docker/ci/Dockerfile
LanderU: Thanks @ruffsl!

osx-compile.sh
LanderU: Only because for now is not needed, and because most of the packages on this repo have COLCON_IGNORE. We can keep this as is till have correctly ported to ROS2.

.docker/ci/docker-entrypoint.sh
LanderU: I prefer to have the de `colcon` command as clean as we can, but if you prefer to skip the packages using this way, I'll change it.

.docker/ci/docker-entrypoint.sh
LanderU: The main reason behind this was this: https://github.com/AcutronicRobotics/moveit2/issues/19#issuecomment-469208116

.docker/ci/Dockerfile
davetcoleman: i do not need to be a maintainer of this anymore

.docker/ci/Dockerfile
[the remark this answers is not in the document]
davetcoleman: For moveit1, we only have to manually install 3 packages:

     sudo apt-get install python-wstool python-catkin-tools clang-format-3.9

And clang-format is actually optional. We should be able to do the same for ros2

https://moveit.ros.org/install/source/

.travis.yml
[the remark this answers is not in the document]
davetcoleman: i think one issue here is that two very different features are being combined in the same PR: migrating to ROS2 and also, adding support for OSX. separating these would make discussion much easier

external-repos.repos
davetcoleman: Does wstool + .rosinstall files not also work with ROS2? I think keeping the moveit1 approach, at least for now, will reduce a lot of headache. I think focusing on the minimum changes for ROS2 support would be the best first approach

henningkayser: `wstool` still works since it's pretty much unrelated to any ROS/ROS2 functionality. `vcstools`  is a new alternative to `wstool` that is supported by OSRF and comes with some improved features (https://github.com/vcstools/wstool/issues/19#issuecomment-72942597) which we don't need for migration. However, the import function of `vcstools` supports .rosinstall files (see [README](https://github.com/dirk-thomas/vcstool)) so we should be able to use it as a drop-in replacement.

davetcoleman:
ping @LanderU do you have any advice on how to get the Acutronics master branch compiling?

vmayoral:
I'm closing this on behalf  of the CI work @mlautman has been leading.

PR #38, opened 2019-03-10
Port robot_state to ROS2

vmayoral:
Follows from https://github.com/ros-planning/moveit2/pull/13


moveit_core/robot_state/include/moveit/robot_state/log.h
```diff
@@ -0,0 +1,53 @@
...
+
+#include "rclcpp/rclcpp.hpp"
+
+namespace moveit
+{
+namespace core
+{
+
+// Logger
+rclcpp::Logger logger_robot_state = rclcpp::get_logger("robot_state");
```
vmayoral: Follows previous code guidelines and creates a single logger `robot_state`.

moveit_core/robot_state/src/conversions.cpp
```diff
@@ -39,24 +39,24 @@
 #include <geometric_shapes/shape_operations.h>
 #include <tf2_eigen/tf2_eigen.h>
 #include <boost/lexical_cast.hpp>
+#include <moveit/robot_state/log.h>
```
vmayoral: Uses logger created before. In case it's desired to create the logger here instead, please say it.

moveit_core/robot_state/test/test_aabb.cpp
```diff
@@ -150,32 +150,34 @@ TEST_F(TestAABB, TestPR2)
...
   // Wait for the publishers to establish connections
   sleep(5);
 
   // Prepare the ROS message we will reuse throughout the rest of the function.
-  visualization_msgs::Marker msg;
-  msg.header.frame_id = pr2_state.getRobotModel()->getRootLinkName();
-  msg.type = visualization_msgs::Marker::CUBE;
-  msg.color.a = 0.5;
-  msg.lifetime.sec = 3000;
+  auto msg = std::make_shared<visualization_msgs::Marker>();
```
henningkayser: I'm not sure if it's always safe to reuse this shared pointer for multiple publish calls. Since we don't rely on performance in this test we might just pass the message by reference.

moveit_core/robot_state/src/conversions.cpp
mlautman: I agree with @henningkayser here. Adding a log.h each time we want a logger is overkill. see comment here: https://github.com/ros-planning/moveit2/pull/40#discussion_r264481130

mlautman:
@anasarrak (Since the original PR was done by them)

moveit_core/robot_state/include/moveit/robot_state/robot_state.h
```diff
@@ -950,114 +947,77 @@ as the new values that correspond to the group */
...
-  setFromIK(const JointModelGroup* group, const Eigen::Isometry3d& pose, const std::string& tip,
-            const std::vector<double>& consistency_limits, unsigned int attempts, double timeout = 0.0,
-            const GroupStateValidityCallbackFn& constraint = GroupStateValidityCallbackFn(),
-            const kinematics::KinematicsQueryOptions& options = kinematics::KinematicsQueryOptions())
-  {
-    return setFromIK(group, pose, tip, consistency_limits, timeout, constraint, options);
-  }
 
-  /** \brief  Warning: This function inefficiently copies all transforms around.
+  /** \brief  ing: This function inefficiently copies all transforms around.
```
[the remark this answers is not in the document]
vmayoral: c377881f6590a3134ae1f97e0e9d1664eb221fb4

moveit_core/robot_state/include/moveit/robot_state/robot_state.h
```diff
@@ -950,114 +947,77 @@ as the new values that correspond to the group */
      set by computing inverse kinematics.
       The pose is assumed to be in the reference frame of the kinematic model. Returns true on success.
       @param pose The pose the last link in the chain needs to achieve
+      @param attempts The number of times IK is attempted
       @param timeout The timeout passed to the kinematics solver on each attempt
       @param constraint A state validity constraint to be required for IK solutions */
-  bool setFromIK(const JointModelGroup* group, const geometry_msgs::Pose& pose, double timeout = 0.0,
-                 const GroupStateValidityCallbackFn& constraint = GroupStateValidityCallbackFn(),
+  bool setFromIK(const JointModelGroup* group, const geometry_msgs::msg::Pose& pose, unsigned int attempts = 0,
```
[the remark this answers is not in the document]
vmayoral: @henningkayser can you help addressing this? what's your suggestion? is this something that will get fixed once we _sync_ with moveit1 master? If so, shall we skip your comment for now?

vmayoral:
I've included @davetcoleman's approach from https://github.com/ros-planning/moveit2/issues/21 here. From @henningkayser's comments, it sounds like some sync is required to move forward with this PR. @henningkayser, you seem to be knowledgeable about what should be in or out in this submodule, can you help addressing your comments above?

moveit_core/robot_state/include/moveit/robot_state/robot_state.h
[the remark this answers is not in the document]
vmayoral: Thanks @henningkayser, that makes sense. I recall @rhaschke pushing some stuff accidentally so it may be from that time since this PR is from a while ago.

I've created a new PR at https://github.com/ros-planning/moveit2/pull/59. Can you please have a look there?

vmayoral:
Follows at https://github.com/ros-planning/moveit2/pull/59.

PR #44, opened 2019-03-13
Update Docker build and add CircleCI + CodeCov support

ruffsl:
This PR begins the process updating legacy docker build process and adding support for CircleCI, addressing https://github.com/ros-planning/moveit2/issues/24 . This falls inline as with other ros2 projects adopting alternative CI integrations as Travis inevitably winds down due to it acquisition from Idera: https://news.ycombinator.com/item?id=19218036

Additional motivation and features are expounded upon in navigation2's respective PR:  
https://github.com/ros-planning/navigation2/pull/534

# Action Items

- DockerHub
  - [x] Create the `ros-planning` DockerHub organization (admin)
  - [x] Invite relevant Collaborators to DockerHub org (admin)
  - [ ] Link GitHub `moveit2` automated repo to DockerHub org (admin)
  - [ ] Configure build rules for automated repo (@ruffs)
    - [ ] Add rule for branchs, e.g. `ros-planning/moveit2:master`
    - [ ] Add repo link for Base Image: `osrf/ros2:nightly`

- CircleCI
  - [ ] Enable GitHub integration (admin)
    - https://circleci.com/gh/organizations/ros-planning/settings#projects
  - [ ] Invite relevant Collaborators to GitHub Team (admin)
  - [ ] Configure Advanced Settings for ci repo (@ruffs)
    - [ ] Only build pull requests (**On**)
      - Save build usage by only testing open PRed branches
    - [ ] Auto-cancel redundant builds (**On**)
      - Save build usage by only testing current HEAD of branch
    - [ ] Build forked pull requests (**On**)
      - Looking at repo's history, most merged PRs seem to come from forks

- CodeCov
  - [ ] Enable GitHub integration (admin)
    - https://codecov.io/gh/ros-planning/movit2
  - [ ] Configure CodeCov tokens with CircleCI (@ruffs)
    - [ ] Only push CodeCov results for debug builds

Ping @davetcoleman , @mlautman , @vmayoral  
As of writing, none of the ROS2 packages within this moveit2 repo build yet, so we have nothing to test.


davetcoleman:
@ruffsl thanks for starting this! I need to do further analysis, but the ycombinator article you shared was very interesting.

At today's sync up meeting between Acutronic and PickNik we were debatting if we should continue to use Travis-based [moveit_ci](https://github.com/ros-planning/moveit_ci/). Do we need both? Our moveit_ci provides a lot of cool stuff:

- build lots of repos from source together
- test shadow-fixed and non-shadow fixed
- test on multiple Ubuntu versions
- fail if clang-format doesn't pass
- fail if clang-tidy doesn't pass

Can CircleCI do all of this?

.circleci/config.yml
```diff
@@ -0,0 +1,240 @@
+version: 2.1
+references:
+  pre_checkout: &pre_checkout
+    run:
+        name: Pre Checkout
+        command: |
+          stat -c '%y' /opt/ros/$ROS_DISTRO/setup.sh | \
+            sha256sum | \
+            awk '{print "distro "$1}' > /opt/ros/$ROS_DISTRO/checksum.txt
+          cp $MOVEIT2_WS/src/moveit2/tools/ros2_dependencies.repos $ROS_WS/
```
davetcoleman: there's already a repos file from @LanderU, but it hasn't been added to ``ros-planning`` yet:

https://github.com/AcutronicRobotics/moveit2/blob/master/moveit2.repos

.circleci/config.yml
```diff
@@ -0,0 +1,240 @@
...
+          rm -rf $MOVEIT2_WS/*
+  post_checkout: &post_checkout
+    run:
+        name: Post Checkout
+        command: |
+          rm $MOVEIT2_WS/src/moveit2/nav2_system_tests/COLCON_IGNORE
```
davetcoleman: nav2_system_tests should be removed

.circleci/config.yml
```diff
@@ -0,0 +1,240 @@
...
+          if ! cmp $MOVEIT2_WS/src/moveit2/tools/ros2_dependencies.repos $ROS_WS/ros2_dependencies.repos >/dev/null 2>&1
```
davetcoleman: remove ros2_dependencies

.circleci/config.yml
```diff
@@ -0,0 +1,240 @@
+version: 2.1
+references:
+  pre_checkout: &pre_checkout
+    run:
+        name: Pre Checkout
+        command: |
+          stat -c '%y' /opt/ros/$ROS_DISTRO/setup.sh | \
+            sha256sum | \
+            awk '{print "distro "$1}' > /opt/ros/$ROS_DISTRO/checksum.txt
```
davetcoleman: could you add some inline comments describing what this checksum.txt is about? i see it used a lot in this script

.circleci/config.yml
```diff
@@ -0,0 +1,240 @@
...
+    store_artifacts:
+        path: log/test
+  report_code_coverage: &report_code_coverage
+    run:
+        name: Report Code Coverage
+        command: |
+          if [ "$COVERAGE_ENABLED" = "True" ]
+          then
+            . install/setup.sh
+            src/moveit2/tools/code_coverage_report.bash codecovio
```
davetcoleman: rather than adding another folder (``tools``) let's put this in our scripts folder:
https://github.com/ros-planning/moveit2/tree/master/moveit/scripts

.circleci/config.yml
```diff
@@ -0,0 +1,240 @@
...
+      - *store_test_logs
+  report_coverage:
+    description: "Report Coverage"
+    steps:
+      - *report_code_coverage
+
+executors:
+  docker_exec:
+    docker:
+      - image: rosplanning/moveit2:master
```
davetcoleman: we've been using the ``moveit`` dockerhub for years:
https://hub.docker.com/r/moveit/moveit/

we also just created:
https://hub.docker.com/r/moveit/moveit2

.circleci/config.yml
```diff
@@ -0,0 +1,240 @@
...
+      - build_source
+      - test_source
+
+workflows:
+  version: 2
+  build-test:
+    jobs:
+      - debug_build
+      - release_build
+  # nightly:
```
davetcoleman: add TODO or some other comment for what this is intended for

Dockerfile
```diff
@@ -0,0 +1,82 @@
+# This dockerfile expects proxies to be set via --build-arg if needed
```
davetcoleman: we have a .docker folder:
https://github.com/ros-planning/moveit2/tree/master/.docker

@mlautman has been setting up docker for ROS2:
https://github.com/ros-planning/moveit2/pull/46

tools/code_coverage_report.bash
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
```
davetcoleman: im so excited to have code coverage reports!

.circleci/config.yml
ruffsl: As long as your cool co locating docker tags for users and CI. I just thought you might want to cleanly separate release and CI repos.

Dockerfile
ruffsl: Optimizations of the CI config heavily rely upon the assumed setup of this Dockerfile, so they go hand-in-hand. Perhaps @mlautman could mention what changes should be added into this CI dockerfile?

tools/code_coverage_report.bash
ruffsl: This is borrowed over from navigation2 and https://github.com/ros-planning/navigation2/pull/553 .
Perhaps @crdelsey would have some recommendations for your cmake setup?

.circleci/config.yml
ruffsl: See yaml header.

tools/code_coverage_report.bash
crdelsey: To have a single `COVERAGE_ENABLED` flag passed as a CMake argument, as @ruffsl has done in this PR, you need to add a few lines to every package's top level CMakelist.txt

```CMake
option(COVERAGE_ENABLED "Enable code coverage" FALSE)
  if(COVERAGE_ENABLED)
    add_compile_options(--coverage)
    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} --coverage")
    set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} --coverage")
  endif()
```

Rather than duplicate this code everywhere, we created a package to hold common CMake macros.
https://github.com/ros-planning/navigation2/tree/master/nav2_common

This package defines a macro with our standard package header, and we just call that in each package. So instead of adding all the lines above, we add two lines to each package:

```CMake
find_package(nav2_common)  # for the macro definintion
```
and then after all the `find_package` lines

```CMake
nav2_package()  # to bring in standard package configuration options including COVERAGE_ENABLED
```

.circleci/config.yml
davetcoleman: Is there a good reason we'd want to separate these? We have not for years:
image

.circleci/config.yml
davetcoleman: awesome thanks

.circleci/config.yml
```diff
@@ -0,0 +1,246 @@
+# `checksum.txt` derives the key for the `save_cache` and `restore_cache` steps.
+# As we can't cache onpon the state of the file system directly,
```
davetcoleman:
```suggestion
# As we can't cache upon the state of the file system directly,
```
??

.docker/ci/Dockerfile
```diff
@@ -1,50 +1,82 @@
...
-    # TODO: in the future we may want to keep this here for further optimization of later containers
-    cd .. && \
-    rm -rf src/ && \
-    #
-    # Clear apt-cache to reduce image size
-    rm -rf /var/lib/apt/lists/*
-
-# Continous Integration Setting
-ENV IN_DOCKER 1
+# This dockerfile expects proxies to be set via --build-arg if needed
```
davetcoleman: @mlautman please review

tools/ros2_dependencies.repos
```diff
@@ -0,0 +1,9 @@
+repositories:
```
davetcoleman: can you move this to the root and call it ``moveit2.repos`` so it matches what will be there soon?

https://github.com/AcutronicRobotics/moveit2/blob/master/moveit2.repos

.circleci/config.yml
ruffsl: A repo on Docker Hub only has one control panel to manage all of its tags, and in the future, it may be the case that two different purpose for the same repo may conflate or interfere with each other. Do you see this `moveit/moveit2` repo for pure public release distribution or continuous integration? Separating the two may be beneficial down the line, like have a separate `ros-planning/moveit2` just for CI and `moveit/moveit2` for users.

For example, I'd think it would be nice to set the source field to the build rules build and tag a respective docker image for each new devel branch using the available regex, while not interfering with build setting for release image tags. 

image


We may also want our CI to re-trigger all CI images, while not needlessly churning the release tags in the same repo, or modify the docker hub build to use custom build hooks customize build args:

https://github.com/ruffsl/navigation2/tree/foo/.dockerhub/devel

tools/code_coverage_report.bash
ruffsl: @davetcoleman would you like to me to port something like `movit2_common` into this PR?

Dockerfile
mlautman: I would recommend testing with the [.docker/ci/Dockerfile](https://github.com/ros-planning/moveit2/blob/90d331e536b9278c3d2c451cd3d1b6ca531f73db/.docker/ci/Dockerfile) in the [MoveIt2 Ci PR](https://github.com/ros-planning/moveit2/pull/46) as it is very close to being merged in.

.circleci/config.yml
```diff
@@ -0,0 +1,246 @@
...
+      - *store_test_logs
+  report_coverage:
+    description: "Report Coverage"
+    steps:
+      - *report_code_coverage
+
+executors:
+  docker_exec:
+    docker:
+      - image: moveit/moveit2:master
```
mlautman: I don't belive we currently build this image. the Base image is `moveit/moveit2:crystal-ci`

.circleci/config.yml
mlautman: Can you elaborate on why this is commented out and set a criteria for when it will acceptable to uncommenting them?

.docker/ci/Dockerfile
```diff
@@ -1,50 +1,82 @@
...
+# It also expects to be contained in the /moveit2 root folder for file copy
+# Example build command:
+# export http_proxy=http://my.proxy.com:80
+# export CMAKE_BUILD_TYPE=Debug
+# docker build -t nav2:latest --build-arg http_proxy --build-arg CMAKE_BUILD_TYPE ./
+FROM osrf/ros2:nightly
+
+# setup keys
+ARG http_proxy
+RUN if [ "$http_proxy" != "" ]; \
```
mlautman: We tend to keep Dockerfiles as a single RUN command to minimize copies

.docker/ci/Dockerfile
```diff
@@ -1,50 +1,82 @@
...
+      cmake \
+      git \
+      python3-colcon-common-extensions \
+      python3-vcstool \
+      wget \
+    && rm -rf /var/lib/apt/lists/*
+
+# copy ros package repo
+ENV MOVEIT2_WS /opt/moveit2_ws
+RUN mkdir -p $MOVEIT2_WS/src
```
mlautman: I don't think we should be putting the workspace in `/opt`. Isn't that generally reserved for system installs? In the past we've been putting the build in something like: `ENV ROS_WS=/root/ws_moveit`

.docker/ci/Dockerfile
```diff
@@ -1,50 +1,82 @@
...
+ARG CMAKE_BUILD_TYPE=Release
+
+RUN . /opt/ros/$ROS_DISTRO/setup.sh && \
+    colcon build \
+      --symlink-install \
+      --cmake-args \
+        -DCMAKE_BUILD_TYPE=$CMAKE_BUILD_TYPE
+
+# install moveit2 package dependencies
+WORKDIR $MOVEIT2_WS
```
mlautman: isn't this redundant

.docker/ci/Dockerfile
```diff
@@ -1,50 +1,82 @@
...
+    apt-get update && \
+    rosdep install -q -y \
+      --from-paths \
+        $ROS_WS/src \
+        src \
+      --ignore-src \
+    && rm -rf /var/lib/apt/lists/*
+
+# build moveit2 package source
+RUN rm $MOVEIT2_WS/src/moveit2/nav2_system_tests/COLCON_IGNORE
```
mlautman: I don't believe that `nav2_system_tests` is a package in `moveit2`

mlautman:
Thank you again for helping migrate MoveIt from travis to CircleCI!

Dockerfile
ruffsl: I made some comments on https://github.com/ros-planning/moveit2/pull/46#pullrequestreview-216485284 that the Dockerfile here also addresses. However, like I mentioned, the Dockerfile here and config go rather hand-in-hand due to the build caching strategy optimizations. That said, I'm not opposed to multiple dockerfiles; we could have one for the existing travis setup and one for CircleCI.

.circleci/config.yml
ruffsl: Please see discussion here https://github.com/ros-planning/moveit2/pull/44#discussion_r266540907

.docker/ci/Dockerfile
ruffsl: > minimize copies

I'm not sure what you mean. Do you mean layers? Are we close to limit for docker storage drivers on our CI? There are justifiable reasons to stage multiple layers to preserve build caches:

https://github.com/ros-planning/moveit2/pull/46#discussion_r267152476

.docker/ci/Dockerfile
ruffsl: I original helped with those dockerfiles as assuming the convention for root, and I think I was mistaken to do so. This makes it tricky to use the same workspace with changing to a different users, e.g. using your host machine user mounted in the container with x forwarding when testing GUIs etc,  

https://github.com/osrf/docker_images/pull/243
https://github.com/ros-planning/moveit2/pull/46#discussion_r267151753

.docker/ci/Dockerfile
ruffsl: Not quite, this is changing the pwd to the second workspace overlay to build moveit2. The first, `ROS2_WS` was  reserved for any upstream dependencies this project necessitates be built from source.

.docker/ci/Dockerfile
ruffsl: I don't know how this keeps escaping my patch commits, thanks!

Dockerfile
mlautman: We certainly can have multiple (we currently have about 4) but it is important that we remove redundancies wherever possible. Looking through the Dockerfile (albeit rather briefly) it seems that the docker container created isn't very different from [moveit/moveit2:crystal-source](https://github.com/ros-planning/moveit2/blob/moveit2-ci/.docker/source/Dockerfile)
The main difference being that this Dockerfile builds the workspace whiel crystal-source does not.  (Although it certainly could)

tools/code_coverage_report.bash
mlautman: That seems like a great optimization. Does it make sense to make a moveit2_common package or would it be better to rename the nav2_common package to something more general like ament_code_coverage and include that?

.docker/ci/Dockerfile
mlautman: I meant layers. Personally, I think that the initial decision to combine layers might have been a pre-mature optimization.

mlautman: https://github.com/ros-planning/moveit2/pull/46#discussion_r267171677

.docker/ci/Dockerfile
mlautman: sounds good

.docker/ci/Dockerfile
mlautman: I se

.docker/ci/Dockerfile
```diff
@@ -1,50 +1,82 @@
...
-    rm -rf /var/lib/apt/lists/*
-
-# Continous Integration Setting
-ENV IN_DOCKER 1
+# This dockerfile expects proxies to be set via --build-arg if needed
+# It also expects to be contained in the /moveit2 root folder for file copy
+# Example build command:
+# export http_proxy=http://my.proxy.com:80
+# export CMAKE_BUILD_TYPE=Debug
+# docker build -t nav2:latest --build-arg http_proxy --build-arg CMAKE_BUILD_TYPE ./
```
mlautman: moveit2:latest

Dockerfile
mlautman: @davetcoleman and myself were just discussing this and it makes sense, at least in the short term, to have multiple dockerfiles with one for the existing travis setup and one for CricleCi

.docker/ci/Dockerfile
ruffsl: I omitted the org so user building don't blow away what they may have pulled.

Dockerfile
ruffsl: Agreed, I would also like to prune Dockerfile redundancies. I have an idea to address this later using build args in the FROM calls and overriding the Dockerhub build hooks. See bfdbc03 and https://docs.docker.com/docker-hub/builds/advanced as I linked to above with https://github.com/ruffsl/navigation2/tree/foo/.dockerhub/devel

.docker/ci/Dockerfile
mlautman: Makes sense. maybe change nav2->moveit2?

Also I really like that the existing dockerfiles have their label in the documentation (eg https://github.com/ros-planning/moveit2/blob/master/.docker/ci/Dockerfile#L1) do you think you can add a simmilar line up top? It could be something like moveit/moveit2:crystal-circle-ci

.docker/ci/Dockerfile
```diff
@@ -1,50 +1,81 @@
-# moveit/moveit:melodic-ci
```
mlautman: Let's keep this file intact as it is already setup with dockerhub and travis. Perhaps we could move your changes to `.docker/circle-ci/Dockerfile`. Once that is done I am comfortable giving this a +1. 

On a related note, if you have access to the ros-planning repo, I would recommend pushing your branch to ros-planning/moveit2 so that we can test that everything works with Dockerhub. I did the same thing for my PR #46 and it has been really helpful for iterating quickly.

.docker/ci/Dockerfile
ruffsl: Rather than have some manual mapping from github branches to dockerhub tags, like `crystal-circle-ci`, I'd recommend we adopt a more programmatic style where release branches and tags on ros-planning/moveit2 github are mapped directly to tag names on ros-planning/moveit2 for CI consumption. There can be procedurally created, while the origin/relation of docker tags to code would be clear and more amenable for CI configs; alluded to in https://github.com/ros-planning/moveit2/pull/44#discussion_r267082024

.docker/ci/Dockerfile
ruffsl: I've restored the old Dockerfile in 6d642ca , but reminded I had to have new Dockerfile at the root of the project so that the build context may encompass the entire repo that it copies from. I've added a .dockerignore file to mitigate build cache breaks from git fetches or editing of the Dockerfile itself.

ruffsl:
Are there any more blocks or unaddressed comments for this?  
I'd like to get this hooked up and live so we can start iterating further.

.circleci/config.yml
```diff
@@ -0,0 +1,247 @@
+# `checksum.txt` derives the key for the `save_cache` and `restore_cache` steps.
```
davetcoleman:
```suggestion
# CircleCi Configuration for MoveIt
# Historically MoveIt has used Travis for continuous integration testing, but we are starting to migrate
# to CircleCi for its greater feature set. In particular, testing coverage metrics.
#
# `checksum.txt` derives the key for the `save_cache` and `restore_cache` steps.
```

davetcoleman: a little background info :)

.circleci/config.yml
```diff
@@ -0,0 +1,247 @@
...
+    executor: docker_exec
+    environment:
+      CMAKE_BUILD_TYPE: "Debug"
+      COVERAGE_ENABLED: "True"
+    steps:
+      - checkout_source
+      - setup_upstream
+      - build_source
+      - test_source
+      # - report_coverage
```
davetcoleman:
```suggestion
      # TODO: re-enable report coverage when...
      # - report_coverage
```
when? or remove comment

.dockerignore
```diff
@@ -0,0 +1,13 @@
+################################################################################
+# Repo
+
+.git/*
+.dockerignore
+.gitignore
+Dockerfile
```
davetcoleman:
```suggestion
Dockerfile
CONTRIBUTING.md
LICENSE.txt
MIGRATION.md
README.md
```
this improves caching, right?

Dockerfile
davetcoleman: Can we move this file to our ``.docker`` folder at least?

davetcoleman:
```suggestion
# Docker specifically for CircleCi setup, not for Travis
#
# This dockerfile expects proxies to be set via --build-arg if needed
```

moveit2.repos
```diff
@@ -0,0 +1,9 @@
+repositories:
+  angles:
+    type: git
+    url: https://github.com/ros/angles.git
+    version: ros2
+  image_common:
+    type: git
+    url: https://github.com/crdelsey/image_common.git
```
davetcoleman: these are junk repos. please use:
https://github.com/ros-planning/moveit2/blob/master/moveit.rosinstall

.circleci/config.yml
ruffsl: Is code coverage working now? I was waiting on this until this repo uses something like nav2's cmake changes to enable codecov at build time.

.dockerignore
ruffsl: I the `.md` and `.txt` files are inconsequential to the build, perhaps is worth blanketing them all in the root directory?
```
*.md
*.txt
```

Dockerfile
ruffsl: > Can we move this file to our .docker folder at least?

The Dockerfile is kept in the root of the project so that the build context may encompass the entire repo so that we can COPY all the separate source files into the image. It is possible to have different Dockerfile and build context locations, but from a user perspective, the expected build context location is no longer obvious. DockerHub has recently revamped the build rules setting, allowing us to specify separate paths, and we can document the use required `pwd` and `--file` for uses building locally.

https://docs.docker.com/engine/reference/commandline/build/#extended-description
https://docs.docker.com/engine/reference/commandline/build/#text-files

.dockerignore
mlautman: sounds good to me

henningkayser:
@mlautman this seems mostly done to me, except for the repos file maybe. What's missing to get this merged?

ruffsl:
As this repo hasn't been active for 3 months, the CircleCI config we use over on ros-planning/navigation2 that this PR derives from has advanced quite a bit. In the coming days I'll try and refactor those fixes to this PR. However I'm guessing the build state of the repo, that I could used to test the CI with, hasn't changed much since June.

henningkayser:
> As this repo hasn't been active for 3 months, the CircleCI config we use over on ros-planning/navigation2 that this PR derives from has advanced quite a bit. In the coming days I'll try and refactor those fixes to this PR. However I'm guessing the build state of the repo, that I could used to test the CI with, hasn't changed much since June.

@ruffsl, thanks a lot for looking into this again!

tylerjw:
> I shortly skimmed through this. Some `lcov` commands are interesting to port over into `moveit_ci`.
> Otherwise, I think we can (and should) keep our hierarchy of MoveIt1 docker files that are optimized for fast CI building. Particularly, the proposed approach here requires to `rosdep install` all packages for each CI build from scratch, doesn't it?
> CircleCI's configuration seems to be amazingly complex. I'm not sure, we should switch from Travis at all.

The advantages of this approach as far as I can tell is actually slight better (and worse) caching.  With this approach the CI run is done in stages and a hash of the input of each stage is calculated.  If possible (all those inputs existed before) it uses a cache of the output of that sage.  

Specifically if the external ros dependencies installed via debians or built from source haven't changed they can be re-used (right now we sort of have that with our ci docker images except we have to rebuild and push them every so often, unless that's done automatically when the ros dependencies update). 

Next, if that first stage (ros dependencies) was able to be reused from a cache it does the same thing with the upstream workspace.  If there was no changes in the inputs for the upstream workspace it can just re-use it from a cache.

Lastly for the target workspace it can do the same thing.  This would happen if you re-ran a job with no code changes it would basically just skip to the testing step and run the tests.

To get that type of caching we would need a ci system that supported pipeline-ing in the way that circle-ci is built to support.  Since this was written CircleCI has the idea of Orbs which are CI routines (similar to what we essentially have in moveit_ci / industrial_ci).  These CI orbs can be published and re-used between projects that just specify parameters to the Orb.  Reference: https://circleci.com/orbs/

This is the same sort of workflow that Gitlab CI is now supporting.  Because of this (and my desire to see faster ci runs) I've submitted an issue to start the discussion of supporting pipeline style ci runs (and therefore caching of steps in the ci process) I've opened this issue: https://github.com/ros-industrial/industrial_ci/issues/548

@ruffsl I'm curious how the navigation2 CI system has changed and if this could be abstracted into a CircleCI Orb we could try to re-use and standardize on.

tylerjw:
One big advantage I see from supporting this sort of caching is if you had two developers submit patches around the same time (where none of the upstream dependencies would have changed) you would basically get to skip that step.

There is a chance (unknown to me) that travis supports this sort of caching and pipe-lining too, we could also roll it ourselves (yuck).

tylerjw:
The answer to the ccache problem was right in front of me in navigation2.  I wonder if we can abstract the circleci config.yaml into an Orb so we can standardize how to do this.

tylerjw:
@ruffsl reading through the CircleCI documentation I don't understand some of your syntax.  My questions relate to this file: https://github.com/ros-planning/navigation2/blob/master/.circleci/config.yml

For starters I can't find the keyword `references` in their documentation or any mention of the syntax of this: `common_environment: &common_environment`.  Can you explain or reference me the documentation for that?

tylerjw:
I take back what I said above... I don't think you are using cache as optimally as you could be.  Also, because of the symbolic link you put in the underlay workspace to /opt/ros/$ROS_DISTRO you don't build the underlay and always skip it in setup.  I'm a bit confused by that.

tylerjw:
Another thing I think I found is that you are using the caching system instead of [persist_to_workspace](https://circleci.com/docs/2.0/configuration-reference/#persist_to_workspace) where that would be a better fitting tool for what you want (when you re-use a workspace from the debug on the release tests).

ruffsl:
> @ruffsl I'm curious how the navigation2 CI system has changed and if this could be abstracted into a CircleCI Orb we could try to re-use and standardize on.

As I've just commented in a ticket elsewhere, with respect to the current state of the Nav2 CI, I'd recommend reading through the docs I've been adding on the subject. It should correct you idea on some of the concepts and features we've been leveraging with CircleCI. Perhaps you may want to comment on the PR with more specific questions after you've read through it, as there may still be some technical details that to me as the author seem self evident, but not so for someone just approaching this.

* https://github.com/ros-planning/navigation2/pull/1731
* https://github.com/ros-industrial/industrial_ci/issues/548#issuecomment-656601260

If there's interest, I can revisit this and quickly back port the improvements I've made to the Nav2 CI to this PR (or just start a new one by now, haha). I'm not sure how much we could push into an orb. Orbs are kind of clunky, given how they are imported, and scope. A lot of the optimizations also come from the style of Docker base image, which an orb wouldn't subplant. As in the other ticket, I'd like to work towards GH actions, given the worker client software is OSS. But last I checked GH actions, the dev story for caching and test results was a bit too under cooked 6 months ago, but that may have changed. I was waiting for a few of the more polished asure CI feature from Microsoft to roll over in the GH actions.

tylerjw:
As to GH Actions.  industrial_ci already supports it natively and it works fairly well.  My motivation to figure out the CircleCI system is that moveit has really long ci times and I think that with a careful use of the caching we could greatly reduce the ci run times.

davetcoleman:
@tylerjw now that migration to Github Actions is basically done, should we close this PR? It is our oldest outstanding PR and does not seem to have momentum anymore.

tylerjw:
> @tylerjw now that migration to Github Actions is basically done, should we close this PR? It is our oldest outstanding PR and does not seem to have momentum anymore.

I'll close it.  Please re-open this PR if you'd like to resume working on it or using work from it to form a new PR.

PR #52, opened 2019-03-19
Fixes to melodic-sync, adapt to ROS2 style

vmayoral:
Follows from https://github.com/ros-planning/moveit2/pull/36

### DIFF
--- a/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
+++ b/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
@@ -42,14 +42,13 @@
 #include <angles/angles.h>
 #include <cmath>
 #include <moveit/trajectory_processing/time_optimal_trajectory_generation.h>
-#include <ros/console.h>
 #include <vector>
-
+#include "rclcpp/rclcpp.hpp"
 #include <iostream>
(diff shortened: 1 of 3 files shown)


moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -42,14 +42,13 @@
...
-#include <ros/console.h>
 #include <vector>
-
+#include "rclcpp/rclcpp.hpp"
 #include <iostream>
 
 namespace trajectory_processing
 {
-const std::string LOGNAME = "trajectory_processing.time_optimal_trajectory_generation";
+rclcpp::Logger logger_trajectory_processing_optimal = rclcpp::get_logger("trajectory_processing.time_optimal_trajectory_generation");
```
mlautman: This could still be const. I don't know why it can't be static, and as a global I believe it should be all caps. I am flexible on this as we can always fix up those concerns repo-wide later on and I would like to keep the ROS2 port moving along

mlautman: https://github.com/ros-planning/moveit2/issues/21

PR #53, opened 2019-03-24
Fix issues form https://github.com/ros-planning/moveit2/pull/47

vmayoral:
Fix compilation issues originated from https://github.com/ros-planning/moveit2/pull/47


henningkayser:
@vmayoral could you resolve the conflicts? This looks like it could be merged easily.

PR #55, opened 2019-03-24
Background_processing module

vmayoral:


mlautman:
Looks good to me!

I would like to see if we could get CI running on this before merging so that we can verify the changes but I would also prefer to not let PR's sit for too long while we wait.

moveit_core/background_processing/src/background_processing.cpp
```diff
@@ -35,12 +35,15 @@
...
 #include <moveit/background_processing/background_processing.h>
-#include <ros/console.h>
+#include "rclcpp/rclcpp.hpp"
 
 namespace moveit
 {
 namespace tools
 {
+// Logger
+rclcpp::Logger LOGGER = rclcpp::get_logger("background_processing");
```
henningkayser:
```suggestion
const rclcpp::Logger LOGGER = rclcpp::get_logger("background_processing");
```

mlautman: We've been down that rode and Acutronics is having issues with const loggers. I think we let it slide for now and make the change to const repo wide after the port

henningkayser: I thought the problem was with having static loggers?

anasarrak: You have the error https://github.com/ros-planning/moveit2/issues/21#issuecomment-478259602 and there is an open issue just for this https://github.com/ros2/rclcpp/issues/679 and an open [PR](https://github.com/ros2/rclcpp/pull/680)

PR #62, opened 2019-03-31
Planning request adapter, port to ROS 2

vmayoral:

PR #67, opened 2019-03-31
Port dynamics_solver to ROS 2

vmayoral:


vmayoral:
See See https://github.com/ros-planning/moveit2/pull/62#issuecomment-480720464

PR #68, opened 2019-03-31
Port moveit_core/utils to ROS 2

vmayoral:
Initial iteration was made on a different "master" status


vmayoral:
See https://github.com/ros-planning/moveit2/pull/62#issuecomment-480720464

moveit_core/utils/CMakeLists.txt
```diff
@@ -2,34 +2,37 @@ set(MOVEIT_LIB_NAME moveit_utils)
 
 add_library(${MOVEIT_LIB_NAME}
   src/lexical_casts.cpp
-  src/xmlrpc_casts.cpp
+  # src/xmlrpc_casts.cpp # TODO: solve the conflicts with xmlrpc dependencies
```
henningkayser: I think `xmlrpc_casts.cpp` will not be used anymore since XML-RPC is not supported in ROS 2.
`parseDouble()` and `isArray()` should be redundant in the future but we have to check where`isStruct()` is used and if we need to reimplement it somehow.
Let's remove `xmlrpc_casts.cpp` and add TODOs everywhere where it is used.
These parts will need to be changed anyway since the same missing XML-RPC dependencies are used.

ahcorde:
@henningkayser can you review it?

ahcorde:
@henningkayser friendly ping

henningkayser:
@ahcorde removing this files won't be enough since the functionality needs to be reimplemented eventually. Those files that `xmlrpc_casts.h`  will fail to compile which is just terrible to debug and could slow down the port.
* moveit_plugins/moveit_simple_controller_manager/src/follow_joint_trajectory_controller_handle.cpp
* moveit_plugins/moveit_simple_controller_manager/src/moveit_simple_controller_manager.cpp
* moveit_core/kinetmatic_constraints/src/utils.cpp
Please fix related includes and add a descriptions for all function calls that need to be replaced somehow. Ideally we would replace the implementation right now but I believe that we still need to have some further discussions on how we handle parameter config files.

ahcorde:
we decided to remove `xmlrpc` because in ROS 2 is not used anymore. This concept of getting parameters with xmlrpc has disappeared. There are new ways the get parameters though ROS 2 and DDS. 

We have replaced all these calls in the code with the corresponding new ROS 2.0 code. I think it's better to discuss if we can include new abstractions for getting parameters than maintaining this old ROS code.

henningkayser:
> we decided to remove `xmlrpc` because in ROS 2 is not used anymore. This concept of getting parameters with xmlrpc has disappeared. There are new ways the get parameters though ROS 2 and DDS.





> we decided to remove `xmlrpc` because in ROS 2 is not used anymore. This concept of getting parameters with xmlrpc has disappeared. There are new ways the get parameters though ROS 2 and DDS.
 
Yes, that's why I suggested to remove `xmlrpc`.

> We have replaced all these calls in the code with the corresponding new ROS 2.0 code. I think it's better to discuss if we can include new abstractions for getting parameters than maintaining this old ROS code.

I wasn't aware that this was already implemented. Is it still under review? If yes I would suggest adding related descriptions and would +1 this PR

ahcorde:
You can open this (huge) [PR](https://github.com/AcutronicRobotics/moveit2/commit/1fc3cf4bb3d1979fe71ac6ff175c3667aa68ace7) and have a look to moveit_planners/ompl/ompl_interface/src/ompl_interface.cpp.  You will see the changes.

Most of these changes are still pending of been ported. But You will find a TODO (anasarrak) or (ahcorde). The parameters which were ported are the most relevant.

moveit_core/utils/include/moveit/utils/robot_model_test_utils.h
```diff
@@ -182,10 +182,10 @@ class RobotModelBuilder
 
 private:
   /** \brief Adds different collision geometries to a link. */
-  void addLinkCollision(const std::string& link_name, const urdf::CollisionSharedPtr& coll, geometry_msgs::Pose origin);
+  void addLinkCollision(const std::string& link_name, urdf::CollisionSharedPtr coll, geometry_msgs::msg::Pose origin);
```
mlautman: Why were these changed from const references to copies if the shared pointer? The code hasn't changed at all. Why should these?

ahcorde: my bad. Fixed

PR #69, opened 2019-04-01
collision_plugin_loader port to ROS 2

vmayoral:
Authored originally by @anasarrak.

### DIFF
--- a/moveit_ros/planning/collision_plugin_loader/CMakeLists.txt
+++ b/moveit_ros/planning/collision_plugin_loader/CMakeLists.txt
@@ -2,7 +2,13 @@ set(MOVEIT_LIB_NAME moveit_collision_plugin_loader)
 
 add_library(${MOVEIT_LIB_NAME} src/collision_plugin_loader.cpp)
 set_target_properties(${MOVEIT_LIB_NAME} PROPERTIES VERSION "${${PROJECT_NAME}_VERSION}")
-target_link_libraries(${MOVEIT_LIB_NAME} ${catkin_LIBRARIES})
+ament_target_dependencies(${MOVEIT_LIB_NAME}
+  ${rclcpp_LIBRARIES}
+  urdf
+  pluginlib)
 
-install(TARGETS ${MOVEIT_LIB_NAME} LIBRARY DESTINATION ${CATKIN_GLOBAL_LIB_DESTINATION})
-install(DIRECTORY include/ DESTINATION ${CATKIN_GLOBAL_INCLUDE_DESTINATION})
+install(TARGETS ${MOVEIT_LIB_NAME}
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
+  RUNTIME DESTINATION bin)
+install(DIRECTORY include/ DESTINATION include)
--- a/moveit_ros/planning/collision_plugin_loader/CMakeLists.txt
+++ b/moveit_ros/planning/collision_plugin_loader/CMakeLists.txt
@@ -3,7 +3,7 @@ set(MOVEIT_LIB_NAME moveit_collision_plugin_loader)
 add_library(${MOVEIT_LIB_NAME} src/collision_plugin_loader.cpp)
 set_target_properties(${MOVEIT_LIB_NAME} PROPERTIES VERSION "${${PROJECT_NAME}_VERSION}")
 ament_target_dependencies(${MOVEIT_LIB_NAME}
-  ${rclcpp_LIBRARIES}
+  rclcpp
   urdf
   pluginlib)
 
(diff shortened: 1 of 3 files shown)


mlautman:
Let's rebase this onto master so that we can start using travis

vmayoral:
@mlautman rebased this and https://github.com/ros-planning/moveit2/pull/71 to check CI. I see https://github.com/ros-planning/moveit2/pull/71 failed so you may want to review that.

vmayoral:
On a second look, it looks like the CI is not building anything just yet right @mlautman? https://github.com/ros-planning/moveit2/blob/master/.docker/ci/Dockerfile

.docker/ci/Dockerfile
```diff
@@ -38,8 +38,8 @@ RUN apt-get -qq update && \
 # NOTE(mlautman): These pip installs are from the ros2 source install instructions. Seeing as not all of them
 #        are installed in the base image I added them here. I have not been able to verify that they are needed
 #        and if they aren't necessary they should be removed.
-RUN python3 -m pip install --upgrade pip && \
-    python3 -m pip install -U \
+RUN python3 -m pip install -q --upgrade pip && \
```
mlautman: Please update to the newest master. I'm not 100% on the best way to do this but one way would be:

```bash
git fetch upstream
git reset --hard upstream/master
git cherry-pick <Your new commits>
```
In this case it would be 
```bash
git fetch upstream
git reset --hard upstream/master
git cherry-pick e2ac85a184eff0dc5f62bc6032e97f979061c47b 5fce9cface567332607a0acf0d786ee35c171068
```

On my local machine this results in a branch that is identical to `AcutronicRobotics:collision-plugin-loader`

mlautman:
I think that building the sub-packages and passing travis is an essential part of the PR process. Ideally each PR would include the changes necessary for travis to build the changes. 

In this case, there are dependencies that will need to be ported before this can be built. (`moveit_core` and `moveit_ros_perception`) If there are open PR's for the dependencies that are blocking this port, then I would recommend that you include links in the PR description. That will help direct our focus towards getting the underlying dependencies done first.

.docker/ci/Dockerfile
vmayoral: I've rebased and pulled. Probably did it wrong. Will have a look again in the future.

vmayoral:
> I think that building the sub-packages and passing travis is an essential part of the PR process. Ideally each PR would include the changes necessary for travis to build the changes.
> 
> In this case, there are dependencies that will need to be ported before this can be built. (`moveit_core` and `moveit_ros_perception`) If there are open PR's for the dependencies that are blocking this port, then I would recommend that you include links in the PR description. That will help direct our focus towards getting the underlying dependencies done first.

Great, thanks. Can you also indicate which changes would be required for travis? Maybe provide an example for a simpler module (e.g. https://github.com/ros-planning/moveit2/pull/67)?

mlautman:
Here are the steps: 
 - Remove the COLCON_IGNORE at [moveit2/moveit_ros/COLCON_IGNORE](https://github.com/ros-planning/moveit2/blob/master/moveit_ros/COLCON_IGNORE)  
- Port the CMakeLists.txt in [moveit2/moveit_ros/CMakeLists.txt](https://github.com/ros-planning/moveit2/blob/master/moveit_ros/CMakeLists.txt) so that it works with colcon. You will want to comment out all of the `add_subdirectory` calls to packages that have yet to be ported, leaving only the ported packages. 
- Port the CMakeLists.txt for the subpackage in question. In this case it would be  
[moveit2/moveit_ros/planning/collision_plugin_loader/CMakeLists.txt](https://github.com/ros-planning/moveit2/blob/master/moveit_ros/planning/collision_plugin_loader/CMakeLists.txt)
- When you remove the COLCON_IGNORE file, travis will automatically try to build the packages.

Let me know if this works for you

.docker/ci/Dockerfile
vmayoral: Totally my bad indeed. Should be fine now @mlautman, thanks!

vmayoral:
@mlautman, I went ahead and performed the steps described over the weekend. Thanks for the guidelines. Unfortunately, the result wasn't satisfactory and @LanderU and I ended up having to modify the Docker image and moveit_ci quite heavily to make it work. 

See https://github.com/AcutronicRobotics/moveit2/issues/34 for the complete discussion or https://github.com/ros-planning/moveit_ci/pull/56 for the proposed changes. Returning to this PR, I'd advice we go ahead and merge these changes since they already pass CI (see https://github.com/ros-planning/moveit2/pull/62#issuecomment-480720464).

mlautman:
> Unfortunately, the result wasn't satisfactory

This is a good thing. The fact that CI failed shows us were there are still issues. Let's fix them in https://github.com/ros-planning/moveit_ci/pull/56 and get these packages building :)

henningkayser:
Closing in favor of #137

PR #74, opened 2019-04-02
kinematics_plugin_loader port to ROS 2

vmayoral:


vmayoral:
Same as https://github.com/ros-planning/moveit2/pull/62#issuecomment-480720464, consider the changes please.

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -287,31 +288,39 @@ robot_model::SolverAllocatorFn KinematicsPluginLoader::getLoaderFunction(const s
...
       {
-        ROS_DEBUG_NAMED(LOGNAME, "Loading settings for kinematics solvers from the ROS param server ...");
+        RCLCPP_DEBUG(LOGGER, "Loading settings for kinematics solvers from the ROS param server ...");
 
         // read data using ROS params
-        ros::NodeHandle nh("~");
-
+        // ros::NodeHandle nh("~");
+        //TODO (anasarrak): Add the appropriate node name for ROS2
+        auto node = rclcpp::Node::make_shared("talker");
```
mlautman: This is a big TODO. Let's do this the right way now

anasarrak: @mlautman any suggestion?

mlautman: I like the approach that was used for the controller manager where the node is passed in during initialization as it seems most flexible.

henningkayser:
Closed in favor of https://github.com/ros-planning/moveit2/pull/107

PR #76, opened 2019-04-05
Port rdf_loader to ROS 2

vmayoral:


vmayoral:
Same as https://github.com/ros-planning/moveit2/pull/62#issuecomment-480720464. Please consider these changes.

moveit_ros/planning/rdf_loader/src/rdf_loader.cpp
```diff
@@ -49,48 +49,56 @@
...
 rdf_loader::RDFLoader::RDFLoader(const std::string& robot_description)
 {
   moveit::tools::Profiler::ScopedStart prof_start;
   moveit::tools::Profiler::ScopedBlock prof_block("RDFLoader(robot_description)");
 
-  ros::WallTime start = ros::WallTime::now();
-  ros::NodeHandle nh("~");
+  auto start = std::chrono::system_clock::now();
+  // TODO (anasarrak): Add the correct node_handle name
+  auto node = rclcpp::Node::make_shared("/");
```
mlautman: This is a big TODO. What is our strategy for Nodes in MoveIt2? Should we always check that rclcpp has been initialized first? Seems like something we should do with intention.

moveit_ros/planning/rdf_loader/src/rdf_loader.cpp
```diff
@@ -160,25 +168,25 @@ bool rdf_loader::RDFLoader::loadXacroFileToString(std::string& buffer, const std
...
   if (!boost::filesystem::exists(path))
   {
-    ROS_ERROR_NAMED("rdf_loader", "File does not exist");
+    RCLCPP_ERROR(LOGGER, "File does not exist");
     return false;
   }
-
-  std::string cmd = "rosrun xacro xacro ";
+  //TODO (anasarrak): Test ros2 xacro https://github.com/ros/xacro/tree/ros2 // https://github.com/bponsler/xacro
+  std::string cmd = "ros2 run xacro xacro";
```
mlautman: Do we have a demonstration of the RDF Loader loading it loading a robot description? Seems like a test that we should really consider adding

moveit_ros/planning/rdf_loader/src/rdf_loader.cpp
mlautman: Suggestion: change the api for rdf_loader so that the user must pass in a node.

moveit_ros/planning/rdf_loader/src/rdf_loader.cpp
```diff
@@ -49,48 +49,56 @@
 #include <fstream>
 #include <streambuf>
 #include <algorithm>
+#include <chrono>
+
+rclcpp::Logger LOGGER = rclcpp::get_logger("moveit").get_child("rdf_loader");
```
mlautman: this should be inside a namespace like the other loggers. 
If you wanted to use `namespace rdf_loader` then you could also find replace `rdf_loader::RDFLoader::` with `RDFLoader::` for every method in the namespace block

henningkayser:
Closed in favor of https://github.com/ros-planning/moveit2/pull/104

PR #83, opened 2019-05-16
Kinematics base port to ROS 2

ahcorde:
- Fixing logger variable 
 - Library as shared
 - porting lookupParam function but it still need more work

### DIFF
--- a/moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
+++ b/moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
@@ -582,6 +582,7 @@ public:
   KinematicsBase();
 
 protected:
+  rclcpp::Node::SharedPtr node_;
   moveit::core::RobotModelConstPtr robot_model_;
   std::string robot_description_;
   std::string group_name_;
--- a/moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
+++ b/moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
@@ -615,51 +615,26 @@ protected:
   template <typename T>
   inline bool lookupParam(std::shared_ptr<rclcpp::Node>& node, const std::string& param, T& val, const T& default_val) const
   {
-    // TODO (ahcorde):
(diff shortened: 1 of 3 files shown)


mlautman:
> porting lookupParam function but it still need more work

@ahcorde Is this a work in progress or is it ready for review?

ahcorde:
@mlautman  reworked!

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -582,6 +582,7 @@ class KinematicsBase
   KinematicsBase();
 
 protected:
+  rclcpp::Node::SharedPtr node_;
```
mlautman: When does this get initialized or used?

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -59,7 +59,7 @@ MOVEIT_CLASS_FORWARD(RobotModel)
 namespace kinematics
 {
 // Logger
-rclcpp::Logger LOGGER = rclcpp::get_logger("kinematics_base");
+static rclcpp::Logger LOGGER_KINEMATICS_BASE = rclcpp::get_logger("kinematics_base");
```
mlautman: Why is this static? It is not in a method or a class so when would their ever be multiple copies?

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
ahcorde: The initialization happens in the kinematics_plugin, `kdl_`, `lma_`, `srv_` make use of node

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
ahcorde: I think I made this static (quick fix) because it used in the cpp and the hpp and otherwise was failing. Suggestions?

ahcorde:
@mlautman did you review my comments?

PR #104, opened 2019-10-21
Port rdf_loader to ROS2 

JafarAbdi:
### Description

This is a continuation of [PR#76](https://github.com/ros-planning/moveit2/pull/76)

The addressed review was added to the master branch [see](https://github.com/AcutronicRobotics/moveit2/commits/master/moveit_ros/planning/rdf_loader) but because I'm not able to cherry-pick the changes for the specific files of rdf_loader, I just applied the changes to the rdf_loader and committed them

PR #107, opened 2019-10-21
Port kinematics_plugin_loader to ROS2 

JafarAbdi:
### Description

This is a continuation of [PR#74](https://github.com/ros-planning/moveit2/pull/74)

I did the same as in this [PR#104](https://github.com/ros-planning/moveit2/pull/104) for applying the changes

### DIFF
--- a/moveit_ros/planning/kinematics_plugin_loader/CMakeLists.txt
+++ b/moveit_ros/planning/kinematics_plugin_loader/CMakeLists.txt
@@ -6,7 +6,10 @@ ament_target_dependencies(${MOVEIT_LIB_NAME}
   moveit_rdf_loader
   rclcpp
   urdf
-  pluginlib)
+  class_loader
+  ament_index_cpp
+  moveit_core
+  Boost)
 
 
 install(TARGETS ${MOVEIT_LIB_NAME}
--- a/moveit_ros/planning/kinematics_plugin_loader/CMakeLists.txt
+++ b/moveit_ros/planning/kinematics_plugin_loader/CMakeLists.txt
@@ -2,10 +2,16 @@ set(MOVEIT_LIB_NAME moveit_kinematics_plugin_loader)
 
 add_library(${MOVEIT_LIB_NAME} src/kinematics_plugin_loader.cpp)
 set_target_properties(${MOVEIT_LIB_NAME} PROPERTIES VERSION "${${PROJECT_NAME}_VERSION}")
-target_link_libraries(${MOVEIT_LIB_NAME} moveit_rdf_loader ${catkin_LIBRARIES})
+ament_target_dependencies(${MOVEIT_LIB_NAME}
+  moveit_rdf_loader
+  rclcpp
+  urdf
+  pluginlib)
+
 
 install(TARGETS ${MOVEIT_LIB_NAME}
-        LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-        ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-        RUNTIME DESTINATION ${CATKIN_GLOBAL_BIN_DESTINATION})
-install(DIRECTORY include/ DESTINATION ${CATKIN_GLOBAL_INCLUDE_DESTINATION})
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
+  RUNTIME DESTINATION bin)
+
+install(DIRECTORY include/ DESTINATION include)
--- a/moveit_ros/planning/kinematics_plugin_loader/include/moveit/kinematics_plugin_loader/kinematics_plugin_loader.h
+++ b/moveit_ros/planning/kinematics_plugin_loader/include/moveit/kinematics_plugin_loader/kinematics_plugin_loader.h
@@ -50,27 +50,29 @@ class KinematicsPluginLoader
 {
 public:
   /** \brief Load the kinematics solvers based on information on the
-      ROS parameter server. Take as optional argument the name of the
+      ROS parameter server. Take node as an argument and as optional argument the name of the
       ROS parameter under which the robot description can be
       found. This is passed to the kinematics solver initialization as
       well as used to read the SRDF document when needed. */
-  KinematicsPluginLoader(const std::string& robot_description = "robot_description",
+  KinematicsPluginLoader(const rclcpp::Node::SharedPtr& node,
+                         const std::string& robot_description = "robot_description",
                          double default_search_resolution = 0.0)
-    : robot_description_(robot_description), default_search_resolution_(default_search_resolution)
+    : node_(node), robot_description_(robot_description), default_search_resolution_(default_search_resolution)
   {
   }
 
   /** \brief Use a default kinematics solver (\e solver_plugin) for
       all the groups in the robot model. The default timeout for the
       solver is \e solve_timeout and the default number of IK attempts
-      is \e ik_attempts. Take as optional argument the name of the ROS
+      is \e ik_attempts. Take node as an argument and as optional argument the name of the ROS
       parameter under which the robot description can be found. This
       is passed to the kinematics solver initialization as well as
       used to read the SRDF document when needed. */
-  KinematicsPluginLoader(const std::string& solver_plugin, double solve_timeout, unsigned int /*ik_attempts*/,
+  KinematicsPluginLoader(const rclcpp::Node::SharedPtr& node, const std::string& solver_plugin, double solve_timeout,
                          const std::string& robot_description = "robot_description",
                          double default_search_resolution = 0.0)
-    : robot_description_(robot_description)
+    : node_(node)
+    , robot_description_(robot_description)
     , default_search_resolution_(default_search_resolution)
     , default_solver_plugin_(solver_plugin)
     , default_solver_timeout_(solve_timeout)
@@ -100,6 +102,7 @@ public:
   void status() const;
 
 private:
+  const rclcpp::Node::SharedPtr node_;
   std::string robot_description_;
   double default_search_resolution_;
 
--- a/moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
+++ b/moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
@@ -47,7 +47,7 @@
 
 namespace kinematics_plugin_loader
 {
-rclcpp::Logger LOGGER = rclcpp::get_logger("kinematics_plugin_loader");
+rclcpp::Logger LOGGER_KINEMATICS_PLUGIN_LOADER = rclcpp::get_logger("kinematics_plugin_loader");
 
 class KinematicsPluginLoader::KinematicsLoaderImpl
 {
@@ -60,9 +60,9 @@ public:
    * \param iksolver_to_tip_links - a map between each ik solver and a vector of custom-specified tip link(s)
    */
   KinematicsLoaderImpl(const std::string& robot_description,
-                       const std::map<std::string, std::vector<std::string> >& possible_kinematics_solvers,
-                       const std::map<std::string, std::vector<double> >& search_res,
-                       const std::map<std::string, std::vector<std::string> >& iksolver_to_tip_links)
+                       const std::map<std::string, std::vector<std::string>>& possible_kinematics_solvers,
+                       const std::map<std::string, std::vector<double>>& search_res,
+                       const std::map<std::string, std::vector<std::string>>& iksolver_to_tip_links)
     : robot_description_(robot_description)
     , possible_kinematics_solvers_(possible_kinematics_solvers)
     , search_res_(search_res)
@@ -75,7 +75,7 @@ public:
     }
     catch (pluginlib::PluginlibException& e)
     {
-      RCLCPP_ERROR(LOGGER, "Unable to construct kinematics loader. Error: %s", e.what());
+      RCLCPP_ERROR(LOGGER_KINEMATICS_PLUGIN_LOADER, "Unable to construct kinematics loader. Error: %s", e.what());
     }
   }
 
@@ -87,22 +87,23 @@ public:
   std::vector<std::string> chooseTipFrames(const robot_model::JointModelGroup* jmg)
   {
     std::vector<std::string> tips;
-    std::map<std::string, std::vector<std::string> >::const_iterator ik_it =
-        iksolver_to_tip_links_.find(jmg->getName());
+    std::map<std::string, std::vector<std::string>>::const_iterator ik_it = iksolver_to_tip_links_.find(jmg->getName());
 
     // Check if tips were loaded onto the rosparam server previously
     if (ik_it != iksolver_to_tip_links_.end())
     {
       // the tip is being chosen based on a corresponding rosparam ik link
-      RCLCPP_DEBUG(LOGGER, "Choosing tip frame of kinematic solver for group %s"
-                                  "based on values in rosparam server.",jmg->getName().c_str());
+      RCLCPP_INFO(LOGGER_KINEMATICS_PLUGIN_LOADER, "Choosing tip frame of kinematic solver for group %s"
+                                                   "based on values in rosparam server.",
+                  jmg->getName().c_str());
       tips = ik_it->second;
     }
     else
     {
       // get the last link in the chain
-      RCLCPP_DEBUG(LOGGER, "Choosing tip frame of kinematic solver for group"
-                                  "based on last link in chain",jmg->getName().c_str());
+      RCLCPP_INFO(LOGGER_KINEMATICS_PLUGIN_LOADER, "Choosing tip frame of kinematic solver for group %s"
+                                                   "based on last link in chain",
+                  jmg->getName().c_str());
 
       tips.push_back(jmg->getLinkModels().back()->getName());
     }
@@ -110,7 +111,7 @@ public:
     // Error check
     if (tips.empty())
     {
-      RCLCPP_ERROR(LOGGER, "Error choosing kinematic solver tip frame(s).");
+      RCLCPP_ERROR(LOGGER_KINEMATICS_PLUGIN_LOADER, "Error choosing kinematic solver tip frame(s).");
     }
 
     // Debug tip choices
@@ -118,7 +119,7 @@ public:
     tip_debug << "Planning group '" << jmg->getName() << "' has tip(s): ";
     for (std::size_t i = 0; i < tips.size(); ++i)
       tip_debug << tips[i] << ", ";
-    RCLCPP_DEBUG(LOGGER, tip_debug.str());
+    RCLCPP_INFO(LOGGER_KINEMATICS_PLUGIN_LOADER, tip_debug.str());
 
     return tips;
   }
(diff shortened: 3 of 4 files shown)


moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -287,78 +295,101 @@ robot_model::SolverAllocatorFn KinematicsPluginLoader::getLoaderFunction(const s
...
+          //     ksolver = node->get_parameter(ksolver_param_name).get_value<std::string>();
+          //
+          //   }
+          // }
 
           std::string ksolver_timeout_param_name;
-          if (nh.searchParam(base_param_name + "/kinematics_solver_timeout", ksolver_timeout_param_name))
+          if (ksolver_params->has_parameter(base_param_name + "/kinematics_solver_timeout"))
           {
+            ksolver_timeout_param_name = node->get_parameter(base_param_name + "/kinematics_solver_timeout").get_value<std::string>();
```
henningkayser: `Parameter::get_value()` will throw a `ParameterTypeException` if the requested type is different from the actual `ParameterType` (http://docs.ros2.org/dashing/api/rclcpp/parameter__value_8hpp_source.html). Apparently, this happens for both `Node::get_parameter()` as well as the parameters client's version. I think if we really want to have a safe version of querying parameters, we have the following options:
1. Wrap any typed parameter queries inside try/catch blocks and handle unexpected results
2. Check if parameter exists, check parameter type, get parameter value using parameter type

If we have flexible parameter types (i.e. `int`, `double` like in the example below) we should be able to just use auto type deduction or let the default type conversion handle it.

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -398,23 +432,25 @@ robot_model::SolverAllocatorFn KinematicsPluginLoader::getLoaderFunction(const s
...
-              }
-              else
-              {
+            ksolver_ik_links_param_name = node->get_parameter(
+                                base_param_name + "/kinematics_solver_ik_links").get_value<std::string>();
+            // XmlRpc::XmlRpcValue ksolver_ik_links;
+            std::vector<std::string> ksolver_ik_links;
+            if(ksolver_params->has_parameter(ksolver_ik_links_param_name)){
+              ksolver_ik_links = node->get_parameter(ksolver_ik_links_param_name).get_value<std::vector<std::string>>();
+              if(node->get_parameter(ksolver_ik_links_param_name).get_type_name().find("array")){
```
henningkayser: It seems like type checking methods are all over the place here. We should clean this up and note the steps as "best practices". @JafarAbdi do you want me to make an attempt?

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -287,78 +295,101 @@ robot_model::SolverAllocatorFn KinematicsPluginLoader::getLoaderFunction(const s
...
-                                known_groups[i].name_.c_str());
-              }
             }
+            // rclcpp::Parameter param =  ksolver_params->get_parameter(base_param_name + ".kinematics_solver");
+            // ksolver_param_name = ksolver_params->get_parameter(base_param_name + ".kinematics_solver").get_value<std::string>();
+            // found = ksolver_params->get_parameter(base_param_name + ".kinematics_solver", parameter);
+            RCLCPP_INFO(LOGGER_KINEMATICS_PLUGIN_LOADER, "%d param is %s ", found, ksolver_param_name.c_str());
+
           }
+          // if (found)
```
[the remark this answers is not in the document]
JafarAbdi: Yes, all the commented codes are from the old implementation [see](https://github.com/AcutronicRobotics/moveit2/blame/master/moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp) I think this should replace the previous `if (found)` block

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
JafarAbdi: Yep, I'm happy to do it -- should I do it in a separate PR .? I would replace the comparison with ` rclcpp::ParameterType::PARAMETER_STRING_ARRAY`

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -287,78 +295,101 @@ robot_model::SolverAllocatorFn KinematicsPluginLoader::getLoaderFunction(const s
...
 
+        //TODO (JafarAbdi): Add the appropriate node name for ROS2
+        auto ksolver_params = std::make_shared<rclcpp::SyncParametersClient>(node, "/dummy_joint_states");
         // read the list of plugin names for possible kinematics solvers
         for (std::size_t i = 0; i < known_groups.size(); ++i)
         {
           std::string base_param_name = known_groups[i].name_;
-          ROS_DEBUG_NAMED("kinematics_plugin_loader", "Looking for param %s ",
-                          (base_param_name + "/kinematics_solver").c_str());
+          RCLCPP_INFO(LOGGER_KINEMATICS_PLUGIN_LOADER, "Looking for param %s ", (base_param_name + ".kinematics_solver").c_str());
```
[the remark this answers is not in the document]
JafarAbdi: I think the previous getting the `kinematics_solver` was broken - could you please check the new changes .?

JafarAbdi:
This PR depends on [PR#104](https://github.com/ros-planning/moveit2/pull/104)

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -324,41 +332,51 @@ robot_model::SolverAllocatorFn KinematicsPluginLoader::getLoaderFunction(const s
...
+                        "Please remove the parameter '%s' from your configuration.",
+                        ksolver_attempts_param_name.c_str());
           }
 
           std::string ksolver_res_param_name;
-          if (nh.searchParam(base_param_name + "/kinematics_solver_search_resolution", ksolver_res_param_name))
+          if (ksolver_params->has_parameter(base_param_name + "/kinematics_solver_search_resolution"))
           {
+            ksolver_res_param_name =
+                node_->get_parameter(base_param_name + "/kinematics_solver_search_resolution").get_value<std::string>();
```
henningkayser: I'm pretty sure this doesn't work as intended. Here the parameter value is assigned to `ksolver_res_param_name` which is then again used as a key for parameter lookup. In the original implementation this made sense as `searchParam()` actually returns the parameter name. As described above, we can't do this here, so we just have to work with the result we got.

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -369,52 +387,66 @@ robot_model::SolverAllocatorFn KinematicsPluginLoader::getLoaderFunction(const s
             }
             else
             {  // handle the case this param is just one value and parsed as a double
-              double res;
-              if (nh.getParam(ksolver_res_param_name, res))
-                search_res[known_groups[i].name_].push_back(res);
-              else
+              if (ksolver_params->has_parameter(ksolver_res_param_name))
```
henningkayser: This condition is always false since we are in the `else` case of the upper condition.

henningkayser: We should probably break up this if/else condition and just query the parameter once. After that we can handle the different cases (string/int/double). Ideally, we should be able to just cast the int to double and only use one line.

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -369,52 +387,66 @@ robot_model::SolverAllocatorFn KinematicsPluginLoader::getLoaderFunction(const s
...
+                }
               }
             }
           }
 
           // Allow a kinematic solver's tip link to be specified on the rosparam server
           // Depreciated in favor of array version now
           std::string ksolver_ik_link_param_name;
-          if (nh.searchParam(base_param_name + "/kinematics_solver_ik_link", ksolver_ik_link_param_name))
+          if (ksolver_params->has_parameter(base_param_name + "/kinematics_solver_ik_link"))
```
henningkayser: Please remove, this parameter is deprecated

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -369,52 +387,66 @@ robot_model::SolverAllocatorFn KinematicsPluginLoader::getLoaderFunction(const s
...
+            ksolver_ik_links_param_name =
+                node_->get_parameter(base_param_name + "/kinematics_solver_ik_links").get_value<std::string>();
+            // XmlRpc::XmlRpcValue ksolver_ik_links;
+            std::vector<std::string> ksolver_ik_links;
+            if (ksolver_params->has_parameter(ksolver_ik_links_param_name))
             {
-              if (ksolver_ik_links.getType() != XmlRpc::XmlRpcValue::TypeArray)
+              ksolver_ik_links =
+                  node_->get_parameter(ksolver_ik_links_param_name).get_value<std::vector<std::string>>();
+              if (node_->get_parameter(ksolver_ik_links_param_name).get_type_name().find("array"))
```
henningkayser: This condition is flipped?

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -369,52 +387,66 @@ robot_model::SolverAllocatorFn KinematicsPluginLoader::getLoaderFunction(const s
...
+                            ksolver_ik_links_param_name.c_str());
               }
               else
               {
                 for (int32_t j = 0; j < ksolver_ik_links.size(); ++j)
                 {
-                  ROS_ASSERT(ksolver_ik_links[j].getType() == XmlRpc::XmlRpcValue::TypeString);
-                  ROS_DEBUG_STREAM_NAMED(LOGNAME, "found tip " << static_cast<std::string>(ksolver_ik_links[j])
-                                                               << " for group " << known_groups[i].name_);
+                  // ROS_ASSERT(ksolver_ik_links[j].getType() == XmlRpc::XmlRpcValue::TypeString);
```
henningkayser: I think we could just use the parameter type `PARAMETER_STRING_ARRAY` to implement this use case.

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -287,31 +294,32 @@ robot_model::SolverAllocatorFn KinematicsPluginLoader::getLoaderFunction(const s
...
+            RCLCPP_INFO(LOGGER_KINEMATICS_PLUGIN_LOADER, "Looking for param %s ", ksolver_param_name.c_str());
+            found = ksolver_params->has_parameter(ksolver_param_name);
           }
           if (found)
           {
-            ROS_DEBUG_NAMED(LOGNAME, "Found param %s", ksolver_param_name.c_str());
-            std::string ksolver;
-            if (nh.getParam(ksolver_param_name, ksolver))
+            RCLCPP_INFO(LOGGER_KINEMATICS_PLUGIN_LOADER, "Found param %s", ksolver_param_name.c_str());
+            auto parameters_and_prefixes = ksolver_params->get_parameters({ ksolver_param_name });
```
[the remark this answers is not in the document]
JafarAbdi: What's your opinion on setting ksolver directly .?
like `std::string ksolver = node_->get_parameter(ksolver_param_name).as_string();` 

when `found` is true `bool Node::get_parameter()` will return true, too - because in both cases the param is declared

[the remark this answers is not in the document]
henningkayser: You're right about the bool. I just had the template function in mind: `node_->get_parameter(ksolver_param_name,  ksolver)`. It seems like both the template function as well as `as_string()` would throw an exception if the type is not string.

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -287,78 +294,71 @@ robot_model::SolverAllocatorFn KinematicsPluginLoader::getLoaderFunction(const s
...
       if (default_solver_plugin_.empty())
       {
-        ROS_DEBUG_NAMED(LOGNAME, "Loading settings for kinematics solvers from the ROS param server ...");
-
-        // read data using ROS params
-        ros::NodeHandle nh("~");
+        RCLCPP_INFO(LOGGER_KINEMATICS_PLUGIN_LOADER,
+                    "Loading settings for kinematics solvers from the ROS param server ...");
 
+        auto ksolver_params = std::make_shared<rclcpp::SyncParametersClient>(node_);
```
henningkayser: Could you remove the `SyncParametersClient` and just use `rclcpp::Node` everywhere? It's much simpler than using a class we don't really need.

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -287,78 +294,71 @@ robot_model::SolverAllocatorFn KinematicsPluginLoader::getLoaderFunction(const s
...
         // read the list of plugin names for possible kinematics solvers
         for (std::size_t i = 0; i < known_groups.size(); ++i)
         {
           std::string base_param_name = known_groups[i].name_;
-          ROS_DEBUG_NAMED("kinematics_plugin_loader", "Looking for param %s ",
-                          (base_param_name + "/kinematics_solver").c_str());
-          std::string ksolver_param_name;
-          bool found = nh.searchParam(base_param_name + "/kinematics_solver", ksolver_param_name);
-          if (!found || !nh.hasParam(ksolver_param_name))
+          std::string ksolver_param_name = base_param_name + "/kinematics_solver";
```
henningkayser: This whole part seems overly complicated to me (it already was in the original implementation). But it's probably not the point for premature optimization here. Could you add a `TODO: revise parameter lookup` here so we can track what parameters we need to test/improve in the future?

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -367,58 +367,39 @@ robot_model::SolverAllocatorFn KinematicsPluginLoader::getLoaderFunction(const s
...
           // Allow a kinematic solver's tip links to be specified on the rosparam server as an array
-          std::string ksolver_ik_links_param_name;
-          if (nh.searchParam(base_param_name + "/kinematics_solver_ik_links", ksolver_ik_links_param_name))
+          std::string ksolver_ik_links_param_name = base_param_name + "/kinematics_solver_ik_links";
+          if (ksolver_params->has_parameter(ksolver_ik_links_param_name))
           {
-            XmlRpc::XmlRpcValue ksolver_ik_links;
-            if (nh.getParam(ksolver_ik_links_param_name, ksolver_ik_links))  // has custom rosparam-based tip link(s)
+            rclcpp::Parameter ksolver_ik_links_param = node_->get_parameter(ksolver_ik_links_param_name);
+            if (ksolver_ik_links_param.get_type() == rclcpp::ParameterType::PARAMETER_STRING_ARRAY)
```
henningkayser: Nice!

moveit_ros/planning/kinematics_plugin_loader/include/moveit/kinematics_plugin_loader/kinematics_plugin_loader.h
```diff
@@ -55,9 +55,10 @@ class KinematicsPluginLoader
       ROS parameter under which the robot description can be
       found. This is passed to the kinematics solver initialization as
       well as used to read the SRDF document when needed. */
-  KinematicsPluginLoader(const std::string& robot_description = "robot_description",
+  KinematicsPluginLoader(const rclcpp::Node::SharedPtr& node,
```
henningkayser: Please update documentation also.

moveit_ros/planning/kinematics_plugin_loader/include/moveit/kinematics_plugin_loader/kinematics_plugin_loader.h
```diff
@@ -68,10 +68,11 @@ class KinematicsPluginLoader
       parameter under which the robot description can be found. This
       is passed to the kinematics solver initialization as well as
       used to read the SRDF document when needed. */
-  KinematicsPluginLoader(const std::string& solver_plugin, double solve_timeout, unsigned int ik_attempts,
-                         const std::string& robot_description = "robot_description",
+  KinematicsPluginLoader(std::shared_ptr<rclcpp::Node> node, const std::string& solver_plugin, double solve_timeout,
```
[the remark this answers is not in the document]
henningkayser: @JafarAbdi Now that I've taken a better look at it: Nope, don't remove it ;) Only remove the the `ik_attempts` parameter.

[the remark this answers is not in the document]
henningkayser: Also, use `const rclcpp::Node::SharedPtr&` just as in the other constructor

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
JafarAbdi: Totally agreed with this point - I added the TODO but I still believe that we should use "." as a delimiter, for example, please take a look at how Navigation2 loads the parameters [yaml](https://github.com/ros-planning/navigation2/blob/6bfa8c48ea0291fff540eb7aead6e5e511130bd7/nav2_bringup/bringup/params/nav2_params.yaml#L113) - [cpp](https://github.com/ros-planning/navigation2/blob/6bfa8c48ea0291fff540eb7aead6e5e511130bd7/nav2_costmap_2d/plugins/obstacle_layer.cpp)

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -87,39 +87,39 @@ class KinematicsPluginLoader::KinematicsLoaderImpl
...
+                                                   "based on values in rosparam server.",
+                  jmg->getName().c_str());
       tips = ik_it->second;
     }
     else
     {
       // get the last link in the chain
-      ROS_DEBUG_STREAM_NAMED("kinematics_plugin_loader", "Choosing tip frame of kinematic solver for group "
-                                                             << jmg->getName() << " based on last link in chain");
+      RCLCPP_INFO(LOGGER_KINEMATICS_PLUGIN_LOADER, "Choosing tip frame of kinematic solver for group"
```
henningkayser:
```suggestion
      RCLCPP_INFO(LOGGER_KINEMATICS_PLUGIN_LOADER, "Choosing tip frame of kinematic solver for group %s"
```

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
henningkayser: I think the yaml you linked doesn't belong to the cpp file. What they did is use [dynamic_reconfigure](https://github.com/ros/dynamic_reconfigure) to generate the parameters: https://github.com/ros-planning/navigation2/blob/6bfa8c48ea0291fff540eb7aead6e5e511130bd7/nav2_costmap_2d/cfg/ObstaclePlugin.cfg. Normal parameters use forward slashes as namespace separators. https://design.ros2.org/articles/ros_parameters.html https://design.ros2.org/articles/topic_and_service_names.html.

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -87,39 +87,39 @@ class KinematicsPluginLoader::KinematicsLoaderImpl
...
+      RCLCPP_ERROR(LOGGER_KINEMATICS_PLUGIN_LOADER, "Error choosing kinematic solver tip frame(s).");
     }
 
     // Debug tip choices
     std::stringstream tip_debug;
     tip_debug << "Planning group '" << jmg->getName() << "' has tip(s): ";
-    for (const std::string& tip : tips)
-      tip_debug << tip << ", ";
-    ROS_DEBUG_STREAM_NAMED("kinematics_plugin_loader", tip_debug.str());
+    for (std::size_t i = 0; i < tips.size(); ++i)
```
henningkayser: Please revert to range-based loop

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -429,8 +410,8 @@ robot_model::SolverAllocatorFn KinematicsPluginLoader::getLoaderFunction(const s
       }
       else
       {
-        ROS_DEBUG_NAMED(LOGNAME, "Using specified default settings for kinematics solvers ...");
-        for (const srdf::Model::Group& known_group : known_groups)
+        RCLCPP_INFO(LOGGER_KINEMATICS_PLUGIN_LOADER, "Using specified default settings for kinematics solvers ...");
```
henningkayser: Please use `RCLCPP_DEBUG` as in the original implementation

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -42,12 +42,12 @@
...
 #include <map>
 #include <memory>
-#include <ros/ros.h>
+#include "rclcpp/rclcpp.hpp"
 #include <moveit/profiler/profiler.h>
 
 namespace kinematics_plugin_loader
 {
-static const std::string LOGNAME = "kinematics_plugin_loader";
+rclcpp::Logger LOGGER_KINEMATICS_PLUGIN_LOADER = rclcpp::get_logger("kinematics_plugin_loader");
```
henningkayser: Please apply the logging best practices https://github.com/ros-planning/moveit2/blob/b96a0259eb8d35364ab7d52f26fd50cdd45ee513/doc/MIGRATION_GUIDE.md

moveit_ros/planning/kinematics_plugin_loader/CMakeLists.txt
```diff
@@ -2,10 +2,19 @@ set(MOVEIT_LIB_NAME moveit_kinematics_plugin_loader)
 
 add_library(${MOVEIT_LIB_NAME} src/kinematics_plugin_loader.cpp)
```
henningkayser:
```suggestion
add_library(${MOVEIT_LIB_NAME} SHARED src/kinematics_plugin_loader.cpp)
```

PR #109, opened 2019-10-27
Port robot_model_loader to ROS2

JafarAbdi:
### Description

I didn't apply this [commit](https://github.com/AcutronicRobotics/moveit2/commit/fdaa83a6a5b02ad32ec84c7466f4da352ee89fac#diff-468186bdf99c33116a2d2ece4e3398c3) nor the changes to CMakeLists.txt from this [commit](https://github.com/AcutronicRobotics/moveit2/commit/5d645ee2e4c47fb4c5063b86482dba8909a40f0f#diff-468186bdf99c33116a2d2ece4e3398c3)

This PR depends on [PR#104](https://github.com/ros-planning/moveit2/pull/104) and [PR#107](https://github.com/ros-planning/moveit2/pull/107).

### DIFF
--- a/moveit_ros/planning/robot_model_loader/CMakeLists.txt
+++ b/moveit_ros/planning/robot_model_loader/CMakeLists.txt
@@ -4,12 +4,19 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
   add_compile_options(-Wno-potentially-evaluated-expression)
 endif()
 
-add_library(${MOVEIT_LIB_NAME} src/robot_model_loader.cpp)
+add_library(${MOVEIT_LIB_NAME} SHARED src/robot_model_loader.cpp)
 set_target_properties(${MOVEIT_LIB_NAME} PROPERTIES VERSION "${${PROJECT_NAME}_VERSION}")
-target_link_libraries(${MOVEIT_LIB_NAME} moveit_rdf_loader moveit_kinematics_plugin_loader ${catkin_LIBRARIES})
+ament_target_dependencies(${MOVEIT_LIB_NAME}
+  moveit_rdf_loader
+  moveit_kinematics_plugin_loader
+  rclcpp
+  urdf
+  Boost
+  moveit_core)
 
 install(TARGETS ${MOVEIT_LIB_NAME}
-        LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-        ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-        RUNTIME DESTINATION ${CATKIN_GLOBAL_BIN_DESTINATION})
-install(DIRECTORY include/ DESTINATION ${CATKIN_GLOBAL_INCLUDE_DESTINATION})
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
+  RUNTIME DESTINATION bin)
+
+install(DIRECTORY include/ DESTINATION include)
--- a/moveit_ros/planning/robot_model_loader/include/moveit/robot_model_loader/robot_model_loader.h
+++ b/moveit_ros/planning/robot_model_loader/include/moveit/robot_model_loader/robot_model_loader.h
@@ -77,9 +77,10 @@ public:
   };
 
   /** @brief Default constructor */
-  RobotModelLoader(const Options& opt = Options());
+  RobotModelLoader(const rclcpp::Node::SharedPtr& node, const Options& opt = Options());
 
-  RobotModelLoader(const std::string& robot_description, bool load_kinematics_solvers = true);
+  RobotModelLoader(const rclcpp::Node::SharedPtr& node, const std::string& robot_description,
+                   bool load_kinematics_solvers = true);
 
   ~RobotModelLoader();
 
@@ -131,5 +132,6 @@ private:
   robot_model::RobotModelPtr model_;
   rdf_loader::RDFLoaderPtr rdf_loader_;
   kinematics_plugin_loader::KinematicsPluginLoaderPtr kinematics_loader_;
+  const rclcpp::Node::SharedPtr node_;
 };
 }
--- a/moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
+++ b/moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
@@ -36,19 +36,23 @@
 
 #include <moveit/robot_model_loader/robot_model_loader.h>
 #include <moveit/profiler/profiler.h>
-#include <ros/ros.h>
+#include "rclcpp/rclcpp.hpp"
 #include <typeinfo>
 
 namespace robot_model_loader
 {
-RobotModelLoader::RobotModelLoader(const std::string& robot_description, bool load_kinematics_solvers)
+static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_ros.robot_model_loader");
+
+RobotModelLoader::RobotModelLoader(const rclcpp::Node::SharedPtr& node, const std::string& robot_description,
+                                   bool load_kinematics_solvers)
+  : node_(node)
 {
   Options opt(robot_description);
   opt.load_kinematics_solvers_ = load_kinematics_solvers;
   configure(opt);
 }
 
-RobotModelLoader::RobotModelLoader(const Options& opt)
+RobotModelLoader::RobotModelLoader(const rclcpp::Node::SharedPtr& node, const Options& opt) : node_(node)
 {
   configure(opt);
 }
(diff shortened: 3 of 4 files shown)


moveit_ros/planning/robot_model_loader/include/moveit/robot_model_loader/robot_model_loader.h
```diff
@@ -78,9 +78,10 @@ class RobotModelLoader
   };
 
   /** @brief Default constructor */
-  RobotModelLoader(const Options& opt = Options());
+  RobotModelLoader(std::shared_ptr<rclcpp::Node>& node, const Options& opt = Options());
```
henningkayser: Please make const (also member) if possible

moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
```diff
@@ -36,20 +36,25 @@
...
+
 namespace robot_model_loader
 {
-RobotModelLoader::RobotModelLoader(const std::string& robot_description, bool load_kinematics_solvers)
+RobotModelLoader::RobotModelLoader(std::shared_ptr<rclcpp::Node>& node, const std::string& robot_description,
+                                   bool load_kinematics_solvers)
 {
   Options opt(robot_description);
   opt.load_kinematics_solvers_ = load_kinematics_solvers;
+  this->node_ = node;
```
henningkayser: Just use copy constructor initialization `node_(node)`.

moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
```diff
@@ -87,12 +103,13 @@ void RobotModelLoader::configure(const Options& opt)
 {
   moveit::tools::Profiler::ScopedStart prof_start;
   moveit::tools::Profiler::ScopedBlock prof_block("RobotModelLoader::configure");
-
-  ros::WallTime start = ros::WallTime::now();
+  //
```
henningkayser: Comment?

moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
```diff
@@ -105,7 +122,8 @@ void RobotModelLoader::configure(const Options& opt)
     moveit::tools::Profiler::ScopedBlock prof_block2("RobotModelLoader::configure joint limits");
 
     // if there are additional joint limits specified in some .yaml file, read those in
-    ros::NodeHandle nh("~");
+    // TODO (anasarrak): Add the correct node name to the .yaml
+    auto additional_joints_parameters = std::make_shared<rclcpp::SyncParametersClient>(node_);
```
henningkayser: I don't see a reason to always initialize the `SyncParametersClient` for simple parameter lookup. We only use it for `has_parameter()` which is also provided by the node. Most queries here (and in the other classes) could be implemented using `bool Node::get_parameter()`, no? Especially in cases like here, we could always use the templated version of `get_parameter()` or `get_parameter_or()`. We only need to adjust for variable parameter types or catch the `rclcpp::ParameterTypeException`.

moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
```diff
@@ -160,8 +184,7 @@ void RobotModelLoader::configure(const Options& opt)
   if (model_ && opt.load_kinematics_solvers_)
     loadKinematicsSolvers();
 
-  ROS_DEBUG_STREAM_NAMED("robot_model_loader", "Loaded kinematic model in " << (ros::WallTime::now() - start).toSec()
-                                                                            << " seconds");
+  RCLCPP_DEBUG(node_->get_logger(), "Loaded kinematic model in %d seconds", (clock.now() - start).seconds());
```
henningkayser: `rclcpp::Time` is much cleaner than using `std::chrono` directly. We should apply this to other packages as well as long as there is no reason against it.

moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
```diff
@@ -181,9 +204,9 @@ void RobotModelLoader::loadKinematicsSolvers(const kinematics_plugin_loader::Kin
     const std::vector<std::string>& groups = kinematics_loader_->getKnownGroups();
     std::stringstream ss;
     std::copy(groups.begin(), groups.end(), std::ostream_iterator<std::string>(ss, " "));
-    ROS_DEBUG_STREAM("Loaded information about the following groups: '" << ss.str() << "'");
+    RCLCPP_DEBUG(LOGGER_ROBOT_MODEL_LOADER, "Loaded information about the following groups: '%S' ", ss.str().c_str());
```
henningkayser: Please change to '%s', '%S' is not a standard specifier. (https://en.cppreference.com/w/cpp/io/c/fprintf)

moveit_ros/planning/robot_model_loader/CMakeLists.txt
```diff
@@ -6,7 +6,15 @@ endif()
 
 add_library(${MOVEIT_LIB_NAME} src/robot_model_loader.cpp)
 set_target_properties(${MOVEIT_LIB_NAME} PROPERTIES VERSION "${${PROJECT_NAME}_VERSION}")
-target_link_libraries(${MOVEIT_LIB_NAME} moveit_rdf_loader moveit_kinematics_plugin_loader ${catkin_LIBRARIES})
+ament_target_dependencies(${MOVEIT_LIB_NAME}
```
henningkayser: There are some dependencies missing, i.e. `moveit_core`.

moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
```diff
@@ -36,19 +36,23 @@
 
 #include <moveit/robot_model_loader/robot_model_loader.h>
 #include <moveit/profiler/profiler.h>
-#include <ros/ros.h>
+#include "rclcpp/rclcpp.hpp"
 #include <typeinfo>
 
+rclcpp::Logger LOGGER_ROBOT_MODEL_LOADER = rclcpp::get_logger("planning").get_child("robot_model_loader");
```
henningkayser: Please move inside namespace

moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
```diff
@@ -116,42 +130,48 @@ void RobotModelLoader::configure(const Options& opt)
         std::string prefix = rdf_loader_->getRobotDescription() + "_planning/joint_limits/" + jlim[j].joint_name + "/";
 
         double max_position;
-        if (nh.getParam(prefix + "max_position", max_position))
+        if (node_->has_parameter(prefix + "max_position"))
         {
+          max_position = node_->get_parameter(prefix + "max_position").get_value<double>();
```
henningkayser: This can throw an exception, as `get_value<double>()` implicitly casts the parameters which could have the wrong type. So it probably makes sense to wrap all of this in a try/catch (probably a big one is enough) and print an error mentioning the failed parameter name. Also, it would be shorter to just use the template function `bool Node::get_parameter(param_name, result)`.

moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
```diff
@@ -115,43 +128,63 @@ void RobotModelLoader::configure(const Options& opt)
...
+          param_name = prefix + "has_acceleration_limits";
+          bool has_acc_limits;
+          if (node_->get_parameter(param_name, has_acc_limits))
+            jlim[j].has_acceleration_limits = has_acc_limits;
+        }
+        catch (const rclcpp::ParameterTypeException& e)
         {
-          jlim[j].has_acceleration_limits = true;
-          jlim[j].max_acceleration = max_acc;
+          RCLCPP_WARN(LOGGER, "When getting the parameter %s: %s", param_name.c_str(), e.what());
```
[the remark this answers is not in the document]
henningkayser: I agree it's an `ERROR` since it detects faulty parameter types in the otherwise valid yaml and requires fixup. Also, missing parameters are still supported like before. I'm just concerned that there is currently no way to handle the error (i.e. returning `false`) and terminating the node/session. But this is still better than silently ignoring faulty values as in the original implementation.

moveit_ros/planning/robot_model_loader/CMakeLists.txt
```diff
@@ -6,7 +6,17 @@ endif()
 
 add_library(${MOVEIT_LIB_NAME} src/robot_model_loader.cpp)
```
henningkayser:
```suggestion
add_library(${MOVEIT_LIB_NAME} SHARED src/robot_model_loader.cpp)
```

moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
```diff
@@ -204,13 +236,13 @@ void RobotModelLoader::loadKinematicsSolvers(const kinematics_plugin_loader::Kin
         }
         else
         {
-          ROS_ERROR("Kinematics solver %s does not support joint group %s.  Error: %s", typeid(*solver).name(),
-                    group.c_str(), error_msg.c_str());
+          RCLCPP_ERROR(LOGGER, "Kinematics solver %s does not support joint group %s.  Error: %s",
+                       typeid(*solver).name(), groups[i].c_str(), error_msg.c_str());
```
henningkayser: With the range-based loop it should be `group` instead of `groups[i]`

moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
```diff
@@ -105,53 +120,71 @@ void RobotModelLoader::configure(const Options& opt)
...
+            jlim[j].has_acceleration_limits = true;
+            jlim[j].max_acceleration = max_acc;
+          }
+
+          param_name = prefix + "has_acceleration_limits";
+          bool has_acc_limits;
+          if (node_->get_parameter(param_name, has_acc_limits))
+            jlim[j].has_acceleration_limits = has_acc_limits;
+        }
+        catch (const rclcpp::ParameterTypeException& e)
```
henningkayser: We should probably add this pattern to as parameter best practices to the migration guide https://github.com/ros-planning/moveit2/blob/b96a0259eb8d35364ab7d52f26fd50cdd45ee513/doc/MIGRATION_GUIDE.md

moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
```diff
@@ -105,53 +120,71 @@ void RobotModelLoader::configure(const Options& opt)
     moveit::tools::Profiler::ScopedBlock prof_block2("RobotModelLoader::configure joint limits");
 
     // if there are additional joint limits specified in some .yaml file, read those in
-    ros::NodeHandle nh("~");
-
-    for (moveit::core::JointModel* joint_model : model_->getJointModels())
+    for (std::size_t i = 0; i < model_->getJointModels().size(); ++i)
```
henningkayser: Please revert to range-based loop

PR #115, opened 2019-10-28
[Don't Merge!] - Diff with Acutronic's fork

JafarAbdi:
Exluded 
- `moveit_core` because it could merged. 
- `moveit_experimental`


moveit_kinematics/kdl_kinematics_plugin/src/kdl_kinematics_plugin.cpp
```diff
@@ -553,11 +549,16 @@ bool KDLKinematicsPlugin::getPositionFK(const std::vector<std::string>& link_nam
   {
     if (fk_solver_->JntToCart(jnt_pos_in, p_out) >= 0)
     {
-      poses[i] = tf2::toMsg(p_out);
+      // poses[i] = tf2::toMsg(p_out);
+      poses[i].position.x = p_out.p[0];
```
henningkayser: This should be available in https://github.com/ros2/geometry2/tree/dashing

moveit_kinematics/lma_kinematics_plugin/include/moveit/lma_kinematics_plugin/lma_kinematics_plugin.h
```diff
@@ -129,13 +129,13 @@ class LMAKinematicsPlugin : public kinematics::KinematicsBase
...
    */
-  bool searchPositionIK(const geometry_msgs::Pose& ik_pose, const std::vector<double>& ik_seed_state, double timeout,
+  bool searchPositionIK(const geometry_msgs::msg::Pose& ik_pose, const std::vector<double>& ik_seed_state, double timeout,
                         std::vector<double>& solution, const IKCallbackFn& solution_callback,
                         moveit_msgs::msg::MoveItErrorCodes& error_code, const std::vector<double>& consistency_limits,
                         const kinematics::KinematicsQueryOptions& options = kinematics::KinematicsQueryOptions()) const;
 
 private:
-  bool timedOut(const ros::WallTime& start_time, double duration) const;
+  bool timedOut(const std::chrono::system_clock::time_point& start_time, double duration) const;
```
henningkayser: We should decide whether we prefer std::chrono or rclcpp::Time variants.

moveit_kinematics/srv_kinematics_plugin/include/moveit/srv_kinematics_plugin/srv_kinematics_plugin.h
```diff
@@ -156,7 +156,10 @@ class SrvKinematicsPlugin : public kinematics::KinematicsBase
 
   int num_possible_redundant_joints_;
 
-  std::shared_ptr<ros::ServiceClient> ik_service_client_;
+  rclcpp::Client<moveit_msgs::srv::GetPositionIK>::SharedPtr ik_service_client_;
```
henningkayser: Do we always have to declare the message type in advance?

moveit_plugins/moveit_fake_controller_manager/src/moveit_fake_controller_manager.cpp
```diff
@@ -49,91 +47,126 @@ namespace moveit_fake_controller_manager
 static const std::string DEFAULT_TYPE = "interpolate";
 static const std::string ROBOT_DESCRIPTION = "robot_description";
 
+struct Controller_list {
```
henningkayser: This changed a lot. We'll have to check if we can use all of these changes

henningkayser: We should probably move this to the back and use a sim controller directly until then (i.e. using ros1_bridge)

moveit_ros/manipulation/move_group_pick_place_capability/src/pick_place_action_capability.h
```diff
@@ -38,67 +38,90 @@
 #define MOVEIT_MOVE_GROUP_PICK_PLACE_ACTION_CAPABILITY_
```
henningkayser: I'm not sure we want to keep this capability as is.

PR #123, opened 2019-11-19
Add locale install to Dockerfile

RoboticsYY:
`srdfdom` requires the locale `nl_NL.UTF-8` on the system. Current CI reports it is missing: 

```cpp
--- stderr: srdfdom
CMake Warning at CMakeLists.txt:67 (message):
  Locale nl_NL not available.  Locale test will not be meaningful.
```

This PR intends to add the locale install to the Dockerfile to suppress this warning.

### DIFF
--- a/.docker/ci/Dockerfile
+++ b/.docker/ci/Dockerfile
@@ -35,6 +35,16 @@ RUN apt-get -qq update && \
     # Clear apt-cache to reduce image size
     rm -rf /var/lib/apt/lists/*
 
+# Required for srdfdom compilation
+# TODO: Remove when srdfdom becomes a system dependency
+RUN apt-get -qq update && \
+    apt-get -qq install -y \
+        locales && \
+    # Clear apt-cache to reduce image size
+    rm -rf /var/lib/apt/lists/* && \
+    # Update system locale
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
Specifically: it's used in [this test](https://github.com/ros-planning/srdfdom/blob/1fb929e4aab58f5fe889d4bf2b4e330423605632/test/srdf_parser_cpp.test) to check for `,` vs `.` issues in the parser.

RoboticsYY:
@rhaschke I agree that install the locale in the moveit docker image may not be a good option. I think there is no need to do this in ROS1, because srdfdom is a system dependency, the locale install can be done in srdfdom's Travis script. Currently, srdfdom is built with moveit2 in the same colcon workspace, I observed this warning in the [CI report](https://travis-ci.org/ros-planning/moveit2/jobs/612461783#L638). No, I think CI can still ignore this warning and pass. This PR just attempts to provide a temporary workaround to suppress the warning.

rhaschke:
It's true that the colcon build spits warnings for diverse packages. However, Travis only complains about warnings in `moveit_core`, which is the only package part of the tested repository. So, everything is fine :smile:

PR #140, opened 2019-12-07
WIP: cleanup

rhaschke:
Several improvements to the ROS2 code migration. Additional changes are required:
- [use of `rclcpp::Clock` for `ros::Time`](https://answers.ros.org/question/287946/ros-2-time-handling)
- ...

### DIFF
--- a/moveit_core/background_processing/src/background_processing.cpp
+++ b/moveit_core/background_processing/src/background_processing.cpp
@@ -35,7 +35,7 @@
 /* Author: Ioan Sucan */
 
 #include <moveit/background_processing/background_processing.h>
-#include "rclcpp/rclcpp.hpp"
+#include <rclcpp/rclcpp.hpp>
 
 namespace moveit
 {
--- a/moveit_core/background_processing/src/background_processing.cpp
+++ b/moveit_core/background_processing/src/background_processing.cpp
@@ -42,7 +42,7 @@ namespace moveit
 namespace tools
 {
 // Logger
-static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_background_processing.background_processing");
+static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit.background_processing.background_processing");
 
 BackgroundProcessing::BackgroundProcessing()
 {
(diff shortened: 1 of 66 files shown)


moveit_core/background_processing/src/background_processing.cpp
```diff
@@ -35,14 +35,14 @@
...
 #include <moveit/background_processing/background_processing.h>
-#include "rclcpp/rclcpp.hpp"
+#include <rclcpp/logging.hpp>
 
 namespace moveit
 {
 namespace tools
 {
 // Logger
-static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_background_processing.background_processing");
```
henningkayser: We should use a systematic naming convention for this (see https://github.com/ros-planning/moveit2/blob/master/doc/MIGRATION_GUIDE.md). Right now we decided to use the library name only (here `moveit_background_processing`) without a separator if the file name is identical. I agree that having a separator would make sense for scoping logger levels. Also, I would argue for including the package name, i.e. `moveit.core.background_processing` following the pattern `moveit.<package>.<library>.<file>` where we drop `moveit` from `<package>.<library>.<file>` and remove redundancies in the namespace if there are any. What do you think of this approach?

moveit_core/background_processing/src/background_processing.cpp
```diff
@@ -35,14 +35,14 @@
 /* Author: Ioan Sucan */
 
 #include <moveit/background_processing/background_processing.h>
-#include "rclcpp/rclcpp.hpp"
```
henningkayser: Yes, this was also on my list ;)

moveit_core/background_processing/src/background_processing.cpp
rhaschke: Adding the package name is a good idea. My main aim was to introduce the top-level `moveit` scope.

mlautman:
correction, removing that line doesn't fix it. You may need to do something else

RoboticsYY:
Hi @rhaschke, I noticed this PR is still titled as WIP. Are you planning to continue this PR? Otherwise, I would like to continue this effort.

rhaschke:
@RoboticsYY, please feel free to continue. I don't have the time to work on ROS2.

rhaschke:
Closing in favor of #170.

PR #153, opened 2019-12-30
Upgrade docker images to ROS 2 Eloquent

henningkayser:
I switched to eloquent as default right away as we are dropping support for dashing in the future.

### DIFF
--- a/.docker/source/Dockerfile
+++ b/.docker/source/Dockerfile
@@ -30,7 +30,7 @@ RUN apt-get -qq update && \
     rm -rf /var/lib/apt/lists/*
 
 # Build the workspace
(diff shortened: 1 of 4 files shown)


.docker/source/Dockerfile
```diff
@@ -30,7 +30,7 @@ RUN apt-get -qq update && \
     rm -rf /var/lib/apt/lists/*
 
 # Build the workspace
-RUN . /opt/ros/dashing/setup.sh &&\
+RUN . /opt/ros/eloquent/setup.sh &&\
```
rhaschke:
```suggestion
RUN . /opt/ros/${ROS_DISTRO}/setup.sh &&\
```

PR #162, opened 2020-01-29
Fix and compile planning_pipeline

henningkayser:
This enables the planning_pipeline and fixes remaining migration steps.

### DIFF
--- a/moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
+++ b/moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
@@ -42,6 +42,8 @@
 #include <boost/algorithm/string/join.hpp>
 #include <sstream>
 
+static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit.ros_planning.planning_pipeline");
+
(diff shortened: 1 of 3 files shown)


moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
```diff
@@ -215,10 +212,7 @@ void planning_pipeline::PlanningPipeline::checkSolutionPaths(bool flag)
     contacts_publisher_.reset();
   else if (!check_solution_paths_ && flag)
   {
-    custom_qos_profile_ = rmw_qos_profile_default;
-    custom_qos_profile_.depth = 100;
-    contacts_publisher_ =
-        node_->create_publisher<visualization_msgs::msg::MarkerArray>(MOTION_CONTACTS_TOPIC, custom_qos_profile_);
+    contacts_publisher_ = node_->create_publisher<visualization_msgs::msg::MarkerArray>(MOTION_CONTACTS_TOPIC, 10);
```
mamoll: not 100, like before?

henningkayser: "before" was already a porting attempt (that to my knowledge never ran in practice) and I don't see a reason to set the value that high here.

PR #163, opened 2020-01-29
Port moveit_cpp to ROS 2

henningkayser:


moveit_ros/planning_interface/CMakeLists.txt
```diff
@@ -8,88 +8,89 @@ if(NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
...
+  tf2_geometry_msgs
+)
+
+#catkin_python_setup()
+
 include_directories(${THIS_PACKAGE_INCLUDE_DIRS}
-                    ${Boost_INCLUDE_DIRS}
-                    ${catkin_INCLUDE_DIRS})
+                    ${rclcpp_INCLUDE_DIRS}
+                    ${Boost_INCLUDE_DIRS})
```
JafarAbdi: Shouldn't these be included automatically when using `ament_target_dependencies` .?

moveit_ros/planning_interface/moveit_cpp/src/moveit_cpp.cpp
```diff
@@ -36,40 +36,36 @@
...
+MoveItCpp::MoveItCpp(const Options& options, const rclcpp::Node::SharedPtr& node,
                      const std::shared_ptr<tf2_ros::Buffer>& tf_buffer)
-  : tf_buffer_(tf_buffer)
+  : node_(node), tf_buffer_(tf_buffer)
 {
+  clock_ = std::make_shared<rclcpp::Clock>();
   if (!tf_buffer_)
-    tf_buffer_.reset(new tf2_ros::Buffer());
-  tf_listener_.reset(new tf2_ros::TransformListener(*tf_buffer_));
+    tf_buffer_ = std::make_shared<tf2_ros::Buffer>(clock_);
```
JafarAbdi: Nit: We could use the node's internal clock `node->get_clock();`

moveit_ros/planning_interface/moveit_cpp/src/planning_component.cpp
```diff
@@ -38,59 +38,50 @@
...
-    ROS_FATAL_STREAM_NAMED(LOGNAME, error);
+    RCLCPP_FATAL_STREAM(LOGGER, error);
     throw std::runtime_error(error);
   }
   planning_pipeline_names_ = moveit_cpp_->getPlanningPipelineNames(group_name);
 }
 
-PlanningComponent::PlanningComponent(const std::string& group_name, const ros::NodeHandle& nh)
-  : nh_(nh), moveit_cpp_(new MoveItCpp(nh)), group_name_(group_name)
+PlanningComponent::PlanningComponent(const std::string& group_name, const rclcpp::Node::SharedPtr& node)
```
JafarAbdi: Nit: For consistency in all MoveIt 2 packages porting we made `node` the first parameter

moveit_ros/planning_interface/package.xml
```diff
@@ -15,28 +15,37 @@
...
-  <depend>moveit_ros_manipulation</depend>
-  <depend>roscpp</depend>
-  <depend>rospy</depend>
-  <depend>rosconsole</depend>
-  <depend>actionlib</depend>
+  <!-- <depend>moveit_ros_warehouse</depend> -->
+  <!-- <depend>moveit_ros_move_group</depend> -->
+  <!-- <depend>moveit_ros_manipulation</depend> -->
+  <depend>rclcpp</depend>
+  <!-- <depend>rospy</depend> -->
```
JafarAbdi:
```suggestion
  <depend>rclpy</depend>
```

moveit_ros/planning_interface/package.xml
```diff
@@ -15,28 +15,37 @@
...
+  <!-- <depend>rosconsole</depend> -->
```
JafarAbdi: This should be removed

moveit_ros/planning_interface/CMakeLists.txt
```diff
@@ -8,88 +8,89 @@ if(NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
...
+#find_package(PythonInterp REQUIRED)
+#find_package(PythonLibs "${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}" REQUIRED)
 
 find_package(Boost REQUIRED)
-if(Boost_VERSION LESS 106700)
-  set(BOOST_PYTHON_COMPONENT python)
-elseif()
-  set(BOOST_PYTHON_COMPONENT python${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR})
-endif()
+#if(Boost_VERSION LESS 106700)
```
mlautman: I don't think we need boost for pointers anymore. Remove?

moveit_ros/planning_interface/CMakeLists.txt
mlautman: +1

moveit_ros/planning_interface/moveit_cpp/CMakeLists.txt
```diff
@@ -3,17 +3,18 @@ set(MOVEIT_LIB_NAME moveit_cpp)
...
 set_target_properties(${MOVEIT_LIB_NAME} PROPERTIES VERSION "${${PROJECT_NAME}_VERSION}")
-target_link_libraries(${MOVEIT_LIB_NAME}
-  moveit_common_planning_interface_objects
-  moveit_planning_scene_interface
-  ${catkin_LIBRARIES}
-  ${Boost_LIBRARIES})
-add_dependencies(${MOVEIT_LIB_NAME} ${catkin_EXPORTED_TARGETS})
+ament_target_dependencies(${MOVEIT_LIB_NAME}
+  rclcpp
+  Boost
```
mlautman: Same as before. (Do we still need boost?)

moveit_ros/planning_interface/CMakeLists.txt
mlautman: ..or is this for boost binding c++ methods to python ones?

moveit_ros/planning_interface/CMakeLists.txt
```diff
@@ -8,88 +8,89 @@ if(NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
...
+
+#add_subdirectory(py_bindings_tools)
+#add_subdirectory(common_planning_interface_objects)
+#add_subdirectory(planning_scene_interface)
+#add_subdirectory(move_group_interface)
+#add_subdirectory(robot_interface)
+#add_subdirectory(test)
 add_subdirectory(moveit_cpp)
+
+ament_export_include_directories(include)
```
mlautman: Does this directory exist at `moveit_ros/planning_interface`?

moveit_ros/planning_interface/CMakeLists.txt
henningkayser: This is from the python wrapper which we didn't touch yet

moveit_ros/planning_interface/moveit_cpp/src/planning_component.cpp
henningkayser: I put it second place because I think we can make it optional in the future

moveit_ros/planning_interface/CMakeLists.txt
henningkayser: No, it exports the shared include directory

moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
```diff
@@ -50,28 +50,28 @@ namespace moveit
...
+MOVEIT_CLASS_FORWARD(MoveItCpp)
 
 class MoveItCpp
 {
 public:
   /// Specification of options to use when constructing the MoveItCpp class
   struct PlanningSceneMonitorOptions
   {
-    void load(const ros::NodeHandle& nh)
+    void load(const rclcpp::Node::SharedPtr& node)
```
mlautman: Any reason to not use the templated approach that is popular in the ros2 code base? 

```   
template<class NodeT>
void load(NodeT && node){
    return load(node->get_node_parameters_interface())
}
void load(rclcpp::node_interfaces::NodeParametersInterface::SharedPtr& node_parameters_interface){
    ...
}
```

moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
```diff
@@ -85,11 +85,11 @@ class MoveItCpp
   /// struct contains the the variables used for loading the planning pipeline
   struct PlanningPipelineOptions
   {
-    void load(const ros::NodeHandle& nh)
+    void load(const rclcpp::Node::SharedPtr& node)
```
mlautman: same

moveit_ros/planning_interface/CMakeLists.txt
henningkayser: It's the install directory. See 
https://github.com/ament/ament_cmake/blob/master/ament_cmake_export_include_directories/cmake/ament_export_include_directories.cmake#L37

moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
```diff
@@ -98,19 +98,20 @@ class MoveItCpp
   /// Parameter container for initializing MoveItCpp
   struct Options
   {
-    Options(const ros::NodeHandle& nh)
+    Options(const rclcpp::Node::SharedPtr& node)
```
mlautman: same

moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
```diff
@@ -98,19 +98,20 @@ class MoveItCpp
...
     }
 
     PlanningSceneMonitorOptions planning_scene_monitor_options;
     PlanningPipelineOptions planning_pipeline_options;
   };
 
   /** \brief Constructor */
-  MoveItCpp(const ros::NodeHandle& nh, const std::shared_ptr<tf2_ros::Buffer>& tf_buffer = {});
-  MoveItCpp(const Options& options, const ros::NodeHandle& nh, const std::shared_ptr<tf2_ros::Buffer>& tf_buffer = {});
+  MoveItCpp(const rclcpp::Node::SharedPtr& node, const std::shared_ptr<tf2_ros::Buffer>& tf_buffer = {});
```
mlautman: same.

moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
```diff
@@ -162,16 +163,16 @@ class MoveItCpp
...
                bool blocking = true);
 
-protected:
-  std::shared_ptr<tf2_ros::Buffer> tf_buffer_;
-  std::shared_ptr<tf2_ros::TransformListener> tf_listener_;
-
 private:
   //  Core properties and instances
-  ros::NodeHandle node_handle_;
+  rclcpp::Node::SharedPtr node_;
```
mlautman: It may be better to save the node by it's components rather than the full node_. That seems to be the approach used by all non-executables

moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
henningkayser: Right now this is just a migration step so I didn't consider it. We're planning to make a bigger refactoring for parameter loading that includes declaring all parameters that we use. However, I don't see a real upside to use the template version, why do you think we should use it?

moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
```diff
@@ -130,7 +131,7 @@ class MoveItCpp
   robot_model::RobotModelConstPtr getRobotModel() const;
 
   /** \brief Get the ROS node handle of this instance operates on */
-  const ros::NodeHandle& getNodeHandle() const;
+  const rclcpp::Node::SharedPtr& getNode() const;
```
mlautman: ahh. Now I understand. If you need to get the node out than you have to have the full shared ptr.

moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
mlautman: nevermind. (see below)

moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
mlautman: nevermind. (see below)

moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
mlautman: nevermind. (see below)

moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
```diff
@@ -115,7 +113,7 @@ class PlanningComponent
   };
 
   /** \brief Constructor */
-  PlanningComponent(const std::string& group_name, const ros::NodeHandle& nh);
+  PlanningComponent(const std::string& group_name, const rclcpp::Node::SharedPtr& node);
```
mlautman: template?

moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
henningkayser: Again, what's the purpose? We are currently sharing the node with all member classes and the use is pretty diverse, so I don't see any good way to restrict the parameters. This belongs to node lifetime/handling efforts we didn't consider for this port.

moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
```diff
@@ -186,7 +184,7 @@ class PlanningComponent
 
 private:
   // Core properties and instances
-  ros::NodeHandle nh_;
+  rclcpp::Node::SharedPtr node_;
```
mlautman: break into components?

moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
henningkayser: Yep ;)

moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
henningkayser: Nope, explanation above

moveit_ros/planning_interface/moveit_cpp/src/moveit_cpp.cpp
```diff
@@ -36,40 +36,35 @@
...
-#include <ros/console.h>
-#include <ros/ros.h>
+#include <rclcpp/rclcpp.hpp>
 
 namespace moveit
 {
 namespace planning_interface
 {
-constexpr char LOGNAME[] = "moveit_cpp";
+static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit.ros_planning_interface.moveit_cpp");
```
mlautman: nit : it may be nice to make the logger a MoveItCpp class variable so that you can namespace it below the name of the Node that gets passed in.

henningkayser:
@mlautman I think you're right about using the node templates for end consumers, but here we are using and sharing (probably) all components of the node. We would have to start by refactoring node use in all used components and then work our way up until these base classes. A big downside of this approach would be that we would heavily restrict what we can do inside the member classes, so for now I wouldn't go this path.

moveit_ros/planning_interface/moveit_cpp/src/moveit_cpp.cpp
henningkayser: Worth a discussion, but for the release this won't happen as this follows guidelines we decided upon for the migration.

moveit_ros/planning_interface/moveit_cpp/src/moveit_cpp.cpp
```diff
@@ -36,40 +36,35 @@
...
-MoveItCpp::MoveItCpp(const Options& options, const ros::NodeHandle& /*unused*/,
+MoveItCpp::MoveItCpp(const rclcpp::Node::SharedPtr& node, const Options& options,
                      const std::shared_ptr<tf2_ros::Buffer>& tf_buffer)
-  : tf_buffer_(tf_buffer)
+  : node_(node), tf_buffer_(tf_buffer)
 {
   if (!tf_buffer_)
-    tf_buffer_.reset(new tf2_ros::Buffer());
-  tf_listener_.reset(new tf2_ros::TransformListener(*tf_buffer_));
+    tf_buffer_ = std::make_shared<tf2_ros::Buffer>(node_->get_clock());
```
mlautman: Something to consider: You may need to establish what time of clock is used by the tf_bufffer. Most every example I found in the ros2 code base uses `auto clock = std::make_shared<rclcpp::Clock>(RCL_ROS_TIME))` to ensure that the TF buffer uses the correct clock

moveit_ros/planning_interface/moveit_cpp/src/moveit_cpp.cpp
```diff
@@ -155,22 +150,22 @@ bool MoveItCpp::loadPlanningSceneMonitor(const PlanningSceneMonitorOptions& opti
 
 bool MoveItCpp::loadPlanningPipelines(const PlanningPipelineOptions& options)
 {
-  ros::NodeHandle node_handle(options.parent_namespace.empty() ? "~" : options.parent_namespace);
+  // ros::NodeHandle node_handle(options.parent_namespace.empty() ? "~" : options.parent_namespace);
```
mlautman: delete

moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
mlautman: It would be nice to be able to use weak pointers but to be honest, I really don't see that as a strong enough use case for going back and templating everything

moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
mlautman: Fair enough

mlautman:
+1 feel free to ignore my comments about templates and node components in other PR's as well.

moveit_ros/planning_interface/moveit_cpp/src/moveit_cpp.cpp
henningkayser: That's actually a good point. I wasn't aware that system time was the default in the clock constructor. However, nodes always use ROS time, so using `Node::get_clock()` and `Node::now()` is valid everywhere you want to use ROS time.

moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
henningkayser: Ok, I see that it can make sense of you structure your library around the node components (which would actually be very nice) but introducing this in an existing library is probably not that easy. I would be interested in some discussions how we can improve node management in MoveIt for later milestones.

moveit_ros/planning_interface/moveit_cpp/src/moveit_cpp.cpp
henningkayser: Added a TODO to use the parent namespace for parameter lookup (I fix this in a follow-up PR with additional parameter refactoring)

henningkayser:
@mlautman I addressed all comments and rebased onto master

moveit_ros/planning_interface/test/CMakeLists.txt
```diff
@@ -1,15 +1,16 @@
...
-  add_executable(test_cleanup test_cleanup.cpp)
-  target_link_libraries(test_cleanup moveit_move_group_interface)
+  # add_executable(test_cleanup test_cleanup.cpp)
+  # target_link_libraries(test_cleanup moveit_move_group_interface)
 
-  add_rostest_gtest(moveit_cpp_test moveit_cpp_test.test moveit_cpp_test.cpp)
-  target_link_libraries(moveit_cpp_test moveit_cpp ${catkin_LIBRARIES})
+  ament_add_gtest(moveit_cpp_test moveit_cpp_test.test moveit_cpp_test.cpp)
+  #ament_target_dependencies(moveit_cpp_test rclcpp)
+  target_link_libraries(moveit_cpp_test moveit_cpp)
```
nbbrooks: moveit_cpp isn't defined in this cmake so shouldn't it be listed under ament_target_dependencies instead?

henningkayser: moveit_cpp is a library that is defined in the same package (this CMakeLists.txt is just included by the parent one). So linking the library here is fine. However, the test is not enabled yet since runtime tests only work with an initialized planning pipeline, making OMPL a dependency.

nbbrooks: Neat!

PR #167, opened 2020-02-04
Fix moveit visualization

JafarAbdi:
### Description

This fixes the undefined symbol error we were getting when loading the RobotState rviz2 plugin

Now loading RobotState plugin gives, which I think is related to rdf_loader (not sure)

```
Thread 1 "rviz2" received signal SIGSEGV, Segmentation fault.
0x00007ffff65af330 in rclcpp::Node::has_parameter(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) const () from /opt/ros/eloquent/lib/librclcpp.so
```

And loading PlanningSene plugin gives  
```
[INFO] [moveit_rdf_loader.rdf_loader]: Robot model parameter not found! Did you remap 'robot_description'?
[ERROR] [moveit_ros.planning_scene_monitor.planning_scene_monitor]: Robot model not loaded
```
### Checklist
- [ ] Document API changes relevant to the user in the MIGRATION.md notes
- [ ] Create tests, which fail without this PR [reference](https://ros-planning.github.io/moveit_tutorials/)

### DIFF
--- a/moveit_ros/planning/rdf_loader/src/rdf_loader.cpp
+++ b/moveit_ros/planning/rdf_loader/src/rdf_loader.cpp
@@ -62,68 +62,51 @@ RDFLoader::RDFLoader(const std::shared_ptr<rclcpp::Node>& node, const std::strin
   moveit::tools::Profiler::ScopedStart prof_start;
   moveit::tools::Profiler::ScopedBlock prof_block("RDFLoader(robot_description)");
 
-  auto start = std::chrono::system_clock::now();
+  auto start = node->now();
 
-  std::string content;
-  // TODO(JafarAbdi): Revise parameter lookup
-  if (node->has_parameter(robot_description))
-  {
-    rclcpp::Parameter robot_description_param = node->get_parameter(robot_description);
-    try
-    {
-      content = robot_description_param.as_string();
-    }
-    catch (const rclcpp::ParameterTypeException& e)
-    {
-      RCLCPP_WARN(LOGGER, "When getting robot_description parameter %s", e.what());
-    }
-  }
-  if (content.empty())
+  // Check if the robot_description parameter is declared, declare it if it's not declared yet
+  if (!node->has_parameter(robot_description))
+    node->declare_parameter(robot_description);
+  std::string robot_description_content;
+  node->get_parameter_or(robot_description, robot_description_content, std::string());
+
+  if (robot_description_content.empty())
   {
     RCLCPP_INFO_ONCE(LOGGER, "Robot model parameter not found! Did you remap '%s'?\n", robot_description.c_str());
     return;
   }
 
   urdf::Model* umodel = new urdf::Model();
-  if (!umodel->initString(content))
+  if (!umodel->initString(robot_description_content))
   {
-    RCLCPP_INFO(LOGGER, "Unable to parse URDF from parameter: '%s'", robot_description_.c_str());
+    RCLCPP_INFO(LOGGER, "Unable to parse URDF from parameter: '%s'", robot_description.c_str());
     return;
   }
   urdf_.reset(umodel);
 
-  const std::string srdf_description(robot_description_ + "_semantic");
-  std::string scontent;
-  // TODO(JafarAbdi): Revise parameter lookup
-  if (node->has_parameter(srdf_description))
-  {
-    rclcpp::Parameter srdf_description_param = node->get_parameter(srdf_description);
-    try
-    {
-      content = srdf_description_param.as_string();
-    }
-    catch (const rclcpp::ParameterTypeException& e)
-    {
-      RCLCPP_WARN(LOGGER, "When getting robot_description_semantic parameter: %s", e.what());
-    }
-  }
-  if (scontent.empty())
+  const std::string srdf_description = robot_description + "_semantic";
+  // Check if the robot_description_semantic parameter is declared, declare it if it's not declared yet
+  if (!node->has_parameter(srdf_description))
+    node->declare_parameter(srdf_description);
+  std::string srdf_description_content;
+  node->get_parameter_or(srdf_description, srdf_description_content, std::string());
+
+  if (srdf_description_content.empty())
   {
     RCLCPP_INFO_ONCE(LOGGER, "Robot semantic description not found. Did you forget to define or remap '%s'?\n",
-                     std::string(robot_description + "_semantic").c_str());
+                     srdf_description.c_str());
     return;
   }
 
   srdf_.reset(new srdf::Model());
-  if (!srdf_->initString(*urdf_, scontent))
+  if (!srdf_->initString(*urdf_, srdf_description_content))
   {
     RCLCPP_ERROR(LOGGER, "Unable to parse SRDF from parameter '%s'", srdf_description.c_str());
     srdf_.reset();
     return;
   }
 
-  RCLCPP_INFO(LOGGER, "Loaded robot model in %ld seconds",
-              std::chrono::duration_cast<std::chrono::seconds>(std::chrono::system_clock::now() - start).count());
+  RCLCPP_INFO_STREAM(LOGGER, "Loaded robot model in " << (node->now() - start).seconds() << " seconds");
 }
 
 RDFLoader::RDFLoader(const std::string& urdf_string, const std::string& srdf_string)
--- a/moveit_ros/visualization/planning_scene_rviz_plugin/include/moveit/planning_scene_rviz_plugin/planning_scene_display.h
+++ b/moveit_ros/visualization/planning_scene_rviz_plugin/include/moveit/planning_scene_rviz_plugin/planning_scene_display.h
@@ -122,7 +122,7 @@ protected Q_SLOTS:
 protected:
   /// This function reloads the robot model and reinitializes the PlanningSceneMonitor
   /// It can be called either from the Main Loop or from a Background thread
-  void loadRobotModel(const rclcpp::Node::SharedPtr& node);
+  void loadRobotModel();
 
   /// This function is used by loadRobotModel() and should only be called in the MainLoop
   /// You probably should not call this function directly
@@ -130,8 +130,7 @@ protected:
 
   /// This function constructs a new planning scene. Probably this should be called in a background thread
   /// as it may take some time to complete its execution
-  virtual planning_scene_monitor::PlanningSceneMonitorPtr
-  createPlanningSceneMonitor(const rclcpp::Node::SharedPtr& node);
+  virtual planning_scene_monitor::PlanningSceneMonitorPtr createPlanningSceneMonitor();
 
   /// This is an event called by loadRobotModel() in the MainLoop; do not call directly
   virtual void onRobotModelLoaded();
@@ -195,6 +194,9 @@ protected:
   rviz_common::properties::FloatProperty* scene_display_time_property_;
   rviz_common::properties::EnumProperty* octree_render_property_;
   rviz_common::properties::EnumProperty* octree_coloring_property_;
+
+  // rclcpp node
+  rclcpp::Node::SharedPtr node_;
 };
 
 }  // namespace moveit_rviz_plugin
--- a/moveit_ros/visualization/planning_scene_rviz_plugin/src/planning_scene_display.cpp
+++ b/moveit_ros/visualization/planning_scene_rviz_plugin/src/planning_scene_display.cpp
@@ -185,6 +185,14 @@ void PlanningSceneDisplay::clearJobs()
 void PlanningSceneDisplay::onInitialize()
 {
   Display::onInitialize();
+  auto ros_node_abstraction = context_->getRosNodeAbstraction().lock();
+  if (!ros_node_abstraction)
+  {
+    RVIZ_COMMON_LOG_WARNING("Unable to lock weak_ptr from DisplayContext in PlanningSceneDisplay constructor");
+    return;
+  }
+  node_ = ros_node_abstraction->get_raw_node();
+  planning_scene_topic_property_->initialize(ros_node_abstraction);
 
   // the scene node that contains everything
   planning_scene_node_ = scene_node_->createChildSceneNode();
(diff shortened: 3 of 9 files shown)


JafarAbdi:
I can verify the seg-fault is related to rdf_loader

moveit_ros/visualization/planning_scene_rviz_plugin/include/moveit/planning_scene_rviz_plugin/planning_scene_display.h
```diff
@@ -195,6 +194,9 @@ protected Q_SLOTS:
   rviz_common::properties::FloatProperty* scene_display_time_property_;
   rviz_common::properties::EnumProperty* octree_render_property_;
   rviz_common::properties::EnumProperty* octree_coloring_property_;
+
+  // rclcpp node
+  const rclcpp::Node::SharedPtr node_ = rclcpp::Node::make_shared("planning_scene_display");
```
henningkayser: Doesn't this imply that rclcpp::init() has been called before, and shouldn't this be rather initialized in the constructor?

moveit_ros/visualization/planning_scene_rviz_plugin/src/planning_scene_display.cpp
```diff
@@ -491,12 +489,11 @@ void PlanningSceneDisplay::unsetLinkColor(rviz_default_plugins::robot::Robot* ro
...
 // Load
 // ******************************************************************************************
-planning_scene_monitor::PlanningSceneMonitorPtr
-PlanningSceneDisplay::createPlanningSceneMonitor(const rclcpp::Node::SharedPtr& node)
+planning_scene_monitor::PlanningSceneMonitorPtr PlanningSceneDisplay::createPlanningSceneMonitor()
 {
   std::shared_ptr<tf2_ros::Buffer> tf_buffer = moveit::planning_interface::getSharedTF();
   return std::make_shared<planning_scene_monitor::PlanningSceneMonitor>(
-      node, robot_description_property_->getStdString(), tf_buffer, getNameStd() + "_planning_scene_monitor");
+      node_, robot_description_property_->getStdString(), tf_buffer, getNameStd() + "_planning_scene_monitor");
```
henningkayser: If this is the only reason for the node, maybe we should provide a PlanningSceneMonitor constructor that provides its own node

moveit_ros/visualization/rviz_plugin_render_tools/CMakeLists.txt
```diff
@@ -22,17 +22,19 @@ add_library(${MOVEIT_LIB_NAME} SHARED
...
   src/trajectory_panel.cpp
   src/mesh_shape.cpp
+  ${HEADERS}
 )
 set_target_properties(${MOVEIT_LIB_NAME} PROPERTIES VERSION "${${PROJECT_NAME}_VERSION}")
 
 target_link_libraries(${MOVEIT_LIB_NAME}
     rviz_ogre_vendor::OgreMain
     Qt5::Widgets
+    rviz_common::rviz_common
```
henningkayser: Why not use `ament_target_dependencies()` for these dependencies?

moveit_ros/visualization/robot_state_rviz_plugin/include/moveit/robot_state_rviz_plugin/robot_state_display.h
```diff
@@ -112,7 +112,7 @@ private Q_SLOTS:
   void fixedFrameChanged() override;
 
   // render the robot
-  rclcpp::Node::SharedPtr node_;
+  const rclcpp::Node::SharedPtr node_ = rclcpp::Node::make_shared("robot_state_display");
```
henningkayser: This should also be defined in the constructor IMO

moveit_ros/visualization/planning_scene_rviz_plugin/include/moveit/planning_scene_rviz_plugin/planning_scene_display.h
```diff
@@ -122,16 +122,15 @@ protected Q_SLOTS:
 protected:
   /// This function reloads the robot model and reinitializes the PlanningSceneMonitor
   /// It can be called either from the Main Loop or from a Background thread
-  void loadRobotModel(const rclcpp::Node::SharedPtr& node);
```
mlautman: You should already have access to a node through the Display parent class's `context_` member. 

Taken from `InteractiveMarkerDisplay.cpp`:

```
  auto ros_node_abstraction = context_->getRosNodeAbstraction().lock();
  if (!ros_node_abstraction) {
    return;
  }

  interactive_marker_namespace_property_->initialize(ros_node_abstraction);

  auto frame_transformer = context_->getFrameManager()->getTransformer();
  rclcpp::Node::SharedPtr node = ros_node_abstraction->get_raw_node();
```

moveit_ros/visualization/planning_scene_rviz_plugin/include/moveit/planning_scene_rviz_plugin/planning_scene_display.h
mlautman: see previous comment about the `context_` member variable

moveit_ros/visualization/planning_scene_rviz_plugin/src/planning_scene_display.cpp
mlautman: The PSM is useless without a node which means that it should probably have one as a member variable. Considering that it's use case is very similar to that of the TF Listener where it should asynchronously subscribe and publish to topics, I think that we should model it after the TF Listener which has a constructor that optionally takes in a node. If you pass in a node, it subscribes and publishes with the node passed in but if you don't provide one, it creates it's own node that spins it with it's own separate executor. See TF Listener + TF Buffer for the full implementation.

moveit_ros/visualization/robot_state_rviz_plugin/include/moveit/robot_state_rviz_plugin/robot_state_display.h
mlautman: We don't need to define our own node because one is available in the `context_` member. see interactive_marker_display.cpp

moveit_ros/visualization/rviz_plugin_render_tools/CMakeLists.txt
mlautman: The rviz_default_plugins package does the linking very explicitly as well. Do we know why? It might just be because rviz2 was ported pre-crystal.

moveit_ros/visualization/planning_scene_rviz_plugin/include/moveit/planning_scene_rviz_plugin/planning_scene_display.h
JafarAbdi: Thanks, Mike, this's exactly what I did, sorry for not pushing my changes -- I'll fix the planning scene plugin and push them all together

JafarAbdi:
Currently, it gives the following error 

`[ERROR] [moveit_ros_visualization.planning_scene_display]: Exception caught executing main loop job: Node has already been added to an executor.` which I think is related to planning scene monitor

moveit_ros/planning/rdf_loader/src/rdf_loader.cpp
```diff
@@ -62,68 +62,51 @@ RDFLoader::RDFLoader(const std::shared_ptr<rclcpp::Node>& node, const std::strin
...
   {
     RCLCPP_INFO_ONCE(LOGGER, "Robot model parameter not found! Did you remap '%s'?\n", robot_description.c_str());
     return;
   }
 
   urdf::Model* umodel = new urdf::Model();
-  if (!umodel->initString(content))
+  if (!umodel->initString(robot_description_content))
   {
-    RCLCPP_INFO(LOGGER, "Unable to parse URDF from parameter: '%s'", robot_description_.c_str());
```
JafarAbdi: I don't see the `robot_description_` parameter is used anymore, should I remove it .?

moveit_ros/visualization/planning_scene_rviz_plugin/src/planning_scene_display.cpp
```diff
@@ -662,7 +665,8 @@ void PlanningSceneDisplay::calculateOffsetPosition()
   Ogre::Vector3 position;
   Ogre::Quaternion orientation;
 
-  context_->getFrameManager()->getTransform(getRobotModel()->getModelFrame(), rclcpp::Time(0), position, orientation);
+  context_->getFrameManager()->getTransform(getRobotModel()->getModelFrame(), rclcpp::Time(0, 0, RCL_ROS_TIME),
```
JafarAbdi: This will fix the exception of two different time resources comparison

moveit_ros/visualization/rviz_plugin_render_tools/CMakeLists.txt
JafarAbdi: I don't know why they did it this way, but using `rviz_default_plugins` gives linking error, I looked at how it's defined in the generated `.cmake` files and the target name is `rviz_default_plugins::rviz_default_plugins`

moveit_ros/planning/rdf_loader/src/rdf_loader.cpp
henningkayser: Yes, makes sense

moveit_ros/visualization/planning_scene_rviz_plugin/src/planning_scene_display.cpp
henningkayser: Ok

moveit_ros/planning/rdf_loader/src/rdf_loader.cpp
```diff
@@ -62,68 +62,51 @@ RDFLoader::RDFLoader(const std::shared_ptr<rclcpp::Node>& node, const std::strin
...
-    }
-    catch (const rclcpp::ParameterTypeException& e)
-    {
-      RCLCPP_WARN(LOGGER, "When getting robot_description_semantic parameter: %s", e.what());
-    }
-  }
-  if (scontent.empty())
+  const std::string srdf_description = robot_description + "_semantic";
+  // Check if the robot_description_semantic parameter is declared, declare it if it's not declared yet
+  if (!node->has_parameter(srdf_description))
```
henningkayser: In general, I don't think we should check `has_parameter()`, except we actually have shared parameters with other classes or instances. It would be much more verbose if we would call `declare_parameter()` with a good description while implicitly disallowing other classes to declare the same parameters. This would follow a single-responsibility principle. Also, `declare_parameter()` already returns the value we need and allows us to set the default, making `get_parameter_or()` redundant. We can keep it in this PR for now, but this is something I would like to refactor for the whole code base in order to really clean up and structure parameter use.

JafarAbdi:
This's how I run the RobotState visualization demo

I used the following packages https://github.com/JafarAbdi/franka_description and https://github.com/JafarAbdi/moveit_resources/tree/pr-panda_xacros

[Launch file and executable](https://gist.github.com/JafarAbdi/be153c4fea6235c11e7a1b275b4da601)

`ros2 launch ./robot_state_demo.launch.py`
Run `dummy_joint_states`

From Rviz2 add RobotState plugin and the robot should be visible

moveit_ros/planning/rdf_loader/src/rdf_loader.cpp
JafarAbdi: The reason I add the checking for a previous declaration is that in Rviz When I add multiple RobotState plugins it was throwing an exception

henningkayser:
> This's how I run the RobotState visualization demo
> 
> I used the following packages https://github.com/JafarAbdi/franka_description and https://github.com/JafarAbdi/moveit_resources/tree/pr-panda_xacros

What did you change in your forks?

> 
> [Launch file and executable](https://gist.github.com/JafarAbdi/be153c4fea6235c11e7a1b275b4da601)
> 
> `ros2 launch ./robot_state_demo.launch.py`
> Run `dummy_joint_states`
> 

For `dummy_joint_states` you used this https://github.com/ros2/demos/blob/master/dummy_robot/dummy_sensors/src/dummy_joint_states.cpp?

JafarAbdi:
> > This's how I run the RobotState visualization demo
> > I used the following packages https://github.com/JafarAbdi/franka_description and https://github.com/JafarAbdi/moveit_resources/tree/pr-panda_xacros
> 
> What did you change in your forks?

I isolated franka_description from the other packages and just ported CMakeLists and package.xml files

The moveit_resources package I added the xacro files from [panda_moveit_config](https://github.com/ros-planning/panda_moveit_config) and changed the xacro includes pointing to `moveit_resources/panda_moveit_config`

> > [Launch file and executable](https://gist.github.com/JafarAbdi/be153c4fea6235c11e7a1b275b4da601)
> > `ros2 launch ./robot_state_demo.launch.py`
> > Run `dummy_joint_states`
> 
> For `dummy_joint_states` you used this https://github.com/ros2/demos/blob/master/dummy_robot/dummy_sensors/src/dummy_joint_states.cpp?

Just changed the joints names and the values, as shown [here](https://gist.github.com/JafarAbdi/be153c4fea6235c11e7a1b275b4da601#file-dummy_joint_states-cpp)

moveit_ros/visualization/rviz_plugin_render_tools/CMakeLists.txt
mlautman: rviz_default_plugins contains a number Displays. I wasn't saying we should link against that library, just that we could use it at a template for creating custom Displays.

moveit_ros/visualization/planning_scene_rviz_plugin/include/moveit/planning_scene_rviz_plugin/planning_scene_display.h
mlautman: @JafarAbdi I think that making `node_` a class variable might end up being a mistake as it defeats the entire reason to use a WeakPtr in the first place. (Ref. https://www.learncpp.com/cpp-tutorial/15-7-circular-dependency-issues-with-stdshared_ptr-and-stdweak_ptr/) As best I can tell, the node_ is only used to initialize the PSM which I think could have it's own node and executor just like the tf_transformer. I can work on a PR to fix this

Currently, loading the PlanningSceneDisplay and then removing it causes rviz2 to crash with:
```
what():  RuntimeAssertionException: name must not be empty in destroySceneNode at /tmp/binarydeb/ros-eloquent-rviz-ogre-vendor-7.0.4/obj-x86_64-linux-gnu/ogre-v1.12.1-prefix/src/ogre-v1.12.1/OgreMain/src/OgreSceneManager.cpp (line 760)
```

JafarAbdi: Thanks @mlautman , Does this [PR](https://github.com/ros-planning/moveit2/pull/168) fixes it .?

PR #168, opened 2020-02-05
MoveIt visualization bug fixes

JafarAbdi:
### Description

This fixes two bugs

1- Using the name of the SceneNode was throwing an exception in the destructor [see Mike's comment](https://github.com/ros-planning/moveit2/pull/167/files#r375475440)
2- Having an empty robot description will pass an empty string to rdf_loader which will cause it to throw an `empty parameter name` exception

### Checklist
- [ ] Document API changes relevant to the user in the MIGRATION.md notes
- [ ] Create tests, which fail without this PR [reference](https://ros-planning.github.io/moveit_tutorials/)


mlautman:
I still think that the robot_description should be able to come from either a parameter, topic, or file like the RobotModelDisplay. Just using parameters doesn't make that much sense as it requires Rviz to be launched with special arguments

PR #204, opened 2020-06-02
[Ready] simple_controller_manager fixes from PR #187, remove spin_until_future_complete

ddengster:
Contains changes made to simple_controller_manager, pulled from PR #187 . ~~Still a work in progress.~~ Expect more sub-PRs from PR#187.

~~Edit: Seems that there was a rebase of master, how should i fix this so the history/changes gets cleaned up?~~

Done! Contains mostly replacements to spin_until_future_complete so that PR#187 can work

### Checklist

### DIFF
--- a/moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
+++ b/moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
@@ -80,7 +80,13 @@ public:
                               const std::string& logger_name)
     : ActionBasedControllerHandleBase(name, logger_name), node_(node), done_(true), namespace_(ns)
   {
-    controller_action_client_ = rclcpp_action::create_client<T>(node_, getActionName());
+    controller_action_client_ = rclcpp_action::create_client<T>(
+      node_->get_node_base_interface(),
+      node_->get_node_graph_interface(),
+      node_->get_node_logging_interface(),
+      node_->get_node_waitables_interface(),
+      getActionName());
+
     unsigned int attempts = 0;
     double timeout;
     node_->get_parameter_or("trajectory_execution.controller_connection_timeout", timeout, 15.0);
@@ -123,6 +129,7 @@ public:
     {
       RCLCPP_INFO_STREAM(LOGGER, "Cancelling execution for " << name_);
       auto cancel_result_future = controller_action_client_->async_cancel_goal(current_goal_);
+      //@todo: replace rclcpp::spin_until_future_complete
       if (rclcpp::spin_until_future_complete(node_, cancel_result_future) !=
           rclcpp::executor::FutureReturnCode::SUCCESS)
       {
@@ -137,9 +144,12 @@ public:
   bool waitForExecution(const rclcpp::Duration& timeout = rclcpp::Duration(0)) override
   {
     auto result_future = controller_action_client_->async_get_result(current_goal_);
-    if (controller_action_client_ && !done_)
-      return (rclcpp::spin_until_future_complete(node_, result_future, timeout.to_chrono<std::chrono::seconds>()) ==
-              rclcpp::executor::FutureReturnCode::SUCCESS);
+
+    rclcpp::Time start_time = node_->now();
+    auto end_time = start_time + timeout;
+    while (!done_ && node_->now() < end_time) {
+      std::this_thread::sleep_for(std::chrono::milliseconds(10));
+    }
     return true;
   }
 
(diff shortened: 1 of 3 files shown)


moveit_plugins/moveit_simple_controller_manager/src/follow_joint_trajectory_controller_handle.cpp
```diff
@@ -54,34 +54,42 @@ bool FollowJointTrajectoryControllerHandle::sendTrajectory(const moveit_msgs::ms
...
+    else
+      RCLCPP_INFO(LOGGER, "Goal request accepted!");
   };
   // Result callback
   send_goal_options.result_callback =
       std::bind(&FollowJointTrajectoryControllerHandle::controllerDoneCallback, this, _1);
   // Send goal
   auto current_goal_future = controller_action_client_->async_send_goal(goal, send_goal_options);
-  if (rclcpp::spin_until_future_complete(node_, current_goal_future) != rclcpp::executor::FutureReturnCode::SUCCESS)
+  std::future_status status = current_goal_future.wait_for(std::chrono::seconds(60));
```
henningkayser, line 79: Why wait for 60 seconds?  the goal doesn't really have a timeout, right?

moveit_plugins/moveit_simple_controller_manager/src/follow_joint_trajectory_controller_handle.cpp
```diff
@@ -54,34 +54,42 @@ bool FollowJointTrajectoryControllerHandle::sendTrajectory(const moveit_msgs::ms
...
+  if (status != std::future_status::ready)
```
henningkayser, line 80: Couldn't the status be deferred as well?

ddengster: changed to `(status == std::future_status::timeout)` 2b1679b

moveit_plugins/moveit_simple_controller_manager/src/follow_joint_trajectory_controller_handle.cpp
```diff
@@ -54,34 +54,42 @@ bool FollowJointTrajectoryControllerHandle::sendTrajectory(const moveit_msgs::ms
...
 
   rclcpp_action::Client<control_msgs::action::FollowJointTrajectory>::SendGoalOptions send_goal_options;
   // Active callback
-  send_goal_options.goal_response_callback = [this](const auto& /* unused-arg */) {
-    RCLCPP_DEBUG_STREAM(LOGGER, name_ << " started execution");
+  send_goal_options.goal_response_callback = [this](const auto& future) {
+    RCLCPP_INFO_STREAM(LOGGER, name_ << " started execution");
+    auto goal_handle = future.get();
+    if (!goal_handle)
+      RCLCPP_INFO(LOGGER, "Goal request rejected");
```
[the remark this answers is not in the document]
ddengster: switched to warn 2b1679b

moveit_plugins/moveit_simple_controller_manager/src/follow_joint_trajectory_controller_handle.cpp
ddengster: i was thinking it would be better to have a timeout instead of spinning forever.  Besides https://github.com/ros-planning/moveit/blob/master/moveit_plugins/moveit_simple_controller_manager/src/follow_joint_trajectory_controller_handle.cpp#L65 seems to be completely async with no checking. Should we follow it as well?

moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
```diff
@@ -137,9 +140,18 @@ class ActionBasedControllerHandle : public ActionBasedControllerHandleBase
...
     auto result_future = controller_action_client_->async_get_result(current_goal_);
-    if (controller_action_client_ && !done_)
-      return (rclcpp::spin_until_future_complete(node_, result_future, timeout.to_chrono<std::chrono::seconds>()) ==
-              rclcpp::executor::FutureReturnCode::SUCCESS);
+
+    if (timeout.seconds() == 0.0)
+      result_future.get();
+    else
+    {
+      rclcpp::Time start_time = node_->now();
```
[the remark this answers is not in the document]
ddengster: switched to wait_for, warn if timeout eb4be53

moveit_plugins/moveit_simple_controller_manager/src/follow_joint_trajectory_controller_handle.cpp
henningkayser: I think we can keep this spinning since it's the job of the TrajectoryExecutionMonitor to validate execution duration (and cancel the goal if execution times out). I wouldn't set the goal to a hard timeout at least.

ddengster: removed timeout 465027a

PR #201, opened 2020-05-26
Port trajectory_rviz_plugin

RoboticsYY:
### Description

This PR intends to port the `moveit_ros_visualization/trajectory_rviz_plugin` to ROS2. The dependencies were ported previously in PR #160.

### Checklist
- [x] **Required by CI**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)

PR #213, opened 2020-06-12
[Don't Merge] CI test rebase of #201

henningkayser:
If CI succeeds and comments are addressed, I will force-push this branch to #201 and merge to master.


henningkayser:
I added nolint tags to the subscription callback as suggested in #214

PR #220, opened 2020-06-19
Port eigen_kdl to ROS2 and add it to moveit_core/utils

JafarAbdi:


JafarAbdi:
> If a simple
> 
> ```
> tf2::fromMsg(ik_pose_chain, tf2::toMsg(ik_eigen_pose));
> ```
> 
> has worse performance (I wouldn't be surprised if it didn't with optimizations on)

I just did benchmark both approaches

```cpp
#include <benchmark/benchmark.h>
#include <eigen_conversions/eigen_kdl.h>
#include <tf2_kdl/tf2_kdl.h>

static void BM_EigenKdl(benchmark::State &state)
{
    KDL::Frame kdl_frame;
    Eigen::Isometry3d eigen_frame;
    for (auto _ : state) {
        tf::transformEigenToKDL(eigen_frame, kdl_frame);
    }
}
BENCHMARK(BM_EigenKdl);

static void BM_TF2(benchmark::State &state)
{
    KDL::Frame kdl_frame;
    Eigen::Isometry3d eigen_frame;
    for (auto _ : state) {
        tf2::fromMsg(tf2::toMsg(eigen_frame), kdl_frame);
    }
}
BENCHMARK(BM_TF2);

BENCHMARK_MAIN();
```

The result

```
Run on (12 X 4100 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x6)
  L1 Instruction 32 KiB (x6)
  L2 Unified 256 KiB (x6)
  L3 Unified 9216 KiB (x1)
Load Average: 0.76, 0.54, 0.42
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
------------------------------------------------------
Benchmark            Time             CPU   Iterations
------------------------------------------------------
BM_EigenKdl       10.8 ns         10.8 ns     61803435
BM_TF2             152 ns          152 ns      4633335
```

>This adds a lot of link dependencies to an otherwise trivial library.
I believe the old geometry stack was meant to drop out - not to be moved in with downstream projects.
Especially not because of a single call to a single function from the header.

> still go for it but add a TODO together with a link to a new pull-request in tf2_kdl adding the missing header.

I did open a PR to port it but it was closed https://github.com/ros2/geometry2/pull/273, my initial proposal was to create a new repo in `ros-planning` and port it there rather than having it in moveit_core/utils but in the MoveIt 2 weekly sync we agreed on the current approach

henningkayser:
@JafarAbdi thanks for testing this

> This adds a lot of link dependencies to an otherwise trivial library.
> I believe the old `geometry` stack was meant to drop out - not to be moved in with downstream projects.
> Especially not because of a single call to a single function from the header.
> 
> If a simple
> 
> ```
> tf2::fromMsg(ik_pose_chain, tf2::toMsg(ik_eigen_pose));
> ```
> 
> has worse performance (I wouldn't be surprised if it didn't with optimizations on), still go for it but add a TODO together with a link to a new pull-request in [tf2_kdl](https://github.com/ros/geometry2/tree/melodic-devel/tf2_kdl/include/tf2_kdl) adding the missing header.

@v4hn I totally agree that your approach seems much nicer, but looking at the performance I would still vote for getting this added here. It's also much more reasonable than adding and maintaining a separate repo for this, IMO. We should add a TODO and probably also an issue referencing this, though. @JafarAbdi could you do this?

v4hn:
> I just did benchmark both approaches

Awesome thank you for taking the time! Did you compile with `-O2` or `-O3`? I would be interested in -O3 results, though I'm not sure our release builds are with `-O3`.

> I did open a PR to port it but it was closed ros2/geometry2#273

The reasoning there does not seem reasonable to me. I [commented there](https://github.com/ros2/geometry2/pull/273#issuecomment-648957071).

Craft like that tends to stay around forever, because, why would you ever look into this code again?
The MoveIt (1) codebase currently has 109 TODOs inline.
As only one of these functions is used in only one location in the whole project I believe the previous solution to add a local function and and issue to release `eigen_kdl`, which you propose to address here, is actually better than moving in the whole header with MoveIt.
After all, why would we provide all these functions in MoveIt's public API?

JafarAbdi:
> Did you compile with `-O2` or `-O3`? I would be interested in -O3 results, though I'm not sure our release builds are with `-O3`.

Yes, I did compile it with `-DCMAKE_BUILD_TYPE=Release`

> After all, why would we provide all these functions in MoveIt's public API?

I would prefer having it in a separate package than having it inside MoveIt codebase since it's a small package and it requires almost zero maintenance

v4hn:
> > Did you compile with `-O2` or `-O3`? I would be interested in -O3 results, though I'm not sure our release builds are with `-O3`.
> 
> Yes, I did compile it with `-DCMAKE_BUILD_TYPE=Release`

Wow, I guess I overexpected the compiler in this case. I don't see why this could not be optimized to the same level.

> I would prefer having it in a separate package than having it inside MoveIt codebase since it's a small package and it requires almost zero maintenance

Yes. I would wait for the geometry2 maintainers to reconsider before having yet another one-source-file repository around.

Either way, this patch does not seem to improve anything w.r.t. the issue of making the standalone package available in ros2.

rhaschke:
I agree with Michael, that this addition should be merged into geometry2.

v4hn:
@JafarAbdi What's the state of the geometry2 pull-request? I believe Tully clearly stated what needs to be done to get this code accepted there. As transition might take time it's a good idea to finish this soon...

JafarAbdi:
Closing since https://github.com/ros2/geometry2/pull/311 is merged

PR #60, opened 2019-03-30
Constraint samplers

vmayoral:
- [ ] Unit tests (pending on addressing https://github.com/AcutronicRobotics/moveit2/issues/13)

PR #226, opened 2020-06-24
Foxy prep fixes

henningkayser:
General fixes that are required for Foxy to build. Since these are very generic, I think they should be merged into master.

### DIFF
--- a/moveit_core/CMakeLists.txt
+++ b/moveit_core/CMakeLists.txt
@@ -29,7 +29,7 @@ find_package(rclcpp REQUIRED)
 if(WIN32)
   set(EXTRA_BOOST_COMPONENTS zlib)
 endif()
-find_package(Boost REQUIRED system filesystem date_time thread iostreams ${EXTRA_BOOST_COMPONENTS})
+find_package(Boost REQUIRED system filesystem date_time thread iostreams regex ${EXTRA_BOOST_COMPONENTS})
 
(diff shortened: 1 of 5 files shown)


moveit_core/constraint_samplers/test/pr2_arm_kinematics_plugin.h
```diff
@@ -84,6 +84,8 @@ class PR2ArmIKSolver : public KDL::ChainIkSolverPos
 
   ~PR2ArmIKSolver() override{};
 
+  void updateInternalDataStructures() override;
```
henningkayser, line 87: For some reason this function was deleted in (https://github.com/ros-planning/moveit2/pull/60/files#diff-8404d089489122c76cd63a908826c88eL90) even though it overrides a virtual function. Not sure why this didn't complain with eloquent.

v4hn:
Could you please keep an eye on and add relevant changes to the moveit repo as well?
I guess at least the `regex` component is missing here too.

moveit_core/constraint_samplers/test/pr2_arm_kinematics_plugin.h
henningkayser: Removed this since `ros-eloquent-orocos-kdl` is at 3.2.1 which doesn't include this. @clalancette are there plans for making another eloquent release?

PR #228, opened 2020-06-25
Cleanup CMakeLists target dependencies

henningkayser:
Started looking into dependency issues with the Foxy branch and ended up cleaning up most of the dependencies.

### DIFF
--- a/moveit_core/CMakeLists.txt
+++ b/moveit_core/CMakeLists.txt
@@ -58,7 +58,7 @@ find_library(LIBFCL_LIBRARIES_FULL ${LIBFCL_LIBRARIES} ${LIBFCL_LIBRARY_DIRS})
 set(LIBFCL_LIBRARIES "${LIBFCL_LIBRARIES_FULL}")
 
 
-find_package(octomap REQUIRED)
+find_package(OCTOMAP REQUIRED)
 find_package(urdfdom REQUIRED)
 find_package(urdf REQUIRED)
 find_package(urdfdom_headers REQUIRED)
--- a/moveit_core/CMakeLists.txt
+++ b/moveit_core/CMakeLists.txt
@@ -169,19 +169,7 @@ include_directories(SYSTEM ${EIGEN3_INCLUDE_DIRS}
   # ${BULLET_INCLUDE_DIRS}
 )
 
-include_directories(${THIS_PACKAGE_INCLUDE_DIRS}
-  ${VERSION_FILE_PATH}
-  ${rclcpp_INCLUDE_DIRS}
-  ${rmw_implementation_INCLUDE_DIRS}
-  ${Boost_INCLUDE_DIRS}
-  ${urdfdom_INCLUDE_DIRS}
-  ${urdfdom_headers_INCLUDE_DIRS}
-  ${OCTOMAP_INCLUDE_DIRS}
-  ${moveit_msgs_INCLUDE_DIRS}
-  ${random_numbers_INCLUDE_DIRS}
-  ${srdfdom_INCLUDE_DIRS}
-  ${geometric_shapes_INCLUDE_DIRS}
-)
+include_directories(${THIS_PACKAGE_INCLUDE_DIRS} ${VERSION_FILE_PATH})
 
 # Generate and install version.h
 string(REGEX REPLACE "^([0-9]+)\\..*" "\\1" MOVEIT_VERSION_MAJOR "${${PROJECT_NAME}_VERSION}")
--- a/moveit_core/constraint_samplers/CMakeLists.txt
+++ b/moveit_core/constraint_samplers/CMakeLists.txt
@@ -23,9 +23,9 @@ target_link_libraries(${MOVEIT_LIB_NAME}
 )
 
 install(TARGETS ${MOVEIT_LIB_NAME}
-        LIBRARY DESTINATION lib
-        ARCHIVE DESTINATION lib
-        RUNTIME DESTINATION bin
+  LIBRARY DESTINATION lib
+  ARCHIVE DESTINATION lib
+  RUNTIME DESTINATION bin
 )
 
 install(DIRECTORY include/ DESTINATION include)
@@ -49,30 +49,13 @@ if(BUILD_TESTING)
     test/pr2_arm_kinematics_plugin.cpp
     test/pr2_arm_ik.cpp
   )
-
   target_include_directories(test_constraint_samplers PUBLIC
     ${geometry_msgs_INCLUDE_DIRS}
   )
-
+  ament_target_dependencies(test_constraint_samplers kdl_parser)
   target_link_libraries(test_constraint_samplers
-    moveit_constraint_samplers
-    moveit_kinematic_constraints
-    moveit_planning_scene
     moveit_test_utils
     moveit_utils
-    moveit_robot_state
     ${MOVEIT_LIB_NAME}
-    ${angles_LIBRARIES}
-    ${urdf_LIBRARIES}
-    ${tf2_LIBRARIES}
-    ${tf2_kdl_LIBRARIES}
-    ${tf2_ros_LIBRARIES}
-    ${kdl_parser_LIBRARIES}
-    ${OCTOMAP_LIBRARIES}
-    ${orcos_kdl_LIBRARIES}
-    ${urdfdom_LIBRARIES}
-    ${urdfdom_headers_LIBRARIES}
-    ${rclcpp_headers_LIBRARIES}
-    ${geometric_shapes_LIBRARIES}
   )
 endif()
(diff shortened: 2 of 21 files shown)


moveit_core/planning_scene/CMakeLists.txt
```diff
@@ -47,21 +42,14 @@ if(BUILD_TESTING)
...
-    ${MOVEIT_LIB_NAME}
-    ${srdfdom_LIBRARIES}
-    ${urdf_LIBRARIES}
-    ${geometric_shapes_LIBRARIES}
-    ${OCTOMAP_LIBRARIES}
-    ${ament_index_cpp_LIBRARIES}
+  ament_target_dependencies(test_planning_scene
+    ament_index_cpp
+    geometric_shapes
+    srdfdom
```
v4hn, line 48: `urdf_parser` is included, but not listed here?

moveit_core/planning_scene/CMakeLists.txt
```diff
@@ -47,21 +42,14 @@ if(BUILD_TESTING)
...
 
   ament_add_gtest(test_multi_threaded test/test_multi_threaded.cpp
     APPEND_LIBRARY_DIRS "${append_library_dirs}")
-  target_link_libraries(test_multi_threaded
-    moveit_planning_scene
-    moveit_test_utils
-    ${MOVEIT_LIB_NAME}
-    ${urdfdom_LIBRARIES}
-    ${urdfdom_headers_LIBRARIES})
+  target_link_libraries(test_multi_threaded moveit_test_utils ${MOVEIT_LIB_NAME})
```
v4hn, line 54: `robot_state`, `robot_model` and `collision_detection` are included, but not listed here?

moveit_core/trajectory_processing/CMakeLists.txt
```diff
@@ -44,19 +44,8 @@ if(BUILD_TESTING)
...
     moveit_test_utils
-    moveit_robot_trajectory
-    moveit_trajectory_processing
-    ${urdfdom_LIBRARIES}
-    ${srdfdom_LIBRARIES}
-    ${urdfdom_headers_LIBRARIES}
     ${MOVEIT_LIB_NAME}
-    ${geometric_shapes_LIBRARIES}
-    resource_retriever::resource_retriever
   )
```
v4hn, line 50: `robot_state`, `robot_model`, and `robot_trajectory` are included but not listed here.

moveit_core/CMakeLists.txt
```diff
@@ -58,7 +58,7 @@ find_library(LIBFCL_LIBRARIES_FULL ${LIBFCL_LIBRARIES} ${LIBFCL_LIBRARY_DIRS})
 set(LIBFCL_LIBRARIES "${LIBFCL_LIBRARIES_FULL}")
 
 
-find_package(octomap REQUIRED)
+find_package(OCTOMAP REQUIRED)
```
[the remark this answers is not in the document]
henningkayser: In 2017 https://github.com/OctoMap/octomap/commit/14581c443dfb13e86e77c500937855d14bad7dd3#diff-5de12f1706f6b5f33b13746d789378c5. Using lowercase wouldn't work with ament_target_dependencies

moveit_core/constraint_samplers/CMakeLists.txt
```diff
@@ -49,30 +49,13 @@ if(BUILD_TESTING)
...
-    ${tf2_LIBRARIES}
-    ${tf2_kdl_LIBRARIES}
-    ${tf2_ros_LIBRARIES}
-    ${kdl_parser_LIBRARIES}
-    ${OCTOMAP_LIBRARIES}
-    ${orcos_kdl_LIBRARIES}
-    ${urdfdom_LIBRARIES}
-    ${urdfdom_headers_LIBRARIES}
-    ${rclcpp_headers_LIBRARIES}
-    ${geometric_shapes_LIBRARIES}
```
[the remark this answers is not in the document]
henningkayser: `moveit_constraint_samplers` is `${MOVEIT_LIB_NAME}`. All other targets are provided with `$MOVEIT_LIB_NAME` so I didn't specify the same deps again for the test.

moveit_core/planning_scene/CMakeLists.txt
```diff
@@ -3,10 +3,13 @@ set(MOVEIT_LIB_NAME moveit_planning_scene)
...
-  ${octomap_msgs_INCLUDE_DIRS}
-  ${octomap_INCLUDE_DIRS}
+ament_target_dependencies(${MOVEIT_LIB_NAME}
+  Boost
+  rclcpp
+  urdfdom
+  urdfdom_headers
+  octomap_msgs
+  OCTOMAP
 )
```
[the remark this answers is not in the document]
henningkayser: It's not PUBLIC in MoveIt 1, so I removed it. We can specify octomap PUBLIC with ament_target_dependencies though.

moveit_core/planning_scene/CMakeLists.txt
henningkayser: provided by `${MOVEIT_LIB_NAME}`

moveit_core/planning_scene/CMakeLists.txt
henningkayser: same

moveit_core/trajectory_processing/CMakeLists.txt
henningkayser: same

henningkayser:
@v4hn thanks for reviewing this. My goal was to remove all the very generic (and often redundant) `target_link_libraries()` deps and includes and use `ament_target_dependencies()` more consistently. A lot of the removed includes were added in Crystal/Dashing where this wouldn't work like it does now. The "missing" dependencies should only be with test targets that really get all dependencies from a library which is specified in the same CMakeLists file. I found that this is just makes more sense is less redundant than specifying the same targets again and again. Considering the circular dependencies in moveit_core, it wouldn't even be possible to always cover all included targets anyway.

moveit_core/CMakeLists.txt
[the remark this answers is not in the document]
henningkayser: Also https://github.com/OctoMap/octomap/issues/138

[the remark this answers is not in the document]
v4hn: I see. Now that's quite a messy package detection in cmake and I wasn't aware of the partially case-sensitive checks.
Makes sense to adjust. :+1:

v4hn:
> My goal was to remove all the very generic (and often redundant) target_link_libraries() deps

Now this is somewhat a matter of build-system philosophy (if such a term exists :sunglasses:). The consistent approach to header includes and linking definitions is to include and link what you use, whether or not you also depend on it transitively. This is because (1) you can syntactically identify all places depending on a package and (2) if the transitive dependency at some point vanishes, the code might still work. I tend to use this approach in my projects, but I do see the benefits of leaving them out.

I will leave it to someone else to accept this as I would not do it that way myself.

henningkayser:
> > My goal was to remove all the very generic (and often redundant) target_link_libraries() deps
> 
> Now this is somewhat a matter of build-system philosophy (if such a term exists ). The consistent approach to header includes and linking definitions is to include and link what you use, whether or not you also depend on it transitively. This is because (1) you can syntactically identify all places depending on a package and (2) if the transitive dependency at some point vanishes, the code might still work. I tend to use this approach in my projects, but I do see the benefits of leaving them out.
> 
> I will leave it to someone else to accept this as I would not do it that way myself.

I totally see your point. But then again, I only applied this "cleanup" to test executables that don't have any downstream dependencies. All other exported targets also define and export their dependencies explicitly (instead of using global includes and library lists as before). So for me it's just a question of linking/including transitive dependencies to tests which are provided by a target in the same CMakeLists file. It probably is really a matter of style/convention at some point, similar to https://github.com/ros-planning/moveit/issues/2170.

henningkayser:
> Out of curiosity I ran catkin_lint on this PR and these are the only cmake errors:
> 
> ```
> moveit_planners_chomp: CMakeLists.txt(16): error: duplicate find_package(catkin)
> moveit_ros_robot_interaction: CMakeLists.txt(54): error: missing test_depend on 'rosunit'
> moveit_setup_assistant: CMakeLists.txt(166): error: missing test_depend on 'rosunit'
> ```
> 
> I'm not sure if these matter, but they might be worth fixing while you are refactoring some of the cmake files. I looked through these changes and they seem to make sense to me. I'll approve it as I don't have anything against these changes.

@tylerjw Thanks for reviewing and testing catkin_lint on this. The failing packages have not been ported yet and have therefore not been checked by `ament_lint_cmake`. The reported errors should be addressed with migration - fixes are automatically enforced by CI in that case. But maybe it would make sense to enable catkin_lint in MoveIt 1.

PR #230, opened 2020-07-08
Planning Scene Monitor Node Executor

jrgnicho:
### Description
Executes internal timer callback using a Single Threaded Executor, this is an attempt to mimic the ROS1 behavior that used an Asynch Spinner and an internal Callback Queue to manage the timer callback invocation.  Need some guidance on how best to test this.

### Checklist


jrgnicho:
I tested this change by running the **run_moveit_cpp** demo
This is ready for review

jrgnicho:
@henningkayser the CI just passed

henningkayser:
@jrgnicho looks good! But shouldn't we then use the private node for running the other callbacks as well? The original implementation had an AsyncSpinner spinning a callback queue assigned to the node handle (which effectively runs the callbacks in a separate thread). Here we initialize the executor in a separate thread and use this for spinning a private node. So for actually running the callbacks in that thread I think we have to initialize publishers/subscribers/services with the private node instead of `node_`. Related to this, there is currently being worked on supporting callback groups (https://github.com/ros2/rclcpp/issues/519) which might at some point allow a more direct replacement of the old behavior.

jrgnicho:
@henningkayser I've made the requested changes

PR #233, opened 2020-07-13
Update Docker build

ruffsl:
### Description

Update to https://github.com/ros-planning/moveit2/pull/44

PR #227, opened 2020-06-24
Compile on ROS2 Foxy

henningkayser:
Once #226 is merged, I will branch off `foxy` and retarget this PR, generate docker images and trigger CI.

PR #237, opened 2020-07-29
Simplify kdl now that kinetic support is dropped

ruffsl:
Otherwise the build fails for `osrf/ros2:nightly` (Ubuntu 20.04) due to the unimplemented virtual function.

Related: https://github.com/ros-planning/moveit2/pull/233

### DIFF
--- a/moveit_kinematics/kdl_kinematics_plugin/include/moveit/kdl_kinematics_plugin/chainiksolver_vel_mimic_svd.hpp
+++ b/moveit_kinematics/kdl_kinematics_plugin/include/moveit/kdl_kinematics_plugin/chainiksolver_vel_mimic_svd.hpp
@@ -62,14 +62,7 @@ public:
                                     const std::vector<kdl_kinematics_plugin::JointMimic>& mimic_joints,
                                     bool position_ik = false, double threshold = 0.001);
 
-// TODO: simplify after kinetic support is dropped
-#define KDL_VERSION_LESS(a, b, c) ((KDL_VERSION) < ((a << 16) | (b << 8) | c))
-#if KDL_VERSION_LESS(1, 4, 0)
-  void updateInternalDataStructures();
-#else
   void updateInternalDataStructures() override;
-#endif
-#undef KDL_VERSION_LESS
 
   ~ChainIkSolverVelMimicSVD() override;
 


moveit_kinematics/kdl_kinematics_plugin/include/moveit/kdl_kinematics_plugin/chainiksolver_vel_mimic_svd.hpp
```diff
@@ -62,14 +62,7 @@ class ChainIkSolverVelMimicSVD : public ChainIkSolverVel
                                     const std::vector<kdl_kinematics_plugin::JointMimic>& mimic_joints,
                                     bool position_ik = false, double threshold = 0.001);
 
-// TODO: simplify after kinetic support is dropped
-#define KDL_VERSION_LESS(a, b, c) ((KDL_VERSION) < ((a << 16) | (b << 8) | c))
-#if KDL_VERSION_LESS(1, 4, 0)
   void updateInternalDataStructures();
```
henningkayser, line 65: Shouldn't this just have an override (https://github.com/ros-planning/moveit/blob/master/moveit_kinematics/kdl_kinematics_plugin/include/moveit/kdl_kinematics_plugin/chainiksolver_vel_mimic_svd.hpp#L70)? Not sure why this was changed in the first place.

henningkayser: Related to https://github.com/ros-planning/moveit2/pull/236

ruffsl: From the git blame, it looks like it was removed here:
https://github.com/ros-planning/moveit2/commit/68c7efc73ec78588a54916bc2f8df93a6831ce3e#diff-3a912bc5f8c8738aafa920927d7dc938

henningkayser: Yeah, looks like it slipped through. But then I would opt for reverting to the function override and only remove the version check.

JafarAbdi: Before foxy, we didn't have this function at all [see](https://github.com/ros2/orocos_kinematics_dynamics/blob/eloquent/orocos_kdl/src/chainiksolver.hpp), they add it to the foxy branch because it was missing [see](https://github.com/ros2/orocos_kinematics_dynamics/blob/foxy/orocos_kdl/src/chainiksolver.hpp)

ruffsl: Feel free to push commits for what you mean, and I'll run it past CI build.

henningkayser: I think it's already included in #227 so if that one runs passes CI on Foxy we should be good.

henningkayser:
@ruffsl I updated this branch, simply removing the version check in `moveit_kinematics` The remaining functions where re-enabled with the MoveIt 1 sync. Since master is targeting Foxy only (see comment in #250) we should be all clear with this solution.

PR #240, opened 2020-07-31
Fix depend for boost date-time

ruffsl:
Fixes: https://github.com/ros-planning/moveit2/issues/239

### DIFF
--- a/moveit_ros/benchmarks/CMakeLists.txt
+++ b/moveit_ros/benchmarks/CMakeLists.txt
@@ -17,7 +17,7 @@ endif()
 
 find_package(ament_cmake REQUIRED)
 find_package(rclcpp REQUIRED)
-find_package(Boost REQUIRED filesystem)
+find_package(Boost REQUIRED date_time filesystem)
 find_package(tf2_eigen REQUIRED)
 find_package(moveit_ros_planning REQUIRED)
 # TODO(YuYan): uncomment after porting moveit_ros_warehouse
--- a/moveit_ros/benchmarks/package.xml
+++ b/moveit_ros/benchmarks/package.xml
@@ -17,12 +17,19 @@
 
   <buildtool_depend>ament_cmake</buildtool_depend>
 
+  <build_depend>libboost-dev</build_depend>
+  <build_depend>libboost-date-time-dev</build_depend>
+  <build_depend>libboost-filesystem-dev</build_depend>
+
   <depend>moveit_ros_planning</depend>
   <!-- TODO(YuYan): uncomment after porting moveit_ros_warehouse -->
   <!-- <depend>moveit_ros_warehouse</depend> -->
   <depend>rclcpp</depend>
   <depend>tf2_eigen</depend>
   <depend version_gte="1.11.2">pluginlib</depend>
+
+  <exec_depend>libboost-date-time</exec_depend>
+  <exec_depend>libboost-filesystem</exec_depend>
 
   <export>
     <build_type>ament_cmake</build_type>


ruffsl:
Not sure though if `build_export_depend` is needed for any of the boost components.

moveit_ros/benchmarks/package.xml
```diff
@@ -17,13 +17,20 @@
 
   <buildtool_depend>ament_cmake</buildtool_depend>
 
+  <build_depend>libboost-dev</build_depend>
+  <build_depend>libboost-date-time-dev</build_depend>
```
henningkayser, line 21: Instead of referencing these packages explicitly, I think we should just use `<depend>boost</depend>` (for libboost-all-dev) just as with `moveit_core`.

moveit_ros/benchmarks/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ endif()
 
 find_package(ament_cmake REQUIRED)
 find_package(rclcpp REQUIRED)
-find_package(Boost REQUIRED filesystem)
```
henningkayser, line 20: Is this required by a change in the boost lib or is this something that might have to be fixed for MoveIt 1 as well?

henningkayser:
> Not sure though if `build_export_depend` is needed for any of the boost components.

I think exporting boost components doesn't work like that, unfortunately.

moveit_ros/benchmarks/package.xml
gavanderhoorn: The point I believe of these changes is actually do *reduce* the size of dependencies. The `boost` key is almost a wildcard when it comes to Boost dependencies, dragging up to 700MB+ of additional (and often unnecessary) dependencies.

See [Generating ‘dev’ and runtime artefacts from ROS packages](https://discourse.ros.org/t/generating-dev-and-runtime-artefacts-from-ros-packages/12448) for some context.

henningkayser: I see, in general it's definitely much better to include all dependencies explicitly, but then I think we should fix this for all packages (including upstream, i.e. ompl) and for MoveIt 1/2 the same way.

henningkayser: > See [Generating ‘dev’ and runtime artefacts from ROS packages](https://discourse.ros.org/t/generating-dev-and-runtime-artefacts-from-ros-packages/12448) for some context.

Thanks, very interesting findings!

Any reasons to not consolidate build/exec depends in a single `<depend>` tag?

gavanderhoorn: > Any reasons to not consolidate build/exec depends in a single `<depend>` tag?

I don't understand your question.

Aren't the reasons to use fine-grained permissions discussed in the thread I (and you) link to?

henningkayser: Well, the major improvement seems to be coming from not using boost-all, dev/non-dev separation seems to only really come into play when creating dev/non-dev ROS packages (which would be a great thing to do eventually). I'm all in favor of specifying dependencies as accurate (or minimal) as possible, but then I would prefer a more coordinated refactoring for all repos/branches.

gavanderhoorn: Taking small steps is probably more feasible though.

moveit_ros/benchmarks/CMakeLists.txt
ruffsl: I'm not sure, I only encountered https://github.com/ros-planning/moveit2/issues/239 if `date_time` boost component was not explicitly declared. @mikaelarguedas hinted to me that this was the cause before.

moveit_ros/benchmarks/package.xml
ruffsl: I'm all for being explicit with dependencies and smaller install footprints.

PR #247, opened 2020-08-14
Update demo launch files to use the new moveit_resources layout

JafarAbdi:
See https://github.com/ros-planning/moveit_resources/pull/41, we should also sync this PR https://github.com/ros-planning/moveit/pull/2199

### DIFF
--- a/moveit_demo_nodes/run_moveit_cpp/src/run_moveit_cpp.cpp
+++ b/moveit_demo_nodes/run_moveit_cpp/src/run_moveit_cpp.cpp
@@ -94,7 +94,7 @@ public:
 
     // Set joint state goal
     RCLCPP_INFO(LOGGER, "Set goal");
-    arm.setGoal("home");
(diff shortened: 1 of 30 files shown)


moveit_demo_nodes/run_moveit_cpp/src/run_moveit_cpp.cpp
```diff
@@ -94,7 +94,9 @@ class MoveItCppDemo
 
     // Set joint state goal
     RCLCPP_INFO(LOGGER, "Set goal");
-    arm.setGoal("home");
+    auto robot_state = *moveit_cpp_->getCurrentState();
```
JafarAbdi, line 97: Since we removed the `home` pose, I used a robot state to set the goal

henningkayser: I switched it to the extended pose

PR #249, opened 2020-08-17
Fix performance-unnecessary-copy-initialization

henningkayser:


henningkayser:
This should be squash-merged

PR #254, opened 2020-08-24
[maint] branch of moveit2 on repos file for eloquent branch

tylerjw:
### Description

fixes the branch for moveit2 on eloquent branch to point to eloquent branch


henningkayser (approved):
Of course this should be fixed. Also, eloquent doesn't support the new `moveit_resources` structure yet. So we should either backport #247 or fix the commit hash.

henningkayser:
Already merged with https://github.com/ros-planning/moveit2/pull/291

PR #256, opened 2020-08-25
Fix moveit_servo test executable

henningkayser:
This test keeps failing for some reason and to me it looks like this should be an executable for a launch test and not a unit test.
https://travis-ci.com/github/ros-planning/moveit2/jobs/377469134
@AdamPettinger, @tylerjw what do you think? I'm wondering why it should have passed in the first place, though.


henningkayser:
Well, doesn't make sense what I did there... closing

PR #257, opened 2020-08-25
[fix] threading issues in PSM and servo

tylerjw:
### Description

Here are a couple of issues I found by building with thread sanitizer and running the servo tests.

### DIFF
--- a/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
+++ b/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
@@ -488,7 +488,7 @@ protected:
   std::unique_ptr<boost::thread> publish_planning_scene_;
   double publish_planning_scene_frequency_;
   SceneUpdateType publish_update_types_;
-  SceneUpdateType new_scene_update_;
+  std::atomic<SceneUpdateType> new_scene_update_;
   boost::condition_variable_any new_scene_update_condition_;
 
   // subscribe to various sources of data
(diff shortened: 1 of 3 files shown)


moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
```diff
@@ -488,7 +488,7 @@ class PlanningSceneMonitor : private boost::noncopyable
   std::unique_ptr<boost::thread> publish_planning_scene_;
   double publish_planning_scene_frequency_;
   SceneUpdateType publish_update_types_;
-  SceneUpdateType new_scene_update_;
+  std::atomic<SceneUpdateType> new_scene_update_;
```
henningkayser, line 491: I'm wondering if this needs to be fixed for ROS1 as well... or if we have some bugs with the mutex locks in here

henningkayser (approved):

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
tylerjw: The issue here is that on one thread we treat this as part of a condition variable and are checking it's state in a loop.  In another thread we do a RMW (read-modify-write) operation on it.  By telling the compiler that this is atomic we make the RMW operation atomoic.

henningkayser: That part is exactly the same as with MoveIt 1, we should probably backport this

tylerjw:
This is failing CI for reasons that will be fixed by @AdamPettinger's next PR:

```
[test_servo_integration-3] Expected: (start_position) != (middle_position), actual: 0.0954851 vs 0.0954851
```

JafarAbdi (approved):

PR #258, opened 2020-08-25
[Servo] Fixes to Servo tests/demos

AdamPettinger:
### Description

Some changes to the tests to hopefully make them less flaky

1. Integration testing: remove the StaleCommandStop test, and instead add that check to the end of SendJointCommand test
2. Unit testing: Change the ServoCalcs unit test to use shared pointers for the ROS node, planning scene monitor, and Servo parameters so they are only loaded at the start of the unit tests instead of for each individual test
3. Minor change to the teleop demo to use a thread to add the collision objects instead of a future
4. Minor spelling corrections

### DIFF
--- a/moveit_ros/moveit_servo/test/unit_test_servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/test/unit_test_servo_calcs.cpp
@@ -74,25 +74,9 @@ FriendServoCalcs::FriendServoCalcs(const rclcpp::Node::SharedPtr& node,
 {
 }
 
-ServoCalcsTestFixture::ServoCalcsTestFixture() : node_(std::make_shared<rclcpp::Node>("servo_calcs_test"))
+ServoCalcsTestFixture::ServoCalcsTestFixture() : node_(TEST_NODE)
 {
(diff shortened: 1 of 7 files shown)


moveit_ros/moveit_servo/test/unit_test_servo_calcs.cpp
```diff
@@ -520,7 +504,33 @@ int main(int argc, char** argv)
   ::testing::InitGoogleTest(&argc, argv);
   rclcpp::init(argc, argv);
 
+  // Set up the shared node
+  TEST_NODE = std::make_shared<rclcpp::Node>("trying_something");
```
tylerjw, line 508: I like the idea.  You should change the name of the node though.

AdamPettinger: Ack, fixed

JafarAbdi (approved):

henningkayser (approved):

henningkayser:
The broken moveit_core test is unrelated to this change. I'll fix it in another PR.

Issue #252, opened 2020-08-20
Duplicate nodes in run_moveit_cpp demo


AdamPettinger:
`###` Description

Running both demo launch files in `run_moveit_cpp` give duplicate nodes:
![image](https://user-images.githubusercontent.com/28165117/90821844-2bee6800-e2f9-11ea-85c9-dff062763695.png)

I believe this is caused by #230 as I can rebase to remove those commits and do not get duplicate nodes.

### Your environment
* ROS Distro: Foxy
* OS Version: Ubuntu 20.04
* Source build: master branch

### Steps to reproduce
[Build moveit2 from source](https://moveit.ros.org/install-moveit2/source/)
`ros2 launch run_moveit_cpp run_moveit_cpp.launch.py`
In a new terminal while that is running:
`ros2 node list`

### Expected behaviour
Rebasing to before that PR was merged gives what I would expect:
![Pre-PSM-PR-node-list](https://user-images.githubusercontent.com/28165117/90822204-b5059f00-e2f9-11ea-97e0-67deec3ef374.png)

henningkayser:
@AdamPettinger are you sure this is still an issue? I thought this was fixed before the referenced PR was merged. I'm not able to reproduce this.

AdamPettinger:
> Possibly useable information: a recent fix for the same kind of problem in navigation2

@mikeferguson That worked like a charm, thanks! Shamelessly copied

PR #262, opened 2020-08-26
Fix duplicate nodes from planning scene monitor private node

AdamPettinger:
### Description

Solves #252. I can change the actual node name if desired. The output of the `ros2 node list` during the demo is now

image

### DIFF
--- a/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
+++ b/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
@@ -152,13 +152,18 @@ PlanningSceneMonitor::PlanningSceneMonitor(const rclcpp::Node::SharedPtr& node,
                                            const std::shared_ptr<tf2_ros::Buffer>& tf_buffer, const std::string& name)
   : monitor_name_(name)
   , node_(node)
-  , pnode_(std::make_shared<rclcpp::Node>(std::string(node_->get_name()) + "_private", node_->get_namespace()))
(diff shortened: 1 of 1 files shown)


moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
```diff
@@ -152,13 +152,18 @@ PlanningSceneMonitor::PlanningSceneMonitor(const rclcpp::Node::SharedPtr& node,
...
   , dt_state_update_(0.0)
   , shape_transform_cache_lookup_wait_time_(0, 0)
   , rm_loader_(rm_loader)
 {
+  std::vector<std::string> new_args = rclcpp::NodeOptions().arguments();
+  new_args.push_back("--ros-args");
+  new_args.push_back("-r");
+  new_args.push_back(std::string("__node:=") + node_->get_name() + "_private");
+  new_args.push_back("--");
+  pnode_ = std::make_shared<rclcpp::Node>("_", "", rclcpp::NodeOptions().arguments(new_args));
```
henningkayser, line 166: Shouldn't the node still go into the same namespace?

henningkayser (commented):
Thanks for fixing this!

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
AdamPettinger: Yep, fixed!

henningkayser (approved):

PR #265, opened 2020-08-26
Poll for service slower in testing to make moveit_servo tests more robust

tylerjw:
### Description

I think these changes should make the tests in moveit_servo more robust.

This is based on my theory that the reason we are seeing failures like this: https://travis-ci.com/github/ros-planning/moveit2/jobs/378055726 is that we are locking up the cpu polling for start service to be ready and once we start doing that we stop loading new librarids in the composable node and the start service is never ready and we get a timeout.

### DIFF
--- a/moveit_ros/moveit_servo/test/servo_launch_test_common.hpp
+++ b/moveit_ros/moveit_servo/test/servo_launch_test_common.hpp
@@ -113,7 +113,7 @@ public:
   {
     // Start client
     client_servo_start_ = node_->create_client<std_srvs::srv::Trigger>("/start_servo");
-    while (!client_servo_start_->wait_for_service(1s))
+    while (!client_servo_start_->service_is_ready())
     {
       if (!rclcpp::ok())
       {
@@ -121,11 +121,12 @@ public:
         return false;
       }
       RCLCPP_INFO(LOGGER, "client_servo_start_ service not available, waiting again...");
+      rclcpp::sleep_for(500ms);
     }
 
     // If we setup the start client, also setup the stop client...
     client_servo_stop_ = node_->create_client<std_srvs::srv::Trigger>("/stop_servo");
-    while (!client_servo_stop_->wait_for_service(1s))
+    while (!client_servo_stop_->service_is_ready())
     {
       if (!rclcpp::ok())
       {
@@ -133,6 +134,7 @@ public:
         return false;
       }
       RCLCPP_INFO(LOGGER, "client_servo_stop_ service not available, waiting again...");
+      rclcpp::sleep_for(500ms);
     }
 
     // Status sub (we need this to check that we've started / stopped)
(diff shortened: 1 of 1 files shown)


henningkayser (commented):
Won't this allow the test to spin forever so that the test times out without us knowing what failed?

AndyZe:
> Won't this allow the test to spin forever so that the test times out without us knowing what failed?

It looks like the potential to get stuck in an infinite loop was there already.

moveit_ros/moveit_servo/test/servo_launch_test_common.hpp
```diff
@@ -113,26 +113,28 @@ class ServoFixture : public ::testing::Test
   {
     // Start client
     client_servo_start_ = node_->create_client<std_srvs::srv::Trigger>("/start_servo");
-    while (!client_servo_start_->wait_for_service(1s))
+    while (!client_servo_start_->service_is_ready())
```
AndyZe, line 116: Check out wait_for_service() defined here:

https://github.com/ros2/rclcpp/blob/bbf2c4cbfb006245c784f39ee8f8f9e56431c91e/rclcpp/include/rclcpp/parameter_client.hpp#L226

It calls wait_for_service_nanoseconds() ... which again calls wait_for_service(). Do you think that recursion is the root cause? In other words, the source of the infinite loop is in rclcpp?

Here's the definition of wait_for_service_nanoseconds():

https://github.com/ros2/rclcpp/blob/633e1157f8aeee9249afddcd0730c3f9bb894fce/rclcpp/src/rclcpp/parameter_client.cpp#L285

moveit_ros/moveit_servo/test/servo_launch_test_common.hpp
```diff
@@ -113,26 +113,28 @@ class ServoFixture : public ::testing::Test
...
     {
       if (!rclcpp::ok())
       {
         RCLCPP_ERROR(LOGGER, "Interrupted while waiting for the service. Exiting.");
         return false;
       }
       RCLCPP_INFO(LOGGER, "client_servo_start_ service not available, waiting again...");
+      rclcpp::sleep_for(500ms);
```
AndyZe, line 124: Putting a sleep in the loop makes a lot of sense

moveit_ros/moveit_servo/test/servo_launch_test_common.hpp
tylerjw: Yes, I think what happens there would be fine if the cpu had enough cores to allow you to lock up one doing that.  It never sleeps in there so it may not ever give the cpu the chance to run servo so it can finish initalizing.

tylerjw:
> > Won't this allow the test to spin forever so that the test times out without us knowing what failed?
> 
> It looks like the potential to get stuck in an infinite loop was there already.

It was already getting stuck in an infinite loop (which is fine because the test timeout will kill it if this never is true).  The improvement here is that instead of a loop that will tie up the cpu constantly checking it goes to sleep and checks every 0.5 seconds to see if it is ready.

henningkayser (approved):

PR #266, opened 2020-08-28
Fix psm/servo parameter conflict

henningkayser:
All parameters are triggering the parameter callback in the PSM node, which results in the PSM rejecting parameters from moveit_servo. I'm not sure if this fix is a good long-term solution, but for now exclusive parameters in the PSM are declared for the private node. Eventually, we might want to reimplement the PSM as a composable node which would allow a cleaner separation.
Additionally, 0a98d55 should fix some warnings with the moveit_servo tests, @AdamPettinger does this make sense to you?

### DIFF
--- a/moveit_ros/moveit_servo/test/launch/servo_launch_test_common.py
+++ b/moveit_ros/moveit_servo/test/launch/servo_launch_test_common.py
@@ -73,7 +73,8 @@ def generate_servo_test_description(*args,
                     package='moveit_servo',
                     plugin='moveit_servo::ServoServer',
                     name='servo_server',
-                    parameters=[servo_params, robot_description, robot_description_semantic])
+                    parameters=[servo_params, robot_description, robot_description_semantic],
+                    extra_arguments=[{'use_intra_process_comm': True}])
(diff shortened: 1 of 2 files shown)


moveit_ros/moveit_servo/test/launch/servo_launch_test_common.py
```diff
@@ -73,7 +73,8 @@ def generate_servo_test_description(*args,
                     package='moveit_servo',
                     plugin='moveit_servo::ServoServer',
                     name='servo_server',
-                    parameters=[servo_params, robot_description, robot_description_semantic])
+                    parameters=[servo_params, robot_description, robot_description_semantic],
+                    extra_arguments=[{'use_intra_process_comm': True}])
```
AdamPettinger, line 77: Oops! Definitely meant to do this for the tests and forgot, thanks Henning!

JensVanhooydonck:
Sorry to comment on this closed issue. But how should one set the parameters moved to the private node? Shouldn't we set these values to the values of the current Node when declaring them?

Issue #268, opened 2020-08-30
High CPU load


AndyZe:
### Description

The moveit_cpp and moveit_servo demos show very high CPU load.

### Your environment
* ROS Distro: Foxy
* OS Version: Ubuntu 20.04
* Source build, master branch

### Steps to reproduce
Run the moveit_servo demo:

`ros2 launch moveit_servo servo_cpp_interface_demo.launch.py`

Run `top`. Observe high CPU load for several nodes (>90%, sometimes >200%). These are the affected nodes: fake_joint_driver, servo_demo, robot_state_pub.

You should see something similar if you run `ros2 launch run_moveit_cpp run_moveit_cpp.launch.py`

### Expected behaviour
Reasonable CPU loads, probably less than 20%.

AndyZe:
Things these 3 nodes have in common:

- They all subscribe to something
- They all publish something

Things they do not have in common:

- They do not all use a MultiThreadedExecutor

AndyZe:
Commenting this line:

`executor->spin();`

And replacing with a loop with sleep drops CPU load by 80%:

```
  while (std::cin.get() != '\n')
  {
    loop_rate.sleep();
  }
```

AndyZe:
Related:  https://github.com/ros-planning/navigation2/pull/866

AdamPettinger:
#270 Does drop the CPU load for me

One thing I have noticed is that the 3 nodes you mentioned (`fake_joint_driver`, `servo_demo`, `robot_state_pub`) all publish or subscribe to `/joint_states`

On both master and with your PR, the `moveit_servo` and `run_moveit_cpp` demos both publish `/joint_states` incredibly fast: 2700 - 3500 hz. I wonder if that is why those nodes are the ones with high load? No other nodes deal with `/joint_states`

AndyZe:
It sounds like we are close! I bet the issue is in fake_joint_driver

AndyZe:
Making this change in fake_joint_driver.cpp is one way to do it. The CPU load is great after this:

```
void FakeJointDriver::update(void)
{
  // only do loopback
  act_dis = cmd_dis;
}
```

```
void FakeJointDriver::update(void)
{
  rclcpp::sleep_for(std::chrono::milliseconds(100));
  // only do loopback
  act_dis = cmd_dis;
}
```

AdamPettinger:
[...]

PR #270, opened 2020-08-30
WIP: Prevent CPU lockup in servo_cpp_interface_demo

AndyZe:
This addresses issue #268 for one executable. If everybody's amenable, there are some other places this type of change might help:

- test/servo_launch_test_common.hpp
- test/test_servo_parameters.cpp
- test/test_servo_parameters_node.cpp
- test/publish_fake_jog_commands.cpp

### DIFF
--- a/moveit_ros/moveit_servo/src/cpp_interface_demo/servo_cpp_interface_demo.cpp
+++ b/moveit_ros/moveit_servo/src/cpp_interface_demo/servo_cpp_interface_demo.cpp
@@ -49,6 +49,7 @@
 using namespace std::chrono_literals;
 
 static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_servo.servo_demo_node.cpp");
+static constexpr double DEFAULT_SPIN_RATE = 1000;
 
 class ServoCppDemo
 {
@@ -186,7 +187,7 @@ int main(int argc, char** argv)
 
   // Spin
   auto executor = std::make_unique<rclcpp::executors::MultiThreadedExecutor>();
-  rclcpp::Rate loop_rate(1000);
+  rclcpp::Rate loop_rate(DEFAULT_SPIN_RATE);
   while (rclcpp::ok())
   {
     executor->spin_node_once(node);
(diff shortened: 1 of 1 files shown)


moveit_ros/moveit_servo/src/cpp_interface_demo/servo_cpp_interface_demo.cpp
```diff
@@ -186,8 +187,12 @@ int main(int argc, char** argv)
 
   // Spin
   auto executor = std::make_unique<rclcpp::executors::MultiThreadedExecutor>();
-  executor->add_node(node);
-  executor->spin();
+  rclcpp::Rate loop_rate(DEFAULT_SPIN_RATE);
+  while (rclcpp::ok())
+  {
+    executor->spin_node_once(node);
```
henningkayser, line 193: Why not use spin_some()?

AndyZe: I'm not aware of the pros/cons of that. Do you have any reason to suggest it? I will try it.

AndyZe: I find that replacing `executor->spin_once(node);` with `executor->spin_some();` does not work -- the robot doesn't move and CPU load is just as high.

AndyZe: Oh, it does work if I do it right. Seems to be functionally equivalent to spin_once() as far as I can tell

AndyZe:
I'm going to add this in other places where it seems useful now.

moveit_ros/moveit_servo/src/cpp_interface_demo/servo_cpp_interface_demo.cpp
mikeferguson: I think the advantage of spin_some() will be when you have multiple nodes on an executor (and possibly some small gains from not adding the node to the executor and then removing it at the end of each call to spin once).

AndyZe:
Marking as WIP until root cause is sorted out better.

PR #286, opened 2020-09-25
Visibility control headers

lilustga:
### Description

Add visibility control headers.
Set the building flag for the respective header in each CMakeLists.txt.
Add public macros in headers where the automatic symbols export is insufficient.

### Checklist


rhaschke (changes requested):
First of all, why did you decided to go for manual visibility control here, but for global visibility in https://github.com/ros-planning/srdfdom/pull/69?

If you go for manual visibility (which in general is more memory efficient and thus should be preferred): Instead of manually writing the boiler-plate visibility-control header files, I suggest to use corresponding cmake functionality, namely [`generate_export_header`](https://cmake.org/cmake/help/v3.0/module/GenerateExportHeader.html).
Please have a look for an example at https://github.com/ros-visualization/rviz/pull/1335

lilustga:
@rhaschke 

These are instances where the global visibility is insufficient. For example global static variables will not get exported properly without manual visibility.

The decision to use the boiler-plate headers follows this logic:
https://github.com/ament/ament_cmake/issues/201#issuecomment-542808202

rhaschke:
> These are instances where the global visibility is insufficient. For example global static variables will not get exported properly without manual visibility.

So, you (need to) mix global and manual visibility? IMO, that's horrible.

> The decision to use the boiler-plate headers follows this logic:
> [ament/ament_cmake#201 (comment)](https://github.com/ament/ament_cmake/issues/201#issuecomment-542808202)

I don't agree with @wjwwood on this. Manually creating those visibility headers is potentially error-prone and auto-creating them isn't that costly I think. Maybe other maintainers can weigh in with their opinion?

wjwwood:
Well we will have to agree to disagree. I don’t think the boilerplate headers are error prone and they haven’t changed in years and years. I won’t repeat my reasoning here since it’s on that issue.

JafarAbdi:
@lilustga I tried to compile this branch with a windows machine but moveit_core is failing because it can't find [boost zlib](https://github.com/ros-planning/moveit2/blob/main/moveit_core/CMakeLists.txt#L32), do you know how to fix it .?

lilustga:
@JafarAbdi That error can be fixed by removing the following from moveit_core/CMakeLists.txt. It is not required for the version of boost that is being used now.

```
# boost::iostreams on Windows depends on boost::zlib	
if(WIN32)	
  set(EXTRA_BOOST_COMPONENTS zlib)	
endif()
```

This PR specifically regards the visibility control headers, I discussed with Henning having specific topic PRs so we could get consensus on the couple main items individually.

This fork contains all the changes to build on windows: https://github.com/lilustga/moveit2/tree/builds_on_windows

moveit_core/robot_model/include/moveit/visibility_control.hpp
```diff
@@ -0,0 +1,72 @@
+// Copyright (c) 2019, Open Source Robotics Foundation, Inc.
```
henningkayser, line 1: Shouldn't this file be in `include/moveit/robot_model`?

moveit_core/robot_state/include/visibility_control.hpp
```diff
@@ -0,0 +1,72 @@
+// Copyright (c) 2019, Open Source Robotics Foundation, Inc.
```
henningkayser, line 1: same for `include/robot_state/visibility_control.hpp`

henningkayser (commented):
@lilustga To me it looks like the include paths might be wrong which could lead to the issues with missing header files you described.

henningkayser (approved):
I see green, awesome!

wolfv:
@lilustga you might be interested in our efforts to add CI to the Moveit project using the packages from the RoboStack project. Our [Foxy stack](https://github.com/robostack/ros-foxy) is currently a bit behind (noetic has much more packages), but with a little help we could create Windows packages so that they could be used as a basis for a CI similar to what we're adding here for Moveit 1: https://github.com/ros-planning/moveit/pull/2604 (specifically this github action: https://github.com/RoboStack/moveit/blob/tests_with_conda_pkgs/.github/workflows/cross_platform_ci.yml)

You can find the currently available packages here: https://robostack.github.io/foxy.html

lilustga:
@wolfv, I'm interested. Cross platform CI would be great! Lets discuss what's needed to get there.

wolfv:
Do you want to join us on our Gitter chat room (Gitter.im/robostack/Lobby)? Or feel free to send me an email and we can discuss by email/video/...
You can find my email in the GitHub profile.

AndyZe:
@ooeygui do you have an opinion on this?

ooeygui:
> @ooeygui do you have an opinion on this?

Hi @AndyZe ,
Lior and I have been working together for MoveIt2 support on Windows. Visibility Control headers are the recommended solution for exposing symbols outside of a binary on Windows. While the `WINDOWS_EXPORT_ALL_SYMBOLS` solution works well for many components (like nodelets), when exposing static member variables from a class (like Qt components), the automatic export does not work, so visibility control is needed. 

We've documented this at the Azure Edge Robotics landing page - https://aka.ms/ros/visibility (and will be adding docs to ros.org as well).

lilustga:
closing because #530 was merged.

PR #291, opened 2020-10-13
moveit2.repos: fix branch to eloquent

rhaschke:
The [docker build for moveit/moveit2:eloquent-ci is broken](https://hub.docker.com/repository/registry-1.docker.io/moveit/moveit2/builds/ec58a1f2-d86e-4e2f-89c2-510b40f4f93a), because it was still referencing to master branch.

### DIFF
--- a/moveit2.repos
+++ b/moveit2.repos
@@ -2,7 +2,7 @@ repositories:
   moveit2:
     type: git
     url: https://github.com/ros-planning/moveit2
-    version: master
+    version: eloquent
   moveit_msgs:
     type: git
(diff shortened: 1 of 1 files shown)


henningkayser (approved):

henningkayser:
We don't actively maintain eloquent, so some changes will still cause CI to fail. Here, we could also freeze `moveit_resources` to an older commit.

rhaschke:
I'm not sure how much effort it is to get eloquent compiling again. If we shouldn't put effort at all, we should in consequence remove the corresponding docker images.

PR #294, opened 2020-10-14
Fix Servo runtime issues

henningkayser:
This resets `ros2_controllers` to the last version that worked with FakeJoint and Servo. We'll switch to the latest versions once there are stable APIs. Also, since the last ros2 package sync, running the component nodes in Servo would throw an InvalidServiceNameException becaues the namespace parameter in the launch files were undefined. Setting them to root "/" fixes this. @AdamPettinger

### DIFF
--- a/moveit2.repos
+++ b/moveit2.repos
@@ -38,7 +38,7 @@ repositories:
   ros2_controllers:
     type: git
     url: https://github.com/ros-controls/ros2_controllers
-    version: d52902538a970e792ecc604416e22ef90a57f29e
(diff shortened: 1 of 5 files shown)


henningkayser:
This PR shouldn't be squashed as it adds to unrelated (but required) fixes to Servo

rhaschke (approved):

rhaschke:
Now you _squashed_ the commits yourself?

henningkayser:
> Now you _squashed_ the commits yourself?

Oh, you're right, that was a fault on my part... I guess it's not worth the effort to reapply this PR

PR #299, opened 2020-10-26
Fix interactive markers not visible in motion planning plugin

JafarAbdi:
### Description

### Checklist

### DIFF
--- a/moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
+++ b/moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
@@ -49,11 +49,14 @@ namespace kinematics_plugin_loader
 {
 rclcpp::Logger LOGGER = rclcpp::get_logger("kinematics_plugin_loader");
 
-rclcpp::ParameterValue declare_parameter(const rclcpp::Node::SharedPtr& node, const std::string& parameter_name)
+rclcpp::Parameter declare_parameter(const rclcpp::Node::SharedPtr& node, const std::string& parameter_name)
 {
   if (!node->has_parameter(parameter_name))
-    return node->declare_parameter(parameter_name);
-  return node->get_parameter(parameter_name).get_parameter_value();
+    node->declare_parameter(parameter_name);
+  rclcpp::Parameter parameter;
+  if (!node->get_parameter(parameter_name, parameter))
+    RCLCPP_DEBUG_STREAM(LOGGER, "Parameter `" << parameter_name << "` doesn't exists");
+  return parameter;
 }
 
 class KinematicsPluginLoader::KinematicsLoaderImpl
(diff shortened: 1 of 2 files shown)


moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -293,27 +303,26 @@ moveit::core::SolverAllocatorFn KinematicsPluginLoader::getLoaderFunction(const
...
-        // TODO(JafarAbdi): revise parameter lookup
         // read the list of plugin names for possible kinematics solvers
         for (const srdf::Model::Group& known_group : known_groups)
         {
           std::string base_param_name = known_group.name_;
           std::string ksolver_param_name = base_param_name + ".kinematics_solver";
           RCLCPP_DEBUG(LOGGER, "Looking for param %s ", ksolver_param_name.c_str());
-          bool found = node_->has_parameter(ksolver_param_name);
-          if (!found)
+          rclcpp::Parameter ksolver_param = declare_parameter(node_, ksolver_param_name);
```
henningkayser, line 312: Why do we need these additional checks? Are we running multiple IK solver instances in the same node namespace? Otherwise, just calling `Node::declare_parameter()` should be sufficient.

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -332,26 +341,23 @@ moveit::core::SolverAllocatorFn KinematicsPluginLoader::getLoaderFunction(const
           }
 
           std::string ksolver_timeout_param_name = base_param_name + ".kinematics_solver_timeout";
+          rclcpp::Parameter ksolver_timeout_param = declare_parameter(node_, ksolver_timeout_param_name);
           if (node_->has_parameter(ksolver_timeout_param_name))
```
henningkayser, line 345: After declaring a parameter, the node parameter will always be set

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -332,26 +341,23 @@ moveit::core::SolverAllocatorFn KinematicsPluginLoader::getLoaderFunction(const
...
           {
-            double ksolver_timeout;
-            rclcpp::Parameter parameter = node_->get_parameter(ksolver_timeout_param_name);
-            if (parameter.get_type() == rclcpp::ParameterType::PARAMETER_DOUBLE)
+            if (ksolver_timeout_param.get_type() == rclcpp::ParameterType::PARAMETER_DOUBLE)
```
henningkayser, line 347: Shouldn't we just declare the parameter as double? That way we don't have to check the different cases

JafarAbdi: It will throw the following exception `rclcpp::exceptions::InvalidParameterTypeException` if the loaded parameter is integer

henningkayser (approved):

JafarAbdi:
I reverted the last commit because the function with one parameter will throw an exception if the parameter is declared and not set https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L680, this's why the tests were throwing exceptions

henningkayser:
This still seems overly complicated to me. A whole lot of code just for loading 4 parameters. But this is an issue that affects a lot of our classes currently. I'll merge this for now as it fixes the interactive marker, but we should still simplify and refactor parameter usage overall.

Issue #261, opened 2020-08-26
Support virtual joints in CurrentStateMonitor


henningkayser:
From https://github.com/ros-planning/moveit2/issues/251#issuecomment-679226135
https://github.com/ros-planning/moveit2/blob/master/moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp#L145

Issue #251, opened 2020-08-18
Planning for a mobile robot


AliaChe:
Hello,

I'm trying to use Moveit2 to plan trajectories for a mobile robot, namely the turtlebot3. I have added a planar virtual joint between the base frame of my robot (base_footprint) and the frame "world" in the srdf file.  I'm struggling now to set goal pose for my robot. I'm using the setGoal() function as follows:

  ` geometry_msgs::msg::PoseStamped goal_pose;
    goal_pose.header.stamp=rclcpp::Clock().now();
    goal_pose.pose.position.x=2;
    goal_pose.pose.position.y=0;
    goal_pose.pose.position.z=0;
    goal_pose.pose.orientation.x=0;
    goal_pose.pose.orientation.y=0;
    goal_pose.pose.orientation.z=0;
    goal_pose.pose.orientation.w=0;
    moveit::planning_interface::PlanningComponent arm("turtlebot", moveit_cpp_);
    arm.setGoal(goal_pose, "base_footprint"); `

Here is the error I got: 
**`[run_TB3_cpp-2] Parsing robot urdf xml string.
[run_TB3_cpp-2] [WARN] [moveit_ros.robot_model_loader]: No kinematics plugins defined. Fill and load kinematics.yaml!
[run_TB3_cpp-2] [WARN] [moveit_ros.planning_scene_monitor.planning_scene_monitor]: The complete state of the robot is not yet known.  Missing virtual_joint
[run_TB3_cpp-2] [WARN] [moveit_ros.planning_scene_monitor.planning_scene_monitor]: The complete state of the robot is not yet known.  Missing virtual_joint
[run_TB3_cpp-2] [WARN] [moveit_ros.planning_scene_monitor.planning_scene_monitor]: The complete state of the robot is not yet known.  Missing virtual_joint
[run_TB3_cpp-2] [WARN] [moveit_kinematic_constraints.kinematic_constraints]: No frame specified for position constraint on link 'base_footprint'!
[run_TB3_cpp-2] [WARN] [moveit_kinematic_constraints.kinematic_constraints]: Orientation constraint for link 'base_footprint' is probably incorrect: 0.000000, 0.000000, 0.000000, 0.000000. Assuming identity instead.
[run_TB3_cpp-2] [WARN] [moveit_kinematic_constraints.kinematic_constraints]: No frame specified for position constraint on link 'base_footprint'!
[...]

PR #310, opened 2020-11-18
Create a transform subscribers to enable virtual joints

JafarAbdi:
### Description

Fix: https://github.com/ros-planning/moveit2/issues/261
Fix: https://github.com/ros-planning/moveit2/issues/251

### Checklist


moveit_ros/move_group/src/move_group.cpp
```diff
@@ -200,7 +200,6 @@ int main(int argc, char** argv)
 
   std::shared_ptr<tf2_ros::Buffer> tf_buffer =
       std::make_shared<tf2_ros::Buffer>(nh->get_clock(), tf2::durationFromSec(10.0));
-  std::shared_ptr<tf2_ros::TransformListener> tfl = std::make_shared<tf2_ros::TransformListener>(*tf_buffer);
```
henningkayser, line 203: We shouldn't rely on the CurrentStateMonitor to fill the tf_buffer. The TransformListener is required for the PlanningSceneMonitor to work without it.

moveit_ros/planning_interface/moveit_cpp/src/moveit_cpp.cpp
```diff
@@ -61,7 +60,6 @@ MoveItCpp::MoveItCpp(const rclcpp::Node::SharedPtr& node, const Options& options
 {
   if (!tf_buffer_)
     tf_buffer_ = std::make_shared<tf2_ros::Buffer>(node_->get_clock());
-  tf_listener_ = std::make_shared<tf2_ros::TransformListener>(*tf_buffer_);
```
henningkayser, line 64: Here the same, we should run a TransformListener independent of the CSM.

henningkayser:
I thought about this and I don't think we should merge this solution. We should have an independent TransformListener running at all times. If that's not possible, we might implement our own listener that supports callbacks.

schornakj:
@henningkayser  I can take a look at this.

henningkayser:
@schornakj TODO: assign

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
```diff
@@ -142,10 +143,14 @@ void planning_scene_monitor::CurrentStateMonitor::startStateMonitor(const std::s
...
-      // tf_connection_.reset(new TFConnection(
-      //     tf_buffer_->_addTransformsChangedListener(std::bind(&CurrentStateMonitor::tfCallback, this))));
+      using callback_t = std::function<void(const tf2_msgs::msg::TFMessage::ConstSharedPtr)>;
+      tf_buffer_->setUsingDedicatedThread(true);
+      transform_subscriber_ = node_->create_subscription<tf2_msgs::msg::TFMessage>(
+          "/tf", tf2_ros::DynamicListenerQoS(25),
+          callback_t(std::bind(&CurrentStateMonitor::transfromCallback, this, std::placeholders::_1, false)));
+      static_transfrom_subscriber_ = node_->create_subscription<tf2_msgs::msg::TFMessage>(
+          "/tf_static", tf2_ros::StaticListenerQoS(25),
+          callback_t(std::bind(&CurrentStateMonitor::transfromCallback, this, std::placeholders::_1, true)));
```
schornakj, line 153: Would it make sense to wrap all this in a new class that takes `std::shared_ptr<rclcpp::Node>` and a `bind` to a callback like `updateMultiDofJoints`? That seems like a useful bit of capability to have as a portable building block. As written it seems like things are a bit scattered between different files.

(I think @henningkayser had suggested something like this previously).

moveit_ros/move_group/src/move_group.cpp
schornakj: I think this comment is obsolete now, since parallel functionality was implemented through subscribers on the `/tf` topic.

moveit_ros/planning_interface/moveit_cpp/src/moveit_cpp.cpp
schornakj: Same with this comment.

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
schornakj: I tried implementing the class I described earlier and I think I've changed my mind about it being especially useful.

The `transformCallback` needs to have some built-in logic specifically designed around handling static and non-static TF callbacks from a pair of subscribers, so it's not as straightforward to allow a user to pass in an arbitrary callback and have it work in an equivalent way to the current method.

DLu:
@JafarAbdi Any thoughts on addressing @schornakj 's comments?

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
```diff
@@ -449,8 +460,13 @@ void planning_scene_monitor::CurrentStateMonitor::tfCallback()
         continue;
       }
 
+      // TODO(JafarAbdi): Replace with insert_or_assign once we have C++17 support
```
henningkayser, line 463: Could you open an issue for C++17? ;) I'd be fully on board with this if it doesn't come with unwanted side effects.

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
```diff
@@ -432,7 +443,7 @@ void planning_scene_monitor::CurrentStateMonitor::tfCallback()
       const std::string& parent_frame =
           joint->getParentLinkModel() ? joint->getParentLinkModel()->getName() : robot_model_->getModelFrame();
 
-      rclcpp::Time latest_common_time;
+      rclcpp::Time latest_common_time = rclcpp::Time(0, 0, RCL_ROS_TIME);
```
henningkayser, line 446:
```suggestion
      rclcpp::Time latest_common_time(0, RCL_ROS_TIME);
```

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
```diff
@@ -488,3 +504,26 @@ void planning_scene_monitor::CurrentStateMonitor::tfCallback()
...
+                               is_static);
+    }
+    catch (tf2::TransformException& ex)
+    {
+      std::string temp = ex.what();
+      RCLCPP_ERROR(LOGGER, "Failure to set recieved transform from %s to %s with error: %s\n",
+                   transform.child_frame_id.c_str(), transform.header.frame_id.c_str(), temp.c_str());
+    }
+  }
+  updateMultiDofJoints();
```
henningkayser, line 528: isn't this redundant in case of a transform exception?

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/current_state_monitor.h
```diff
@@ -182,7 +180,8 @@ class CurrentStateMonitor
 
 private:
   void jointStateCallback(const sensor_msgs::msg::JointState::ConstSharedPtr joint_state);
-  void tfCallback();
+  void updateMultiDofJoints();
+  void transfromCallback(const tf2_msgs::msg::TFMessage::ConstSharedPtr msg, bool is_static);
```
henningkayser, line 184:
```suggestion
  void transformCallback(const tf2_msgs::msg::TFMessage::ConstSharedPtr msg, bool is_static);
```

moveit_ros/move_group/src/move_group.cpp
henningkayser: How is this comment obsolete? The subscribers are initialized inside the CSM so unless we call `PSM::startStateMonitor()` there won't be any listeners updating the buffer. A reasonable workaround would be to move the transform subscribers into the PSM and to call `CSM::updateMultiDofJoints` from there. What do you think?

henningkayser (commented):
Sorry for letting the reviews pend for so long. I think this PR is really getting there but the original complaint doesn't seem to be fixed, which is that the PSM doesn't update TF if the CSM is not initialized (please correct me if I'm wrong). This could for instance break the planning scene display as that one doesn't listen for joint states explicitly. The callback implementation looks good to me on its own but please consider moving them into the PSM and updating the transforms from there (see comment).

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
JafarAbdi: What if the exception happened after adding few transforms to the buffer .?

moveit_ros/move_group/src/move_group.cpp
JafarAbdi: We do always call [startStateMonitor](https://github.com/ros-planning/moveit2/blob/main/moveit_ros/planning/moveit_cpp/src/moveit_cpp.cpp#L108) when constructing move_group

moveit_ros/planning_interface/moveit_cpp/src/moveit_cpp.cpp
JafarAbdi: See my previous comment

JafarAbdi:
I think this PR is ready to be merged, regarding having an independent transform listener for PSM I didn't change the behavior from ROS1, the user have to provide it themselves unless a current state monitor is also enabled in that case a warning message will be print telling that the transforms will be added by both transform listener & current state monitor's callbacks

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
```diff
@@ -401,8 +408,11 @@ void CurrentStateMonitor::tfCallback()
...
       }
 
+      if (auto joint_time_it = joint_time_.find(joint); joint_time_it == joint_time_.end())
+        joint_time_.emplace(joint, rclcpp::Time(0, 0, RCL_ROS_TIME));
+
       // allow update if time is more recent or if it is a static transform (time = 0)
-      if (latest_common_time <= joint_time_[joint] && latest_common_time > rclcpp::Time(0))
+      if (latest_common_time <= joint_time_[joint] && latest_common_time > rclcpp::Time(0, 0, RCL_ROS_TIME))
         continue;
       joint_time_[joint] = latest_common_time;
```
nbbrooks, line 417: `// allow update if time is more recent or if it is a static transform (time = 0)`

This block confuses me. 
- `latest_common_time` is always going to be 0, so seems like a misnomer
- because `latest_common_time` doesn't change, the second condition in the line 415 if statement will always be false, making the if never trigger. So why do we have it?

nbbrooks, line 417:
```suggestion
      // allow update if time is more recent or if it is a static transform (time = 0)
      joint_time_[joint] = latest_common_time;
```

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
```diff
@@ -401,8 +408,11 @@ void CurrentStateMonitor::tfCallback()
         continue;
       }
 
+      if (auto joint_time_it = joint_time_.find(joint); joint_time_it == joint_time_.end())
+        joint_time_.emplace(joint, rclcpp::Time(0, 0, RCL_ROS_TIME));
+
       // allow update if time is more recent or if it is a static transform (time = 0)
```
nbbrooks, line 414:
```suggestion
      // skip if time is not more recent and is not a static transform (time = 0)
```
Comments should use the same logic as the conditionals, don't make the person parsing it invert conditions.

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
JafarAbdi: How it's not changing .? if you look after few lines we set it to `transf.header.stamp`

nbbrooks: Bizarre, I thought I had everything expanded and ctrl + f still wasn't yielding anything.

henningkayser (approved):
LGTM! Also, thanks for the tests @JafarAbdi. This fix is good for now, but I think this PR exposes serious flaws in the PSM API that should be addressed.

Using an external TF listener is discouraged now in case CSM is used since transforms callbacks would be duplicate. This should be documented and ideally checked for inside the PSM constructor. If the CSM is not used, an external listener is strictly required to keep transforms updated. For the PSM it means that it's highly questionable if tf_buffer receives updated transforms or not. Obviously, some of this confusion was already present in ROS 1, but this PR makes it just a little bit more confusing.

One way to guarantee that PSM has a valid listener would be to:
* Remove/Deprecate tf_buffer from PSM constructors
* Initialize a private tf_listener inside PSM, possibly using startTFListener() to make this optional
* Stop tf_listener with `startStateMonitor()`, restart it with `stopStateMonitor()`

(this is definitely out of scope of this PR)

tylerjw (approved):
Thank you for adding this.  I generally agree with Henning that this part of the code needs a closer look to make the correct use of the API more obvious.

PR #320, opened 2020-12-09
port pose tracking

Abishalini:
### Description

Ported pose tracking from moveit_servo to ROS2

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/CMakeLists.txt
+++ b/moveit_ros/moveit_servo/CMakeLists.txt
@@ -76,16 +76,16 @@ target_link_libraries(${CPP_DEMO_NAME} ${SERVO_LIB_NAME})
 ## Library for servoing toward a pose ##
 ########################################
 
-# TODO(henningkayser): Port pose tracking and demos in cpp_interface_example to ROS 2
-# add_library(${POSE_TRACKING} SHARED src/pose_tracking_demo.cpp)
-# ament_target_dependencies(${POSE_TRACKING} ${THIS_PACKAGE_INCLUDE_DEPENDS})
-# add_dependencies(${POSE_TRACKING} ${catkin_EXPORTED_TARGETS})
-# target_link_libraries(${POSE_TRACKING} ${SERVO_LIB_NAME})
-#
+set(POSE_TRACKING pose_tracking)
+add_library(${POSE_TRACKING} SHARED src/pose_tracking.cpp)
+ament_target_dependencies(${POSE_TRACKING} ${THIS_PACKAGE_INCLUDE_DEPENDS})
+target_link_libraries(${POSE_TRACKING} ${SERVO_LIB_NAME})
+
 # An example of pose tracking
-# set(POSE_TRACKING_DEMO_NAME servo_pose_tracking_demo)
-# add_executable(${POSE_TRACKING_DEMO_NAME} src/cpp_interface_example/pose_tracking_demo.cpp)
-# ament_target_dependencies(${POSE_TRACKING_DEMO_NAME} ${THIS_PACKAGE_INCLUDE_DEPENDS})
+set(POSE_TRACKING_DEMO_NAME servo_pose_tracking_demo)
+add_executable(${POSE_TRACKING_DEMO_NAME} src/cpp_interface_demo/pose_tracking_demo.cpp)
+target_link_libraries(${POSE_TRACKING_DEMO_NAME} ${POSE_TRACKING})
+ament_target_dependencies(${POSE_TRACKING_DEMO_NAME} ${THIS_PACKAGE_INCLUDE_DEPENDS})
 
 # Add executable for using a controller
 set(SERVO_CONTROLLER_INPUT servo_controller_input)
@@ -110,6 +110,8 @@ install(
     ${SERVO_CONTROLLER_INPUT}
     ${CPP_DEMO_NAME}
     ${FAKE_SERVO_CMDS_NAME}
+    ${POSE_TRACKING}
+    ${POSE_TRACKING_DEMO_NAME}
   ARCHIVE DESTINATION lib
   LIBRARY DESTINATION lib
   RUNTIME DESTINATION bin
@@ -190,15 +192,13 @@ if(BUILD_TESTING)
   #   ${catkin_LIBRARIES}
   # )
 
-  # # pose_tracking
-  # add_rostest_gtest(pose_tracking_test
-  #   test/pose_tracking_test.test
-  #   test/pose_tracking_test.cpp
-  # )
-  # target_link_libraries(pose_tracking_test
-  #   pose_tracking
-  #   ${catkin_LIBRARIES}
-  # )
+  # pose_tracking
+  ament_add_gtest_executable(test_servo_pose_tracking
+        test/pose_tracking_test.cpp
+  )
+  ament_target_dependencies(test_servo_pose_tracking ${THIS_PACKAGE_INCLUDE_DEPENDS})
+  target_link_libraries(test_servo_pose_tracking ${POSE_TRACKING})
+  add_ros_test(test/launch/test_servo_pose_tracking.test.py ARGS "test_binary_dir:=${CMAKE_CURRENT_BINARY_DIR}")
 
   # # low_latency
   # add_rostest_gtest(servo_low_latency_test
--- a/moveit_ros/moveit_servo/include/moveit_servo/make_shared_from_pool.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/make_shared_from_pool.h
@@ -38,7 +38,7 @@
 
 #pragma once
 
-#include <boost/make_shared.hpp>
+#include <memory>
 #include <boost/pool/pool_alloc.hpp>
 
 namespace moveit
@@ -47,10 +47,10 @@ namespace util
 {
 // Useful template for creating messages from a message pool
 template <typename T>
-boost::shared_ptr<T> make_shared_from_pool()
+std::shared_ptr<T> make_shared_from_pool()
 {
-  using allocator_t = boost::fast_pool_allocator<boost::shared_ptr<T>>;
-  return boost::allocate_shared<T, allocator_t>(allocator_t());
+  using allocator_t = boost::fast_pool_allocator<std::shared_ptr<T>>;
+  return std::allocate_shared<T, allocator_t>(allocator_t());
 }
 
 }  // namespace util
--- a/moveit_ros/moveit_servo/include/moveit_servo/pose_tracking.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/pose_tracking.h
@@ -39,13 +39,13 @@
 #pragma once
 
 #include <atomic>
-#include <boost/optional.hpp>
-#include <control_toolbox/pid.h>
+#include <control_toolbox/pid.hpp>
 #include <moveit_servo/make_shared_from_pool.h>
 #include <moveit_servo/servo.h>
-#include <rosparam_shortcuts/rosparam_shortcuts.h>
+#include <moveit_servo/servo_parameters.h>
 #include <tf2_geometry_msgs/tf2_geometry_msgs.h>
 #include <tf2_ros/transform_listener.h>
+#include <rclcpp/rclcpp.hpp>
 
 // Conventions:
 // Calculations are done in the planning_frame_ unless otherwise noted.
@@ -86,7 +86,8 @@ class PoseTracking
 {
 public:
   /** \brief Constructor. Loads ROS parameters under the given namespace. */
-  PoseTracking(const ros::NodeHandle& nh, const planning_scene_monitor::PlanningSceneMonitorPtr& planning_scene_monitor);
+  PoseTracking(const rclcpp::Node::SharedPtr& node, const ServoParametersPtr& servo_parameters,
+               const planning_scene_monitor::PlanningSceneMonitorPtr& planning_scene_monitor);
 
   PoseTrackingStatusCode moveToPose(const Eigen::Vector3d& positional_tolerance, const double angular_tolerance,
                                     const double target_pose_timeout);
@@ -113,7 +114,7 @@ public:
    * @param transform the transform that will be calculated
    * @return true if a valid transform was available
    */
-  bool getCommandFrameTransform(geometry_msgs::TransformStamped& transform);
+  bool getCommandFrameTransform(geometry_msgs::msg::TransformStamped& transform);
 
   /** \brief Re-initialize the target pose to an empty message. Can be used to reset motion between waypoints. */
   void resetTargetPose();
@@ -138,7 +139,7 @@ private:
   bool satisfiesPoseTolerance(const Eigen::Vector3d& positional_tolerance, const double angular_tolerance);
 
   /** \brief Subscribe to the target pose on this topic */
-  void targetPoseCallback(const geometry_msgs::PoseStampedConstPtr& msg);
+  void targetPoseCallback(const geometry_msgs::msg::PoseStamped::ConstSharedPtr msg);
 
   /** \brief Update PID controller target positions & orientations */
   void updateControllerSetpoints();
@@ -147,23 +148,24 @@ private:
   void updateControllerStateMeasurements();
 
   /** \brief Use PID controllers to calculate a full spatial velocity toward a pose */
-  geometry_msgs::TwistStampedConstPtr calculateTwistCommand();
+  geometry_msgs::msg::TwistStamped::ConstSharedPtr calculateTwistCommand();
 
   /** \brief Reset flags and PID controllers after a motion completes */
   void doPostMotionReset();
 
-  ros::NodeHandle nh_;
+  rclcpp::Node::SharedPtr node_;
+  moveit_servo::ServoParametersPtr servo_parameters_;
 
   planning_scene_monitor::PlanningSceneMonitorPtr planning_scene_monitor_;
-  robot_model::RobotModelConstPtr robot_model_;
+  moveit::core::RobotModelConstPtr robot_model_;
   const moveit::core::JointModelGroup* joint_model_group_;
   // Joint group used for controlling the motions
   std::string move_group_name_;
 
-  ros::Rate loop_rate_;
+  rclcpp::Rate loop_rate_;
 
   // ROS interface to Servo
-  ros::Publisher twist_stamped_pub_;
+  rclcpp::Publisher<geometry_msgs::msg::TwistStamped>::SharedPtr twist_stamped_pub_;
 
   std::vector<control_toolbox::Pid> cartesian_position_pids_;
   std::vector<control_toolbox::Pid> cartesian_orientation_pids_;
@@ -172,12 +174,12 @@ private:
 
   // Transforms w.r.t. planning_frame_
   Eigen::Isometry3d command_frame_transform_;
-  ros::Time command_frame_transform_stamp_;
-  geometry_msgs::PoseStamped target_pose_;
+  rclcpp::Time command_frame_transform_stamp_ = rclcpp::Time(0, 0, RCL_ROS_TIME);
+  geometry_msgs::msg::PoseStamped target_pose_;
   mutable std::mutex target_pose_mtx_;
 
   // Subscribe to target pose
-  ros::Subscriber target_pose_sub_;
+  rclcpp::Subscription<geometry_msgs::msg::PoseStamped>::SharedPtr target_pose_sub_;
 
   tf2_ros::Buffer transform_buffer_;
   tf2_ros::TransformListener transform_listener_;
--- a/moveit_ros/moveit_servo/launch/pose_tracking_example.launch.py
+++ b/moveit_ros/moveit_servo/launch/pose_tracking_example.launch.py
@@ -1,0 +1,88 @@
+import os
+import yaml
+from launch import LaunchDescription
+from launch_ros.actions import Node
+from ament_index_python.packages import get_package_share_directory
+
+def load_file(package_name, file_path):
+    package_path = get_package_share_directory(package_name)
+    absolute_file_path = os.path.join(package_path, file_path)
+
+    try:
+        with open(absolute_file_path, 'r') as file:
+            return file.read()
+    except EnvironmentError: # parent of IOError, OSError *and* WindowsError where available
+        return None
+
+def load_yaml(package_name, file_path):
+    package_path = get_package_share_directory(package_name)
+    absolute_file_path = os.path.join(package_path, file_path)
+
+    try:
+        with open(absolute_file_path, 'r') as file:
+            return yaml.safe_load(file)
+    except EnvironmentError: # parent of IOError, OSError *and* WindowsError where available
+        return None
+
+
+def generate_launch_description():
+    # Get URDF and SRDF
+    robot_description_config = load_file('moveit_resources_panda_description', 'urdf/panda.urdf')
+    robot_description = {'robot_description' : robot_description_config}
+
+    robot_description_semantic_config = load_file('moveit_resources_panda_moveit_config', 'config/panda.srdf')
+    robot_description_semantic = {'robot_description_semantic' : robot_description_semantic_config}
+
+    # Get parameters for the Pose Tracking node
+    pose_tracking_yaml = load_yaml('moveit_servo', 'config/pose_tracking_settings.yaml')
+    pose_tracking_params = { 'moveit_servo' : pose_tracking_yaml }
+
+    # Get parameters for the Servo node
+    servo_yaml = load_yaml('moveit_servo', 'config/panda_simulated_config_pose_tracking.yaml')
+    servo_params = { 'moveit_servo' : servo_yaml }
+
+    kinematics_yaml = load_yaml('moveit_resources_panda_moveit_config', 'config/kinematics.yaml')
+    
+    #RViz
+    rviz_config_file = get_package_share_directory('moveit_servo') + "/config/demo_rviz_pose_tracking.rviz"
+    rviz_node = Node(package='rviz2',
+                     executable='rviz2',
+                     name='rviz2',
+                     #prefix=['xterm -e gdb -ex run --args'],
+                     output='log',
+                     arguments=['-d', rviz_config_file],
+                     parameters=[robot_description, robot_description_semantic, kinematics_yaml])
+
+    # Publishes tf's for the robot
+    robot_state_publisher = Node(
+        package='robot_state_publisher',
+        executable='robot_state_publisher',
+        output='screen',
+        parameters=[robot_description]
+    )
+
+    # A node to publish world -> panda_link0 transform
+    static_tf = Node(package='tf2_ros',
+                     executable='static_transform_publisher',
+                     name='static_transform_publisher',
+                     output='log',
+                     arguments=['0.0', '0.0', '0.0', '0.0', '0.0', '0.0', 'world', 'panda_link0'])
+
+    pose_tracking_node = Node(
+        package='moveit_servo',
+        executable='servo_pose_tracking_demo',
+        #prefix=['xterm -e gdb -ex run --args'],
+        output='screen',
+        parameters=[robot_description, robot_description_semantic, kinematics_yaml, pose_tracking_params, servo_params]
+    )
+
+    # Fake joint driver
+    fake_joint_driver_node = Node(package='fake_joint_driver',
+                                  executable='fake_joint_driver_node',
+                                  parameters=[{'controller_name': 'fake_joint_trajectory_controller'},
+                                              os.path.join(get_package_share_directory("moveit_servo"), "config", "panda_controllers.yaml"),
+                                              os.path.join(get_package_share_directory("moveit_servo"), "config", "start_positions.yaml"),
+                                              robot_description])
+
+    return LaunchDescription([rviz_node, static_tf, pose_tracking_node, fake_joint_driver_node, robot_state_publisher])
+
--- a/moveit_ros/moveit_servo/src/cpp_interface_demo/pose_tracking_demo.cpp
+++ b/moveit_ros/moveit_servo/src/cpp_interface_demo/pose_tracking_demo.cpp
@@ -87,7 +87,7 @@ int main(int argc, char** argv)
 
   rclcpp::executors::SingleThreadedExecutor executor;
   executor.add_node(node);
-  std::thread([&executor]() { executor.spin(); }).detach();
+  std::thread executor_thread([&executor]() { executor.spin(); });
 
   moveit_servo::ServoParametersPtr servo_parameters;
   servo_parameters = std::make_shared<moveit_servo::ServoParameters>();
@@ -179,6 +179,10 @@ int main(int argc, char** argv)
   // Make sure the tracker is stopped and clean up
   tracker.stopMotion();
   move_to_pose_thread.join();
+
+  // Kill executor thread before shutdown
+  executor.cancel();
+  executor_thread.join();
 
   rclcpp::shutdown();
   return EXIT_SUCCESS;
--- a/moveit_ros/moveit_servo/src/cpp_interface_demo/pose_tracking_demo.cpp
+++ b/moveit_ros/moveit_servo/src/cpp_interface_demo/pose_tracking_demo.cpp
@@ -116,7 +116,11 @@ int main(int argc, char** argv)
   planning_scene_monitor->startPublishingPlanningScene(planning_scene_monitor::PlanningSceneMonitor::UPDATE_SCENE);
 
   // Wait for Planning Scene Monitor to setup
-  rclcpp::sleep_for(std::chrono::seconds(5));
+  if (!planning_scene_monitor->waitForCurrentRobotState(node->now(), 5.0 /* seconds */))
+  {
+    RCLCPP_ERROR_STREAM(LOGGER, "Error waiting for current robot state in PlanningSceneMonitor.");
+    exit(EXIT_FAILURE);
+  }
 
   // Create the pose tracker
   moveit_servo::PoseTracking tracker(node, servo_parameters, planning_scene_monitor);
@@ -172,9 +176,6 @@ int main(int argc, char** argv)
 
     loop_rate.sleep();
   }
-
-  rclcpp::sleep_for(std::chrono::seconds(5));
-  RCLCPP_DEBUG_STREAM(LOGGER, "Timer expired and unfinished motion will be terminated");
 
   // Make sure the tracker is stopped and clean up
   tracker.stopMotion();
--- a/moveit_ros/moveit_servo/src/cpp_interface_demo/pose_tracking_demo.cpp
+++ b/moveit_ros/moveit_servo/src/cpp_interface_demo/pose_tracking_demo.cpp
@@ -178,7 +178,6 @@ int main(int argc, char** argv)
   }
 
   // Make sure the tracker is stopped and clean up
-  tracker.stopMotion();
   move_to_pose_thread.join();
 
   // Kill executor thread before shutdown
--- a/moveit_ros/moveit_servo/src/cpp_interface_demo/pose_tracking_demo.cpp
+++ b/moveit_ros/moveit_servo/src/cpp_interface_demo/pose_tracking_demo.cpp
@@ -36,39 +36,43 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  *******************************************************************************/
 
-#include <std_msgs/Int8.h>
-#include <geometry_msgs/TransformStamped.h>
+#include <std_msgs/msg/int8.hpp>
+#include <geometry_msgs/msg/transform_stamped.hpp>
 
 #include <moveit_servo/servo.h>
 #include <moveit_servo/pose_tracking.h>
 #include <moveit_servo/status_codes.h>
+#include <moveit_servo/servo_parameters.h>
+#include <moveit_servo/servo_parameters.cpp>
 #include <moveit_servo/make_shared_from_pool.h>
 #include <thread>
 
-static const std::string LOGNAME = "cpp_interface_example";
+static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_servo.pose_tracking_demo");
 
 // Class for monitoring status of moveit_servo
 class StatusMonitor
 {
 public:
-  StatusMonitor(ros::NodeHandle& nh, const std::string& topic)
+  StatusMonitor(const rclcpp::Node::SharedPtr& node, const std::string& topic)
   {
-    sub_ = nh.subscribe(topic, 1, &StatusMonitor::statusCB, this);
+    sub_ = node->create_subscription<std_msgs::msg::Int8>(
+        topic, 1, std::bind(&StatusMonitor::statusCB, this, std::placeholders::_1));
   }
 
 private:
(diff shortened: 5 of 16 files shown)


moveit_ros/moveit_servo/src/pose_tracking.cpp
```diff
@@ -64,37 +66,37 @@ PoseTracking::PoseTracking(const ros::NodeHandle& nh,
...
 PoseTrackingStatusCode PoseTracking::moveToPose(const Eigen::Vector3d& positional_tolerance,
                                                 const double angular_tolerance, const double target_pose_timeout)
 {
   // Wait a bit for a target pose message to arrive.
   // The target pose may get updated by new messages as the robot moves (in a callback function).
-  const ros::Time start_time = ros::Time::now();
+  const rclcpp::Time start_time = node_->now();
   while ((!haveRecentTargetPose(target_pose_timeout) || !haveRecentEndEffectorPose(target_pose_timeout)) &&
-         ((ros::Time::now() - start_time).toSec() < target_pose_timeout))
+         ((node_->now() - start_time).to_chrono<std::chrono::duration<double>>().count() < target_pose_timeout))
```
henningkayser, line 88:
```suggestion
         ((node_->now() - start_time).seconds() < target_pose_timeout))
```

Can't we just use [seconds()](http://docs.ros2.org/foxy/api/rclcpp/classrclcpp_1_1Duration.html#a2294b94b1434fc532d2f153757ed4ba6)?

moveit_ros/moveit_servo/src/pose_tracking.cpp
```diff
@@ -64,37 +66,37 @@ PoseTracking::PoseTracking(const ros::NodeHandle& nh,
...
   {
     if (servo_->getCommandFrameTransform(command_frame_transform_))
     {
-      command_frame_transform_stamp_ = ros::Time::now();
+      command_frame_transform_stamp_ = node_->now();
     }
-    ros::Duration(0.001).sleep();
+    std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
henningkayser, line 94: How about using rclcpp::Duration with chrono literals?
```suggestion
    rclcpp::Duration(1ms).sleep();
```

moveit_ros/moveit_servo/src/pose_tracking.cpp
```diff
@@ -137,63 +139,70 @@ PoseTrackingStatusCode PoseTracking::moveToPose(const Eigen::Vector3d& positiona
...
-  while (ros::ok() && !nh.hasParam("planning_frame") && ((ros::Time::now() - begin).toSec() < ROS_STARTUP_WAIT))
+  rclcpp::Time begin = node->now();
+  while (rclcpp::ok() && !node->has_parameter("planning_frame") && ((node->now() - begin).to_chrono<std::chrono::duration<double>>().count() < ROS_STARTUP_WAIT))
   {
-    ROS_WARN_STREAM_NAMED(LOGNAME, "Waiting for parameter: "
-                                       << "planning_frame");
-    ros::Duration(0.1).sleep();
+    RCLCPP_WARN_STREAM(LOGGER, "Waiting for parameter: "
+                                   << "planning_frame");
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
henningkayser, line 162: same

moveit_ros/moveit_servo/src/pose_tracking.cpp
```diff
@@ -137,63 +139,70 @@ PoseTrackingStatusCode PoseTracking::moveToPose(const Eigen::Vector3d& positiona
...
     ++error;
-    ROS_ERROR_STREAM_NAMED(LOGNAME, "Unable to find the specified joint model group: " << move_group_name_);
+    RCLCPP_ERROR_STREAM(LOGGER, "Unable to find the specified joint model group: " << move_group_name_);
   }
 
   double publish_period;
-  error += !rosparam_shortcuts::get(LOGNAME, nh, "publish_period", publish_period);
-  loop_rate_ = ros::Rate(1 / publish_period);
+  error += !PoseTracking::rosparam_shortcut_get(LOGGER, node, "publish_period", publish_period);
+  rclcpp::Rate loop_rate_(publish_period);
```
henningkayser, line 177: I think double values are initialized as rate, so we should probably stick to using `1 / publish_period`.

moveit_ros/moveit_servo/test/pose_tracking_test.cpp
```diff
@@ -63,24 +63,48 @@ class PoseTrackingFixture : public ::testing::Test
...
+      bool have_message = false;
+      auto sub = node_.create_subscription<sensor_msgs::msg::JointState>(
+          "joint_states", 1,
+          std::function<void(const sensor_msgs::msg::JointState::ConstSharedPtr)>(
+              [have_message](const auto msg) mutable { have_message = true; }));
+      while (!have_message)
+        ;
+    
+
+    while (!node_.has_parameter("/robot_description") && rclcpp::ok())
```
henningkayser, line 87: Parameter names don't use "/" anymore. parameter separators are "." in ROS 2
```suggestion
    while (!node_.has_parameter("robot_description") && rclcpp::ok())
```

henningkayser: However, we can probably remove this while loop. It shouldn't be necessary at all.

moveit_ros/moveit_servo/test/pose_tracking_test.cpp
```diff
@@ -63,24 +63,48 @@ class PoseTrackingFixture : public ::testing::Test
   void SetUp() override
   {
     // Wait for several key topics / parameters
-    ros::topic::waitForMessage<sensor_msgs::JointState>("/joint_states");
-    while (!nh_.hasParam("/robot_description") && ros::ok())
+    
+    //TODO: Need to figure out replacement. How to use WaitSet?
```
henningkayser, line 67: No need to do it this way anymore. Simply call [`PlanningSceneMonitor::waitForCurrentRobotState()`](http://docs.ros.org/en/noetic/api/moveit_ros_planning/html/classplanning__scene__monitor_1_1PlanningSceneMonitor.html#a75135c3e95ce2c6dd9a5c379ca4bfdb5) after initializing the PSM.

moveit_ros/moveit_servo/test/pose_tracking_test.cpp
```diff
@@ -90,21 +114,23 @@ class PoseTrackingFixture : public ::testing::Test
   }
 
 protected:
-  ros::NodeHandle nh_{ "~" };
+  //ros::NodeHandle nh_{ "~" };
+  rclcpp::Node node_{ "~" };
```
henningkayser, line 118: We should probably have a proper node name and initialize it in the constructor

henningkayser (commented):
Thanks for this first attempt! Looks good so far

install/.colcon_install_layout
```diff
@@ -0,0 +1 @@
+isolated
```
JafarAbdi, line 1: @Abishalini Please remove this directory

Abishalini: I have removed the log directories and updated ./colcon_install_layout

moveit_ros/moveit_servo/include/moveit_servo/pose_tracking.h
```diff
@@ -188,6 +191,12 @@ class PoseTracking
   std::atomic<bool> stop_requested_;
 
   double angular_error_;
+  
+  //rosparam_shortcut
+  template <class T>
+  bool rosparam_shortcut_get(const rclcpp::Logger &parent_name, const rclcpp::Node::SharedPtr& node, const std::string &param_name, T &value);
```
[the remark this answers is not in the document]
Abishalini: This has been replaced with the declareOrGetParam function from servo_parameters.cpp

moveit_ros/moveit_servo/src/pose_tracking.cpp
Abishalini: rclcpp::Duration doesn't seem to have a sleep function

install/.colcon_install_layout
```diff
@@ -0,0 +1 @@
+
```
henningkayser, line 1: This looks like colcon artifacts which should be removed

moveit_ros/moveit_servo/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 cmake_minimum_required(VERSION 3.5)
 project(moveit_servo)
+#add_compile_options(-g)
```
henningkayser, line 3:
```suggestion
```

moveit_ros/moveit_servo/cmake-build-debug/CMakeCache.txt
```diff
@@ -0,0 +1,398 @@
+# This is the CMakeCache file.
```
henningkayser, line 1: Also artifacts that don't belong in here

moveit_ros/moveit_servo/config/pose_tracking_settings.yaml
```diff
@@ -2,6 +2,7 @@
 # PID parameters for pose seeking
 #################################
 
+
```
henningkayser, line 5:
```suggestion
```

moveit_ros/moveit_servo/include/moveit_servo/pose_tracking.h
```diff
@@ -171,12 +174,12 @@ class PoseTracking
 
   // Transforms w.r.t. planning_frame_
   Eigen::Isometry3d command_frame_transform_;
-  ros::Time command_frame_transform_stamp_;
-  geometry_msgs::PoseStamped target_pose_;
+  rclcpp::Time command_frame_transform_stamp_ = rclcpp::Time(0, 0, RCL_ROS_TIME);
```
henningkayser, line 177: does this need a default value?

moveit_ros/moveit_servo/src/cpp_interface_demo/pose_tracking_demo.cpp
```diff
@@ -36,39 +36,43 @@
...
 #include <moveit_servo/servo.h>
 #include <moveit_servo/pose_tracking.h>
 #include <moveit_servo/status_codes.h>
+#include <moveit_servo/servo_parameters.h>
+#include <moveit_servo/servo_parameters.cpp>
 #include <moveit_servo/make_shared_from_pool.h>
 #include <thread>
 
-static const std::string LOGNAME = "cpp_interface_example";
+static const std::string LOGNAME = "pose_tracking_demo";
```
henningkayser, line 50: I think the logname is not needed anymore, `LOGGER` is used instead

moveit_ros/moveit_servo/src/cpp_interface_demo/pose_tracking_demo.cpp
```diff
@@ -36,39 +36,43 @@
...
+static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_servo.pose_tracking_demo");
 
 // Class for monitoring status of moveit_servo
 class StatusMonitor
 {
 public:
-  StatusMonitor(ros::NodeHandle& nh, const std::string& topic)
+  StatusMonitor(rclcpp::Node::SharedPtr node, const std::string& topic)
```
henningkayser, line 57: Better pass by const ref

moveit_ros/moveit_servo/test/launch/test_servo_pose_tracking.test.py
```diff
@@ -0,0 +1,171 @@
...
+    #                  output='log',
+    #                  arguments=['-d', rviz_config_file],
+    #                  parameters=[robot_description, robot_description_semantic, kinematics_yaml])
+
+    # A node to publish world -> panda_link0 transform
+    # static_tf = Node(package='tf2_ros',
+    #                  executable='static_transform_publisher',
+    #                  name='static_transform_publisher',
+    #                  output='log',
+    #                  arguments=['0.0', '0.0', '0.0', '0.0', '0.0', '0.0', 'world', 'panda_link0'])
```
tylerjw, line 85: remove commented out code

moveit_ros/moveit_servo/test/launch/test_servo_pose_tracking.test.py
```diff
@@ -0,0 +1,171 @@
...
+                                              robot_description]
+                                )
+    
+    # pose_tracking_node = Node(
+    #     package='moveit_servo',
+    #     executable='servo_pose_tracking_demo',
+    #     #prefix=['xterm -e gdb -ex run --args'],
+    #     output='screen',
+    #     parameters=[robot_description, robot_description_semantic, kinematics_yaml, pose_tracking_params, servo_params]
+    # )
```
tylerjw, line 102: remove commented out code

moveit_ros/moveit_servo/test/launch/test_servo_pose_tracking.test.py
```diff
@@ -0,0 +1,171 @@
...
+                #     parameters=[pose_tracking_params, servo_params, robot_description, robot_description_semantic, kinematics_yaml])
+                #     extra_arguments=[{'use_intra_process_comm': True}])
+            ],
+            output='screen',
+    )
+
+    # pose_tracking_gtest = launch_testing.actions.GTest(
+    #             path=PathJoinSubstitution([LaunchConfiguration('test_binary_dir'), gtest_name]),
+    #             timeout=40.0, output='screen'
+    #     )
```
tylerjw, line 133: remove commented out code

moveit_ros/moveit_servo/test/launch/test_servo_pose_tracking.test.py
```diff
@@ -0,0 +1,171 @@
...
+
+    pose_tracking_gtest = launch_ros.actions.Node(
+                executable=PathJoinSubstitution([LaunchConfiguration('test_binary_dir'), gtest_name]),
+                parameters=[robot_description, robot_description_semantic, pose_tracking_params, servo_params]
+        )
+
+    #return LaunchDescription([static_tf, fake_joint_driver_node, robot_state_publisher])
```
tylerjw, line 140: here too

moveit_ros/moveit_servo/test/launch/test_servo_pose_tracking.test.py
```diff
@@ -0,0 +1,171 @@
...
+    return launch.LaunchDescription([
+        launch.actions.DeclareLaunchArgument(name='test_binary_dir',
+                                             description='Binary directory of package '
+                                                         'containing test executables'),
+        fake_joint_driver_node,
+        #robot_state_publisher,
+        #static_tf,
+        #rviz_node,
```
tylerjw, line 148: -

moveit_ros/moveit_servo/test/launch/test_servo_pose_tracking.test.py
```diff
@@ -0,0 +1,171 @@
...
+        test_container,
+        pose_tracking_gtest,
+        launch_testing.actions.ReadyToTest()
+    ]), {'test_container': test_container ,'servo_gtest': pose_tracking_gtest, 'fake_joint_driver_node': fake_joint_driver_node }
+    #{'robot_stablte_puisher': robot_state_publisher, 'static_tf': static_tf, 'servo_gtest': pose_tracking_gtest, 'fake_joint_driver_node': fake_joint_driver_node }
```
tylerjw, line 153: -

moveit_ros/moveit_servo/test/pose_tracking_test.cpp
```diff
@@ -62,49 +64,92 @@ class PoseTrackingFixture : public ::testing::Test
 public:
   void SetUp() override
   {
-    // Wait for several key topics / parameters
-    ros::topic::waitForMessage<sensor_msgs::JointState>("/joint_states");
-    while (!nh_.hasParam("/robot_description") && ros::ok())
+    // Wait for several key topics / parameters // nodes to start
+    rclcpp::sleep_for(std::chrono::seconds(5));
```
tylerjw, line 68: Is this sleep needed?

moveit_ros/moveit_servo/test/pose_tracking_test.cpp
```diff
@@ -62,49 +64,92 @@ class PoseTrackingFixture : public ::testing::Test
...
         planning_scene_monitor::PlanningSceneMonitor::DEFAULT_COLLISION_OBJECT_TOPIC,
         planning_scene_monitor::PlanningSceneMonitor::DEFAULT_PLANNING_SCENE_WORLD_TOPIC,
         false /* skip octomap monitor */);
+    planning_scene_monitor_->startStateMonitor(parameters_->joint_topic);
+    planning_scene_monitor_->startPublishingPlanningScene(planning_scene_monitor::PlanningSceneMonitor::UPDATE_SCENE);
 
-    tracker_ = std::make_shared<moveit_servo::PoseTracking>(nh_, planning_scene_monitor_);
-
-    target_pose_pub_ = nh_.advertise<geometry_msgs::PoseStamped>("target_pose", 1 /* queue */, true /* latch */);
+    rclcpp::sleep_for(std::chrono::seconds(2));
```
tylerjw, line 95: Do we need this sleep?  What is it waiting for?

moveit_ros/moveit_servo/test/pose_tracking_test.cpp
```diff
@@ -135,15 +180,14 @@ TEST_F(PoseTrackingFixture, OutgoingMsgTest)
...
     {
-      target_pose_pub_.publish(target_pose);
-      ros::Duration(0.01).sleep();
+      target_pose_pub_->publish(target_pose);
+      std::this_thread::sleep_for(std::chrono::milliseconds(10));;
     }
   });
 
-  ros::Duration(ROS_PUB_SUB_DELAY).sleep();
+  std::this_thread::sleep_for(std::chrono::milliseconds(ROS_PUB_SUB_DELAY*100));
```
tylerjw, line 188:
```suggestion
  std::this_thread::sleep_for(std::chrono::seconds(ROS_PUB_SUB_DELAY));
```

Instead of doing the conversion manually with a multiply, you can just use the seconds type.

moveit_ros/moveit_servo/test/pose_tracking_test.cpp
```diff
@@ -155,12 +199,10 @@ TEST_F(PoseTrackingFixture, OutgoingMsgTest)
 
 int main(int argc, char** argv)
 {
-  ros::init(argc, argv, LOGNAME);
-  testing::InitGoogleTest(&argc, argv);
-
-  ros::AsyncSpinner spinner(8);
-  spinner.start();
+  ::testing::InitGoogleTest(&argc, argv);
+  rclcpp::init(argc, argv);
```
tylerjw, line 203: Should the ros init be called before the google test init?

tylerjw (commented):
I left a handful of comments.  Let me know if any of these don't make sense or don't work as changes so I can help.  Ping me when you are ready for another review of this.

moveit_ros/moveit_servo/include/moveit_servo/pose_tracking.h
```diff
@@ -81,11 +82,12 @@ const std::unordered_map<PoseTrackingStatusCode, std::string> POSE_TRACKING_STAT
  * Class PoseTracking - subscribe to a target pose.
  * Servo toward the target pose.
  */
-class PoseTracking
+class PoseTracking 
 {
 public:
   /** \brief Constructor. Loads ROS parameters under the given namespace. */
-  PoseTracking(const ros::NodeHandle& nh, const planning_scene_monitor::PlanningSceneMonitorPtr& planning_scene_monitor);
+  PoseTracking(const rclcpp::Node::SharedPtr& node, const ServoParametersPtr& parameters,
```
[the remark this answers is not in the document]
Abishalini: Fixed this in pose tracking demo and test as well

moveit_ros/moveit_servo/launch/pose_tracking_example.launch.py
```diff
@@ -0,0 +1,92 @@
+import os
+import yaml
+from launch import LaunchDescription
+from launch_ros.actions import Node
+from ament_index_python.packages import get_package_share_directory
+from launch_ros.actions import ComposableNodeContainer
+from launch_ros.descriptions import ComposableNode
```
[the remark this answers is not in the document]
Abishalini: robot_state_publisher and tf nodes use`ComposableNode` in the test file. I can fix this in the demo when we refactor PoseTracking

moveit_ros/moveit_servo/src/cpp_interface_demo/pose_tracking_demo.cpp
```diff
@@ -78,82 +82,107 @@ class StatusMonitor
...
   target_pose.header.frame_id = current_ee_tf.header.frame_id;
   target_pose.pose.position.x = current_ee_tf.transform.translation.x;
   target_pose.pose.position.y = current_ee_tf.transform.translation.y;
   target_pose.pose.position.z = current_ee_tf.transform.translation.z;
   target_pose.pose.orientation = current_ee_tf.transform.rotation;
 
+
   // Modify it a little bit
-  target_pose.pose.position.x += 0.1;
+  target_pose.pose.position.x += 0.01;
```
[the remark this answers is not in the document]
Abishalini: It works. I just reduced it so the arm doesn't go out of bound and reach singularity.
I've reverted it back to 0.1 now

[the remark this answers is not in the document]
Abishalini: Strangely that caused the seg fault at end that you were getting

moveit_ros/moveit_servo/src/cpp_interface_demo/pose_tracking_demo.cpp
```diff
@@ -78,82 +82,107 @@ class StatusMonitor
...
-  ros::Rate loop_rate(50);
+  rclcpp::Rate loop_rate(50);
   for (size_t i = 0; i < 500; ++i)
   {
     // Modify the pose target a little bit each cycle
     // This is a dynamic pose target
-    target_pose.pose.position.z += 0.0004;
-    target_pose.header.stamp = ros::Time::now();
-    target_pose_pub.publish(target_pose);
+    target_pose.pose.position.z += 0.0001;
```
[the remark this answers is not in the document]
Abishalini: That makes my arm overshoot and so there is an emergency stop (since it gets very close to a singularity).
I thought the position is in terms of meters but I'm not sure looking at how far the arm reaches out with just 0.001 change in the z direction on RViz.

moveit_ros/moveit_servo/src/cpp_interface_demo/pose_tracking_demo.cpp
```diff
@@ -78,82 +82,107 @@ class StatusMonitor
...
+    target_pose.header.stamp = node->now();
+    target_pose_pub->publish(target_pose);
 
     loop_rate.sleep();
   }
 
+  rclcpp::sleep_for(std::chrono::seconds(10));
+  RCLCPP_DEBUG_STREAM(LOGGER, "Timer expired and unfinished motion will be terminated");
```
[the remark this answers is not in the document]
Abishalini: @nbbrooks added the sleep to make sure the arm reaches the final position before the stopMotion() function gets executed.

moveit_ros/moveit_servo/test/pose_tracking_test.cpp
Abishalini: I initially put it to set up the planning scene monitor but it's not necessary. I'll remove it

moveit_ros/moveit_servo/test/pose_tracking_test.cpp
Abishalini: It's not. I'll removed it

moveit_ros/moveit_servo/test/pose_tracking_test.cpp
Abishalini: I don't think so (ROS1 example - [http://wiki.ros.org/gtest])

moveit_ros/moveit_servo/include/moveit_servo/make_shared_from_pool.h
```diff
@@ -38,7 +38,8 @@
 
 #pragma once
 
-#include <boost/make_shared.hpp>
+//#include <boost/make_shared.hpp>
```
tylerjw, line 41: delete this

moveit_ros/moveit_servo/include/moveit_servo/pose_tracking.h
```diff
@@ -81,11 +82,12 @@ const std::unordered_map<PoseTrackingStatusCode, std::string> POSE_TRACKING_STAT
  * Class PoseTracking - subscribe to a target pose.
  * Servo toward the target pose.
  */
-class PoseTracking
+class PoseTracking
```
tylerjw, line 85: Remove space at the end of this line.

https://stackoverflow.com/questions/30884131/remove-trailing-spaces-automatically-or-with-a-shortcut

moveit_ros/moveit_servo/src/cpp_interface_demo/pose_tracking_demo.cpp
```diff
@@ -78,46 +82,59 @@ class StatusMonitor
...
   planning_scene_monitor->startSceneMonitor();
   planning_scene_monitor->startWorldGeometryMonitor(
       planning_scene_monitor::PlanningSceneMonitor::DEFAULT_COLLISION_OBJECT_TOPIC,
       planning_scene_monitor::PlanningSceneMonitor::DEFAULT_PLANNING_SCENE_WORLD_TOPIC,
       false /* skip octomap monitor */);
-  planning_scene_monitor->startStateMonitor();
+  planning_scene_monitor->startStateMonitor(servo_parameters->joint_topic);
+  planning_scene_monitor->startPublishingPlanningScene(planning_scene_monitor::PlanningSceneMonitor::UPDATE_SCENE);
+
+  rclcpp::sleep_for(std::chrono::seconds(5));
```
tylerjw, line 118: Add comment to figure out later how to wait for the planning scene monitor to be ready.

moveit_ros/moveit_servo/test/pose_tracking_test.cpp
```diff
@@ -41,19 +41,21 @@
...
 #include <moveit_servo/make_shared_from_pool.h>
+#include <moveit_servo/servo_parameters.cpp>
+
+static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_servo.pose_tracking_test");
 
-static const std::string LOGNAME = "servo_cpp_interface_test";
 static constexpr double TRANSLATION_TOLERANCE = 0.01;  // meters
 static constexpr double ROTATION_TOLERANCE = 0.1;      // quaternion
-static constexpr double ROS_PUB_SUB_DELAY = 4;         // allow for subscribers to initialize
+static constexpr u_int64_t ROS_PUB_SUB_DELAY = 4;      // allow for subscribers to initialize
```
tylerjw, line 58: Add the unit to the comment.

tylerjw (approved):

tylerjw:
You need to rebase it again on the `main` branch.

moveit_ros/moveit_servo/src/pose_tracking.cpp
```diff
@@ -235,7 +233,7 @@ void PoseTracking::targetPoseCallback(const geometry_msgs::msg::PoseStamped::Con
     try
     {
       geometry_msgs::msg::TransformStamped target_to_planning_frame = transform_buffer_.lookupTransform(
-          planning_frame_, target_pose_.header.frame_id, rclcpp::Time(0), rclcpp::Duration(0.1));
```
tylerjw: Are you sure you want no timeout on this?  I'm not sure what that does but I imagine that causes it to block forever or return imidiatly.  I would assume what you want is something of a timeout (as there was before).

tylerjw:
The more I think about this I think more it'd be good to leave this in the state it is and not merge it yet until we have time to fix the segfault, review and document the ros interfaces, and design a tutorial for using this.

I'd like this to be as ready for use as servo is now when we merge it.  Thank you for getting us this far @Abishalini.  I know we don't have time to do the things I am requesting now.  If someone wants to use this they can through this PR and when we have time to work on it again we can get it to the same level of readiness as servo is in moveit2.

davetcoleman:
We're going to need this feature for Studio soon, fyi

tylerjw:
After much internal discussion there are really only two things I am willing to block merging this over:

1. The segfault in the demo.
2. Sleeps in initialization/testing.  This will make these tests and initialization flaky.  Instead, we need to block (wait) on the transitioning of state in other nodes.  To do this we can either publish to some state topic that lathes or we can wait for indicators of state transitions (see how the tests in moveit_servo wait on the start service to be available to know when the servo node has finished initializing).

I will create issues for the other things I'd like done with pose tracker before we publically announce it.  Specifically:

1. Parameters refactored to use the standard @henningkayser is developing.
2. Component node interface
3. Pose Tracker moved into its own ros package/directory
4. Complete code coverage with unit test that tests the functionality of pose tracker by itself (without servo)
5. Passes all standard ROS2 linters
6. A tutorial with an example application similar to servo
7. Standardized documentation

henningkayser:
@Abishalini @tylerjw please review https://github.com/Abishalini/moveit2/pull/1

moveit_ros/moveit_servo/src/cpp_interface_demo/pose_tracking_demo.cpp
```diff
@@ -78,46 +82,60 @@ class StatusMonitor
...
   planning_scene_monitor->startWorldGeometryMonitor(
       planning_scene_monitor::PlanningSceneMonitor::DEFAULT_COLLISION_OBJECT_TOPIC,
       planning_scene_monitor::PlanningSceneMonitor::DEFAULT_PLANNING_SCENE_WORLD_TOPIC,
       false /* skip octomap monitor */);
-  planning_scene_monitor->startStateMonitor();
+  planning_scene_monitor->startStateMonitor(servo_parameters->joint_topic);
+  planning_scene_monitor->startPublishingPlanningScene(planning_scene_monitor::PlanningSceneMonitor::UPDATE_SCENE);
+
+  // Wait for Planning Scene Monitor to setup
+  rclcpp::sleep_for(std::chrono::seconds(5));
```
tylerjw, line 119: @henningkayser here is a sleep for waiting for the planning scene we should figure out how to remove..

moveit_ros/moveit_servo/src/cpp_interface_demo/pose_tracking_demo.cpp
```diff
@@ -132,28 +150,36 @@ int main(int argc, char** argv)
...
     target_pose.pose.position.z += 0.0004;
-    target_pose.header.stamp = ros::Time::now();
-    target_pose_pub.publish(target_pose);
+    target_pose.header.stamp = node->now();
+    target_pose_pub->publish(target_pose);
 
     loop_rate.sleep();
   }
 
+  rclcpp::sleep_for(std::chrono::seconds(5));
```
tylerjw, line 176: Here is another one we should remove.

moveit_ros/moveit_servo/src/pose_tracking.cpp
```diff
@@ -225,35 +223,35 @@ bool PoseTracking::satisfiesPoseTolerance(const Eigen::Vector3d& positional_tole
...
-
   // If the target pose is not defined in planning frame, transform the target pose.
   if (target_pose_.header.frame_id != planning_frame_)
   {
     try
     {
-      geometry_msgs::TransformStamped target_to_planning_frame = transform_buffer_.lookupTransform(
-          planning_frame_, target_pose_.header.frame_id, ros::Time(0), ros::Duration(0.1));
+      geometry_msgs::msg::TransformStamped target_to_planning_frame = transform_buffer_.lookupTransform(
+          planning_frame_, target_pose_.header.frame_id, rclcpp::Time(0), rclcpp::Duration(0));
```
tylerjw, line 236: Here is the documentation for `rclcpp::Duration`.  This should be some value other than 0.  It is no longer constructed with a double but is either (seconds, nanoseconds) or (nanoseconds).

moveit_ros/moveit_servo/test/pose_tracking_test.cpp
```diff
@@ -135,15 +176,15 @@ TEST_F(PoseTrackingFixture, OutgoingMsgTest)
...
-      target_pose_pub_.publish(target_pose);
-      ros::Duration(0.01).sleep();
+      target_pose_pub_->publish(target_pose);
+      std::this_thread::sleep_for(std::chrono::milliseconds(10));
+      ;
     }
   });
 
-  ros::Duration(ROS_PUB_SUB_DELAY).sleep();
+  std::this_thread::sleep_for(std::chrono::seconds(ROS_PUB_SUB_DELAY));
```
tylerjw, line 185: This sleep should be removed.

moveit_ros/moveit_servo/test/pose_tracking_test.cpp
```diff
@@ -135,15 +176,15 @@ TEST_F(PoseTrackingFixture, OutgoingMsgTest)
     size_t msg_count = 0;
     while (++msg_count < 100)
     {
-      target_pose_pub_.publish(target_pose);
-      ros::Duration(0.01).sleep();
+      target_pose_pub_->publish(target_pose);
+      std::this_thread::sleep_for(std::chrono::milliseconds(10));
```
tylerjw, line 180: This should be done with a rate instead of a sleep.

moveit_ros/moveit_servo/src/cpp_interface_demo/pose_tracking_demo.cpp
henningkayser: this is fixed in the PR linked above

moveit_ros/moveit_servo/src/pose_tracking.cpp
henningkayser: I'd change this to `0.1s` using chrono literals

moveit_ros/moveit_servo/test/pose_tracking_test.cpp
henningkayser: I agree!

moveit_ros/moveit_servo/test/pose_tracking_test.cpp
henningkayser: Poll for subscription count? https://docs.ros2.org/foxy/api/rclcpp/classrclcpp_1_1PublisherBase.html#a60c84260b2ff0fb364fad688ed91234f

tylerjw: Maybe that would work here.  In the ros1 tests for servo we count messages produced by servo on its output for this.  This is the one I'm not sure what the right approach is (as I'm not sure why this sleep is here at all).  This sleep is after the loop that publishes command messages.

moveit_ros/moveit_servo/src/pose_tracking.cpp
tylerjw: or `100ms`

tylerjw:
I pushed a rebase of this on main and with @henningkayser 's changes to fix the segfault.

We still need to resolve a couple of issues I commented on above.

moveit_ros/moveit_servo/src/pose_tracking.cpp
Abishalini: Changed it to 100ms

MarqRazz:
Nit picking but do we always miss our control rate when we first startup?
`[moveit_servo.pose_tracking]: Target control rate was missed`

Also after the target pose is reached why do we get a warning about a `Stale command`?
<pre>
[moveit_servo.pose_tracking]: The target pose is achieved!
[moveit_servo.pose_tracking_demo]: Pose tracker exited with status: Success
[moveit_ros.planning_scene_monitor.planning_scene_monitor]: The complete state of the robot is not yet known.  Missing virtual_joint
<b>[moveit_servo.servo_calcs]: Stale command. Try a larger 'incoming_command_timeout' parameter?</b>
[moveit_ros.planning_scene_monitor.planning_scene_monitor]: Stopped publishing maintained planning scene.
[moveit_ros.planning_scene_monitor.planning_scene_monitor]: Stopping world geometry monitor
[moveit_ros.planning_scene_monitor.planning_scene_monitor]: Stopping planning scene monitor
</pre>

moveit_ros/moveit_servo/test/pose_tracking_test.cpp
henningkayser: Looking at the test logic I don't think we should have a delay here at all. The subscription callback should be synchronized or timed out using a future (preferred imo) or condition variable.

Abishalini: I removed the sleep because otherwise the difference between header timestamp and current time was larger than the timeout which caused the program to think the target pose was not updating

tylerjw:
fixed clang_format issue and rebased on main, will merge if this passes.

tylerjw:
> ~Nit picking but do we always miss our control rate when we first startup?~
> `[moveit_servo.pose_tracking]: Target control rate was missed`
> 
> ~Also after the target pose is reached why do we get a warning about a `Stale command`?~
> 
> [moveit_servo.pose_tracking]: The target pose is achieved!
> [moveit_servo.pose_tracking_demo]: Pose tracker exited with status: Success
> [moveit_ros.planning_scene_monitor.planning_scene_monitor]: The complete state of the robot is not yet known.  Missing virtual_joint
> [moveit_servo.servo_calcs]: Stale command. Try a larger 'incoming_command_timeout' parameter?
> [moveit_ros.planning_scene_monitor.planning_scene_monitor]: Stopped publishing maintained planning scene.
> [moveit_ros.planning_scene_monitor.planning_scene_monitor]: Stopping world geometry monitor
> [moveit_ros.planning_scene_monitor.planning_scene_monitor]: Stopping planning scene monitor
> 
> Sorry I believe that both of these issues were due to the sleeping in the demo program. After updating my branch they are gone.

tylerjw:
The pose tracking test seems to be failing by timing out on travis.  Maybe we need to extend the test timeout for travis.  It doesn't look like it got stuck, just took a full minute during init.  Yet one more reason we should leave travis if there is something better.

schornakj:
Can we retry CI for this PR? I was going to experiment with increasing the test timeout [in this PR](https://github.com/ros-planning/moveit2/pull/356/checks) but the previously-failing test succeeded and took a similar amount of time as when I ran it locally on my PC (about 20 seconds). I'm wondering if this is an issue with some aspect of the CI runner being flaky, rather than something actually being wrong with the code or tests.

AndyZe:
i restarted it. Ping me on slack if I need to do that again

AndyZe:
It looks like the most important set of tests is passing, though:  `1490.2 [Foxy] - gcc (build,test)`

schornakj:
Looks like the job that was failing previously passed in the most recent run: `1490.3 [Foxy] - clang (build,test,clang-tidy-fix,test_ikfast_plugins)`.

I tried to reproduce the failure in a variety of other contexts (building from source, running moveit_ci locally, sending in additional PRs against this repo and other repos, building both with and without caching) and in all those cases it successfully built and tested everything. In my mind that means there's potentially something up with the specific context of the jobs in this PR, such as the time when they were run or the state of the cache.

In my opinion this is goes into the "keep an eye out for it going forward" category.

AndyZe (approved):
I took a quick look. Seems good to me! I think it's been nitpicked enough.

PR #323, opened 2021-01-05
MTC compatiblity fixes

JafarAbdi:
### Description

These are some fixes for bugs I encountered while [porting MTC](https://github.com/ros-planning/moveit_task_constructor/pull/170) ~+ backported the first two commits~

### Checklist


moveit_core/utils/include/moveit/utils/robot_model_test_utils.h
```diff
@@ -164,6 +166,9 @@ class RobotModelBuilder
    */
   void addGroup(const std::vector<std::string>& links, const std::vector<std::string>& joints, const std::string& name);
 
+  void addEndEffector(const std::string& name, const std::string& parent_link, const std::string& parent_group = "",
+                      const std::string& component_group = "");
+
```
tylerjw, line 171: should this hav a doc string?

moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
```diff
@@ -148,10 +145,17 @@ class ActionBasedControllerHandle : public ActionBasedControllerHandleBase
...
-      std::future_status status = result_future.wait_for(timeout.to_chrono<std::chrono::seconds>());
+      std::future_status status = result_future.wait_for(timeout.to_chrono<std::chrono::duration<double>>());
       if (status == std::future_status::timeout)
+      {
         RCLCPP_WARN(LOGGER, "waitForExecution timed out");
+        return false;
+      }
     }
+    // workaround for controllerDoneCallback being called a little bit late
+    rclcpp::Time deadline = node_->now() + rclcpp::Duration::from_seconds(0.1);  // limit waiting to 0.1s
```
tylerjw, line 156: Should this be a constant (same with the 100ns below)?

tylerjw (approved):
Just a couple small questions.  Overall looks good.

henningkayser:
This PR includes cherry-picked commits from Moveit 1 master, right? I think we should only apply your added fixes and run another sync to add the remaining changes.

moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
henningkayser: Please don't add constant delays. Better use a way to explicitly wait for the done callback (i.e. condition variable, promise).

moveit_ros/planning_interface/planning_scene_interface/src/planning_scene_interface.cpp
```diff
@@ -49,7 +49,8 @@ static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_ros.planning_sce
 class PlanningSceneInterface::PlanningSceneInterfaceImpl
 {
 public:
-  explicit PlanningSceneInterfaceImpl(const std::string& ns = "", bool wait = true) : node_(new rclcpp::Node(ns))
+  explicit PlanningSceneInterfaceImpl(const std::string& ns = "", bool wait = true)
+    : node_(new rclcpp::Node("planning_scene_interface", ns))
```
henningkayser, line 53: Does this work the way it should? I would think that this will cause the same duplicate node issues as with the PSM.

moveit_ros/visualization/CMakeLists.txt
```diff
@@ -79,6 +79,16 @@ pluginlib_export_plugin_description_file(rviz_common robot_state_rviz_plugin_des
 #  add_rostest(test/moveit_joy.test)
 #endif()
 
+ament_export_libraries(moveit_motion_planning_rviz_plugin_core)
+ament_export_libraries(moveit_motion_planning_rviz_plugin)
```
henningkayser, line 83: I don't think the plugins should be exported (what's the purpose), the core libs should be sufficient.

moveit_ros/visualization/CMakeLists.txt
```diff
@@ -79,6 +79,16 @@ pluginlib_export_plugin_description_file(rviz_common robot_state_rviz_plugin_des
...
+ament_export_libraries(moveit_planning_scene_rviz_plugin_core)
+ament_export_libraries(moveit_planning_scene_rviz_plugin)
+ament_export_libraries(moveit_robot_state_rviz_plugin)
+ament_export_libraries(moveit_robot_state_rviz_plugin_core)
+ament_export_libraries(moveit_rviz_plugin_render_tools)
+ament_export_libraries(moveit_trajectory_rviz_plugin_core)
+ament_export_libraries(moveit_trajectory_rviz_plugin)
+ament_export_include_directories(include)
```
henningkayser, line 91: Shouldn't this list the actual include directories of the plugins?

moveit_ros/visualization/CMakeLists.txt
JafarAbdi: You're right, I removed them

moveit_ros/visualization/CMakeLists.txt
JafarAbdi: No, this will export the main installed `include` directory

moveit_ros/planning_interface/planning_scene_interface/src/planning_scene_interface.cpp
```diff
@@ -49,7 +49,8 @@ static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_ros.planning_sce
 class PlanningSceneInterface::PlanningSceneInterfaceImpl
 {
 public:
-  explicit PlanningSceneInterfaceImpl(const std::string& ns = "", bool wait = true) : node_(new rclcpp::Node(ns))
+  explicit PlanningSceneInterfaceImpl(const std::string& ns = "", bool wait = true)
+    : node_(new rclcpp::Node("planning_scene_interface_" + std::to_string(reinterpret_cast<std::size_t>(this)), ns))
```
henningkayser, line 53: What's with this node name? I still think that this will clash with the application node if there is one. Shouldn't we rather pass through a node with default initialization or create a "private" node as for the PlanningSceneMonitor? There we use ros args to set the node name since it's not applied otherwise. Also, there doesn't seem to be an executor anywhere, but that's probably not required as we are spinning the service calls explicitly.

JafarAbdi: Super sorry, I forgot about the node name bug which needs an explicit remapping

> Also, there doesn't seem to be an executor anywhere, but that's probably not required as we are spinning the service calls explicitly.

We don't need an executor here, we do spin the node only after sending a request or publishing a message

henningkayser (approved):

PR #326, opened 2021-01-10
Refine local planner component

sjahr:
### Description

Refines the local planner component and makes it generic. Therefore, the behavior of the local planner component will be defined by two plugins. One to identify the local problem each loop run and to manage the reference trajectory (trajectory matching and blending) and a second plugin to implement the local constraint solver.

### Checklist

### DIFF
--- a/moveit_ros/hybrid_planning/constraint_solver_plugins/src/handle_imminent_collision.cpp
+++ b/moveit_ros/hybrid_planning/constraint_solver_plugins/src/handle_imminent_collision.cpp
@@ -70,10 +70,6 @@ HandleImminentCollision::solve(robot_trajectory::RobotTrajectory local_trajector
                                std::vector<moveit_msgs::msg::Constraints> local_constraints,
                                std::shared_ptr<moveit_msgs::action::LocalPlanner::Feedback> feedback)
 {
-  // Transform local_trajectory into msg data structure to use isPathValid()
-  moveit_msgs::msg::RobotTrajectory local_trajectory_msg;
-  local_trajectory.getRobotTrajectoryMsg(local_trajectory_msg);
-
   // Clone current planning scene
   planning_scene_monitor_->updateFrameTransforms();
   planning_scene_monitor_->lockSceneRead();  // LOCK planning scene
@@ -81,14 +77,11 @@ HandleImminentCollision::solve(robot_trajectory::RobotTrajectory local_trajector
       planning_scene_monitor_->getPlanningScene());  // TODO remove expensive planning scene cloning
   planning_scene_monitor_->unlockSceneRead();        // UNLOCK planning scene
 
-  // Get Current State
-  moveit_msgs::msg::RobotState current_state_msg;
-  robotStateToRobotStateMsg(planning_scene->getCurrentState(), current_state_msg);
-
   robot_trajectory::RobotTrajectory robot_command(local_trajectory.getRobotModel(), local_trajectory.getGroupName());
+  std::vector<std::size_t>* invalid_index;
 
   // Check if path is valid
-  if (planning_scene->isPathValid(current_state_msg, local_trajectory_msg))
+  if (planning_scene->isPathValid(local_trajectory, local_trajectory.getGroupName(), false, invalid_index))
   {
     if (feedback_send_)
     {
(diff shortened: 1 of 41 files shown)


moveit_ros/hybrid_planning/local_planner/include/moveit/local_planner/local_planner_component.h
```diff
@@ -46,43 +48,134 @@
...
+  {
+    RCLCPP_ERROR_STREAM(node->get_logger(),
+                        "Error getting parameter \'" << param_name << "\', check parameter type in YAML file");
+    throw e;
+  }
+}
+
+/// Internal local planner states
+/// TODO Use lifecycle node?
 enum class LocalPlannerState : int8_t
```
henningkayser, line 83: Are these generic enough to go into the message type?

moveit_ros/hybrid_planning/local_planner/src/constraint_solver_interface.cpp
```diff
@@ -0,0 +1,40 @@
...
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+#include <moveit/local_planner/constraint_solver_interface.h>
+
+namespace moveit_hybrid_planning
```
henningkayser, line 37: Are these even needed?

moveit_ros/hybrid_planning/local_planner/include/moveit/local_planner/local_planner_component.h
sjahr: Probably, but I think this should be replaced by a lifecycle component node with integrated states in the long run. Do you think we should keep it this way in the mean time?

moveit_ros/hybrid_planning/constraint_solver_plugins/src/handle_imminent_collision.cpp
```diff
@@ -70,16 +70,18 @@ HandleImminentCollision::solve(robot_trajectory::RobotTrajectory local_trajector
...
                                std::shared_ptr<moveit_msgs::action::LocalPlanner::Feedback> feedback)
 {
-  // Clone current planning scene
+  // Read current planning scene
   planning_scene_monitor_->updateFrameTransforms();
   planning_scene_monitor_->lockSceneRead();  // LOCK planning scene
-  planning_scene::PlanningScenePtr planning_scene = planning_scene::PlanningScene::clone(
-      planning_scene_monitor_->getPlanningScene());  // TODO remove expensive planning scene cloning
-  planning_scene_monitor_->unlockSceneRead();        // UNLOCK planning scene
+  planning_scene::PlanningScenePtr planning_scene = planning_scene_monitor_->getPlanningScene();
```
henningkayser, line 75: This won't do what you want (the planning scene instance may change after unlocking so your collision check won't be protected). The recommended way to do this would be to initialize a `LockedPlanningSceneRO` which gives you read-locked access to the planning scene until it goes out of scope. http://docs.ros.org/en/melodic/api/moveit_ros_planning/html/classplanning__scene__monitor_1_1LockedPlanningSceneRO.html

moveit_ros/hybrid_planning/local_planner/include/moveit/local_planner/local_planner_component.h
henningkayser: No need to. Using this for node lifecycle management at some point makes sense to me.

moveit_ros/hybrid_planning/local_planner/include/moveit/local_planner/constraint_solver_interface.h
```diff
@@ -0,0 +1,87 @@
...
+#include <moveit_msgs/action/local_planner.hpp>
+
+#include <trajectory_msgs/msg/joint_trajectory.h>
+
+namespace moveit_hybrid_planning
+{
+/**
+ * Class ConstraintSolverInterface - Base class for a local constrain solver.
+ */
+class ConstraintSolverInterface
```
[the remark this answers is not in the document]
sjahr: Good idea, do you think I should just rename the class but keep the file names to avoid them from getting too long?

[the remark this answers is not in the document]
henningkayser: I would keep them the same. It's ok to have long file names imo.

henningkayser:
please fix CI, then lgtm

henningkayser (approved):

PR #340, opened 2021-01-22
Enable ament_lint tests

tylerjw:
Signed-off-by: Tyler Weaver <tylerjw@gmail.com>

### Description

This enables the default set of linters in [ament_lint_common](https://github.com/ament/ament_lint/blob/master/ament_lint_common/doc/index.rst).  It selectively disables all the tests that do not currently pass.  This should make it easy for people to fix the issues found by these linters by removing the one line and running the test to see what they need to fix.  This PR enables over 200 lint tests that already pass.

These could create excellent first-timer-only issues or moveit wold day issues.

### DIFF
--- a/moveit_common/CMakeLists.txt
+++ b/moveit_common/CMakeLists.txt
@@ -3,6 +3,16 @@ project(moveit_common NONE)
 
 find_package(ament_cmake REQUIRED)
 
+if(BUILD_TESTING)
+  find_package(ament_lint_auto REQUIRED)
+
+  # These don't pass yet, disable them for now
+  set(ament_cmake_copyright_FOUND TRUE)
+
+  # Run all lint tests in package.xml except those listed above
+  ament_lint_auto_find_test_dependencies()
+endif()
+
 ament_package(
   CONFIG_EXTRAS "moveit_common-extras.cmake"
 )
--- a/moveit/CMakeLists.txt
+++ b/moveit/CMakeLists.txt
@@ -1,4 +1,17 @@
 cmake_minimum_required(VERSION 3.10.2)
 project(moveit)
 find_package(ament_cmake REQUIRED)
+
+if(BUILD_TESTING)
+  find_package(ament_lint_auto REQUIRED)
+
+  # These don't pass yet, disable them for now
+  set(ament_cmake_copyright_FOUND TRUE)
+  set(ament_cmake_flake8_FOUND TRUE)
+  set(ament_cmake_pep257_FOUND TRUE)
+
+  # Run all lint tests in package.xml except those listed above
+  ament_lint_auto_find_test_dependencies()
+endif()
+
 ament_package()
--- a/moveit/package.xml
+++ b/moveit/package.xml
@@ -33,6 +33,9 @@
   <exec_depend>moveit_setup_assistant</exec_depend>
   -->
 
+  <test_depend>ament_lint_auto</test_depend>
+  <test_depend>ament_lint_common</test_depend>
+
   <export>
     <build_type>ament_cmake</build_type>
   </export>
--- a/moveit_common/package.xml
+++ b/moveit_common/package.xml
@@ -12,6 +12,9 @@
 
   <buildtool_export_depend>ament_cmake</buildtool_export_depend>
 
+  <test_depend>ament_lint_auto</test_depend>
+  <test_depend>ament_lint_common</test_depend>
+
   <export>
     <build_type>ament_cmake</build_type>
   </export>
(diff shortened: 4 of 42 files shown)


bostoncleek (approved):

moveit_common/CMakeLists.txt
```diff
@@ -3,6 +3,16 @@ project(moveit_common NONE)
 
 find_package(ament_cmake REQUIRED)
 
+if(BUILD_TESTING)
```
henningkayser, line 6: Can't we simply set and enable all tests in `moveit_common` only, possibly providing an extra macro? That way we don't have linters enabled all over the place and we can enforce what checks need to pass in all MoveIt packages. Also, what do you think about using a different build flag than `BUILD_TESTING` so that we can separate the CI stages?

tylerjw: Using the `colcon mixin` system when linters are enabled this way there is an explicit way to run only linters or only non-linters.

Here is the documentation for how to enable colcon mixin: https://github.com/colcon/colcon-mixin-repository

When running tests with colon you can then only linters:
```bash
colcon test --mixin linters-only
```

Skip running linters:
```bash
colcon test --mixin linters-skip
```

tylerjw: > Can't we simply set and enable all tests in `moveit_common` only

I tried this and I couldn't figure out how to get them to propagate.  I really wish there was a way to specify linters per-repo instead of per-ros-package.  I asked about this here: https://github.com/colcon/colcon-core/issues/421 to see if I could instead find some other explicit per-repo way to specify them.

The other thing is that even if we figured out how to enable them for all packages from moveit_common, until all packages pass these linters we need a way to exclude them per-package.

tylerjw:
One thing you'll notice is that I am not enabling clang-tidy or clang-format this way.  Eventually, I'd like to use the ament system for those linters as well.  However, there are existing issues with the ament scripts for how we would like to use those linters.  I have gotten some PRs merged to resolve some of those issues but they are not 100% there.  Specifically, these are the issues that remain:

## clang-tidy
1. Execution time is really long to run this linter so in our CI system we have a workaround where we run it only on packages that have changed in the given PR.  I have a colcon/ament solution to this here: https://github.com/tylerjw/moveit_ci_tools  While the --packages-select arg PR got merged there is another issue.
2. clang-tidy-10 is not supported by ament_clang_tidy.  The issue is the regex for building the xunit files only works with clang-tidy-6.  I am locally just modifying ament_clang_tidy to use clang-tidy-10 for my local testing but before we can use this method we would that regex fixed.  I just haven't had time to work on that yet.
3. The config file.  We currently have a different .clang-tidy config from what is used by ROS2 itself.  I think eventually it'd be really nice to standardize moveit2 to use the same clang-tidy config as ROS2.

## clang-format
1. There was an issue with using clang-format-10 and I got it fixed in a PR.  One thing to note is that until ament_lint is released if we wanted to use this we'd need to depend on the source of ament_lint. 
2. The config file, ROS2 (and therefore ament_clang_format) uses a different config file (partially because they are using clang-6 vs 10).  There is a solution to use our config file, it would just require me to write some CMake code for moveit_common and I haven't gotten around to doing that yet.

Long term I'd like the CI system to lint packages changed by a PR and run tests on those packages, and the dependencies of those packages.  This could greatly speed up testing in CI.  The testing part could impact how code coverage reporting works and I haven't thought of a fix for that yet.

Lastly, it would be really nice if there was an easy way for people to locally run tests and linters on packages they have changed.  The colcon_list_packages_changed script I created and linked to is my current thoughts on that.  Eventually, I'd like to contribute this to be a first-class feature of colcon.

henningkayser (approved):
This is awesome! Somehow, I ran into an issue where `moveit_ros_planning_interface` fails to find `MONGODB` when compiling. Not sure how this could possibly be related to this PR, though.

JafarAbdi (approved):
I fixed a conflict, will merge this once CI pass

tylerjw:
There is something wrong with the dependencies of the latest released ament_lint_auto:

```
--- stderr: moveit_ros_planning_interface                                   
CMake Error at /home/ubuntu/code/ws_moveit2/install/warehouse_ros_mongo/share/warehouse_ros_mongo/cmake/ament_cmake_export_dependencies-extras.cmake:21 (find_package):
  By not providing "FindMONGODB.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "MONGODB", but
  CMake did not find one.

  Could not find a package configuration file provided by "MONGODB" with any
  of the following names:

    MONGODBConfig.cmake
    mongodb-config.cmake

  Add the installation prefix of "MONGODB" to CMAKE_PREFIX_PATH or set
  "MONGODB_DIR" to a directory containing one of the above files.  If
  "MONGODB" provides a separate development package or SDK, be sure it has
  been installed.
  /home/ubuntu/code/ws_moveit2/install/warehouse_ros_mongo/share/warehouse_ros_mongo/cmake/warehouse_ros_mongoConfig.cmake:41 (include)
  /opt/ros/foxy/share/ament_lint_auto/cmake/ament_lint_auto_find_test_dependencies.cmake:36 (find_package)
  CMakeLists.txt:125 (ament_lint_auto_find_test_dependencies)
```

Issue #306, opened 2020-11-10
servo_parameters.cpp shouldn't be in the include folder


AndyZe:
See https://github.com/ros-planning/moveit2/tree/main/moveit_ros/moveit_servo/include/moveit_servo

servo_parameters.cpp should be moved to the src folder

JafarAbdi:
Fixed with #341

PR #341, opened 2021-01-22
Refactor moveit_servo launch testing

tylerjw:
### Description

This change is taking into account what we learned here about launching gtests with parameters:  https://github.com/ros2/launch/issues/466

This removes the workaround that @AdamPettinger found and makes the tests depend directly on the parameter loading.  This is useful because there are no longer two copies of the config, therefore the test that compared the struct version with the loaded yaml version is no longer necessary.

I also broke the parameter loading into a separate library so it could be included by tests that needed the parameters but otherwise did not have a reason to link in the servo library itself.

Next, I made the parameters a shared_ptr to a constant struct.  The advantage of this is the params are no longer mutable during runtime.  This reduces the number of mutable variables and therefore the number of things the reader has to keep track of when trying to grok the source.  This especially useful for reasoning about immutability when debugging threading issues.  One side-effect of this is the unit test for servo_cals that does modify the parameter struct during runtime has to now use const_cast.  This is a reasonable thing to do in a test imo.

Lastly, because the struct is gone the unit test for servo_calcs has to be launched with a launch file because it needs the parameters to be loaded for it.

I hope that this can be a good example of how to write gtests (and their launch files) for both integration and unit tests, and how to load yaml parameters into those tests.  My primary motivation in doing this test was to learn about how the launch system works for gtests in ros2.

### Checklist


moveit_ros/moveit_servo/test/servo_launch_test_common.hpp
```diff
@@ -75,26 +74,38 @@ class ServoFixture : public ::testing::Test
...
-    , TIMEOUT_ITERATIONS(10 * PUBLISH_HZ)
-    , publish_loop_rate_(PUBLISH_HZ)
-    , servo_server_node_name_("/servo_server")
   {
+    // read parameters and store them in shared pointer to constant
+    {
+      auto servo_parameters = std::make_shared<moveit_servo::ServoParameters>();
+      servo_parameters_loaded_success_ = moveit_servo::readParameters(servo_parameters, node_, LOGGER);
+      servo_parameters_ = servo_parameters;
+    }
```
tylerjw, line 91: @henningkayser here is a way to load parameters into const storage.  This uses the shared_ptr to a const as a class variable: `servo_parameters_` and a local non-const one.  The parameters are loaded into the temporary non-const parameters and then the shared pointer is given the ownership and the local non-const goes out of scope.

moveit_ros/moveit_servo/test/unit_test_servo_calcs.cpp
```diff
@@ -526,8 +526,19 @@ int main(int argc, char** argv)
...
+  {
+    auto servo_parameters = std::make_shared<moveit_servo::ServoParameters>();
+    if (!moveit_servo::readParameters(servo_parameters, TEST_NODE, LOGGER))
+    {
+      RCLCPP_FATAL(LOGGER, "Failed to load the servo parameters");
+      rclcpp::shutdown();
+      return 1;
+    }
+    TEST_PARAMS = servo_parameters;
+  }
```
tylerjw, line 539: Here I am also doing the parameter initialization, just this time it is in a main, and when it fails I can log and terminate.  In the case above where it is in a constructor, the workaround is I have to store a variable for the return on readParmeters and later fail if that variable is false.  Generally, I think we should keep reading parameters out of constructors except in the case of test fixtures.

tylerjw:
The reason this is failing is an initialization timing issue.  The integration test launches servo the same way the demo does with a component container and then the gtest sits around waiting for the start service to be registered.  If you read the logs for some reason the servo component never gets loaded.  I think we should find some way in the launch file to launch the gtest node only after the servo component gets loaded.  I posted a question to ros answers here about this: https://answers.ros.org/question/370324/launch-node-when-component-node-is-loaded/

The issue with this PR exists outside of this PR but I'm ok waiting on this until we figure out the answer to how to launch the test when a component node loads.

moveit_ros/moveit_servo/src/collision_check.cpp
```diff
@@ -53,27 +53,27 @@ namespace moveit_servo
 CollisionCheck::CollisionCheck(rclcpp::Node::SharedPtr node, const ServoParametersPtr& parameters,
                                const planning_scene_monitor::PlanningSceneMonitorPtr& planning_scene_monitor)
   : node_(node)
-  , parameters_(parameters)
+  , servo_parameters_(parameters)
```
henningkayser, line 56: Not sure if this renaming is really worth it, it touches a lot of lines and will definitely make syncing difficult in the future.

moveit_ros/moveit_servo/src/servo_parameters.cpp
```diff
@@ -87,7 +87,7 @@ void declareOrGetParam(T& output_value, const std::string& param_name, const rcl
  * @return true if all parameters were loaded and verified successfully, false otherwise
  */
 bool readParameters(ServoParametersPtr& parameters, const rclcpp::Node::SharedPtr& node, const rclcpp::Logger& logger,
-                    std::string ns = "moveit_servo")
+                    std::string ns /* = "moveit_servo"*/)
```
henningkayser, line 90: no default? Then I would remove this comment

moveit_ros/moveit_servo/test/servo_launch_test_common.hpp
henningkayser: Nice solution! I had something similar in mind (more inside the struct for const/non-const parameters) but this is probably the way to go for the beginning.

moveit_ros/moveit_servo/test/test_servo_interface.cpp
```diff
@@ -162,7 +165,7 @@ TEST_F(ServoFixture, SendJointServoTest)
   log_time_start = node_->now();
 
   // Allow the last command to go stale and measure the output position
-  const int sleep_time = 1.5 * 1000 * parameters_->incoming_command_timeout;
+  const int sleep_time = 1.5 * 1000 * servo_parameters_->incoming_command_timeout;
```
henningkayser, line 168: nit: using seconds would give you a `duration` in exchange for a `1000 *` and a `milli`. Or maybe even nicer to read would be `rclcpp::sleep_for(1.5s *  servo_parameters_->incoming_command_timeout);`...

moveit_ros/moveit_servo/test/unit_test_servo_calcs.cpp
henningkayser: very useful example

henningkayser (approved):
I added some suggestions for improvements, but in general I approve these changes

AndyZe (changes requested):
Given that Servo has been broken for some time (in MoveIt2), I would like to start testing all changes. Blocking until I have time for that.

moveit_ros/moveit_servo/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ cmake_minimum_required(VERSION 3.5)
 project(moveit_servo)
 
 set(SERVO_LIB_NAME moveit_servo)
```
AndyZe, line 4: Please add a comment for what each of these is. Remember that to most people, it's hard to tell the difference between `moveit_servo` and `servo_server`. I think that should be:

```
# Core Servo library, for calculations and collision checking. Provides a C++ interface
set(SERVO_LIB_NAME moveit_servo)
# Library for parameter loading
set(SERVO_PARAM_LIB_NAME ${SERVO_LIB_NAME}_parameters)
# A library to run as a ROS node component, and receive commands via topics
set(COMPOSABLE_NODE_NAME servo_server)
```

moveit_ros/moveit_servo/CMakeLists.txt
```diff
@@ -47,17 +48,22 @@ include_directories(
 ## Library to process realtime twist or joint commands ##
 #########################################################
 
+# This library provides a way of loading parameters for servo
+add_library(${SERVO_PARAM_LIB_NAME} SHARED src/servo_parameters.cpp)
+set_target_properties(${SERVO_PARAM_LIB_NAME} PROPERTIES VERSION "${${PROJECT_NAME}_VERSION}")
+ament_target_dependencies(${SERVO_PARAM_LIB_NAME} ${THIS_PACKAGE_INCLUDE_DEPENDS})
+
 # This library provides an interface for sending realtime twist or joint commands to a robot
```
AndyZe, line 56: Let's specify that it's a C++ interface, not a ROS2 topic interface:

`# This core library provides a C++ interface for sending realtime twist or joint commands to a robot`

moveit_ros/moveit_servo/CMakeLists.txt
```diff
@@ -47,17 +48,22 @@ include_directories(
...
   src/servo_calcs.cpp
   src/collision_check.cpp
   src/low_pass_filter.cpp
-  include/moveit_servo/servo_parameters.cpp
 )
 set_target_properties(${SERVO_LIB_NAME} PROPERTIES VERSION "${${PROJECT_NAME}_VERSION}")
 ament_target_dependencies(${SERVO_LIB_NAME} ${THIS_PACKAGE_INCLUDE_DEPENDS})
+target_link_libraries(${SERVO_LIB_NAME} ${SERVO_PARAM_LIB_NAME})
 
 # Add and export library for composable node server
```
AndyZe, line 68: `# Add and export library to run as a ROS node component, and receive commands via topics`

moveit_ros/moveit_servo/CMakeLists.txt
```diff
@@ -202,13 +184,33 @@ if(BUILD_TESTING)
 
 endif()
 
+ament_export_include_directories(include)
+ament_export_libraries(
+  ${SERVO_LIB_NAME}
+  ${SERVO_PARAM_LIB_NAME}
```
AndyZe, line 190: Why not export `COMPOSABLE_NODE_NAME` too? This is the library that provides a ROS node component

moveit_ros/moveit_servo/doc/servo_tutorial.md
```diff
@@ -35,7 +35,7 @@ Inputs to `moveit_servo` include
 
 
 ### Interface Options
-In ROS2, `moveit_servo` offers 2 interfacing options: a direct C++ API, and as a standalone composable node. 
+In ROS2, `moveit_servo` offers 2 interfacing options: a direct C++ API, and as a standalone composable node.
```
AndyZe, line 38: nit: `standalone node component that receives commands as ROS topics`

moveit_ros/moveit_servo/doc/servo_tutorial.md
```diff
@@ -69,15 +69,15 @@ The bare minimum requirements for running `moveit_servo` on your robot include:
 Because the kinematics are handled by the core parts of `moveit`, it is recommended that you have a valid `moveit_config` package for your robot and you can run the demo launch file included with it.
 
 #### A note on controllers
-As of now, there is no [ros2_controls controller](https://github.com/ros-controls/ros2_controllers) that can work with `moveit_servo`. ROS1 controllers that worked include `position_controllers/JointGroupPositionControllers` and `velocity_controllers/JointGroupVelocityControllers`. 
+As of now, there is no [ros2_controls controller](https://github.com/ros-controls/ros2_controllers) that can work with `moveit_servo`. ROS1 controllers that worked include `position_controllers/JointGroupPositionControllers` and `velocity_controllers/JointGroupVelocityControllers`.
```
AndyZe, line 72: This is actually outdated now. `position_controllers/JointGroupPositionControllers` and `velocity_controllers/JointGroupVelocityControllers` are available. Suggest simply deleting this line.

AndyZe:
This would close #306

moveit_ros/moveit_servo/include/moveit_servo/servo.h
```diff
@@ -86,20 +86,12 @@ class Servo
   /** \brief Get the parameters used by servo node. */
   const std::shared_ptr<moveit_servo::ServoParameters>& getParameters() const;
 
-  /** \brief Change the controlled link. Often, this is the end effector
-   * This must be a link on the robot since MoveIt tracks the transform (not tf)
-   */
-  void changeRobotLinkCommandFrame(const std::string& new_command_frame)
```
AndyZe, line 92: Why delete this?

moveit_ros/moveit_servo/src/collision_check.cpp
AndyZe: yeah, i tend to agree

moveit_ros/moveit_servo/test/servo_launch_test_common.hpp
AndyZe: Can you add a further comment at L86.5 about why that matters? For those of us who are new to ROS2   ;)

Maybe something like:

`The constant can be shared by many individual node components`

moveit_ros/moveit_servo/test/servo_launch_test_common.hpp
```diff
@@ -454,10 +467,21 @@ class ServoFixture : public ::testing::Test
...
   rclcpp::Node::SharedPtr node_;
   rclcpp::Executor::SharedPtr executor_;
-  moveit_servo::ServoParametersPtr parameters_;
   std::thread executor_thread_;
+  std::shared_ptr<const moveit_servo::ServoParameters> servo_parameters_;
+  bool servo_parameters_loaded_success_ = false;
+
+  struct TestParameters
+  {
+    double PUBLISH_HZ;
```
AndyZe, line 476: The ROS2 style guide refers to the Google style guide here. It says the variable names should be camel_case.

https://google.github.io/styleguide/cppguide.html#Variable_Names

AndyZe:
Overall I really like the simplifications here. I tested that `servo_cpp_interface_demo.launch.py` still works. Hopefully both demos work after rebasing.

moveit_ros/moveit_servo/CMakeLists.txt
tylerjw: What do you think of calling the node just `servo` or `servo_node`?  The word `server` doesn't mean much in ros and seems confusing to me.

moveit_ros/moveit_servo/src/collision_check.cpp
tylerjw: Ok, I'll remove the rename.  I thought it would make it clearer because there is a pose tracker parameters struct also.  I agree that this is unrelated to this PR and shouldn't be here.

moveit_ros/moveit_servo/src/servo_parameters.cpp
tylerjw: I think it is nice to have the default args as comments in the code when they are now specified in a header file.  If you have a function that's defined in a header with default args you can't also specify them in the source file so this is the tactic I've used to show when looking at the code that the parameter has a default value.

moveit_ros/moveit_servo/test/test_servo_interface.cpp
tylerjw: I agree, I'll clean this up in a follow-on PR.  I am going to remove the rename.

moveit_ros/moveit_servo/include/moveit_servo/servo.h
tylerjw: I was going to replace this function call with a service (so it could be done through the component interface also).  However, I don't think that change belongs in this PR.  I'll put this back.

moveit_ros/moveit_servo/test/servo_launch_test_common.hpp
tylerjw: You are right.  I named these this way to make it clear that they are constants everywhere they are used.  I agree though that this is not following the style guide and I will fix it.

moveit_ros/moveit_servo/CMakeLists.txt
tylerjw: Or maybe even servo_component?

AndyZe: i like `servo_component`  :+1:

moveit_ros/moveit_servo/CMakeLists.txt
tylerjw: The way I'm using it here is even wrong (these shouldn't be library names).  Here is the issue I created about my confusion around how to use this and the official documentation.  https://github.com/ros2/ros2_documentation/issues/1094

The short of it is this line should not be here and the correct (as far as I understand now) should be this instead:

```cmake
# Install Targets
install(
  TARGETS
    ${SERVO_LIB_NAME}
    ${SERVO_PARAM_LIB_NAME}
    ${COMPONENT_NODE_NAME}
    ${SERVO_CONTROLLER_INPUT}
    ${CPP_DEMO_NAME}
    ${FAKE_SERVO_CMDS_NAME}
  EXPORT
    export_${PROJECT_NAME}
  ARCHIVE DESTINATION lib
  LIBRARY DESTINATION lib
  RUNTIME DESTINATION bin
  INCLUDES DESTINATION include
)

# This calls the install(EXPORT ...) cmake functions to export the target libraries
# so other pojects can depend on libraries from moveit_servo.
ament_export_targets(export_${PROJECT_NAME} HAS_LIBRARY_TARGET)
ament_export_dependencies(${THIS_PACKAGE_INCLUDE_DEPENDS})
```

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -1127,9 +1128,4 @@ void ServoCalcs::setPaused(bool paused)
   paused_ = paused;
 }
 
-void ServoCalcs::changeRobotLinkCommandFrame(const std::string& new_command_frame)
-{
-  parameters_->robot_link_command_frame = new_command_frame;
-}
-
```
tylerjw, line 1134: @AndyZe This is why I removed this function.  This is no longer a constant parameter.  This parameter needs to be handled separately from the rest of the parameters if it can be changed during runtime from the rest of the parameters.

What if this changes between the use of it here: https://github.com/ros-planning/moveit2/blob/bbe859ee6c2bd35c9ba01f666ff9e9a8a4915e1f/moveit_ros/moveit_servo/src/servo_calcs.cpp#L309-L314 and here: https://github.com/ros-planning/moveit2/blob/bbe859ee6c2bd35c9ba01f666ff9e9a8a4915e1f/moveit_ros/moveit_servo/src/servo_calcs.cpp#L463

Should the value of this be changed only between runs of the servo calcs loop or can it be changed while it is being used?

tylerjw:
@AndyZe @henningkayser I responded to your reviews and am now rebasing against main.  I removed the renames and will make separate PRs for them.  I made my response to comments a separate commit so it would be easy to look at it separately however I expect to squash this before landing it.

moveit_ros/moveit_servo/include/moveit_servo/servo.h
AndyZe: I'm glad you commented it. This is a useful function that I would like to keep.

moveit_ros/moveit_servo/src/servo_calcs.cpp
AndyZe: ideally it could be changed at any time, not just while Servo is paused

moveit_ros/moveit_servo/test/launch/unit_test_servo_calcs.test.py
```diff
@@ -0,0 +1,39 @@
...
+                                             description='Binary directory of package '
+                                                         'containing test executables')
+
+    servo_gtest = launch_ros.actions.Node(
+        executable=launch.substitutions.PathJoinSubstitution(
+            [launch.substitutions.LaunchConfiguration('test_binary_dir'), 'unit_test_servo_calcs']
+        ),
+        parameters=[servo_params],
+        output='screen',
+        # prefix="kitty gdb -e run --args"
```
AndyZe, line 26: nit: what is this?

AndyZe (approved):

moveit_ros/moveit_servo/test/launch/unit_test_servo_calcs.test.py
tylerjw: This was leftover from me debugging, I can remove it if you'd like.  It is in the test launch and is useful to run gdb in a separate window.  I can remove it or add a comment about it, I don't care either way.

tylerjw:
@AndyZe sorry to keep adding to this PR.  I didn't think it would be good to merge this removing the functionality of `changeRobotLinkCommandFrame` so I went and learned about how the dynamic parameter system works in ROS2.  This should be a good example for how to use it.

@henningkayser take a look at the last commit I just added to this.  It is my idea for how to add dynamic parameter handling to a parameter struct.

Another interesting thing is I realized the parameter builder struct could return a shared_ptr<const and that would guarantee it is used the way I expect.

The only thing I don't like about this dynamic parameter interface is the object that cares about the dynamic parameter has to handle the state of the parameter instead of the parameter struct handling it.  If I could have some more time to think about this maybe I could find another way around that.

tylerjw:
I refactored this and it is now passing reliably locally and I can no longer reproduce the segfault (let's hope it was an upstream bug that is now fixed).

PR #343, opened 2021-01-25
Mk/fix moveit servo linkerr

MoriKen254:
### Description
When I tried to use moveit_servo from a self created package outside from this repository, by including the related hedders as follows:

```cpp
#include <moveit_servo/servo_parameters.cpp>
#include <moveit_servo/servo.h>
```

the following link error due to multiple definition of `moveit_servo::readParameters` did not stop.

```bash
--- stderr: self_created_pkg                                                                                                                                                           
self_created_pkg: You did not request a specific build type: Choosing 'Release' for maximum performance
/usr/bin/ld: CMakeFiles/self_created_code.dir/src/self_created_code_node.cpp.o: in function `moveit_servo::readParameters(std::shared_ptr<moveit_servo::ServoParameters>&, std::shared_ptr<rclcpp::Node> const&, rclcpp::Logger const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >)':
self_created_code_node.cpp:(.text+0xc0): multiple definition of `moveit_servo::readParameters(std::shared_ptr<moveit_servo::ServoParameters>&, std::shared_ptr<rclcpp::Node> const&, rclcpp::Logger const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >)'; CMakeFiles/self_created_code.dir/src/self_created_code.cpp.self_created_code.cpp:(.text+0x970): first defined here
make[2]: *** [CMakeFiles/self_created_code.dir/build.make:358: self_created_code] Error 1
make[1]: *** [CMakeFiles/Makefile2:80: CMakeFiles/self_created_code.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
---
Failed   <<< self_created_pkg [43.3s, exited with code 2]
```
I'm pretty sure that directly including `servo_parameters.cpp` containing the definition of `moveit_servo::readParameters` causes this link error.

Thus I modified the structure of `servo_parameters` and updated the related demo codes to include `servo_parameters.h` instead of `servo_parameters.cpp`.

After applying this modification, the following demo script still works fine.

```bash
$  ros2 launch moveit_servo servo_cpp_interface_demo.launch.py
```

In addition, our original code became to also include `servo_parameters.h` and confirmed the servo have properly worked.

Please have a check of the codes. I carefully changed the codes so that the differences should be minimum.

Thank you in advance.

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo_parameters.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo_parameters.h
@@ -89,4 +89,46 @@ struct ServoParameters
 
 using ServoParametersPtr = std::shared_ptr<ServoParameters>;
 
+template <typename T>
+void declareOrGetParam(T& output_value, const std::string& param_name, const rclcpp::Node::SharedPtr& node,
+                       const rclcpp::Logger& logger)
+{
+  try
+  {
+    if (node->has_parameter(param_name))
+      node->get_parameter<T>(param_name, output_value);
+    else
+      output_value = node->declare_parameter<T>(param_name, T{});
+  }
+  catch (const rclcpp::exceptions::InvalidParameterTypeException& e)
+  {
+    // Catch a <double> parameter written in the yaml as "1" being considered an <int>
+    if (std::is_same<T, double>::value)
+    {
+      node->undeclare_parameter(param_name);
+      output_value = node->declare_parameter<int>(param_name, 0);
+    }
+    else
+    {
+      RCLCPP_ERROR_STREAM(logger, "Error getting parameter \'" << param_name << "\', check parameter type in YAML file");
+      throw e;
+    }
+  }
+  RCLCPP_INFO_STREAM(logger, "Found parameter - " << param_name << ": " << output_value);
+}
+
+/**
+ * Declares, reads, and validates parameters used for moveit_servo
+ * @param node Shared ptr to node that will the parameters will be declared in. Params should be defined in
+ * launch/config files
+ * @param logger Logger for outputting warnings about the parameters
+ * @param parameters The set up parameters that will be updated. After this call, they can be used to start a Servo
+ * instance
+ * @param ns Parameter namespace (as loaded in launch files). Defaults to "moveit_servo" but can be changed to allow
+ * multiple arms/instances
+ * @return true if all parameters were loaded and verified successfully, false otherwise
+ */
+bool readParameters(ServoParametersPtr& parameters, const rclcpp::Node::SharedPtr& node, const rclcpp::Logger& logger,
+                    std::string ns = "moveit_servo");
+
 }  // namespace moveit_servo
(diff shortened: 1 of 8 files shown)


moveit_ros/moveit_servo/include/moveit_servo/servo_parameters.h
```diff
@@ -89,4 +89,46 @@ struct ServoParameters
 
 using ServoParametersPtr = std::shared_ptr<ServoParameters>;
 
+template <typename T>
```
henningkayser, line 92: I'm not sure this function is supposed to be exposed since it's only used inside `servo_parameters.cpp` which is not meant to be a generic parameter utility library. If you are looking for something similar, maybe consider https://github.com/PickNikRobotics/rosparam_shortcuts/pull/15.

henningkayser (commented):
@MoriKen254 Thanks for preparing this fixup, this actually resolves https://github.com/ros-planning/moveit2/issues/306. In general I would be fine with making `declareOrGetParam()` public but we are planning to refactor parameter loading in MoveIt and this function will probably be removed in the not too distant future. If you need it it's more future proof to include it in your own code directly or switch to `rosparam_shortcuts`.

MoriKen254:
@henningkayser 
Thanks for you review.

Ok, I got what you mean.

Maybe just moving the `servo_parameters.cpp` to `src` dir will be a right correction for this PR.

In our original packages, we'are gonna use `rosparam_shortcuts` or implement a simple lib similar to `servo_parameters.cpp`.

moveit_ros/moveit_servo/include/moveit_servo/servo_parameters.h
MoriKen254: I agree. `declareOrGetParam` should not be exported in this case.

I will switch to just move `servo_parameters.cpp` to `src` in order to fix https://github.com/ros-planning/moveit2/issues/306.

tylerjw:
@MoriKen254 I created this PR that does something similar to yours: https://github.com/ros-planning/moveit2/pull/341

The background behind why loading parameters is as weird as it is in servo has to do with how the launching of tests with parameters was.  It previously wasn't possible to run a gtest with parameters loaded into it through the launchfile.  After getting some feedback on this from the authors of ros2/launch we now know how to go about it.

In my PR I converted the parameter loading into a shared lib that you can depend on and use as the header is also installed with the public interface.

Ideally you can use servo through ros interfaces by building your node that controls it as a component node.  If you do that (like the xbox controller demo) you will only need at most to load the same parameters into your controlling node and everything else can go through ros interfaces that will use shared memory transport (because component node).

moveit_ros/moveit_servo/include/moveit_servo/servo_parameters.h
AndyZe: If you move this function definition to servo_parameters.cpp so it's not exposed as Henning said, I would approve the PR. This looks like a nice step forward and a fix to #306.

henningkayser:
@MoriKen254 any thoughts on what @tylerjw is suggesting? I think providing a separate library for the parameters makes sense in this case and it indeed seems to solve the same problem

MoriKen254:
Thank you for your comments.

I agree with providing a separate lib.
I think this PR is supposed to be closed.

henningkayser:
Thanks @MoriKen254

PR #344, opened 2021-01-25
Unable to load move_group capabilities and controller manager

bostoncleek:
The cmake requires `EXPORT` the shared library under `INSTALL`
otherwise if building using `colcon build --symlink-install` will
fail to load the corresponding library. This issue has been fixed
for the `moveit_ros_move_group` and `moveit_simple_controller_manager`
packages.

Please see [building a library](https://index.ros.org/doc/ros2/Tutorials/Ament-CMake-Documentation/#building-a-library).

### Checklist


moveit_plugins/moveit_simple_controller_manager/CMakeLists.txt
```diff
@@ -29,20 +31,28 @@ ament_target_dependencies(${PROJECT_NAME}
...
+)
 
 install(DIRECTORY include/ DESTINATION include)
 
 pluginlib_export_plugin_description_file(moveit_core moveit_simple_controller_manager_plugin_description.xml)
+
 ament_export_libraries(${PROJECT_NAME})
+
+ament_export_dependencies(
+  Boost
```
henningkayser, line 49: I like this grouped export better as well. What do you think about using a list variable for exported dependencies that can be used for `ament_target_dependencies` as well?

henningkayser (commented):
Do we need to check for symlink installs with CI?

moveit_ros/move_group/CMakeLists.txt
```diff
@@ -49,31 +48,59 @@ add_library(moveit_move_group_default_capabilities SHARED
...
-  DESTINATION lib
-  LIBRARY
-  DESTINATION lib
-  RUNTIME
-  DESTINATION bin)
+install(
+  TARGETS
+    moveit_move_group_default_capabilities
+    moveit_move_group_capabilities_base
+  # EXPORT PROJECT_NAME or both shared libs ?
```
henningkayser, line 85: I think that only the `moveit_move_group_capabilities_base` lib needs to be exported (for linking to custom capabilities), but not `moveit_move_group_default_capabilities` since those are only loaded at runtime. BTW, `move_group` and `list_move_group_capabilities` are not exported in ROS 1 and I don't think they should. Both are executables that are not used as build dependencies anywhere. We should only have to export libraries that were listed under `catkin_package{LIBRARIES ...]`

henningkayser (approved):

henningkayser (approved):

PR #349, opened 2021-01-27
Server timeout set to infinity

bostoncleek:
By setting the timeout to infinity ensures the move group interface
is will be ready to use. This change is simply a ros2 migration
change.

### Checklist

### DIFF
--- a/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
+++ b/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
@@ -143,12 +143,8 @@ public:
 
     current_state_monitor_ = getSharedStateMonitor(node_, robot_model_, tf_buffer_);
 
-    rclcpp::Time timeout_for_servers = pnode_->now() + wait_for_servers;
-    if (wait_for_servers == rclcpp::Duration(0.0))
-      timeout_for_servers = rclcpp::Time();  // wait for ever
-
     move_action_client_ = rclcpp_action::create_client<moveit_msgs::action::MoveGroup>(pnode_, move_group::MOVE_ACTION);
-    move_action_client_->wait_for_action_server(std::chrono::nanoseconds(timeout_for_servers.nanoseconds()));
+    move_action_client_->wait_for_action_server(wait_for_servers.to_chrono<std::chrono::duration<double>>());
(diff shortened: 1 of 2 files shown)


moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
```diff
@@ -144,8 +144,10 @@ class MoveGroupInterface::MoveGroupInterfaceImpl
     current_state_monitor_ = getSharedStateMonitor(node_, robot_model_, tf_buffer_);
 
     rclcpp::Time timeout_for_servers = pnode_->now() + wait_for_servers;
```
henningkayser, line 146: Doesn't `wait_for_action_server()` expect an actual timeout duration instead of a future time stamp? I thought we could just set the default value to `rclcpp::Duration(-1)` (in MGI constructor), call `client_->wait_for_action_server(wait_for_servers.toChrono())` and everyone would be happy.

tylerjw (approved):

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
bostoncleek: I agree that suggestion makes more sense. There is one issue. The function [to_chrono()](https://docs.ros2.org/foxy/api/rclcpp/classrclcpp_1_1Duration.html#a671704beeb2de5c6b48b4082c38a7804) is listed here but the compiler is unable to find it. I get the error: `error: no matching function for call to ‘rclcpp::Duration::to_chrono() const`.

bostoncleek: By default [http://docs.ros2.org/foxy/api/rclcpp_action/classrclcpp__action_1_1ClientBase.html#a725bde769458d64cd875d21b1582f589()](http://docs.ros2.org/foxy/api/rclcpp_action/classrclcpp__action_1_1ClientBase.html#a725bde769458d64cd875d21b1582f589) the timeout is infinite. I suggest we remove the `wait_for_servers` argument from the MGI constructor.

henningkayser (approved):
Yeah, that's what I meant. Thanks!

PR #350, opened 2021-01-28
[fix] Add EXPORT command to plugin library installs

SyllogismRXS:
### Description

When running the move_group node, I noticed that several plugins weren't being loaded correctly. I found that the plugin libraries were only being built in the "build" directory and not installed in the "install" directory. Looking at the recent changes in the git log, I saw that the EXPORT command was added to other moveit plugins, so I did the same for the missing plugins, which fixed the problem. I'm not sure why the EXPORT install is required for plugins in the MoveIt package, since it isn't required by other pluginlib libraries in other ament packages.

### Checklist

### DIFF
--- a/moveit_plugins/moveit_simple_controller_manager/CMakeLists.txt
+++ b/moveit_plugins/moveit_simple_controller_manager/CMakeLists.txt
@@ -30,6 +30,7 @@ ament_target_dependencies(${PROJECT_NAME}
 )
 
 install(TARGETS ${PROJECT_NAME}
+        EXPORT  ${PROJECT_NAME}
         LIBRARY DESTINATION lib
         ARCHIVE DESTINATION lib
         RUNTIME DESTINATION bin)
--- a/moveit_ros/move_group/CMakeLists.txt
+++ b/moveit_ros/move_group/CMakeLists.txt
@@ -63,6 +63,7 @@ install(TARGETS move_group list_move_group_capabilities
 
 install(TARGETS moveit_move_group_capabilities_base
   moveit_move_group_default_capabilities
+  EXPORT moveit_move_group_default_capabilities
   ARCHIVE
   DESTINATION lib
   LIBRARY


henningkayser:
Thanks for filing this fix. It seems like this is also covered by #344, correct?

SyllogismRXS:
It's close. I'm not sure if #344 is missing the `EXPORT moveit_move_group_default_capabilities` call. #344 uses the `EXPORT moveit_move_group_capabilities_base` call, but that's a different shared library.

Do you know what is different about the moveit2 CMake packages that the `EXPORT` command is required in the `INSTALL` function? This doesn't seem to be required in other ROS2 packages and I couldn't find any obvious reasons.

tylerjw:
> Do you know what is different about the moveit2 CMake packages that the `EXPORT` command is required in the `INSTALL` function? This doesn't seem to be required in other ROS2 packages and I couldn't find any obvious reasons.

I think it is required, it just is that the documentation doesn't seem clear (to me at least) how it should be done.  Here is the official documentation where the EXPORT can be seen in the install: https://index.ros.org/doc/ros2/Tutorials/Ament-CMake-Documentation/#building-a-library

tylerjw (approved):

henningkayser:
> It's close. I'm not sure if #344 is missing the `EXPORT moveit_move_group_default_capabilities` call. #344 uses the `EXPORT moveit_move_group_capabilities_base` call, but that's a different shared library.
> 
> Do you know what is different about the moveit2 CMake packages that the `EXPORT` command is required in the `INSTALL` function? This doesn't seem to be required in other ROS2 packages and I couldn't find any obvious reasons.

From my understanding it doesn't make sense to export plugin libraries that are never used as a build dependencies in other cmake packages. For the base class it makes sense since other non-default capaiblities (like the ones from Pilz) actually depend on it. @SyllogismRXS could you check if #344 solved your initial issue? If not, I'm fine with adding the default capabilities as well, but it would be great if we could get to the core of why this is necessary.

tylerjw:
Thank you again for your PR.  After much more reading I've created this PR which should fix this issue for every package in moveit: https://github.com/ros-planning/moveit2/pull/372  I'd appreciate if you wouldn't mind reviewing and testing it to see if it fixes the issues you experienced.

SyllogismRXS:
@tylerjw I rebuilt moveit2 from a clean build using your PR (https://github.com/ros-planning/moveit2/pull/372) and it installed the missing libraries correctly (I used --symlink-install).  If your commit is merged into main, we can close this PR.

tylerjw:
> From my understanding it doesn't make sense to export plugin libraries that are never used as a build dependencies in other cmake packages.

I think the cmake exports are what can be used in tests.  For this reason, if tests run the plugin then it is required to export them.  This might also just be a side affect as the library target needs to be copied to the install location but not actually put in the cmake extras export, I'm not sure how to install but not export a library.

henningkayser:
> @tylerjw I rebuilt moveit2 from a clean build using your PR (#372) and it installed the missing libraries correctly (I used --symlink-install). If your commit is merged into main, we can close this PR.

#372 is merged now, closing this PR

PR #353, opened 2021-02-01
various osx/clang touchups

Karsten1987:
### Description

I am not sure exactly if there's already an open issue/PR for fixes for OSX. If so, please disregard. 
I tried to give moveit2 a shot while testing the new ros2 control API with gazebo and ran into compilation issues.
The set of changes in this PR are really just various touchups, mostly a bit of cmake and some clang C++14 deprecation warnings.

### Checklist

### DIFF
--- a/moveit_kinematics/cached_ik_kinematics_plugin/include/moveit/cached_ik_kinematics_plugin/detail/NearestNeighborsGNAT.h
+++ b/moveit_kinematics/cached_ik_kinematics_plugin/include/moveit/cached_ik_kinematics_plugin/detail/NearestNeighborsGNAT.h
@@ -551,7 +551,9 @@ protected:
         std::vector<int> permutation(children_.size());
         for (unsigned int i = 0; i < permutation.size(); ++i)
           permutation[i] = i;
-        std::random_shuffle(permutation.begin(), permutation.end());
+        std::random_device rng;
+        std::mt19937 urng(rng());
+        std::shuffle(permutation.begin(), permutation.end(), urng);
 
         for (unsigned int i = 0; i < children_.size(); ++i)
           if (permutation[i] >= 0)
@@ -605,7 +607,9 @@ protected:
         std::vector<int> permutation(children_.size());
         for (unsigned int i = 0; i < permutation.size(); ++i)
           permutation[i] = i;
-        std::random_shuffle(permutation.begin(), permutation.end());
+        std::random_device rng;
+        std::mt19937 urng(rng());
+        std::shuffle(permutation.begin(), permutation.end(), urng);
 
         for (unsigned int i = 0; i < children_.size(); ++i)
           if (permutation[i] >= 0)
--- a/moveit_planners/ompl/ompl_interface/CMakeLists.txt
+++ b/moveit_planners/ompl/ompl_interface/CMakeLists.txt
@@ -1,5 +1,10 @@
 set(MOVEIT_LIB_NAME moveit_ompl_interface)
 
+if (APPLE)
+  # needed to bring ode on the linker path exported by homebrew's OMPL
+  link_directories(${OMPL_LIBRARY_DIRS})
+endif()
+
 add_library(${MOVEIT_LIB_NAME} SHARED
   src/ompl_interface.cpp
   src/planning_context_manager.cpp
--- a/moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
+++ b/moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
@@ -46,6 +46,9 @@
 
 #include <memory>
 
+using boost::placeholders::_1;
+using boost::placeholders::_2;
+
 namespace occupancy_map_monitor
 {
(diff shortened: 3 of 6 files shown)


moveit_ros/warehouse/warehouse/CMakeLists.txt
```diff
@@ -9,9 +9,10 @@ add_library(${MOVEIT_LIB_NAME} SHARED
   src/state_storage.cpp
   src/warehouse_connector.cpp
 )
+target_include_directories(${MOVEIT_LIB_NAME} PUBLIC include)
 set_target_properties(${MOVEIT_LIB_NAME} PROPERTIES VERSION "${${PROJECT_NAME}_VERSION}")
 ament_target_dependencies(${MOVEIT_LIB_NAME} rclcpp Boost moveit_core warehouse_ros moveit_ros_planning)
-target_link_libraries(${MOVEIT_LIB_NAME} ${OPENSSL_CRYPTO_LIBRARY})
+#target_link_libraries(${MOVEIT_LIB_NAME} ${OPENSSL_CRYPTO_LIBRARY})
```
Karsten1987, line 15: I left this comment in on purpose, mainly because I wasn't sure why it was in there in the first place but maybe Linux requires it.

moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
```diff
@@ -46,6 +46,9 @@
 
 #include <memory>
 
+using boost::placeholders::_1;
```
henningkayser, line 49: I would prefer having these specified in the bind commands explicitly. Eventually, we will switch to `std::bind()` or lambdas anyway.

moveit_ros/warehouse/warehouse/CMakeLists.txt
henningkayser: I'm also not sure about this. @RoboticsYY do you know if/why this was required?

moveit_kinematics/cached_ik_kinematics_plugin/include/moveit/cached_ik_kinematics_plugin/detail/NearestNeighborsGNAT.h
```diff
@@ -551,7 +551,9 @@ class NearestNeighborsGNAT : public NearestNeighbors<_T>
         std::vector<int> permutation(children_.size());
         for (unsigned int i = 0; i < permutation.size(); ++i)
           permutation[i] = i;
-        std::random_shuffle(permutation.begin(), permutation.end());
+        std::random_device rng;
```
henningkayser, line 554: Random device and generator could be member variables that can be reused. Also, are these relevant for any test? Otherwise, we should consider allowing to use a constant seed to prevent flaky CI (however, `std::random_shuffle` doesn't use a seed either).

moveit_planners/ompl/ompl_interface/CMakeLists.txt
```diff
@@ -1,5 +1,10 @@
 set(MOVEIT_LIB_NAME moveit_ompl_interface)
 
+if (APPLE)
+  # needed to bring ode on the linker path exported by homebrew's OMPL
+  link_directories(${OMPL_LIBRARY_DIRS})
```
henningkayser, line 5: Interesting, I didn't know there were any issues with this. Do we need a vendor package, then? @mamoll any suggestions?

henningkayser (commented):
I'm fine with merging all of these per se, just some things I'd like to first clarify in the comments.

moveit_planners/ompl/ompl_interface/CMakeLists.txt
```diff
@@ -1,5 +1,10 @@
 set(MOVEIT_LIB_NAME moveit_ompl_interface)
 
+if (APPLE)
```
DLu, line 3: To fix `ament_lint_cmake` error
```suggestion
if(APPLE)
```

moveit_ros/warehouse/warehouse/CMakeLists.txt
DLu: This is failing CI right now because of an undefined reference to MD5. 

```
--- stderr: moveit_ros_warehouse

/usr/bin/ld: libmoveit_warehouse.so.: undefined reference to `MD5'

make[2]: *** [warehouse/CMakeFiles/moveit_warehouse_broadcast.dir/build.make:493: warehouse/moveit_warehouse_broadcast] Error 1

make[1]: *** [CMakeFiles/Makefile2:189: warehouse/CMakeFiles/moveit_warehouse_broadcast.dir/all] Error 2

make: *** [Makefile:141: all] Error 2

---

Failed   <<< moveit_ros_warehouse [2min 15s, exited with code 2]
```

However, I can't find any reference to MD5.

DLu (changes requested):
Didn't pass CI originally, and is now out of date. Please resolve the conflicts when convenient.

tylerjw:
@sea-bass, here is a PR to improve something about cartesian planning. Would you mind looking over this work and letting us know if you think we should rebase this work and get it merged or if there is a newer approach to solving this problem we should document and close this PR?

tylerjw:
Issue for continuing this work here:

- https://github.com/ros-planning/moveit2/issues/1725

PR #356, opened 2021-02-03
[WIP] Port pose tracking, and fix CI timeout

schornakj:
### Description

Continues from #320. For now, increase the timeout of the failing test.

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/test/pose_tracking_test.test
+++ b/moveit_ros/moveit_servo/test/pose_tracking_test.test
@@ -15,7 +15,7 @@
     <param name="publish_frequency" type="double" value="50.0"/>
   </node>
 
-  <test pkg="moveit_servo" type="pose_tracking_test" test-name="pose_tracking_test" time-limit="300
+  <test pkg="moveit_servo" type="pose_tracking_test" test-name="pose_tracking_test" time-limit="60
     " args="">
     <param name="parameter_ns" type="string" value="" />
     <rosparam command="load" file="$(find moveit_servo)/test/config/servo_settings.yaml"/>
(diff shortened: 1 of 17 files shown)


moveit_ros/moveit_servo/test/pose_tracking_test.test
```diff
@@ -15,7 +15,7 @@
     <param name="publish_frequency" type="double" value="50.0"/>
   </node>
 
-  <test pkg="moveit_servo" type="pose_tracking_test" test-name="pose_tracking_test" time-limit="60
+  <test pkg="moveit_servo" type="pose_tracking_test" test-name="pose_tracking_test" time-limit="300
```
henningkayser: Is that really all that's needed? I'm a little concerned since the test should really not take that long, something might be blocking it that we are not aware of.

henningkayser (commented):
I think this branch needs a rebase (sorry...), but the conflicts should be fairly simple and won't break anything, at least everything worked when I tried. Also, should we continue this effort here or in #320 ?

schornakj:
> I think this branch needs a rebase (sorry...), but the conflicts should be fairly simple and won't break anything, at least everything worked when I tried. Also, should we continue this effort here or in #320 ?

Let's focus on #320. I will close this PR.

moveit_ros/moveit_servo/test/pose_tracking_test.test
schornakj: I believe that the test timeout duration is not related to the CI issue we had encountered in #320. Unfortunately that doesn't leave us any further ahead than we were previously -- while we were able to get the CI pipeline for #320 succeeding after restarting the job, we don't have a conclusive answer about why it was failing.

PR #357, opened 2021-02-04
Fix components demo by moving servo_cpp_demo out of composable container

### LINKED ISSUE #321: Running moveit servo demo - Component Demo - Teleoperation Demo
### Description
I'm completly new to ros moveit so I hope this is just an issue with my understanding.
I installed moveit servo following the instructions from  
https://github.com/ros-planning/moveit2/blob/main/moveit_ros/moveit_servo/doc/running_the_demos.md
and than tried the following demos.

If I start the Component Demo, rViz is opening with the robot in the default position but if I start the command publisher nothing is happing.
The same problem is coming up using the Teleoperation Demo

The Cpp Interface Demo on the otherside is following the expected behaviour.
The error from Possible error (below) is also coming up but seems to have no influence

### Your environment
* ROS Distro: ROS Foxy
* OS Version: Ubuntu 20.04
* Source build
* branch: main, commit: 0bd801f3dad80486455e8e9f4073720a7e5b32f8, tag: 2.1.0

### Steps to reproduce
Starting the demos right after installation

### Expected behaviour
The Component Demo should listen to the command publisher

### Actual behaviour
Nothing

Output servo_server_demo:
https://gist.github.com/yannischa/da26a53c97e4f7581b31e4605a6b43f4

Output fake_command_publisher:
```
yannick@yannick-System-Product-Name:~$ source ~/ws_ros2/install/setup.bash
yannick@yannick-System-Product-Name:~$ ros2 run moveit_servo fake_command_publisher
```

### Possible error:
[rviz2-1] [INFO] [1607863798.535862084] [moveit_ros.planning_scene_monitor.planning_scene_monitor]: Failed to call service get_planning_scene, have you launched move_group or called psm.providePlanningSceneService()?

Thank you really much for the help!

AndyZe:
### Description

Fixes #321

It's not a root cause fix, more of a band-aid.

### DIFF
--- a/moveit_ros/moveit_servo/launch/servo_server_demo.launch.py
+++ b/moveit_ros/moveit_servo/launch/servo_server_demo.launch.py
@@ -55,6 +55,18 @@ def generate_launch_description():
                                               os.path.join(get_package_share_directory("moveit_servo"), "config", "start_positions.yaml"),
                                               robot_description])
 
+
+
+
+    # The servo cpp interface demo
+    # Creates the Servo node and publishes commands to it
+    servo_node = Node(
+        package='moveit_servo',
+        executable='servo_cpp_demo',
+        output='screen',
+        parameters=[servo_params, robot_description, robot_description_semantic ]
+    )
+
     # Launch as much as possible in components
     container = ComposableNodeContainer(
             name='moveit_servo_demo_container',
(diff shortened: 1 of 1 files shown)


henningkayser (changes requested):
I think we should fix the bug instead of stepping back from using component nodes.

henningkayser:
Not needed as of #358

PR #360, opened 2021-02-09
WIP: Add dynamic_window trajectory operator plugin

sjahr:
### Description

Add trajectory operator plugin for the hybrid planning architecture that samples the local planning problem from the global reference trajectory based on the worst-case stopping time. 

### Checklist


moveit_ros/hybrid_planning/trajectory_operator_plugins/src/dynamic_window_operator.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+  // Calculate worst case stopping time
+  double worst_case_stop_time = calculateWorstCaseStopTime(current_state);
+
+  // Construct local trajectory
+  robot_trajectory::RobotTrajectory local_trajectory(reference_trajectory_->getRobotModel(),
+                                                     reference_trajectory_->getGroupName());
+
+  // Upper limit of the dynamic time window
+  double max_duration_from_start =
```
henningkayser, line 103: I think this approach only makes sense if we are able to closely follow the desired timestamps. With non-trivial local solvers, this will most certainly not be the case. A more robust approach would be to start with the current dynamic state and iterate the waypoints forward while applying the maximum deceleration until we find a waypoint that allows coming to a stop. The scaling can be applied to either the acceleration limits or the start velocity.

moveit_ros/hybrid_planning/trajectory_operator_plugins/src/dynamic_window_operator.cpp
```diff
@@ -0,0 +1,182 @@
...
+{
+  moveit::core::JointModel::Bounds kinematic_bounds;
+  double acceleration_limit = 0;
+  double joint_velocity = 0;
+  double worst_case_stop_time = 0;
+
+  for (auto joint_model : state.getRobotModel()->getActiveJointModels())
+  {
+    // Skip virtual joint
+    if (joint_model->getName() == "virtual_joint")
```
henningkayser, line 141: This shouldn't be necessary

sjahr:
This PR became outdated and the proposed functionality for Hybrid Planning Architecture should be properly implemented in  a new PR

PR #361, opened 2021-02-09
Use fake_components::GenericSystem from ros2_control

JafarAbdi:
### Description

Depends on https://github.com/ros-controls/ros2_control/pull/323 (merged)

Allows us to stop using the `fake_joint` package

### Checklist
- [x] Update servo to use the `FakeSystem`

PR #369, opened 2021-02-21
Add joystick to servo generic component

destogl:
### Description

The current demonstration component `JoyToServo` is tailor-made for the Panda manipulator, going so far that the joint names are hard-coded. It also does not provide control of all joints and all cartesian directions.

So in this PR, I propose a solution using a coupling with [ros2/teleop_twist_joy](https://github.com/ros2/teleop_twist_joy) package for the cartesian control and implementing only joint-based control using a joystick. The implementation is currently tested with a non-public manipulator using FakeSystem (#361) implementation in ros2_control. I added only a configuration file for the [Xbox 360 Wired](https://wiki.ros.org/joy?distro=noetic#Microsoft_Xbox_360_Wired_Controller_for_Linux) joystick since this is the one currently available to me.

For now, this PR draft to share my work, which could be useful for someone. Before making it a proper PR, the following need to be done:

- [ ] Add a section about this node in the documentation
- [ ] Add support for FakeSystem in Panda URDF - partially from (#361)
- [ ] Update example launch file
- [ ] Test example
- [ ] Update documentation

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/config/joystick_to_servo.yaml
+++ b/moveit_ros/moveit_servo/config/joystick_to_servo.yaml
@@ -1,9 +1,12 @@
-
 ##########################################################################
 # Modify all parameters related to joystickk to servo here
 ##########################################################################
-robot_eef_frame: tool0    # Name of the robot's end effector link defined in URDF/SRDF
-robot_base_frame: base_link  # Name of the robot's base frame used by MoveIt (usually: base_link or world)
+
+# Setup for Microsoft Xbox 360 Wired Controller
+enable_joint_commands_button: 4  # LB
+enable_twist_commands_button: 5  # RB
+switch_cartesian_command_frame_button: 1   # B
+switch_joint_axis_mode_button: 2  # X
 
 cartesian_command_frames:
   - base_link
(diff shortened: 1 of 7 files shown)


moveit_ros/moveit_servo/config/joystick_to_servo.yaml
```diff
@@ -0,0 +1,51 @@
...
+  - joint1
+  - joint2
+  - joint3
+  - joint4
+  - joint5
+  - joint6
+  - joint7
+
+# Do not change this names. The joints are order as in the list above
+num_joint_axis_control_modes: 2
```
AndyZe, line 26: I think you could use a comment here. This confused me initially:

`# Each joystick axis has up to 2 different behaviors, depending on the control mode`

moveit_ros/moveit_servo/src/teleop_demo/joystick_to_servo.cpp
```diff
@@ -0,0 +1,198 @@
+/*********************************************************************
```
AndyZe, line 1: I'm glad you put this in its own `teleop_demo` folder

AndyZe (commented):
At a high level, I like what this PR does. It goes above and beyond what you can do with teleop_tools. So I would support merging it when it works with the Panda.

moveit_ros/moveit_servo/config/joystick_to_servo.yaml
destogl, line 26:
```suggestion
# Each joystick axis has up to 2 different behaviors, depending on the control mode
num_joint_axis_control_modes: 2 
```

AndyZe:
After some more thought, I'm not sure this really belongs in the Servo package. We already have the `teleop_tools` method here as well as the `joystick_to_servo_example.cpp` (a very basic demo.

What would you think about moving this to a different repo? I created [this one ](https://github.com/PickNikRobotics/xbox_translator) and was thinking about porting it there.

moveit_ros/moveit_servo/config/xbox360_wired.yaml
```diff
@@ -0,0 +1,28 @@
+axis_linear:
```
AndyZe, line 1: One of the questions I have when looking at this is:  Why are there 2 config files? It seems like `joystick_to_servo.yaml` is for joint commands, `xbox360_wired.yaml` is for Cartesian commands?

PR #374, opened 2021-02-28
moveit servo: fix constructing duration from double & fix bug in insertRedundantPointsIntoTrajectory function

JafarAbdi:
### Description

Fixes two bugs:
1- Constructing a duration from a double by using `rclcpp::Duration::from_seconds` otherwise the values will cast to int
2- In `insertRedundantPointsIntoTrajectory` function we were starting from the third point, not the second one, so I fixed the index in the for-loop

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/config/panda_simulated_config.yaml
+++ b/moveit_ros/moveit_servo/config/panda_simulated_config.yaml
@@ -1,7 +1,7 @@
 ###############################################
 # Modify all parameters related to servoing here
 ###############################################
-use_gazebo: true # Whether the robot is started in a Gazebo simulation environment
+use_gazebo: false # Whether the robot is started in a Gazebo simulation environment
 
 ## Properties of incoming commands
 command_in_type: "unitless" # "unitless"> in the range [-1:1], as if from joystick. "speed_units"> cmds are in m/s and rad/s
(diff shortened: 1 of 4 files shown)


AndyZe (approved):
Thanks @JafarAbdi! Looks good to me.

AndyZe:
Before merging, can you try setting `use_gazebo=false` in this config file? I think it was used as a Band-Aid fix for the issue you uncovered but we shouldn't need it.

https://github.com/ros-planning/moveit2/blob/main/moveit_ros/moveit_servo/config/panda_simulated_config.yaml

JafarAbdi:
> Before merging, can you try setting `use_gazebo=false` in this config file? I think it was used as a Band-Aid fix for the issue you uncovered but we shouldn't need it.
> 
> https://github.com/ros-planning/moveit2/blob/main/moveit_ros/moveit_servo/config/panda_simulated_config.yaml

I changed it to false, the robot is working without any problem

JafarAbdi:
> Before merging, can you try setting `use_gazebo=false` in this config file? I think it was used as a Band-Aid fix for the issue you uncovered but we shouldn't need it.
> 
> https://github.com/ros-planning/moveit2/blob/main/moveit_ros/moveit_servo/config/panda_simulated_config.yaml

I changed it to false, the robot is working without any problem

PR #377, opened 2021-03-10
WIP: WMD 2021 Cartesian Planning

bostoncleek:
This PR  instantiates a BaseConstraint-derived class called, say,
ToolPathConstraint from a CartesianTrajectory.msg that represents a tool path.
The class finely interpolates points along the tool path and inserts these points in a nearest neighbor data structure.
For any random joint state, we can use FK to compute end effector pose and the nearest point on the tool path.

### Checklist

### DIFF
--- a/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/detail/ompl_constraints.h
+++ b/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/detail/ompl_constraints.h
@@ -40,10 +40,12 @@
 
 #include <ompl/base/Constraint.h>
 
+#include <moveit/ompl_interface/detail/NearestNeighborsGNAT.h>
 #include <moveit/ompl_interface/detail/threadsafe_state_storage.h>
 #include <moveit/robot_model/robot_model.h>
 #include <moveit/macros/class_forward.h>
 #include <moveit_msgs/msg/constraints.hpp>
+#include <moveit_msgs/msg/cartesian_trajectory.hpp>
 
 namespace ompl_interface
 {
@@ -333,5 +335,62 @@ Bounds positionConstraintMsgToBoundVector(const moveit_msgs::msg::PositionConstr
 std::shared_ptr<BaseConstraint> createOMPLConstraint(const moveit::core::RobotModelConstPtr& robot_model,
                                                      const std::string& group,
                                                      const moveit_msgs::msg::Constraints& constraints);
+
+/******************************************
+ * Toolpath Constraint
+ * ****************************************/
+/** \brief A path constraint for the EEF useful for sanding, painting, drawing, welding, etc
+ * */
+class ToolPathConstraint : public BaseConstraint
+{
+public:
+  ToolPathConstraint(const moveit::core::RobotModelConstPtr& robot_model, const std::string& group,
+                     unsigned int num_dofs, const unsigned int num_cons = 6, double step = 0.01);
+
+  void parseConstraintMsg(const moveit_msgs::msg::Constraints& constraints) override;
+
+  void setPath(const moveit_msgs::msg::CartesianTrajectory& msg);
+
+  void function(const Eigen::Ref<const Eigen::VectorXd>& joint_values, Eigen::Ref<Eigen::VectorXd> out) const override;
+
+private:
+  // TODO: Why won't NearestNeighborsGNAT allow me to store pose as Eigen Transform
+  mutable NearestNeighborsGNAT<geometry_msgs::msg::Pose> pose_nn_;  // store poses
+  const double step_;                                               // step size between each waypoint
+};
+
+/**
+ * @brief Distance between two poses
+ * @param pose1 - first pose
+ * @param pose2 - second pose
+ * @return arc length between two quaternion + Eulcidean distance between positions
+ */
+double poseDistance(const geometry_msgs::msg::Pose& pose1, const geometry_msgs::msg::Pose& pose2);
+
+/**
+ * @brief Compose Euclidean distance between two poses
+ * @param pose1 - first pose
+ * @param pose2 - second pose
+ * @return Euclidean distance
+ */
+double euclideanDistance(const geometry_msgs::msg::Pose& pose1, const geometry_msgs::msg::Pose& pose2);
+
+/**
+ * @brief Compose arc length between two quaternions
+ * @param pose1 - first pose
+ * @param pose2 - second pose
+ * @return Arc length
+ */
+double arcLength(const geometry_msgs::msg::Pose& pose1, const geometry_msgs::msg::Pose& pose2);
+
+/**
+ * @brief Interpolated between two poses
+ * @param pose1 - first pose
+ * @param pose2 - second pose
+ * @param step - step from first pose to pose 2 [0 1]
+ * @return Interpolated pose
+ */
+geometry_msgs::msg::Pose poseInterpolator(const geometry_msgs::msg::Pose& pose1, const geometry_msgs::msg::Pose& pose2,
+                                          double step);
 
 }  // namespace ompl_interface
--- a/moveit_planners/ompl/ompl_interface/src/detail/ompl_constraints.cpp
+++ b/moveit_planners/ompl/ompl_interface/src/detail/ompl_constraints.cpp
@@ -45,6 +45,8 @@ static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_planners_ompl.om
 
 namespace ompl_interface
 {
+static const double MAX_QUATERNION_NORM_ERROR = 1e-9;
+
 Bounds::Bounds() : size_(0)
 {
 }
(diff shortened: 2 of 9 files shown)


mamoll:
Thanks for taking this on, @bostoncleek!

moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/detail/ompl_constraints.h
```diff
@@ -334,4 +336,43 @@ std::shared_ptr<BaseConstraint> createOMPLConstraint(const moveit::core::RobotMo
...
+private:
+  mutable NearestNeighborsGNAT<geometry_msgs::msg::Pose> pose_nn_;  // store poses
+  const unsigned int num_steps_;                                    // number of step between each waypoint
+};
+
+/**
+ * @brief Distance between two poses
+ * @param pose1 - first pose
+ * @param pose2 - second pose
+ * @return arclength between two quaternion + eulcidean distance between positions
```
mamoll, line 364:
```suggestion
 * @return arc length between two quaternion + Euclidean distance between positions
```

moveit_planners/ompl/ompl_interface/src/detail/ompl_constraints.cpp
```diff
@@ -385,4 +385,101 @@ std::shared_ptr<BaseConstraint> createOMPLConstraint(const moveit::core::RobotMo
...
+ * Toolpath Constraint
+ * ****************************************/
+ToolPathConstraint::ToolPathConstraint(const moveit::core::RobotModelConstPtr& robot_model, const std::string& group,
+                                       unsigned int num_dofs, unsigned int steps)
+  : BaseConstraint(robot_model, group, num_dofs), num_steps_(steps)
+{
+  NearestNeighbors<geometry_msgs::msg::Pose>::DistanceFunction func =
+      std::bind(&poseDistance, std::placeholders::_1, std::placeholders::_2);
+
+  pose_nn_.setDistanceFunction(func);
```
mamoll, line 399: See the Effective Modern C++ book or [this](https://github.com/peter-can-write/cpp-notes/blob/master/prefer-lambdas-over-std::bind.md).
```suggestion
  pose_nn_.setDistanceFunction(
   [this](const geometry_msgs::msg::Pose& pose1, const geometry_msgs::msg::Pose& pose2) {
      return poseDistance(pose1, pose2);
   }
 )
```

moveit_planners/ompl/ompl_interface/src/detail/ompl_constraints.cpp
```diff
@@ -385,4 +385,101 @@ std::shared_ptr<BaseConstraint> createOMPLConstraint(const moveit::core::RobotMo
...
+  // Interpolate along waypoints and insert into pose_nn_;
+  const auto dt = 1.0 / static_cast<double>(num_steps_);
+  for (unsigned int i = 0; i < msg.points.size() - 1; i++)
+  {
+    // Insert waypoints
+    pose_nn_.add(msg.points.at(i).point.pose);
+
+    // Insert interpolated poses
+    auto step = dt;
+    for (unsigned int j = 0; j < num_steps_ - 1; j++)
```
mamoll, line 418: it might make more sense to have a variable that indicates the max spacing between waypoints and use that to determine how many interpolated poses you need between any pair of waypoints.

moveit_planners/ompl/ompl_interface/src/detail/ompl_constraints.cpp
```diff
@@ -385,4 +385,101 @@ std::shared_ptr<BaseConstraint> createOMPLConstraint(const moveit::core::RobotMo
...
+  // Insert last waypoint
+  pose_nn_.add(msg.points.at(msg.points.size() - 1).point.pose);
+}
+
+
+void ToolPathConstraint::function(const Eigen::Ref<const Eigen::VectorXd>& joint_values, Eigen::Ref<Eigen::VectorXd> out) const
+{
+  // TODO: Add path constraint function
+  // Ideal: F(q_actual) - F(q_neared_pose) = 0
+
```
mamoll, line 435: * Use forward kinematics on joint_values to compute EEF pose `pose1`
* Use `pose_nn_` to find pose `pose2` that is nearest to `pose1`
* Return `poseDistance(pose1, pose2)`

moveit_planners/ompl/ompl_interface/src/detail/ompl_constraints.cpp
```diff
@@ -385,4 +385,101 @@ std::shared_ptr<BaseConstraint> createOMPLConstraint(const moveit::core::RobotMo
...
+  // Distance between two quaternions
+  // https://math.stackexchange.com/questions/90081/quaternion-distance
+
+  // dot product between q1 and q2
+  const auto qdotprod = pose1.orientation.w * pose2.orientation.w + pose1.orientation.x * pose2.orientation.x +
+                        pose1.orientation.y * pose2.orientation.y + pose1.orientation.z * pose2.orientation.z;
+
+  // Euclidean + Quaternion Distance
+  return std::sqrt(dx * dx + dy * dy + dz * dz) + (1.0 - qdotprod * qdotprod);
```
mamoll, line 454: quaternion distance is not right. See https://github.com/ompl/ompl/blob/master/src/ompl/base/spaces/src/SO3StateSpace.cpp#L254-L262
The `fabs` and `acos` are both important.

bostoncleek:
@mamoll Thank you this sort of feedback is very beneficial.

gautz:
If I understand well, your goal is to use the constrained planner to plan e.g. a spraying path? The constraints would be defined by discrete trajectory points? Do you plan to use the Cartesian Trajectory proposed interface? https://fzi-robot-interface-proposal.readthedocs.io/en/latest/proposal/index.html#tldr-proposed-interface

I am familiar with ompl_constraints.cpp but not with the classical cartesian planning. I never used the CartesianTrajectory.msg, thus it would be nice if you could explain the idea here.. 
Where would your input come from? A 3D path generated on a CAD file in a CAM system? In the form of linear, circular and spline segment?

I am working on possibly a similar problem, just want to avoid dual work.

bostoncleek:
@gautz The idea is that the end effector would follow a path (spraying, welding, drawing, etc). The `CartesianTrajectory.msg` should contain a list of waypoints for the end effector to follow. 

As of now the user must specify the waypoints. But as you can imagine the waypoints could come from a CAM system or another path generator. You could even run canny edge detection on an imagine and convert the pixels representing contours to Cartesian space. The robot would then ideally be able to draw the picture.

moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/detail/ompl_constraints.h
```diff
@@ -334,4 +336,61 @@ std::shared_ptr<BaseConstraint> createOMPLConstraint(const moveit::core::RobotMo
...
+                     unsigned int num_dofs, const unsigned int num_cons = 6, double step = 0.01);
+
+  void parseConstraintMsg(const moveit_msgs::msg::Constraints& constraints) override;
+
+  void setPath(const moveit_msgs::msg::CartesianTrajectory& msg);
+
+  void function(const Eigen::Ref<const Eigen::VectorXd>& joint_values, Eigen::Ref<Eigen::VectorXd> out) const override;
+
+private:
+  // TODO: Why won't NearestNeighborsGNAT allow me to store pose as Eigen Transform
```
schornakj, line 357: @bostoncleek What kind of error do you get when you try to use an Eigen Transform as the template type?

gautz:
@bostoncleek any plans to implement a jacobian() with this generic Cartesian path approach? Or should the function() be sufficient for performance?

bostoncleek:
> @bostoncleek any plans to implement a jacobian() with this generic Cartesian path approach? Or should the function() be sufficient for performance?

Ompl can compose the derivative numerically. Of course the jacobian will allows provide an increase in performance.

tylerjw:
@sea-bass this one, sorry about that other issue. Could you review this one and advise?

moveit_planners/ompl/ompl_interface/src/detail/ompl_constraints.cpp
```diff
@@ -385,4 +387,151 @@ std::shared_ptr<BaseConstraint> createOMPLConstraint(const moveit::core::RobotMo
...
+  eef_pose.position.y = trans.y();
+  eef_pose.position.z = trans.z();
+
+  eef_pose.orientation.w = eef_quat.w();
+  eef_pose.orientation.x = eef_quat.x();
+  eef_pose.orientation.y = eef_quat.y();
+  eef_pose.orientation.z = eef_quat.z();
+
+  // Nearest end-effector pose in trajectory
+  const geometry_msgs::msg::Pose eef_nearest_pose = pose_nn_.nearest(eef_pose);
```
sea-bass, line 464: I wonder if it makes sense to call `nearestK` instead, get the 2 nearest points, and then interpolate between those?

That would give you a nice tradeoff between having to cache a super fine set of points up front, and doing a *little* more at runtime to get something nearer to the desired pose goal.

sea-bass (commented):
I think having a nearest neighbors based path constraint for planning is worthwhile to pull in.

However, my overarching comments to get this in would be:
1. Do we really need to copy-paste and modify 3 OMPL files headers here, or can we make the necessary changes to OMPL itself?
2. A lot of the distance functions are already available in Eigen, and since we're converting anyway we may as well just use Eigen instead of geometry_msgs.
3. The pose interpolation function does a lot of superfluous conversions from geometry_msgs to Eigen types, but combined with the above point, this could be easily addressed

And then, we want a compelling demo that maybe sets e.g., a circular path and allows users to plan within that constraint.

sea-bass:
By request, I'm closing this PR and capturing that we should pick up this work in a new issue : https://github.com/ros-planning/moveit2/issues/1740

moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/detail/ompl_constraints.h
sea-bass: I was curious about this and it turns out the issue is there is no `==` operator implemented for these Eigen types:

```
/opt/ros/humble/include/ompl-1.6/ompl/datastructures/NearestNeighborsGNAT.h:551:102: error: no match for ‘operator==’ (operand types are ‘const Eigen::Transform<double, 3, 1>’ and ‘const Eigen::Transform<double, 3, 1>’)
  551 |                 if (dist < nbh.top().first || (dist < std::numeric_limits<double>::epsilon() && data == key))
```

PR #382, opened 2021-03-12
Remove ROS_WS env variable from foxy-ci which is causing the overlay to be installed on top of the underlay

JafarAbdi:
### Description

As you can see in https://travis-ci.com/github/ros-planning/moveit_task_constructor/jobs/489509921 having `ROS_WS` is causing the overlay ws to be installed on top of the underlay hence compiling the underlay again

### Checklist

### DIFF
--- a/.docker/ci-testing/Dockerfile
+++ b/.docker/ci-testing/Dockerfile
@@ -16,21 +16,3 @@ RUN apt-get -qq update && \
     #
     # Clear apt-cache to reduce image size
     rm -rf /var/lib/apt/lists/*
-
-# Download moveit source and fetch all its dependencies
-# TODO: remove all this when the migration is finished and all dependencies are install from ci base image
-RUN mkdir src && \
-    cd $ROS_WS/src && \
-    wget -q https://raw.githubusercontent.com/ros-planning/moveit2/main/moveit2.repos && \
-    vcs import < moveit2.repos && \
-    # Remove folders declared as COLCON_IGNORE
-    find -L . -name COLCON_IGNORE -printf "%h\0" | xargs -0 rm -rf && \
-    apt-get -qq update && \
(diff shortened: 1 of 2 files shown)


rhaschke (approved):
Yeah, we don't define ROS_WS for MoveIt 1 as well.

.docker/ci-testing/Dockerfile
```diff
@@ -16,21 +16,3 @@ RUN apt-get -qq update && \
...
-    find -L . -name COLCON_IGNORE -printf "%h\0" | xargs -0 rm -rf && \
-    apt-get -qq update && \
-    rosdep update -q && \
-    rosdep install -q -y --from-paths . --ignore-src --rosdistro ${ROS_DISTRO} --as-root=apt:false && \
-    # Clear apt-cache to reduce image size
-    rm -rf /var/lib/apt/lists/* && \
-    #
-    # Remove the source code from this container
-    cd $ROS_WS && \
-    rm -rf src/
```
rhaschke, line 36: Implementing the TODO and removing all this is unrelated to this PR and should be a separate commit.

JafarAbdi: These lines use `ROS_WS` this's why I combined the changes in the same PR

JafarAbdi: I separated the changes into different commits, I don't think we need to implement anything here since we already run the commands in https://github.com/ros-planning/moveit2/blob/main/.docker/ci/Dockerfile

rhaschke: In the past, it was important to fetch all dependencies from source here as well. Now, that everything has settled, we can drop this again (as stated in the TODO).

rhaschke (approved):

PR #386, opened 2021-03-17
Add initialize function for moveit_sensor_manager plugin

JafarAbdi:
### Description

Currently, there's no way to pass parameters to the sensor manager when creating it, so I added an initialize function similar to all other MoveIt plugins

### Checklist


mamoll (approved):

moveit_ros/planning/plan_execution/src/plan_with_sensing.cpp
```diff
@@ -107,7 +106,11 @@ plan_execution::PlanWithSensing::PlanWithSensing(
       try
       {
         if (node_->get_parameter("moveit_sensor_manager", manager))
+        {
           sensor_manager_ = sensor_manager_loader_->createUniqueInstance(manager);
+          if (!sensor_manager_->initialize(node_))
+            RCLCPP_ERROR_STREAM(LOGGER, "Failed to initialize " << manager);
```
henningkayser, line 112: Shouldn't the sensor manager be unset in that case?

henningkayser (approved):

moveit_ros/planning/plan_execution/src/plan_with_sensing.cpp
JafarAbdi: Nice catch, thanks!

PR #389, opened 2021-03-18
Add Github Actions and pre-commit for Format checking

tylerjw:
### Description

This adds the very awesome pre-commit solution to clang-format and various other issues to this repo.  This awesome feature was shared with me by @rhaschke [here](https://github.com/ros-planning/warehouse_ros/pull/54#issuecomment-801326902).  He used it in MTC and I added it to warehouse_ros.  This PR does two things:

* Add the pre-commit config file so people can use it locally to fix clang-format and other issues when they try to commit.
* Add a github action that tests using this same config.

To test running this locally you need to install [pre-commit](https://pre-commit.com/):

```bash
pip3 install pre-commit
```

Then you need to install it in your git environment:

```bash
pre-commit install
```

Then when you try to commit if the pre-commit config exists it runs the tests in it.  Alternatively you can run the tests themselves like this:

```bash
pre-commit run -a
```

One last thing to note.  The pre-commit checks are simplified from the config used in MTC.  That is because the xml checker is failing and I can't figure out how to fix the two issues it finds, and secondly the `MoveIt` spelling checker fails in the MIGRATION.md document in a way I'm not sure how to fix.  I figured those two things could be done afterwards in separate PRs.


gavanderhoorn:
> Add Github Actions and pre-commit for Fromat checking 

If only we had had some fromat checking earlier .. ;)

tylerjw:
> If only we had had some fromat checking earlier .. ;)

What is great about this is that it unblocks migrating to circle_ci as ament_clang_format isn't usable yet and this offloads the testing of formatting to github actions.

tylerjw:
The end of line thing, it will be so nice to stop having to ask people to configure their editors correctly as it'll be done for them.

gavanderhoorn:
I was hoping the smiley was enough, but you misspelled *format* as *fromat* ;)

tylerjw:
I thought you were making a joke about how I was adding a _currently_ redundant clang-format CI run (I didn't remove it from travis yet as I plan on replacing travis).  Thank-you for clarifying.  I do think this feature is really awesome and am going to try to add it everywhere.

tylerjw:
> And perhaps a more useful comment: I personally try to separate whitespace/formatting changes from actual content/structural changes.
> 
> You have a single commit here, which mixes both.

That is always an issue when you enable new format checkers.  I included the changes needed so the code passes CI after this is merged.  Would you prefer I made the CI change that first that will fail separately or the format changes first and then the CI change?

I guess I figured the changes were minor enough it was ok.  The formatting changes seem really trivial to me.  They are newlines at end of file, python3 print statements, and yaml file fixes to our .clang-format file.

gavanderhoorn:
> That is always an issue when you enable new format checkers. I included the changes needed so the code passes CI after this is merged. Would you prefer I made the CI change that first that will fail separately or the format changes first and then the CI change?

CI runs against `HEAD` of a *branch*, not individual commits.

Wouldn't it have been possible to commit your CI change/addition first, and then add the fixup commit, then push?

tylerjw:
> CI runs against `HEAD` of a _branch_, not individual commits.
> 
> Wouldn't it have been possible to commit your CI change/addition first, and then add the fixup commit, then push?

I always forget this is an option because I try to make each change as a single commit so it can be squash-merged (because of how badly merges are handled by the changelog script).

All that being said I'll make this change to this PR and we can do a normal merge with it.

gavanderhoorn:
> I always forget this is an option because I try to make each change as a single commit so it can be squash-merged

this is starting to off-topic, but this is not a really nice thing to do.

We have history for a reason.

A single misbehaving tool is not a sufficient reason to throw that all away.

tylerjw:
Back to the change at hand, what do you think of this approach for clang-format and other formatting checking?

JafarAbdi (approved):

PR #390, opened 2021-03-18
Planar Joint Properties

DLu:
### Description

Set the joint properties (angular distance factor and motion model) for planar joint models. 

### Checklist


moveit_core/robot_model/src/planar_joint_model.cpp
```diff
@@ -139,39 +141,109 @@ void PlanarJointModel::getVariableRandomPositionsNearBy(random_numbers::RandomNu
...
+void computeTurnDriveTurnGeometry(const double* from, const double* to, double& dx, double& dy, double& initial_turn,
+                                  double& drive_angle, double& final_turn)
+{
+  dx = to[0] - from[0];
+  dy = to[1] - from[1];
+  drive_angle = atan2(dy, dx);
+
+  initial_turn = angles::shortest_angular_distance(from[2], drive_angle);
+  drive_angle = from[2] + initial_turn;
+  final_turn = to[2] - drive_angle;
```
mamoll, line 153: don't you need angles::shortest_angular_distance here as well?

mamoll (approved):
I think there's one tiny bug, but otherwise this looks good to me.

moveit_core/robot_model/src/planar_joint_model.cpp
DLu: No, and I should add comments to the code explicitly stating why.

moveit_core/robot_model/src/planar_joint_model.cpp
```diff
@@ -139,39 +141,128 @@ void PlanarJointModel::getVariableRandomPositionsNearBy(random_numbers::RandomNu
   normalizeRotation(values);
 }
 
+void computeTurnDriveTurnGeometry(const double* from, const double* to, const double min_translational_distance,
+                                  double& dx, double& dy, double& initial_turn, double& drive_angle, double& final_turn)
+{
+  dx = to[0] - from[0];
+  dy = to[1] - from[1];
+  // If the translational distance between from & to states is very small, it will cause an unnecessary rotations since
```
AndyZe, line 149: rotation, no **s**

moveit_core/robot_model/src/planar_joint_model.cpp
```diff
@@ -139,39 +141,128 @@ void PlanarJointModel::getVariableRandomPositionsNearBy(random_numbers::RandomNu
   normalizeRotation(values);
 }
 
+void computeTurnDriveTurnGeometry(const double* from, const double* to, const double min_translational_distance,
```
AndyZe, line 144: I'm not a fan of double* type. Would prefer a std::vector. Maybe @jliukkonen or @tylerjw has a strong opinion

I see you inherited the previous style, though

moveit_core/robot_model/include/moveit/robot_model/planar_joint_model.h
```diff
@@ -75,13 +82,52 @@ class PlanarJointModel : public JointModel
...
+  MotionModel motion_model_;
+  /// Only used for the differential drive motion model @see computeTurnDriveTurnGeometry
+  double min_translational_distance_;
 };
+
+/**
+ * @brief Compute the geometry to turn toward the target point, drive straight and then turn to target orientation
+ * @param[in]  from                       A vector representing the initial position [x0, y0, theta0]
+ * @param[in]  to                         A vector representing the target position  [x1, y1, theta1]
+ * @param[in]  min_translational_distance If the translational distance between \p from and \p to is less than this
```
AndyZe, line 121: units are meters?

moveit_core/robot_model/src/planar_joint_model.cpp
```diff
@@ -139,39 +141,128 @@ void PlanarJointModel::getVariableRandomPositionsNearBy(random_numbers::RandomNu
...
+    double drive_d = hypot(dx, dy);
+    double final_d = fabs(final_turn) * angular_distance_weight_;
+
+    double total_d = initial_d + drive_d + final_d;
+
+    double initial_frac = initial_d / total_d;
+    double drive_frac = drive_d / total_d;
+    double final_frac = final_d / total_d;
+
+    double pct;
```
AndyZe, line 221: would prefer "percent" so I don't have to spend 2 seconds guessing what the variable means

moveit_core/robot_model/src/planar_joint_model.cpp
```diff
@@ -139,39 +141,128 @@ void PlanarJointModel::getVariableRandomPositionsNearBy(random_numbers::RandomNu
...
+    if (t <= initial_frac)
```
AndyZe, line 222: I'm confused about what's going on here. A time is being compared to a distance fraction? Doesn't seem to make sense.

moveit_core/robot_model/src/planar_joint_model.cpp
mamoll: This is a consequence of the interpolate signature in the base class where double* is used.

moveit_core/robot_model/src/planar_joint_model.cpp
mamoll: `t` is not really time but a fraction (0<=t<=1). This is the convention used in the ompl_interface code. Perhaps this could be documented better...

mamoll (approved):

moveit_core/robot_model/src/planar_joint_model.cpp
AndyZe: Ah, I even made a PR to document that in MoveIt 1 but I guess it has not been ported to MoveIt2 yet.

https://github.com/ros-planning/moveit/pull/2528

AndyZe (approved):

DLu:
:smile:

PR #391, opened 2021-03-18
WIP: Use Reflexxes to smooth commands in moveit_servo

AndyZe:
Reflexxes Type II is a publicly-available library that provides acceleration and velocity-constrained realtime command smoothing. It's kind of like a low-pass filter except more configurable, and it runs fast (>1kHz). This PR deletes the low pass filter stuff from Servo and replaces it with Reflexxes. The Reflexxes library is wrapped for ROS [here](https://github.com/PickNikRobotics/ros_reflexxes).


moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -732,60 +770,51 @@ double ServoCalcs::velocityScalingFactorForSingularity(const Eigen::VectorXd& co
...
-  Eigen::ArrayXd velocity = delta_theta / parameters_->publish_period;
-
-  std::size_t joint_delta_index = 0;
-  // Track the smallest velocity scaling factor required, across all joints
-  double velocity_limit_scaling_factor = 1;
-
-  for (auto joint : joint_model_group_->getActiveJointModels())
+  // Update current state in Reflexxes
+  // For current state, we know the current position well. For vel and accel, use the previous output from Reflexxes
+  // This assumes the hardware tracks the requested vel/accel well.
```
gavanderhoorn, line 777: Yes, a rather strong assumption.

I've run into this before and it doesn't always work.

tylerjw (commented):
Read through the changes and I didn't see any issues.  :+1: 

It needs to be rebased because my refactoring of parameters got merged.  I'm sorry about that, it shouldn't be too difficult though.  Request another review from me once you've done that.

I'm not sure how to test these changes.  I'll assume you are using it on a project and have tests for it.

moveit_ros/moveit_servo/src/servo_calcs.cpp
AndyZe: OK, I understand. Not sure I will change it right now.

BTW I saw the 2 posts on ROS Discourse about jerk-limited smoothing. I would love to use one of those but unfortunately this PR was 90% done when I saw the posts. At least this PR will serve as a template, if somebody (maybe me!) wants to use one of those libraries instead.

Another thing I saw is that Reflexxes tends to overshoot depending on the acceleration limit, its current velocity, and target velocity. Basically it needs to overshoot to give itself more time to accelerate to the target velocity. This is pretty counterintuitive and might be a surprise for many people. The robot can even more backward if the wrong control frequency is used.

It takes careful parameter tuning. I found that running at a lower control frequency helped.

AndyZe: I'm interested to know which of these _you_ think moveit_servo should use:

- Reflexxes II
- Simple lowpass filter
- opt_control
- pantor/ruckig

The lowpass filter has a lot of benefits. It's basically idiot-proof but it doesn't work on some robots.

v4hn:
Reflexxes is of course no "command smoothing", but a full trajectory generator, though you can use it to smooth if you know what you're doing (You already pointed out it can behave in rather unexpected ways when accelerations are involved).

> Reflexxes Type II is a publicly-available library

That's true, but it seems to be released under LGPL. That's worth keeping in mind as there have been efforts to keep MoveIt at BSD 3 clause in the past.

> I'm interested to know which of these you think moveit_servo should use:

To me this variety of filter alternatives sounds very much like a plugin interface. MoveIt is good at these, so why force the use of one algorithm here?

AndyZe:
> To me this variety of filter alternatives sounds very much like a plugin interface. MoveIt is good at these, so why force the use of one algorithm here?

Agree! Will work on that.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -149,16 +153,53 @@ ServoCalcs::ServoCalcs(rclcpp::Node::SharedPtr node, const ServoParametersPtr& p
...
   tf_moveit_to_robot_cmd_frame_ = empty_matrix;
+
+  // Reflexxes for command smoothing
+  reflexxes_ = std::make_unique<ReflexxesAPI>(num_joints_, parameters_->publish_period, 0 /* additional threads */);
+  reflexxes_position_input_param_ = std::make_unique<RMLPositionInputParameters>(num_joints_);
+  reflexxes_position_output_param_ = std::make_shared<RMLPositionOutputParameters>(num_joints_);
+  // All joints are active
+  reflexxes_wrapper::setSelectionVectorAllTrue(reflexxes_position_input_param_.get(), num_joints_);
+  // Initialize positions/vels/accels to zeros. Current positions will be updated later
+  reflexxes_wrapper::initializePositionInputStateToZeros(reflexxes_position_input_param_.get(), num_joints_);
```
destogl, line 171: Is it safe to initialize the position vector to 0? Position should correspond to the current state of the robot. In ros2_control we are initializing positions to `std::numeric_limits<double>::quiet_NaN()` in constructors and then when the robot states are updated to real values assign those. Using this approach we can distinguish if data are initialized from the hardware or only variable initialization is done. Otherwise, unintended movement toward positions 0 could happen. (May is this protected with another variable, if so, ignore this comment).

moveit_ros/moveit_servo/src/servo_calcs.cpp
destogl: @AndyZe: you could use this assumption only for robots not providing velocity and acceleration in their `joint_state` message. Or what you think about calculating those from positions? (I know that derivative are not so simple to calculate, but with sufficient frequency, they should be sufficiently correct).

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -732,60 +770,51 @@ double ServoCalcs::velocityScalingFactorForSingularity(const Eigen::VectorXd& co
...
+                                    target_velocities);
+
+  // Call the Reflexxes algorithm
+  int result = reflexxes_->RMLPosition(*reflexxes_position_input_param_, reflexxes_position_output_param_.get(),
+                                       reflexxes_flags_);
+
+  // Calculate the new delta_theta from Reflexxes output
+  for (size_t joint_idx = 0; joint_idx < num_joints_; ++joint_idx)
+  {
+    delta_theta[joint_idx] = reflexxes_position_output_param_->NewPositionVector->VecData[joint_idx] -
```
destogl, line 807: Is this correct? You are actually calculating `delta_theta` from the step *k* and using it to set reference in step *k+1*.

I am not certain if this is plausible? Are you trying to keep the robot at its velocity limit?

Actually, thinking about this.... This could cause the robot moving backward if the frequency is not sufficient. When we used reflexxes before for the trajectory smoothing in controllers then we always set as goal current position reference (target state) and if needed, updated it at every control update. This leverage the trajectory generation of reflexxes and always worked very well. Maybe it is worth trying, or do you see some issue with that approach?

P.S. Knowing the author of reflexxes personally, I could check with him what is the right way to use the library.

destogl (commented):
The changes look good, but I am not sure reflexxes library is used properly which could lead to "strange" behavior or robot moving backwards. I wrote a longer comment and explanation in the corresponding line.

As an idea, you can use reflexxes for limiting accelerations in Cartesian space also, if there is not solved yet.

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
```diff
@@ -43,6 +43,10 @@
 #include <mutex>
 #include <thread>
 
+// Reflexxes, for smoothing
+#include <libreflexxestype2/ReflexxesAPI.h>
```
AndyZe, line 47: TODO: wrapper pkg should include this dep already. Delete it here.

moveit_ros/moveit_servo/src/servo_calcs.cpp
AndyZe: good idea - done!

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -149,16 +153,53 @@ ServoCalcs::ServoCalcs(rclcpp::Node::SharedPtr node, const ServoParametersPtr& p
...
+  reflexxes_position_input_param_ = std::make_unique<RMLPositionInputParameters>(num_joints_);
+  reflexxes_position_output_param_ = std::make_shared<RMLPositionOutputParameters>(num_joints_);
+  // All joints are active
+  reflexxes_wrapper::setSelectionVectorAllTrue(reflexxes_position_input_param_.get(), num_joints_);
+  // Initialize positions/vels/accels to zeros. Current positions will be updated later
+  reflexxes_wrapper::initializePositionInputStateToZeros(reflexxes_position_input_param_.get(), num_joints_);
+  // Set velocity and acceleration limits
+  std::vector<double> velocity_limits;
+  std::vector<double> acceleration_limits;
+  for (auto joint : joint_model_group_->getActiveJointModels())
```
AndyZe, line 175: joint& : ...

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -149,16 +153,53 @@ ServoCalcs::ServoCalcs(rclcpp::Node::SharedPtr node, const ServoParametersPtr& p
...
+  {
+    // Some joints do not have bounds defined
+    const auto bound = joint->getVariableBounds(joint->getName());
```
AndyZe, line 178: bound&

moveit_ros/moveit_servo/src/servo_calcs.cpp
AndyZe, line 175: const auto&

moveit_ros/moveit_servo/src/servo_calcs.cpp
gavanderhoorn: > I'm interested to know which of these *you* think moveit_servo should use:

I don't really have a preference. Getting `moveit_servo` to output jerk limited control signals would be great, so anything which does that would seem to be acceptable.

My comment was really just about the comment I commented on (hah).

It's a rather strong assumption which is easily violated, and it's buried somewhere in the code.

@destogl wrote:
> Or what you think about calculating those from positions? (I know that derivative are not so simple to calculate, but with sufficient frequency, they should be sufficiently correct).

I've seen some really weird things happen with derivatives and reflexxes / other trajectory generators.

moveit_ros/moveit_servo/CMakeLists.txt
```diff
@@ -53,7 +57,6 @@ add_library(${SERVO_LIB_NAME} SHARED
   src/servo.cpp
   src/servo_calcs.cpp
   src/collision_check.cpp
-  src/low_pass_filter.cpp
```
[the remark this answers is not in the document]
AndyZe: I'd rather keep the filter as it is. It's very lightweight and designed specifically for this task (first order Butterworth, that has the nice property of not overshooting.)

moveit_ros/moveit_servo/src/servo_calcs.cpp
AndyZe: I'm pretty sure that is correct. It was copied from an example.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -46,6 +46,10 @@
 
 static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_servo.servo_calcs");
 constexpr size_t ROS_LOG_THROTTLE_PERIOD = 30 * 1000;  // Milliseconds to throttle logs inside loops
+constexpr double SAFE_DEFAULT_VELOCITY_LIMIT =
+    2;  // rad/s, should be safe for most robots that don't have custom limits defined
```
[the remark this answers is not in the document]
AndyZe: Sure, PI/2 sounds good to me.

AndyZe:
This is working quite well now. I suggest merging and replacing the old low-pass filter implementation.

v4hn:
> > To me this variety of filter alternatives sounds very much like a plugin interface. MoveIt is good at these, so why force the use of one algorithm here?
> 
> Agree! Will work on that.

...

> This is working quite well now. I suggest merging and replacing the old low-pass filter implementation.

Was there some internal discussion on this that rejected the plugin?
I still don't support hardcoding a (relatively new) core-feature of MoveIt to a particular third-party component, especially if it does not use the same licensing.

AndyZe:
> Was there some internal discussion on this that rejected the plugin?
> I still don't support hardcoding a (relatively new) core-feature of MoveIt to a particular third-party component, especially if it does not use the same licensing.

Nope, no internal discussion. I just see the benefits of Reflexxes --

- Fine-grain control over each joint's limits
- Easier to tune than a low-pass filter (where you only specify one coefficient for everything)
- The strange backward motion I saw before is gone since using velocity mode (ignore positions)

I'm willing to write the plugin but it's not gonna happen in the next week or so.

AndyZe:
Closing in favor of #571 571

AndyZe:
@v4hn I'm finally circling back to try to implement this as a plugin. I agreed with your comment in principle but ran into some issues in practice:

> To me this variety of filter alternatives sounds very much like a plugin interface. MoveIt is good at these, so why force the use of one algorithm here?

I made the plugin base class:

> pluginlib::ClassLoader<moveit_servo::SmoothingBaseClass> smoothing_loader("moveit_servo", "moveit_servo::SmoothingBaseClass");

I create the shared pointer to my particular type of plugin that inherits from the base class:

> smoother_ = smoothing_loader.createSharedInstance("moveit_servo/LowPassFilter");

Now, the problem comes when I want to call a unique function from my child class:

>   if (SMOOTHER_PLUGIN_NAME == "moveit_servo/LowPassFilter")
>   {
>     smoother_->initialize(num_joints_, parameters_->low_pass_filter_coeff);  // This function is unique to the child class
>   }

This is important because most algorithms get initialized differently. The error is (paraphrasing):   "SmoothingBaseClass has no member named ‘initialize’"

So, how would I initialize each algorithm differently?

A couple PickNikers were suggesting `std::static_pointer_cast` and I tried `dynamic_cast`, but no luck so far.

gavanderhoorn:
Inversion of control.

Call a generic `init(..)` in your plugins and let them retrieve their parameters 'somewhere'.

PR #392, opened 2021-03-20
Add black formatter

tylerjw:
### Description

This adds the `black` python formatter to CI.  This formatter is like `clang-format` in that it will automatically reformat all code to match a pre-defined style.  This makes reading the code much nicer.  It will also enforce that all the python code is python3.

I know this is a large reformatting PR.  I will make the same PR against the ros1 moveit repo to add the same functionality there for this and the other pre-commit tests.  That way this will not impact being able to sync between the two repos.

### DIFF
--- a/moveit/scripts/create_readme_table.py
+++ b/moveit/scripts/create_readme_table.py
@@ -41,50 +41,56 @@ import requests
 
 def create_header(ros_ubuntu_dict):
     ros_distros = sorted(ros_ubuntu_dict.keys())
-    section_header = '### ROS Buildfarm\n'
-    header = 'MoveIt Package'
-    header_lines = '-'*len(header)
+    section_header = "### ROS Buildfarm\n"
+    header = "MoveIt Package"
+    header_lines = "-" * len(header)
     for ros in ros_distros:
-        source = ' '.join([ros.capitalize(), 'Source'])
-        debian = ' '.join([ros.capitalize(), 'Debian'])
-        header = ' | '.join([header, source, debian])
-        header_lines = ' | '.join([header_lines, '-'*len(source), '-'*len(debian)])
-    return '\n'.join([section_header, header, header_lines])
+        source = " ".join([ros.capitalize(), "Source"])
+        debian = " ".join([ros.capitalize(), "Debian"])
+        header = " | ".join([header, source, debian])
+        header_lines = " | ".join([header_lines, "-" * len(source), "-" * len(debian)])
+    return "\n".join([section_header, header, header_lines])
 
 
 def define_urls(target, params):
-    if target == 'src':
-        params['job'] = '{R}src_u{U}__{package}__ubuntu_{ubuntu}__source'.format(**params)
-        params['url'] = '{base_url}/view/{R}src_u{U}/job/{job}'.format(**params)
-    elif target == 'bin':
-        params['job'] = '{R}bin_u{U}64__{package}__ubuntu_{ubuntu}_amd64__binary'.format(**params)
-        params['url'] = '{base_url}/view/{R}bin_u{U}64/job/{job}'.format(**params)
+    if target == "src":
+        params["job"] = "{R}src_u{U}__{package}__ubuntu_{ubuntu}__source".format(
+            **params
+        )
+        params["url"] = "{base_url}/view/{R}src_u{U}/job/{job}".format(**params)
+    elif target == "bin":
+        params[
+            "job"
+        ] = "{R}bin_u{U}64__{package}__ubuntu_{ubuntu}_amd64__binary".format(**params)
+        params["url"] = "{base_url}/view/{R}bin_u{U}64/job/{job}".format(**params)
 
 
 def create_line(package, ros_ubuntu_dict):
     ros_distros = sorted(ros_ubuntu_dict.keys())
-    line_format = ' | [![Build Status]({base_url}/buildStatus/icon?job={job})]({url})'
-    line = '\n' + package
+    line_format = " | [![Build Status]({base_url}/buildStatus/icon?job={job})]({url})"
+    line = "\n" + package
     print(package, file=sys.stderr)
     for ros in ros_distros:
         ubuntu = ros_ubuntu_dict[ros]
         params = {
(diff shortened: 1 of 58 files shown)


moveit_commander/bin/moveit_commander_cmdline.py
```diff
@@ -136,30 +155,55 @@ def run_service(group_name):
...
+        usage="""%(prog)s [options] [<group_name>]""",
+        description="Command Line Interface to MoveIt",
+    )
+    parser.add_argument(
+        "-i",
+        "--interactive",
+        action="store_true",
+        dest="interactive",
+        default=True,
+        help="Run the command processing script in interactive mode (default)",
```
felixvd, line 190: I didn't know ending a list of arguments with an open comma like this wasn't a syntax error. I guess it's just hideous. The more you know.

There's no setting for this in black, right?

moveit_commander/test/python_time_parameterization.py
```diff
@@ -59,34 +59,52 @@ def plan(self):
...
-
     def test_plan_and_time_parameterization(self):
         plan = self.plan()
-        retimed_plan = self.time_parameterization(plan, "iterative_time_parameterization")
-        self.assertTrue(len(retimed_plan.joint_trajectory.points) > 0, "Retimed plan is invalid")
-        retimed_plan = self.time_parameterization(plan, "iterative_spline_parameterization")
-        self.assertTrue(len(retimed_plan.joint_trajectory.points) > 0, "Retimed plan is invalid")
-        retimed_plan = self.time_parameterization(plan, "time_optimal_trajectory_generation")
-        self.assertTrue(len(retimed_plan.joint_trajectory.points) > 0, "Retimed plan is invalid")
+        retimed_plan = self.time_parameterization(
```
felixvd, line 81: Kind of a lot of lines, but it's not critical.

felixvd (commented):
I just had a quick scan through this. Looks good overall though.

Could you check if there's a setting to delete the trailing comma in a list of arguments? I think that's just atrocious.

moveit_commander/bin/moveit_commander_cmdline.py
gavanderhoorn: there is a good reason for this: it avoids syntax errors when (named) arguments get shuffled around -- and the person doing that forgets to add the comma and then spends I-don't-know-how-long figuring out where the syntax error is.

moveit_ros/planning_interface/test/robot_state_update.py
```diff
@@ -33,7 +37,9 @@ def test(self):
             error_code1, plan, time = self.plan(target)
             error_code = MoveItErrorCodes()
             error_code.deserialize(error_code1)
-            if (error_code.val == MoveItErrorCodes.SUCCESS) and self.group.execute(plan):
+            if (error_code.val == MoveItErrorCodes.SUCCESS) and self.group.execute(
```
henningkayser, line 40: This seems a bit out of place

moveit/scripts/create_readme_table.py
```diff
@@ -41,50 +41,56 @@
...
             # we want to show a particular OS's badges to indicate they are not
             # released / working yet
-            if response < 400 or ubuntu == 'focal':  # success
+            if response < 400 or ubuntu == "focal":  # success
                 line += line_format.format(**params)
             else:  # error
-                line += ' | '
-                print('  {}: {} {}'.format(ros, response, params['url']), file=sys.stderr)
+                line += " | "
+                print(
```
henningkayser, line 91: also a bit odd

moveit_commander/src/moveit_commander/planning_scene_interface.py
```diff
@@ -119,7 +129,9 @@ def add_plane(self, name, pose, normal=(0, 0, 1), offset=0):
         co.plane_poses = [pose.pose]
         self.__submit(co, attach=False)
 
-    def attach_mesh(self, link, name, pose=None, filename='', size=(1, 1, 1), touch_links=[]):
+    def attach_mesh(
```
henningkayser, line 132: should we change the max time length?

moveit_ros/planning_interface/test/robot_state_update.py
```diff
@@ -6,7 +6,9 @@
 import rostest
 import os
 
-from moveit_ros_planning_interface._moveit_move_group_interface import MoveGroupInterface
+from moveit_ros_planning_interface._moveit_move_group_interface import (
```
henningkayser, line 9: not sure, this also seems like a weird rule for a single import.

henningkayser (approved):
Really nice! Comments are mostly nits, maybe the config could be tweaked a little

rhaschke:
black - by design - [_"is not configurable"_](https://black.readthedocs.io/en/stable/the_black_code_style.html). I think the only option is `--line-length`. It's configured like this:
```yaml
  - repo: https://github.com/psf/black
    rev: 20.8b1
    hooks:
      - id: black
        args: ["--line-length", "100"]
```

tylerjw:
> black - by design - [_"is not configurable"_](https://black.readthedocs.io/en/stable/the_black_code_style.html). I think the only option is `--line-length`. It's configured like this:
> 
> ```yaml
>   - repo: https://github.com/psf/black
>     rev: 20.8b1
>     hooks:
>       - id: black
>         args: ["--line-length", "100"]
> ```

@henningkayser imo we should just take black the way it is.  The line length setting included.  One thing that's great about black is if you install pre-commit locally you can just stop trying to edit the formatting of your python code and it'll do it for you and we'll all have a nice consistent style.  I honestly wish clang-format only had --style options and not a config file.

moveit_commander/bin/moveit_commander_cmdline.py
felixvd: I guess. Although when I see a comma followed by a closed bracket I look like this:

Screen Shot 2021-03-23 at 12 40 13

I also don't know how often people reorder arguments vs how often code is read, but it doesn't matter. It's the setting in black and I don't have strong objections.

felixvd (approved):
LGTM even with the idiosyncrasies.

PR #396, opened 2021-03-25
Migrate from Travis to GitHub Actions

tylerjw:
### Description

This is finally here.  There two features of this that are temporary.  First `clang-tidy` filtering is implemented in my PR to industrial_ci here: https://github.com/ros-industrial/industrial_ci/pull/649.  I'd appreciate any reviews of that to get it merged and cleaned up.  But until that happens we can use my fork.  

Secondly, preparing the code coverage report is being done by a script.  Eventually after [this PR on industrial_ci](https://github.com/ros-industrial/industrial_ci/pull/504) is merged we won't need that script.

Lastly, this PR is here on the repo itself (and not from my fork) so the Actions jobs will run here.

Thank you @rhaschke for helping me get this here.  I'll make a copy of this over on moveit1 also.

Also, this disables a test in moveit_servo.  That test has become flaky recently and I don't know why.  I'm disabling it here so it doesn't get in the way.  I will work on re-enabling and debugging it in the future.


.github/workflows/industrial_ci_action.yml
```diff
@@ -0,0 +1,95 @@
...
+    strategy:
+      matrix:
+        env:
+          - IMAGE: 'foxy-ci'
+            CCOV: true
+            AFTER_RUN_TARGET_TEST: './.ci.prepare_codecov'
+          - IMAGE: 'foxy-ci-testing'
+            BEFORE_TARGET_TEST_EMBED: 'set +u && source moveit_kinematics/test/test_ikfast_plugins.sh && set -u'
+            CLANG_TIDY: true
+            CLANG_TIDY_BASE_REF: ${{ github.base_ref || github.ref }}
```
rhaschke, line 24: Please move this line down to the "computed" env section.

.github/workflows/industrial_ci_action.yml
```diff
@@ -0,0 +1,95 @@
...
+    env:
+      DOCKER_IMAGE: moveit/moveit2:${{ matrix.env.IMAGE }}
+      UPSTREAM_WORKSPACE: moveit2.repos
+      CCOV: ${{ matrix.env.CCOV }}
+      AFTER_SETUP_UPSTREAM_WORKSPACE: 'vcs pull $BASEDIR/upstream_ws/src'
```
rhaschke, line 29: Wasn't there a `UNDERLAY_WS` feature?

.github/workflows/industrial_ci_action.yml
```diff
@@ -0,0 +1,95 @@
...
+
+    name: "${{ matrix.env.IMAGE }}${{ matrix.env.CCOV && ' + ccov' || ''}}${{ matrix.env.CLANG_TIDY && ' + clang-tidy' || '' }}"
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Free up disk space"
+        if: ${{ env.CCOV }}
+        run: |
+          sudo apt-get -qq purge build-essential ghc*
+          sudo apt-get clean
+          docker system prune -af
```
rhaschke, line 45: Add some comments:
```suggestion
          # cleanup docker images not used by us
          docker system prune -af
```

.github/workflows/industrial_ci_action.yml
```diff
@@ -0,0 +1,95 @@
...
+          sudo mkdir /mnt/ccache
```
rhaschke, line 46:
```suggestion
          # shift ccache folder to /mnt on a separate disk
          sudo mkdir /mnt/ccache
```

.github/workflows/industrial_ci_action.yml
```diff
@@ -0,0 +1,95 @@
...
+          mkdir ${{ env.CCACHE_DIR }}
+          sudo mount --bind ${{ env.CCACHE_DIR }} /mnt/ccache
+          sudo rm -rf /usr/local/*
```
rhaschke, line 49:
```suggestion
          # free up a lot of stuff from /usr/local
          sudo rm -rf /usr/local/*
```

.github/workflows/industrial_ci_action.yml
```diff
@@ -0,0 +1,95 @@
...
+        with:
+          path: ${{ env.BASEDIR }}/upstream_ws
+          key: upstream_ws-${{ env.CACHE_PREFIX }}-${{ hashFiles('moveit2.repos') }}-${{ github.run_id }}
+          restore-keys: |
+            upstream_ws-${{ env.CACHE_PREFIX }}-${{ hashFiles('moveit2.repos') }}
+      - name: cache target_ws
+        if: ${{ ! env.CCOV }}
+        uses: pat-s/always-upload-cache@v2.1.3
+        with:
+          path: ${{ env.BASEDIR }}/target_ws
```
rhaschke, line 63: You don't want to restore the src folder from cache. Otherwise you overwrite what you checked out, don't you?
Or is the checked out source symlinked into the source folder? If so, maybe add a comment here, to clarify this for our future selves.

moveit2.repos
```diff
@@ -1,8 +1,4 @@
 repositories:
-  moveit2:
-    type: git
-    url: https://github.com/ros-planning/moveit2
-    version: main
```
rhaschke, line 5: I see your point of only listing upstream stuff here. But, removing the main source will probably invalidate our build-from-source instructions, which essentially reads like this: "Fetch moveit2.repos and build." They should be adapted then.

rhaschke (commented):
Thanks for this great work. I have some open questions, though.

.github/workflows/industrial_ci_action.yml
tylerjw: That is for an underlay that is included in the docker image.  That works with the source images.  This is for the upstream workspace being cached.

moveit2.repos
tylerjw: I'll prepare a PR for that.  This being here breaks the CI system because it creates a vcs error when it tries to import this file and moveit2 is already there.

tylerjw: https://github.com/ros-planning/moveit.ros.org/pull/577

.github/workflows/industrial_ci_action.yml
```diff
@@ -0,0 +1,100 @@
...
+      - main
+
+jobs:
+  industrial_ci:
+    strategy:
+      matrix:
+        env:
+          - IMAGE: 'foxy-ci'
+            CCOV: true
+            AFTER_RUN_TARGET_TEST: './.ci.prepare_codecov'
```
rhaschke, line 20: Same as in MoveIt 1

.github/workflows/industrial_ci_action.yml
```diff
@@ -0,0 +1,100 @@
...
+      CCACHE_DIR: "${{ github.workspace }}/.ccache"
+      BASEDIR: ${{ github.workspace }}/.work
+      CACHE_PREFIX: "${{ matrix.env.IMAGE }}${{ matrix.env.CCOV && '-ccov' || '' }}"
+      CLANG_TIDY_BASE_REF: "${{ github.base_ref || github.ref }}"
+
+    name: "${{ matrix.env.IMAGE }}${{ matrix.env.CCOV && ' + ccov' || ''}}${{ matrix.env.CLANG_TIDY && ' + clang-tidy' || '' }}"
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Free up disk space"
+        if: ${{ env.CCOV }}
```
rhaschke, line 41:
```suggestion
        if: ${{ matrix.env.CCOV }}
```

.github/workflows/industrial_ci_action.yml
```diff
@@ -0,0 +1,100 @@
...
+        uses: pat-s/always-upload-cache@v2.1.3
+        with:
+          path: ${{ env.BASEDIR }}/upstream_ws
+          key: upstream_ws-${{ env.CACHE_PREFIX }}-${{ hashFiles('moveit2.repos') }}-${{ github.run_id }}
+          restore-keys: |
+            upstream_ws-${{ env.CACHE_PREFIX }}-${{ hashFiles('moveit2.repos') }}
+      # The target directory cache doesn't include the source directory because
+      # that comes from the checkout.  See "prepare target_ws for cache" task below
+      - name: cache target_ws
+        if: ${{ ! env.CCOV }}
```
rhaschke, line 65:
```suggestion
        if: ${{ ! matrix.env.CCOV }}
```

.github/workflows/industrial_ci_action.yml
```diff
@@ -0,0 +1,100 @@
...
+          - IMAGE: 'foxy-ci'
+            CCOV: true
+            AFTER_RUN_TARGET_TEST: './.ci.prepare_codecov'
+          - IMAGE: 'foxy-ci-testing'
+            BEFORE_TARGET_TEST_EMBED: 'set +u && source moveit_kinematics/test/test_ikfast_plugins.sh && set -u'
+            CLANG_TIDY: true
+    env:
+      DOCKER_IMAGE: moveit/moveit2:${{ matrix.env.IMAGE }}
+      UPSTREAM_WORKSPACE: moveit2.repos
+      CCOV: ${{ matrix.env.CCOV }}
```
rhaschke, line 27: This is confusing. Just refer to matrix.env.CCOV everywhere.
```suggestion
```

.github/workflows/industrial_ci_action.yml
```diff
@@ -0,0 +1,100 @@
...
+        with:
+          name: test-results
+          path: ${{ env.BASEDIR }}/target_ws/**/test_results/**/*.xml
+      - name: upload codecov report
+        uses: codecov/codecov-action@v1
+        if: ${{ env.CCOV }}
+        with:
+          files: ${{ env.BASEDIR }}/coverage.info
+      - name: prepare target_ws for cache
+        if: ${{ always() && ! env.CCOV }}
```
rhaschke, line 95: What's the meaning of `always()` here? Seems to be redundant.

.github/workflows/industrial_ci_action.yml
rhaschke: Can you comment on this, please?

rhaschke (commented):
More comments.

.github/workflows/industrial_ci_action.yml
tylerjw: Always causes this to run even when previous steps failed (such as tests in industrial_ci).

.github/workflows/industrial_ci_action.yml
tylerjw: It isn't always defined, which causes the if statements to fail in the steps.  I could set it to false for every not codecov job then I wouldn't need this.

tylerjw: That is what I remembered why I did that, I pushed this change though and it seems to be working so far.

.github/workflows/industrial_ci_action.yml
tylerjw: I left a comment on line 62-63

.github/workflows/industrial_ci_action.yml
tylerjw: It looks like it worked, this change is now in.

tylerjw:
This one has been updated per feedback and is ready for another review @rhaschke.

tylerjw:
Rebased and pushed as there was a different change merged.

.github/workflows/industrial_ci_action.yml
```diff
@@ -0,0 +1,100 @@
...
+    strategy:
+      matrix:
+        env:
+          - IMAGE: 'foxy-ci'
+            CCOV: true
+          - IMAGE: 'foxy-ci-testing'
+            IKFAST_TEST: true
+            CLANG_TIDY: true
+    env:
+      DOCKER_IMAGE: moveit/moveit2:${{ matrix.env.IMAGE }}
```
henningkayser, line 24: nice

.github/workflows/industrial_ci_action.yml
```diff
@@ -0,0 +1,100 @@
...
+      - name: cache ccache
+        uses: pat-s/always-upload-cache@v2.1.3
+        with:
+          path: ${{ env.CCACHE_DIR }}
+          key: ccache-${{ env.CACHE_PREFIX }}-${{ github.sha }}-${{ github.run_id }}
+          restore-keys: |
+            ccache-${{ env.CACHE_PREFIX }}-${{ github.sha }}
+            ccache-${{ env.CACHE_PREFIX }}
+      - name: industrial_ci
+        uses: 'tylerjw/industrial_ci@clang-tidy-modified-filter'
```
henningkayser, line 81: Better keep a note/issue to update this

moveit2.repos
henningkayser: What about `vcs import --skip-existing`?

.github/workflows/format.yml
```diff
@@ -7,6 +7,8 @@ on:
   workflow_dispatch:
   pull_request:
   push:
+    branches:
+      - main
```
henningkayser, line 11: Should this include other branches as well, for instance `feature/hybrid-planning` after that one has been rebased?

henningkayser (approved):
> Rebased and pushed as there was a different change merged.

Sorry ;)

.github/workflows/format.yml
tylerjw: I didn't know there were other branches.  I'll add that.

tylerjw:
@henningkayser 

> What about vcs import --skip-existing?

no, because I'm not calling vcs import, industrial_ci is.

.github/workflows/industrial_ci_action.yml
tylerjw: Added a comment to link to the open PR.  I'm hopeful this won't take long to get merged and when it does it'll be trivial to update here to point back to industrial_ci

tylerjw:
> Sorry ;)
> 
> ... PoseTracking is also flaky, you might need to disable that one as well.

I disabled it too for now.  I'll create issues for the flaky tests in servo.

tylerjw:
@henningkayser The newest failure seems maybe related to one of the things you merged `test_launch_move_group_ompl_constraints`

Do you think that caused it?  Any ideas on how to fix this?

tylerjw:
I pushed changes suggested by @JafarAbdi.  Hopefully this fixes the newly broken test.

tylerjw:
It is still failing, @henningkayser @JafarAbdi any ideas?

JafarAbdi:
@tylerjw Trajectory execution timeouts, can you change the following values in this [file](https://github.com/ros-planning/moveit2/blob/main/moveit_ros/planning_interface/test/launch/move_group_launch_test_common.py#L96)

```
"trajectory_execution.allowed_execution_duration_scaling": 1.5,
"trajectory_execution.allowed_goal_duration_margin": 1.0,
"trajectory_execution.allowed_start_tolerance": 0.01,
```

rhaschke (approved):

moveit2.repos
JafarAbdi: This caused the docker images to not have moveit2 source code, I think we should revert and apply @henningkayser suggestion, @rhaschke @tylerjw what do you think .?

rhaschke: As @tylerjw already pointed out, `vcs import` is performed by `industrial_ci`. What about adding the MoveIt source explicitly in the docker image?

tylerjw: Because the docker build is running from the repo you should be able to use the docker COPY command to get the moveit repo in docker.  See this line in the nav2 docker file: https://github.com/ros-planning/navigation2/blob/e13459441dd4895450b2f2073f73bf340823fcfe/Dockerfile#L26

PR #398, opened 2021-03-26
Fix build directory for foxy-source image

JafarAbdi:
### Description

This's a bug introduced by my previous PR, I forgot to change the directory before building causing the `install`, `build`, and `log` folders to be inside `src`

### Checklist

### DIFF
--- a/.docker/source/Dockerfile
+++ b/.docker/source/Dockerfile
@@ -27,7 +27,7 @@ RUN \
     rm -rf /var/lib/apt/lists/*
 
 # Build the workspace
-RUN . /opt/ros/${ROS_DISTRO}/setup.sh &&\
+RUN cd .. && . /opt/ros/${ROS_DISTRO}/setup.sh &&\
     colcon build \
             --cmake-args -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTING=OFF -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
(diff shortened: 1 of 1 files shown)


.docker/source/Dockerfile
```diff
@@ -27,7 +27,7 @@ RUN \
     rm -rf /var/lib/apt/lists/*
 
 # Build the workspace
-RUN . /opt/ros/${ROS_DISTRO}/setup.sh &&\
+RUN cd .. && . /opt/ros/${ROS_DISTRO}/setup.sh &&\
```
tylerjw, line 30: To make it clearer where you are changing directory to, could you use the variable ROS_DISTRO?

tylerjw (approved):

.docker/source/Dockerfile
JafarAbdi: There's no need to use `ROS_DISTRO`, the previous current directory is `/root/ws_moveit/src`

tylerjw: What I meant was to use `ROS_UNDERLAY`:

`cd $ROS_UNDERLAY/..`

I'm sorry, not ROS_DISTRO.

JafarAbdi, line 30:
```suggestion
RUN cd $ROS_UNDERLAY/.. && . /opt/ros/${ROS_DISTRO}/setup.sh &&\
```

tylerjw (approved):
:+1:

PR #402, opened 2021-03-29
[Servo] Halt Servo command on Pose Tracking stop (#2501)

AndyZe:
This is a port of this MoveIt1 PR: https://github.com/ros-planning/moveit/pull/2501

I figure it is important to port because it has some safety ramifications.


tylerjw (approved):
Does this mean that it sends an all zero `TwistStamped` to the output?  Could this cause issues if the robot was being controlled by a different controller or is that not an issue?

henningkayser (commented):
Doesn't this allow `stopMotion()` to complete before `stop_requested_` has been read in the main control loop? That would lead to a race condition in the order of commands published to the controller. Better call the `stopMotion()` from the main loop to make sure it's always published last.

AndyZe:
> Doesn't this allow `stopMotion()` to complete before `stop_requested_` has been read in the main control loop? ...

Thanks, that's a good point

AndyZe:
Actually that was already covered. Here from the main loop:

```
    if (stop_requested_)
    {
      RCLCPP_INFO_STREAM(LOGGER, "Halting servo motion, a stop was requested.");
      doPostMotionReset();  --> Calls **stopMotion()**, among other things
      return PoseTrackingStatusCode::STOP_REQUESTED;
    }
```

moveit_ros/moveit_servo/src/pose_tracking.cpp
```diff
@@ -300,8 +300,23 @@ geometry_msgs::msg::TwistStamped::ConstSharedPtr PoseTracking::calculateTwistCom
...
+  stop_requested_ = true;
+
+  // Send a 0 command to Servo to halt arm motion
+  auto msg = moveit::util::make_shared_from_pool<geometry_msgs::msg::TwistStamped>();
+  {
+    std::lock_guard<std::mutex> lock(target_pose_mtx_);
+    msg->header.frame_id = target_pose_.header.frame_id;
+  }
+  msg->header.stamp = node_->now();
+  twist_stamped_pub_->publish(*msg);
```
vatanaksoytezer, line 314: @AndyZe maybe we should change this to a regular message or a unique_ptr as discussed [here](https://github.com/ros-planning/moveit2/pull/428#discussion_r618461474)

vatanaksoytezer (commented):
I've read through the code and  left a tiny comment regarding `make_shared_from_pool`

sjahr (commented):
Would it make sense to add a test that checks whether the last published command to Servo is always 0 when the motion is supposed to finish?

moveit_ros/moveit_servo/src/pose_tracking.cpp
AndyZe: Done!

henningkayser:
@AndyZe same as with https://github.com/ros-planning/moveit2/pull/483, the corresponding PR has been synced into main. Please verify it's all there what you need. If there are additional changes required, they can be added with this PR or in another follow-up.

vatanaksoytezer:
I'm closing this PR since, the functionaity is merged with the last sync per Henning's last comment. Feel free to re-open this or open a new one if needed

PR #408, opened 2021-04-01
Declare joint limit parameters

bostoncleek:
Joint position, velocity, and accleration parameter limits
where not declared in the `robot_model_loader`. This prevented
the use of joint limits. 

### Checklist

### DIFF
--- a/moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
+++ b/moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
@@ -139,7 +139,7 @@ void RobotModelLoader::configure(const Options& opt)
           bool has_position_limits = false;
           if (!node_->has_parameter(param_name))
           {
-            node_->declare_parameter(param_name, false);
+            node_->declare_parameter(param_name);
           }
           node_->get_parameter(param_name, has_position_limits);
 
@@ -147,7 +147,7 @@ void RobotModelLoader::configure(const Options& opt)
           bool has_vel_limits = false;
           if (!node_->has_parameter(param_name))
           {
-            node_->declare_parameter(param_name, false);
+            node_->declare_parameter(param_name);
           }
           node_->get_parameter(param_name, has_vel_limits);
 
@@ -155,7 +155,7 @@ void RobotModelLoader::configure(const Options& opt)
           bool has_acc_limits = false;
           if (!node_->has_parameter(param_name))
           {
-            node_->declare_parameter(param_name, false);
+            node_->declare_parameter(param_name);
           }
           node_->get_parameter(param_name, has_acc_limits);
 
@@ -165,19 +165,24 @@ void RobotModelLoader::configure(const Options& opt)
 
           if (has_position_limits && canSpecifyPosition(joint_model, joint_id))
           {
-            param_name = prefix + "max_position";
-            if (!node_->has_parameter(param_name))
+            const auto max_pos_name = prefix + "max_position";
+            if (!node_->has_parameter(max_pos_name))
             {
-              node_->declare_parameter(param_name, 0.0);
+              node_->declare_parameter(max_pos_name);
             }
-            node_->get_parameter(param_name, joint_limit[joint_id].max_position);
 
-            param_name = prefix + "min_position";
-            if (!node_->has_parameter(param_name))
+            const auto min_pos_name = prefix + "min_position";
+            if (!node_->has_parameter(min_pos_name))
             {
-              node_->declare_parameter(param_name, 0.0);
+              node_->declare_parameter(min_pos_name);
             }
-            node_->get_parameter(param_name, joint_limit[joint_id].min_position);
+
+            if (!node_->get_parameter(max_pos_name, joint_limit[joint_id].max_position) ||
+                !node_->get_parameter(min_pos_name, joint_limit[joint_id].min_position))
+            {
+              RCLCPP_ERROR(LOGGER, "Specified a position limit for joint: %s but did not set a max or min position",
+                           joint_limit[joint_id].joint_name.c_str());
+            }
 
             joint_limit[joint_id].has_position_limits = has_position_limits;
           }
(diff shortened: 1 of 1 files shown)


moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
```diff
@@ -154,30 +154,30 @@ void RobotModelLoader::configure(const Options& opt)
...
             joint_limit[joint_id].max_velocity = max_velocity;
           }
 
           param_name = prefix + "has_velocity_limits";
-          bool has_vel_limits;
+          bool has_vel_limits = node_->declare_parameter(param_name, false);
           if (node_->get_parameter(param_name, has_vel_limits))
           {
             joint_limit[joint_id].has_velocity_limits = has_vel_limits;
           }
```
AndyZe, line 169: This was a previously-existing issue, not your code, but maybe now is a good time to fix it.

joint_limit[joint_id].has_velocity_limits is already set at L160 so I think you could delete this block

moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
```diff
@@ -154,30 +154,30 @@ void RobotModelLoader::configure(const Options& opt)
...
           {
             joint_limit[joint_id].has_acceleration_limits = true;
             joint_limit[joint_id].max_acceleration = max_acc;
           }
 
           param_name = prefix + "has_acceleration_limits";
-          bool has_acc_limits;
+          bool has_acc_limits = node_->declare_parameter(param_name, false);
           if (node_->get_parameter(param_name, has_acc_limits))
             joint_limit[joint_id].has_acceleration_limits = has_acc_limits;
```
AndyZe, line 182: I think you could delete this block of code, too

AndyZe (commented):
Generally looks good! I'm not very familiar with ROS2 so somebody else should def review this.

moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
bostoncleek: If it does not have limits should we set these to default values?

AndyZe (approved):

moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
AndyZe: I think not. Hopefully @JafarAbdi or @henningkayser  can confirm

moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
```diff
@@ -131,55 +131,69 @@ void RobotModelLoader::configure(const Options& opt)
         std::string param_name;
         try
         {
-          param_name = prefix + "max_position";
-          double max_position;
-          if (node_->get_parameter(param_name, max_position))
+          param_name = prefix + "has_position_limits";
```
destogl, line 134: If prefix ends with "." than this should work.

destogl (approved):
This looks good. Just a small comment to check. 

Also adding some tests would be nice.

moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
```diff
@@ -131,55 +131,69 @@ void RobotModelLoader::configure(const Options& opt)
...
+          bool has_position_limits = node_->declare_parameter(param_name, false);
+          if (node_->get_parameter(param_name, has_position_limits))
```
JafarAbdi, line 136: This will always return true since we're declaring it with a default value, also we need to make sure that a parameter wasn't declared before otherwise, it will throw an exception

Same for the other parameters

moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
bostoncleek: It ends with a "." see `        std::string prefix =
            rdf_loader_->getRobotDescription() + "_planning.joint_limits." + joint_limit[joint_id].joint_name + ".";`

JafarAbdi (commented):
LGTM

But I'm not sure if we should explicitly specify the default values (0.0 & false) when declaring the parameters, 

Currently, if the user didn't load the joint limit parameters and in another function, they called `get_parameter` they will get 0.0/false rather than `PARAMETER_NOT_SET` which may not be expected

@AndyZe @destogl @henningkayser what do you think .?

moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
```diff
@@ -131,89 +131,78 @@ void RobotModelLoader::configure(const Options& opt)
...
 
             if (!node_->has_parameter(param_name))
             {
-              max_acc = node_->declare_parameter(param_name, 0.0);
+              node_->declare_parameter(param_name, 0.0);
             }
 
-            node_->get_parameter(param_name, max_acc);
-            joint_limit[joint_id].max_acceleration = max_acc;
+            node_->get_parameter(param_name, joint_limit[joint_id].max_acceleration);
```
AndyZe, line 205: Sorry to nitpick but what if the user puts:

`has_acceleration_limit = true;`

But does not include `max_acceleration` in the yaml file?

I think we should print an error message:

`RCLCP_ERROR(logger, "has_acceleration_limits was set, but max_acceleration was not");`

Or just delete `has_vel/accel_limits` altogether?

AndyZe:
> Currently, if the user didn't load the joint limit parameters and in another function, they called `get_parameter` for one of the joint limit parameters they will get 0.0/false rather than `PARAMETER_NOT_SET` which may not be expected

I would rather not set defaults

bostoncleek:
> > Currently, if the user didn't load the joint limit parameters and in another function, they called `get_parameter` for one of the joint limit parameters they will get 0.0/false rather than `PARAMETER_NOT_SET` which may not be expected
> 
> I would rather not set defaults

That is a good point. Will update to remove the default parameters.

tylerjw (approved):
lgtm, would it be difficult to add a test of this?  Just have a test that sets the parameters using the c++ ros parameter interface or via loading a yaml and checks that those are parsed correctly by this code?

bostoncleek:
> lgtm, would it be difficult to add a test of this? Just have a test that sets the parameters using the c++ ros parameter interface or via loading a yaml and checks that those are parsed correctly by this code?

Yes I will certainly add a test.

AndyZe:
3x +1's --> let's get this merged and make an issue to write a test

PR #410, opened 2021-04-05
Servo: do not cause a joint jump when SuddenHalt() is called

AndyZe:
We found out the hard way, when the suddenHalt() function was called for a robot in velocity control mode, it would send a command vector of all-zero positions. That could cause a really big joint jump depending what type of ros2_controller is active. This fixes that.

Also (really minor fix) it sets trajectory accelerations to 0 to match the comment.


tylerjw (approved):
ltm

PR #419, opened 2021-04-13
Unify PickNik name in copyrights

tylerjw:
### Description

Various copyrights mention `PickNik Robotics` (these are my fault), `PickNik LLC` (from moveit1, we should fix these there too), `PickNik Consulting` (only one of these).  There is a legal requirement that these are a legal name.  This fixes that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ This open source project is maintained by supporters from around the world — s
   <img src="https://picknik.ai/assets/images/logo.jpg" width="168">
 </a>
 
-[PickNik Robotics](https://picknik.ai/) is leading and organizing the development of MoveIt 2.
+[PickNik Inc.](https://picknik.ai/) is leading and organizing the development of MoveIt 2.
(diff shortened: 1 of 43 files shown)


JafarAbdi (approved):

tylerjw:
moveit_servo tests continue to be flaky, re-running jobs.  We should probably fix the flakyness of those before trying to release for foxy or it will probably plague the official buildfarm.

nbbrooks:
To be completely safe, I think we should use "PickNik, Inc."

With this search
gr "PickNik" | grep -v "PickNik, Inc\." | grep -v "\.ai"

I picked up a few things that should be resolved with
findreplace "PickNik, LLC" "PickNik, Inc."
findreplace "PickNik nor" "PickNik, Inc. nor"
findreplace "PickNik Inc" "PickNik, Inc"

nbbrooks (approved):
lgtm thanks!

Issue #417, opened 2021-04-13
Moveit2 should use move_group namespace


vatanaksoytezer:
### Description

move_group namespace is not used in moveit2 but rviz2 defaults to it when opening up MotionPlanning.

### Your environment
* ROS Distro: Foxy
* Ubuntu 20.04
* Source Build from master

### Steps to reproduce
Launch an empty rviz instance and run_move_group.demo.launch.py. Then add MotionPlanning to the rviz

### Expected behaviour
MotionPlanning topics should be published including move_group namespace.
### Actual behaviour
MotionPlanning topics does not currently include move_group, topics are published without move_group namespace.

JafarAbdi:
Closed by https://github.com/ros-planning/moveit2/pull/420

PR #420, opened 2021-04-13
Remove move_group namespace

vatanaksoytezer:
Related to #417, @JafarAbdi and I discussed and looked for solutions. We tried to add ~/ to the nodes though these result in a namespace of move_group_private which we thought would have to change and break quite a bit of stuff. So we thought to leave it the way it is right now and removed move_group namespaces from the relevant files. @henningkayser can you check if it makes sense, or we still want the private node stuff?

If you approve I will send a PR to moveit2_tutorials and moveit_resources related to this. I've tested this with run_move_group, run_moveit_cpp and quickstart rviz (didn't send a PR for this one yet) all seem to work fine.

### DIFF
--- a/moveit_demo_nodes/run_move_group/launch/run_move_group.rviz
+++ b/moveit_demo_nodes/run_move_group/launch/run_move_group.rviz
@@ -310,7 +310,7 @@ Visualization Manager:
         Show Trail: false
         State Display Time: 0.05 s
         Trail Step Size: 1
-        Trajectory Topic: /move_group/display_planned_path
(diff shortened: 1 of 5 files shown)


henningkayser (approved):
This works for now to prevent irritations. We can change it back when re-enabling proper namespaces in a follow-up.

vatanaksoytezer:
> This works for now to prevent irritations. We can change it back when re-enabling proper namespaces in a follow-up.

That sounds great! Thanks for checking up.

PR #423, opened 2021-04-15
Re-re-enable servo tests.

vatanaksoytezer:
Ongoing attempt to see what is going with the tests and fixing it

### DIFF
--- a/moveit_ros/moveit_servo/test/unit_test_servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/test/unit_test_servo_calcs.cpp
@@ -505,8 +505,8 @@ TEST_F(ServoCalcsTestFixture, TestComposeOutputMsg)
 
 int main(int argc, char** argv)
 {
-  ::testing::InitGoogleTest(&argc, argv);
   rclcpp::init(argc, argv);
+  ::testing::InitGoogleTest(&argc, argv);
 
   // Set up the shared node
   TEST_NODE = std::make_shared<rclcpp::Node>("servo_calcs_test");
--- a/moveit_ros/moveit_servo/test/unit_test_servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/test/unit_test_servo_calcs.cpp
@@ -360,17 +360,17 @@ TEST_F(ServoCalcsTestFixture, TestEnforceVelLimits)
   for (size_t i = 0; i < 7; ++i)
   {
     // We need to check vs radians-per-loop allowable rate (not rad/s)
-    EXPECT_GE(desired_velocity[i], vel_limits[i] * servo_calcs_->parameters_->publish_period);
+    EXPECT_LE(desired_velocity[i], vel_limits[i] * servo_calcs_->parameters_->publish_period);
   }
 
-  // Let's check the negative velocities too
+  // Let's check negative velocity limits too
   desired_velocity *= -1;
   servo_calcs_->prev_joint_velocity_ = desired_velocity;
   servo_calcs_->enforceVelLimits(desired_velocity);
   for (size_t i = 0; i < 7; ++i)
   {
     // We need to check vs radians-per-loop allowable rate (not rad/s)
-    EXPECT_LE(desired_velocity[i], -1 * vel_limits[i] * servo_calcs_->parameters_->publish_period);
+    EXPECT_GE(desired_velocity[i], -1 * vel_limits[i] * servo_calcs_->parameters_->publish_period);
   }
 }
 
--- a/moveit_ros/moveit_servo/test/unit_test_servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/test/unit_test_servo_calcs.cpp
@@ -360,7 +360,8 @@ TEST_F(ServoCalcsTestFixture, TestEnforceVelLimits)
   for (size_t i = 0; i < 7; ++i)
   {
     // We need to check vs radians-per-loop allowable rate (not rad/s)
-    EXPECT_LE(desired_velocity[i], vel_limits[i] * servo_calcs_->parameters_->publish_period);
+    // was EXPECT_LE
+    EXPECT_GE(desired_velocity[i], vel_limits[i] * servo_calcs_->parameters_->publish_period);
   }
 
   // Let's check the negative velocities too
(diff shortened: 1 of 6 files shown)


moveit_ros/moveit_servo/test/unit_test_servo_calcs.cpp
```diff
@@ -360,7 +360,7 @@ TEST_F(ServoCalcsTestFixture, TestEnforceVelLimits)
   for (size_t i = 0; i < 7; ++i)
   {
     // We need to check vs radians-per-loop allowable rate (not rad/s)
-    EXPECT_LE(desired_velocity[i], vel_limits[i] * servo_calcs_->parameters_->publish_period);
+    EXPECT_GE(desired_velocity[i], vel_limits[i] * servo_calcs_->parameters_->publish_period);
```
henningkayser, line 363: I don't think this test condition should be flipped. After calling `enforceVelLimits()` the desired velocity should be within the limits.

vatanaksoytezer: I see, I thought it was vice versa, but now makes sense. So probably this means we have something wrong in enforceVelLimits?

henningkayser: Yes, seems like it. Did you compare the values? Maybe it's just a rounding error and we need to add an epsilon

vatanaksoytezer: 0.102 vs 0.081345 for positive
-0.102 vs -0.097614 for negative

AndyZe: It would be a good idea to port this MoveIt1 PR that really simplified the limit enforcement before digging too much farther into this

https://github.com/ros-planning/moveit/pull/2471

vatanaksoytezer: @AndyZe I can port it quickly if you want seem pretty straightforward to port. Do you want me to include basic functionality tests and remove (/ rename) low latency tests?

AndyZe: Some of the existing tests seem just fine to me, including:

TestEnforcePosLimits
TestEnforceVelLimits

Can you link me to exactly what tests you might change?

AndyZe: :+1: to porting that PR

vatanaksoytezer: I meant the naming from the PR, it seems you guys have changed the naming from low_latency to basic_servo_tests

AndyZe: I see. I think `basic_servo_tests` is definitely a better, more descriptive name.

vatanaksoytezer: Tested in my local with the one on ros-planning/moveit#2471 though it seems to give the exact same results for `enforceVelocityLimits()`

vatanaksoytezer:
After debugging with @JafarAbdi we traced the bug (for the `enforceVelLimits` function) to appear after #408. We don't exactly now what is causing this behaviour but we might want to check it out.

vatanaksoytezer:
Also flakiness seems to be something from run_test.py of colcon not being able to verify the result of the tests. Test itself passes but verification fails so it seems to be failing

AndyZe:
I just started noticing warnings like this. Might be relevant here --

`parameter 'moveit_servo.lower_singularity_threshold' has invalid type: expected [double] got [integer]`

AndyZe:
There's a change this moveit_resources PR will fix the issue:  https://github.com/ros-planning/moveit_resources/pull/63

AndyZe:
@vatanaksoytezer I think it would be really great if you started a new PR for your port of joint_limits handling (porting this [MoveIt1 PR](https://github.com/ros-planning/moveit/pull/2471))

vatanaksoytezer:
@AndyZe I will open up a PR for this tomorrow (monday), I was working on porting moveit2_tutorials to github actions, and didn't had time to complete this on Friday.

moveit_ros/moveit_servo/test/unit_test_servo_calcs.cpp
```diff
@@ -370,7 +370,7 @@ TEST_F(ServoCalcsTestFixture, TestEnforceVelLimits)
   for (size_t i = 0; i < 7; ++i)
   {
     // We need to check vs radians-per-loop allowable rate (not rad/s)
-    EXPECT_GE(desired_velocity[i], -1 * vel_limits[i] * servo_calcs_->parameters_->publish_period);
+    EXPECT_LE(desired_velocity[i], -1 * vel_limits[i] * servo_calcs_->parameters_->publish_period);
```
AndyZe, line 373: This should be "greater than" because we're dealing with negative numbers

AndyZe, line 373: This should be "greater than" because we're dealing with negative numbers

AndyZe (changes requested):

moveit_ros/moveit_servo/test/unit_test_servo_calcs.cpp
vatanaksoytezer: Yeah, you are right I thought maybe the tests were wrong since the outcome was vice versa but Henning helped me figure out. I'll change it back

AndyZe: I went ahead and reverted the changes to this file so we can see if CI passes reliably with moveit_resources 63

vatanaksoytezer: Sure, thanks a lot

vatanaksoytezer:
@AndyZe @tylerjw @JafarAbdi I think I found the source of the flakiness. It seems we are starting google test before, we are satarting the ros nodes, which causes comm problems sometimes and creates flakiness. I've observed this in three tests (singularity, collision and interface) which were the ones that are flaky. For eg. https://github.com/ros-planning/moveit2/blob/main/moveit_ros/moveit_servo/test/test_servo_interface.cpp#L225-L226 here, order of the lines 225 and 226 should be vice versa. See if this makes sense, and I am including these in this pr

vatanaksoytezer:
I also disabled collision test for now since @JafarAbdi and I traced it back to #408 .

AndyZe (approved):

JafarAbdi (approved):

vatanaksoytezer:
https://github.com/ros-planning/moveit2/actions/runs/764040882 seems to fail again :/

PR #303, opened 2020-11-09
WIP: Enable moveit_kinematics launch tests

sjahr:
### Description

Continued working on #212  to enable moveit_kinematics launch tests with ROS 2's [launch_testing](https://github.com/ros2/launch/tree/master/launch_testing).

### Checklist
- [ ] **Required by CI**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)
- [X] Fix test_kinematics_plugin
- [ ] Port fanuc-kdl.test
- [ ] Port fanuc-ikfast.test
- [ ] Port panda-kdl.test
- [ ] Port panda-ikfast.test

To build and run the tests:
1. Build package with testing enabled
```
colcon build --packages-select moveit_kinematics --cmake-args -DCMAKE_BUILD_TYPE=Release -BUILD_TESTING=true
```

2. Run the tests:
```
colcon test --packages-select moveit_kinematics
```

3. Show results
```
colcon test-result --all --verbose
```

PR #435, opened 2021-04-22
Enable LMA and KDL kinematic launch tests

vatanaksoytezer:
An attempt to reproduce and solve problems on enabling kinematic launch tests. Continued work on #303 .


griswaldbrooks (commented):
Does https://github.com/ros/resource_retriever/blob/kinetic-devel/package.xml need to be added to the `package.xml`?

moveit_kinematics/test/test_kinematics_plugin.cpp
```diff
@@ -98,58 +94,41 @@ class SharedData
...
+    ASSERT_TRUE(timeout_ > 0.0) << "'ik_timeout' must be more than 0.0 seconds";
+    node_->get_parameter_or("tolerance", tolerance_, tolerance_);
+    ASSERT_TRUE(tolerance_ > 0.0) << "'tolerance' must be greater than 0.0";
+    node_->get_parameter_or("num_fk_tests", num_fk_tests_, 0);
+    node_->get_parameter_or("num_ik_cb_tests", num_ik_cb_tests_, 0);
+    node_->get_parameter_or("num_ik_tests", num_ik_tests_, 0);
+    node_->get_parameter_or("num_ik_multiple_tests", num_ik_multiple_tests_, 0);
+    node_->get_parameter_or("num_nearest_ik_tests", num_nearest_ik_tests_, 0);
+    node_->get_parameter_or("ik_plugin_name", ik_plugin_name_, UNDEFINED);
+    node_->get_parameter_or("publish_trajectory", publish_trajectory_, false);
```
nbbrooks, line 131: I'm out of the loop on param retrieval and not sure how we want to standardize moveit2 on "this required param is not set in the namespace I'm expecting" - getParam is a nice pattern, but ends up being copied to a bunch of files, rather than being a single place "rosparam shortcuts package?"

Regardless, removing the getParam pattern + ASSERT_TRUE is not going to catch cases where a required param is not set. Do we have another test which will catch that? Or are these not truly "required params"? By required params I mean params which we cannot provide a "good" default value for.

vatanaksoytezer: I am currently not aware if we have such tests that will catch this behavior other than getParam+ ASSERT_TRUE. I'll have to  look further for it. Maybe @tylerjw or @JafarAbdi knows? If we have such thing I would love to use it as well.

vatanaksoytezer:
> Does https://github.com/ros/resource_retriever/blob/kinetic-devel/package.xml need to be added to the `package.xml`?

Thanks for asking! I think we need `resource_retriever`, we had a problem with this package today breaking moveit2 builds see #442 .

sjahr (commented):
I've tested it locally and the bugs I described in the previous PR are gone. Awesome work! Thanks a lot for picking that up!

vatanaksoytezer:
This is ready for review if this CI pass.

moveit_kinematics/test/launch/fanuc-kdl-singular.test.py
```diff
@@ -0,0 +1,199 @@
...
+    robot_description_kinematics = {"robot_description_kinematics": kinematics_yaml}
+    test_param = load_yaml("moveit_kinematics", "config/fanuc-kdl-test.yaml")
+
+    singular_private_params = {
+        "seed": [0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
+        "num_fk_tests": 0,
+        "num_ik_tests": 0,
+    }
+
+    singular_unit_tests_poses = {
```
henningkayser, line 60: Could you move these poses into a yaml file, maybe even the new test yamls that you created?

moveit_kinematics/test/test_kinematics_plugin.cpp
```diff
@@ -412,353 +390,358 @@ TEST_F(KinematicsTest, randomWalkIK)
...
+{
+  static const std::string TEST_POSES_PARAM = "unit_test_poses";
+  size_t expected_test_poses = 0;
+  node_->get_parameter_or(TEST_POSES_PARAM + ".size", expected_test_poses, expected_test_poses);
+
+  std::vector<double> sol;
+  const std::vector<std::string>& tip_frames = kinematics_solver_->getTipFrames();
+  moveit::core::RobotState robot_state(robot_model_);
+  robot_state.setToDefaultValues();
+  robot_state.setJointGroupPositions(jmg_, seed_);
```
henningkayser, line 437: The original test checks if `seed_` is empty, is that case impossible now?

moveit_kinematics/test/config/fanuc-ikfast-test.yaml
```diff
@@ -0,0 +1,19 @@
+tip_link: "tool0"
+root_link: "base_link"
+group: "manipulator"
+ik_timeout: 0.2
+tolerance: 0.1 ## TODO(henningkayser): set arg
```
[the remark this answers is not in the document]
vatanaksoytezer: I think you or Sebestian left this one before. I can just delete it I guess.

moveit_kinematics/test/launch/fanuc-kdl-singular.test.py
vatanaksoytezer: Sure

moveit_kinematics/test/test_kinematics_plugin.cpp
vatanaksoytezer: We can check if it is empty or not, but every test declares a seed anyway.

moveit_kinematics/test/config/fanuc-ikfast-test.yaml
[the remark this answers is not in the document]
henningkayser: Yeah, I don't see the point in using launch args for this

moveit_kinematics/test/test_kinematics_plugin.cpp
henningkayser: Could you add an assert that checks if `seed_` has the correct size?

vatanaksoytezer: Sure thing

vatanaksoytezer: There was already one: https://github.com/vatanaksoytezer/moveit2/blob/feature/enable-kinematic-tests/moveit_kinematics/test/test_kinematics_plugin.cpp#L117

henningkayser (approved):

moveit_kinematics/test/test_kinematics_plugin.cpp
```diff
@@ -412,353 +390,358 @@ TEST_F(KinematicsTest, randomWalkIK)
...
+    pose_values.clear();
+    node_->get_parameter_or(pose_param + ".pose", pose_values, pose_values);
+    ASSERT_TRUE(pose_values.size() == 6 || pose_values.size() == 7)
+        << "Test pose '" << pose_name << "' has invalid 'pose' vector size";
+
+    Eigen::Isometry3d pose;
+    ASSERT_TRUE(parsePose(pose_values, pose)) << "Failed to parse 'pose' vector in: " << pose_name;
+    if (pose_values.size() == 7)
+    {
+      RCLCPP_DEBUG(LOGGER, "Pose Input (7): %f %f %f %f %f %f %f\n", pose_values.at(0), pose_values.at(1),
```
sjahr, line 512: I think the debug output can be removed

moveit_kinematics/test/test_kinematics_plugin.cpp
```diff
@@ -412,353 +390,358 @@ TEST_F(KinematicsTest, randomWalkIK)
...
+          (Eigen::Map<const Eigen::VectorXd>(s.data(), s.size()) - seed_eigen).array().abs().sum();
+      if (error_multiple_ik <= smallest_error_multiple_ik)
+        smallest_error_multiple_ik = error_multiple_ik;
+    }
+    EXPECT_NEAR(smallest_error_multiple_ik, error_get_ik, tolerance_);
+  }
+}
 
 int main(int argc, char** argv)
 {
```
sjahr, line 747: Maybe add a  rclcpp::shutdown() before returning the result to properly shutdown the context

sjahr (approved):
I've tested it on my local setup: the tests pass and seem to work. The code looks good to me, I've commented only on a few minor things you could change.

moveit_kinematics/test/test_kinematics_plugin.cpp
vatanaksoytezer: I think the shutdown is called automatically in the end even if you don't write it in the context of tests.

moveit_kinematics/test/test_kinematics_plugin.cpp
vatanaksoytezer: Sure

vatanaksoytezer: > I think this is still relevant, maybe you could use ASSERT_TRUE(node_->get_parameter(...)) without setting a default value. Then the tests would fail if a required parameter is not avaiilable

This makes sense I will add it.

Issue #441, opened 2021-04-26
colcon build fails at moveit_core can't find target (ROS2 Foxy)


newcanopies:
### Description

colcon fails to build moveit_core
find_package() calls seem to work but targets aren't picked up

sourced foxy install before build
$ . /opt/ros/foxy/setup.bash
$ colcon build --event-handlers desktop_notification- status- --cmake-args -DCMAKE_BUILD_TYPE=Release  

full output log is in [this Gist](https://gist.github.com/newcanopies/c1c95cc27ec4f8dd7a002572e6849d8c) and moveit_core excertp below. 

### Your environment
* ROS2 Foxy
* Ubuntu 20.04 / WSL2
* Binary
* ?

### Backtrace or Console output

```
- \\ -

--- stderr: moveit_core
CMake Error at utils/CMakeLists.txt:16 (add_library):
  Target "moveit_test_utils" links to target
  "resource_retriever::resource_retriever" but the target was not found.
  Perhaps a find_package() call is missing for an IMPORTED target, or an
  ALIAS target is missing?


CMake Error at /opt/ros/foxy/share/ament_cmake_gtest/cmake/ament_add_gtest_executable.cmake:50 (add_executable):
  Target "test_robot_model" links to target
  "resource_retriever::resource_retriever" but the target was not found.
  Perhaps a find_package() call is missing for an IMPORTED target, or an
  ALIAS target is missing?
Call Stack (most recent call first):
  /opt/ros/foxy/share/ament_cmake_gtest/cmake/ament_add_gtest_executable.cmake:37 (_ament_add_gtest_executable)
  /opt/ros/foxy/share/ament_cmake_gtest/cmake/ament_add_gtest.cmake:68 (ament_add_gtest_executable)
  robot_model/CMakeLists.txt:35 (ament_add_gtest)

CMake Generate step failed.  Build files cannot be regenerated correctly.
---
Failed   <<< moveit_core [59.4s, exited with code 1]
Aborted  <<< diff_drive_controller [1min 27s]
Aborted  <<< joint_state_broadcaster [1min 33s]
Aborted  <<< position_controllers [20.5s]
Aborted  <<< velocity_controllers [20.6s]
Aborted  <<< joint_trajectory_controller [1min 39s]
Aborted  <<< effort_controllers [25.7s]

Summary: 22 packages finished [6min 9s]
  1 package failed: moveit_core
[...]

Issue #440, opened 2021-04-26
colcon build problem at moveit_core 


AtariaLee:
### Description

After  i update my moveit2 files today,some problems happend in build.When i run `colcon build --event-handlers desktop_notification- status- --cmake-args -DCMAKE_BUILD_TYPE=Release` to build my workspace,some problems happend.
 I dont know what happend,and how to solve it.I will be very grateful to you for your help!!!
### Your environment
* ROS Distro: RO2 FOXY
* OS Version: Ubuntu 20.04
* maybe binary?

### Steps to reproduce
When i run `colcon build --event-handlers desktop_notification- status- --cmake-args -DCMAKE_BUILD_TYPE=Release` to build my workspace,some problems happend.
### Expected behaviour
 All files are successful!

### Backtrace or Console output
--- stderr: moveit_core
CMake Error at utils/CMakeLists.txt:16 (add_library):
  Target "moveit_test_utils" links to target
  "resource_retriever::resource_retriever" but the target was not found.
  Perhaps a find_package() call is missing for an IMPORTED target, or an
  ALIAS target is missing?

CMake Error at /opt/ros/foxy/share/ament_cmake_gtest/cmake/ament_add_gtest_executable.cmake:50 (add_executable):
  Target "test_robot_model" links to target
  "resource_retriever::resource_retriever" but the target was not found.
  Perhaps a find_package() call is missing for an IMPORTED target, or an
  ALIAS target is missing?
Call Stack (most recent call first):
  /opt/ros/foxy/share/ament_cmake_gtest/cmake/ament_add_gtest_executable.cmake:37 (_ament_add_gtest_executable)
  /opt/ros/foxy/share/ament_cmake_gtest/cmake/ament_add_gtest.cmake:68 (ament_add_gtest_executable)
  robot_model/CMakeLists.txt:35 (ament_add_gtest)

CMake Error at robot_model/CMakeLists.txt:4 (add_library):
  Target "moveit_robot_model" links to target
  "resource_retriever::resource_retriever" but the target was not found.
  Perhaps a find_package() call is missing for an IMPORTED target, or an
  ALIAS target is missing?

CMake Error at /opt/ros/foxy/share/ament_cmake_gtest/cmake/ament_add_gtest_executable.cmake:50 (add_executable):
  Target "test_all_valid" links to target
  "resource_retriever::resource_retriever" but the target was not found.
  Perhaps a find_package() call is missing for an IMPORTED target, or an
  ALIAS target is missing?
[...]

PR #442, opened 2021-04-26
WIP: Fix missing dependency resource_retriever

tylerjw:
### Description

This should fix these two issues:
* https://github.com/ros-planning/moveit2/issues/440
* https://github.com/ros-planning/moveit2/issues/441

The dependency `resource_retriever` must have previously been found via one of our dependencies.  I'm not sure which one (as moveit_core has a ton of dependencies) but whichever one it was they no longer have that dependency but we still do.

### DIFF
--- a/moveit_core/CMakeLists.txt
+++ b/moveit_core/CMakeLists.txt
@@ -206,6 +206,13 @@ if(BUILD_TESTING)
 
   # Run all lint tests in package.xml except those listed above
   ament_lint_auto_find_test_dependencies()
+
+  # This has to be exported here because it is needed to link for the test
+  #  in collision_distance_field but exporting only works from the top level
+  #  cmake file (this one).  This means that if tests link external libraries
+  #  they need to be exported like this or libraries that depend on this one
+  #  will throw a cmake error about a missing find_package when they build.
+  ament_export_dependencies(resource_retriever)
 endif()
 
 ament_package(CONFIG_EXTRAS ConfigExtras.cmake)
--- a/moveit_core/package.xml
+++ b/moveit_core/package.xml
@@ -59,6 +59,7 @@
   <test_depend>orocos_kdl</test_depend>
   <test_depend>ament_cmake_gtest</test_depend>
   <test_depend>ament_index_cpp</test_depend>
+  <test_depend>resource_retriever</test_depend>
 
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>


tylerjw:
Here is the error message this is fixing:

```

--- stderr: moveit_simple_controller_manager
CMake Error at CMakeLists.txt:32 (add_library):
  Target "moveit_simple_controller_manager" links to target
  "resource_retriever::resource_retriever" but the target was not found.
  Perhaps a find_package() call is missing for an IMPORTED target, or an
  ALIAS target is missing?


CMake Error at CMakeLists.txt:32 (add_library):
  Target "moveit_simple_controller_manager" links to target
  "resource_retriever::resource_retriever" but the target was not found.
  Perhaps a find_package() call is missing for an IMPORTED target, or an
  ALIAS target is missing?


CMake Generate step failed.  Build files cannot be regenerated correctly.
make: *** [Makefile:296: cmake_check_build_system] Error 1
---
```

I believe this issue was caused by the latest Foxy sync and the issue is actually upstream of us.  The reason I'm saying this is I can't find anywhere that we are linking this library.  Someone upstream of us is probably not exporting it as a dependency.

tylerjw:
I tried searching through upstream repos and I'm not sure they are to blame.  This might have something to do with depending on moveit_core which contains the test `test_collision_distance_field` which does link with `resource_retriever::resource_retriever`.  However that's only when BUILD_TESTING is enabled.  Does this mean we need to export test dependencies too?

tylerjw:
I think I figured it out (no idea why it is just showing up now).  `test_collision_distance_field` links this library that is added as a test_depend on moveit_core.  We correctly find_package it within `collision_distance_field` under the if build testing block.  However, because this is an external library we are linking it has to be exported if build testing is enabled.   Because of a limitation of the `ament_export_dependencies` cmake macro this has to be done at the top level cmake file in `moveit_core`.

The reason for the errors inside the other two libraries is that they link `moveit_core` and define `if(BUILD_TESTING)` sections themselves.

I believe this is now ready and will work correctly.  This was not the fault of something upstream, but instead something internal to moveit_core.

v4hn:
@dirk-thomas This requires your expertise.

At this point, `moveit_core` find_packages `resource_retriever` in two places: [here](https://github.com/ros-planning/moveit2/blob/3a41a20374e7c9a4cfe0e2e1b472688159b59313/moveit_core/robot_state/CMakeLists.txt#L34) and [here](https://github.com/ros-planning/moveit2/blob/main/moveit_core/collision_distance_field/CMakeLists.txt#L30).
The former seems unnecessary anyway, but the latter one is actually linked to a target created with `ament_add_gtest`.
It's a single binary which is not used in any package downstream.
Still it seems, [this unrelated target](https://github.com/ros-planning/moveit2/blob/3a41a20374e7c9a4cfe0e2e1b472688159b59313/moveit_plugins/moveit_simple_controller_manager/CMakeLists.txt#L30) which adds [`moveit_core` via `ament_target_dependencies`](https://github.com/ros-planning/moveit2/blob/3a41a20374e7c9a4cfe0e2e1b472688159b59313/moveit_plugins/moveit_simple_controller_manager/CMakeLists.txt#L37) ends up linking to `resource_retriever` for no reason and @tylerjw's fix here [is to export this external dependency](https://github.com/ros-planning/moveit2/pull/442/files#diff-34e4003b17ce0718e3672073b6f4a88c54d09d9b4a965f8a791ad01d9f687abeR215), which is not meant to be a public dependency at all.

Looks fishy to me and I suspect either a bug in ament or a restriction due to its design.

AndyZe (approved):
I think we should merge this soon-ish because it's blocking all other PR's.

tylerjw:
Please don't merge this.  I believe this is the real fix: https://github.com/ros-planning/geometric_shapes/pull/182

dirk-thomas:
> @dirk-thomas This requires your expertise.

@v4hn Sorry, I am not actively involved in the project [anymore](https://discourse.ros.org/t/thank-you-farewell/17297) .

v4hn:
> I am not actively involved in the project anymore.

I'm more than sorry to hear that. As others told you many times, you will be missed. :bouquet: 
For future reference, who can we ping for intricate questions about catkin/ament/colcon/dozens-of-other-things?

Anyway, it looks like @tylerjw found an [unexpected other culprit](https://github.com/ros-planning/geometric_shapes/pull/182) that might explain the mystery I pinged you about.

dirk-thomas:
> For future reference, who can we ping for intricate questions about catkin/ament/colcon/dozens-of-other-things?

I don't think there is a single person. I can only recommend to reach out via discourse / ROS answers / issue trackers of the various repositories to hopefully get feedback from the right person. Each repository has a dedicated maintainer assigned.

tylerjw:
@dirk-thomas thank you for all you've done.  I wish you success in whatever you are doing now.  Thank you again.

I'm closing this PR as I fixed it at the source in geometric_shapes.

PR #428, opened 2021-04-19
Refactor Servo velocity bounds enforcement. Disable flaky unit tests.

vatanaksoytezer:
Port of https://github.com/ros-planning/moveit/pull/2471 and an attempt to fix flakiness (again). Overall this PR:

- Ports 2471.
- Fixes most flakiness of the tests.
- Adds low latency (named basic_servo_tests)

PR #451, opened 2021-05-04
Improvement on Servo enforcePositionLimits()

AndyZe:
Previously, the incoming robot state velocity was used to check if the robot would move beyond the joint limit. That was a problem if the robot state did not contain velocity data, i.e. if "copy dynamics" hadn't been turned on.

This PR uses the prospective robot velocity from Servo itself to check if the joint would move beyond the position limit and halt, if so. This velocity should always be available regardless of whether "copy dynamics" is ON.

### DIFF
--- a/moveit_core/robot_model/src/revolute_joint_model.cpp
+++ b/moveit_core/robot_model/src/revolute_joint_model.cpp
@@ -163,7 +163,9 @@ bool RevoluteJointModel::satisfiesPositionBounds(const double* values, const Bou
   if (continuous_)
     return true;
   else
+  {
     return !(values[0] < bounds[0].min_position_ - margin || values[0] > bounds[0].max_position_ + margin);
+  }
 }
 
 bool RevoluteJointModel::harmonizePosition(double* values, const JointModel::Bounds& other_bounds) const
--- a/moveit_ros/moveit_servo/src/servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/src/servo_calcs.cpp
@@ -582,7 +582,7 @@ bool ServoCalcs::internalServoUpdate(Eigen::ArrayXd& delta_theta,
   composeJointTrajMessage(internal_joint_state_, joint_trajectory);
 
   // Enforce SRDF position limits, might halt if needed, set prev_vel to 0
-  if (!enforcePositionLimits())
+  if (!enforcePositionLimits(joint_trajectory))
   {
     suddenHalt(joint_trajectory);
     status_ = StatusCode::JOINT_BOUND;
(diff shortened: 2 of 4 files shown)


moveit_core/robot_model/src/revolute_joint_model.cpp
```diff
@@ -163,7 +163,9 @@ bool RevoluteJointModel::satisfiesPositionBounds(const double* values, const Bou
   if (continuous_)
     return true;
   else
+  {
     return !(values[0] < bounds[0].min_position_ - margin || values[0] > bounds[0].max_position_ + margin);
+  }
```
tylerjw, line 168: Thank you for this.  Can you also add the brackets around the if statement on the lines above?

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -809,43 +809,52 @@ void ServoCalcs::enforceVelLimits(Eigen::ArrayXd& delta_theta)
...
+        // Check if pending velocity command is moving in the right direction
+        auto joint_itr =
+            std::find(joint_trajectory.joint_names.begin(), joint_trajectory.joint_names.end(), joint->getName());
+        auto joint_idx = std::distance(joint_trajectory.joint_names.begin(), joint_itr);
+
+        if ((joint_trajectory.points[0].velocities[joint_idx] < 0 &&
              (joint_angle < (limits[0].min_position + parameters_->joint_limit_margin))) ||
-            (current_state_->getJointVelocities(joint)[0] > 0 &&
+            (joint_trajectory.points[0].velocities[joint_idx] > 0 &&
              (joint_angle > (limits[0].max_position - parameters_->joint_limit_margin))))
```
tylerjw, line 845: Is there a current test that tests the behavior of this?  If not, could you extend the test to test this?

tylerjw (commented):
Generally approve; left a couple of minor comments.

moveit_core/robot_model/src/revolute_joint_model.cpp
AndyZe: this was just leftover from debugging, but OK

AndyZe: this was just leftover from debugging, but OK

moveit_ros/moveit_servo/src/servo_calcs.cpp
AndyZe: good idea, done

tylerjw (approved):
thank you for adding the test, this looks good to me :1st_place_medal:

moveit_ros/moveit_servo/test/test_servo_status.cpp
```diff
@@ -90,6 +90,21 @@ TEST_F(ServoFixture, ReachSingular)
...
+  {
+    loop_duration = node_->now() - start_time;
+    auto joint_msg = std::make_unique<control_msgs::msg::JointJog>();
+    joint_msg->header.stamp = node_->now();
+    joint_msg->header.frame_id = "panda_link3";
+    joint_msg->velocities.push_back(2.0);
+    pub_joint_cmd_->publish(std::move(joint_msg));
+    publish_loop_rate.sleep();
+  }
+  EXPECT_EQ(getLatestStatus(), moveit_servo::StatusCode::JOINT_BOUND);
```
vatanaksoytezer, line 106: This test seems to fail in CI where the status code that `getLatestStatus()` returns is 0 but `moveit_servo::StatusCode::JOINT_BOUND` is 5 (see https://github.com/ros-planning/moveit2/pull/451/checks?check_run_id=2506237803). Is there a specific reason why we are publishing 2.0 velocity? Maybe adding a comment about what the joint limits are would be nice.

vatanaksoytezer (commented):
I've read through the code and checked the failing CI. I left a small comment with a question

moveit_ros/moveit_servo/test/test_servo_status.cpp
AndyZe: You're right. I'll refactor the test.

No reason to choose 2.0 for velocity other than it's a reasonable value.

moveit_ros/moveit_servo/src/servo_calcs.cpp
AndyZe: @tylerjw please realize you are asking a lot here. The tests are currently commented with this note:

`# TODO(henningkayser): Port tests to ROS2`

So I effectively need to fix some other tests to add this one.

AndyZe: Getting #428 merged would unblock this PR. The test we need (`TestEnforcePosLimits`) is already there

AndyZe: Let's get this merged and we are working to get `TestEnforcePosLimits` operational in #428.

Issue #315, opened 2020-12-04
Migrate Pilz industrial motion planner to ROS 2


henningkayser:
The new Pilz planner has been merged with https://github.com/ros-planning/moveit2/commit/7d903094a8031a07847b8d6d38793df477ba2716 and needs to be ported to ROS 2.

- [ ] fix failing tests

PR #452, opened 2021-05-05
Port Pilz Industrial Motion Planner

henningkayser:
Fixes https://github.com/ros-planning/moveit2/issues/315

You can test the planner using the PRBT [demo.launch.py](https://github.com/henningkayser/moveit_resources/blob/pr-port_prbt_packages/prbt_moveit_config/launch/demo.launch.py)

**Update**
* ~~(almost)~~ All unit tests are finally succeeding locally for me now (thanks @sjahr for all the restructuring and preparing the launch scripts)
* Gripper tests are still disabled (they were implemented using a test suite with robot models for arm-only and arm+gripper). Support for that requires adding more boilerplate to the launch files. Those tests are also somewhat specific to the PRBT so IMO we should probably drop them and just prepare different launch setups for different robots to run the existing tests.
* I reduced redundancy in the build structure. Some source files were compiled over and over again (different plugins, different tests) and there was even an extra library that included almost all source files just to provide test dependencies again. I broke that down into smaller libraries that can be reused and configured more specific target dependencies for most libraries and tests. There is still some cleanup that could be done, but I would expect that this reduces compile time significantly. I was planning to compare build times with the old setup, maybe this is something we want to backport to MoveIt 1.

**Issues (Future work)**

Some stuff I encountered while working on this. I'll file issues for these the next days.

* There is a bunch of boilerplate code in the tests that could be moved into moveit_test_utils
* I ran into weird runtime issues where plugins could not be loaded even though they were reported as "declared" by the ClassLoader. E.g. MoveIt capabilites were found but failed to load only when the Pilz planner has been loaded before. I was only able to fix this by letting all plugins use a plugin name (removing plugin names from all plugins also worked). @AndyZe this might very well be related to your weird issue where the smoothing plugin only didn't load inside a component node, which also is a plugin. I still want to confirm this with a reproducible example, but this is probably just a bug upstream. Since plugin names are not required anymore in ROS 2, we could also drop them from all the existing ones (or enforce adding them) if the issue persist.
* RViz dies when switching planners during execution
* I'm frequently running into a std::bad_alloc when loading a robot model, looking into this next


moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -640,7 +640,7 @@ bool JointModelGroup::canSetStateFromIK(const std::string& tip) const
     // remove frame reference, if specified
     const std::string& tip_local = tip[0] == '/' ? tip.substr(1) : tip;
     const std::string& tip_frame_local = tip_frame[0] == '/' ? tip_frame.substr(1) : tip_frame;
-    RCLCPP_WARN(LOGGER, "comparing input tip: %s to this groups tip: %s ", tip_local.c_str(), tip_frame_local.c_str());
```
henningkayser, line 643: not sure why this was a warning, was spamming everywhere when enabling the tests. I changed it back how it was since MoveIt 1.

moveit_planners/pilz_industrial_motion_planner/src/joint_limits_aggregator.cpp
```diff
@@ -39,22 +39,33 @@
...
   for (auto joint_model : joint_models)
   {
     JointLimit joint_limit;
 
-    // If there is something defined for the joint on the parameter server
-    if (joint_limits_interface::getJointLimits(joint_model->getName(), nh, joint_limit))
+    // NOTE: declareParameters fails (=returns false) if the parameters have already been declared.
+    // The function should be checked in the if condition below when we disable
+    // 'NodeOptions::automatically_declare_parameters_from_overrides(true)'
+    joint_limits_interface::declareParameters(joint_model->getName(), param_namespace, node);
```
henningkayser, line 65: We should make a decision if we want to enforce declaring parameters everywhere or if we are fine with allowing automatic overrides moving forward.

henningkayser:
clang-tidy doesn't like the copied joint_limits headers and the included ikfast plugin for the PRBT. Both should be ignored. Otherwise, CI seems happy with the current version.

henningkayser:
Update:
* All unit tests pass for the PRBT package
* I think the migration is ready to get merged before we branch off rolling

However, I don't think that it makes sense to enable the PRBT in `moveit_resources` and enabling the tests for a different robot is a bigger undertaking. For now, I switched to the Panda robot, but obviously a lot of the tests fail since there are still plenty of hardcoded values in the tests and the Panda doesn't provide deterministic IK at this point. The failing tests are disabled with the last commit.
I still think that we should merge this before the branch-off and then work on fixing all the tests for a different robot as suggested in this issue https://github.com/ros-planning/moveit/issues/2947.

henningkayser:
Based on @tylerjw's feedback, I switched this back to use PRBT to keep the tests enabled. We still need to migrate to a different robot, but keeping the old setup around is worth it in order to also keep the tests enabled.

moveit_planners/pilz_industrial_motion_planner/CMakeLists.txt
```diff
@@ -1,512 +1,225 @@
...
+ament_export_libraries(
+  ${PROJECT_NAME}
+  joint_limits_common
+  planning_context_loader_base
+  planning_context_loader_ptp
+  planning_context_loader_lin
+  planning_context_loader_circ
+  sequence_capability
+  trajectory_generation_common
+)
```
JafarAbdi, line 203: I don't think these are needed since we export the targets

moveit_planners/pilz_industrial_motion_planner/CMakeLists.txt
```diff
@@ -1,512 +1,225 @@
...
+ament_export_targets(export_${PROJECT_NAME})
```
JafarAbdi, line 204:
```suggestion
ament_export_targets(export_${PROJECT_NAME} HAS_LIBRARY_TARGET)
```

moveit_planners/pilz_industrial_motion_planner/src/move_group_sequence_action.cpp
```diff
@@ -47,56 +47,80 @@
...
+        RCLCPP_DEBUG(LOGGER, "Canceling action goal");
+        preemptMoveCallback();
+        return rclcpp_action::CancelResponse::ACCEPT;
+      },
+      [this](
+          const std::shared_ptr<rclcpp_action::ServerGoalHandle<moveit_msgs::action::MoveGroupSequence>> goal_handle) {
+        // Return quickly to avoid blocking executor
+        std::thread(std::bind(&MoveGroupSequenceAction::executeSequenceCallback, this, std::placeholders::_1),
+                    goal_handle)
+            .detach();
```
JafarAbdi, line 85: I don't think this's a good solution, why not have a separate callback group for this action?

henningkayser: Switched to callbackgroup  with ebf83c3

moveit_planners/pilz_industrial_motion_planner/CMakeLists.txt
henningkayser: Fixed

moveit_planners/pilz_industrial_motion_planner/CMakeLists.txt
henningkayser: Fixed

henningkayser:
@JafarAbdi thanks for your review, all requests are fixed. I already merged the moveit_resources PR and updated the installed repo here as well. IMO, this is ready to get merged if CI succeeds, but please don't squash-merge.

.github/workflows/ci.yaml
```diff
@@ -27,6 +27,7 @@ jobs:
       CXXFLAGS: "-Wall -Wextra -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls -Wno-deprecated-copy"
       DOCKER_IMAGE: ghcr.io/ros-planning/moveit2:${{ matrix.env.IMAGE }}
       UPSTREAM_WORKSPACE: moveit2.repos $(f="moveit2_$(sed 's/-.*$//' <<< "${{ matrix.env.IMAGE }}").repos"; test -r $f && echo $f)
+      TARGET_WORKSPACE: $TARGET_REPO_PATH github:henningkayser/moveit_resources#pr-port_prbt_packages
```
tylerjw, line 30: Shouldn't you drop this change now that it is merged into moveit_resources?

henningkayser: I thought I just did this... I must have failed to commit it

henningkayser: Should be fixed now. We still need moveit_resources in the target workspace because of the circular dependency (see #885). It's the same issue with MoveIt 1

tylerjw:
I pushed a small change to fix a clang-tidy warning in one of the tests.

tylerjw (approved):

Issue #444, opened 2021-04-28
Motion planning cannot avoid Octomap in planning scene


PrakarnJ:
### Description
I am working on robot motion planning with perception from point cloud topic (3d camera). I found that when I planned robot path, the robot cannot avoid the collision in planning scene as shown in Figure ![1](https://media.giphy.com/media/1Zpr9iZJ5JtrevjaUl/giphy.gif).

I tried to solve this problem and found that one of the solution is it need to add something to planning scene using Scene objects tab UI and push Publish button as shown in Figure ![Screenshot from 2021-04-28 20-49-43](https://user-images.githubusercontent.com/31307798/116421532-20304980-a869-11eb-83de-efdb25b10e91.jpg)

After that I planned robot path again and you can see robot can avoid the obstacle 
![3](https://media.giphy.com/media/AL9p2sgD0iPethJTOp/giphy.gif).

Do anyone have the way to solve this problem or the way to check this issue?

### Your environment
* ROS Distro: Foxy
* OS Version: Ubuntu 20.04

swilcock0:
This looks like #407 and #429

henningkayser:
This doesn't look like #407 to me, but it might really be related to #429 which also looks like planning scene updates not being synced properly. I'll check if there is something odd going on, collision checking working after the manual update is very suspicious.

swilcock0:
@henningkayser the manual update thing is the same with #407

PrakarnJ:
@JafarAbdi Yeah It can fixed.

Issue #429, opened 2021-04-19
move_group is not respecting the attached collision object


MarqRazz:
### Description

We are working on porting the [tutorials](https://github.com/ros-planning/moveit2_tutorials/pull/69) and I recently noticed issues with the robot not respecting the attached collision object.
![collision_error](https://user-images.githubusercontent.com/25058794/115306944-c7083d80-a125-11eb-9b91-df89b8bcbba1.gif)

This feature was working for me a few weeks ago.
@JafarAbdi any ideas? @vatanaksoytezer said you had recently seen issues in moveit1.

### Your environment
* ROS Distro: [Foxy]
* OS Version: e.g. Ubuntu 20.04
* Source build?
* If binary, which release version?
* If source, which branch?

### Steps to reproduce
Tell us how to reproduce this issue. Attempt to provide a working demo, perhaps using Docker.

### Expected behaviour
Tell us what should happen

### Actual behaviour
Tell us what happens instead

### Backtrace or Console output

Use [gist.github.com](gist.github.com) to copy-paste the console output or segfault backtrace using gdb.

vatanaksoytezer:
Thanks a lot for the issue! I will have a look at this tomorrow and see if I can figure it out.

Issue #407, opened 2021-03-31
Collision objects not publishing properly with PlanningSceneInterface as class member


swilcock0:
### Is there a method to "publish" planning_scene_interface changes from move_group? 

Currently I've been using this code below to create a subscriber which adds collision bricks to the scene, as part of a larger file. They appear in RVIZ, and until recently it was working well. Recently however the objects are being passed through. Republishing the objects through the motion planning plugin in RVIZ makes the planner recognise them properly.

UPDATE:
See next comment for link to a working demonstration code for this problem

```cpp
  
#include <streambuf>
#include <string>
#include <cmath>
#include <iostream>
#include <fstream>

#include <rclcpp/rclcpp.hpp>
#include <rclcpp/time_source.hpp>

#include <ament_index_cpp/get_package_share_directory.hpp>

#include <moveit/move_group_interface/move_group_interface.h>
#include <moveit/planning_scene_interface/planning_scene_interface.h>

#include <moveit_msgs/msg/display_robot_state.hpp>
#include <moveit_msgs/msg/display_trajectory.hpp>
#include <moveit_msgs/msg/attached_collision_object.hpp>
#include <moveit_msgs/msg/collision_object.hpp>

#include <geometry_msgs/msg/pose.h>
#include <geometry_msgs/msg/pose_stamped.h>

namespace ph = std::placeholders;
using namespace std::chrono_literals;

auto SUCCESS = moveit::planning_interface::MoveItErrorCode::SUCCESS;

// Parameters
const std::string PLANNING_GROUP = "panda_arm";

const std::string node_name = "move_group_interface_tutorial";

static const rclcpp::Logger logger_ = rclcpp::get_logger(node_name);

class moveServer : public rclcpp::Node
{
public:
  moveServer()
    : Node(node_name, rclcpp::NodeOptions().automatically_declare_parameters_from_overrides(true))
    , move_group_(std::shared_ptr<rclcpp::Node>(std::move(this)), PLANNING_GROUP)
    , planning_scene_interface_()
  {
    brick_num = 0;
    attached = false;

    // Config moveit params
    this->move_group_.setPlanningTime(5.0);
    this->move_group_.setMaxVelocityScalingFactor(1.0);
    this->move_group_.setMaxAccelerationScalingFactor(1.0);
    this->move_group_.setNumPlanningAttempts(2);
    this->move_group_.allowReplanning(true);

    // Configure callback groups for multithreading
[...]

PR #456, opened 2021-05-07
Revert "Eliminate ability to keep multiple collision detectors updated"

JafarAbdi:
Fix: https://github.com/ros-planning/moveit2/issues/407
Fix: https://github.com/ros-planning/moveit2/issues/444
Fix: https://github.com/ros-planning/moveit2/issues/429

Reverts ros-planning/moveit2#364

I wasn't able to trace where the bug is actually happening (it only happen when the planning scene have a parent scene), so just reverting it for now

### DIFF
--- a/moveit_core/collision_detection/include/moveit/collision_detection/collision_plugin.h
+++ b/moveit_core/collision_detection/include/moveit/collision_detection/collision_plugin.h
@@ -87,7 +87,7 @@ public:
   /**
    * @brief This should be used to load your collision plugin.
    */
-  virtual bool initialize(const planning_scene::PlanningScenePtr& scene) const = 0;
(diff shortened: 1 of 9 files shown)


AndyZe:
I don't think we should merge this, even if it does fix the issues. It adds about 10% more code and a fair bit of complexity to the planning scene   :(

vatanaksoytezer:
I tested this in tutorials and this fix works for me. The robot now respects the attached collision object as intended. However as @AndyZe said we may need to a little deep and try to find which part of the original PR caused the issue.

swilcock0:
This does also fix #407

PrakarnJ:
It can fixed #444

PR #464, opened 2021-05-18
Fix PlanningScene CollisionDetector diff handling

### LINKED ISSUE #429: move_group is not respecting the attached collision object

henningkayser:
This fixes how PlanningScene diffs are created and managed (replaces #463 and #456).
https://github.com/ros-planning/moveit2/pull/364 introduced a bug where a child scene would not copy-construct the collision environments so that it wouldn't include any of the parent's link geometries or attached objects. The link geometries aren't that bad because they would be constructed from the robot model anyway, but there was no way to sync back the attached objects. This PR fixes that by providing a private allocate function that allows passing the parent detector which implements the old behavior. I tested this and it fixes #429 and #407 for me. I didn't test #444 yet.

However, there is still a bug (or irritation) where the start state of the MotionPlanningPanel doesn't include the attached object right away after publishing the collision objects. The reason is that the internal start state of the panel is not updated automatically which is also exactly the case for MoveIt 1. The requested plan then solves a trajectory which is then being rejected right away because it's in collision. To make the plan succeed, the start state needs to be updated either by selecting "\<current\>" again or by changing it to a different state. I actually don't think that this needs fixing and it's definitely out of scope of this PR.

### DIFF
--- a/moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
+++ b/moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
@@ -912,7 +912,7 @@ public:
   /** \brief Clone a planning scene. Even if the scene \e scene depends on a parent, the cloned scene will not. */
   static PlanningScenePtr clone(const PlanningSceneConstPtr& scene);
 
-  /** \brief Replace previous collision detector with a new collision detector type (or create new, if none previously).
+  /** \brief Allocate a new collision detector and replace the previous one if there was any.
    *
    * The collision detector type is specified with (a shared pointer to) an
    * allocator which is a subclass of CollisionDetectorAllocator.  This
@@ -922,10 +922,13 @@ public:
    * A new PlanningScene uses an FCL collision detector by default.
    *
    * example: to add FCL collision detection (normally not necessary) call
-   *   planning_scene->setCollisionDetectorType(collision_detection::CollisionDetectorAllocatorFCL::create());
+   *   planning_scene->allocateCollisionDetector(collision_detection::CollisionDetectorAllocatorFCL::create());
    *
    * */
-  void setCollisionDetectorType(const collision_detection::CollisionDetectorAllocatorPtr& allocator);
+  void allocateCollisionDetector(const collision_detection::CollisionDetectorAllocatorPtr& allocator)
+  {
+    allocateCollisionDetector(allocator, nullptr /* no parent_detector */);
+  }
 
 private:
   /* Private constructor used by the diff() methods. */
@@ -948,6 +951,10 @@ private:
   static void poseMsgToEigen(const geometry_msgs::msg::Pose& msg, Eigen::Isometry3d& out);
 
   MOVEIT_STRUCT_FORWARD(CollisionDetector)
+
+  /* Construct a new CollisionDector from allocator, copy-construct environments from parent_detector if specified */
+  void allocateCollisionDetector(const collision_detection::CollisionDetectorAllocatorPtr& allocator,
+                                 const CollisionDetectorPtr& parent_detector);
 
   /* \brief A set of compatible collision detectors */
   struct CollisionDetector
--- a/moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
+++ b/moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
@@ -952,7 +952,7 @@ private:
 
   MOVEIT_STRUCT_FORWARD(CollisionDetector)
 
-  /* Construct a new CollisionDector from allocator, copy-construct environments from parent_detector if specified */
+  /* Construct a new CollisionDector from allocator, copy-construct environments from parent_detector if not null */
   void allocateCollisionDetector(const collision_detection::CollisionDetectorAllocatorPtr& allocator,
                                  const CollisionDetectorPtr& parent_detector);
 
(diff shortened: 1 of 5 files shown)


moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -223,36 +223,37 @@ void PlanningScene::CollisionDetector::copyPadding(const PlanningScene::Collisio
...
-  collision_detector_->cenv_ = collision_detector_->alloc_->allocateEnv(world_, getRobotModel());
-  collision_detector_->cenv_const_ = collision_detector_->cenv_;
-  collision_detector_->cenv_unpadded_ = collision_detector_->alloc_->allocateEnv(world_, getRobotModel());
-  collision_detector_->cenv_unpadded_const_ = collision_detector_->cenv_unpadded_;
 
-  // Copy padding from the previous collision detector
-  if (have_previous_coll_detector)
+  // If parent_detector is specified, copy-construct collision environments (copies link shapes and attached objects)
+  // Otherwise, construct new collision environment from world and robot model
+  if (parent_detector)
```
henningkayser, line 238: This is the key part of this fix. The collision environment of the parent is used to copy-construct the child environment so that attached objects are available as well.

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -295,11 +296,8 @@ void PlanningScene::clearDiffs()
   if (current_world_object_update_callback_)
     current_world_object_update_observer_handle_ = world_->addObserver(current_world_object_update_callback_);
 
-  if (parent_)
-  {
-    collision_detector_->copyPadding(*parent_->collision_detector_);
-  }
-  collision_detector_->cenv_const_ = collision_detector_->cenv_;
+  // Reset collision detector to the the parent's version
```
henningkayser, line 299: `clearDiffs()` now also just allocates a fresh collision detector from parent

moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
```diff
@@ -949,6 +952,10 @@ class PlanningScene : private boost::noncopyable, public std::enable_shared_from
 
   MOVEIT_STRUCT_FORWARD(CollisionDetector)
 
+  /* Construct a new CollisionDector from allocator, copy-construct environments from parent_detector if specified */
+  void allocateCollisionDetector(const collision_detection::CollisionDetectorAllocatorPtr& allocator,
+                                 const CollisionDetectorPtr& parent_detector);
```
AndyZe, line 957: I'm not sure this `parent_detector` argument is needed. Either there is a non-null `parent_` or there isn't.

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -223,36 +223,37 @@ void PlanningScene::CollisionDetector::copyPadding(const PlanningScene::Collisio
...
-    have_previous_coll_detector = true;
-    prev_coll_detector = *collision_detector_;
-  }
-
-  // TODO(andyz): uncomment this for a small speed boost when another collision detector type is available in MoveIt2
-  // For now, it is useful in the switchCollisionDetectorType() unit test
-  //  const std::string& name = allocator->getName();
-  //  if (name == getCollisionDetectorName())  // already using this collision detector
-  //    return;
+  // Temporary store the previous (if any) collision detector to copy padding from
```
AndyZe, line 229: Temporarily

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -223,36 +223,37 @@ void PlanningScene::CollisionDetector::copyPadding(const PlanningScene::Collisio
...
+  CollisionDetectorPtr prev_coll_detector = collision_detector_;
```
AndyZe, line 230: This just copies the pointer, right? I don't think it will copy the data in collision_detector_

moveit_core/planning_scene/src/planning_scene.cpp
AndyZe: I think this check should be:

`if (parent_)`

AndyZe (changes requested):

moveit_core/planning_scene/src/planning_scene.cpp
henningkayser: It copies the shared pointer which keeps the original object of `collision_detector_` alive while the pointer `collision_detector_` get reset right below. So yes, it doesn't copy the data, it just keeps the old object accessible for the scope of this function using the new shared pointer.

AndyZe: I see.

AndyZe:
Tested with the move_group_interface tutorial. Looks good to me!

AndyZe (approved):

PrakarnJ:
@henningkayser I already test this with octomap . It can fixed #444

PR #468, opened 2021-05-20
PlanningRequestAdapter helper method getParam 

DLu:
### Description

The same logic for resolving parameter names, checking for the existence of the parameter and printing the results was invoked several times in the different `PlanningRequestAdapter`s. This PR modularizes that logic and adds it for TOTG. 

I admit that the `getParam` method parameters are not the cleanest. There's an argument that instead, `node` and `parameter_namespace` are made class members and there's a base implementation of `initialize` that sets them. However, you'd still have to pass the LOGGER around (or create another instance). 

### Checklist


henningkayser (approved):
This is a good cleanup. We will at some point apply a repo-wide refactoring for declaring and loading all parameters. Wrapping the getters and checks in clean functions like this will help a lot with that.

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -993,6 +993,10 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
 
   // Compute sample count
   size_t sample_count = std::ceil(parameterized.getDuration() / resample_dt_);
+  if (sample_count < num_points)
```
henningkayser, line 996: By increasing `sample_count` to `num_points` the trajectory is filled with duplicate trajectory points with the same time stamps at the end which breaks CI (ros2_control isn't happy about that). The reason is that the computed timestamps in the loop below exceed the trajectory duration (see https://github.com/ros-planning/moveit2/pull/468/files#diff-227cef8a99e7eaf48f74355541c0f673279e65839065e5f29646319063377cdeR1008). Is it necessary for you to have at least `num_points` waypoints? If so, you would have to recompute `resample_dt_` as `parameterized.getDuration() / num_points` to match the trajectory duration. But I think the better option would be to just specify a resampling step that works in general and remove this change.

DLu: > But I think the better option would be to just specify a resampling step that works in general and remove this change.

Do you mean an entirely separate PlanningRequestAdapter that does the resampling? Because I agree, but don't know how to move forward with that in a way that doesn't break everyone's existing configurations.

henningkayser: No, I meant setting the resampling_dt parameter small enough so that you don't have to worry about trajectory waypoint count at all.

henningkayser:
@DLu Awesome, you broke the curse!

PR #471, opened 2021-05-22
Delete MoveIt fake_controller_manager

AndyZe:
This deletes the `fake_controller_manager` plugin. I don't believe it's needed anymore because ros2_control supports simulation in a fairly easy way. All you need to do is specify fake components in `your_robot.ros2_control.xacro`. ([example](https://github.com/ros-planning/moveit_resources/blob/ros2/panda_moveit_config/config/panda.ros2_control.xacro))

```
<hardware>
    <plugin>fake_components/GenericSystem</plugin>
</hardware>
```

Then the fake controllers get launched just like a real hardware controller from a launch file:

> ros2_controllers_path = os.path.join(
>     get_package_share_directory("moveit_resources_panda_moveit_config"),
>    "config",
>    "panda_ros_controllers.yaml",
> )
> ros2_control_node = Node(
>     package="controller_manager",
>     executable="ros2_control_node",
>     parameters=[robot_description, ros2_controllers_path],
> )

This is great! It's only a few words difference between hardware and simulation.

I also think it would be good to delete this from MoveIt2 sooner rather than later, to minimize the number of people using fake_controller_manager.

Tagging @v4hn and @rhaschke because they were the code owners in MoveIt1.


henningkayser (approved):
I fully agree. Making ros2_control GenericSystem the default minimizes the changes for running setups on hardware and it directly supports all kinds of controllers. I also don't see the use case for the fake controller anymore. I'd still wait for a second opinion by either one of the code owners.

.github/CODEOWNERS.disabled
```diff
@@ -33,7 +33,6 @@
 # FOLDER-SPECIFIC REVIEWERS:
 
 /moveit_plugins/moveit_ros_control_interface/               @ipa-mdl @bmagyar
```
JafarAbdi, line 35: Should we remove `moveit_ros_control_interface` .? we're not using it at all

tylerjw (approved):
This needs to be rebased for us to merge it.

.github/CODEOWNERS.disabled
AndyZe: How about in a new PR?

PR #479, opened 2021-05-27
Fix unused launch parameter

DLu:
### Description

I believe this is why CI isn't working. `robot_description_kinematics` was loaded but not used anywhere. 

### Checklist


vatanaksoytezer:
I think this may fix the bug where the test is actually failing in assert statement. The result file bug, which also caused this PR to fail is something else, and I opened up a PR to disable this test until I (or someone else) figures that out. See #481 and #482 .

henningkayser (commented):
You might be onto something here. Actually `move_group` is loading the kinematics yaml, but maybe it should be using the parameter you added here. I don't know if the controller manager actually uses this parameter, though.

Issue #475, opened 2021-05-24
Sync MoveIt 1 `master` into `main` pre 2.2


henningkayser:

JafarAbdi:
Fixed by https://github.com/ros-planning/moveit2/pull/505

PR #483, opened 2021-05-30
Port MoveIt1 PR 2682 (Fix collision normals and order of reported bodies)

AndyZe:
See https://github.com/ros-planning/moveit/pull/2682

This was a pretty straight-forward port

### DIFF
--- a/moveit_core/collision_detection/include/moveit/collision_detection/test_collision_common_panda.h
+++ b/moveit_core/collision_detection/include/moveit/collision_detection/test_collision_common_panda.h
@@ -304,7 +304,64 @@ TYPED_TEST_P(DistanceCheckPandaTest, DistanceSingle)
   }
 }
 
+template <class CollisionAllocatorType>
+class DistanceFullPandaTest : public CollisionDetectorPandaTest<CollisionAllocatorType>
+{
+};
+
+TYPED_TEST_CASE_P(DistanceFullPandaTest);
+
+/** \brief Tests if nearest points are computed correctly. */
+TYPED_TEST_P(DistanceFullPandaTest, DistancePoints)
+{
+  // Adding a box right in front of the robot hand
+  shapes::Box* shape = new shapes::Box(0.1, 0.1, 0.1);
+  shapes::ShapeConstPtr shape_ptr(shape);
+
+  Eigen::Isometry3d pos{ Eigen::Isometry3d::Identity() };
+  pos.translation().x() = 0.43;
+  pos.translation().y() = 0;
+  pos.translation().z() = 0.55;
+  this->cenv_->getWorld()->addToObject("box", shape_ptr, pos);
+
+  collision_detection::DistanceRequest req;
+  req.acm = &*this->acm_;
+  req.type = collision_detection::DistanceRequestTypes::SINGLE;
+  req.enable_nearest_points = true;
+  req.max_contacts_per_body = 1;
+
+  collision_detection::DistanceResult res;
+  this->cenv_->distanceRobot(req, res, *this->robot_state_);
+
+  // Checks a particular point is inside the box
+  auto checkInBox = [&](const Eigen::Vector3d& p) {
+    Eigen::Vector3d inBox = pos.inverse() * p;
+
+    constexpr double EPS = 1e-5;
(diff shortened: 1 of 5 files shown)


henningkayser (commented):
Thanks for preparing this. Is there anything special to this port that requires applying it separate from the [next sync](https://github.com/ros-planning/moveit2/issues/475)?

henningkayser:
@AndyZe could you confirm that this is correctly ported with #505 and then close this PR?

moveit_core/collision_detection/include/moveit/collision_detection/test_collision_common_panda.h
```diff
@@ -304,7 +304,64 @@ TYPED_TEST_P(DistanceCheckPandaTest, DistanceSingle)
   }
 }
 
+template <class CollisionAllocatorType>
+class DistanceFullPandaTest : public CollisionDetectorPandaTest<CollisionAllocatorType>
```
DLu, line 308: In #505 this is extended from `DistanceCheckPandaTest<CollisionAllocatorType>`

moveit_core/collision_detection_fcl/include/moveit/collision_detection_fcl/collision_env_fcl.h
```diff
@@ -154,6 +154,9 @@ class CollisionEnvFCL : public CollisionEnv
   /** \brief Callback function executed for each change to the world environment */
   void notifyObjectChange(const ObjectConstPtr& obj, World::Action action);
 
+  /** \brief Check if enable_nearest_distance_points is available */
+  void checkFCLNearestDistCapability(const DistanceRequest& req) const;
```
DLu, line 158: This is not added to #505

moveit_core/collision_detection_fcl/src/collision_common.cpp
```diff
@@ -585,8 +591,23 @@ bool distanceCallback(fcl::CollisionObjectd* o1, fcl::CollisionObjectd* o2, void
 #else
         dist_result.nearest_points[0] = Eigen::Vector3d(contact.pos.data.vs);
         dist_result.nearest_points[1] = Eigen::Vector3d(contact.pos.data.vs);
-        dist_result.normal = Eigen::Vector3d(contact.normal.data.vs);
+        Eigen::Vector3d normal(contact.normal.data.vs);
```
DLu, line 594: #505 does not set the normal

DLu (commented):
There are a number of places where #505 has a different implementation that you and @JafarAbdi should iron out.

moveit_core/collision_detection_fcl/src/collision_common.cpp
henningkayser: this looks like a bug, normal is redefined in the condition below

henningkayser: also, the logic seems to be fixed in the sync

moveit_core/collision_detection_fcl/include/moveit/collision_detection_fcl/collision_env_fcl.h
henningkayser: this one also doesn't exist in https://github.com/ros-planning/moveit/pull/2682/files

henningkayser:
To me it looks like the relevant changes were all added with #505 (at least with the latest update). Also, this PR does not really match the original changes. In general, backports and syncs should cherry-pick or merge the original changes to apply the same changes and maintain authorship. If there is still anything missing from #505, please add a review to that PR. Additional changes not part of the original contribution should be added in a fixup PR.

moveit_core/collision_detection/include/moveit/collision_detection/test_collision_common_panda.h
henningkayser: https://github.com/ros-planning/moveit/pull/2682/files also does so

AndyZe:
If I'd known that MoveIt1 PR's were still being systematically ported to MoveIt2, I wouldn't have opened this PR.

There are some small things I like better about this version. I think the function name `checkFCLNearestDistCapability()` is better than `checkFCLCapabilites()`, for example. But, I don't care enough to actually finish the PR   ;)

Issue #433, opened 2021-04-22
Hybrid Planning Feature


sjahr:
This issue tracks the current status of development for the new [hybrid planning feature for MoveIt 2](https://github.com/ros-planning/moveit2/tree/feature/hybrid_planning). After a basic version with a proof-of-concept demo was implemented last winter (#300), the current version of hybrid planning can be found on a [feature branch of MoveIt2](https://github.com/ros-planning/moveit2/tree/feature/hybrid_planning/moveit_ros/hybrid_planning). The next goal is to include the feature into the main version of MoveIt 2. The associated PR #488 is currently under review. Additionally, there is PR open,  which will add a [tutorial for the Hybrid Planning feature](https://github.com/sjahr/moveit2_tutorials/blob/add_hybrid_planning_tutorial/doc/hybrid_planning/hybrid_planning_tutorial.rst) and there is a [demo in progress ](https://github.com/PickNikRobotics/UR10e_welding_demo)that uses Hybrid Planning in a welding application. 

 - [X] Update use newest MoveIt 2 version and improvements: #426 
 - [ ] Create unit tests
 - [X] Refactor proof-of-concept plugins: #447 
 - [X] Improve documentation and a tutorial: [moveit2_tutorials #97](https://github.com/ros-planning/moveit2_tutorials/pull/97)
 - [x] Merge feature branch into main: #813 
 - [ ] Include more sophicticated algorithms into the hybrid planning pipeline
-  [x] Create a demo with real hardware: [UR10e_welding_demo](https://github.com/PickNikRobotics/UR10e_welding_demo)

Feel free to leave a comment if you have any ideas or questions!

AndyZe:
This has all been done now!  :+1:

PR #488, opened 2021-06-08
Add hybrid planning feature

sjahr:
### Description

This PR adds the new hybrid planning architecture to MoveIt2. The related issues that document the development process are #300 and #433.

To run the demo's you simply need to build moveit2 from source and make sure you use [sjahr/moveit_msgs's hybrid_planning branch](https://github.com/sjahr/moveit_msgs/tree/hybrid_planning) until the message definitions get merged into ros_planning/moveit_msgs.

The current setup includes a demo to verify the architecture's correct behavior. To run it use the following command:
```
ros2 launch moveit_hybrid_planning hybrid_planning.launch.py
```
The expected behavior looks like this:

replan

You can change the configuration of the hybrid planning architecture by switching the different plugins used by the architecture's main components in the associated [config file](https://github.com/ros-planning/moveit2/tree/feature/hybrid_planning/moveit_ros/hybrid_planning/test/config). 
 
### Checklist
- [X] Extend the tutorials / documentation [reference](http://moveit.ros.org/documentation/contributing/)


moveit_ros/hybrid_planning/local_planner/local_planner_component/include/moveit/local_planner/trajectory_operator_interface.h
```diff
@@ -0,0 +1,107 @@
...
+   */
+  virtual bool initialize(const rclcpp::Node::SharedPtr& node, const moveit::core::RobotModelConstPtr& robot_model,
+                          const std::string& group_name) = 0;
+
+  /**
+   * Add a new reference trajectory segment to the vector of global trajectory segments to process
+   * @param new_trajectory New reference trajectory segment to add
+   * @return True if segment was successfully added
+   */
+  virtual bool addTrajectorySegment(const robot_trajectory::RobotTrajectory& new_trajectory) = 0;
```
AdamPettinger, line 75: I think this function should include feedback (`moveit_msgs::action::LocalPlanner::Feedback`), perhaps as the return instead of a bool

moveit_ros/hybrid_planning/local_planner/local_planner_component/include/moveit/local_planner/trajectory_operator_interface.h
```diff
@@ -0,0 +1,107 @@
...
+
+  /**
+   * Return the current local constraints based on the newest robot state
+   * @param current_state Current RobotState
+   * @return Current local constraints that define the local planning goal
+   */
+  virtual robot_trajectory::RobotTrajectory getLocalTrajectory(const moveit::core::RobotState& current_state) = 0;
```
AdamPettinger, line 82: I think this function should also be able to give the local planner feedback, as right now nothing in the `trajectory_operator_interface` can give feedback (that can be used to trigger a global replanning, for instance). Consider: 
```suggestion
  virtual moveit_msgs::action::LocalPlanner::Feedback getLocalTrajectory(const moveit::core::RobotState& current_state, robot_trajectory::RobotTrajectory& local_trajectory) = 0;
```

moveit_ros/hybrid_planning/local_planner/trajectory_operator_plugins/src/simple_sampler.cpp
```diff
@@ -0,0 +1,121 @@
...
+    // Check if state reached
+    if (next_desired_goal_state.distance(current_state) <= 0.1)
+    {
+      // Update index (and thus desired robot state)
+      next_waypoint_index_ += 1;
+    }
+
+    // Construct local trajectory containing the next three global trajectory waypoints
+    local_trajectory.addSuffixWayPoint(reference_trajectory_->getWayPoint(next_waypoint_index_),
+                                       reference_trajectory_->getWayPointDurationFromPrevious(next_waypoint_index_));
```
AdamPettinger, line 103: I am pretty sure the comment needs updating to "the next global trajectory waypoint"?

moveit_ros/hybrid_planning/local_planner/trajectory_operator_plugins/src/simple_sampler.cpp
```diff
@@ -0,0 +1,121 @@
...
+  return true;
+}
+robot_trajectory::RobotTrajectory SimpleSampler::getLocalTrajectory(const moveit::core::RobotState& current_state)
+{
+  robot_trajectory::RobotTrajectory local_trajectory(reference_trajectory_->getRobotModel(),
+                                                     reference_trajectory_->getGroupName());
+  if (pass_through_)
+  {
+    // Use reference_trajectory as local trajectory
+    local_trajectory.swap(*reference_trajectory_);
```
AdamPettinger, line 87: If I set the `pass_through` parameter `true` in the config file, I get a segfault, I believe at this line

moveit_ros/hybrid_planning/test/config/hybrid_planning_manager.yaml
```diff
@@ -0,0 +1 @@
+planner_logic_plugin_name: "moveit_hybrid_planning/ReplanInvalidatedTrajectory"
```
AdamPettinger, line 1: Is it possible to combine all these config files into 1 file? Do we even want to?

moveit_ros/hybrid_planning/test/launch/hybrid_planning.launch.py
```diff
@@ -0,0 +1,199 @@
...
+            "stdout": "screen",
+            "stderr": "screen",
+        },
+    )
+
+    # Load controllers
+    load_controllers = []
+    for controller in [
+        "joint_state_controller",
+        "panda_joint_group_position_controller",
```
AdamPettinger, line 170: I would move to using the `panda_arm_controller` here as the default, because that is what the rest of the MoveIt tutorials and demos use

moveit_ros/hybrid_planning/test/config/demo_controller.yaml
```diff
@@ -0,0 +1,28 @@
+controller_manager:
+  ros__parameters:
+    update_rate: 500  # Hz
```
AdamPettinger, line 3: Changing this (to 100) has strange effects on the demo, which I cannot explain. Sometimes the robot moves a bit and then stops

moveit_ros/hybrid_planning/test/config/local_planner.yaml
```diff
@@ -0,0 +1,14 @@
...
+local_planning_frequency: 100.0
+global_solution_topic: "global_trajectory"
+local_solution_topic: "/panda_joint_group_position_controller/commands" # or panda_arm_controller/joint_trajectory
+local_solution_topic_type: "std_msgs/Float64MultiArray" # or trajectory_msgs/JointTrajectory
+
+# SimpleSampler param
+pass_through: false
+
+# ForwardTrajectory param
+stop_before_collision: false
```
AdamPettinger, line 14: I think this should be `true` so the demonstration runs like the provided gif.

If this is false, the demo robot passes through the objects when I try it

AdamPettinger (commented):
Went through mostly the local planner stuff, but skimmed the manager and global as well.

This looks great! Very clean, and I think the plugin architecture will allow a lot of flexibility

moveit_ros/hybrid_planning/test/config/demo_controller.yaml
```diff
@@ -0,0 +1,28 @@
...
+    panda_arm_controller:
+      type: joint_trajectory_controller/JointTrajectoryController
+    joint_state_controller:
+      type: joint_state_controller/JointStateController
+    panda_joint_group_position_controller:
+      type: position_controllers/JointGroupPositionController
+
+panda_arm_controller:
+  ros__parameters:
```
AdamPettinger, line 12: Need the joint names to use the other controller

```suggestion
  ros__parameters:
      joints:
      - panda_joint1
      - panda_joint2
      - panda_joint3
      - panda_joint4
      - panda_joint5
      - panda_joint6
      - panda_joint7
```

moveit_ros/hybrid_planning/test/config/demo_controller.yaml
sjahr: I've experienced the same thing and it started when I switched from fake joint to ros2_control based fake joint. This fixed it for me. Initially, I thought it might be related to an insufficient frequency of joint state updates but maybe I need to take a look at the connection between the local planner - and underlying controller frequency

moveit_ros/hybrid_planning/test/config/hybrid_planning_manager.yaml
sjahr: I guess this is debatable. My thought was that it might be more clearly arranged if I use one config file per component node especially when plugins are used, that have many parameters. But I don't have a strong opinion on this and I also see advantages in using one config file. It would be great to hear the opinion of other reviewers on this

moveit_ros/hybrid_planning/test/launch/hybrid_planning.launch.py
sjahr: I am not sure if the joint_trajectory_controller is well suited for high-frequent goal updates as it is the case for the replanning demo with point-wise trajectory forwarding. What are your experiences with the controller in the demo setup?

moveit_ros/hybrid_planning/CMakeLists.txt
```diff
@@ -0,0 +1,88 @@
...
+  control_toolbox
+)
+
+set(THIS_PACKAGE_INCLUDE_DIRS
+  global_planner/include
+  hybrid_planning_manager/hybrid_planning_manager_component/include
+  hybrid_planning_manager/planner_logic_plugins/include
+  local_planner/local_planner_component/include
+  local_planner/trajectory_operator_plugins/include
+  local_planner/local_constraint_solver_plugins/include
```
AdamPettinger, line 56:
```suggestion
  global_planner/include/
  hybrid_planning_manager/hybrid_planning_manager_component/include/
  hybrid_planning_manager/planner_logic_plugins/include/
  local_planner/local_planner_component/include/
  local_planner/trajectory_operator_plugins/include/
  local_planner/local_constraint_solver_plugins/include/
```

Without the trailing slash, these are installed at `install/moveit_hybrid_planning/include/include/moveit/.....` (notice the extra include) and other packages cannot find them without adding an include in the path

moveit_ros/hybrid_planning/CMakeLists.txt
```diff
@@ -0,0 +1,95 @@
...
+find_package(rclcpp_action REQUIRED)
+find_package(moveit_msgs REQUIRED)
+find_package(moveit_ros_planning REQUIRED)
+find_package(moveit_ros_planning_interface REQUIRED)
+find_package(moveit_core REQUIRED)
+find_package(std_srvs REQUIRED)
+find_package(std_msgs REQUIRED)
+find_package(tf2_ros REQUIRED)
+find_package(trajectory_msgs REQUIRED)
+find_package(control_toolbox REQUIRED)
```
AndyZe, line 21: Everything else is nicely alphabetized. You might as well alphabetize this, too.

(Also, check that control_toolbox is actually needed. I don't see it used anywhere)

moveit_ros/hybrid_planning/README.md
```diff
@@ -0,0 +1,23 @@
+# Hybrid Planning
+Implementation of MoveIt 2's new hybrid planning architecture. You find more information in the project's issues[#300](https://github.com/ros-planning/moveit2/issues/300), [#433](https://github.com/ros-planning/moveit2/issues/433)  and on the [MoveIt 2 roadmap](https://moveit.ros.org/documentation/contributing/roadmap/).
```
AndyZe, line 2:
```suggestion
A hybrid planning architecture. You can find more information in the project's issues[#300](https://github.com/ros-planning/moveit2/issues/300), [#433](https://github.com/ros-planning/moveit2/issues/433)  and on the [MoveIt 2 roadmap](https://moveit.ros.org/documentation/contributing/roadmap/).
```

AndyZe: Remove the word "new" because it won't be new for long. Best to keep time-based statements out of the documentation.

moveit_ros/hybrid_planning/README.md
```diff
@@ -0,0 +1,23 @@
...
+   with the local planner. In case the local planner detects a collision the global planner is rerun to update the
+   invalidated global trajectory.
+  - single_plan_execution: Run the global planner once and starts executing the global solution
+    with the local planner
+- trajectory operator plugins:
+  - simple_sampler: Samples the next global trajectory waypoint as local goal constraint
+   based on the current robot state. When the waypoint is reached the index that marks the current local goal constraint
+   is updated to the next global trajectory waypoint. Global trajectory updates simply replace the reference trajectory.
+- local solver plugins:
+  - forward_trajectory: Forwards the next waypoint of the sampled local trajectory.
```
AndyZe, line 22: LOL. So basically equivalent functionality with a normal MoveIt trajectory at this point. That's OK, I understand this is a work in progress.

moveit_ros/hybrid_planning/global_planner/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+add_subdirectory(global_planner_component)
+add_subdirectory(global_planner_plugins)
```
AndyZe, line 2: I like how CMakeLists subdirectories are set up  :+1:

moveit_ros/hybrid_planning/global_planner/global_planner_component/include/moveit/global_planner/global_planner_component.h
```diff
@@ -0,0 +1,87 @@
...
+class GlobalPlannerComponent : public rclcpp::Node
+{
+public:
+  GlobalPlannerComponent(const rclcpp::NodeOptions& options);
+
+  // TODO(sjahr) implement get_last_solution service
+
+private:
+  rclcpp::TimerBase::SharedPtr timer_;
+  bool initialized_{ false };
```
AndyZe, line 61: Usually we initialize everything all at once, all in the same place. That place is the constructor. So move this initialization to the initializer list.

moveit_ros/hybrid_planning/global_planner/global_planner_component/include/moveit/global_planner/global_planner_interface.h
```diff
@@ -0,0 +1,83 @@
...
+  /**
+   * Reset global planner plugin
+   * @return True if reset was successful
+   */
+  virtual bool reset() = 0;
+  virtual ~GlobalPlannerInterface(){};
+
+protected:
+  /** \brief Constructor */
+  GlobalPlannerInterface(){};
```
AndyZe, line 81: Griz has told me, this is not necessary for a pure virtual class.

`You don't need this as pure virtual classes cannot be instantiated`

moveit_ros/hybrid_planning/global_planner/global_planner_component/include/moveit/global_planner/global_planner_component.h
```diff
@@ -0,0 +1,87 @@
...
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Sebastian Jahr
```
AndyZe, line 35: Please add a description. (The same for every other header file).

With this particular file, I'm confused about the difference between global_planner_interface.h and global_planner_component.h

moveit_ros/hybrid_planning/global_planner/global_planner_component/src/global_planner_component.cpp
```diff
@@ -0,0 +1,160 @@
...
+using namespace std::chrono_literals;
+const rclcpp::Logger LOGGER = rclcpp::get_logger("global_planner_component");
+const std::string UNDEFINED = "<undefined>";
+
+GlobalPlannerComponent::GlobalPlannerComponent(const rclcpp::NodeOptions& options)
+  : Node("global_planner_component", options)
+{
+  // Initialize global planner after construction
+  // TODO(sjahr) Remove once life cycle component nodes are available
+  timer_ = this->create_wall_timer(1ms, [this]() {
```
AndyZe, line 57: Geez, that's a pretty bad hack. I guess you tested this and it's necessary? You can't just call init() once and expect it to work?

Prob a ROS2 issue.

moveit_ros/hybrid_planning/global_planner/global_planner_component/src/global_planner_component.cpp
```diff
@@ -0,0 +1,160 @@
...
+        timer_->cancel();
+        throw std::runtime_error("Failed to initialize global planner");
+      }
+    }
+  });
+}
+
+bool GlobalPlannerComponent::init()
+{
+  auto node_ptr = shared_from_this();
```
AndyZe, line 76: This is only used in one place (L124). Suggest moving closer to where it's used (L123).

moveit_ros/hybrid_planning/local_planner/local_planner_component/include/moveit/local_planner/local_constraint_solver_interface.h
```diff
@@ -0,0 +1,93 @@
...
+  /**
+   * Reset local constraint solver to some user-defined initial state
+   * @return True if reset was successful
+   */
+  virtual bool reset() = 0;
+  virtual ~LocalConstraintSolverInterface(){};
+
+protected:
+  /** \brief Constructor */
+  LocalConstraintSolverInterface(){};
```
AndyZe, line 91: Not necessary. Delete it.

moveit_ros/hybrid_planning/local_planner/local_planner_component/include/moveit/local_planner/local_constraint_solver_interface.h
```diff
@@ -0,0 +1,93 @@
...
+  solve(const robot_trajectory::RobotTrajectory& local_trajectory,
+        const std::shared_ptr<const moveit_msgs::action::LocalPlanner::Goal> local_goal,
+        trajectory_msgs::msg::JointTrajectory& local_solution) = 0;
+
+  /**
+   * Reset local constraint solver to some user-defined initial state
+   * @return True if reset was successful
+   */
+  virtual bool reset() = 0;
+  virtual ~LocalConstraintSolverInterface(){};
```
AndyZe, line 87: Probably not necessary (I'm not 100% sure). Please try deleting it.

moveit_ros/hybrid_planning/local_planner/local_planner_component/include/moveit/local_planner/local_planner_component.h
```diff
@@ -0,0 +1,194 @@
...
+  /**
+   * Initialize and start planning scene monitor to listen to the planning scene topic.
+   * Load trajectory_operator and constraint solver plugin.
+   * Initialize ROS 2 interfaces
+   * @return true if scene monitor and plugins are successfully initialized
+   */
+  bool initialize();
+
+  /**
+   * Handle the planners current job based on the internal state each loop run when the planner is started.
```
AndyZe, line 143:
```suggestion
   * Handle the planner's current job based on the internal state each iteration when the planner is started.
```

moveit_ros/hybrid_planning/local_planner/local_planner_component/include/moveit/local_planner/local_constraint_solver_interface.h
```diff
@@ -0,0 +1,93 @@
...
+  /**
+   * Initialize local constraint solver
+   * @return True if initialization was successful
+   */
+  virtual bool initialize(const rclcpp::Node::SharedPtr& node,
+                          const planning_scene_monitor::PlanningSceneMonitorPtr& planning_scene_monitor,
+                          const std::string& group_name) = 0;
+
+  /**
+   * Solve local planning problem for the current loop run
```
AndyZe, line 71: Please do a find-replace of "loop run" with "iteration".

Also check "loop-run".

I just don't hear that phrase in English. Must be a German->English translation issue, I guess.

moveit_ros/hybrid_planning/local_planner/local_planner_component/include/moveit/local_planner/trajectory_operator_interface.h
```diff
@@ -0,0 +1,113 @@
...
+   * @return A value between 0.0 (start) to 1.0 (completion).
+   */
+  virtual double getTrajectoryProgress(const moveit::core::RobotState& current_state) = 0;
+
+  /**
+   * Reset trajectory operator to some user-defined initial state
+   * @return True if reset was successful
+   */
+  virtual bool reset() = 0;
+  virtual ~TrajectoryOperatorInterface(){};
```
AndyZe, line 103: Again, delete the constructor and destructor

moveit_ros/hybrid_planning/global_planner/global_planner_component/include/moveit/global_planner/global_planner_interface.h
```diff
@@ -0,0 +1,83 @@
...
+   */
+  virtual moveit_msgs::msg::MotionPlanResponse plan(
+      const std::shared_ptr<rclcpp_action::ServerGoalHandle<moveit_msgs::action::GlobalPlanner>> global_goal_handle) = 0;
+
+  /**
+   * Reset global planner plugin
+   * @return True if reset was successful
+   */
+  virtual bool reset() = 0;
+  virtual ~GlobalPlannerInterface(){};
```
AndyZe, line 77: Not necessary. Delete this.

moveit_ros/hybrid_planning/hybrid_planning_manager/planner_logic_plugins/src/replan_invalidated_trajectory.cpp
```diff
@@ -0,0 +1,65 @@
...
+ *********************************************************************/
+
+/* Author: Sebastian Jahr
+ */
+
+#include <moveit/planner_logic_plugins/replan_invalidated_trajectory.h>
+
+namespace moveit_hybrid_planning
+{
+const rclcpp::Logger LOGGER = rclcpp::get_logger("hybrid_planning_manager");
```
AndyZe, line 42: I believe this should go in an anonymous namespace so it doesn't possibly conflict with other variables in the moveit_hybrid_planning namespace. Like this:

```
namespace {
  const rclcpp::Logger LOGGER = rclcpp::get_logger("hybrid_planning_manager");
}
```

Maybe @tylerjw can confirm?

moveit_ros/hybrid_planning/hybrid_planning_manager/planner_logic_plugins/src/replan_invalidated_trajectory.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+ReactionResult ReplanInvalidatedTrajectory::react(const std::string& event)
+{
+  if (event == "collision_ahead")
```
AndyZe, line 46: I'm surprised you are passing strings. Why not use enums?

AndyZe (commented):
I'll do another pass after you address this round of reviews. Looks good! I'm excited to see it develop further.

moveit_ros/hybrid_planning/README.md
sjahr: Yes, these are simple proof of concept plugins. Another motivation for this plugin was, that it should be possible to create MoveIt's "normal" sense-plan-act behavior with the architecture

moveit_ros/hybrid_planning/local_planner/local_planner_component/include/moveit/local_planner/local_constraint_solver_interface.h
sjahr: Ok, thanks for pointing that out. Iteration is much better

moveit_ros/hybrid_planning/hybrid_planning_manager/planner_logic_plugins/src/replan_invalidated_trajectory.cpp
sjahr: The MIGRATION_GUIDE.md suggests:
```
static const rclcpp::Logger LOGGER = rclcpp::get_logger("logger_name");
```
When I understand it correctly this does the same as the anonymous namespace. Is one variant more preferable? (and why?)

moveit_ros/hybrid_planning/global_planner/global_planner_component/src/global_planner_interface.cpp
```diff
@@ -0,0 +1,43 @@
...
+ *********************************************************************/
+
+/* Author: Sebastian Jahr
+ */
+
+#include <moveit/global_planner/global_planner_interface.h>
+
+namespace moveit_hybrid_planning
+{
+// Empty because this library only defines an interface
```
AndyZe, line 42: I've found out in my own journey to replace a ROS2 plugin -- you do not need a cpp file for the base class.

If you read the [ROS2 pluginlib tutorial](https://docs.ros.org/en/foxy/Tutorials/Pluginlib.html) carefully, you'll see that the base class (RegularPolygon) only has a header file. There's no library in the CMakeLists.

So, delete this file.

I think this comment applies to several places in this PR.

moveit_ros/hybrid_planning/hybrid_planning_manager/planner_logic_plugins/src/replan_invalidated_trajectory.cpp
AndyZe: My reasoning on strings vs enums is:

These strings are basically constants that are used in multiple files. If somebody were to change a string in one file but not another, the compiler would not warn them.

If an enum were used, they would get a compiler error.

moveit_ros/hybrid_planning/local_planner/trajectory_operator_plugins/src/simple_sampler.cpp
```diff
@@ -0,0 +1,136 @@
...
+                               const std::string& group_name)
+{
+  // Load parameter & initialize member variables
+  if (node->has_parameter("pass_through"))
+  {
+    node->get_parameter<bool>("pass_through", pass_through_);
+  }
+  else
+  {
+    pass_through_ = node->declare_parameter<bool>("pass_through", false);
```
tylerjw, line 59: Please add a filled out ParameterDescription to all parameter declarations.  This helps get information on use of the parameter through the ros interfaces.

moveit_ros/hybrid_planning/package.xml
```diff
@@ -0,0 +1,48 @@
+<package format="2">
+  <name>moveit_hybrid_planning</name>
+  <version>0.0.1</version>
+  <description>Hybrid planning components of MoveIt 2</description>
+  <author email="sebastian.jahr@picknik.ai">Sebastian Jahr</author>
+
+  <maintainer email="todo@to.do">TODO</maintainer>
```
tylerjw, line 7: Please set this to the release team.

```suggestion
  <maintainer email="moveit_releasers@googlegroups.com">MoveIt Release Team</maintainer>
```

moveit_ros/hybrid_planning/package.xml
```diff
@@ -0,0 +1,48 @@
...
+
+  <license>BSD</license>
+
+  <url type="website">http://moveit.ros.org</url>
+  <url type="bugtracker">https://github.com/ros-planning/moveit/issues</url>
+  <url type="repository">https://github.com/ros-planning/moveit</url>
```
tylerjw, line 13: moveit2

moveit_ros/hybrid_planning/package.xml
```diff
@@ -0,0 +1,48 @@
+<package format="2">
+  <name>moveit_hybrid_planning</name>
+  <version>0.0.1</version>
```
tylerjw, line 3: Please set this to the version used in the rest of the repo (all packages have the same version).

moveit_ros/hybrid_planning/test/config/demo_controller.yaml
tylerjw: Is this an int type and not a floating point?

moveit_ros/hybrid_planning/test/config/local_planner.yaml
```diff
@@ -0,0 +1,15 @@
+
```
tylerjw, line 1: remove blank line at start of file

moveit_ros/hybrid_planning/test/hybrid_planning_test_node.cpp
```diff
@@ -0,0 +1,276 @@
...
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Sebastian Jahr
+ */
```
tylerjw, line 36: description

moveit_ros/hybrid_planning/test/hybrid_planning_test_node.cpp
```diff
@@ -0,0 +1,276 @@
...
+#include <moveit_msgs/action/hybrid_planning.hpp>
+#include <moveit_msgs/msg/display_robot_state.hpp>
+#include <moveit_msgs/msg/motion_plan_response.hpp>
+
+using namespace std::chrono_literals;
+namespace
+{
+const rclcpp::Logger LOGGER = rclcpp::get_logger("test_hybrid_planning_client");
+}
+class HybridPlanningDemo
```
tylerjw, line 59: [nit] add space after close of namespace

moveit_ros/hybrid_planning/test/hybrid_planning_test_node.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+    box_1_.type = box_1_.BOX;
+    box_1_.dimensions = { 0.5, 0.8, 0.01 };
+
+    box_2_.type = box_2_.BOX;
+    box_2_.dimensions = { 1.0, 0.4, 0.01 };
+
+    // Add new collision object as soon as global trajectory is available.
+    global_solution_subscriber_ = node_->create_subscription<moveit_msgs::msg::MotionPlanResponse>(
+        "global_trajectory", 1, [this](const moveit_msgs::msg::MotionPlanResponse::SharedPtr msg) {
```
tylerjw, line 85: again, please consider using standard QoS and not a queue of 1

moveit_ros/hybrid_planning/test/hybrid_planning_test_node.cpp
```diff
@@ -0,0 +1,276 @@
...
+  }
+
+  void run()
+  {
+    RCLCPP_INFO(LOGGER, "Initialize Planning Scene Monitor");
+    tf_buffer_ = std::make_shared<tf2_ros::Buffer>(node_->get_clock());
+
+    planning_scene_monitor_ = std::make_shared<planning_scene_monitor::PlanningSceneMonitor>(
+        node_, "robot_description", tf_buffer_, "planning_scene_monitor");
+    if (planning_scene_monitor_->getPlanningScene() != nullptr)
```
tylerjw, line 125: why would this be a nullptr?  if it should you do something about that?

moveit_ros/hybrid_planning/test/hybrid_planning_test_node.cpp
```diff
@@ -0,0 +1,276 @@
...
+  std::shared_ptr<tf2_ros::Buffer> tf_buffer_;
+};
+
+int main(int argc, char** argv)
+{
+  rclcpp::init(argc, argv);
+  rclcpp::NodeOptions node_options;
+  node_options.automatically_declare_parameters_from_overrides(true);
+
+  rclcpp::Node::SharedPtr node = rclcpp::Node::make_shared("hybrid_planning_test_node", "", node_options);
```
tylerjw, line 264: why can't we use std::make_shared here?

moveit_ros/hybrid_planning/test/hybrid_planning_test_node.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+  HybridPlanningDemo demo(node);
+  std::thread run_demo([&demo]() {
+    rclcpp::sleep_for(5s);
```
tylerjw, line 268: this seems fairly fragile, have you tested this on slow machines?

moveit_ros/hybrid_planning/test/hybrid_planning_test_node.cpp
```diff
@@ -0,0 +1,276 @@
...
+    demo.run();
+  });
+
+  rclcpp::spin(node);
+  run_demo.join();
+
+  return 0;
```
tylerjw, line 275: rclcpp::shutdown() before returning from main

tylerjw:
The lack of unit tests to test the functionality inside the core classes of this is sad.  Please consider adding unit tests to this to verify the logic in the core classes of this.

moveit_ros/hybrid_planning/hybrid_planning_manager/planner_logic_plugins/src/replan_invalidated_trajectory.cpp
AndyZe: One way to do this:

```
enum ErrorCodeEnum
{
  NO_ERROR = 0,
  DESIRED_DURATION_TOO_SHORT = 1
};

/**
 * \brief Use this map to look up human-readable strings for each error code
 */
const std::unordered_map<uint, std::string> ERROR_CODE_MAP(
    { { NO_ERROR, "No error, trajectory generation was successful" },
      { DESIRED_DURATION_TOO_SHORT, "Desired duration is too short, cannot have less than one timestep in a "
                                    "trajectory" }
```


Usage:
```
    if (error_code != ErrorCodeEnum::NO_ERROR)
    {
      std::cout << "Error code: " << ERROR_CODE_MAP.at(error_code) << std::endl;
      return -1;
    }
```

.github/workflows/ci.yaml
```diff
@@ -8,7 +8,7 @@ on:
   pull_request:
   push:
     branches:
-      - main
```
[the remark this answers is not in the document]
sjahr: I'll add a TODO to remove this. As I cannot directly push on the feature branch, I need to open PRs against it and from my understanding, without this change, these PRs wouldn't get checked by CI

moveit_ros/hybrid_planning/CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
...
+install(TARGETS ${LIBRARIES} hybrid_planning_test_node
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
+  RUNTIME DESTINATION lib/${PROJECT_NAME}
+  INCLUDES DESTINATION include)
+
+install(DIRECTORY ${THIS_PACKAGE_INCLUDE_DIRS} DESTINATION include)
+
+install(DIRECTORY test/launch DESTINATION share/${PROJECT_NAME})
+install(DIRECTORY test/config DESTINATION share/${PROJECT_NAME})
```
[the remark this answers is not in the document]
sjahr: I've exported the Libraries a couple of lines below but I'll move that up and add an EXPORT_NAME for better clarity

moveit_ros/hybrid_planning/global_planner/global_planner_component/include/moveit/global_planner/global_planner_interface.h
```diff
@@ -0,0 +1,78 @@
...
+ * Class GlobalPlannerInterface - Base class for a global planner implementation
+ */
+class GlobalPlannerInterface
+{
+public:
+  /**
+   * Initialize global planner
+   * @return True if initialization was successful
+   */
+  virtual bool initialize(const rclcpp::Node::SharedPtr& node) = 0;
```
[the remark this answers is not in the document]
sjahr: Global planner plugin implementations need to inherit from this class and from my understanding, it is not possible to pass arguments to the constructor during the creation of a plugin instance with the plugin loader. Is that right?

moveit_ros/hybrid_planning/hybrid_planning_manager/planner_logic_plugins/src/replan_invalidated_trajectory.cpp
```diff
@@ -0,0 +1,68 @@
...
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Sebastian Jahr
```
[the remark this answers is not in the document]
sjahr: Do the source files also need a description when I give one in the associated header file? (I am aware that I did not do that consistently in all header files but I've fixed that in the next PR that will address these reviews).

moveit_ros/hybrid_planning/test/launch/hybrid_planning.launch.py
```diff
@@ -0,0 +1,199 @@
...
+    hybrid_planning_manager_param = load_yaml(
+        "moveit_hybrid_planning", "config/hybrid_planning_manager.yaml"
+    )
+
+    # Generate launch description with multiple components
+    container = ComposableNodeContainer(
+        name="hybrid_planning_container",
+        namespace="/",
+        package="rclcpp_components",
+        executable="component_container",
```
henningkayser, line 83: Taking note of https://github.com/ros-planning/moveit2/pull/723, you should probably also use `component_container_mt`

moveit_ros/hybrid_planning/global_planner/global_planner_component/include/moveit/global_planner/global_planner_component.h
```diff
@@ -0,0 +1,87 @@
...
+
+namespace moveit_hybrid_planning
+{
+// Component node containing the global planner
+class GlobalPlannerComponent : public rclcpp::Node
+{
+public:
+  GlobalPlannerComponent(const rclcpp::NodeOptions& options);
+
+  // TODO(sjahr) implement get_last_solution service
```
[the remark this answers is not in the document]
sjahr: I've removed it

AndyZe:
@sjahr I can't push to your branch. If I update and make a PR against your branch, can you merge it quickly? Otherwise I'll probably just open a new PR like Tyler did already with #763

Overall, this still works with a few tweaks but bit rot is starting to set in. Would like to get it merged sooner rather than later.

moveit_ros/hybrid_planning/global_planner/global_planner_component/include/moveit/global_planner/global_planner_component.h
```diff
@@ -0,0 +1,87 @@
...
+  rclcpp::TimerBase::SharedPtr timer_;
+  bool initialized_{ false };
+
+  std::string global_planner_name_;
+
+  // Global planner plugin loader
+  std::unique_ptr<pluginlib::ClassLoader<moveit_hybrid_planning::GlobalPlannerInterface>> global_planner_plugin_loader_;
+
+  // Global planner instance
+  std::shared_ptr<moveit_hybrid_planning::GlobalPlannerInterface> global_planner_instance_;
```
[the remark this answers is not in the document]
AndyZe: A unique_ptr here leads to an issue when creating the plugin here. I'm not quite sure how to fix this.

`global_planner_instance_ = global_planner_plugin_loader_->createUniqueInstance(global_planner_name_);`

The error:

`error: no match for ‘operator=’ (operand types are ‘std::unique_ptr<moveit_hybrid_planning::GlobalPlannerInterface>’ and ‘pluginlib::UniquePtr<moveit_hybrid_planning::GlobalPlannerInterface>’ {aka ‘std::unique_ptr<moveit_hybrid_planning::GlobalPlannerInterface, std::function<void(moveit_hybrid_planning::GlobalPlannerInterface*)> >’})`

moveit_ros/hybrid_planning/global_planner/global_planner_plugins/src/moveit_planning_pipeline.cpp
```diff
@@ -0,0 +1,138 @@
...
+  node->declare_parameter<std::string>(PLAN_REQUEST_PARAM_NS + "planning_pipeline", UNDEFINED);
+  node->declare_parameter<int>(PLAN_REQUEST_PARAM_NS + "planning_attempts", 5);
+  node->declare_parameter<double>(PLAN_REQUEST_PARAM_NS + "planning_time", 1.0);
+  node->declare_parameter<double>(PLAN_REQUEST_PARAM_NS + "max_velocity_scaling_factor", 1.0);
+  node->declare_parameter<double>(PLAN_REQUEST_PARAM_NS + "max_acceleration_scaling_factor", 1.0);
+
+  node_ptr_ = node;
+
+  // Initialize MoveItCpp API
+  moveit_cpp_ = std::make_shared<moveit_cpp::MoveItCpp>(node);
```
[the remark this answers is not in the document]
AndyZe: @henningkayser do you have an example of this type of usage? This doesn't work:

```
  moveit_cpp::MoveItCpp::Options moveit_cpp_options(node);
  moveit_cpp_ = std::make_shared<moveit_cpp::MoveItCpp>(node, moveit_cpp_options);
```

The error is:

```
[component_container-1] [ERROR] [1637183065.035467313] [moveit.ros_planning_interface.moveit_cpp]: Failed to load any planning pipelines.
[component_container-1] [FATAL] [1637183065.035482501] [moveit.ros_planning_interface.moveit_cpp]: Failed to load planning pipelines from parameter server
```

moveit_ros/hybrid_planning/global_planner/global_planner_component/include/moveit/global_planner/global_planner_component.h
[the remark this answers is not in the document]
henningkayser: pluginlib has its own unique_ptr type which basically provides unique access to the single instance managed by the plugin loader. So you either need to use `pluginlib::UniquePtr` or create an unmanaged plugin instance. Thinking about this more, I'm not sure if any of these versions are better than simply using shared_ptr though. @tylerjw any thoughts?

moveit_ros/hybrid_planning/global_planner/global_planner_plugins/src/moveit_planning_pipeline.cpp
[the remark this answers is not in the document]
henningkayser: To me it looks like no planning pipeline could be loaded which means that no pipelines were defined in the parameters. MoveItCpp doesn't use defaults here, so you need to verify the config. The options type expects the `planning_pipelines.pipeline_names` parameter to include the pipeline configuration namespaces, like "ompl" (https://github.com/ros-planning/moveit2/blob/main/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h#L89). Then, MoveItCpp tries to initialize a planning pipeline from that namespace. Both the `pipeline_names` and the actual plugin parameters must be present for this to work. See the moveit_cpp demo for an example.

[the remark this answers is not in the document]
AndyZe: I find an example in move_group.cpp. Loading the pipeline and moveit_cpp there takes about 99 lines. Yikes. This feels way harder than it should be.

https://github.com/ros-planning/moveit2/blob/b426a1c6f957522f8261f3500ee11f82199cdef6/moveit_ros/move_group/src/move_group.cpp#L217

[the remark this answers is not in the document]
AndyZe: Ah, I got it working. This is not so bad when it's figured out.

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/include/moveit/hybrid_planning_manager/hybrid_planning_manager.h
```diff
@@ -0,0 +1,132 @@
...
+   * @param success Indicates whether hybrid planning was successful
+   */
+  void sendHybridPlanningResponse(bool success);
+
+private:
+  // Planner logic plugin loader
+  std::unique_ptr<pluginlib::ClassLoader<moveit_hybrid_planning::PlannerLogicInterface>> planner_logic_plugin_loader_;
+
+  // Planner logic instance to implement reactive behavior
+  std::shared_ptr<moveit_hybrid_planning::PlannerLogicInterface> planner_logic_instance_;
```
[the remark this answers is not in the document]
AndyZe: Same issue with plugin loading, discussed above

moveit_ros/hybrid_planning/hybrid_planning_manager/planner_logic_plugins/src/replan_invalidated_trajectory.cpp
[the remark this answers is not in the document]
AndyZe: Agree with @sjahr, I don't think descriptions should be duplicated in .cpp and .h files. I'll make sure the .h files all have descriptions, though

AndyZe:
Closing in favor of #811. I'll still track whether the comments get addressed here

moveit_ros/hybrid_planning/CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
...
+find_package(std_msgs REQUIRED)
+find_package(std_srvs REQUIRED)
+find_package(tf2_ros REQUIRED)
+find_package(trajectory_msgs REQUIRED)
+
+set(LIBRARIES
+  # Components
+  moveit_global_planner_component
+  moveit_hybrid_planning_manager
+  moveit_local_planner_component
```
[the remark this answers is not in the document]
AndyZe: I don't have nearly as strong an aversion to the word Manager when it's used well. I think `moveit_hybrid_planning_manager` is more specific than `moveit_hybrid_planning_node`, for sure.

This manager is managing the interaction of the other 2 components so it makes perfect sense to me.

moveit_ros/hybrid_planning/hybrid_planning_manager/planner_logic_plugins/src/single_plan_execution.cpp
```diff
@@ -0,0 +1,93 @@
...
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Sebastian Jahr
+ */
```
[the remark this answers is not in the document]
AndyZe: Discussed above -- we'll not duplicate descriptions in headers and cpp

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/include/moveit/hybrid_planning_manager/planner_logic_interface.h
```diff
@@ -0,0 +1,148 @@
...
+class PlannerLogicInterface
+{
+public:
+  /**
+   * Initialize the planner logic
+   * @param hybrid_planning_manager The hybrid planning manager instance to initialize this logic with.
+   * @return true if initialization was successful
+   */
+  virtual bool
+  initialize(const std::shared_ptr<moveit_hybrid_planning::HybridPlanningManager>& hybrid_planning_manager) = 0;
```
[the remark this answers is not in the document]
AndyZe: That makes sense but this is a plugin

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/src/hybrid_planning_manager.cpp
```diff
@@ -0,0 +1,307 @@
...
+  // Load planning logic plugin
+  try
+  {
+    planner_logic_plugin_loader_ =
+        std::make_unique<pluginlib::ClassLoader<moveit_hybrid_planning::PlannerLogicInterface>>(
+            "moveit_hybrid_planning", "moveit_hybrid_planning::PlannerLogicInterface");
+  }
+  catch (pluginlib::PluginlibException& ex)
+  {
+    RCLCPP_ERROR(LOGGER, "Exception while creating planner logic plugin loader '%s'", ex.what());
```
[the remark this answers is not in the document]
AndyZe: @tylerjw I don't quite understand what you mean

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/src/hybrid_planning_manager.cpp
```diff
@@ -0,0 +1,307 @@
...
+    std::shared_ptr<rclcpp_action::ServerGoalHandle<moveit_msgs::action::HybridPlanning>> goal_handle)
+{
+  // Pass goal handle to class member
+  hybrid_planning_goal_handle_ = std::move(goal_handle);
+
+  // React on incoming planning request
+  ReactionResult reaction_result = planner_logic_instance_->react(
+      moveit_hybrid_planning::BasicHybridPlanningEvent::HYBRID_PLANNING_REQUEST_RECEIVED);
+  if (reaction_result.error_code.val != moveit_msgs::msg::MoveItErrorCodes::SUCCESS)
+  {
```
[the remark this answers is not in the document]
AndyZe: I think another PR to moveit_msgs and waiting for the review cycle there is way out of scope at this point

moveit_ros/hybrid_planning/local_planner/local_planner_component/include/moveit/local_planner/local_planner_component.h
```diff
@@ -0,0 +1,196 @@
...
+  READY = 1,
+  UNCONFIGURED = 2,
+  AWAIT_GLOBAL_TRAJECTORY = 3,
+  LOCAL_PLANNING_ACTIVE = 4
+};
+
+/**
+ * Class LocalPlannerComponent - ROS 2 component node that implements a local planner.
+ */
+class LocalPlannerComponent : public rclcpp::Node
```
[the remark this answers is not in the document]
AndyZe: Disagree about the naming. So you're proposing we name everything a **node**:

- LocalPlannerNode
- GlobalPlannerNode
- HybridPlanningNode

This is a lot more descriptive. HybridPlanning surely should not be named the same way as the other two:

- LocalPlannerComponent
- GlobalPlannerComponent
- HybridPlanningManager

I think you are following this rule of thumb about "name everything a node" too rigidly.

moveit_ros/hybrid_planning/global_planner/global_planner_component/src/global_planner_component.cpp
```diff
@@ -0,0 +1,164 @@
...
+    {
+      timer_->cancel();
+    }
+    else
+    {
+      initialized_ = this->init();
+      if (!initialized_)
+      {
+        timer_->cancel();
+        throw std::runtime_error("Failed to initialize global planner");
```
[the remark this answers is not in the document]
sjahr: Fixed with [b027f62](https://github.com/ros-planning/moveit2/pull/811/commits/b027f625442b7f53091ff3b7af6f26bae7a3e9fd)

moveit_ros/hybrid_planning/local_planner/local_planner_component/include/moveit/local_planner/local_planner_component.h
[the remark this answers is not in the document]
tylerjw: Thinking about this more, I like the suffix Component, but I still don't like the suffix Manager.  We should be able to find some word that is more descriptive than Manager.

Issue #473, opened 2021-05-22
Re-enable Bullet, finalize Migration & Tests


henningkayser:
Bullet is still [disabled](https://github.com/ros-planning/moveit2/blob/main/moveit_core/CMakeLists.txt#L180) at this point even though the port has been completed with #322. We should remove the `BULLET_ENABLE` flag by default and always compile it. I'm open to support a build flag that allows configuring what collision environment should be compiled. Things to consider:

* Should we switch to Bullet as default plugin instead of FCL?
* Do we want to move bullet + fcl into a separate package `moveit_collision_detection_plugins` for faster compile times (`moveit_core` is a bottle-neck for multi-core systems)

AndyZe:
I don't think Bullet is ready for prime time yet.

- Seg fault issue:  https://github.com/ros-planning/moveit/issues/2485

- Memory leak issue?  https://github.com/ros-planning/moveit_task_constructor/issues/232

- Also, I haven't personally seen any performance improvement.

PR #489, opened 2021-06-08
Removed the BULLET_ENABLE flag

jrgnicho:
### Description

Addresses issue #473
Removed the BULLET_ENABLE flag
Properly installs the bullet collision plugin

### Checklist

### DIFF
--- a/moveit_core/CMakeLists.txt
+++ b/moveit_core/CMakeLists.txt
@@ -18,6 +18,11 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/CMakeModules")
 # TODO: Move collision detection into separate packages
 find_package(Bullet 2.87 REQUIRED)
 
+if(BULLET_ENABLE)
+  set(BULLET_INC "collision_detection_bullet/include")
+  set(BULLET_LIB "moveit_collision_detection_bullet")
+endif()
+
 find_package(PkgConfig REQUIRED)
 
 pkg_check_modules(LIBFCL REQUIRED "fcl>=0.5.0")
@@ -128,7 +133,6 @@ set(THIS_PACKAGE_INCLUDE_DEPENDS
   Eigen3
   eigen3_cmake_module
   OCTOMAP
-  ${BULLET_ENABLE}
 )
 pluginlib_export_plugin_description_file(moveit_core collision_detector_fcl_description.xml)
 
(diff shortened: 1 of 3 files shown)


moveit_core/CMakeLists.txt
```diff
@@ -128,9 +127,10 @@ set(THIS_PACKAGE_INCLUDE_DEPENDS
   Eigen3
   eigen3_cmake_module
   OCTOMAP
-  ${BULLET_ENABLE}
```
henningkayser, line 131: Bullet needs to be listed as include depend, no?

moveit_core/collision_detection_bullet/CMakeLists.txt
```diff
@@ -25,29 +25,27 @@ target_link_libraries(${MOVEIT_LIB_NAME}
   ${BULLET_LIBRARIES}
```
henningkayser, line 25: Can you add `Bullet` to the `ament_target_dependencies` block above? The variables seem to match the right pattern, so this should work.

moveit_core/collision_detection_bullet/CMakeLists.txt
```diff
@@ -25,29 +25,27 @@ target_link_libraries(${MOVEIT_LIB_NAME}
...
-  ${MOVEIT_LIB_NAME}
-  moveit_planning_scene
-)
 
+install(DIRECTORY include/ DESTINATION include)
 install(TARGETS ${MOVEIT_LIB_NAME} EXPORT ${MOVEIT_LIB_NAME}
   TARGETS collision_detector_bullet_plugin EXPORT collision_detector_bullet_plugin
   LIBRARY DESTINATION lib
   ARCHIVE DESTINATION lib
   RUNTIME DESTINATION bin
```
henningkayser, line 47: I think the runtime destination should be `lib/${PROJECT_NAME}`, otherwise all executables will be accessible in the global bin path. That's a change that we should probably apply for the whole codebase.

henningkayser (commented):
Thanks for fixing this! I'm glad this really just seems to work without any code fixes. I'll test this for a bit before approving.

moveit_core/collision_detection_bullet/CMakeLists.txt
jrgnicho: Isn't that only necessary for ros2 dependencies?

moveit_core/CMakeLists.txt
jrgnicho: I think I omitted it because I was mirroring the FCL setup.  In fact the FCL include dependency is not in this list but I can add the Bullet one if that's what's preferred.

moveit_core/collision_detection_bullet/CMakeLists.txt
jrgnicho: OK, I can change this

moveit_core/collision_detection_bullet/CMakeLists.txt
henningkayser: > Isn't that only necessary for ros2 dependencies?

no, it works for all kinds of packages as long as they provide the right variables (_LIBRARIES, _INCLUDE_DIRS, ...).

henningkayser (approved):
@jrgnicho I hope you don't mind me hijacking this PR. The missing piece was that the bullet includes weren't configured correctly. `ament_target_dependencies` does this implicitly. The `SYSTEM` keyword was added to silence build warnings in bullet headers. I tested this with the MoveItCpp demo and it worked just fine for me.

I'd wait with merging this until the ROS 1 sync that @JafarAbdi is preparing is ready and merged.

jrgnicho:
@henningkayser thanks for fixing that, do you need me to test anything else?

nkalupahana:
Just an FYI, this PR breaks compilation on macOS. I'll bring in a fix when I have time to debug the issue (though @jrgnicho I would appreciate it if you could take a look at the CI logs to see if this is something you know how to fix off the bat). Obviously this isn't anyone's fault, since there's no macOS CI -- all it really shows is the need for it :) I'll make an issue for that later to track my progress. 

https://github.com/nkalupahana/ros2-foxy-macos/runs/2936771527?check_suite_focus=true

PR #492, opened 2021-06-09
follow_joint_trajectory_controller_handle: publish new multi_dof_trajectory field

JafarAbdi:
### Description

Use the new [multi_dof_trajectory](https://github.com/ros-controls/control_msgs/blob/galactic-devel/control_msgs/action/FollowJointTrajectory.action#L4) field, this have to be merged after galactic support

### Checklist


henningkayser (approved):
Changes look reasonable to me. It's a little odd that MoveIt would complain about multi-dof trajectories even if this should be up to the controller implementation. Do we actually have a controller that supports this? In the future we should probably remove or at least simplify the controller plugins now that we don't use the fake controller anymore.

DLu (approved):

moveit_plugins/moveit_simple_controller_manager/src/follow_joint_trajectory_controller_handle.cpp
```diff
@@ -48,18 +48,14 @@ bool FollowJointTrajectoryControllerHandle::sendTrajectory(const moveit_msgs::ms
...
-  }
-
   if (done_)
     RCLCPP_INFO_STREAM(LOGGER, "sending trajectory to " << name_);
   else
     RCLCPP_INFO_STREAM(LOGGER, "sending continuation for the currently executed trajectory to " << name_);
 
   control_msgs::action::FollowJointTrajectory::Goal goal = goal_template_;
   goal.trajectory = trajectory.joint_trajectory;
+  goal.multi_dof_trajectory = trajectory.multi_dof_joint_trajectory;
```
henningkayser, line 58: so... this doesn't seem to exist?

JafarAbdi: This only exists in galactic

henningkayser: ok, so we need to hold this for now until we branch off Foxy

PR #493, opened 2021-06-10
Solve the problem that the click stop during the execute path process in MotionPlanning cannot respond in time

vimior:
Problem: In the operation of rviz->MotionPlanning->planning, when you click stop during the execute process, you cannot respond in time (the corresponding stop operation will be executed after the execute ends)

PR solution: add a Node to subscribe (stop) event, and spin operation in a new thread


JafarAbdi (commented):
The cause of the bug is the default callback group of the node is MutuallyExclusive which prevent two callback from being called at the same time, I don't think adding a new node & thread is a good solution for this problem, I created a new PR to fix it, do you mind reviewing it .?

Thanks!

vimior:
ok, i pull and test at next work day.
Thanks！

Issue #481, opened 2021-05-27
Fix failing OMPL constraint test


vatanaksoytezer:
Fix failing OMPL constraint test

jrgnicho:
I'm getting a CI error reporting a failure in a OMPL constraint unit test
https://github.com/ros-planning/moveit2/pull/489/checks?check_run_id=2775556215

vatanaksoytezer:
Thanks for reporting Jorge @jrgnicho this is one of the newer flaky tests and will work on a fix very soon. Meanwhile you can just re-trigger the CI

vatanaksoytezer:
The disabled OMPL test was under planning interface, which should be solved similar with #397

PR #495, opened 2021-06-10
Disable actually failing flaky OMPL test

vatanaksoytezer:
Disable actually failing flaky OMPL test that causes CI failures all over moveit2. See https://github.com/ros-planning/moveit2/runs/2798018498?check_suite_focus=true for a sample output.

### DIFF
--- a/moveit_planners/ompl/ompl_interface/CMakeLists.txt
+++ b/moveit_planners/ompl/ompl_interface/CMakeLists.txt
@@ -82,9 +82,11 @@ if(BUILD_TESTING)
   ament_target_dependencies(test_planning_context_manager moveit_core tf2_eigen OMPL Boost Eigen3)
   target_link_libraries(test_planning_context_manager ${MOVEIT_LIB_NAME})
 
-  ament_add_gtest(test_ompl_constraints test/test_ompl_constraints.cpp)
-  ament_target_dependencies(test_ompl_constraints moveit_core OMPL Boost Eigen3)
-  target_link_libraries(test_ompl_constraints ${MOVEIT_LIB_NAME})
+  # Disabling flaky test
+  # TODO (vatanaksoytezer): Uncomment once this is fixed
+  # ament_add_gtest(test_ompl_constraints test/test_ompl_constraints.cpp)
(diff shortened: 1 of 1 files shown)


vatanaksoytezer:
Disables #481

henningkayser (approved):

davetcoleman:
>  # TODO (vatanaksoytezer): Uncomment once this is fixed

I think this is a pretty bad habit to get into... the chances of you ever re-enabling this flakey test is close to zero. Did a Clubhouse story get created to tackle this issue in the next sprint? Do we have a plan to fix it? 

MoveIt already has low test coverage, but disabling existing ones only makes that worse. 

Tagging @mamoll as this pertains to OMPL.

vatanaksoytezer:
> I think this is a pretty bad habit to get into... the chances of you ever re-enabling this flakey test is close to zero. Did a Clubhouse story get created to tackle this issue in the next sprint? Do we have a plan to fix it?

Thanks for the heads up @davetcoleman. I have a clubhouse story to re-enable this. The reason I suggested to disable this was we didn't know the cause and this flakiness was becoming an issue with a lot of PRs failing in CI in the release sprint. https://github.com/ros-planning/moveit2/pull/511 has more information from my debugging session. I think @mamoll maybe the right person to look into this.

For the OMPL and servo tests that we disabled I am trying to look whenever I can and really want to enable them back asap.

v4hn:
just a side remark: if you have to disable tests, you should usually **not** disable compiling the whole unit.
Instead you can [selectively disable the affected gtests](https://stackoverflow.com/questions/7208070/googletest-how-to-skip-a-test). The code will still be compiled (thus break on API changes) and during execution gtest explicitly mentions the disabled tests.

vatanaksoytezer:
> just a side remark: if you have to disable tests, you should usually **not** disable compiling the whole unit.
> Instead you can [selectively disable the affected gtests](https://stackoverflow.com/questions/7208070/googletest-how-to-skip-a-test). The code will still be compiled (thus break on API changes) and during execution gtest explicitly mentions the disabled tests.

Thanks a lot @v4hn. Definitely learned something new today! I will make sure I do it this way if we need to disable any test in the future. I might take my time to change this as well if I can't figure out a fix soon.

PR #497, opened 2021-06-14
moveit_servo: Fix segfault when publish_joint_velocities set to false and a joint is close to position limit

JafarAbdi:
### Description

moveit_servo is segfaulting when reaching limits if `publish_joint_velocities` set to `false`. the reason is in [ServoCalcs::enforcePositionLimits](https://github.com/ros-planning/moveit2/blob/main/moveit_ros/moveit_servo/src/servo_calcs.cpp#L780) we do access the velocities despite being an empty vector [see](https://github.com/ros-planning/moveit2/blob/main/moveit_ros/moveit_servo/src/servo_calcs.cpp#L810-L813)

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/src/servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/src/servo_calcs.cpp
@@ -578,16 +578,16 @@ bool ServoCalcs::internalServoUpdate(Eigen::ArrayXd& delta_theta,
   // Mark the lowpass filters as updated for this cycle
   updated_filters_ = true;
 
-  // compose outgoing message
-  composeJointTrajMessage(internal_joint_state_, joint_trajectory);
-
   // Enforce SRDF position limits, might halt if needed, set prev_vel to 0
-  if (!enforcePositionLimits(joint_trajectory))
+  if (!enforcePositionLimits(internal_joint_state_))
   {
-    suddenHalt(joint_trajectory);
+    suddenHalt(internal_joint_state_);
     status_ = StatusCode::JOINT_BOUND;
     prev_joint_velocity_.setZero();
   }
+
+  // compose outgoing message
+  composeJointTrajMessage(internal_joint_state_, joint_trajectory);
 
   // Modify the output message if we are using gazebo
   if (parameters_->use_gazebo)
@@ -777,7 +777,7 @@ void ServoCalcs::enforceVelLimits(Eigen::ArrayXd& delta_theta)
   delta_theta = velocity_scaling_factor * velocity * parameters_->publish_period;
 }
 
-bool ServoCalcs::enforcePositionLimits(trajectory_msgs::msg::JointTrajectory& joint_trajectory) const
+bool ServoCalcs::enforcePositionLimits(sensor_msgs::msg::JointState& joint_state) const
 {
   // Halt if we're past a joint margin and joint velocity is moving even farther past
   bool halting = false;
@@ -803,21 +803,13 @@ bool ServoCalcs::enforcePositionLimits(trajectory_msgs::msg::JointTrajectory& jo
       if (!limits.empty())
       {
         // Check if pending velocity command is moving in the right direction
-        auto joint_itr =
-            std::find(joint_trajectory.joint_names.begin(), joint_trajectory.joint_names.end(), joint->getName());
-        auto joint_idx = std::distance(joint_trajectory.joint_names.begin(), joint_itr);
+        auto joint_itr = std::find(joint_state.name.begin(), joint_state.name.end(), joint->getName());
+        auto joint_idx = std::distance(joint_state.name.begin(), joint_itr);
 
-        const bool near_min_position =
-            parameters_->publish_joint_velocities ?
-                joint_trajectory.points[0].velocities[joint_idx] < 0 &&
-                    (joint_angle < (limits[0].min_position + parameters_->joint_limit_margin)) :
-                joint_angle < (limits[0].min_position + parameters_->joint_limit_margin);
-        const bool near_max_position =
-            parameters_->publish_joint_velocities ?
-                joint_trajectory.points[0].velocities[joint_idx] > 0 &&
-                    (joint_angle > (limits[0].max_position - parameters_->joint_limit_margin)) :
-                joint_angle > (limits[0].max_position - parameters_->joint_limit_margin);
-        if (near_min_position || near_max_position)
+        if ((joint_state.velocity.at(joint_idx) < 0 &&
+             (joint_angle < (limits[0].min_position + parameters_->joint_limit_margin))) ||
+            (joint_state.velocity.at(joint_idx) > 0 &&
+             (joint_angle > (limits[0].max_position - parameters_->joint_limit_margin))))
         {
           rclcpp::Clock& clock = *node_->get_clock();
           RCLCPP_WARN_STREAM_THROTTLE(LOGGER, clock, ROS_LOG_THROTTLE_PERIOD,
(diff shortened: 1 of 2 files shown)


AndyZe (commented):
This makes a lot of sense. Let me test it.

schornakj (commented):
I tested this and it does stop Servo from segfaulting when I reach a joint position limit.

However, once I'm at a limit I'm not able to move back in the opposite direction or move any other joint in the robot. I get this warning message:

```
[moveit_servo.servo_calcs]: servo_server wrist_3_joint close to a position limit. Halting.
```
Seems like this is related to the conditional that sets `halting = true`. I think it should check if the commanded direction of rotation is still going towards the limit and still allow rotation if it goes away from the limit. Could that be resolved within the scope of this PR or does that require a lot of extra work?

AndyZe:
Maybe we should always store joint position/vel/accel internally so we don't have to worry about this type of thing. Then `publish_joint_velocities` and `publish_joint_positions` would only be used on the outgoing message. Does that seem like a good path forward?

AdamPettinger:
I will test it too, but seems reasonable

> I think it should check if the commanded direction of rotation is still going towards the limit and still allow rotation if it goes away from the limit. Could that be resolved within the scope of this PR or does that require a lot of extra work?

I think the original (seg faulting) code here was checking for this, so I would say the same behavior should be included in this PR

JafarAbdi:
> I tested this and it does stop Servo from segfaulting when I reach a joint position limit.
> 
> However, once I'm at a limit I'm not able to move back in the opposite direction or move any other joint in the robot. I get this warning message:
> 
> ```
> [moveit_servo.servo_calcs]: servo_server wrist_3_joint close to a position limit. Halting.
> ```
> 
> Seems like this is related to the conditional that sets `halting = true`. I think it should check if the commanded direction of rotation is still going towards the limit and still allow rotation if it goes away from the limit. Could that be resolved within the scope of this PR or does that require a lot of extra work?

This also happens to me when the robot reaches collisions or joint limits and once it's there it will be stuck, but I think this's outside the scope of this PR since it requires changing the halting behavior

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -807,10 +807,17 @@ bool ServoCalcs::enforcePositionLimits(trajectory_msgs::msg::JointTrajectory& jo
             std::find(joint_trajectory.joint_names.begin(), joint_trajectory.joint_names.end(), joint->getName());
         auto joint_idx = std::distance(joint_trajectory.joint_names.begin(), joint_itr);
 
-        if ((joint_trajectory.points[0].velocities[joint_idx] < 0 &&
-             (joint_angle < (limits[0].min_position + parameters_->joint_limit_margin))) ||
-            (joint_trajectory.points[0].velocities[joint_idx] > 0 &&
-             (joint_angle > (limits[0].max_position - parameters_->joint_limit_margin))))
+        const bool near_min_position =
+            parameters_->publish_joint_velocities ?
+                joint_trajectory.points[0].velocities[joint_idx] < 0 &&
```
tylerjw, line 812: If you change all the `operator[]` to `at()` it should help catch any future occurrences of this too.

AndyZe:
> This also happens to me when the robot reaches collisions or joint limits and once it's there it will be stuck, but I think this's outside the scope of this PR since it requires changing the halting behavior

However, there is a regression because it could reverse away from joint limits previously. Sorry, but I think we'll have to change the way this PR works. I'll make a pull request against your branch, @JafarAbdi.

AdamPettinger:
> > This also happens to me when the robot reaches collisions or joint limits and once it's there it will be stuck, but I think this's outside the scope of this PR since it requires changing the halting behavior
> 
> However, there is a regression because it could reverse away from joint limits previously. Sorry, but I think we'll have to change the way this PR works. I'll make a pull request against your branch, @JafarAbdi.

@AndyZe if it is not too late, I think a workable solution is to switch the order `composeJointTrajMessage` and `enforcePositionLimits` is called so that `enforcePositionLimits` works on a JointState and uses `internal_joint_state_` which holds position and velocity information regardless of what should be published

AndyZe:
@AdamPettinger do you mind making a PR? That sounds better than the way I did it.

AdamPettinger:
Yep! See https://github.com/JafarAbdi/moveit2/pull/13

AdamPettinger (approved):
Does joint limit checking probably the way it should have been all along :sweat_smile: 

Tested, and doesn't seg fault when `publish_joint_velocities` is false, and also allows backing away from a joint limit

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -862,6 +861,13 @@ void ServoCalcs::suddenHalt(trajectory_msgs::msg::JointTrajectory& joint_traject
   }
 }
 
+void ServoCalcs::suddenHalt(sensor_msgs::msg::JointState& joint_state) const
+{
+  // Set the position to the original position, and velocity to 0
+  joint_state.position = original_joint_state_.position;
+  joint_state.velocity.assign(joint_state.position.size(), 0.0);
```
nbbrooks, line 868: Is joint_state.velocity.size() always the size of the joints? (i.e. do we ever not provide joint_state.velocity values? If so should we check that before populating the velocity fields here?)

nbbrooks (commented):
+1 switch to at()

moveit_ros/moveit_servo/src/servo_calcs.cpp
AdamPettinger: The `original_joint_state_` [contains only active joint](https://github.com/ros-planning/moveit2/blob/eb2c8b97862e9acb760dd0c7882d06d04e4ecfae/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h#L284-L285) so this should return a joint state that only includes the active joints, which is what the rest of the calculations are expecting

nbbrooks: I think a better phrased question is "will incoming_joint_state_ always have velocity information"?

e.g. is there a servo configuration where `joint_state.velocity` would be empty here and we may be introducing side effects in ros control by resizing velocity?

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -785,17 +785,17 @@ bool ServoCalcs::enforcePositionLimits(sensor_msgs::msg::JointState& joint_state
 
   for (auto joint : joint_model_group_->getActiveJointModels())
   {
-    for (std::size_t c = 0; c < original_joint_state_.name.size(); ++c)
+    for (std::size_t c = 0; c < joint_state.name.size(); ++c)
     {
       // Use the most recent robot joint state
-      if (original_joint_state_.name[c] == joint->getName())
+      if (joint_state.name[c] == joint->getName())
```
AndyZe, line 791: It makes sense to use Servo's internal join position here, rather than `original_joint_state_`. We do the same thing for velocity at L809  :+1:

moveit_ros/moveit_servo/src/servo_calcs.cpp
AndyZe: I can't think of any weird edge cases where the positions vector and the velocities vector would be different sizes (other than if one of them is completely empty).

Also, `internal_joint_state_` previously had the velocities filled ALWAYS while internal calculations are happening. So this looks OK to me. (https://github.com/ros-planning/moveit2/blob/eb2c8b97862e9acb760dd0c7882d06d04e4ecfae/moveit_ros/moveit_servo/src/servo_calcs.cpp#L158)

Velocities get dropped at L673 before the final message goes out, if configured to do so:

>   if (parameters_->publish_joint_velocities)
>     point.velocities = joint_state.velocity;

AndyZe (approved):

PR #500, opened 2021-06-16
Address hybrid planning review

sjahr:
### Description

Address changes requested by @AdamPettinger 

### Checklist

### DIFF
--- a/moveit_ros/hybrid_planning/global_planner/src/global_planner_component.cpp
+++ b/moveit_ros/hybrid_planning/global_planner/src/global_planner_component.cpp
@@ -80,6 +80,7 @@ GlobalPlannerComponent::GlobalPlannerComponent(const rclcpp::NodeOptions& option
   global_trajectory_pub_ = this->create_publisher<moveit_msgs::msg::MotionPlanResponse>("global_trajectory", 1);
 
   // Initialize global planner after construction
+  // TODO(sjahr) Remove once life cycle component nodes are available
   timer_ = this->create_wall_timer(1ms, [this]() {
     if (initialized_)
     {
--- a/moveit_ros/hybrid_planning/global_planner/src/global_planner_component.cpp
+++ b/moveit_ros/hybrid_planning/global_planner/src/global_planner_component.cpp
@@ -164,7 +164,6 @@ void GlobalPlannerComponent::globalPlanningRequestCallback(
   const auto goal = goal_handle->get_goal();
 
   // Plan global trajectory
-  RCLCPP_INFO(LOGGER, "Start global planning");
   moveit_msgs::msg::MotionPlanResponse planning_solution = plan(goal->request);
 
   // Publish global planning solution to the local planner
--- a/moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/include/moveit/hybrid_planning_manager/planner_logic_interface.h
+++ b/moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/include/moveit/hybrid_planning_manager/planner_logic_interface.h
@@ -39,11 +39,76 @@
 #pragma once
 
 #include <rclcpp/rclcpp.hpp>
-
+#include <moveit_msgs/msg/move_it_error_codes.hpp>
 #include <moveit/hybrid_planning_manager/hybrid_planning_events.h>
 
 namespace moveit_hybrid_planning
 {
+// TODO(sjahr): Move this into utility package
+class MoveItErrorCode : public moveit_msgs::msg::MoveItErrorCodes
+{
+public:
+  MoveItErrorCode()
+  {
+    val = 0;
+  }
+  MoveItErrorCode(int code)
+  {
+    val = code;
+  }
+  MoveItErrorCode(const moveit_msgs::msg::MoveItErrorCodes& code)
+  {
+    val = code.val;
+  }
+  explicit operator bool() const
+  {
+    return val == moveit_msgs::msg::MoveItErrorCodes::SUCCESS;
+  }
+  bool operator==(const int code) const
+  {
+    return val == code;
+  }
+  bool operator!=(const int code) const
+  {
+    return val != code;
+  }
+};
+
+// Describes the outcome of a reaction to an event in the hybrid planning architecture
+struct ReactionResult
+{
+  ReactionResult(const BasicHybridPlanningEvent& planning_event, const std::string& error_msg, const int& error_code)
+    : error_message(error_msg), error_code(error_code)
+  {
+    switch (planning_event)
+    {
+      case moveit_hybrid_planning::BasicHybridPlanningEvent::HYBRID_PLANNING_REQUEST_RECEIVED:
+        event = "Hybrid planning request received";
+        break;
+      case moveit_hybrid_planning::BasicHybridPlanningEvent::GLOBAL_PLANNING_ACTION_FINISHED:
+        event = "Global planning action finished";
+        break;
+      case moveit_hybrid_planning::BasicHybridPlanningEvent::GLOBAL_SOLUTION_AVAILABLE:
+        event = "Global solution available";
+        break;
+      case moveit_hybrid_planning::BasicHybridPlanningEvent::LOCAL_PLANNING_ACTION_FINISHED:
+        event = "Local planning action finished";
+        break;
+    }
+  };
+  ReactionResult(const std::string& event, const std::string& error_msg, const int& error_code)
+    : event(event), error_message(error_msg), error_code(error_code){};
+
+  // Event that triggered the reaction
+  std::string event;
+
+  // Additional error description
+  std::string error_message;
+
+  // Error code
+  MoveItErrorCode error_code;
+};
+
 class HybridPlanningManager;  // Forward declaration
 
 /**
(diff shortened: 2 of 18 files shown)


moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
```diff
@@ -261,21 +268,25 @@ void LocalPlannerComponent::executePlanningLoopRun()
...
 
-      if (!local_feedback->feedback.empty())
+      if (!local_planner_feedback_->feedback.empty())
       {
-        local_planning_goal_handle_->publish_feedback(local_feedback);
+        local_planning_goal_handle_->publish_feedback(local_planner_feedback_);
       }
 
-      // Use the same message interface to controllers like MoveIt servo to use application dependend controller
+      // Use the configuarable message interface like MoveIt serve
```
henningkayser, line 289:
```suggestion
      // Use the configurable message interface like MoveIt servo
```

gavanderhoorn:
In the context of https://github.com/ros-planning/moveit.ros.org/pull/603 and https://github.com/ros-planning/moveit.ros.org/pull/606: something like e911df8 immediately complicates actual RT behaviour of MoveIt. Unwinding stacks when dealing with exceptions is not RT safe. It's possible to fix it (see the Apex AI presentation for one such approach), but by default exceptions are typically considered RT unsafe.

moveit_ros/hybrid_planning/local_planner/trajectory_operator_plugins/src/simple_sampler.cpp
```diff
@@ -77,14 +80,18 @@ bool SimpleSampler::reset()
...
-                                                     reference_trajectory_->getGroupName());
+  // Delete previous local trajectory
+  local_trajectory.clear();
+
+  // Determine current local trajectory based on configured behavior
   if (pass_through_)
   {
     // Use reference_trajectory as local trajectory
-    local_trajectory.swap(*reference_trajectory_);
+    local_trajectory = robot_trajectory::RobotTrajectory(*reference_trajectory_);
```
henningkayser, line 94: So, this constructs a new robot trajectory each time, so clearing `local_trajectory` would be redundant. But I'd guess that clearing and then copying the waypoints over would be more efficient.

moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
```diff
@@ -181,13 +183,18 @@ bool LocalPlannerComponent::initialize()
         moveit::core::RobotState start_state(planning_scene_monitor_->getRobotModel());
         moveit::core::robotStateMsgToRobotState(msg->trajectory_start, start_state);
         new_trajectory.setRobotTrajectoryMsg(start_state, msg->trajectory);
-        trajectory_operator_instance_->addTrajectorySegment(new_trajectory);
+        *local_planner_feedback_ = trajectory_operator_instance_->addTrajectorySegment(new_trajectory);
```
henningkayser, line 186: Not sure sharing and initializing the feedback like this is ideal, but I don't have a better idea right now. Also, feedback messages like this should probably not be empty making the check below unnecessary.

moveit_ros/hybrid_planning/global_planner/src/global_planner_component.cpp
```diff
@@ -136,15 +137,15 @@ bool GlobalPlannerComponent::init()
   {
     if (planning_pipelines_.count(planning_pipeline_name) > 0)
     {
-      RCLCPP_WARN(LOGGER, "Skipping duplicate entry for planning pipeline '%s'.", planning_pipeline_name.c_str());
+      RCLCPP_WARN(LOGGER, "Skipping duplicate entry for planning pipeline \'%s\'.", planning_pipeline_name.c_str());
```
henningkayser, line 140: Why are the escapes necessary? We haven't used them before, afaik

henningkayser:
> In the context of [ros-planning/moveit.ros.org#603](https://github.com/ros-planning/moveit.ros.org/pull/603) and [ros-planning/moveit.ros.org#606](https://github.com/ros-planning/moveit.ros.org/pull/606): something like [e911df8](https://github.com/ros-planning/moveit2/commit/e911df8a503ca76bc3042d5e344dbd120683dce8) immediately complicates actual RT behaviour of MoveIt. Unwinding stacks when dealing with exceptions is not RT safe. It's possible to fix it (see the Apex AI presentation for one such approach), but by default exceptions are typically considered RT unsafe.
> 
> Edit: this obviously does not mean anything in the context of the review of this PR, or devalues the work by @sjahr, so I'll mark this comment as _off-topic_.
> 
> I thought it would be interesting to point this out though, given our recent discussions about this topic.

@gavanderhoorn I think that's a really good point! I'm also not convinced that exeptions are the way to go here. @sjahr what was the reasoning to use exceptions over a result type that wraps an error code and result message?

moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
henningkayser: or "Use a ..."?

gavanderhoorn:
I'm not claiming exceptions are bad necessarily.

I just wanted to point out this PR alone makes the claim which was added wrt RT-ness of MoveIt 2 (in https://github.com/ros-planning/moveit.ros.org/pull/603 and https://github.com/ros-planning/moveit.ros.org/pull/606) contentious.

And it is possible to fix/work-around that -- as evidenced by the use of exceptions in Apex' code -- but it will require special care, which is not something done in MoveIt 2.

My comment was really more of a `don't-make-such-claims-too-quickly` comment than anything else.

henningkayser:
> I just wanted to point out this PR alone makes the claim which was added wrt RT-ness of MoveIt 2 (in [ros-planning/moveit.ros.org#603](https://github.com/ros-planning/moveit.ros.org/pull/603) and [ros-planning/moveit.ros.org#606](https://github.com/ros-planning/moveit.ros.org/pull/606)) contentious.

I see, you're absolutely right. We didn't describe properly what "realtime" even means in context of MoveIt, and there has already been a huge confusion between RT-computing and -messaging constraints and "realtime motion planning" as used in reasearch (not necessarily including RTC). I think the now fixed wording is more clear about this? 
 
> And it is possible to fix/work-around that -- as evidenced by the use of exceptions in Apex' code -- but it will require special care, which is not something done in MoveIt 2.

Thanks for pointing us at it, I'll review in what context this would be applicable for us.

> My comment was really more of a `don't-make-such-claims-too-quickly` comment than anything else.
> 
> Edit: I'm also not obsessed with the RT claim. It's just that this happens to pop-up, and it's a claim which is easy to _negate_.

It's an important point, though. The original claim was too general and vague and therefore somewhat deceptive. Official documentation and posts should be more exact and specific in its terminology.

sjahr:
Thanks a lot for this valuable input @gavanderhoorn! I was not aware of this when I decided to change it. My intention was to improve debugging options by this, especially if within the planner logic plugins internally utilize parts/algorithms from other libraries. I guess the solution suggested by @henningkayser to introduce error codes is a better solution to achieve both.

gavanderhoorn:
I would suggest to perhaps take a step back and evaluate the options wrt the short/long term goals of MoveIt 2 (and probably PickNik). If actual determinism or certification for particular use-cases does not have (high) priority right now, introducing these exceptions here might not be an actual problem or contribute to one.

Again, exceptions are fine and definitely have benefits (such as easier to follow control flow according to proponents as it doesn't require frequent `if(..) return ..;`s everywhere).

v4hn:
> Again, exceptions are fine and definitely have benefits (such as easier to follow control flow according to proponents as it doesn't require frequent if(..) return ..;s everywhere).

I would strictly prefer to see all expected code paths of a usual run to be handled without exceptions - that's why they are called "exceptions" after all. ;) That way users can always compile with [`-fno-exceptions`](https://stackoverflow.com/questions/32797607/is-there-a-way-to-enable-disable-the-exception-handling-in-the-c-source-file) to get rid of the overhead.
I'm not entirely sure but it looks like the introduced exception will only be triggered by a broken setup, so it should be fine.
No definition of "realtime support" would enforce "realtime handling of misconfigurations".

That being said I'm not one to talk. Robert sprinkled `boost::any_cast` throughout the Task Constructor code base and its interface is limited to exceptions...

moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
AdamPettinger: Maybe it should not be empty (e.g. it contains some `SUCCESS` feedback or something), but we might want to check anyway so we don't publish every cycle. This is running pretty fast, and publishing feedback here triggers a reaction by the `HybridPlanningManager`. The reaction could be trivial for trivial feedback, but it is still a lot of activity (i.e. ROS messages and plugin class function calls) for a message we care about like 1% of the time

moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
```diff
@@ -119,11 +121,11 @@ bool LocalPlannerComponent::initialize()
     if (!trajectory_operator_instance_->initialize(node_ptr, planning_scene_monitor_->getRobotModel(),
                                                    "panda_arm"))  // TODO(sjahr) add default group param
```
AdamPettinger, line 122: This TODO (and the other hardcoded `"panda_arm"` in this file) might be important to do before this is merged. It would be useful if you want to use hybrid planning on something besides the example panda moveit config

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/include/moveit/hybrid_planning_manager/planner_logic_interface.h
```diff
@@ -67,14 +67,14 @@ class PlannerLogicInterface
    * @param event Basic hybrid planning event
    * @return true if reaction was successful
```
MarqRazz, line 68: Need to update this comment and the one on line 75

moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
sjahr: I agree with Adam that it would create a message overhead when we would send "Null-feedback". Do you think it would be better to initialize the feedback and check for some kind of null-value within the local planner component before sending
 @henningkayser?

henningkayser (approved):

PR #501, opened 2021-06-16
Servo: Allow a negative joint margin

AndyZe:
Servo has a `joint_limit_margin` parameter that provides a little padding around joint limits from the URDF. This is useful to make sure the robot has time to stop before the actual joint limits.

We have an unusual use case where one joint has very narrow joint limits and the encoders are pretty noisy. Also, there are hard stops, so it's no big deal if the software joint limit gets exceeded. For that, we want to allow a negative joint limit margin. Although we still throw a warning because this is so unusual.


AdamPettinger (approved):
The idea seems OK to me, and I do think spitting out a warning is good

destogl (approved):
I can confirm the use-case :)

vatanaksoytezer (approved):
This also seems reasonable to me

nbbrooks (approved):
Some more context on why a servo specific solution is needed. Some smoothing algorithms appropriate for use with servo do not consider position limits and will "smooth" commands beyond them. 

On the other hand, time parameterization algorithms would handle pos as well as vel+acc+... limits directly.

Modifying joint_limits.yaml would impact both of these when we only want to adjust this for servoing. So this servo specific solution is needed.

moveit_ros/moveit_servo/src/servo_parameters.cpp
```diff
@@ -200,9 +200,9 @@ ServoParameters::SharedConstPtr ServoParameters::makeServoParameters(const rclcp
   }
   if (parameters->joint_limit_margin < 0.)
   {
-    RCLCPP_WARN(logger, "Parameter 'joint_limit_margin' should be "
-                        "greater than or equal to zero. Check yaml file.");
-    return nullptr;
+    RCLCPP_WARN(logger, "Parameter 'joint_limit_margin' is usually "
+                        "greater than or equal to zero, although negative values are allowed for special use cases. "
+                        "Check yaml file.");
```
v4hn, line 205:
```suggestion
    RCLCPP_WARN(logger, "Parameter 'joint_limit_margin' should usually be  greater than or equal to zero, "
                        "although negative values can be used if the specified joint limits are actually soft. "
                        "Check yaml file.");
```

v4hn (commented):
This parameter in general is just a way to simulate soft limits, which should be avoided, but can be moved out of. I would love to have support for them in MoveIt generally, but that's a bigger effort...

henningkayser (approved):

PR #502, opened 2021-06-16
Add prerelease tests

vatanaksoytezer:


.github/workflows/prerelease.yaml
```diff
@@ -0,0 +1,27 @@
+# This config uses industrial_ci (https://github.com/ros-industrial/industrial_ci.git).
+# For troubleshooting, see readme (https://github.com/ros-industrial/industrial_ci/blob/master/README.rst)
+
+name: pre-release
+
+on:
+  workflow_dispatch:
+  pull_request:
```
tylerjw, line 8:
```suggestion
```

tylerjw (approved):

MarqRazz (approved):

.github/workflows/prerelease.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+    env:
+      ROS_DISTRO: ${{ matrix.distro }}
+      PRERELEASE: true
+      BASEDIR: ${{ github.workspace }}/.work
+
+    name: "${{ matrix.distro }}"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
tylerjw, line 24: This will probably need the clean up disk space step like this: https://github.com/ros-planning/moveit/blob/95ab9eef58b552fe366f0513487a6f6db5a3e71e/.github/workflows/prerelease.yaml#L27

Have you tested this in your branch?

vatanaksoytezer: I did. It seemed fine without cleanup so I left it as is. If it is speeding up things, I can definitely add this. See https://github.com/vatanaksoytezer/moveit2/runs/2909699212?check_suite_focus=true for a sample run. (Galactic fail is related with buildfarm failures in `geometric_shapes`)

tylerjw: sounds good to me then, merging...

tylerjw: I need you to rebase this on main for some reason :(

henningkayser (approved):

PR #518, opened 2021-06-28
Fix commit that got removed during the sync

JafarAbdi:
### Description

I accidentally reverted the commits that got merged after I started the sync 11 days ago

PR #519, opened 2021-06-29
Fix main sync

henningkayser:
I think #518 is a scenario were we should consider actually fixing history. This branch squashes the fixup into the merge commit so that the affected lines show up as changed by the original PRs. The PRs that have been merged after the sync were rebased without any conflicts.

I'm not sure how we can prevent things like this in the future since it's not really obvious to see in the sync diff. @JafarAbdi any thoughts?


JafarAbdi:
> I'm not sure how we can prevent things like this in the future since it's not really obvious to see in the sync diff. @JafarAbdi any thoughts?

I think one possible solution would be to freeze the main branch when we have sync and wait until it's merged

henningkayser:
merged per agreement in standup meeting

PR #520, opened 2021-06-29
WIP: Fix build warnings

henningkayser:


henningkayser:
putting this on hold for now since this would require branching off Foxy. I'll revive this once we have a plan to do this.

Issue #522, opened 2021-06-30
Basic tutorial doesn't work


nkalupahana:
### Description

With the latest `moveit2` changes (42efad9), our team's MoveIt code suddenly stopped working. To see if it was an issue with our configuration, we tried the tutorials as well -- but they also seem to be broken. See "Steps to reproduce" for more.

### Your environment
* ROS Distro: Foxy
* OS Version: Ubuntu 20
* Source or Binary build? Source (42efad9)

### Steps to reproduce
- I created an empty Ubuntu 20 VM to start.
- [Install `ros-foxy-desktop`](https://docs.ros.org/en/foxy/Installation/Ubuntu-Install-Debians.html)
- [Follow the getting started instructions to install MoveIt2.](http://moveit2_tutorials.picknik.ai/doc/getting_started/getting_started.html)
- Run the MoveIt2 quickstart: `ros2 launch moveit2_tutorials demo.launch.py`

### Expected behaviour
You should get a screen that looks something like this (this is from my main Ubuntu VM, which has MoveIt2 at `a7d7798181cafd6a4e4614aeb0ab7b03e20aa370`:
![Working](https://cdn.discordapp.com/attachments/783194785975369738/859602517321842688/Screen_Shot_2021-06-29_at_6.13.26_PM.png)

### Actual behaviour
Instead, you get this:
![Not Working 1](https://cdn.discordapp.com/attachments/783194785975369738/859602341589680128/Screen_Shot_2021-06-29_at_5.59.56_PM.png)
In the "Context" tab of the MotionPlanning RViz plugin:
![Not Working 2](https://cdn.discordapp.com/attachments/783194785975369738/859602351462809621/Screen_Shot_2021-06-29_at_6.00.02_PM.png)

### Backtrace or Console output
Tutorial launch logs: https://gist.github.com/nkalupahana/09af17aa7f242bc73be5bd8495a7add5
My team's code launch logs: https://gist.github.com/nkalupahana/274d59bdaa42636109c00992cb124903
- Both the tutorial and my team's code work perfectly on MoveIt2 at `a7d7798`.

This is a serious issue for anyone getting started for MoveIt, since the basic tutorials don't work -- and considering my team's code has now stopped working, I think it's a pretty big issue for all users of MoveIt2.

I've now tested this on two clean Ubuntu VMs, and the issue persists -- this is a significant regression, especially for the Galactic release.

nkalupahana:
[...]

PR #524, opened 2021-06-30
move_group: Fix default_planning_pipeline for deprecated behavior

JafarAbdi:
### Description

Fix: https://github.com/ros-planning/moveit2/issues/522

We only enter this else statement when `default_planning_pipeline` is empty which will cause to not load any planning pipeline

### Checklist

### DIFF
--- a/moveit_ros/move_group/src/move_group.cpp
+++ b/moveit_ros/move_group/src/move_group.cpp
@@ -264,6 +264,7 @@ int main(int argc, char** argv)
     else
     {
       RCLCPP_WARN(LOGGER, "Falling back to using the the move_group node namespace (deprecated behavior).");
+      default_planning_pipeline = "move_group";
       moveit_cpp_options.planning_pipeline_options.pipeline_names = { default_planning_pipeline };
       moveit_cpp_options.planning_pipeline_options.parent_namespace = nh->get_effective_namespace();
     }


moveit_ros/move_group/src/move_group.cpp
```diff
@@ -264,6 +264,7 @@ int main(int argc, char** argv)
     else
     {
       RCLCPP_WARN(LOGGER, "Falling back to using the the move_group node namespace (deprecated behavior).");
+      default_planning_pipeline = "move_group";
       moveit_cpp_options.planning_pipeline_options.pipeline_names = { default_planning_pipeline };
       moveit_cpp_options.planning_pipeline_options.parent_namespace = nh->get_effective_namespace();
```
henningkayser, line 269: I think in MoveIt 1 we simply had an empty name while the parent namespace is set to `move_group`, which is the private namespace. I guess this somehow results in the same behavior as long as MoveGroup is not launched inside a subnamespace. Otherwise, we would have to set the parent namespace as node namespace + node name

henningkayser: `nh` should be renamed to `node`

JafarAbdi: > I think in MoveIt 1 we simply had an empty name while the parent namespace is set to move_group

Do you mean `moveit_cpp_options.planning_pipeline_options.parent_namespace` .? I don't think this variable is used at all (there's a TODO to make use of it)

henningkayser: oops, you're right. By default we were loading parameters from the private namespace, though. Which would result in the same behavior. It seems like this isn't the case here anymore?

JafarAbdi: Yes, that's right for ROS1, in moveit2 we made it such that the default namespace is `move_group` since there's no private namespace for the parameters, another solution would be to remove `move_group` from ompl configs like [this](https://github.com/ros-planning/moveit2_tutorials/pull/95) but that would break all existing launch files for the users, 

I think we should go with the solution in this PR to avoid breaking existing setups

henningkayser (approved):

moveit_ros/move_group/src/move_group.cpp
henningkayser: ... ok I think you're right that we shouldn't break existing setups. The old behavior will be deprecated and remove anyway eventually.

Issue #474, opened 2021-05-22
MotionPlanningDisplay not loading initially, requires reset


henningkayser:
### Description

On bad days the MotionPlanning display doesn't initialize correctly for me so that start and goal states as well as planner data are not available. After disabling and enabling the display again, everything works fine. 

### Your environment
* ROS Distro: Foxy
* OS Version: Ubuntu 20.04
* Version: Source build, main branch

### Steps to reproduce
Launch the MoveGroup demo: https://github.com/ros-planning/moveit2/tree/main/moveit_demo_nodes/run_move_group

### Expected behaviour
The display should just load and work from the start.

### Actual behaviour
It doesn't work (not always, but sometimes).

### Backtrace or Console output

Will deliver...

PR #525, opened 2021-06-30
Fix race condition when declaring parameters

JafarAbdi:
### Description

Fix: https://github.com/ros-planning/moveit2/issues/474

Not an optimal solution at all, but at least it works :D

### Checklist

### DIFF
--- a/moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
+++ b/moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
@@ -137,25 +137,34 @@ void RobotModelLoader::configure(const Options& opt)
           // Check if parameter has been declared to avoid exception
           param_name = prefix + "has_position_limits";
           bool has_position_limits = false;
-          if (!node_->has_parameter(param_name))
+          try
           {
             node_->declare_parameter(param_name);
+          }
+          catch (const rclcpp::exceptions::ParameterAlreadyDeclaredException&)
+          {
           }
           node_->get_parameter(param_name, has_position_limits);
 
           param_name = prefix + "has_velocity_limits";
           bool has_vel_limits = false;
-          if (!node_->has_parameter(param_name))
+          try
           {
             node_->declare_parameter(param_name);
+          }
+          catch (const rclcpp::exceptions::ParameterAlreadyDeclaredException&)
+          {
           }
           node_->get_parameter(param_name, has_vel_limits);
 
           param_name = prefix + "has_acceleration_limits";
           bool has_acc_limits = false;
-          if (!node_->has_parameter(param_name))
+          try
           {
             node_->declare_parameter(param_name);
+          }
+          catch (const rclcpp::exceptions::ParameterAlreadyDeclaredException&)
+          {
           }
           node_->get_parameter(param_name, has_acc_limits);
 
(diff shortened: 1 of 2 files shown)


tylerjw:
Why would a node have already declared these parameters?

JafarAbdi:
> Why would a node have already declared these parameters?

Not 100% sure, but I think this mainly happens for RViz since each display will load its own robot model and they all share the same node causing multiple declarations

moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
```diff
@@ -166,15 +175,21 @@ void RobotModelLoader::configure(const Options& opt)
           if (has_position_limits && canSpecifyPosition(joint_model, joint_id))
           {
             const auto max_pos_name = prefix + "max_position";
-            if (!node_->has_parameter(max_pos_name))
+            try
+            {
+              node_->declare_parameter(param_name);
```
henningkayser, line 180: Here you declare `param_name` while ignoring `max_pos_name` from above. Same for `min_pos_name`

henningkayser (commented):
So... RViz is initializing the plugins in separate threads with the same node instance? Well.. this is definitely another requirement for our WIP parameter API

v4hn:
We actually have [interfaces for sharing a RobotModel(Loader)](https://github.com/ros-planning/moveit2/blob/main/moveit_ros/planning_interface/common_planning_interface_objects/include/moveit/common_planning_interface_objects/common_objects.h#L49) in static memory.
Wouldn't it make sense to migrate the plugins to use `moveit::planning_interface::getSharedRobotModel()` instead of fixing the racing condition via catching exceptions?

JafarAbdi:
I ended up fixing it by adding a shared robot model loaders, hopefully, this's a better solution than the previous one

moveit_ros/planning_interface/common_planning_interface_objects/src/common_objects.cpp
```diff
@@ -109,6 +110,24 @@ std::shared_ptr<tf2_ros::Buffer> getSharedTF()
...
+  SharedStorage& s = getSharedStorage();
+  boost::mutex::scoped_lock slock(s.lock_);
+  auto it = s.robot_model_loaders_
+                .insert(std::make_pair(node->get_fully_qualified_name() + robot_description,
+                                       robot_model_loader::RobotModelLoaderWeakPtr()))
+                .first;
+  auto rml = it->second.lock();
+  if (!rml)
+  {
+    rml = std::make_shared<RobotModelLoader>(node, robot_description);
```
henningkayser, line 125: I assume you removed `opt.load_kinematics_solvers_ = true` on purpose since we don't call `setFromIK()` anywhere? But I think that RobotInteraction depends on IK solvers being loaded.

henningkayser (commented):
Solution looks good to me. Please verify if we should or should not load IK solvers for the shared robot model

moveit_ros/planning_interface/common_planning_interface_objects/src/common_objects.cpp
JafarAbdi: Sorry for not clarifying it, I did remove it because loading the kinematics solvers is the default for robot model loader https://github.com/ros-planning/moveit2/blob/main/moveit_ros/planning/robot_model_loader/include/moveit/robot_model_loader/robot_model_loader.h#L56, if you want me to add it back I can do it immediately

henningkayser: Oh, I wasn't aware of them being enabled by default. All good, thanks!

henningkayser (approved):

PR #533, opened 2021-07-01
Add new option to MoveGroupInterface to specify move group node's namespace

JafarAbdi:
### Description

Currently, it's not possible to specify the move group node's namespace in the move group interface, which makes it not possible to have different motion planning displays that connect to different move groups, could be tested with https://github.com/JafarAbdi/moveit2_tutorials/tree/pr-multi_arm

Fix: https://answers.ros.org/question/381276/ros2-rviz-node-two-robots-with-different-namespaces/

### Checklist

### DIFF
--- a/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
+++ b/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
@@ -110,8 +110,10 @@ public:
   /** \brief Specification of options to use when constructing the MoveGroupInterface class */
   struct Options
   {
-    Options(std::string group_name, std::string desc = ROBOT_DESCRIPTION)
-      : group_name_(std::move(group_name)), robot_description_(std::move(desc))
+    Options(std::string group_name, std::string move_group_namespace = "", std::string desc = ROBOT_DESCRIPTION)
+      : group_name_(std::move(group_name))
+      , robot_description_(std::move(desc))
+      , move_group_namespace_(std::move(move_group_namespace))
     {
     }
 
@@ -123,6 +125,9 @@ public:
 
     /// Optionally, an instance of the RobotModel to use can be also specified
     moveit::core::RobotModelConstPtr robot_model_;
+
+    /// The namespace for the move group node
+    std::string move_group_namespace_;
   };
 
   MOVEIT_STRUCT_FORWARD(Plan);
--- a/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
+++ b/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
@@ -57,6 +57,7 @@
 #include <moveit_msgs/srv/grasp_planning.hpp>
 #include <moveit_msgs/srv/get_planner_params.hpp>
 #include <moveit_msgs/srv/set_planner_params.hpp>
+#include <moveit/utils/rclcpp_utils.h>
 // TODO(JafarAbdi): Enable once moveit_ros_manipulation is ported
 // #include <moveit_msgs/msg/place_location.hpp>
 // #include <moveit_msgs/action/pickup.hpp>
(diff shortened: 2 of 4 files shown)


moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
```diff
@@ -110,8 +110,10 @@ class MoveGroupInterface
   /** \brief Specification of options to use when constructing the MoveGroupInterface class */
   struct Options
   {
-    Options(std::string group_name, std::string desc = ROBOT_DESCRIPTION)
-      : group_name_(std::move(group_name)), robot_description_(std::move(desc))
+    Options(std::string group_name, std::string move_group_namespace = "", std::string desc = ROBOT_DESCRIPTION)
```
henningkayser, line 113: `move_group_namespace` should be last so that it's not confused with `desc`

davetcoleman:
@JafarAbdi your link to the tutorials implies this is for multi-arm support. But I think we should be clear whether this is for:

- Two arms on one robot
- Two robots with one arm each

For the former, users should be using one unified URDF and one MoveGroup, IMHO. 

Which use case is this for?

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
```diff
@@ -137,13 +138,18 @@ class MoveGroupInterface::MoveGroupInterfaceImpl
     pose_reference_frame_ = getRobotModel()->getModelFrame();
 
     trajectory_event_publisher_ = pnode_->create_publisher<std_msgs::msg::String>(
-        trajectory_execution_manager::TrajectoryExecutionManager::EXECUTION_EVENT_TOPIC, 1);
+        rclcpp::names::append(opt_.move_group_namespace_,
```
davetcoleman, line 141: Could we document what this does?
```suggestion
        // Append the slash between two topic components
        rclcpp::names::append(opt_.move_group_namespace_,
```

JafarAbdi:
> @JafarAbdi your link to the tutorials implies this is for multi-arm support. But I think we should be clear whether this is for:
> 
> * Two arms on one robot
> * Two robots with one arm each
> 
> For the former, users should be using one unified URDF and one MoveGroup, IMHO.
> 
> Which use case is this for?

@davetcoleman This's for the second case, I agree that having one MoveGroup node is what we should do in the first case

davetcoleman (commented):
Joe and I reviewed this together and like this change. However we think a unit test should be written to demonstrate that everything loads up correctly with and without the namespace. Could you set this up?

JafarAbdi:
Reported from discord

> Hey, I'm not able to initialize MoveGroupInterface using https://github.com/ros-planning/moveit2/pull/533
Calling MoveGroupInterface constructor cause node freeze due to wait_for_servers params equals infinity. With mentioned PR above, move_group action_servers exists within changed path:
/$(namespace)/execute_trajectory
/$(namespace)/move_action
and MoveGroupInterface stuck here https://github.com/ros-planning/moveit2/blob/f601eef5e66e7b6c62e77b55d9f5e9e6706f7335/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp#L153  It is possible to set wait_for_servers for certain time, but then it make impossible to execute robot commands. How to initialize MoveGroupInterface within namespace?
>
> You may reproduce it by using mentioned above PR with corresponding launch file https://github.com/JafarAbdi/moveit2_tutorials/blob/pr-multi_arm/doc/quickstart_in_rviz/launch/demo.launch.py 
and the calling move_group interface, e.g. with this launch https://github.com/ros-planning/moveit2_tutorials/blob/main/doc/move_group_interface/launch/move_group_interface_tutorial.launch.py
Adding namespace here or/and modifying yaml params doesn't work.

> Sry, now it works, i forgot about slash in constructor
auto mgi_options = moveit::planning_interface::MoveGroupInterface::Options("arm_1_ur_manipulator", "/arm_1/", "arm_1.robot_description");

TODO: Handle slashes for the namespace parameter

henningkayser (approved):

PR #534, opened 2021-07-01
Remove rolling repos

vatanaksoytezer:
With the latest sync we no longer need a seperate _rolling.repos for builds.


davetcoleman (commented):
Makes sense to me, but @tylerjw I think is the best approver of this.

henningkayser (approved):

tylerjw (approved):

PR #539, opened 2021-07-05
Add missing test dependencies

jspricke:


vatanaksoytezer (approved):
Looks good to me

tylerjw (approved):

moveit_planners/ompl/package.xml
```diff
@@ -34,6 +34,7 @@
   <test_depend>tf2_eigen</test_depend>
   <buildtool_depend>eigen3_cmake_module</buildtool_depend>
   <test_depend>ament_cmake_gtest</test_depend>
+  <test_depend>ros_testing</test_depend>
```
JafarAbdi, line 37: Looking at https://github.com/ros-planning/moveit2/blob/main/moveit_planners/ompl/ompl_interface/CMakeLists.txt we are not using any of `ros_testing` functionality, I think we should remove https://github.com/ros-planning/moveit2/blob/main/moveit_planners/ompl/ompl_interface/CMakeLists.txt#L68 rather than adding it here

```suggestion
```

jspricke: done.

JafarAbdi (approved):

moveit_planners/ompl/package.xml
henningkayser: still there?

jspricke: uhm, no?

moveit_ros/planning_interface/package.xml
```diff
@@ -46,6 +46,8 @@
   <test_depend>moveit_planners_ompl</test_depend>
   <test_depend>rviz2</test_depend>
   <test_depend>ros_testing</test_depend>
```
henningkayser, line 48:
```suggestion
  <test_depend>ros_testing</test_depend>
```

moveit_planners/ompl/package.xml
henningkayser: ... I confused the files

henningkayser (approved):

PR #540, opened 2021-07-06
Refactor out velocity limit enforcement with test

tylerjw:
### Description

While working on my parameters refactor I pulled this function (and it's test) out of the code that depends on ros so it could be tested by unit tests.

### Checklist


AndyZe:
I can see that this gets rid of the `friend class`, so that makes testing a little simpler. I don't think it's a huge improvement but I guess I'm OK with it.

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
```diff
@@ -154,9 +154,6 @@ class ServoCalcs
   void suddenHalt(trajectory_msgs::msg::JointTrajectory& joint_trajectory) const;
   void suddenHalt(sensor_msgs::msg::JointState& joint_state) const;
 
-  /** \brief  Scale the delta theta to match joint velocity/acceleration limits */
-  void enforceVelLimits(Eigen::ArrayXd& delta_theta);
-
   /** \brief Avoid overshooting joint limits */
   bool enforcePositionLimits(sensor_msgs::msg::JointState& joint_state) const;
```
AndyZe, line 158: Seems like this should be moved to the new `enforce_limits.cpp too`

tylerjw: I agree.  This was more about converting the test to a unit test instead of a ros test (as those take forever and are flaky on foxy).

Long term I think it would be nice to separate all the math/logic from the ros code so it could be easily tested.

AdamPettinger (commented):
I like this, and agree we should move as much as possible of the logic to more testable places. Probably outside the scope of this PR, but good candidates are probably:

- `calculateWorstCaseStopTime`
- `checkValidCommand`
- `scaleCartesianCommand`
- `scaleJointCommand`
- Maybe `velocityScalingFactorForSingularity`
- `composeJointTrajMessage`
- `insertRedundantPointsIntoTrajectory`
- `removeDimension`
- `removeDriftDimensions`
- `enforceControlDimensions`

I think a lot of unit tests for those are already written, but use a friend class. So transferring the tests should not be too bad

tylerjw:
@AndyZe I rebased this on main.  Would you mind reviewing this again and let me know if you like this change.

AdamPettinger (approved):
+1 getting rid of the ros test in favor of unit test for internal joint limit checking

AndyZe (approved):

Issue #507, opened 2021-06-17
trajectory_event_publisher_ in MoveGroupInterface seems redundant


AndyZe:
In move_group_interface.cpp, there is a `move_action_client_` and a `trajectory_event_publisher_`. It's not clear to me why we have both. The `trajectory_event_publisher_` is used to interrupt trajectories but nothing else. Shouldn't the `move_action_client_` be able to cancel trajectories too?

(The discussion started [here](https://github.com/ros-planning/moveit2/pull/506#issuecomment-863531348).)

henningkayser:
The `trajectory_event_publisher_` allows stopping trajectories when using `asyncExecute()` or `asyncMove()`. I don't think it's redundant. However, it might be possible to implement the async calls by storing the active goal and future and using that one to terminate the action call. This would require quite some refactoring and testing, though.

PR #545, opened 2021-07-08
Port moveit ros control interface to ROS2

JafarAbdi:
### Description

- Port moveit_ros_control_interface to ROS2
- trajectory_execution_manager: Multiple bug fixes

Currently setting `moveit_manage_controllers` to `True` will not work since controller_manager doesn't report joints names for the inactive controllers

### Checklist


moveit_plugins/moveit_ros_control_interface/CMakeLists.txt
```diff
@@ -1,101 +1,85 @@
...
     moveit_core
+    moveit_simple_controller_manager
+    pluginlib
     trajectory_msgs
-  DEPENDS Boost
 )
 
-include_directories(include)
-include_directories(SYSTEM ${catkin_INCLUDE_DIRS} ${Boost_INCLUDE_DIRS})
+foreach(dependency IN ITEMS ${THIS_PACKAGE_INCLUDE_DEPENDS})
```
henningkayser, line 15: This is a really nice way of using the include depends. Maybe a best practice we want to apply everywhere?

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -1765,11 +1767,11 @@ bool TrajectoryExecutionManager::ensureActiveControllers(const std::vector<std::
         for (const std::string& controller_to_activate : controllers_to_activate)
         {
           ControllerInformation& ci = known_controllers_[controller_to_activate];
-          ci.last_update_ = rclcpp::Time();
+          ci.last_update_ = rclcpp::Time(0, 0, RCL_ROS_TIME);
```
henningkayser, line 1770: Looks like this should be set to `rclcpp::Time::now()`, no?

moveit_plugins/moveit_ros_control_interface/moveit_ros_control_interface_plugins.xml
```diff
@@ -1,18 +1,6 @@
...
-    </class>
-    <class name="effort_controllers/JointTrajectoryController" type="moveit_ros_control_interface::JointTrajectoryControllerAllocator" base_class_type="moveit_ros_control_interface::ControllerHandleAllocator">
-        <description></description>
-    </class>
-    <class name="pos_vel_controllers/JointTrajectoryController" type="moveit_ros_control_interface::JointTrajectoryControllerAllocator" base_class_type="moveit_ros_control_interface::ControllerHandleAllocator">
-        <description></description>
-    </class>
-    <class name="pos_vel_acc_controllers/JointTrajectoryController" type="moveit_ros_control_interface::JointTrajectoryControllerAllocator" base_class_type="moveit_ros_control_interface::ControllerHandleAllocator">
+  <library path="moveit_ros_control_interface_trajectory_plugin">
+    <class name="joint_trajectory_controller/JointTrajectoryController" type="moveit_ros_control_interface::JointTrajectoryControllerAllocator" base_class_type="moveit_ros_control_interface::ControllerHandleAllocator">
```
henningkayser, line 3: for better understanding.. All `JointTrajectoryController` types will use the same plugin in the future?

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -1071,6 +1071,8 @@ bool TrajectoryExecutionManager::configure(TrajectoryExecutionContext& context,
     // empty trajectories don't need to configure anything
     return true;
   }
+
+  reloadControllerInformation();
```
henningkayser, line 1075: Why is this needed here?

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -62,18 +67,32 @@ namespace moveit_ros_control_interface
...
 }
 
-MOVEIT_CLASS_FORWARD(MoveItControllerManager)  // Defines MoveItControllerManagerPtr, ConstPtr, WeakPtr... etc
+/**
+ * \brief Get joint name from resource name reported by ros2_control, since claimed_interfaces return by ros2_control
+ * will have the interface name as suffix joint_name/INTERFACE_TYPE
+ * @param[in] claimed_interface claimed interface as joint_name/INTERFACE_TYPE
+ * @return joint_name part of the /p claimed_interface
+ */
+std::string getJointName(const std::string& claimed_interface)
```
henningkayser, line 87:
```suggestion
std::string parseJointNameFromResource(const std::string& claimed_interface)
```
or something similar

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -62,18 +67,32 @@ namespace moveit_ros_control_interface
...
+{
+  const auto index = claimed_interface.find('/');
+  if (index == std::string::npos)
+    return claimed_interface;
```
henningkayser, line 91: should this be handled in any way?

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -204,14 +232,27 @@ class MoveItControllerManager : public moveit_controller_manager::MoveItControll
...
+  {
+    node_ = node;
+    if (!ns_.empty())
+    {
+      if (!node_->has_parameter("ros_control_namespace"))
+        ns_ = node_->declare_parameter<std::string>("ros_control_namespace", "/");
+    }
+    list_controllers_service_ = node_->create_client<controller_manager_msgs::srv::ListControllers>(
+        getAbsName("controller_manager/list_controllers"));
+    switch_controller_service_ = node_->create_client<controller_manager_msgs::srv::SwitchController>(
```
henningkayser, line 245: Do we need to wait for the services here to synchronize with ROS2-control?

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -338,31 +373,29 @@ class MoveItControllerManager : public moveit_controller_manager::MoveItControll
...
-      if (!ros::service::call(getAbsName("controller_manager/switch_controller"), srv))
+      auto result_future = switch_controller_service_->async_send_request(request);
+      if (result_future.wait_for(std::chrono::duration<double>(SERVICE_CALL_TIMEOUT)) == std::future_status::timeout)
       {
-        ROS_ERROR_STREAM("Could switch controllers at " << ns_);
+        RCLCPP_ERROR_STREAM(LOGGER, "Could switch controllers at " << switch_controller_service_->get_service_name());
       }
       discover(true);
-      return srv.response.ok;
+      return result_future.get()->ok;
```
henningkayser, line 398: In case of a timeout `get()` shouldn't be called anymore since it's using an internal wait call.

henningkayser:
@JafarAbdi doesn't this PR address https://github.com/ros-planning/moveit2/issues/507

moveit_plugins/moveit_ros_control_interface/moveit_ros_control_interface_plugins.xml
JafarAbdi: I'm not 100% sure, but I think currently this's the only supported controller, @destogl what do you think .?

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
JafarAbdi: Just in case, do you want me to throw an exception if this happens .?

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
JafarAbdi: This's related to https://github.com/ros-controls/ros2_control/issues/464 causing `known_controllers_` to have an empty vector of joints for the newly activated controller, calling this will update the joint list

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
JafarAbdi: I think both should work, this how it was previously

AndyZe (commented):
Is there a high-level overview of the difference between these packages somewhere? Frankly, they all sound the same:

- moveit_ros_control_interface:  "ros_control controller manager interface for MoveIt"
- moveit_controller_manager: I see that this is COLCON-ignored, probably will be deleted.
- moveit_simple_controller_manager: "A generic, simple controller manager plugin for MoveIt."

It seems like `moveit_ros_control_interface` and `moveit_simple_controller_manager` should be combined into one package.

AndyZe (commented):
I also think a bit of specific documentation about what these packages do would be great. Do they switch controllers when a different move group is used, or ...?

moveit_plugins/moveit_ros_control_interface/moveit_ros_control_interface_plugins.xml
destogl: Currently, there is only one `JointTrajectoryController` which interfaces are configurable. We are planning to add specializations into different namespaces, but the current one will always be used at the end. The specializations basically only limit parameterization possibilities.

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -62,18 +67,32 @@ namespace moveit_ros_control_interface
  * @param[in] force force timeout
  * @return True if timeout duration was passed
  */
-bool checkTimeout(ros::Time& t, double timeout, bool force = false)
+bool checkTimeout(const rclcpp::Clock::SharedPtr& clock, rclcpp::Time& t, double timeout, bool force = false)
```
henningkayser, line 70: This seems overly complicated and the semantics of `force` is somewhat weird here...
I would replace this function with a single-line check like:
```
static const rclcpp::Duration CONTROLLERS_TIMEOUT = rclcpp::Duration::from_seconds(1.0);

...
void discover(bool forceUpdate = false)
{
  if (forceUpdate || (node_->now() - controllers_stamp_ >= CONTROLLERS_TIMEOUT))
  {
    controllers_stamp_ = node_now();
    ...
  }
}
```

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
henningkayser: I think that would make a lot of sense. Especially if we don't have a better way to validate the resource at this point.

moveit_plugins/moveit_ros_control_interface/src/gripper_controller_plugin.cpp
```diff
@@ -0,0 +1,62 @@
...
+ * \brief Simple allocator for moveit_simple_controller_manager::FollowJointTrajectoryControllerHandle instances.
+ */
+class GripperControllerAllocator : public ControllerHandleAllocator
+{
+public:
+  moveit_controller_manager::MoveItControllerHandlePtr alloc(const rclcpp::Node::SharedPtr& node,
+                                                             const std::string& name,
+                                                             const std::vector<std::string>& /* resources */) override
+  {
+    return std::make_shared<moveit_simple_controller_manager::GripperControllerHandle>(node, name, "gripper_cmd");
```
henningkayser, line 55: should "gripper_cmd" be dynamic, or is it always going to be this value?

henningkayser (commented):
testing this right now...

henningkayser (approved):

PR #551, opened 2021-07-11
Add crypto library to avoid MD5 build error

fmrico:
Signed-off-by: Francisco Martín Rico <fmrico@gmail.com>

### Description

`moveit_ros_warehouse` package fails in Ubuntu/Foxy because MD5 is not defined:

```
--- stderr: moveit_ros_warehouse                                                                                                      
/usr/bin/ld: libmoveit_warehouse.so.2.2.0: reference to `MD5' undefined
```

This PR adds a `target_link_libraries` instruction to compile with crypto lib.

I hope it helps!!
Francisco

### Checklist

### DIFF
--- a/moveit_ros/warehouse/warehouse/CMakeLists.txt
+++ b/moveit_ros/warehouse/warehouse/CMakeLists.txt
@@ -11,6 +11,7 @@ add_library(${MOVEIT_LIB_NAME} SHARED
 )
 set_target_properties(${MOVEIT_LIB_NAME} PROPERTIES VERSION "${${PROJECT_NAME}_VERSION}")
 ament_target_dependencies(${MOVEIT_LIB_NAME} rclcpp Boost moveit_core warehouse_ros moveit_ros_planning)
+target_link_libraries(${MOVEIT_LIB_NAME} crypto)
 
 add_executable(moveit_warehouse_broadcast src/broadcast.cpp)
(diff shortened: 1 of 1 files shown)


tylerjw:
Do we also need to add this to the package.xml and the list of exported dependencies?

fmrico:
> Do we also need to add this to the package.xml and the list of exported dependencies?

Hi @tylerjw. I think it's not necessary.

tylerjw (approved):

JafarAbdi (changes requested):
This's a bug in `ament_cmake` where it picks the system installed package rather than the source one, could you please test my last comment from this [answer](https://answers.ros.org/question/381730/colcon-build-moveit2-failure-because-of-moveit_ros_warehouse-collect2-error/?answer=381792#post-id-381792) .?

The current release for `warehouse_ros` doesn't export this dependency correctly, it should be fixed after the next foxy sync

AndyZe:
I had the same build error when installing MoveIt2 from source, at first. After following the directions more carefully, the build error went away. So, I tend to agree with Jafar- this probably is not needed.

fmrico:
Yes, definitely

Thanks for the pointer @JafarAbdi ;)

AndyZe:
@fmrico thanks for submitting, regardless!

PR #553, opened 2021-07-12
Fix joint's position limits loading

JafarAbdi:
### Description

This PR fixes two things

1- The previous PR that added the declaration for this parameterw introduced a new parameter called `has_position_limits` and it has to be true to override the max/min_positions which breaks all the existing setups, so I did revert it
2- Similar to https://github.com/ros-planning/moveit2/commit/45dfbae5baab6c1d1dddfff04af68f83cbff7464 it fixes unwanted overrides for the position limits

### Checklist

### DIFF
--- a/moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
+++ b/moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
@@ -131,18 +131,37 @@ void RobotModelLoader::configure(const Options& opt)
         std::string param_name;
         try
         {
-          // Need to check before setting to true
-          joint_limit[joint_id].has_position_limits = false;
+          const auto max_pos_name = prefix + "max_position";
+          if (!node_->has_parameter(max_pos_name))
+          {
+            node_->declare_parameter(max_pos_name);
+          }
+          double max_position;
+          if (node_->get_parameter(max_pos_name, max_position))
+          {
+            if (canSpecifyPosition(joint_model, joint_id))
+            {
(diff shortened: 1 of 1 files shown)


nbbrooks (approved):
lgtm but have a couple thoughts
- Should we warn the user if they have specified a min_position but not a max_position (or vice versa)?
- Should we check that, if min and max are defined, that min < max? Or is that done elsewhere?

JafarAbdi:
> * Should we warn the user if they have specified a min_position but not a max_position (or vice versa)?

I think in that case the value from URDF will be used for the other unset position

> * Should we check that, if min and max are defined, that min < max? Or is that done elsewhere?

I'm not sure, maybe other maintainers have thoughts on this

nbbrooks:
> > * Should we warn the user if they have specified a min_position but not a max_position (or vice versa)?
> 
> I think in that case the value from URDF will be used for the other unset position
> 
I forgot about that, current behavior sounds good to me then.

> > * Should we check that, if min and max are defined, that min < max? Or is that done elsewhere?
> 
> I'm not sure, maybe other maintainers have thoughts on this

Jafar looked into this and found there's no error during the configuration but when the user plan they will get the following error 
```
moveit_constraint_samplers.default_constraint_samplers:` The constraints for joint 'panda_joint1' are such that there are no possible values for the joint: min_bound: 0.5, max_bound: 0.1. Failing.
```
I'm happy with that behavior. If we want parse-time warnings/errors we should tackle that in a separate issue.

+1

PR #555, opened 2021-07-13
Fix MoveIt Servo compilation on macOS

### LINKED ISSUE #491: macOS moveit_servo build issue
### Description

When building MoveIt2 from source, I'm getting a strange error when building MoveIt Servo:

```
--- stderr: moveit_servo
Undefined symbols for architecture x86_64:
  "void moveit_servo::declareOrGetParam<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::shared_ptr<rclcpp::Node> const&, rclcpp::Logger const&)", referenced from:
      moveit_servo::PoseTracking::readROSParams() in pose_tracking.cpp.o
  "void moveit_servo::declareOrGetParam<double>(double&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::shared_ptr<rclcpp::Node> const&, rclcpp::Logger const&)", referenced from:
      moveit_servo::PoseTracking::readROSParams() in pose_tracking.cpp.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [libpose_tracking.dylib] Error 1
make[1]: *** [CMakeFiles/pose_tracking.dir/all] Error 2
make: *** [all] Error 2
```
This build was run on CI, so full configuration can be viewed here: https://github.com/nkalupahana/ros2-foxy-macos/runs/2786139463?check_suite_focus=true

Any ideas about what this might be about?

### Your environment
* ROS Distro: Foxy
* OS Version: macOS
* Build details: From source (main)

nkalupahana:
### Description

This PR fixes MoveIt Servo compilation on macOS. Closes #491
Broken CI: https://github.com/nkalupahana/ros2-foxy-macos/runs/3026534213?check_suite_focus=true
Now working: https://github.com/nkalupahana/ros2-foxy-macos/runs/3053981174?check_suite_focus=true

### Checklist


moveit_ros/moveit_servo/CMakeLists.txt
```diff
@@ -85,7 +85,7 @@ set_target_properties(${SERVO_LIB_NAME} PROPERTIES VERSION "${${PROJECT_NAME}_VE
 ament_target_dependencies(${SERVO_LIB_NAME} ${THIS_PACKAGE_INCLUDE_DEPENDS})
 target_link_libraries(${SERVO_LIB_NAME} ${SERVO_PARAM_LIB_NAME})
 
-add_library(${POSE_TRACKING} SHARED src/pose_tracking.cpp)
+add_library(${POSE_TRACKING} SHARED src/pose_tracking.cpp src/servo_parameters.cpp)
 ament_target_dependencies(${POSE_TRACKING} ${THIS_PACKAGE_INCLUDE_DEPENDS})
 target_link_libraries(${POSE_TRACKING} ${SERVO_LIB_NAME})
```
JafarAbdi, line 90: I think we should link it against the already built library rather than adding the source files to the target

```suggestion
add_library(${POSE_TRACKING} SHARED src/pose_tracking.cpp)
ament_target_dependencies(${POSE_TRACKING} ${THIS_PACKAGE_INCLUDE_DEPENDS})
target_link_libraries(${POSE_TRACKING} ${SERVO_LIB_NAME} ${SERVO_PARAM_LIB_NAME})
```

nkalupahana: You can't link against the already-built library, because as you can see from the failing CI, this is a template issue -- the already-built library doesn't have the correct templated classes built in, which is why linker fails. Thus, the source has to be built each time (as is done here) to allow for any additional template classes that need to be built.

Failing CI of this change: https://github.com/nkalupahana/ros2-foxy-macos/runs/3058543555?check_suite_focus=true

JafarAbdi: > I think including the implementation (cpp) in the .h file would also work; that's the pattern I've seen for template classes in the past, at least. Let me know what you want to see.

Yes, I think this should be the solution

moveit_ros/moveit_servo/include/moveit_servo/servo_parameters.h
```diff
@@ -155,4 +155,31 @@ struct ServoParameters
   rcl_interfaces::msg::SetParametersResult setParametersCallback(std::vector<rclcpp::Parameter> parameters);
 };
 
+// Helper template for declaring and getting ros param
+// Must be declared here to ensure template class is built for required templates when included
+template <typename T>
+void declareOrGetParam(T& output_value, const std::string& param_name, const rclcpp::Node::SharedPtr& node,
+                       const rclcpp::Logger& logger)
```
JafarAbdi, line 162: Why did you remove the default value parameter .?

moveit_ros/moveit_servo/CMakeLists.txt
nkalupahana: Alright, so it looks like you can't just include the .cpp file because not everything in there is templated so if you include it, then you'll get duplicate symbols. To fix that, I just moved the template function into the .h file (the forward declaration is already in there, so I figured that wouldn't be too big of a deal). Let me know if this works!

CI: https://github.com/nkalupahana/ros2-foxy-macos/actions/runs/1027725851

moveit_ros/moveit_servo/include/moveit_servo/servo_parameters.h
nkalupahana: Oh sorry, not sure. I'll fix that.

nkalupahana: Okay, I think this is good.

moveit_ros/moveit_servo/include/moveit_servo/servo_parameters.h
```diff
@@ -155,4 +150,31 @@ struct ServoParameters
...
+{
+  try
+  {
+    if (node->has_parameter(param_name))
+    {
+      node->get_parameter<T>(param_name, output_value);
+    }
+    else
+    {
+      output_value = node->declare_parameter<T>(param_name, T{});
```
JafarAbdi, line 167:
```suggestion
      output_value = node->declare_parameter<T>(param_name, default_value);
```

nkalupahana:
@JafarAbdi any idea what's going on with CI? I don't think that failure is due to this PR.

JafarAbdi:
> @JafarAbdi any idea what's going on with CI? I don't think that failure is due to this PR.

Look like the build farm is broken for rolling, I'll retrigger the workflow after few hours to see if it got fixed

nkalupahana:
@JafarAbdi looks like a slightly different CI issue, has this been reported elsewhere? I didn't see any issues for it on ros2_control.

AndyZe (approved):

nkalupahana:
@JafarAbdi rolling CI passed on 42ba2fa, but now it's broken again? Not sure what's going on. If it's okay with you, I think we should just merge this in -- rolling CI has now passed, and these breaks in the CI now have nothing to do with the PR :)

moveit_ros/moveit_servo/CMakeLists.txt
davetcoleman: I'd like to recommend the rationale for this be added as a comment to this CMake file... just copy-paste your explanation here, into the code :-)

davetcoleman:
@tylerjw agrees we can merge this because rolling is broken...

moveit_ros/moveit_servo/CMakeLists.txt
nkalupahana: Added! :)

tylerjw (approved):
This does not change the functionality and will be part of the code I'll be replacing with my parameters refactor.  I'll merge this without rolling-testing passing (it is a ros infrastructure problem).

Issue #529, opened 2021-06-30
Delete rolling and galactic repos files once there are syncs with the latest upstream releases


vatanaksoytezer:
We needed those files because some of these packages weren't available for rolling and galactic main repositories. This will be fixed with the syncs. Rolling will have it's sync tomorrow: https://discourse.ros.org/t/preparing-for-rolling-sync-2021-07-01/21169 and Galactic sync is in two weeks from now: https://discourse.ros.org/t/when-is-the-next-galactic-sync/21146

vatanaksoytezer:
#534 deletes these files for rolling, I will send a follow up PR for galactic.

PR #558, opened 2021-07-19
Remove galactic repos from CI builds

vatanaksoytezer:
First Galactic sync is [out](https://discourse.ros.org/t/new-packages-for-galactic-geochelone-2021-07-16/21418) and now we should be able to build without a seperate galactic repos file.


vatanaksoytezer:
Should close #529

tylerjw (approved):

PR #562, opened 2021-07-21
Tests for CurrentStateMonitor using dependency injection

tylerjw:
### Description

This is mostly about demonstrating a pattern for using dependency injection to be able to unit test sections of the code with ros dependencies in a unit test.

### Checklist


moveit_ros/planning/planning_scene_monitor/test/current_state_monitor_tests.cpp
```diff
@@ -0,0 +1,106 @@
...
+struct MockRclInterface : public planning_scene_monitor::CurrentStateMonitor::RclInterface
+{
+  MOCK_METHOD(rclcpp::Time, now, (), (const, override));
+  MOCK_METHOD(void, createJointStateSubscription,
+              (const std::string& topic, planning_scene_monitor::JointStateUpdateCallback callback), (override));
+  MOCK_METHOD(void, resetJointStateSubscription, (), (override));
+  MOCK_METHOD(std::string, getJointStateTopicName, (), (const, override));
+};
+
+TEST(StartStopTest, CurrentStateMonitorTests)
```
gbrooks2-pn, line 57: Consider BDD or TDD comments. My preference is BDD
```
// GIVEN preconditions
// WHEN event happens
// THEN expected result
```

moveit_ros/planning/planning_scene_monitor/test/current_state_monitor_tests.cpp
```diff
@@ -0,0 +1,106 @@
...
+{
+  auto mock_rcl_interface = std::make_unique<MockRclInterface>();
+  auto clock = std::make_shared<rclcpp::Clock>();
+  moveit::core::RobotModelPtr robot_model = moveit::core::loadTestingRobotModel("panda");
+  auto tf_buffer = std::make_shared<tf2_ros::Buffer>(clock);
```
gbrooks2-pn, line 62: Consider
```
auto tf_buffer = std::make_shared<tf2_ros::Buffer>{std::make_shared<rclcpp::Clock>{}};
```
unless you think it's too long, to limit variable scope

moveit_ros/planning/planning_scene_monitor/test/current_state_monitor_tests.cpp
```diff
@@ -0,0 +1,106 @@
...
+              (const std::string& topic, planning_scene_monitor::JointStateUpdateCallback callback), (override));
+  MOCK_METHOD(void, resetJointStateSubscription, (), (override));
+  MOCK_METHOD(std::string, getJointStateTopicName, (), (const, override));
+};
+
+TEST(StartStopTest, CurrentStateMonitorTests)
+{
+  auto mock_rcl_interface = std::make_unique<MockRclInterface>();
+  auto clock = std::make_shared<rclcpp::Clock>();
+  moveit::core::RobotModelPtr robot_model = moveit::core::loadTestingRobotModel("panda");
```
gbrooks2-pn, line 61: Does this `auto`?

moveit_ros/planning/planning_scene_monitor/test/current_state_monitor_tests.cpp
```diff
@@ -0,0 +1,106 @@
...
+  auto tf_buffer = std::make_shared<tf2_ros::Buffer>(clock);
+
+  EXPECT_CALL(*mock_rcl_interface, createJointStateSubscription);
+  EXPECT_CALL(*mock_rcl_interface, resetJointStateSubscription);
+  EXPECT_CALL(*mock_rcl_interface, now);
+
+  planning_scene_monitor::CurrentStateMonitor current_state_monitor(std::move(mock_rcl_interface), robot_model,
```
gbrooks2-pn, line 68: Along the lines of scope limit
```
auto mock_rcl_interface = std::make_unique<MockRclInterface>{};
EXPECT_CALL(*mock_rcl_interface, createJointStateSubscription);
EXPECT_CALL(*mock_rcl_interface, resetJointStateSubscription);
EXPECT_CALL(*mock_rcl_interface, now);

planning_scene_monitor::CurrentStateMonitor current_state_monitor{
    std::move{mock_rcl_interface},
    moveit::core::loadTestingRobotModel{"panda"},
    std::make_shared<tf2_ros::Buffer>{std::make_shared<rclcpp::Clock>{}}};
```
might be a way to go, unless you think it's too noisy

moveit_ros/planning/planning_scene_monitor/test/current_state_monitor_tests.cpp
```diff
@@ -0,0 +1,106 @@
...
+                                                                    tf_buffer);
+  current_state_monitor.startStateMonitor();
+
+  EXPECT_TRUE(current_state_monitor.isActive());
+
+  current_state_monitor.stopStateMonitor();
+
+  EXPECT_FALSE(current_state_monitor.isActive());
```
gbrooks2-pn, line 76: Personally, I would make this 3-5 tests, with the BDD comments, so the focus of what's being tested is very limited.

moveit_ros/planning/planning_scene_monitor/test/current_state_monitor_tests.cpp
```diff
@@ -0,0 +1,106 @@
...
+}
+
+TEST(NoModelTest, CurrentStateMonitorTests)
+{
+  auto mock_rcl_interface = std::make_unique<MockRclInterface>();
+  auto clock = std::make_shared<rclcpp::Clock>();
+  moveit::core::RobotModelPtr robot_model = nullptr;
+  auto tf_buffer = std::make_shared<tf2_ros::Buffer>(clock);
+
+  try
```
gbrooks2-pn, line 86: Use `EXPECT_THROW` http://google.github.io/googletest/reference/assertions.html#EXPECT_THROW

moveit_ros/planning/planning_scene_monitor/test/current_state_monitor_tests.cpp
```diff
@@ -0,0 +1,106 @@
...
+  {
+    planning_scene_monitor::CurrentStateMonitor current_state_monitor(std::move(mock_rcl_interface), robot_model,
```
gbrooks2-pn, line 88: uniform initialization

moveit_ros/planning/planning_scene_monitor/test/current_state_monitor_tests.cpp
```diff
@@ -0,0 +1,106 @@
...
+  current_state_monitor.stopStateMonitor();
+
+  EXPECT_FALSE(current_state_monitor.isActive());
+}
+
+TEST(NoModelTest, CurrentStateMonitorTests)
+{
+  auto mock_rcl_interface = std::make_unique<MockRclInterface>();
+  auto clock = std::make_shared<rclcpp::Clock>();
+  moveit::core::RobotModelPtr robot_model = nullptr;
```
gbrooks2-pn, line 83: This is an example where I think BDD is helpful, since the focus of the test is in the parameters
```
// GIVEN an uninitialized robot model
moveit::core::RobotModelPtr robot_model = nullptr;
// WHEN the CurrentStateMonitor is constructed with it
// THEN we expect the monitor to throw because of the invalid model
EXPECT_THROW(planning_scene_monitor::CurrentStateMonitor current_state_monitor(.., robot_model, ...);, std::invalid_argument);
```

gbrooks2-pn (changes requested):

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -60,10 +60,16 @@ RobotState::RobotState(const RobotModelConstPtr& robot_model)
   , has_velocity_(false)
   , has_acceleration_(false)
   , has_effort_(false)
-  , dirty_link_transforms_(robot_model_->getRootJoint())
```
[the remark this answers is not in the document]
tylerjw: I'm up for doing this in a separate PR.  I'm hoping to keep this PR as minimal as possible so I can point people to it as an example.

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -60,10 +60,16 @@ RobotState::RobotState(const RobotModelConstPtr& robot_model)
...
+  , dirty_link_transforms_(nullptr)
   , dirty_collision_body_transforms_(nullptr)
   , rng_(nullptr)
 {
+  if (!robot_model)
```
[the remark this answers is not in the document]
tylerjw: Changing to `== nullptr` because it is more explicit, and I generally side with more explicit and the automatic conversion to bool isn't as obvious.

moveit_ros/planning/package.xml
```diff
@@ -41,6 +41,9 @@
 
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
+  <test_depend>ament_cmake_gmock</test_depend>
+  <test_depend>ament_cmake_gtest</test_depend>
+  <test_depend>moveit_core</test_depend>
```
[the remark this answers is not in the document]
tylerjw: this was a mistake, there is not a linter for ros2 for these things

moveit_ros/planning/planning_scene_monitor/CMakeLists.txt
```diff
@@ -36,3 +37,19 @@ install(TARGETS demo_scene
 )
 
 install(DIRECTORY include/ DESTINATION include)
+
+if(BUILD_TESTING)
+  find_package(ament_cmake_gtest REQUIRED)
+  find_package(ament_cmake_gmock REQUIRED)
+  find_package(ros_testing REQUIRED)
```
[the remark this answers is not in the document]
tylerjw: I was mistaken that ros_testing was needed, removing

moveit_ros/planning/planning_scene_monitor/CMakeLists.txt
```diff
@@ -36,3 +37,19 @@ install(TARGETS demo_scene
...
+  find_package(moveit_core REQUIRED)
```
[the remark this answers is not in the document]
tylerjw: same answer, you found a mistake I made

moveit_ros/planning/planning_scene_monitor/test/current_state_monitor_tests.cpp
tylerjw: Can you help me understand how you would this these up into smaller tests?

moveit_ros/planning/planning_scene_monitor/test/current_state_monitor_tests.cpp
tylerjw: changed

moveit_ros/planning/planning_scene_monitor/test/current_state_monitor_tests.cpp
tylerjw: can't get it work along with the EXPECT_THROW,

results in a "EXPECT_THROW takes 2 arguments and you provided 4"

moveit_ros/planning/planning_scene_monitor/test/current_state_monitor_tests.cpp
tylerjw: I'll ask you for help understanding what you mean by this.

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/current_state_monitor.h
```diff
@@ -47,7 +47,7 @@
 
 namespace planning_scene_monitor
 {
-using JointStateUpdateCallback = boost::function<void(const sensor_msgs::msg::JointState::ConstSharedPtr&)>;
+using JointStateUpdateCallback = std::function<void(sensor_msgs::msg::JointState::ConstSharedPtr)>;
```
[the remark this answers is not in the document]
tylerjw: done

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/current_state_monitor.h
```diff
@@ -56,6 +56,53 @@ class CurrentStateMonitor
...
+     */
+    virtual rclcpp::Time now() const = 0;
+
+    /**
+     * @brief      Creates a joint state subscription
+     *
+     * @param[in]  topic     The topic
+     * @param[in]  callback  The callback
+     */
+    virtual void createJointStateSubscription(const std::string& topic, JointStateUpdateCallback callback) = 0;
```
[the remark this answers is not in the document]
tylerjw: done

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/current_state_monitor_rcl_interface.hpp
```diff
@@ -0,0 +1,89 @@
...
+   */
+  rclcpp::Time now() const override;
+
+  /**
+   * @brief      Creates a joint state subscription
+   *
+   * @param[in]  topic     The topic
+   * @param[in]  callback  The callback
+   */
+  void createJointStateSubscription(const std::string& topic, JointStateUpdateCallback callback) override;
```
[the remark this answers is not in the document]
tylerjw: done

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/current_state_monitor_rcl_interface.hpp
```diff
@@ -0,0 +1,89 @@
...
+  /**
+   * @brief      Get the joint state topic name
+   *
+   * @return     The joint state topic name.
+   */
+  std::string getJointStateTopicName() const override;
+
+private:
+  rclcpp::Node::SharedPtr node_;
+  rclcpp::Subscription<sensor_msgs::msg::JointState>::SharedPtr joint_state_subscription_;
```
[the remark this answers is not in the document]
tylerjw: done

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
```diff
@@ -35,20 +35,25 @@
...
+}
 
-planning_scene_monitor::CurrentStateMonitor::CurrentStateMonitor(const rclcpp::Node::SharedPtr& node,
-                                                                 const moveit::core::RobotModelConstPtr& robot_model,
-                                                                 const std::shared_ptr<tf2_ros::Buffer>& tf_buffer)
-  : node_(node)
+CurrentStateMonitor::CurrentStateMonitor(std::unique_ptr<CurrentStateMonitor::RclInterface> rcl_interface,
+                                         const moveit::core::RobotModelConstPtr& robot_model,
+                                         const std::shared_ptr<tf2_ros::Buffer>& tf_buffer)
+  : rcl_interface_(std::move(rcl_interface))
```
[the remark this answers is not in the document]
tylerjw: yes, I really want to keep the scope of this limit

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor_rcl_interface.cpp
```diff
@@ -0,0 +1,80 @@
...
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Tyler Weaver */
+
+#include <moveit/planning_scene_monitor/current_state_monitor.h>
```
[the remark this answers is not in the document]
tylerjw: you aren't, that was a mistake

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor_rcl_interface.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include <moveit/planning_scene_monitor/current_state_monitor_rcl_interface.hpp>
+
```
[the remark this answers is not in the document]
tylerjw: done

moveit_ros/planning/planning_scene_monitor/test/current_state_monitor_tests.cpp
```diff
@@ -0,0 +1,106 @@
...
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Tyler Weaver */
+
+#include <memory>
```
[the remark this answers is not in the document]
tylerjw: done

moveit_ros/planning/planning_scene_monitor/test/current_state_monitor_tests.cpp
tylerjw: replaced by single line construction.

moveit_ros/planning/planning_scene_monitor/test/current_state_monitor_tests.cpp
tylerjw: I like this.  I changed to doing this.  You can't use the {} initialization with std::move (it is a function) or std::make_shared (same reason I think).

moveit_ros/planning/planning_scene_monitor/test/current_state_monitor_tests.cpp
tylerjw: it does but now I'm just doing it in the parameter of the custructor

moveit_ros/planning/planning_scene_monitor/test/current_state_monitor_tests.cpp
gbrooks2-pn: Resolved

gbrooks2-pn (approved):

griswaldbrooks (approved):

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/current_state_monitor.h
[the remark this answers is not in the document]
davetcoleman: this looks like an API change that could cause some noise by people use are implementing this call back, but let them make noise! it's only moveit2 :-)

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/current_state_monitor.h
```diff
@@ -56,6 +57,53 @@ class CurrentStateMonitor
   using TFConnection = boost::signals2::connection;
 
 public:
+  /**
+   * @brief      This class contains the rcl interfaces for easier testing
+   */
+  class RclInterface
```
davetcoleman, line 63: Should there be a better name? 
```suggestion
  class RclTestInterface
```

davetcoleman: This is just for testing right?

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/current_state_monitor_rcl_interface.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+#include <moveit/planning_scene_monitor/current_state_monitor.h>
+#include <rclcpp/rclcpp.hpp>
+#include <sensor_msgs/msg/joint_state.hpp>
+#include <string>
+
+namespace planning_scene_monitor
+{
+/**
+ * @brief      This class contains the rcl interfaces for CurrentStateMontior
```
davetcoleman, line 47: This is not very descriptive... the words you used here are all in the file name. Be more helpful to future developers, like:
```suggestion
 * @brief      Dependency injection method for testing the CurrentStateMonitor
```

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/current_state_monitor.h
davetcoleman: This class seems like a generic tool that should not be specific to the CurrentStateMonitor. Maybe it should live in moveit_core/utils ?

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor_rcl_interface.cpp
```diff
@@ -0,0 +1,81 @@
...
+std::string CurrentStateMonitorRclInterface::getJointStateTopicName() const
+{
+  if (joint_state_subscription_)
+  {
+    return joint_state_subscription_->get_topic_name();
+  }
+  else
+  {
+    return "";
```
davetcoleman, line 77: should this also throw/print an error, if the string is undefined?

davetcoleman (commented):
RCL is already a wrapper around DDS or other middlewares. Now we are creating a wrapper for the wrapper, in the name of testing? Does this kind of functionality belong in core ROS2, rather than a one-off implementation in this one MoveIt class? This type of new complexity in the name of testing does not seem like a good idea to me, in terms of long term maintenance. What's a solution we can use for all of MoveIt? Or for all ROS applications?

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor_rcl_interface.cpp
tylerjw: This is not a change in behavior (this is a straight copy-paste of code from current_state_monitor)

tylerjw:
> RCL is already a wrapper around DDS or other middlewares. Now we are creating a wrapper for the wrapper, in the name of testing? Does this kind of functionality belong in core ROS2, rather than a one-off implementation in this one MoveIt class? 

My original idea was to use the abstract interfaces in `rclcpp::node_interfaces`.  The problem with this became apparent quickly when trying to create a subscription.  The underlying behavior in ros is done by a bunch of c functions.  These functions depend on a concrete node class existing.  There is no way to get dependency injection using types already in rclcpp.

> This type of new complexity in the name of testing does not seem like a good idea to me, in terms of long term maintenance. 

Increasing maintenance cost (more complex code) in exchange for better testing seems like a good tradeoff to me.  With out dependency injection, we can't test these parts of the code without creating rostests.  The downside to rostests is they are executing a bunch of code within ros and therefore our tests can go flaky when there are changes within ros.  This is the situation with a the tests in moveit_servo and the ompl constraint planner on foxy.  With dependency injection we can write tests that only test the code in moveit without testing ros itself.

> What's a solution we can use for all of MoveIt? Or for all ROS applications?

This pattern is something we can use for all of moveit.  It adds complexity to the implementation of various classes without changing the interfaces of those classes in exchange for tests of the behavior in those classes.  We are currently debugging a concurrency issue inside PlanningSceneMonitor for a client that with this pattern we will be able to add tests to isolate (and fix) those issues.

This PR is very limited in its scope and does not include in depth testing of CurrentStateMonitor.  That was intentional as we are using this PR as a pattern for how to add the infrastructure for dependency injection based unit tests.

Another benefit of this which I think @v4hn (and @henningkayser and @JafarAbdi who have been doing the syncs) will especially like is that it should make syncing improvements between ros1 and ros2 versions of moveit much easier (after these tests are backported, which should be fairly trivial).  The reason for this is it isolates the ROS code to these MiddlewareHandle implementation classes from the actual logic in moveit itself.

henningkayser (approved):

PR #563, opened 2021-07-22
Optional thread deadline scheduling for Servo

AndyZe:
This allows users to allocate plenty of CPU time to a thread. Description of thread scheduling is here:

https://nicolovaligi.com/articles/concurrency-and-parallelism-in-ros1-and-ros2-linux-kernel-tools/

(Under The Deadline Scheduler section)

@zultron does this make sense to you?

### DIFF
--- a/moveit_ros/moveit_servo/src/servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/src/servo_calcs.cpp
@@ -43,6 +43,9 @@
 #include <mutex>
 
 #include <std_msgs/msg/bool.h>
+// Optionally schedule a thread deadline for closer-to-realtime performance
+#include <linux/sched.h>
+#include <sys/syscall.h>
 
 // #include <moveit_servo/make_shared_from_pool.h> // TODO(adamp): create an issue about this
 #include <moveit_servo/servo_calcs.h>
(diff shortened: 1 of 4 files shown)


moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -43,6 +43,9 @@
 #include <mutex>
 
 #include <std_msgs/msg/bool.h>
+// Optionally schedule a thread deadline for closer-to-realtime performance
+#include <linux/sched.h>
```
AdamPettinger, line 47: We just had a discussion today about MoveIt 2 on Windows, does this make Windows support complicated?

AndyZe: yes...  :/

AndyZe: How about wrapping this stuff in a preprocessor directive to check if it's on a Unix system?

zultron:
I don't really understand the context here, or what the problem is that this code addresses.  Sounds like you have a thread that's not getting enough CPU time because it's getting preempted?

I haven't used the `SCHED_DEADLINE` scheduling policy in the past, so I'm not sure exactly how it behaves.  The 100% will allow the thread to run 100% of the time on its CPU if it wants to.  Does this thread need that much CPU?  If it's possible to run more than one of these threads, there's a risk of locking up the machine.  Consider setting the limit to 95% with a shorter period so that the system may be recoverable if something bad happens.

Most of my experience with scheduling policy surrounds `SCHED_FIFO` for scheduling high priority tasks, like periodic servo update routines.  In these scenarios, the issue is scheduling latency rather than CPU bandwidth.

AndyZe:
We were trying to debug jerky motion on a UR5e but it looks like moveit_servo isn't the cause. OK, thanks @zultron.

Closing for now.

PR #564, opened 2021-07-22
Delete "stop distance"-based collision checking

AndyZe:
This was an optional collision prevention method that was based on how long it would take the robot to come to a stop. It was flaky and never really worked great, in my experience. Probably due to latency -- by the time the robot is very close to collision, a 10ms delay due to a ROS message hop is pretty severe.

I would like to delete it just as a matter of cleaning up the code.


tylerjw (approved):
Generally approve of removing code if you don't think it worked well.  Do you think you'll backport this?

AndyZe:
Thanks. Prob not worth the effort of backporting in my opinion.

v4hn:
> Probably due to latency -- by the time the robot is very close to collision, a 10ms delay due to a ROS message hop is pretty severe.

Isn't that the main point about ROS2? :)
Did you ever test this feature in a ROS2 project with single-process moveit_servo+control_loop setup (and possibly with low acc/deceleration values)?

> I would like to delete it just as a matter of cleaning up the code.

Fair enough either way.

AndyZe:
@tylerjw how are we handling the foxy branch now? Merge to main and make a separate PR for merging to foxy?

AndyZe:
@v4hn I compared the default "threshold distance" collision check to this "stop distance" that I want to delete. This was done in ROS2. It's pretty obvious the "threshold distance" performs better. "stop distance" might work alright if run at a very high rate but at this 10 Hz rate it wasn't effective at all.

threshold distance

https://user-images.githubusercontent.com/11284393/128218777-cd356a48-a974-4d3f-8346-b483360ce55e.mp4

stop distance

https://user-images.githubusercontent.com/11284393/128218796-bd4ca5e6-a7a3-41d9-8ed4-eb93ddc77418.mp4

v4hn:
Thanks for going through the trouble to make these videos @AndyZe :medal_sports:

PR #565, opened 2021-07-22
Update repos files for CI

tylerjw:
### Description

This updates the repos files and CI jobs to reflect what is released and needed for doing CI jobs.


tylerjw:
The basic idea here is that so long as the released versions of our upstream repos doesn't have a breaking change we depend on we can install them via the debains in our tests.  As soon as someone makes a change where they need the repos file and we haven't yet released they can comment them back out until those are synced into the buildfarm.  Currently the only upstream repos we need to build are these:

rolling: 
* warehouse_ros (waiting on sync)
* warehouse_ros_mongo (waiting on sync)
* geometric_shapes (broken on rolling, needs PR to fix it)
galactic: none
foxy: none

vatanaksoytezer (approved):
As long as we don't forget to ask people to uncomment this back when they want to merge a PR that is depending on our upstreams, this is good, adds more reliablity to binaries and increases the CI time.

tylerjw:
@henningkayser could you give this a review?  Do you like the general idea of this?

henningkayser (approved):

PR #568, opened 2021-07-23
Fix crash if MoveIt Servo collision check is disabled

schornakj:
### Description

If the Servo collision check parameter is set to `false`, `CollisionCheck`'s `rclcpp::TimerBase::SharedPtr` `timer_` member is never initialized. Since `~CollisionCheck()` calls `timer_->cancel()`, requesting that Servo start if collision checking was not enabled causes a crash.

This fixes the crash by checking if `timer_` was actually set before trying to cancel it.

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/include/moveit_servo/collision_check.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/collision_check.h
@@ -71,7 +71,10 @@ public:
 
   ~CollisionCheck()
   {
-    timer_->cancel();
+    if (timer_)
+    {
+      timer_->cancel();
+    }
   }
 
   /** \brief start the Timer that regulates collision check rate */


jliukkonen (approved):
Why is the timer `shared_ptr` in the first place? In MoveIt1 it is just `ros::Timer`. This will fix the problem but I think the fix could also be unifying the timer initialization to match the MoveIt1.

JafarAbdi (approved):

schornakj:
> Why is the timer `shared_ptr` in the first place? In MoveIt1 it is just `ros::Timer`. This will fix the problem but I think the fix could also be unifying the timer initialization to match the MoveIt1.

In ROS1, the NodeHandle's `createTimer` function only returns a `ros::Timer`. The equivalent function in ROS2 is Node's `create_wall_timer`, which only returns a `shared_ptr`. ROS2 made a deliberate decision to interact with publishers, subscribers, timers, etc. through `shared_ptr`s, and while the rationale behind that is still debatable (`unique_ptr` would be preferable in many cases) that's the approach which is favored by the API.

AndyZe (approved):
Thanks! Good catch.

PR #569, opened 2021-07-23
OccupancyMapMonitor tests using Dependency Injection

tylerjw:
### Description

Dependency Injection for OccupancyMapMonitor.  More tests to come.

### Checklist


moveit_ros/occupancy_map_monitor/test/occupancy_map_monitor_tests.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+  // THEN we expect it to call these methods on the MiddlewareHandle
+  EXPECT_CALL(*mock_middleware_handle, getMapFrameParameter);
+  EXPECT_CALL(*mock_middleware_handle, getSensorListParameter);
+  EXPECT_CALL(*mock_middleware_handle, getMapResolutionParameter);
+  EXPECT_CALL(*mock_middleware_handle, createSaveMapService);
+  EXPECT_CALL(*mock_middleware_handle, createLoadMapService);
+
+  // WHEN we start the current state monitor
+  occupancy_map_monitor::OccupancyMapMonitor occupancy_map_monitor{
```
jliukkonen, line 72: Isn't it GIVEN, WHEN, THEN? Or am I missing something?

moveit_ros/occupancy_map_monitor/test/occupancy_map_monitor_tests.cpp
```diff
@@ -0,0 +1,81 @@
...
+              (const std::string& sensor_plugin), (override));
+  MOCK_METHOD(void, initializeOccupancyMapUpdater,
+              (occupancy_map_monitor::OccupancyMapUpdaterPtr occupancy_map_updater), (override));
+  MOCK_METHOD(void, createSaveMapService, (occupancy_map_monitor::SaveMapServiceCallback callback), (override));
+  MOCK_METHOD(void, createLoadMapService, (occupancy_map_monitor::LoadMapServiceCallback callback), (override));
+};
+
+TEST(OccupancyMapMonitorTests, ConstructorTest)
+{
+  // GIVEN a mocked middlware handle
```
jliukkonen, line 61: Nit: typo in middleware.

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
```diff
@@ -39,124 +39,95 @@
...
 
   if (map_frame_.empty())
-    if (!node_->get_parameter("octomap_frame", map_frame_))
-      if (tf_buffer_)
-        RCLCPP_WARN(LOGGER, "No target frame specified for Octomap. No transforms will be applied to received data.");
+  {
+    map_frame_ = middleware_handle_->getMapFrameParameter();
+  }
 
+  if (tf_buffer_ && map_frame_.empty())
```
jliukkonen, line 89: Nit: comparing pointers explicitly to `nullptr`s is clearer than using the overload.

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
```diff
@@ -39,124 +39,95 @@
...
+
+OccupancyMapMonitor::OccupancyMapMonitor(const rclcpp::Node::SharedPtr& node,
+                                         const std::shared_ptr<tf2_ros::Buffer>& tf_buffer,
+                                         const std::string& map_frame, double map_resolution)
+  : OccupancyMapMonitor(std::make_unique<OccupancyMapMonitorMiddlewareHandle>(node), tf_buffer, map_frame,
+                        map_resolution)
+{
+}
+
 void OccupancyMapMonitor::initialize()
```
jliukkonen, line 79: I guess it would break the API and is therefore a no-go but I've been starting to lean on removing initialization functions since it is the constructor whose job it is to make sure the object is in usable state.

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
```diff
@@ -39,124 +39,95 @@
...
+#include <moveit/occupancy_map_monitor/occupancy_map_monitor_middleware_handle.hpp>
 #include <boost/bind.hpp>
 
 namespace occupancy_map_monitor
 {
-static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit.ros.occupancy_map_monitor");
-
-OccupancyMapMonitor::OccupancyMapMonitor(const rclcpp::Node::SharedPtr& node, double map_resolution)
-  : map_resolution_(map_resolution), debug_info_(false), mesh_handle_count_(0), node_(node), active_(false)
+namespace
```
jliukkonen, line 47: Nit: Should this namespace be outside the occupancy_map_monitor namespace for better readability? My personal preference is to have them be outside any namespace to underscore they "anonymousness".

moveit_ros/occupancy_map_monitor/include/moveit/occupancy_map_monitor/occupancy_map_monitor.h
```diff
@@ -137,26 +244,22 @@ class OccupancyMapMonitor
   bool getShapeTransformCache(std::size_t index, const std::string& target_frame, const rclcpp::Time& target_time,
                               ShapeTransformCache& cache) const;
 
+  std::unique_ptr<MiddlewareHandle> middleware_handle_;
   std::shared_ptr<tf2_ros::Buffer> tf_buffer_;
   std::string map_frame_;
   double map_resolution_;
-  boost::mutex parameters_lock_;
+  std::mutex parameters_lock_;
```
jliukkonen, line 251: This could be marked `mutable`.

moveit_ros/occupancy_map_monitor/include/moveit/occupancy_map_monitor/occupancy_map_monitor.h
```diff
@@ -36,30 +36,137 @@
...
+     * @return     The sensor plugin parameter.
+     */
+    virtual std::string getSensorPluginParameter(const std::string& sensor) = 0;
+
+    /**
+     * @brief      Gets the sensor list parameter.
+     *
+     * @return     The sensor list parameter.
+     */
+    virtual std::vector<std::string> getSensorListParameter() = 0;
```
jliukkonen, line 103: Every getter here can be marked `const` right?

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
```diff
@@ -39,124 +39,95 @@
...
   tree_const_ = tree_;
 
-  std::vector<std::string> sensor_list;
-  if (node_->get_parameter("sensors", sensor_list))
+  std::vector<std::string> sensor_list = middleware_handle_->getSensorListParameter();
+
+  for (const auto& sensor : sensor_list)
   {
-    if (!sensor_list.empty())
+    std::string sensor_plugin = middleware_handle_->getSensorPluginParameter(sensor);
```
jliukkonen, line 106: `const std::string` ?

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
```diff
@@ -39,124 +39,95 @@
...
                         "No transforms will be applied to received data.");
+  }
 
-  tree_.reset(new OccMapTree(map_resolution_));
+  tree_ = std::make_shared<OccMapTree>(map_resolution_);
   tree_const_ = tree_;
 
-  std::vector<std::string> sensor_list;
-  if (node_->get_parameter("sensors", sensor_list))
+  std::vector<std::string> sensor_list = middleware_handle_->getSensorListParameter();
```
jliukkonen, line 102: `const`?

jliukkonen (commented):
My initial review. I think I want to leave another round of comments later when I have time to go through it more thoroughly.

moveit_ros/occupancy_map_monitor/test/occupancy_map_monitor_tests.cpp
tylerjw: Yes, but because of how gmock works it goes in this order.  The expect macro has to come before the creation of the OccupancyMapMonitor where the mock object is moved into it.

moveit_ros/occupancy_map_monitor/test/occupancy_map_monitor_tests.cpp
tylerjw: agree, changing

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
tylerjw: In this case, initialize is a private method that is only called by the constructor.  I'm moving its code into the constructor and removing the method to make it more clear.  That change does not break API in this case.

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
tylerjw: I'm just copying a pattern I've seen in other parts of the MoveIt code base.  I would prefer consistency, although I agree that having a separate anonymous namespace seems to be clearer.  For consistency sake I want to leave this as is.

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
tylerjw: sure, wouldn't hurt here

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
tylerjw: :+1:

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
griswaldbrooks: > but I've been starting to lean on removing initialization functions

:100:

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
griswaldbrooks: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rconst-const

I would argue that it always hurts when a scope variable that can be `const`, isn't.

moveit_ros/occupancy_map_monitor/include/moveit/occupancy_map_monitor/occupancy_map_monitor.h
tylerjw: I'm not sure this mutex is even needed (if you look at its use it seems silly).  However, I was trying to not include a refactor of that in this change as I want to refactor all the parameter stuff at once.

Instead, I am just doing a basic boost->stl refactor along with these changes where it makes sense.  I can make this mutable but that wouldn't matter much.

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
```diff
@@ -39,124 +39,90 @@
...
+  , active_(false)
 {
-  /* load params from param server */
-  if (map_resolution_ <= std::numeric_limits<double>::epsilon())
-    if (!node_->get_parameter("octomap_resolution", map_resolution_))
-    {
-      map_resolution_ = 0.1;
-      RCLCPP_WARN(LOGGER, "Resolution not specified for Octomap. Assuming resolution = %g instead", map_resolution_);
-    }
+  map_resolution_ = middleware_handle_->getMapResolutionParameter();
```
griswaldbrooks, line 76: It looks like the original version checks the ctor set value first, whereas this version overrides it regardless. Was this the intent? If so, what's the point of the ctor argument?

griswaldbrooks (changes requested):

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
griswaldbrooks: Just putting it out there (not trying to change things), this is an area where cppcoreguidelines and abseil disagree

https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es87-dont-add-redundant--or--to-conditions

https://abseil.io/tips/141

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
tylerjw: This is in the cpp so I would assume they wouldn't export

moveit_ros/occupancy_map_monitor/include/moveit/occupancy_map_monitor/occupancy_map_monitor.h
```diff
@@ -36,30 +36,137 @@
...
-#include <vector>
-#include <string>
+#include <functional>
+#include <memory>
+#include <mutex>
 #include <rclcpp/rclcpp.hpp>
+#include <string>
 #include <tf2_ros/buffer.h>
-#include <pluginlib/class_loader.hpp>
+#include <vector>
```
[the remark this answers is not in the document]
tylerjw: I reordered them to be alphabetically and to be in two groups (not-moveit, moveit).  I can split them so they are (stl, not-moveit, moveit) if you'd like?

moveit_ros/occupancy_map_monitor/include/moveit/occupancy_map_monitor/occupancy_map_monitor.h
```diff
@@ -36,30 +36,137 @@
...
+     * @brief      Gets the map frame parameter.
+     *
+     * @return     The map frame parameter.
+     */
+    virtual std::string getMapFrameParameter() = 0;
+
+    /**
+     * @brief      Gets the sensor plugin parameter.
+     *
+     * @param[in]  sensor  The sensor
```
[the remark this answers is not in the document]
tylerjw: I generated these comments using a tool.  I can remove them as they do basically seem like noise.

moveit_ros/occupancy_map_monitor/include/moveit/occupancy_map_monitor/occupancy_map_monitor.h
```diff
@@ -36,30 +36,137 @@
...
 #include <moveit/occupancy_map_monitor/occupancy_map.h>
 #include <moveit/occupancy_map_monitor/occupancy_map_updater.h>
 
-#include <boost/thread/mutex.hpp>
-
-#include <memory>
-
 namespace occupancy_map_monitor
 {
+using SaveMapServiceCallback = std::function<bool(const std::shared_ptr<rmw_request_id_t> request_header,
```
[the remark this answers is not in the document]
tylerjw: I like that idea, as it is only used by the middleware handle.

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor_middleware_handle.cpp
```diff
@@ -0,0 +1,138 @@
...
+  {
+    RCLCPP_FATAL_STREAM(LOGGER, "Exception while creating octomap updater plugin loader " << ex.what());
+    throw ex;
+  }
+}
+
+double OccupancyMapMonitorMiddlewareHandle::getMapResolutionParameter()
+{
+  double map_resolution = 0.1;
+  if (!node_->get_parameter("octomap_resolution", map_resolution))
```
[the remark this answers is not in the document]
tylerjw: I was able to put all of them but getSensorPluginParameter in the constructor and have them be const.

moveit_ros/occupancy_map_monitor/include/moveit/occupancy_map_monitor/occupancy_map_monitor.h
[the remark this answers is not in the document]
griswaldbrooks: I would rather they exist with better comments

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor_middleware_handle.cpp
```diff
@@ -0,0 +1,138 @@
...
+  }
+  return sensor_plugin;
+}
+
+OccupancyMapUpdaterPtr OccupancyMapMonitorMiddlewareHandle::loadOccupancyMapUpdater(const std::string& sensor_plugin)
+{
+  OccupancyMapUpdaterPtr occupancy_map_updater = nullptr;
+  try
+  {
+    occupancy_map_updater = updater_plugin_loader_->createUniqueInstance(sensor_plugin);
```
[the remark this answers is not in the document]
tylerjw: I generally try to avoid multiple returns if I can, but in this case, it seems simple enough and that is clearer.  Also, I would trust that the compiler would help optimize this local variable out.

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
```diff
@@ -39,124 +39,90 @@
...
 namespace occupancy_map_monitor
 {
+namespace
+{
 static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit.ros.occupancy_map_monitor");
+}
 
 OccupancyMapMonitor::OccupancyMapMonitor(const rclcpp::Node::SharedPtr& node, double map_resolution)
-  : map_resolution_(map_resolution), debug_info_(false), mesh_handle_count_(0), node_(node), active_(false)
+  : OccupancyMapMonitor(std::make_unique<OccupancyMapMonitorMiddlewareHandle>(node), nullptr, "", map_resolution)
```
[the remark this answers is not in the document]
tylerjw: yes

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
```diff
@@ -39,124 +39,90 @@
...
-  , node_(node)
-{
-  initialize();
-}
-
-void OccupancyMapMonitor::initialize()
+  , active_(false)
 {
-  /* load params from param server */
-  if (map_resolution_ <= std::numeric_limits<double>::epsilon())
```
tylerjw, line 70: I missed this.  The default value is 0.0 which causes it to look for the parameter.

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
tylerjw: I didn't mean to change the behavior, putting the old  behavior back.

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor_middleware_handle.cpp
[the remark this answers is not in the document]
griswaldbrooks: I don't usually consider what the compiler would do most of the time.
Maybe because of how I usually organize my code, multiple returns don't seem so annoying to me (I try to only return at the beginning or end). For me, the thread of extra non-const local variables is greater.

Is this why you avoid multiple returns? https://softwareengineering.stackexchange.com/questions/118703/where-did-the-notion-of-one-return-only-come-from

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
```diff
@@ -39,124 +39,90 @@
...
-    }
+  map_resolution_ = middleware_handle_->getMapResolutionParameter();
   RCLCPP_DEBUG(LOGGER, "Using resolution = %lf m for building octomap", map_resolution_);
 
   if (map_frame_.empty())
-    if (!node_->get_parameter("octomap_frame", map_frame_))
-      if (tf_buffer_)
-        RCLCPP_WARN(LOGGER, "No target frame specified for Octomap. No transforms will be applied to received data.");
+  {
+    map_frame_ = middleware_handle_->getMapFrameParameter();
```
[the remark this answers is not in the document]
tylerjw: I've updated this to more closely follow the original logic.  lmk what you think.

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor_middleware_handle.cpp
```diff
@@ -0,0 +1,138 @@
...
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Tyler Weaver */
+
+#include <moveit/occupancy_map_monitor/occupancy_map_monitor_middleware_handle.hpp>
+#include <pluginlib/class_loader.hpp>
+#include <string>
+#include <vector>
+
```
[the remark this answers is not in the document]
tylerjw: done

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor_middleware_handle.cpp
```diff
@@ -0,0 +1,138 @@
...
+namespace occupancy_map_monitor
+{
+namespace
+{
+static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit.ros.occupancy_map_monitor.middleware_handle");
+}
+
+OccupancyMapMonitorMiddlewareHandle::OccupancyMapMonitorMiddlewareHandle(const rclcpp::Node::SharedPtr& node)
+  : node_(node)
```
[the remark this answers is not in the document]
tylerjw: done

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor_middleware_handle.cpp
```diff
@@ -0,0 +1,138 @@
...
+{
+  try
+  {
+    updater_plugin_loader_ = std::make_unique<pluginlib::ClassLoader<OccupancyMapUpdater> >(
```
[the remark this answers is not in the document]
tylerjw: you don't, fixed... our clang_format will allow either apparently

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor_middleware_handle.cpp
```diff
@@ -0,0 +1,138 @@
...
+  {
+    RCLCPP_ERROR_STREAM(LOGGER, "Exception while loading octomap updater '" << sensor_plugin
+                                                                            << "': " << exception.what() << std::endl);
+  }
+  return occupancy_map_updater;
+}
+
+void OccupancyMapMonitorMiddlewareHandle::initializeOccupancyMapUpdater(OccupancyMapUpdaterPtr occupancy_map_updater)
+{
+  if (!occupancy_map_updater->initialize(node_))
```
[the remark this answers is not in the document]
tylerjw: I agree... something is going to have to be done about these pluginlib things when we do some more real testing.  Also they need tests themselves.

moveit_ros/occupancy_map_monitor/include/moveit/occupancy_map_monitor/occupancy_map_monitor.h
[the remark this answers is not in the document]
griswaldbrooks: Looks like you did the second, which what I'm used to 
https://github.com/fetchrobotics/fetch_open_auto_dock/blob/melodic-devel/include/fetch_auto_dock/linear_pose_filter_2d.h#L22
 but splitting the stl is the minimum for me.

Though, when I wrote that, our format was
```
// C++ Standard Library
...
// Boost
...
// ROS
...
```
I'm not sure why they felt that `includes.` was necessary as it's really obvious that they are includes.
```

File: moveit_ros/occupancy_map_monitor/include/moveit/occupancy_map_monitor/occupancy_map_monitor.h [hunk 23]
```diff
@@ -36,30 +36,80 @@
...
+                                                      std::shared_ptr<moveit_msgs::srv::SaveMap::Response> response)>;
+    using LoadMapServiceCallback = std::function<bool(const std::shared_ptr<rmw_request_id_t> request_header,
+                                                      const std::shared_ptr<moveit_msgs::srv::LoadMap::Request> request,
+                                                      std::shared_ptr<moveit_msgs::srv::LoadMap::Response> response)>;
+
+    virtual ~MiddlewareHandle() = default;
+    virtual double getMapResolutionParameter() const = 0;
+    virtual std::string getMapFrameParameter() const = 0;
+    virtual std::vector<std::string> getSensorListParameter() const = 0;
+    virtual std::string getSensorPluginParameter(const std::string& sensor) = 0;
```
griswaldbrooks, line 75: Where did the documentation go? Also, this couldn't be made `const`? Why am I not surprised? Is it because of the node handle? Would it make sense to make it `mutable` so that anything else would have `const` enforcement?

moveit_ros/occupancy_map_monitor/include/moveit/occupancy_map_monitor/occupancy_map_monitor.h
```diff
@@ -36,30 +36,80 @@
...
+   * @brief      Occupancy map monitor constructor with the MiddlewareHandle
+   *
+   * @param[in]  middleware_handle   The rcl interface
+   * @param[in]  tf_buffer           The tf buffer
+   */
+  OccupancyMapMonitor(std::unique_ptr<MiddlewareHandle> middleware_handle,
+                      const std::shared_ptr<tf2_ros::Buffer>& tf_buffer);
+
+  /**
+   * @brief      Occupancy map montior constructor with Node
```
griswaldbrooks, line 92: ~montior~ `monitor`

moveit_ros/occupancy_map_monitor/include/moveit/occupancy_map_monitor/occupancy_map_monitor.h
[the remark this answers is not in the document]
griswaldbrooks: :point_up:

moveit_ros/occupancy_map_monitor/include/moveit/occupancy_map_monitor/occupancy_map_monitor_middleware_handle.hpp
```diff
@@ -0,0 +1,128 @@
...
+
+#pragma once
+
+#include <moveit/occupancy_map_monitor/occupancy_map_monitor.h>
+#include <moveit/occupancy_map_monitor/occupancy_map_updater.h>
+#include <pluginlib/class_loader.hpp>
+#include <rclcpp/rclcpp.hpp>
+#include <string>
+#include <vector>
+
```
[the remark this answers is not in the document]
griswaldbrooks: :point_up:

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
griswaldbrooks: But do they?

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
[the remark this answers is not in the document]
griswaldbrooks: jsyk, I still don't know how to resolve comments.

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor_middleware_handle.cpp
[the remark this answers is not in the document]
tylerjw: It isn't something from writing assembly (I never have done much of that).  It was simply a style I found more readable.  Although in this case, I do like it better with multiple after changing it.

moveit_ros/occupancy_map_monitor/include/moveit/occupancy_map_monitor/occupancy_map_monitor.h
tylerjw: It is because it has to call get_parameter... I could move more logic into the constructor of the handle but I was trying to keep it as logic free as possible.

moveit_ros/occupancy_map_monitor/CMakeLists.txt
```diff
@@ -86,6 +87,16 @@ if(BUILD_TESTING)
 
   # Run all lint tests in package.xml except those listed above
   ament_lint_auto_find_test_dependencies()
+
+  find_package(ament_cmake_gtest REQUIRED)
```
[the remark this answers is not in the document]
tylerjw: no, fixed

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
tylerjw: How would I know for sure? I have just assumed that anything declared as a file global without an export in some header had only file scope.

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor_middleware_handle.cpp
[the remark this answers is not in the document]
griswaldbrooks: Not assembly, but C

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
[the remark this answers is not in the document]
griswaldbrooks: Looks a lot better

moveit_ros/occupancy_map_monitor/include/moveit/occupancy_map_monitor/occupancy_map_monitor.h
griswaldbrooks: Also, I just noticed, that in the ctor of `OccupancyMapMonitor`, `getSensorListParameter` is called, and then immediately passed to `getSensorPluginParameter`, and isn't stored. Why is it better to have this logic in the `OccupancyMapMonitor` ctor, rather than the `MiddlewareHandle`, which also has logic. You could also then remove `getSensorListParameter` from the public interface.

griswaldbrooks (commented):
Added comments. Also, not having the ability to resolve things is very frustrating :rocket:

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
gbrooks2-pn: Well, as cppreference points out, things in anonymous namespaces can be imported.
In your gtest, I would try to access something in your anon namespace defined in your cpp

griswaldbrooks: InverseKinematics234!

griswaldbrooks: Well, as cppreference points out, things in anonymous namespaces can be imported.
In your gtest, I would try to access something in your anon namespace defined in your cpp

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor_middleware_handle.cpp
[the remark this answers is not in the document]
v4hn: Historical background:
The code base has/had a lot of looong if blocks just because the original authors disliked `if(!...) return;`.
I brought this up in a maintainer meeting years ago and we agreed that early return should be preferred over five layers of conditional blocks.

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
v4hn: cppreference:
> Even though names in an unnamed namespace may be declared with external linkage, they are never accessible from other translation units because their namespace name is unique. 

If that's what you mean by export here. But I guess you knew that before.
This only affects the individual cpp file and if anyone defines a second LOGGER there, you should change the names anyway to reflect the different meanings.

Usually I prefer anonymous namespaces **inside** file-wide namespaces to avoid having to spell out the file-wide namespace each time I need to access other symbols from there. While this is not used here, it makes some sense to keep the style consistent.

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor_middleware_handle.cpp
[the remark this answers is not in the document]
griswaldbrooks: Exactly. A previous codebase I worked on had a huge chain of nested blocks, which, inevitably when someone tried to maintain it, they'd get wrong (ie me). I read that part of that comes from a C style (which this codebase had originated as) where you didn't have RAII and have to worry about cleanup lest you leak resources. Nowadays I try to organize my code as
```
// Preconditions
if (condition1()) return;
if (condition2()) return;
if (condition3()) return;

// Body
do_work();

// Postconditions
if (condition4()) return 4;
return ultimate;
```
or something like that. I try to avoid returning in the body if I can.

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
griswaldbrooks: Yeah, it didn't seem likely that anything would be exported if declared in the cpp because of the external linkage. I guess it was a bit of sticker shock since I didn't know that anon-namespace anything could be exported.

file-wide namespace?

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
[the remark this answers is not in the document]
tylerjw: I know... I've been resolving some of them (including this one) when I think I've fixed them.  Although I think it is broken to have me resolving them when you were the one who created the thread.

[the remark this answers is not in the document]
griswaldbrooks: Yeah, but @nbbrooks found https://docs.github.com/en/github/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/about-pull-request-reviews#resolving-conversations so, not much to do in the short term

moveit_ros/occupancy_map_monitor/include/moveit/occupancy_map_monitor/occupancy_map_monitor.h
tylerjw: fixed

tylerjw:
> Added comments. Also, not having the ability to resolve things is very frustrating rocket

I agree, this seems like a github bug.  https://twitter.com/weaveringrally/status/1420517755380436997  I couldn't find a github issue and found someone saying that tweeting at github has worked for them before at getting their attention.

I updated all the files I've touched with the same format for includes and consistent doxygen comments with descriptions as best as I understand.

moveit_ros/occupancy_map_monitor/include/moveit/occupancy_map_monitor/occupancy_map_monitor_middleware_handle.hpp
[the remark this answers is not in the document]
tylerjw: done

moveit_ros/occupancy_map_monitor/include/moveit/occupancy_map_monitor/occupancy_map_monitor.h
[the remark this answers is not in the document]
tylerjw: I put them back with mostly manually written comments to the best of my knowledge of the various pieces.

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor_middleware_handle.cpp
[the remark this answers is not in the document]
tylerjw: done

moveit_ros/occupancy_map_monitor/include/moveit/occupancy_map_monitor/occupancy_map_monitor.h
tylerjw: done, lmk what you think of my implementation of this.

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
v4hn: > file-wide namespace?

It's not a technical term. Some maintainers introduced namespace declarations that span virtually the whole file for many packages in MoveIt (e.g., [this one](https://github.com/ros-planning/moveit/commit/e01314474b2b06065daaea2a54b9d592f0f7e3c8)) to avoid the full namespace declaration for each symbol.

tylerjw:
It seems that re-ordering the includes broke the Foxy CI run.  I've installed foxy locally and am working on figuring this one out.

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
```diff
@@ -34,129 +34,104 @@
...
-  tree_.reset(new OccMapTree(map_resolution_));
+  tree_ = std::make_shared<OccMapTree>(map_resolution_);
   tree_const_ = tree_;
 
-  std::vector<std::string> sensor_list;
-  if (node_->get_parameter("sensors", sensor_list))
+  for (const auto& pair : middleware_handle_->getSensorPluginsParameter())
   {
-    if (!sensor_list.empty())
+    const auto& sensor_name = pair.first;
```
henningkayser, line 107: C++17 anyone? structured binding would be nice here

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
```diff
@@ -34,129 +34,104 @@
...
-      RCLCPP_WARN(LOGGER, "Resolution not specified for Octomap. Assuming resolution = %g instead", map_resolution_);
-    }
-  RCLCPP_DEBUG(LOGGER, "Using resolution = %lf m for building octomap", map_resolution_);
+  if (middleware_handle_ == nullptr)
+  {
+    throw std::invalid_argument("OccupancyMapMonitor cannot be constructed with nullptr MiddlewareHandle");
+  }
+
+  // Get the parameters
+  map_frame_ = middleware_handle_->getMapFrameParameter();
```
henningkayser, line 87: Things like this make me think that classes should just use structs for their parameter configs. The middleware handle could simply initialize those from `rclcpp::Parameters` or manually in case of unit tests.

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
tylerjw: done

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
tylerjw: done, lmk what you think of the implementation

moveit_ros/occupancy_map_monitor/include/moveit/occupancy_map_monitor/occupancy_map_monitor.h
```diff
@@ -36,127 +36,302 @@
...
-  std::string map_frame_;
-  double map_resolution_;
-  boost::mutex parameters_lock_;
-
-  OccMapTreePtr tree_;
-  OccMapTreeConstPtr tree_const_;
+  std::unique_ptr<MiddlewareHandle> middleware_handle_; /*!< The abstract interface to ros */
+  std::shared_ptr<tf2_ros::Buffer> tf_buffer_;          /*!< TF buffer */
+  Parameters parameters_;
+  std::mutex parameters_lock_; /*!< Mutex for synchronizing access to parameters */
```
jliukkonen-pn, line 323: I just realized that sometimes the access to parameters is protected by mutex but most of the time it isn't. Is that a bug or am I just reading the code wrong?

tylerjw: This mutex is never actually used to synchronize anything.  I think someone had a plan to do that but didn't.  I am separately working on implementing dynamic parameters for moveit2 and will clean this up in an unrelated change.

gbrooks2-pn (approved):

griswaldbrooks (approved):

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor_middleware_handle.cpp
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    if (sensor_plugin.empty() || sensor_plugin[0] == '~')
+    {
+      RCLCPP_INFO_STREAM(LOGGER, "Skipping octomap updater plugin '" << sensor_plugin << "'");
+      continue;
+    }
+    else
+    {
+      sensor_plugins_.push_back(std::make_pair(sensor_name, sensor_plugin));
```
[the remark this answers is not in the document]
tylerjw: Sure, generally I avoid emplace_back because it is easy to create unintended situations with it.  However in this case I think it does make sense.

https://abseil.io/tips/112

henningkayser (approved):

PR #570, opened 2021-07-23
 Tests for TrajectoryMonitor using dependency injection

Abishalini:
### Description

### Checklist


moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/trajectory_monitor_middleware_handle.hpp
```diff
@@ -0,0 +1,66 @@
...
+ */
+class TrajectoryMonitorMiddlewareHandle : public TrajectoryMonitor::MiddlewareHandle
+{
+public:
+  /**
+   * @brief      Constructor
+   *
+   * @param[in]  node  The ros node
+   */
+  TrajectoryMonitorMiddlewareHandle(const double sampling_frequency);
```
jliukkonen, line 56: The type can be just `double`. Constness isn't required.

moveit_ros/planning/planning_scene_monitor/src/trajectory_monitor_middleware_handle.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+/* Author: Abishalini Sivaraman */
+
+#include <moveit/planning_scene_monitor/trajectory_monitor_middleware_handle.hpp>
+#include <rclcpp/rclcpp.hpp>
+
+namespace planning_scene_monitor
+{
+planning_scene_monitor::TrajectoryMonitorMiddlewareHandle::TrajectoryMonitorMiddlewareHandle(
+    const double sampling_frequency)
```
jliukkonen, line 43: `const` can be removed.

moveit_ros/planning/planning_scene_monitor/test/trajectory_monitor_tests.cpp
```diff
@@ -0,0 +1,111 @@
...
+TEST(TrajectoryMonitorTests, test1)
+{
+  auto mock_trajectory_monitor_middleware_handle = std::make_unique<MockTrajectoryMonitorMiddlewareHandle>();
+  auto mock_current_state_monitor_middleware_handle = std::make_unique<MockCurrentStateMonitorMiddlewareHandle>();
+
+  // THEN we expect it to call addSleep function
+  EXPECT_CALL(*mock_trajectory_monitor_middleware_handle, sleep).Times(::testing::AtLeast(1));
+
+  // GIVEN a TrajectoryMonitor is started
+  planning_scene_monitor::CurrentStateMonitorPtr current_state_monitor;
```
jliukkonen, line 88: Should it be GIVEN, WHEN and THEN after those two?

moveit_ros/planning/planning_scene_monitor/test/trajectory_monitor_tests.cpp
```diff
@@ -0,0 +1,111 @@
...
+                                                                std::move(mock_trajectory_monitor_middleware_handle),
+                                                                10.0 };
+
+  std::atomic<bool> callback_called{ false };
+  trajectory_monitor.setOnStateAddCallback(
+      [&](const moveit::core::RobotStateConstPtr& robot_state, const rclcpp::Time& time) { callback_called = true; });
+
+  // WHEN we call the startTrajectoryMonitor function
+  trajectory_monitor.startTrajectoryMonitor();
+  waitFor(10s, [&]() { return callback_called == true; });
```
jliukkonen, line 104: Will this test run for 10 seconds? Probably should be in millisecond range? Or?

moveit_ros/planning/planning_scene_monitor/CMakeLists.txt
```diff
@@ -46,6 +47,12 @@ if(BUILD_TESTING)
     test/current_state_monitor_tests.cpp
   )
   target_link_libraries(current_state_monitor_tests
+  ${MOVEIT_LIB_NAME}
+  )
```
jliukkonen, line 51: Nit: the formatting looks a bit weird here.

jliukkonen (commented):
Good to have tests for this component :)

moveit_ros/planning/planning_scene_monitor/test/trajectory_monitor_tests.cpp
tylerjw: @jliukkonen the pattern order is this way because of how gmock works.

moveit_ros/planning/planning_scene_monitor/test/trajectory_monitor_tests.cpp
```diff
@@ -0,0 +1,111 @@
...
+};
+
+struct MockCurrentStateMonitorMiddlewareHandle : public planning_scene_monitor::CurrentStateMonitor::MiddlewareHandle
+{
+  MOCK_METHOD(rclcpp::Time, now, (), (const, override));
+  MOCK_METHOD(void, createJointStateSubscription,
+              (const std::string& topic, planning_scene_monitor::JointStateUpdateCallback callback), (override));
+  MOCK_METHOD(void, resetJointStateSubscription, (), (override));
+  MOCK_METHOD(std::string, getJointStateTopicName, (), (const, override));
+};
```
tylerjw, line 64: @griswaldbrooks If we start using these in a bunch of tests would it be normal to put them in a header that's included by our tests?

moveit_ros/planning/planning_scene_monitor/test/trajectory_monitor_tests.cpp
```diff
@@ -0,0 +1,111 @@
...
+  {
+    if (done())
+    {
+      return;
+    }
+    std::this_thread::sleep_for(1ms);
+  }
+}
+
+TEST(TrajectoryMonitorTests, test1)
```
tylerjw, line 79: Name the test something.  Maybe `SleepAtleastOnce`?

moveit_ros/planning/planning_scene_monitor/test/trajectory_monitor_tests.cpp
tylerjw: It will wait for up to 10s for callback_called to equal true.  The actual time this takes is unknown but should be <1ms.

tylerjw (approved):
Generally approve.  The only thing I saw was the const in the parameter list that @jliukkonen pointed out and the name of the test.  Thank you for doing this!

moveit_ros/planning/planning_scene_monitor/test/trajectory_monitor_tests.cpp
Abishalini: The mock middleware handles are moved when creating the `current_state_monitor` and `trajectory_monitor` too

tylerjw (approved):

tylerjw:
@griswaldbrooks will you give this a quick review.  I believe this is ready for merging.

tylerjw:
The failing rolling-testing CI job is a problem with the OSRF debians for rolling testing.

griswaldbrooks:
@tylerjw will do

moveit_ros/planning/planning_scene_monitor/test/trajectory_monitor_tests.cpp
griswaldbrooks: It would, assuming they're exactly the same and reused many times. Normally, my threshold for reuse is about 2. For these, my gut says about 3.

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/trajectory_monitor.h
```diff
@@ -41,11 +41,11 @@
 #include <moveit/robot_trajectory/robot_trajectory.h>
 #include <boost/thread.hpp>
 #include <memory>
+#include <functional>
 
 namespace planning_scene_monitor
 {
-using TrajectoryStateAddedCallback =
-    boost::function<void(const moveit::core::RobotStateConstPtr&, const rclcpp::Time&)>;
+using TrajectoryStateAddedCallback = std::function<void(const moveit::core::RobotStateConstPtr&, const rclcpp::Time&)>;
```
griswaldbrooks, line 48: Include-what-you-use would want
```
#include <rclcpp/rclcpp.hpp> // or #include <rclcpp/time.hpp>
```
and whatever declares `moveit::core::RobotStateConstPtr` but I can't seem to find that...

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/trajectory_monitor.h
```diff
@@ -54,9 +54,28 @@ MOVEIT_CLASS_FORWARD(TrajectoryMonitor);  // Defines TrajectoryMonitorPtr, Const
...
+
+    /**
+     * @brief      Add sleep using rate
+     */
+    virtual void sleep() = 0;
+  };
+
   /** @brief Constructor.
    */
   TrajectoryMonitor(const CurrentStateMonitorConstPtr& state_monitor, double sampling_frequency = 0.0);
```
griswaldbrooks, line 76: I could be convinced otherwise, but I would think that while we're here, we would use proper doxygen comments, especially on the new ctor.

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/trajectory_monitor.h
```diff
@@ -96,6 +115,7 @@ class TrajectoryMonitor
   void recordStates();
 
   CurrentStateMonitorConstPtr current_state_monitor_;
+  std::unique_ptr<MiddlewareHandle> middleware_handle_;
```
griswaldbrooks, line 118: @tylerjw where is moveit style on class member documentation.
https://moveit.ros.org/documentation/contributing/code/ implies http://wiki.ros.org/CppStyleGuide#Documentation to me
> All functions, methods, classes, class variables, enumerations, and constants should be documented.

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/trajectory_monitor.h
```diff
@@ -54,9 +54,28 @@ MOVEIT_CLASS_FORWARD(TrajectoryMonitor);  // Defines TrajectoryMonitorPtr, Const
...
+  class MiddlewareHandle
+  {
+  public:
+    /**
+     * @brief      Destroys the object.
+     */
+    virtual ~MiddlewareHandle() = default;
+
+    /**
+     * @brief      Add sleep using rate
```
griswaldbrooks, line 69: What is `rate`? I don't see that anywhere else in the public interface or documentation for `MiddlewareHandle`

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/trajectory_monitor_middleware_handle.hpp
```diff
@@ -0,0 +1,67 @@
...
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Abishalini Sivaraman */
+
+#pragma once
+
+#include <moveit/planning_scene_monitor/trajectory_monitor.h>
+#include <moveit/planning_scene_monitor/current_state_monitor.h>
```
griswaldbrooks, line 40:
```suggestion

```

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/trajectory_monitor_middleware_handle.hpp
```diff
@@ -0,0 +1,67 @@
...
+public:
+  /**
+   * @brief      Constructor for TrajectoryMonitor
+   *
+   * @param[in]  sampling_frequency  Used to create ROS2 Rate
+   */
+  TrajectoryMonitorMiddlewareHandle(double sampling_frequency);
+
+  /**
+   * @brief      sleep function using Rate
```
griswaldbrooks, line 60: ~sleep function using Rate~ `Sleeps for time specified by @p sampling_frequency`

moveit_ros/planning/planning_scene_monitor/src/trajectory_monitor.cpp
```diff
@@ -44,7 +45,16 @@ static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_ros.planning_sce
...
 planning_scene_monitor::TrajectoryMonitor::TrajectoryMonitor(const CurrentStateMonitorConstPtr& state_monitor,
                                                              double sampling_frequency)
+  : TrajectoryMonitor(state_monitor, std::make_unique<TrajectoryMonitorMiddlewareHandle>(sampling_frequency),
+                      sampling_frequency)
+{
+}
+
+planning_scene_monitor::TrajectoryMonitor::TrajectoryMonitor(
+    const CurrentStateMonitorConstPtr& state_monitor,
+    std::unique_ptr<TrajectoryMonitor::MiddlewareHandle> middleware_handle, double sampling_frequency)
```
griswaldbrooks, line 55: Judging from https://github.com/ros-planning/moveit2/pull/570/files#diff-f6042f47b79910249d738bcd543e8424d07fb3f45995bc4d567053d00e222e3aR69

It looks like the middleware handle will have to have an interface to change it's frequency since that's what was happening https://github.com/ros-planning/moveit2/pull/570/files#diff-f6042f47b79910249d738bcd543e8424d07fb3f45995bc4d567053d00e222e3aL111 on every loop

griswaldbrooks: Wait, wth???
https://github.com/ros-planning/moveit2/pull/570/files#diff-f6042f47b79910249d738bcd543e8424d07fb3f45995bc4d567053d00e222e3aR71
```
if (sampling_frequency != sampling_frequency_)
    return;  // silently return if nothing changes
```
Doesn't this mean you can never change the frequency? @tylerjw am I reading this wrong?

moveit_ros/planning/planning_scene_monitor/src/trajectory_monitor.cpp
```diff
@@ -35,6 +35,7 @@
 /* Author: Ioan Sucan */
 
 #include <moveit/planning_scene_monitor/trajectory_monitor.h>
+#include <moveit/planning_scene_monitor/trajectory_monitor_middleware_handle.hpp>
 #include <moveit/trajectory_processing/trajectory_tools.h>
 #include <rclcpp/rate.hpp>
```
griswaldbrooks, line 40: I don't think this is needed anymore

moveit_ros/planning/planning_scene_monitor/test/trajectory_monitor_tests.cpp
```diff
@@ -0,0 +1,111 @@
...
+struct MockCurrentStateMonitorMiddlewareHandle : public planning_scene_monitor::CurrentStateMonitor::MiddlewareHandle
+{
+  MOCK_METHOD(rclcpp::Time, now, (), (const, override));
+  MOCK_METHOD(void, createJointStateSubscription,
+              (const std::string& topic, planning_scene_monitor::JointStateUpdateCallback callback), (override));
+  MOCK_METHOD(void, resetJointStateSubscription, (), (override));
+  MOCK_METHOD(std::string, getJointStateTopicName, (), (const, override));
+};
+
+void waitFor(std::chrono::seconds timeout, std::function<bool()> done)
```
griswaldbrooks, line 66: Doxygen comment would be good.
Does passing `timeout` by `const&` make sense @tylerjw https://stackoverflow.com/questions/45740179/why-arent-stdchronoduration-instances-passed-by-value-in-the-standard-libra or are we going to assume we know it's a POD and there's no point or...

moveit_ros/planning/planning_scene_monitor/test/trajectory_monitor_tests.cpp
```diff
@@ -0,0 +1,111 @@
...
+  auto mock_trajectory_monitor_middleware_handle = std::make_unique<MockTrajectoryMonitorMiddlewareHandle>();
+  auto mock_current_state_monitor_middleware_handle = std::make_unique<MockCurrentStateMonitorMiddlewareHandle>();
+
+  // THEN we expect it to call sleep function
+  EXPECT_CALL(*mock_trajectory_monitor_middleware_handle, sleep).Times(::testing::AtLeast(1));
+
+  // GIVEN a TrajectoryMonitor is started
+  planning_scene_monitor::CurrentStateMonitorPtr current_state_monitor;
+
+  current_state_monitor = std::make_unique<planning_scene_monitor::CurrentStateMonitor>(
```
griswaldbrooks, line 90: Why is the declaration and instantiation on different lines? (this sort of practice opens you up to problems).

~Also, I'll bet you one keyboard that `CurrentStateMonitorPtr` is `std::shared_ptr<CurrentStateMonitor>` which would mean you're converting a unique to a shared??? Can you just use `auto` or `std::unique_ptr<CurrentStateMonitor>` because even if `CurrentStateMonitorPtr` is a `unique_ptr` the ambiguity is frustrating since the different pointers are used for different ownership lifetimes.~ 

https://stackoverflow.com/questions/37884728/does-c11-unique-ptr-and-shared-ptr-able-to-convert-to-each-others-type

Can you change this to `make_shared`?

moveit_ros/planning/planning_scene_monitor/test/trajectory_monitor_tests.cpp
```diff
@@ -0,0 +1,111 @@
...
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Abishalini Sivaraman */
+
```
griswaldbrooks, line 36:
```suggestion
#include <atomic>
```

moveit_ros/planning/planning_scene_monitor/test/trajectory_monitor_tests.cpp
```diff
@@ -0,0 +1,111 @@
...
+#include <memory>
+#include <chrono>
+#include <thread>
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+#include "moveit/planning_scene_monitor/trajectory_monitor.h"
+#include "moveit/planning_scene_monitor/current_state_monitor.h"
+#include "moveit/utils/robot_model_test_utils.h"
+#include "rclcpp/node_interfaces/node_clock_interface.hpp"
+#include "rclcpp/node_interfaces/node_topics_interface.hpp"
```
griswaldbrooks, line 46: Do you need any of these?

moveit_ros/planning/planning_scene_monitor/test/trajectory_monitor_tests.cpp
```diff
@@ -0,0 +1,111 @@
...
+  planning_scene_monitor::CurrentStateMonitorPtr current_state_monitor;
+
+  current_state_monitor = std::make_unique<planning_scene_monitor::CurrentStateMonitor>(
+      std::move(mock_current_state_monitor_middleware_handle), moveit::core::loadTestingRobotModel("panda"),
+      std::make_shared<tf2_ros::Buffer>(std::make_shared<rclcpp::Clock>()));
+
+  planning_scene_monitor::TrajectoryMonitor trajectory_monitor{ current_state_monitor,
+                                                                std::move(mock_trajectory_monitor_middleware_handle),
+                                                                10.0 };
+
```
griswaldbrooks, line 97: Can you add
```
// Set the trajectory monitor callback so we can block until at least one sample was taken.
```

In the future @tylerjw , I feel like this is one reuse away from being put into a fixture. Used only once is debatable. I do give pause though after reading https://abseil.io/tips/122

griswaldbrooks (changes requested):
@Abishalini @tylerjw

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/trajectory_monitor.h
Abishalini: I think it is `#include <moveit/robot_state/robot_state.h>`

griswaldbrooks: Could be, though I suspect it's that `MOVEIT` macro, which might be there?

moveit_ros/planning/planning_scene_monitor/src/trajectory_monitor.cpp
```diff
@@ -44,7 +45,16 @@ static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_ros.planning_sce
 
 planning_scene_monitor::TrajectoryMonitor::TrajectoryMonitor(const CurrentStateMonitorConstPtr& state_monitor,
                                                              double sampling_frequency)
+  : TrajectoryMonitor(state_monitor, std::make_unique<TrajectoryMonitorMiddlewareHandle>(sampling_frequency),
```
griswaldbrooks, line 48: Can we use uniform initialization?

moveit_ros/planning/planning_scene_monitor/src/trajectory_monitor.cpp
```diff
@@ -44,7 +45,16 @@ static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_ros.planning_sce
...
+                      sampling_frequency)
+{
+}
+
+planning_scene_monitor::TrajectoryMonitor::TrajectoryMonitor(
+    const CurrentStateMonitorConstPtr& state_monitor,
+    std::unique_ptr<TrajectoryMonitor::MiddlewareHandle> middleware_handle, double sampling_frequency)
   : current_state_monitor_(state_monitor)
+  , middleware_handle_(std::move(middleware_handle))
```
griswaldbrooks, line 57: Consider (though a separate PR also makes sense) converting to uniform initialization

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/trajectory_monitor.h
tylerjw: I would specifically disagree with the idea of documenting class variables.  I would prefer we focus on documenting the public interface.  I'm not sure what a useful comment on this line would be?

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/trajectory_monitor.h
tylerjw: Specify the unit (Hz).

moveit_ros/planning/planning_scene_monitor/src/trajectory_monitor.cpp
tylerjw: lol, that looks like a bug (based on the comment).

It should probably be `(sampling_frequency == sampling_frequency_)`.  I think this is prof that you will need to add an interface to change the frequency used.

moveit_ros/planning/planning_scene_monitor/src/trajectory_monitor.cpp
tylerjw:
```suggestion
  , middleware_handle_{std::move(middleware_handle)}
```

:arrow_up: that is all Griz means btw.

moveit_ros/planning/planning_scene_monitor/test/trajectory_monitor_tests.cpp
tylerjw: I think we assume these are POD and leave it as is.  I didn't know about that... the more you know :rainbow:

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/trajectory_monitor.h
griswaldbrooks: Consider
```
CurrentStateMonitorConstPtr current_state_monitor_;  //!< Samples joint states.
std::unique_ptr<MiddlewareHandle> middleware_handle_;  //!< Interface for communicating with ROS.
```
It may not be part of the public interface, but it is a public facing member.

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/trajectory_monitor.h
griswaldbrooks: No, my point is that in this interface, there is no discussion of rate, sampling frequency, or anything like that. Furthermore, `sleep` could be implemented without respect to a ROS rate, so a more abstract comment like
```
/**
 * @brief      Sleep the handle for some prescribed amount of time.
 */
 virtual void sleep() = 0;
```

moveit_ros/planning/planning_scene_monitor/src/trajectory_monitor.cpp
griswaldbrooks: Yes, but for all

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/trajectory_monitor.h
griswaldbrooks: Also, what do you mean by "focus on documenting the public interface"? Everything public facing should be documented imho

tylerjw: You are correct that these are not in an impl class.  I would still however focus on documenting the parts of a class that the user can access without inheriting or making it a friend class first (that is what I meant by public interface).

v4hn: MoveIt is not properly pimpl'ed (even in the few places where it is it's somewhat smelly), otherwise this would be a different matter. As MoveIt is generally extendable/hackable on any level, I'm no big fan of having pimpl/private declarations in the framework for most members anyway, but that's an independent discussion.

Either way, adding brief inline comments makes sense when they add additional information and adding them takes less resources than discussing in long threads whether focusing on other things makes more sense :smirk: 
I don't feel the proposed comments add relevant information, but then that's usually just maintainer bias.
 `[csm] //!< Samples robot states.` is somewhat more accurate, because CSM also updates from TF where required.

moveit_ros/planning/planning_scene_monitor/test/trajectory_monitor_tests.cpp
v4hn: @griswaldbrooks I love your recent in-depths-backed-up-by-references reviews! They make for great reading! :heart:

If @tylerjw feels it's unnecessary let's leave it as is. My main complain would be the interface itself, because condition variables/void futures should usually be preferred over polling arbitrary condition functions. :cat2: 
But in this scope I think it's actually bikeshedding. :shrug:

moveit_ros/planning/planning_scene_monitor/src/trajectory_monitor.cpp
v4hn: MoveIt does not have a fixed policy on uniform initialization at the moment. I know @rhaschke does not like it very much. I on the other hand am quite a big fan.
It's almost only used in the `pilz_industrial_planner` sources (which are currently `COLCON_IGNORE`'ed in moveit2 afaik)
and while I would love to see a clang-tidy PR for the whole code base :sunglasses:, let's keep the style consistent unless it actually produces cleaner code to use it (e.g. in-class initialization of floating points, etc.).

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/trajectory_monitor.h
griswaldbrooks: > even in the few places where it is it's somewhat smelly
:smirk: 

I probably wouldn't advocate for arbitrary use of pimpl (Especially since I generally advocate against pointers where dynamic memory isn't required. Additionally I think ROS developers over use pointers because that's what they were taught).
I know it has a purpose in maintaining a stable ABI.
I am a big fan of composition, though not simply using pimpl without the "p".

> Either way, adding brief inline comments makes sense when they add additional information

Mainly I'm just I'm just trying to clarify a point of policy/bike shedding, since the moveit style guide said/implied that it follows the ROS style guide. At previous companies, we documented every member and interface with no ill effects. One might even argue that if you can't come up with a logical and simple comment for the member, perhaps it shouldn't be there.
Regardless it seems that the style that member comments are optional. Perhaps I should update the style webpage @tylerjw? I wonder what else deviates.

> [csm] //!< Samples robot states. is somewhat more accurate, because CSM also updates from TF where required.

This was mainly a dig to @tylerjw from last week at our hackathon where we started this since every one was making fun of the name and didnt' **really** know what the class did until we read the entire implementation.

moveit_ros/planning/planning_scene_monitor/test/trajectory_monitor_tests.cpp
griswaldbrooks: > If @tylerjw feels it's unnecessary let's leave it as is

Agreed. Especially since this is test code. I feel like test code, while still requiring high quality, can skip on a few things. Not in this case but in others I might write two functions that did something **very** similar with the same name in two different namespaces in the same file.
In production code I would never do such a thing, but I think since the scope of test code tends towards being very local, the value of the implementation being nearby and specific, rather than having to understand different initialization options for example if preferred. Though I might be able to be convinced otherwise.

>  because condition variables/void futures 

I also prefer that but something about it in this scope didn't quite hit right. I think it had something to do with being in the same thread but that sounds funny. The atomic is doing something similar-ish. The function isn't arbitrary since it's the only interface the caller has to know if any samples were collected.

:bicyclist:

moveit_ros/planning/planning_scene_monitor/src/trajectory_monitor.cpp
griswaldbrooks: Arguably, uniform initialization actually solves a problem and is not a manner of style.
While C++ has multiple ways to initialize, the rules can be funny https://isocpp.org/wiki/faq/cpp11-language#uniform-init
And then of course the "most vexing parse" is always avoided.
Personally, I think within a class, the same initialization technique should be used for uniformity (unless of course you're initialing a `std::vector` or some such class which has specific initialization semantics for the parens or braces) but I'm not sure why updating to modern idioms shouldn't happen on a class-by-class basis for a codebase. If moveit originally used C-style arrays (such as amcl), and then was ported to C++, I'm not sure why one wouldn't start using `std::array` or `std::vector`.

moveit_ros/planning/planning_scene_monitor/test/trajectory_monitor_tests.cpp
griswaldbrooks: > They make for great reading! 

I'm trying to prevent myself from picking up habits that aren't based on particular guidance.
I've lived through a few codebases where things were done they way they are because they were done they way they were.

moveit_ros/planning/planning_scene_monitor/src/trajectory_monitor.cpp
tylerjw: > I'm not sure why updating to modern idioms shouldn't happen on a class-by-class basis for a codebase.
:+1: 

It would be really nice if we could clang-tidy this change to the whole codebase but doing it in each piece as we touch it also seems fine to me.  Leaving a piece of code a bit nicer because you were there seems like an ok thing to do.

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/trajectory_monitor.h
v4hn: > Additionally I think ROS developers over use pointers because that's what they were taught

Not just taught, but also forced by existing shared_ptr APIs. Fully agreed. :worried: 

> One might even argue that if you can't come up with a logical and simple comment for the member, perhaps it shouldn't be there.

One might just as well argue that if you can't come up with a comment that's more descriptive than the variable/class name, it might just be that the author did a good job naming them :cat2: 

> we documented every member and interface with no ill effects

Adding explanations always makes sense. Period. Adding comments for syntactic reasons, on the other hand, if they don't do more than repeat the identifier, amounts to a `NOLINT` flag for the line, so reviewers don't complain about missing comments. :)
It's hard to find the line between `if (exists) { // if it exists` and `CurrentStateMonitor csm; //!< Samples states`, especially when you're working with the code for years.

moveit_ros/planning/planning_scene_monitor/src/trajectory_monitor.cpp
v4hn: > Arguably, uniform initialization actually solves a problem and is not a manner of style. [...] the rules can be funny

Absolutely! It's just that the places you proposed it for are not among the funny rules. And especially for those regular cases opinions differ on what is nicer. Because the whole codebase does not use uniform initialization yet, it makes sense to discuss this separately (maybe add it as a point to the next maintainer meeting and file a separate issue).

--- offtopic ---
> actually solves a problem

<details><summary>Currently my favorite problem it *creates* is this</summary>

```
$ cat > foo.cpp <<EOF
#include <iostream>

struct A {
        A(const std::initializer_list<double>& x, bool critical= true){
                std::cout << critical << "\n";
        }
};

int main(){
   A a1({0,1,2}, false);
   A a2{ {0,1,2}, false };
   A a3({0}, false);
   A a4{ {0}, false };
        return 0;
}
EOF
make foo
./foo
0
0
0
1
```
</details>

tylerjw: Ah, the fun initializer_list constructor parameter.  It can both make the creating of objects more intuitive and create a trap for users.

tylerjw:
If you rebase this on main it should no longer fail CI on foxy (it will also no longer run a foxy test) as main is for galactic/rolling now.

moveit_ros/planning/planning_scene_monitor/src/trajectory_monitor.cpp
griswaldbrooks: Ah, I see https://godbolt.org/z/GYYqWjoxv
because it's casting the boolean to a double in the initializer list.

A different, but also frustrating example
https://godbolt.org/z/orz8oT4js

v4hn: > A different, but also frustrating example

Yeah, but at least that's the "classic" one everyone knows about (or at least should after reading Effective Modern C++) :)

griswaldbrooks (approved):

tylerjw:
I'm sorry this isn't working because I added another function to the middleware interface for current state monitor.  You should be able to fix this by just copying one line for the mock out of the current state monitor test.

tylerjw (approved):
Thank you for updating this.  I think we should merge this before it goes stale again.

PR #571, opened 2021-07-26
Ruckig trajectory_processing plugin

AndyZe:
This adds a Ruckig plugin that's similar to TOTG with these differences:

- Allows nonzero initial/final velocities and accelerations
- Allows jerk limits
- Does not compute timestamps like TOTG does. It's meant to run after TOTG

Occasionally  Ruckig fails to solve so I made [an issue](https://github.com/pantor/ruckig/issues/58) there. Let's not be too negative about this, though... Realistically it is already better than the IterativeParabolic and IterativeSpline plugins that MoveIt was using previously.

My vision is that Servo can start using this same plugin.

Test with `ros2 launch moveit_resources_panda_moveit_config demo.launch.py`. It requires [this branch](https://github.com/AndyZe/moveit_resources/tree/andyz/change_time_parameterization) of moveit_resources to change one config file.

https://user-images.githubusercontent.com/11284393/127777777-98336e8c-523f-4216-b74d-b0e25b2b26fd.mp4


moveit_core/trajectory_processing/include/moveit/trajectory_processing/ruckig_traj_smoothing.h
```diff
@@ -0,0 +1,53 @@
...
+ * All rights reserved.
+ *
+ * Author: Tobias Kunz <tobias@gatech.edu>
+ * Date: 05/2012
+ *
+ * Humanoid Robotics Lab      Georgia Institute of Technology
+ * Director: Mike Stilman     http://www.golems.org
+ *
+ * Algorithm details and publications:
+ * http://www.golems.org/node/1570
```
gavanderhoorn, line 12: This seems strange.

AndyZe:
I'd say this is at the point of being useful for some people. It's not perfect, but it is optional. Taking it out of WIP status.

moveit_core/trajectory_processing/include/moveit/trajectory_processing/ruckig_traj_smoothing.h
davetcoleman: How so? I took classes with Mike Stilman back at Georgia Tech and remember that website!

moveit2.repos
```diff
@@ -23,6 +23,10 @@ repositories:
     type: git
     url: https://github.com/ros-controls/ros2_controllers
     version: master
+  ruckig:
+    type: git
+    url: https://github.com/pantor/ruckig.git
+    version: master
```
davetcoleman, line 29: Is this project stable enough to be in the main moveit2.repos file?

moveit_core/package.xml
```diff
@@ -53,6 +53,7 @@
   <depend>tf2_geometry_msgs</depend>
   <depend>trajectory_msgs</depend>
   <depend>visualization_msgs</depend>
+  <depend>ruckig</depend>
```
davetcoleman, line 56: I had assumed this would be an optional plugin, but it appears it will be a required dependency. Given the experimental state of the results you presented in this PR, is there a way to reduce the impact of this?

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -0,0 +1,169 @@
...
+  ruckig_output.new_position = ruckig_input.current_position;
+  ruckig_output.new_velocity = ruckig_input.current_velocity;
+  ruckig_output.new_acceleration = ruckig_input.current_acceleration;
+
+  // Kinematic limits (vel/accel/jerk)
+  const std::vector<std::string>& vars = group->getVariableNames();
+  const moveit::core::RobotModel& rmodel = group->getParentModel();
+  for (size_t i = 0; i < num_dof; ++i)
+  {
+    // TODO(andyz): read this from the joint group if/when jerk becomes supported
```
davetcoleman, line 108: can you clarify in the comment... when jerk becomes supported by the moveit trajectory_processing API? or by something that ruckig changes?

moveit_core/package.xml
davetcoleman: Can this be added instead to [planning_request_adapter_plugins](https://github.com/ros-planning/moveit2/tree/main/moveit_ros/planning/planning_request_adapter_plugins/src)?

moveit_core/trajectory_processing/include/moveit/trajectory_processing/ruckig_traj_smoothing.h
AndyZe: I think he was pointing out we copied the license from TOTG but forgot to change it...

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -984,7 +984,7 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
   }
 
   // Now actually call the algorithm
-  Trajectory parameterized(Path(points, path_tolerance_), max_velocity, max_acceleration, 0.001);
+  Trajectory parameterized(Path(points, path_tolerance_), max_velocity, max_acceleration, DEFAULT_TIMESTEP);
```
davetcoleman, line 987: nice cleanup work!

moveit_ros/planning/planning_request_adapters_plugin_description.xml
```diff
@@ -38,12 +38,19 @@
 
   <class name="default_planner_request_adapters/AddIterativeSplineParameterization" type="default_planner_request_adapters::AddIterativeSplineParameterization" base_class_type="planning_request_adapter::PlanningRequestAdapter">
     <description>
+      A deprecated time parameterization algorithm.
```
davetcoleman, line 41: Off topic and not related to this PR, but can we make a moveit2 issue to remove this algorithm from moveit2?

davetcoleman (commented):
This is an exciting work, but I think is too experimental to be merged in here. I chatted with @tylerjw and he's going to propose a better location for it.

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
AndyZe: We need to get jerk added and parsed from joint_limits.yaml

moveit2.repos
AndyZe: We can use a fixed hash, as we did with ros2_control until not long ago.

moveit_ros/planning/planning_request_adapters_plugin_description.xml
AndyZe: Well, are you sure it should be removed? I googled around and checked out MoveIt1 issues. Did not see any issue reports.

AndyZe: I'm unclear if it's IterativeSpline or IterativeParabolic that has major issues. Or maybe both/neither ...?

AndyZe: Per this issue from the UR repo, both have issues and probably should be deleted.

https://github.com/ros-planning/moveit/issues/1711

AndyZe: Made this issue:  https://github.com/ros-planning/moveit2/issues/573

AndyZe:
@davetcoleman I hope I didn't give the wrong impression by showing a video of some trajectory failure cases. That's in my nature, I guess. I'm looking at this mainly to be used with Servo, as a way to start standardizing Servo with the trajectory pipeline.

moveit2.repos
tylerjw: My main concern is that it hasn't been released yet and with each new version of ros we'll need to get the author's of this to release before we can release if moveit depends on it.  If we put this plugin in a separate repo it can still be included in tutorials and linked to by servo examples but it won't be a build/release dependency.

AndyZe: OK. We can move this somewhere else, I guess. Then we also need to replace this plugin with a different plugin based on the lowpass filter that's currently in Servo so it has a default plugin to use.

Basically, what is left to do is...  (the list is getting long)
- Add functionality for parsing jerk and storing it (in the Joint Model, I think)
- Move this plugin somewhere else
- Delete the lowpass filter class from Servo and make it a plugin
- Refactor Servo to make it compatible with these plugins

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -0,0 +1,169 @@
...
+  if (!group)
+  {
+    RCLCPP_ERROR(LOGGER, "It looks like the planner did not set the group the plan was computed for");
+    return false;
+  }
+
+  size_t num_waypoints = trajectory.getWayPointCount();
+  if (num_waypoints < 2)
+  {
+    RCLCPP_WARN_STREAM(LOGGER, "Trajectory does not have enough points to smooth with Ruckig");
```
henningkayser, line 66: This aborts the `applySmoothing()` function, so this should be an error like above

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -0,0 +1,169 @@
...
+    return false;
+  }
+
+  const size_t num_dof = group->getVariableCount();
```
henningkayser, line 70: This should probably use `group->getActiveJointModels().size()`. Group variables include mimic joints, floating joints etc...

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+  // This lib does not actually work properly when angles wrap around, so we need to unwind the path first
+  trajectory.unwind();
+
+  // Instantiate the smoother
+  double timestep = trajectory.getAverageSegmentDuration();
+  ruckig::Ruckig<0> ruckig{ num_dof, timestep };
```
henningkayser, line 77: what's with the `0`

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -0,0 +1,169 @@
...
+  ruckig::InputParameter<0> ruckig_input{ num_dof };
+  ruckig::OutputParameter<0> ruckig_output{ num_dof };
+
+  // Initialize the smoother
+  //  ruckig_input.synchronization = ruckig::Synchronization::Time;
+  //  ruckig_input.interface = ruckig::Interface::Velocity;
+  const std::vector<int>& idx = group->getVariableIndexList();
```
henningkayser, line 84: same here, instead of iterating all variables, better use `group->getActiveJointModelNames()` and iterate those. Otherwise this could have odd effects with robots that have mimic joints (e.g. controller rejecting trajectories)

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
henningkayser: > We need to get jerk added and parsed from joint_limits.yaml

If we get to really supporting it, this sounds like a great idea!

henningkayser (commented):
This looks really good to me so far. I think you can ignore my concerns about the non-active group variables for now, the other plugins do just the same. I'd still like to verify at some point if we should take care of this. Regarding the ruckig dependency: IMO there is nothing wrong with maintaining plugins like these in a separate repo for a while (like STOMP for example). Honestly, I would even suggest moving all the plugins out of `moveit_core` since they are no build dependency for downstream packages.

On servo: Do you have any thoughts how planner adapters could be used for smoothing servo commands? I guess it would require a different form of plugin structure. To me it would not be such a bad idea to use the same (or related) plugin structure for that, possibly by providing an alternative interface. That way we could keep related algorithms in one place.

AndyZe:
I figured out some extra logic to eliminate the backward motion that sometimes occurs. Looking much better now. I'll now move on to address the rest of the code reviews here.


https://user-images.githubusercontent.com/11284393/127777714-c1b231c8-fa03-4648-acb9-50da3ff9872e.mp4

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
AndyZe: done, good catch

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
AndyZe: Well, this bit was directly copied from TOTG. Open an issue to figure it out later?

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
AndyZe: The template argument was originally supposed to be num_dof, but that caused some issues if (for example) num_dof wasn't known at compile time.

So the Ruckig author made [this PR](https://github.com/pantor/ruckig/pull/47) to allow dynamic num_dof. As far as I know, this is correct usage even if it looks a little funny.

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
AndyZe: https://github.com/ros-planning/moveit2/issues/586

AndyZe:
A joint angle plot while running Ruckig.

with_ruckig

pantor:
Hi, author of Ruckig here. If there is anything I can do to bring this closer to merging, I'm happy to help!

The recent video with the fix for the backward motion looks great! Regarding the goal pose: Isn't it possible to set the goal pose as the last target pose for Ruckig (when updating the input parameters the final time), so that the goal pose is always reached exactly?

With the recent commits of Ruckig, the numeric stability in particular for high limits (tested with `|x| < 1e6`) should have improved significantly. If you get any more errors, please file an issue with the input parameters so I can check what's going on exactly. By the way, I've added a `ruckig::DynamicDOFs` constant to replace the `<0>` for dynamic DoFs to improve the readability of the code.

AndyZe:
Hey @pantor, how would you feel about creating a ROS2 release? That will make it a lot easier for us to use your library.

Without a ROS release, we will probably move this plugin to a different repository because we don't want to (essentially) copy/paste your library into moveit.

There are instructions [here](https://docs.ros.org/en/galactic/Guides/Releasing-a-ROS-2-package-with-bloom.html). I would probably target Rolling Ridley (Ubuntu 20.04). Does that sound right, @tylerjw?

tylerjw:
If your library is released for both rolling and galactic we can then depend on it in our main branch.  If you also release for foxy we can backport this to foxy.

pantor:
I've now created a ROS2 release for Ruckig (with PRs for [rolling](https://github.com/ros/rosdistro/pull/30422) and [galactic](https://github.com/ros/rosdistro/pull/30421)).

gavanderhoorn:
> I've now created a ROS2 release for Ruckig (with PRs for [rolling](https://github.com/ros/rosdistro/pull/30422) and [galactic](https://github.com/ros/rosdistro/pull/30421)). I'll add foxy later.

One for Melodic & Noetic would then also be appreciated.

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -0,0 +1,226 @@
...
+#include <moveit/trajectory_processing/ruckig_traj_smoothing.h>
+#include <rclcpp/rclcpp.hpp>
+#include <vector>
+
+namespace trajectory_processing
+{
+static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_trajectory_processing.ruckig_traj_smoothing");
+constexpr double DEFAULT_MAX_VELOCITY = 5;       // rad/s
+constexpr double DEFAULT_MAX_ACCELERATION = 10;  // rad/s^2
+constexpr double DEFAULT_MAX_JERK = 20;          // rad/s^3
```
tylerjw, line 49: Thank you for adding comments with units, that is really helpful.  Consider putting this in an anonymous namespace.

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -0,0 +1,226 @@
...
+  //  ruckig_input.synchronization = ruckig::Synchronization::Time;
+  //  ruckig_input.interface = ruckig::Interface::Velocity;
+  const std::vector<int>& idx = group->getVariableIndexList();
+  std::vector<double> current_positions_vector(num_dof);
+  std::vector<double> current_velocities_vector(num_dof);
+  std::vector<double> current_accelerations_vector(num_dof);
+  moveit::core::RobotStatePtr waypoint = trajectory.getFirstWayPointPtr();
+  for (size_t i = 0; i < num_dof; ++i)
+  {
+    current_positions_vector.at(i) = waypoint->getVariablePosition(idx[i]);
```
tylerjw, line 90: use `at()` for accessing into idx vector

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -0,0 +1,226 @@
...
+  ruckig_output.new_velocity = ruckig_input.current_velocity;
+  ruckig_output.new_acceleration = ruckig_input.current_acceleration;
+
+  // Kinematic limits (vel/accel/jerk)
+  const std::vector<std::string>& vars = group->getVariableNames();
+  const moveit::core::RobotModel& rmodel = group->getParentModel();
+  for (size_t i = 0; i < num_dof; ++i)
+  {
+    // TODO(andyz): read this from the joint group if/when jerk limits are added to the JointModel
+    ruckig_input.max_jerk[i] = DEFAULT_MAX_JERK;
```
tylerjw, line 108: use `at()`

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -0,0 +1,226 @@
...
+      next_waypoint->setVariablePosition(idx[joint], ruckig_output.new_position[joint]);
+      next_waypoint->setVariableVelocity(idx[joint], ruckig_output.new_velocity[joint]);
+      next_waypoint->setVariableAcceleration(idx[joint], ruckig_output.new_acceleration[joint]);
+    }
+  }
+
+  return true;
+}
+
+double RuckigSmoothing::getTargetVelocityMagnitude(const ruckig::InputParameter<0>& ruckig_input, const size_t num_dof)
```
tylerjw, line 181: size_t parameter shouldn't be const

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -0,0 +1,226 @@
...
+      return true;
+      break;
+    }
+  }
+  return false;
+}
+
+void RuckigSmoothing::getNextCurrentTargetStates(ruckig::InputParameter<0>& ruckig_input,
+                                                 ruckig::OutputParameter<0>& ruckig_output,
+                                                 const moveit::core::RobotStatePtr next_waypoint, const size_t num_dof,
```
tylerjw, line 210: Const shared pointers can force copies because to actually share the pointer you need to be able to update the reference count.  Use either a const& or a non-const shared ptr.  Also, get rid of the const on the size_t

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -49,7 +49,8 @@ namespace trajectory_processing
 static const rclcpp::Logger LOGGER =
     rclcpp::get_logger("moveit_trajectory_processing.time_optimal_trajectory_generation");
 
-constexpr double EPS = 0.000001;
+constexpr double DEFAULT_TIMESTEP = 0.001;
+constexpr double EPS = 1e-6;
```
tylerjw, line 53: rebase on the PR we merged so this change isn't here.

moveit_ros/planning/planning_request_adapter_plugins/src/add_ruckig_traj_smoothing.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+  bool adaptAndPlan(const PlannerFn& planner, const planning_scene::PlanningSceneConstPtr& planning_scene,
+                    const planning_interface::MotionPlanRequest& req, planning_interface::MotionPlanResponse& res,
+                    std::vector<std::size_t>& /*added_path_index*/) const override
+  {
+    bool result = planner(planning_scene, req, res);
+    if (result && res.trajectory_)
+    {
+      RCLCPP_DEBUG(LOGGER, " Running '%s'", getDescription().c_str());
+      RuckigSmoothing smoother;
```
tylerjw, line 71: Is there a reason this is created with each service call and not part of the class and re-used with each service call?

tylerjw (commented):
Please add some basic unit tests to make sure your use of arrays in index-based for loops doesn't use memory past the end of the vectors under normal operation.  You can do this by changing you array access to using `at` and then writing a test that will exercise the code in a normal way.  That way the test will pass assuming it doesn't throw an exception.

pantor:
Ruckig is now released as a [ROS package](https://index.ros.org/search/?term=ruckig)  for galactic, foxy, rolling, noetic, and melodic.

tylerjw:
I recognize that some of the things I asked for changes to might require changes to the plugin interface.  If that is the case please point that out and I'll make a separate PR to make that change (this being targeted towards rolling/galactic we should be able to make interface changes where they make sense).

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
AndyZe: You could have just said, replace `[]` with `.at()`   ;)

tylerjw: sorry... there are a ton of comments in here.  There is a bunch of array/vector access with operator[] and generally I'm trying to get moveit to use `at` because it'll make it much easier to see what went wrong if the indexes get confused at some point.  Technically operator[] is more performant (slightly) and in some really hot sections of the code we should use it after carefully making sure the indexes can never do weird things.  You probably already know this.

moveit_ros/planning/planning_request_adapter_plugins/src/add_ruckig_traj_smoothing.cpp
AndyZe: There is a reason. The planning_request_adapter this inherits from takes a const "Planner."

`bool adaptAndPlan(const PlannerFn& planner, const planning_scene::PlanningSceneConstPtr& planning_scene`

The change you're asking for would require non-const

AndyZe: Prob would be good to change that and update TOTG in the same way...

tylerjw: I see that now.  Thank you for clarifying.  It was the only thing I saw that seemed a bit odd.

moveit2.repos
tylerjw: This is needed for now for CI to pass but after the next sync we can remove it as it was bloomed: https://github.com/ros/rosdistro/pulls?q=is%3Apr+Ruckig+is%3Aclosed

AndyZe:
2 more issues I'm aware of:

- write a test
- spewed warnings about updated collision body transforms

I'll try to fix the warnings this weekend

dirty_warning

tylerjw (approved):
Assuming this passes CI I'll merge this.  Then we should backport it to Foxy, Noetic, and Melodic if possible.

tylerjw:
I just saw your message about the warnings.  I'll wait on merging this until you get a chance to clean that up.  Thank you.

moveit_ros/planning/planning_request_adapter_plugins/src/add_ruckig_traj_smoothing.cpp
AndyZe: I'll make an issue so we don't forget. It's a good idea

https://github.com/ros-planning/moveit2/issues/610

moveit_core/trajectory_processing/CMakeLists.txt
```diff
@@ -1,8 +1,11 @@
 set(MOVEIT_LIB_NAME moveit_trajectory_processing)
 
+find_package(ruckig REQUIRED)
```
henningkayser, line 3: This is a `moveit_core` dependency now, please move into the top-level CMakeLists.txt along with the other `find_package()` calls. (I see you already added ruckig to `THIS_PACKAGE_INCLUDE_DEPENDS`, also).

moveit_core/trajectory_processing/CMakeLists.txt
```diff
@@ -20,6 +23,7 @@ ament_target_dependencies(${MOVEIT_LIB_NAME}
 target_link_libraries(${MOVEIT_LIB_NAME}
   moveit_robot_state
   moveit_robot_trajectory
+  ruckig::ruckig
```
henningkayser, line 26: nit, doesn't `ament_target_dependencies(... ruckig)` work? `target_link_libraries()` doesn't take care about include directories

henningkayser (commented):
@AndyZe really almost there! Please address the minor comments on the CMakeLists, then you get a +1 from me

moveit_core/package.xml
henningkayser: All trajectory processing algorithms are implemented in moveit_core, only the plugin wrappers are provided by the referenced package. I think it's ok to add this dependency here, especially since it has been released in the meantime.

moveit_core/trajectory_processing/include/moveit/trajectory_processing/ruckig_traj_smoothing.h
```diff
@@ -0,0 +1,72 @@
...
+#include <list>
+#include <moveit/robot_trajectory/robot_trajectory.h>
+#include <ruckig/ruckig.hpp>
+
+namespace trajectory_processing
+{
+class RuckigSmoothing
+{
+public:
+  bool applySmoothing(robot_trajectory::RobotTrajectory& trajectory, const double max_velocity_scaling_factor = 1.0,
```
henningkayser, line 48: I don't see any members in this class. `applySmoothing()` could be const or even static, both would allow you to make `RuckigSmoothing` a const member of the `PlanningRequestAdapter` addressing [this](https://github.com/ros-planning/moveit2/pull/571/files#r686447993 ).

AndyZe:
@personal-fork-robot  it's very easy. You only need to add the plugin to demo.launch.py (or whatever launch file you're using.) See the example here.  https://github.com/AndyZe/moveit_resources/commit/40dabcab15b70aeeec89a5bede7a69ff2bfe95e7#diff-6393b58b4a7a0e7352f796c30f408dac2b85d7cd4b4293aaedc6c479754c8973R71

AndyZe:
@pantor the Rolling debian of Ruckig is now available. Thanks!

Unfortunately I get a compilation error now from this bit of code:

```
  ruckig::Ruckig<0> ruckig{ num_dof, timestep };
  ruckig::InputParameter<0> ruckig_input{ num_dof };
  ruckig::OutputParameter<0> ruckig_output{ num_dof };
```

> /home/andy/ws_ros2/src/moveit2/moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp: In member function ‘bool trajectory_processing::RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory&, double, double)’:
> /home/andy/ws_ros2/src/moveit2/moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp:79:47: error: no matching function for call to ‘ruckig::Ruckig<0>::Ruckig(<brace-enclosed initializer list>)’
>    79 |   ruckig::Ruckig<0> ruckig{ num_dof, timestep };

I have the Ruckig debian installed, not building Ruckig from source. I wonder if [this commit ](https://github.com/pantor/ruckig/pull/47) didn't get included in the ROS release?

AndyZe:
Maybe the Ruckig library isn't installed correctly by the debian or I screwed up the CMakeLists. Will circle back to look more closely in a few hours.

AndyZe:
OK, the problem is that the ROS debian doesn't install anything and it doesn't even use colcon at all (https://github.com/ros/rosdistro/blob/be4e7376eee40fb6903492371ba00ec4175c1e06/rolling/distribution.yaml#L3622). So there's not even a ruckig folder installed in /opt/ros/lib

@tylerjw @henningkayser @JafarAbdi do you have any example of a pure CMake package and how that should be used with ROS2?

tylerjw:
@jspricke do you know what they should do?  I will look into this on Monday when I'm back at my computer too.

AndyZe:
It's strange that it finds the library --> no linker error. The result of this is attached:

`sudo apt install ros-rolling-rucking`
`dpkg -L ros-rolling-ruckig`

Makes me think it must be an older version of the library that got released. Prior to 7d2322f

[dpkg_ruckig.txt](https://github.com/ros-planning/moveit2/files/7026392/dpkg_ruckig.txt)

gavanderhoorn:
> OK, the problem is that the ROS debian doesn't install anything and it doesn't even use colcon at all

packages don't "use Colcon". That's the build tool.

From the `dpkg -L` output it looks like everything is installed correctly. As I would expect for a pure-CMake package.

pantor:
Hi @AndyZe, it seems that tag 0.3.3 was released (I assumed it would release the master branch), so the feature of dynamic DoFs is indeed not available. I'll release a new version tomorrow.

pantor:
Version 0.4.0 that should fix this issue is now released.

AndyZe:
OK, the correct Ruckig Debian has finally been released. I tested it locally and things look fine. If it passes CI, let's get this merged

moveit_core/trajectory_processing/CMakeLists.txt
AndyZe: I have tested that. Indeed, it does not work. I think it's because ruckig is installed like a system dependency, not like a normal ROS2 pkg.

moveit_core/trajectory_processing/include/moveit/trajectory_processing/ruckig_traj_smoothing.h
AndyZe: Good idea. Done

AndyZe:
I've been testing this for a couple hours and tweaking the kinematics calculations. I find that it succeeds practically 100% of the time when (accel scaling = 1.00, vel scaling = 0.05). It often fails if (accel scaling = 1.00, vel scaling = 1.00). Which makes sense, because TOTG runs first and it isn't aware of jerk limits. So TOTG is making requests that just aren't possible when jerk limits are taken into account.

I think we can improve this (later) by stretching out the timesteps from TOTG and re-running it, but it's not necessarily something I want to dive into in this PR.

https://user-images.githubusercontent.com/11284393/133514024-1951301b-a207-478b-8e75-5ba3368361c3.mp4

tylerjw (approved):
I looked through your latest changes and this looks good.

gavanderhoorn:
> I think we can improve this for high speeds (later) by stretching out the timesteps from TOTG and re-running it

Starts to sound a bit like the IPTP works, IIRC?

PR #575, opened 2021-07-29
SortIncludes: true

tylerjw:
### Description

I know, I know... don't change the `.clang-format` file.  I am tired of manually formatting includes and would really like to stop doing it.  Please, please tell me why this is a bad idea.  I'm happy to make this PR into every repo we have.

:warning: :warning: :warning: 

I honestly don't care what the rules are, just that I can stop formatting includes manually and that there is a consistent style.  Those that care are welcome to fight here about what is right but I want this decided to be something other than 1000 different manually created styles in our code.  :bell: :bell: :bell: now Fight!


moveit_core/collision_detection/include/moveit/collision_detection/world.h
```diff
@@ -36,13 +36,15 @@
...
-#include <moveit/macros/class_forward.h>
+#include <map>
 #include <string>
 #include <vector>
-#include <map>
-#include <boost/function.hpp>
+
 #include <Eigen/Geometry>
+#include <boost/function.hpp>
 #include <eigen_stl_containers/eigen_stl_vector_container.h>
```
v4hn, line 45: I'm not really happy about sorting third-party includes by alphabet (at least ROS packages should in theory go after system libs), but that's annoying to achieve. I guess separate including groups for our moveit-specific libs (geometric_shapes, ...), as well as Eigen and boost (possibly also fcl and OMPL) might suffice for 95% of the cases?

v4hn (commented):
In general I love it (as long as you backport it as well).

My harsh criticism is in the inline comment. ;)

vatanaksoytezer (approved):
Agreeing with @v4hn's comments on seperating Eigen, boost, fcl and OMPL I really love this change, and feel very useful.

jliukkonen (commented):
I like this change a lot too. Vatan's proposal sounded good. Major libraries and their includes can be their own separate blocks.

moveit_core/collision_detection/include/moveit/collision_detection/world.h
tylerjw: What do you think of my latest revision, I tried to accomplish what you asked as best I could.  I can't tell what is/is-not a ros library (other than maybe rclcpp) so instead, I just made a block for our external ones and another block for the libraries you listed.

tylerjw:
I had to reverse the order (moveit -> system) to get some of the regex to work correctly.  This is because it matches from most specific to least.  The current ordering I have now is:

1. moveit
2. other ros-planning repos
3. boost/Eigen/fcl/ompl
4. libraries with a / in the name (ros, other ros repos, etc)
5. system

tylerjw:
Of course this doesn't work because the order of includes matters :/

If anyone has time to track down any of these issues it would be really nice.  Ideally, if everything just included what it used this wouldn't be an issue.

jliukkonen (approved):
Looks good to me. I'm sure that LinkModel/JointModel circular dependency can be fixed with forward declarations.

By the way, does someone know why there is a `.inc` file there?

moveit_core/collision_detection/include/moveit/collision_detection/collision_tools.h
```diff
@@ -36,16 +36,17 @@
...
 #include "rcl/time.h"
 #include "rclcpp/clock.hpp"
+#include "rclcpp/rclcpp.hpp"
 #include "rclcpp/time.hpp"
 
+#include <moveit/collision_detection/collision_common.h>
+#include <moveit_msgs/msg/contact_information.hpp>
+#include <moveit_msgs/msg/cost_source.hpp>
+
+#include <visualization_msgs/msg/marker_array.hpp>
```
v4hn, line 48: Would it make sense to keep *all* message includes in a separate block?

moveit_core/collision_detection_fcl/test/test_fcl_env.cpp
```diff
@@ -34,19 +34,17 @@
...
 #include <moveit/robot_model/robot_model.h>
 #include <moveit/robot_state/robot_state.h>
 #include <moveit/utils/robot_model_test_utils.h>
 
-#include <moveit/collision_detection_fcl/collision_common.h>
-#include <moveit/collision_detection_fcl/collision_env_fcl.h>
+#include <geometric_shapes/shape_operations.h>
 
+#include <gtest/gtest.h>
 #include <urdf_parser/urdf_parser.h>
```
v4hn, line 47: would it make sense to keep `gtest` separate from everything else as well?

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -34,18 +34,22 @@
...
 #include <moveit/collision_detection/collision_tools.h>
-#include <moveit/trajectory_processing/trajectory_tools.h>
-#include <moveit/robot_state/conversions.h>
+#include <moveit/collision_detection_fcl/collision_detector_allocator_fcl.h>
 #include <moveit/exceptions/exceptions.h>
+#include <moveit/planning_scene/planning_scene.h>
 #include <moveit/robot_state/attached_body.h>
+#include <moveit/robot_state/conversions.h>
+#include <moveit/trajectory_processing/trajectory_tools.h>
 #include <moveit/utils/message_checks.h>
```
v4hn, line 44: The downside of alphabetical ordering is of course that our own hierarchies in MoveIt are not represented in the includes. I'm ok with that, but it also doesn't make me happy :eyes:

v4hn (commented):
> Of course this doesn't work because the order of includes matters :/

You found it, you fix it. ;)

> I had to reverse the order (moveit -> system) to get some of the regex to work correctly. This is because it matches from most specific to least. The current ordering I have now is:

I'm not sure I understand why you had to do that. Isn't the order of the include groups independent of the priorities of the regexes? Either way, I'm not sure which way around I actually prefer...

@jliukkonen wrote:
> By the way, does someone know why there is a .inc file there?

Probably because `.hpp` was not a standard term when this part was written. :roll_eyes: 
Could you fix this for both repositories please? The file was probably ignored by formatting filters as well...

tylerjw:
> You found it, you fix it. ;)

That's fair... I would say the same thing.

I was up really late last night trying to fix this.  I eventually tried to use include-what-you-use from google: https://github.com/include-what-you-use/include-what-you-use

Annoyingly it does dumb things with a bunch of the ros includes (messages and libraries inside ros).  At some point I might try making a fork of it or see if there is something I could configure to make it work much better.

The short story is I got past the `.inc` file fairly quickly doing what you suggest but there are still tons of other issues and to make it more annoying it is different on foxy and rolling.

tylerjw:
Replaced by https://github.com/ros-planning/moveit2/pull/583

PR #576, opened 2021-07-29
Temporarily disable rolling-testing in CI

tylerjw:
### Description

`rolling-testing` does not seem to be fixed yet (one week after it was broken) even with all the work that @jspricke did helping osrf figure out what is wrong with it.  I propose we disable testing on it until the core packages are more stable on it.  The point of rolling-testing is to get API breaking changes sooner so we can adapt to them, instead what we've gotten is buildfarm breaking changes and there is no way for us to adapt other than stop testing on `rolling-testing`.

https://github.com/osrf/infrastructure/issues/5


JafarAbdi (approved):

vatanaksoytezer (approved):
Can we open an issue to enable this back?

PR #578, opened 2021-07-30
Fix Foxy CI

vatanaksoytezer:
Foxy CI is broken because we've branched off for geometric_shapes. We need to use the correct branch of geometric_shapes for foxy to prevent CI failures.


vatanaksoytezer:
CI will be fixed by #565 instead.

PR #579, opened 2021-07-30
Upgrade to C++17

tylerjw:
### Description

Galactic and newer use C++17

https://www.ros.org/reps/rep-2000.html#galactic-geochelone-may-2021-november-2022


henningkayser (approved):

v4hn:
All for C++17 support, but it *does* seem weird to upgrade the compile flags when there is not a single line requiring it yet...

PR #580, opened 2021-07-30
Remove Rolling and Galactic CI testing in foxy branch

tylerjw:
### Description

Foxy branch now only supports ROS 2 Foxy, no need to test Galactic and Rolling for CI on foxy branch.


henningkayser (commented):
I think the build badges on the README still need to get updated.

tylerjw:
> I think the build badges on the README still need to get updated.

Pushed a commit with this change (here and in the other PR)

henningkayser (approved):

PR #581, opened 2021-07-30
Add missing exec dependencies to demo packages

henningkayser:
Backport to:
- [x] Foxy


henningkayser:
Blocked by https://github.com/ros-controls/ros2_control/issues/421. @tylerjw what do you think of enabling repos for this? I'm also wondering about the servo tests that we are planning to re-enable, those should also depend on ros2_control at this point.

tylerjw (approved):

PR #584, opened 2021-07-31
Minor documentation and cleanup of TOTG plugin

AndyZe:
Pulling some changes out of #571 so they can be merged separately.

- Document some limitations of TOTG in the plugin description.
- Minor code cleanup.
- Mark iterative spline and iterative parabola plugins as deprecated since they have some known issues (https://github.com/ros-planning/moveit/issues/1711).

### DIFF
--- a/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
+++ b/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
@@ -49,7 +49,8 @@ namespace trajectory_processing
 static const rclcpp::Logger LOGGER =
     rclcpp::get_logger("moveit_trajectory_processing.time_optimal_trajectory_generation");
 
-constexpr double EPS = 0.000001;
+constexpr double DEFAULT_TIMESTEP = 0.001;
+constexpr double EPS = 1e-6;
 class LinearPathSegment : public PathSegment
 {
 public:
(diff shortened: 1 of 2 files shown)


moveit_ros/planning/planning_request_adapters_plugin_description.xml
```diff
@@ -33,17 +33,19 @@
 
   <class name="default_planner_request_adapters/AddTimeParameterization" type="default_planner_request_adapters::AddTimeParameterization" base_class_type="planning_request_adapter::PlanningRequestAdapter">
     <description>
+      A deprecated time parameterization plugin.
```
sjahr, line 36: So does this mean TOTG is the only algorithm recommended to use? Maybe this information should be added to the tutorials as well.

AndyZe: Yep, I agree.

henningkayser: I don't think anyone will ever notice this description. Could you add some warnings if the old plugins are being used? I'm also not convinced the old plugins should be deprecated, switching the default to TOTG and hinting on the shortcomings seems sufficient to me.

AndyZe: In the interest of getting this merged, I'll just revert the changes to `TimeParameterization` and `IterativeSplineParameterization`

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -49,7 +49,8 @@ namespace trajectory_processing
 static const rclcpp::Logger LOGGER =
     rclcpp::get_logger("moveit_trajectory_processing.time_optimal_trajectory_generation");
 
-constexpr double EPS = 0.000001;
+constexpr double DEFAULT_TIMESTEP = 0.001;
```
tylerjw, line 52: [nit] use `1e-3` to mach the style of the next line.  Also, put this in an anonymous namespace unless it is an exported symbol somewhere?

tylerjw (approved):
Generally, approve this PR.  Thank you for separating out this small change!  :rocket:

tylerjw (approved):

PR #590, opened 2021-08-03
Wait for complete state duration fix

tylerjw:
### Description

This fixes a bug I found while trying to use the servo demo on rolling.

### DIFF
--- a/moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
+++ b/moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
@@ -251,13 +251,13 @@ bool CurrentStateMonitor::waitForCurrentState(const rclcpp::Time& t, double wait
 
 bool CurrentStateMonitor::waitForCompleteState(double wait_time) const
 {
-  double slept_time = 0.0;
-  double sleep_step_s = std::min(0.05, wait_time / 10.0);
-  rclcpp::Duration sleep_step(sleep_step_s);
-  while (!haveCompleteState() && slept_time < wait_time)
+  const double sleep_step_s = std::min(0.05, wait_time / 10.0);
+  const auto sleep_step_duration = rclcpp::Duration::from_seconds(sleep_step_s);
+  double slept_time_s = 0.0;
+  while (!haveCompleteState() && slept_time_s < wait_time)
   {
-    rclcpp::sleep_for(sleep_step.to_chrono<std::chrono::nanoseconds>());
-    slept_time += sleep_step_s;
+    rclcpp::sleep_for(sleep_step_duration.to_chrono<std::chrono::nanoseconds>());
+    slept_time_s += sleep_step_s;
   }
   return haveCompleteState();
 }
--- a/moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
+++ b/moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
@@ -225,11 +225,11 @@ bool CurrentStateMonitor::haveCompleteStateHelper(const rclcpp::Time& oldest_all
   return (missing_joints == nullptr) || missing_joints->empty();
 }
 
-bool CurrentStateMonitor::waitForCurrentState(const rclcpp::Time& t, double wait_time) const
+bool CurrentStateMonitor::waitForCurrentState(const rclcpp::Time& t, double wait_time_s) const
 {
   rclcpp::Time start = middleware_handle_->now();
   rclcpp::Duration elapsed(0, 0);
-  rclcpp::Duration timeout = rclcpp::Duration::from_seconds(wait_time);
+  rclcpp::Duration timeout = rclcpp::Duration::from_seconds(wait_time_s);
 
   std::unique_lock<std::mutex> lock(state_update_lock_);
   while (current_state_time_ < t)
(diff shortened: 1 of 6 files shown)


moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
```diff
@@ -251,13 +251,13 @@ bool CurrentStateMonitor::waitForCurrentState(const rclcpp::Time& t, double wait
...
+  double slept_time_s = 0.0;
+  while (!haveCompleteState() && slept_time_s < wait_time)
   {
-    rclcpp::sleep_for(sleep_step.to_chrono<std::chrono::nanoseconds>());
-    slept_time += sleep_step_s;
+    rclcpp::sleep_for(sleep_step_duration.to_chrono<std::chrono::nanoseconds>());
+    slept_time_s += sleep_step_s;
   }
   return haveCompleteState();
 }
```
tylerjw, line 263: What happens with the old logic (on rolling):

wait_time = 10 (seconds)
slept_time = 0.0
sleep_time_s = 1
sleep_step = 1ns (constructor for double seconds is gone and the single argument constructor for Duration is now nanoseconds, there is a compiler warning that prints saying this is deprecated)

sleeps for 10 1ns amounts and returns...

As a result servo never hasCompleteState and calls `exit(EXIT_FAILURE);` in the constructor of Servo.  I moved the wait for complete state into the start function.

As an aside I really wish I had colorized ros2 output.  Has anyone figured out how to get that?  It would have been much quicker to debug if the error it printed was red.

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
```diff
@@ -251,13 +251,13 @@ bool CurrentStateMonitor::waitForCurrentState(const rclcpp::Time& t, double wait
 
 bool CurrentStateMonitor::waitForCompleteState(double wait_time) const
 {
-  double slept_time = 0.0;
-  double sleep_step_s = std::min(0.05, wait_time / 10.0);
-  rclcpp::Duration sleep_step(sleep_step_s);
-  while (!haveCompleteState() && slept_time < wait_time)
+  const double sleep_step_s = std::min(0.05, wait_time / 10.0);
+  const auto sleep_step_duration = rclcpp::Duration::from_seconds(sleep_step_s * 1e9);
```
JafarAbdi, line 255:
```suggestion
  const auto sleep_step_duration = rclcpp::Duration::from_seconds(sleep_step_s);
```

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
JafarAbdi: > As an aside I really wish I had colorized ros2 output. Has anyone figured out how to get that? It would have been much quicker to debug if the error it printed was red.

You could do it with `export RCUTILS_COLORIZED_OUTPUT=1`

AndyZe: ^Thanks @JafarAbdi! Adding to my bashrc

tylerjw:
This now comes with a test, restoring the old code in waitForCompleteState fails like I'd expect:

```
[ RUN      ] CurrentStateMonitorTests.WaitForCompleteStateWaits
[INFO] [1628090240.874086555] [moveit_robot_model.robot_model]: Loading robot model 'panda'...
/home/tyler/code/moveit2/ws_moveit2/src/moveit2/moveit_ros/planning/planning_scene_monitor/test/current_state_monitor_tests.cpp:187: Failure
Expected: (std::chrono::duration_cast<std::chrono::nanoseconds>(end - start).count()) > (5e+7), actual: 1946203 vs 5e+07
[  FAILED  ] CurrentStateMonitorTests.WaitForCompleteStateWaits (12 ms)
```

JafarAbdi (approved):

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
```diff
@@ -251,13 +251,13 @@ bool CurrentStateMonitor::waitForCurrentState(const rclcpp::Time& t, double wait
...
-  while (!haveCompleteState() && slept_time < wait_time)
+  const double sleep_step_s = std::min(0.05, wait_time / 10.0);
+  const auto sleep_step_duration = rclcpp::Duration::from_seconds(sleep_step_s);
+  double slept_time_s = 0.0;
+  while (!haveCompleteState() && slept_time_s < wait_time)
   {
-    rclcpp::sleep_for(sleep_step.to_chrono<std::chrono::nanoseconds>());
-    slept_time += sleep_step_s;
+    rclcpp::sleep_for(sleep_step_duration.to_chrono<std::chrono::nanoseconds>());
+    slept_time_s += sleep_step_s;
```
jliukkonen, line 260: Nit about variable naming. I know ROS is "by default" SI units and the base unit for time is second and that should always be the implicit default. Yet, I find it much easier to read code and API if the units are documented as part of the variable names. In this case, it would mean that `wait_time` would be `wait_time_s` and `sleep_step_duration` would be `sleep_step_duration_s`. Thoughts?

vatanaksoytezer (commented):
Should this (and all this DI stuff) be backported to foxy, or we don't care about that?

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
vatanaksoytezer: I didn't thought this before but really liked it 10/10 will use when I need

moveit_ros/planning/planning_scene_monitor/test/current_state_monitor_tests.cpp
```diff
@@ -156,6 +157,36 @@ TEST(CurrentStateMonitorTests, NoModelTest)
...
+TEST(CurrentStateMonitorTests, HaveCompleteStateConstructFalse)
+{
+  // GIVEN a CurrentStateMonitor
+  planning_scene_monitor::CurrentStateMonitor current_state_monitor{
+    std::make_unique<MockMiddlewareHandle>(), moveit::core::loadTestingRobotModel("panda"),
+    std::make_shared<tf2_ros::Buffer>(std::make_shared<rclcpp::Clock>())
+  };
+
+  // WHEN it is constructed
+  // THEN we expect haveCompleteState to be false
```
AndyZe, line 169: nit: this weird capitalization is weird/hard to read. Not sure it adds anything.  WHEN, THEN, etc.

AndyZe (approved):
Approving assuming you update variable names like Jere suggested

moveit_ros/planning/planning_scene_monitor/test/current_state_monitor_tests.cpp
nbbrooks: Does Doxygen support pulling out GWT conditions in some format? That would be a nice motivation for this explicit formatting.

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
tylerjw: sleep_step_duration is a Duration object and is not specifically in the unit seconds, it is in whatever unit you want.  I will change wait_time though.

tylerjw:
> Should this (and all this DI stuff) be backported to foxy, or we don't care about that?

I don't really care to backport it to foxy, but we can if we are using foxy for some project that uses servo.  I think the issue with the constructor for Duration no longer taking double seconds is a rolling/galactic problem.

moveit_ros/planning/planning_scene_monitor/test/current_state_monitor_tests.cpp
tylerjw: The given-when-then comes from BDD (https://en.wikipedia.org/wiki/Behavior-driven_development) and is a practice of testing.  I'll be the first to admit I don't fully understand BDD but that it does seem to provide helpful patterns for writing nice tests.

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
davetcoleman: +1, Google style is to include units

davetcoleman:
So exciting to see test coverage growing!

PR #591, opened 2021-08-03
Add moveit_configs_utils package to simplify loading paramters

JafarAbdi:
### Description

Second attempt to simplify loading parameters

This's based on https://github.com/JafarAbdi/moveit_resources/commit/44a82c49d34bdf60ef5956bb52de57be8a660eb8

Replace: https://github.com/ros-planning/moveit2/pull/209

### Checklist


moveit_demo_nodes/run_move_group/launch/run_move_group.launch.py
```diff
@@ -5,111 +5,35 @@
...
-    # planning_context
-    robot_description_config = xacro.process_file(
-        os.path.join(
-            get_package_share_directory("moveit_resources_panda_moveit_config"),
-            "config",
-            "panda.urdf.xacro",
-        )
+    moveit_config = (
+        MoveItConfigsBuilder("moveit_resources_panda")
+        .robot_description(file_name="panda.urdf.xacro")
```
AdamPettinger, line 14: I think the `str` args here should be filepaths within the package instead of just the name. In `MoveItConfigsBuilder`, this filename expands to `config/panda.urdf.xacro`. That might be OK for all the moveit config packages, but if somebody wanted to use this for their own (e.g. modified) config package, they would have to put files in specific places or modify the configs builder

moveit_demo_nodes/run_move_group/launch/run_move_group.launch.py
```diff
@@ -5,111 +5,35 @@
...
-            moveit_controllers,
-            planning_scene_monitor_parameters,
-            joint_limits_yaml,
+            moveit_config.robot_description,
+            moveit_config.robot_description_semantic,
+            moveit_config.robot_description_kinematics,
+            moveit_config.planning_pipelines,
+            moveit_config.trajectory_execution,
+            moveit_config.planning_scene_monitor,
+            moveit_config.joint_limits,
```
AdamPettinger, line 36: We could probably convert `moveit_config` to a list here so we just need `parameters = [ moveit_config, <some_other_parameter>]`, right?

moveit_demo_nodes/run_move_group/launch/run_move_group.launch.py
```diff
@@ -124,11 +48,11 @@ def generate_launch_description():
...
-            robot_description,
-            robot_description_semantic,
-            ompl_planning_pipeline_config,
-            kinematics_yaml,
-            joint_limits_yaml,
+            moveit_config.robot_description,
+            moveit_config.robot_description_semantic,
+            moveit_config.robot_description_kinematics,
+            moveit_config.planning_pipelines,
+            moveit_config.joint_limits,
```
AdamPettinger, line 55: Above comment about doing `moveit_config` -> list wouldn't help here if we only want to send some of the parameters. (Side note: is there a downside passing all the moveit config parameters here?)

moveit_demo_nodes/run_move_group/launch/run_move_group_interface.launch.py
```diff
@@ -3,53 +3,18 @@
...
-    # planning_context
-    robot_description_config = load_file(
-        "moveit_resources_panda_description", "urdf/panda.urdf"
+    moveit_config = (
+        MoveItConfigsBuilder("moveit_resources_panda")
+        .robot_description(file_name="panda.urdf.xacro")
+        .robot_description_semantic()
+        .robot_description_kinematics()
+        .joint_limits()
+        .moveit_configs()
```
AdamPettinger, line 16: Calling like this seems a bit strange, but I like that it allows you to decide which functions to call, and pass parameters for each.

We should make it clear that you need `.moveit_configs()` last no matter what

AdamPettinger (commented):
Overall I like this. I think there are pros/cons about this style of implementation but this (as written now) is already better than the current way of doing the same configs setup in every launch file

moveit_demo_nodes/run_move_group/launch/run_move_group_interface.launch.py
tylerjw: I don't know how to do it in python but the approach normally used in cpp is to make so the builder object has a conversion operator to the final type so it can be converted.  In python the types are dynamic so it means you probably need a function like this that'll return the moveit configs object.  Maybe there was some sort of to_dict() or something you could overload so the moveit_config object would just be converted into a dictionary when you use it like one?

tylerjw:
I like the pattern here.  I wonder if there is a way to make it more generic so the sytax was more like:

```python
parameters = ParameterBuilder('robot_config_package')
  .yaml('moveit_cpp.yaml')                                                          # loads a yaml file into parameters
  .file_parameter('robot_description', 'xacro/panda.urdf.xacro')  # fills out full path based on package
  .parameter('planning_timeout', 0.1)                                         # sets a parameter
  .to_dict()
```

Then it could be used for any sort of parameters, not just a moveit config.  Then maybe there could be a moveit config builder that used the ParameterBuilder and was just a function or something like that.

AdamPettinger:
> I like the pattern here. I wonder if there is a way to make it more generic so the sytax was more like:
> 
> ```python
> parameters = ParameterBuilder('robot_config_package')
>   .yaml('moveit_cpp.yaml')                                        # loads a yaml file into parameters
>   .file_parameter('robot_description', 'xacro/panda.urdf.xacro')  # fills out full path based on package
>   .parameter('planning_timeout', 0.1)                             # sets a parameter
>   .to_dict()
> ```
> 
> Then it could be used for any sort of parameters, not just a moveit config. Then maybe there could be a moveit config builder that used the ParameterBuilder and was just a function or something like that.

I like this a lot. It would let you do multiple inside one launch file (e.g. when getting parameters from multiple packages), and would also let you use it to return a specific set of all parameters for a config package (e.g. moveit_configs, or if some other user wants to have their own bringup/config package they could provide a `getPackageConfigs()` call)

moveit_configs_utils/src/moveit_configs_utils/moveit_configs_builder.py
```diff
@@ -0,0 +1,196 @@
...
+#     MoveItConfigsBuilder("ROBOT_NAME")
+#         .robot_description()
+#         .robot_description_semantic()
+#         .robot_description_kinematics()
+#         .joint_limits()
+#         .trajectory_execution(file_name="panda_gripper_controllers.yaml")
+#         .planning_scene_monitor()
+#         .planning_pipelines()
+#         .moveit_configs()
+# )
```
JafarAbdi: Remove

moveit_configs_utils/setup.py
```diff
@@ -0,0 +1,30 @@
...
+    keywords=['ROS2'],
+    classifiers=[
+        'Intended Audience :: Developers',
+        'License :: OSI Approved :: BSD',
+        'Programming Language :: Python',
+        'Topic :: Software Development',
+    ],
+    description='TODO',
+    license='BSD',
+)
```
JafarAbdi: This need update

moveit_configs_utils/package.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>moveit_configs_utils</name>
+  <version>0.0.0</version>
+  <description>TODO: Package description</description>
+  <maintainer email="cafer.abdi@gmail.com">jafar_abdi</maintainer>
+  <license>TODO: License declaration</license>
```
JafarAbdi, line 8: Update

moveit_ros/moveit_servo/test/launch/unit_test_servo_calcs.test.py
```diff
@@ -4,15 +4,18 @@
...
 sys.path.append(os.path.dirname(__file__))
-from servo_launch_test_common import load_yaml
 
 
 def generate_test_description():
     # Get parameters using the demo config file
-    servo_yaml = load_yaml("moveit_servo", "config/panda_simulated_config.yaml")
-    servo_params = {"moveit_servo": servo_yaml}
+    servo_params = {
+        "moveit_servo": ParameterBuilder("moveit_servo")
```
JafarAbdi, line 15: Use `parameter_namespace`

moveit_configs_utils/src/moveit_configs_utils/moveit_configs_builder.py
```diff
@@ -0,0 +1,201 @@
...
+            "planning_pipelines": pipelines,
+            "default_planning_pipeline": default_planning_pipeline,
+        }
+        for pipeline in pipelines:
+            self.__moveit_configs.planning_pipelines[pipeline] = load_yaml(
+                self._package_path / "config" / (pipeline + "_planning.yaml")
+            )
+        return self
+
+    def moveit_configs(self):
```
henningkayser, line 173: This function name is not really intuitive to me. Maybe use something like `to_config()` along with `to_dict()`?

moveit_configs_utils/src/moveit_configs_utils/moveit_configs_builder.py
```diff
@@ -0,0 +1,201 @@
...
+        return self.__moveit_configs
+
+    def to_dict(self, include_moveit_configs: bool = True):
+        parameters = self._parameters
```
henningkayser, line 177: Where does `self._parameters` come from?

moveit_demo_nodes/run_move_group/launch/run_move_group.launch.py
henningkayser: So this could use `to_dict()` then, no? I like the explicitness, though.

henningkayser (commented):
This is awesome, all launch files look so much cleaner! I'm just not sure if we should remove the default values from the builder functions.

Some thought: Let's say all moveit config packages provide python scripts that export their custom moveit configs. Couldn't we then simply build config hierarchies by taking another included moveit config as default and just modifying specific parts, say the kinematics yaml or something else? Given this PR it seems like a really small step to get this done. The only thing we need to watch out for is circular dependencies.

moveit_configs_utils/src/moveit_configs_utils/moveit_configs_builder.py
```diff
@@ -0,0 +1,201 @@
...
+    robot_description = None
+    robot_description_semantic = None
+    robot_description_planning = None
+    robot_description_kinematics = None
+    planning_pipelines = None
+    trajectory_execution = None
+    planning_scene_monitor = None
+    move_group_capabilities = None
+    move_group = None
+    joint_limits = None
```
sjahr, line 27: Some planners (at least the PILZ planner) require Cartesian limits which look i.e. like this:
```
cartesian_limits:
  max_trans_vel: 1
  max_trans_acc: 2.25
  max_trans_dec: -5
  max_rot_vel: 1.57
```
Maybe you could add this as a default field too.

moveit_configs_utils/src/moveit_configs_utils/moveit_configs_builder.py
JafarAbdi: We decided to use `to_moveit_configs()`

moveit2_rolling.repos
```diff
@@ -11,3 +15,7 @@ repositories:
     type: git
     url: https://github.com/ros-planning/warehouse_ros_mongo
     version: ros2
+  parameter_tools:
+    type: git
+    url: https://github.com/JafarAbdi/parameter_tools
```
JafarAbdi, line 20: Update to upstream

moveit2_galactic.repos
```diff
@@ -1,5 +1,9 @@
 # TODO(henningkayser): Remove once ros2_control is released (https://github.com/ros-controls/ros2_control/issues/421)
 repositories:
+  moveit_resources:
+    type: git
+    url: https://github.com/JafarAbdi/moveit_resources
+    version: pr-moveit_configs_utils
```
tylerjw, line 6: Because of how CI works now if you put something in moveit2.repos you don't also need it here.  Robert (using some shell scripting magic) made so the ros version-specific repos files extend the base one.  Putting it here might be the cause of some of the CI failures.

moveit2_galactic.repos
```diff
@@ -8,3 +12,7 @@ repositories:
     type: git
     url: https://github.com/ros-controls/ros2_controllers
     version: master
+  parameter_tools:
+    type: git
+    url: https://github.com/PickNikRobotics/parameter_tools
+    version: main
```
tylerjw, line 18: :point_up:

moveit_demo_nodes/run_move_group/launch/run_move_group.launch.py
```diff
@@ -1,115 +1,37 @@
...
         parameters=[
-            robot_description,
-            robot_description_semantic,
-            kinematics_yaml,
-            ompl_planning_pipeline_config,
-            trajectory_execution,
-            moveit_controllers,
-            planning_scene_monitor_parameters,
-            joint_limits_yaml,
+            moveit_config.robot_description,
```
tylerjw, line 28: It would be nice if you could just write:

```python
parameters = [moveit_config.to_dict()]
```
or something like that?

moveit_demo_nodes/run_move_group/launch/run_move_group.launch.py
```diff
@@ -124,11 +46,11 @@ def generate_launch_description():
...
-            robot_description,
-            robot_description_semantic,
-            ompl_planning_pipeline_config,
-            kinematics_yaml,
-            joint_limits_yaml,
+            moveit_config.robot_description,
+            moveit_config.robot_description_semantic,
+            moveit_config.robot_description_kinematics,
+            moveit_config.planning_pipelines,
+            moveit_config.joint_limits,
```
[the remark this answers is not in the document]
tylerjw: Probably out of the scope of this PR but it would be really nice if the rviz plugin had the option of being pointed at the ros node that had the moveit config loaded in it (maybe even as something that could be changed in the gui) and it would use the `get_parameters` service that ros starts up to read the moveit config from that other node.

tylerjw (approved):
Overall I really like this PR.  I just left a couple of comments.

moveit_configs_utils/src/moveit_configs_utils/moveit_configs_builder.py
```diff
@@ -0,0 +1,201 @@
...
+    __srdf_filename: str
+    __robot_description: str
+
+    # Look-up for robot_name_moveit_config package
+    def __init__(self, robot_name: str, robot_description="robot_description"):
+        super().__init__(robot_name + "_moveit_config")
+        self.__robot_name = robot_name
+        setup_assistant_file = self._package_path / ".setup_assistant"
+        if not setup_assistant_file.exists():
+            # TODO: Should this throw an exception .?
```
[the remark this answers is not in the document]
JafarAbdi: I think we should, all the configs files are associated with the urdf/srdf parameters from the `.setup_assistant` file

moveit_configs_utils/src/moveit_configs_utils/moveit_configs_builder.py
```diff
@@ -0,0 +1,201 @@
...
+            )
+        }
+        return self
+
+    def robot_description_semantic(
+        self, file_path: Optional[str] = None, mappings: dict = None
+    ):
+        self.__moveit_configs.robot_description_semantic = {
+            self.__robot_description
+            + "_semantic": load_xacro(
```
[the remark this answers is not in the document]
JafarAbdi: Yes, it can

moveit_configs_utils/src/moveit_configs_utils/moveit_configs_builder.py
```diff
@@ -0,0 +1,201 @@
...
+                mappings=mappings,
+            )
+        }
+        return self
+
+    def robot_description_kinematics(self, file_path: Optional[str] = None):
+        self.__moveit_configs.robot_description_kinematics = {
+            self.__robot_description
+            + "_kinematics": load_yaml(
+                self._package_path / (file_path or "config/kinematics.yaml")
```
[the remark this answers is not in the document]
JafarAbdi: You mean `config/kinematics.yaml` right .? I'll update the class to have them

moveit_demo_nodes/run_move_group/launch/run_move_group.launch.py
```diff
@@ -1,115 +1,37 @@
...
-        "moveit_resources_panda_moveit_config", "config/ompl_planning.yaml"
-    )
-    ompl_planning_pipeline_config["ompl"].update(ompl_planning_yaml)
-
-    # Trajectory Execution Functionality
-    moveit_simple_controllers_yaml = load_yaml(
-        "moveit_resources_panda_moveit_config", "config/panda_controllers.yaml"
+    moveit_config = (
+        MoveItConfigsBuilder("moveit_resources_panda")
+        .robot_description(file_path="config/panda.urdf.xacro")
```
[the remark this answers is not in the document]
JafarAbdi: You could pass it to the robot description function `.robot_description(file_path="config/panda.urdf.xacro", mappings={"param1": value, ....})`

moveit_configs_utils/src/moveit_configs_utils/moveit_configs_builder.py
JafarAbdi: Done

moveit_demo_nodes/run_move_group/launch/run_move_group.launch.py
JafarAbdi: I'll update it now

henningkayser (commented):
@JafarAbdi I think this really looks about to get ready to be merged. Before merging and adding the dependency to `parameter_tools`, I'd suggest switching to a more specific package name that distinguishes it from c++ packages.
Something like
* launch_parameter_builder
* parameter_tools_py

JafarAbdi:
> @JafarAbdi I think this really looks about to get ready to be merged. Before merging and adding the dependency to `parameter_tools`, I'd suggest switching to a more specific package name that distinguishes it from c++ packages.
> Something like
> 
> * launch_parameter_builder
> * parameter_tools_py

The `parameter_tools` will have two packages in it, the current Python one and a C++ one

henningkayser:
> The `parameter_tools` will have two packages in it, the current Python one and a C++ one

I know, I'm talking about the package name `parameter_builder`, not the repo (which also could use a reference to ROS 2, but that's another topic). It's just very easy to get lost in these kinds of helper packages if they have very generic names. I'm just thinking that we still have drafts for a C++ config builder API based on `ParameterSet` which should be completed in the future. Who will be able to distinguish these when all packages are named `parameter_set`, `parameter_builder`, `config_builder`, etc...
Names like `rclcpp_parameter_set`, `launch_parameter_builder` (just suggestions) would be much less confusing.

moveit_configs_utils/moveit_configs_utils/moveit_configs_builder.py
```diff
@@ -0,0 +1,303 @@
...
+        for pipeline in pipelines:
+            self.__moveit_configs.planning_pipelines[pipeline] = load_yaml(
+                self._package_path / "config" / (pipeline + "_planning.yaml")
+            )
+        return self
+
+    def cartesian_limits(self, file_path: Optional[str] = None):
+        self.__moveit_configs.cartesian_limits = load_yaml(
+            self._package_path / (file_path or "config/cartesian_limits.yaml")
+        )
```
sjahr, line 293: Thanks for adding the cartesian limits! Similar to the joint limits, cartesian limits where within the robot_description_planning namespace i.e. in the [panda_moveit_config](https://github.com/ros-planning/moveit_resources/blob/master/panda_moveit_config/launch/planning_context.launch). I guess it would make sense to keep it that way for the sake of uniformity. What do you think?
```suggestion
        self.__moveit_configs.cartesian_limits = {
            self.__robot_description
            + "_planning": load_yaml(
                self._package_path / (file_path or "config/cartesian_limits.yaml")
            )
        }
```

moveit_configs_utils/moveit_configs_utils/moveit_configs_builder.py
```diff
@@ -0,0 +1,303 @@
...
+        self.planning_scene_monitor = {}
+        self.move_group_capabilities = {}
+        self.move_group = {}
+        self.joint_limits = {}
+        self.moveit_cpp = {}
+        self.cartesian_limits = {}
+
+    @property
+    def robot_description(self):
+        return self.__robot_description
```
gavanderhoorn, line 41: [PEP-8](https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables) seems to suggest using double underscore prefixes for simple member variables is not recommended.

davetcoleman:
> Using catchall names such as utils should be avoided as they do not scope what goes into the package or what should be outside the package.

Can we rename this package to something more specific, per this [ROS Convention](http://wiki.ros.org/ROS/Patterns/Conventions)?

Something like:
- moveit_config_building
- moveit_param_builder

moveit2.repos
```diff
@@ -15,3 +15,7 @@ repositories:
     type: git
     url: https://github.com/ros-planning/warehouse_ros_mongo
     version: ros2
+  parameter_tools:
```
davetcoleman, line 18: @JafarAbdi This package name is also a bit ambiguous per http://wiki.ros.org/ROS/Patterns/Conventions

moveit_configs_utils/moveit_configs_utils/moveit_configs_builder.py
```diff
@@ -0,0 +1,303 @@
+from pathlib import Path
+from typing import Optional, List
+import logging
+from ament_index_python.packages import get_package_share_directory
+
+from parameter_builder import ParameterBuilder, load_yaml, load_xacro
+
```
davetcoleman, line 7: Can we add inline documentation at the top describing how this file is used, please? Just copy Tyler's first comment in this PR with his example code.

moveit_configs_utils/package.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>moveit_configs_utils</name>
+  <version>0.0.0</version>
+  <description>Python library for loading moveit config parameters in launch files</description>
+  <maintainer email="moveit_releasers@googlegroups.com">MoveIt Release Team</maintainer>
```
davetcoleman, line 7: Please add yourself as author @JafarAbdi

moveit_configs_utils/package.xml
```diff
@@ -0,0 +1,19 @@
...
+  <test_depend>ament_lint_auto</test_depend>
+  <test_depend>ament_lint_common</test_depend>
+
+  <exec_depend>ament_index_python</exec_depend>
+  <exec_depend>parameter_builder</exec_depend>
+
+  <export>
+    <build_type>ament_python</build_type>
+  </export>
+</package>
```
davetcoleman, line 19: Why is the file below empty?
> moveit_configs_utils/resource/moveit_configs_utils

moveit_configs_utils/setup.py
```diff
@@ -0,0 +1,23 @@
+from setuptools import setup
+
+package_name = "moveit_configs_utils"
+
+setup(
+    name=package_name,
+    version="0.0.0",
```
davetcoleman, line 7: Can we make the initial version 0.0.1? For some reason in MoveIt 1, [we used 0.3.0](https://github.com/ros-planning/moveit2/blob/main/moveit_setup_assistant/templates/moveit_config_pkg_template/package.xml.template). Either way, 0.0.0 feels wrong to me.

moveit_configs_utils/setup.py
```diff
@@ -0,0 +1,23 @@
...
+    packages=[package_name],
+    data_files=[
+        ("share/ament_index/resource_index/packages", ["resource/" + package_name]),
+        ("share/" + package_name, ["package.xml"]),
+    ],
+    install_requires=["setuptools"],
+    zip_safe=True,
+    maintainer="MoveIt Release Team",
+    maintainer_email="moveit_releasers@googlegroups.com",
+    description="Python library for loading moveit config parameters in launch files",
```
davetcoleman, line 17:
```suggestion
    description="Python library for loading MoveIt config parameters in launch files",
```

moveit_demo_nodes/run_move_group/launch/run_move_group.launch.py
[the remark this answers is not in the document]
davetcoleman: Please add documentation for this soemwhere, perhaps in a new MoveIt tutorial? I think I would actually prefer just a section here: https://moveit.ros.org/documentation/contributing/code/

davetcoleman (commented):
Great work!
Should this be a moveit-agnostic package?

moveit2.repos
tylerjw: Do you have a suggestion for a better name?  

The intent of the repo `parameter_tools` is infrastructure for the use of ros2 parameters (similar to rosparam_shortcuts was for ros1).  The reason we wanted to create a new repo is that rosparam_shortcuts was ported to ros2 but I didn't think the approach worked well to solve the new api and pain points of the ros2 parameter API.  Secondly, rosparam_shortcuts is one ros package in a repo and I wanted to have a repo with multiple ros packages as what Jafar created (parameter_builder) is a python package and what I'm working on (parameter_sets) is a C++ package that solve separate problems around the ros2 parameters.

I didn't want to conflict with the rosparam_shortcuts name and wanted a repo that contained tools for making ros2 paramters better.  A big reason why I don't want to put it in moveit2 itself is to avoid creating yet another circular dependency similar to the issues with the moveit_resources package.  moveit2 config packages (like moveit_resources) will need to depend on this package so it would be much nicer (for release testing) if this code is in a separate repo from moveit2.

tylerjw:
> Can we rename this package to something more specific, per this [ROS Convention](http://wiki.ros.org/ROS/Patterns/Conventions)?
> 
> Something like:
> 
>     * moveit_config_building
> 
>     * moveit_param_builder

It might be worth separating parameter_builder from the moveit specific config portions of it eventually to make the more general pattern here apply outside of moveit.  One obvious flaw in the name `parameter_tools` and `parameter_builder` is that it does not tell the user that it was created for loading moveit configs.

moveit_configs_utils/moveit_configs_utils/moveit_configs_builder.py
```diff
@@ -0,0 +1,303 @@
...
+                mappings=mappings,
+            )
+        }
+        return self
+
+    def robot_description_kinematics(self, file_path: Optional[str] = None):
+        self.__moveit_configs.robot_description_kinematics = {
+            self.__robot_description
+            + "_kinematics": load_yaml(
+                self._package_path / (file_path or "config/kinematics.yaml")
```
v4hn, line 215: All file paths with explicit `/` in a string will fail on windows, won't they?
I guess you have to use a filepath class for these defaults too?

moveit_demo_nodes/run_move_group/launch/run_move_group_interface.launch.py
```diff
@@ -1,55 +1,17 @@
...
+from moveit_configs_utils.moveit_configs_builder import MoveItConfigsBuilder
 
 
 def generate_launch_description():
-    # planning_context
-    robot_description_config = load_file(
-        "moveit_resources_panda_description", "urdf/panda.urdf"
+    moveit_config = (
+        MoveItConfigsBuilder("moveit_resources_panda")
+        .robot_description(file_path="config/panda.urdf.xacro")
```
v4hn, line 9: See my other comment, but this one is on the user-side which is even worse usability from my perspective..
Not sure anything can be done about that.

moveit_demo_nodes/run_move_group/launch/run_move_group_interface.launch.py
```diff
@@ -1,55 +1,17 @@
...
+        .robot_description_semantic()
+        .robot_description_kinematics()
+        .joint_limits()
+        .to_moveit_configs()
```
v4hn, line 13: I know you might not need all parameters everywhere, but why did you decide to specify every single parameter for each use? I think I might prefer
```
MoveItConfigsBuilder("moveit_resources_panda")
.robot_description(file_path="config/panda.urdf.xacro")
.to_moveit_configs()
```
to load all other parameters with their defaults.

v4hn (commented):
I just looked into this PR in more detail for the first time. This work is such a gem to make the ROS2 launch file mess usable!

Some spontaneous feedback inline.

moveit_demo_nodes/run_move_group/launch/run_move_group_interface.launch.py
tylerjw: :+1: to this suggestion.  I think one of the large barriers to the usability of moveit is many things don't have defaults that should.  Here is one case where the default behavior (what parts of the moveit config to load) could be defined somewhere and the builder could then just override various parts of it.

vatanaksoytezer:
Is there anything holding us merging this?

henningkayser:
> Is there anything holding us merging this, except for conflicts?

I think there are still several open requests and suggestions. But I agree that we should get something merged soon.

moveit_configs_utils/package.xml
JafarAbdi: This's needed for ament

JafarAbdi:
I would really appreciate a final review from you @jrgnicho @henningkayser @AndyZe @tylerjw

moveit_configs_utils/moveit_configs_utils/moveit_configs_builder.py
```diff
@@ -0,0 +1,445 @@
...
+        self.planning_pipelines = {}
+        # A dictionary contains parameters for trajectory execution & moveit controller managers.
+        self.trajectory_execution = {}
+        # A dictionary that have the planning scene monitor's parameters.
+        self.planning_scene_monitor = {}
+        # A dictionary containing move_group's non-default capabilities.
+        self.move_group_capabilities = {}
+        # A dictionary containing the overridden position/velocity/acceleration limits.
+        self.joint_limits = {}
+        # A dictionary MoveItCpp related parameters.
```
AndyZe, line 93:
```suggestion
        # A dictionary containing MoveItCpp related parameters.
```

moveit_configs_utils/moveit_configs_utils/moveit_configs_builder.py
```diff
@@ -0,0 +1,445 @@
...
+        "__move_group_capabilities",
+        "__joint_limits",
+        "__moveit_cpp",
+        "__cartesian_limits",
+    ]
+
+    def __init__(self):
+        # A dictionary that have the content of the URDF file.
+        self.robot_description = {}
+        # A dictionary that have the content of the SRDF file.
```
AndyZe, line 79:
```suggestion
        # A dictionary that has the contents of the URDF file.
        self.robot_description = {}
        # A dictionary that has the contents of the SRDF file.
```

moveit_configs_utils/moveit_configs_utils/moveit_configs_builder.py
AndyZe: Ping to fix this

AndyZe (commented):
I think this is a great idea. I've done something similar on a client project and it helped a lot.

Will this work with namespaces for dual arms? Will approve if the answer is yes.

stephanie-eng (commented):
Small nit - some comments reference robot_name_moveit_config**s** as the default package name, which caused me a bit of confusion.

moveit_configs_utils/moveit_configs_utils/moveit_configs_builder.py
```diff
@@ -0,0 +1,445 @@
...
+    @property
+    def planning_scene_monitor(self):
+        return self.__planning_scene_monitor
+
+    @planning_scene_monitor.setter
+    def planning_scene_monitor(self, value):
+        self.__planning_scene_monitor = value
+
+    @property
+    def move_group_capabilities(self):
```
stephanie-eng, line 147: Add move group capabilities loader function

moveit_configs_utils/moveit_configs_utils/moveit_configs_builder.py
```diff
@@ -0,0 +1,445 @@
...
+            robot_description_file_path = self._package_path / file_path
+        self.__moveit_configs.robot_description = {
+            self.__robot_description: load_xacro(
+                robot_description_file_path, mappings=mappings
+            )
+        }
+        return self
+
+    def robot_description_semantic(
+        self, file_path: Optional[str] = None, mappings: dict = None
```
stephanie-eng, line 262: Support passing LaunchConfiguration as a mapping

AndyZe (approved):

AndyZe:
I just spot-checked this on some of the tutorials. It still works well  :+1:

We still haven't tested it on a dual-arm system due to some unexpected issues, but I don't want to block a great PR forever. We can always fix up whatever issues emerge.

moveit_configs_utils/moveit_configs_utils/moveit_configs_builder.py
```diff
@@ -0,0 +1,445 @@
...
+                    .robot_description_semantic(Path.home() / "my_config" / "new_file.srdf")
+                    .to_moveit_configs()
+"""
+
+from pathlib import Path
+from typing import Optional, List
+import logging
+from ament_index_python.packages import get_package_share_directory
+
+from parameter_builder import ParameterBuilder, load_yaml, load_xacro
```
tylerjw, line 57: should you rename you package to launch_param_builder now that it has that repo name?

tylerjw (approved):

PR #594, opened 2021-08-05
WIP / Discussion: Move automated build for dockers to Github Registry

vatanaksoytezer:
Github's Docker Registry is free and well integrated into Github Action ecosystem. This PR is an attempt to start automated docker builds in with Github Actions and pushes the dockers to Github's Docker Registry at [ghcr.io](https://github.com/features/packages). Feel free to discuss / comment on if this would be a good way of doing automated docker builds.

I'll comment out the runs on pull-requests before merging.


vatanaksoytezer:
I would need to include a private access token for the ros-planning organization that has push access

vatanaksoytezer:
Also see https://github.com/vatanaksoytezer/moveit2/pkgs/container/moveit2 on how these look in Github

tylerjw:
This looks really nice.  @henningkayser could you create the token so we can try using it for this?

dockerfiles/ci-testing/Dockerfile
```diff
@@ -1,8 +1,8 @@
-# moveit/moveit2:foxy-ci-testing
+# ghcr.io/ros-planning/moveit2:${ROS_DISTRO}-ci-testing
```
rhaschke, line 1: What's the benefit of switching the container registry?
Does that require changing all our docs that refer to docker containers?

.github/workflows/build_dockers.yaml
```diff
@@ -0,0 +1,53 @@
...
+  pull_request:
+  push:
+    branches:
+      - main
+
+jobs:
+  build_docker:
+    strategy:
+      matrix:
+        env:
```
rhaschke, line 14: `matrix.env` requires all combinations to be explicitly listed. Use [real matrix options](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix).

rhaschke (commented):
Hm. I don't yet see the benefits of moving docker images from dockerhub to ghcr. 
Could you explain more? I guess you plan to disable dockerhub eventually?

Why do you move `.docker` to `dockerfiles`?
You should also define [interdependencies between jobs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idneeds).

.github/workflows/build_dockers.yaml
vatanaksoytezer: > `matrix.env` requires all combinations to be explicitly listed. Use [real matrix options](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix).

Thanks for the heads up! Will change

dockerfiles/ci-testing/Dockerfile
vatanaksoytezer: > Does that require changing all our docs that refer to docker containers?

For MoveIt 2 we don't actually have any documentation on dockers. For MoveIt, I think we only refer them in the website, which I will send a PR, for sure to update docker location.

vatanaksoytezer:
Thanks a lot for the review @rhaschke!

> Hm. I don't yet see the benefits of moving docker images from dockerhub to ghcr.
That's mostly my liking to put everything in Github environment, I am pretty sure everything I write here is doable with a nice GHA that triggers dockerhub builds but following is a list of benefits that I can think of:

- Everything stays in Github.
- Good integration with GitHub Actions (we can proabably integrate dockerhub the same way as well with a few secret variables)
- Docker is being built in GHA, so easy to track and resolve errors.
- We can make workflows and get all the benefits from GHA (maybe daily builds?) as you mentioned.
- Not important but, it is free and unlimited space for open source projects.

> Could you explain more? I guess you plan to disable dockerhub eventually?
Yes, if we can move everything, we can disable dockerhub.

> Why do you move `.docker` to `dockerfiles`?
AFAIK .docker folder is associated with Dockerhub builds, it is just something that is not `.docker`. I am open to any suggestions on this.

> You should also define [interdependencies between jobs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idneeds).
Sure, I'll add this once I get the PAT, and make sure I can build within ros-planning.

rhaschke:
I think all of your arguments apply for dockerhub as well. I suggest sticking with dockerhub, but building images via GHA (essentially using your PR, but pushing to dockerhub instead of ghcr). The benefit of dockerhub is that this is the default image repository - users don't need to refer to the new one.

henningkayser (commented):
Well, look at that https://github.com/ros-planning/moveit2/pkgs/container/moveit2. @vatanaksoytezer This is great! I think it's much better than having to configure the auto-builds on dockerhub. IMO, there is nothing wrong with having the images on ghcr and simply copying them over to dockerhub for convenience. For CI, I would rely on ghcr images so that we don't run into the rate limitation issue again. So what's still missing for this to get merged?

* build release and source docker images
* enable caching?

vatanaksoytezer:
> So what's still missing for this to get merged?
> 
> * build release and source docker images

Release and source dockers are essential before merging this. I'll add it to this PR.

> * enable caching?

I am not super experienced with caching but I noted that to myself as well. I'll look into what I can do for this. But before caching we can still merge, and then leave it to another PR as well.

Also one more thing I would need to disable builds on PR.

vatanaksoytezer:
@rhaschke @tylerjw @henningkayser I've been playing with how I can this the cleanest, but ended up doing the ugly thing to make jobs dependent to each other. So the docker dependencies should be as follows:

- release: does not depend on anything
- ci: does not depend on anything
- ci_testing: depends on ci
- source: depends on ci_testing

So release can run parallel with the rest, but ci->ci_testing->source should be in order. 

- If we are OK using previously built dockers for CI, I can parallelize these and just use one matrix for the whole thing. 
- I can also change the dockerfiles to not inherit the other dockers and just copy and paste content, so docker gets built each time from scratch. That would remove the dependencies, and I can just run everything in parallel.
- If we want sequential jobs in every run (as @rhaschke mentioned), I couldn't find a cleaner way. I am happy to implement or look into if you have a cleaner solution.

I've been looking at [concurrency](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency) (see https://github.com/ros-planning/moveit2/pull/594/commits/786a3a3f58759efc253aad79ac786c68659cf390) for making concurrent jobs so that we can run rolling and galactic in parallel but rest in sequantial, but doesn't seem very consistent and build matrix doesn't seem to always choose the first job (ci in this case.  I've also looked at [composite runs](https://github.blog/changelog/2020-08-07-github-actions-composite-run-steps/), for reducing the repeated lines with steps but I think this not we want. I've found a few hacky ways to use steps again with yaml scripting but I feel that is too unnecessary and will reduce the maintainability of CI.

Let me know what you think!

rhaschke (approved):
Unfortunately, GHA don't support yaml _anchors_, which would easily allow avoiding the ugly code duplication. However, I'm afraid without anchors, we don't have another choice...
Given the rather long job names (e.g. `Build Dockers / build_release_docker (galactic, release)`), I pushed an improved naming scheme.

Note, that the jobs fail with permission issues pushing the built docker images!

rhaschke:
Note: the CI jobs need to be adapted to use the new docker images as well!

vatanaksoytezer:
Thanks a ton for giving a hand @rhaschke!

> Note, that the jobs fail with permission issues pushing the built docker images!

@henningkayser  opened up a test branch for this, we will test it there and then after seeing all green hopefully merge this.

> Note: the CI jobs need to be adapted to use the new docker images as well!

Sure thing, I'll update the CI jobs as well.

v4hn:
Does anyone plan to move the MoveIt docker images as well or is this supposed to be a MoveIt2-only migration?
I seem to recall Henning planned to migrate the moveit repository as well.

vatanaksoytezer:
@v4hn I can, once we are all good settled with moveit2's migration.

PR #583, opened 2021-07-30
Format includes using clang-format

tylerjw:
### Description

Replaces https://github.com/ros-planning/moveit2/pull/575

Uses clang-format to format includes and fixes missing includes.

I will backport the .clang-format change to moveit1 and fix any missing includes there too.

PR #598, opened 2021-08-10
Clean galactic / rolling warnings

vatanaksoytezer:
Since we've branched off we can now get rid of the warnings. We also need to merge https://github.com/ros-planning/warehouse_ros/pull/85 (this fails in foxy and galactic but passes on rolling) and open up foxy branches for `rviz_visual_tools` and `moveit_visual_tools` and make some cleanup there as well


vatanaksoytezer:
Bad news from CI is that they deprecated these headers in rolling, but they do not exist in galactic... So we either need branch off for galactic as well, or live with the warnings ...

vatanaksoytezer:
Good news, is that we can solve the header problems with https://en.cppreference.com/w/cpp/preprocessor/include as Henning suggested

henningkayser (commented):
Looks good! I find the duration constructor a little too clunky, `rclcpp::Duration::from_seconds()` or the chrono literals would be cleaner, imo. @tylerjw what do you think, should we get #583 merged before or after this one?

vatanaksoytezer:
> I find the duration constructor a little too clunky, `rclcpp::Duration::from_seconds()` or the chrono literals would be cleaner, imo.

Sure, I think chrono literals is fair. I will add this soon.

moveit_core/kinematic_constraints/src/kinematic_constraint.cpp
```diff
@@ -951,7 +955,7 @@ void VisibilityConstraint::getMarkers(const moveit::core::RobotState& state,
   mk.pose.orientation.y = 0;
   mk.pose.orientation.z = 0;
   mk.pose.orientation.w = 1;
-  mk.lifetime = rclcpp::Duration(60);
+  mk.lifetime = rclcpp::Duration(std::chrono::seconds(60));
```
tylerjw, line 958: I agree with Henning about these.  I think doing it like this is nicer:

```suggestion
  mk.lifetime = rclcpp::Duration::from_seconds(60);
```

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -1489,7 +1493,7 @@ bool TrajectoryExecutionManager::executePart(std::size_t part_index)
       if (context.trajectory_parts_[longest_part].joint_trajectory.points.size() >=
           context.trajectory_parts_[longest_part].multi_dof_joint_trajectory.points.size())
       {
-        rclcpp::Duration d(0.0);
+        rclcpp::Duration d(std::chrono::seconds(0));
```
tylerjw, line 1496: In this case I would do this:

```suggestion
        auto d = rclcpp::Duration::from_seconds(0);
```

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
```diff
@@ -393,7 +393,8 @@ void MotionPlanningFrame::changePlanningGroupHelper()
       // /std::shared_ptr<tf2_ros::Buffer> tf_buffer = context_->getFrameManager()->getTF2BufferPtr();
       std::shared_ptr<tf2_ros::Buffer> tf_buffer = moveit::planning_interface::getSharedTF();
 #endif
-      move_group_.reset(new moveit::planning_interface::MoveGroupInterface(node_, opt, tf_buffer, rclcpp::Duration(30)));
+      move_group_.reset(new moveit::planning_interface::MoveGroupInterface(node_, opt, tf_buffer,
+                                                                           rclcpp::Duration(std::chrono::seconds(30))));
```
tylerjw, line 397: Assuming this is a shared pointer I'd change this to:

```suggestion
      move_group_ = std::make_shared<moveit::planning_interface::MoveGroupInterface>(node_, opt, tf_buffer,
                                                                           rclcpp::Duration::from_seconds(30));
```

tylerjw (approved):
Looked through all these changes.  Thank you for doing this.  I left a couple minor nits.

tylerjw:
> Looks good! I find the duration constructor a little too clunky, `rclcpp::Duration::from_seconds()` or the chrono literals would be cleaner, imo. @tylerjw what do you think, should we get #583 merged before or after this one?

I don't care either way.  I think after Vatan makes a couple of changes this one might be easier to merge (unless you just want to merge format change into moveit2 one without waiting for the moveit1 version).  We can do it in either order, but if we are waiting on the moveit1 format change, let's not hold this up.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
vatanaksoytezer: I've fixed this one but there are a ton of (around ~100) `new`s used to reset shared pointers in MoveIt 2. I am gonna go ahead and open an issue to track this and send a fix soon-ish.

tylerjw (approved):

JafarAbdi (approved):
LGTM, we should open an issue to remove the `__has_include`s once we branch galactic

PR #600, opened 2021-08-10
Format Includes with clang-format

tylerjw:
### Description

Backport of https://github.com/ros-planning/moveit2/pull/583 to make it easier to sync other changes.


henningkayser (approved):

vatanaksoytezer:
Are waiting on anything to merge this?

AndyZe (approved):
These changes look positive to me

AndyZe:
Do we still want to target `foxy`? I think it should target `main` at this point.

vatanaksoytezer:
> Do we still want to target `foxy`? I think it should target `main` at this point.

I agree with that we should target main, maybe we can do it in both?

v4hn:
> Are we waiting on anything to merge this?

Did you read the PR description? This is a **backport** of an unmerged PR for `main`. So yes, it waits for the actual PR to be merged. And that one is conflicting and left with open review comments right now.

I'll close this for now, though it might make sense to create a new one if Tyler decides that it makes sense to pick this up again.

tylerjw:
As @v4hn said, this is unfinished work.  Basically I gave up on this as I saw it as less important than other work.  This was a minor-ish annoyance of mine and I was hoping it would be easy to get buy-in for blanket formatting of the includes with fairly simple rules.  In some cases the re-ordering of the includes didn't make much sense and I was asked to make the rules more complex to fix that.  I honestly would prefer it be really simple (maybe it should just split out system includes vs everything else and alphabetically sort).  I honestly don't care what the list of includes look like (so long as I don't have to manually sort them).

Issue #597, opened 2021-08-10
missing run dependencies for servo demos


tylerjw:
### Description

This is some of the output of running: `ros2 launch moveit_servo servo_teleop.launch.py`.

```
[ros2_control_node-2] [INFO] [1628569196.077672859] [controller_manager]: Loading controller 'joint_state_controller'
[ros2_control_node-2] [ERROR] [1628569196.077743017] [controller_manager]: Available classes:
[ros2_control_node-2] [ERROR] [1628569196.077755482] [controller_manager]: controller_manager/test_controller
[ros2_control_node-2] [ERROR] [1628569196.077762294] [controller_manager]: controller_manager/test_controller_failed_init
[ros2_control_node-2] [ERROR] [1628569196.077768948] [controller_manager]: controller_manager/test_controller_with_interfaces
[ros2_control_node-2] [ERROR] [1628569196.077775506] [controller_manager]: Loader for controller 'joint_state_controller' not found.
[ros2_control_node-2] [INFO] [1628569196.083868697] [controller_manager]: Loading controller 'panda_arm_controller'
[ros2_control_node-2] [ERROR] [1628569196.083905842] [controller_manager]: Available classes:
[ros2_control_node-2] [ERROR] [1628569196.083915106] [controller_manager]: controller_manager/test_controller
[ros2_control_node-2] [ERROR] [1628569196.083921690] [controller_manager]: controller_manager/test_controller_failed_init
[ros2_control_node-2] [ERROR] [1628569196.083927818] [controller_manager]: controller_manager/test_controller_with_interfaces
[ros2_control_node-2] [ERROR] [1628569196.083934001] [controller_manager]: Loader for controller 'panda_arm_controller' not found.
[ERROR] [ros2 run controller_manager spawner.py joint_state_controller-5]: process has died [pid 135003, exit code 1, cmd 'ros2 run controller_manager spawner.py joint_state_controller'].
[ERROR] [ros2 run controller_manager spawner.py panda_arm_controller-4]: process has died [pid 135001, exit code 1, cmd 'ros2 run controller_manager spawner.py panda_arm_controller'].
```

What are the run dependencies we are missing?  Have these been released for rolling/galactic?

tylerjw:
I just realized that `servo_teleop.launch.py` is a file only I have locally :facepalm: it is a copy of the file from the tutorial for my own testing.

AndyZe:
[...]

PR #602, opened 2021-08-11
Add missing dependency to servo

tylerjw:
### Description

This is part of the answer to https://github.com/ros-planning/moveit2/issues/597

I still don't know where `panda_arm_controller` and the teleop demo doesn't seem to work (frame is wrong in rviz and it prints a ton of errors about collisions).  I am testing on rolling.


vatanaksoytezer (approved):
I've also had this error in my servo test PR in #603, I can confirm that this fixes the CI and my local runs.

AdamPettinger:
I believe the Panda controller is from [here](https://github.com/ros-planning/moveit_resources/blob/ros2/panda_moveit_config/config/panda_controllers.yaml), and loaded into the Servo demos like [this](https://github.com/ros-planning/moveit2_tutorials/blob/84af78cff8f63cc9d67e39df1df2e995310c068d/doc/realtime_servo/launch/servo_teleop.launch.py#L70-L75)

> teleop demo doesn't seem to work (frame is wrong in rviz and it prints a ton of errors about collisions)

I am trying to replicate. Just want to verify you mean the [servo_teleop](https://github.com/ros-/planning/moveit2_tutorials/blob/main/doc/realtime_servo/launch/servo_teleop.launch.py) demo over in `moveit2_tutorials`, yes?

AdamPettinger (approved):
I found an issue with the TF static publisher frame names and fixed it in https://github.com/ros-planning/moveit2_tutorials/pull/104. Will you take a look and try it out? Otherwise this dependency addition worked

henningkayser (commented):
The packages inside `moveit_demo_nodes` and `moveit_ros_planning_interface` are missing this one as well. Could you add it to them?

tylerjw:
closing as another change fixed this

vatanaksoytezer:
> The packages inside `moveit_demo_nodes` and `moveit_ros_planning_interface` are missing this one as well. Could you add it to them?

@tylerjw isn't this still valid?

PR #604, opened 2021-08-11
[main] Fix missing include in servo example

vatanaksoytezer:
This error also exists in Foxy, and should be backported to Foxy.


vatanaksoytezer:
> @vatanaksoytezer so the demo just did not work previous to this fix? Does it work now with this fix?

It never launched before, throwing an undefined function error. I've tested this in my local rolling and worked fine.

tylerjw (approved):

AdamPettinger (approved):
Tested before/after change and works as expected after

PR #605, opened 2021-08-11
Document the difference between Servo pause/unpause and start/stop

AndyZe:
Some engineers on our team got confused on the difference between Start/Stop and Pause/Unpause, and it's easy to see why. It wasn't documented at all.


JafarAbdi (approved):

vatanaksoytezer (approved):
Thanks a lot for the clarification looks much better

tylerjw (approved):
Thank you for clarifying this!

PR #611, opened 2021-08-13
Check for nullptr on getGlobalLinkTransform

AdamPettinger:
### Description
Checks for `nullptr` when looking up a link transformation. The `LinkModel*` is set as `nullptr` [here](https://github.com/ros-planning/moveit2/blob/bf6357e895e364fa055da218ad7fcbd9ec3591f2/moveit_core/robot_model/src/robot_model.cpp#L1256-L1280) if the link name isn't known. This PR makes it throw in this case, instead of Seg Faulting. This should probably be backported to Moveit1 as well

I found this by modifying [this line](https://github.com/ros-planning/moveit2_tutorials/blob/9b1e55f2d665d12742388588f9b0ab2d1b9e0cb0/doc/realtime_servo/src/servo_keyboard_input.cpp#L74) to be an unknown frame, then running the [servo teleop demo](http://moveit2_tutorials.picknik.ai/doc/realtime_servo/realtime_servo_tutorial.html) with keyboard (arrow keys) input

### Checklist
- [X] Extend the tutorials / documentation [reference](http://moveit.ros.org/documentation/contributing/)

### DIFF
--- a/moveit_core/robot_state/include/moveit/robot_state/robot_state.h
+++ b/moveit_core/robot_state/include/moveit/robot_state/robot_state.h
@@ -1291,7 +1291,7 @@ public:
    *   A related, more comprehensive function is |getFrameTransform|, which additionally to link frames
    *   also searches for attached object frames and their subframes.
    *
-   *   This will throw a std::runtime_error if the passed link is not found
+   *   This will throw an exception if the passed link is not found
    *
    *  The returned transformation is always a valid isometry.
    */
--- a/moveit_core/robot_state/include/moveit/robot_state/robot_state.h
+++ b/moveit_core/robot_state/include/moveit/robot_state/robot_state.h
@@ -1291,6 +1291,8 @@ public:
    *   A related, more comprehensive function is |getFrameTransform|, which additionally to link frames
    *   also searches for attached object frames and their subframes.
    *
+   *   This will throw a std::runtime_error if the passed link is not found
+   *
    *  The returned transformation is always a valid isometry.
    */
   const Eigen::Isometry3d& getGlobalLinkTransform(const std::string& link_name)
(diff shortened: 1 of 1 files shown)


tylerjw (approved):
Thank you for fixing this.  At some point we should add a test for this.

AndyZe:
Good catch!

Are you sure you want to throw a std::runtime_error, though?

I don't see that done anywhere else in robot_state.h or robot_state.cpp. I think it should be like this, instead:

`throw Exception("Invalid link");`

To match what's done [here](https://github.com/ros-planning/moveit2/blob/de0946d02aed6a4a49a88c17a184c48a0238d2f3/moveit_core/robot_state/src/robot_state.cpp#L1253).

AdamPettinger:
> Good catch!
> 
> Are you sure you want to throw a std::runtime_error, though?
> 
> I don't see that done anywhere else in robot_state.h or robot_state.cpp. I think it should be like this, instead:
> 
> `throw Exception("Invalid link");`
> 
> To match what's done [here](https://github.com/ros-planning/moveit2/blob/de0946d02aed6a4a49a88c17a184c48a0238d2f3/moveit_core/robot_state/src/robot_state.cpp#L1253).

Seems good to me, switched over to that

moveit_core/robot_state/include/moveit/robot_state/robot_state.h
```diff
@@ -1291,6 +1291,8 @@ class RobotState
    *   A related, more comprehensive function is |getFrameTransform|, which additionally to link frames
    *   also searches for attached object frames and their subframes.
    *
+   *   This will throw a std::runtime_error if the passed link is not found
```
AndyZe, line 1294:
```suggestion
   *   This will throw an exception if the passed link is not found
```

AndyZe (approved):

tylerjw (approved):

PR #603, opened 2021-08-11
Enable Servo Tests

vatanaksoytezer:
With the recent fixes on Rolling and Galactic's `ComposobleNodeContainer`, we should be able to enable these tests safely. Closes #397.

PR #612, opened 2021-08-14
Servo: Check frames are known before getting their TFs

AdamPettinger:
### Description
Noticed this when I was looking at #611 originally. Servo should check the frames it needs to do TF math with exist in the robot state before trying to look them up. 

Can test this before/after by: modifying [this line](https://github.com/ros-planning/moveit2_tutorials/blob/9b1e55f2d665d12742388588f9b0ab2d1b9e0cb0/doc/realtime_servo/src/servo_keyboard_input.cpp#L74) to be an unknown frame, then running the[ servo teleop demo](http://moveit2_tutorials.picknik.ai/doc/realtime_servo/realtime_servo_tutorial.html) with keyboard (arrow keys) input. After this change, Servo will not crash if the command frame is invalid, but will instead print a warning and skip the command.

Note that this might break the test on `checkValidCommand()` when it is re-enabled with #603. If that is merged first, I will rebase and make sure the test passes (and add a test case for this invalid frame)

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/src/servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/src/servo_calcs.cpp
@@ -246,6 +246,19 @@ void ServoCalcs::start()
   last_sent_command_ = std::move(initial_joint_trajectory);
 
   current_state_ = planning_scene_monitor_->getStateMonitor()->getCurrentState();
+
+  // Check that all links are known to the robot
+  auto check_link_is_known = [this](const std::string& frame_name) {
+    if (!current_state_->knowsFrameTransform(frame_name))
+    {
+      std::string error_string = "Unknown frame: " + frame_name;
+      throw std::runtime_error(error_string);
+    }
+  };
+  check_link_is_known(parameters_->planning_frame);
+  check_link_is_known(parameters_->ee_frame_name);
+  check_link_is_known(robot_link_command_frame_);
+
   tf_moveit_to_ee_frame_ = current_state_->getGlobalLinkTransform(parameters_->planning_frame).inverse() *
                            current_state_->getGlobalLinkTransform(parameters_->ee_frame_name);
   tf_moveit_to_robot_cmd_frame_ = current_state_->getGlobalLinkTransform(parameters_->planning_frame).inverse() *
@@ -532,6 +545,16 @@ rcl_interfaces::msg::SetParametersResult ServoCalcs::robotLinkCommandFrameCallba
 {
   const std::lock_guard<std::mutex> lock(main_loop_mutex_);
   rcl_interfaces::msg::SetParametersResult result;
+
+  // Check that the new frame is known
+  if (!current_state_->knowsFrameTransform(parameter.as_string()))
+  {
+    RCLCPP_ERROR_STREAM(LOGGER, "Failed to change robot_link_command_frame. Passed frame '" << parameter.as_string()
+                                                                                            << "' is unknown");
+    result.successful = false;
+    return result;
+  }
+
   result.successful = true;
   robot_link_command_frame_ = parameter.as_string();
   RCLCPP_INFO_STREAM(LOGGER, "robot_link_command_frame changed to: " + robot_link_command_frame_);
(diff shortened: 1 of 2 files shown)


henningkayser (commented):
This seems reasonable to me. Out of scope: Wouldn't it make sense to throw the exception from RobotState instead of simply printing an error message and returning an identity transform?

henningkayser (approved):

tylerjw (changes requested):
This is now causing some of the servo tests to fail.  Please update the tests to make sure they are either not causing this to happen or test that this error handling works.

AndyZe (commented):
Greatly support the idea behind this PR! Will approve when CI passes.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -210,6 +210,19 @@ void ServoCalcs::start()
   last_sent_command_ = std::move(initial_joint_trajectory);
 
   current_state_ = planning_scene_monitor_->getStateMonitor()->getCurrentState();
+
+  // Check that all links are known to the robot
+  auto check_link_is_known = [this](const std::string& frame_name) {
+    if (!current_state_->knowsFrameTransform(frame_name))
+    {
+      std::string error_string = "Unknown frame: " + frame_name;
+      throw std::runtime_error(error_string);
```
jliukkonen, line 219: Can be simplified: `throw std::runtime_error{"Unknown frame: " + frame_name};` Also, I take it that changing the frames on the fly might be a possibility? Still, are there any alternatives to simple terminating a process if a frame cannot be found? Is it documented anywhere that `start()` can suddenly throw an exception?

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -918,6 +941,15 @@ bool ServoCalcs::checkValidCommand(const geometry_msgs::msg::TwistStamped& cmd)
     }
   }
 
+  // Check that the command frame is known
+  if (!current_state_->knowsFrameTransform(cmd.header.frame_id))
+  {
+    rclcpp::Clock& clock = *node_->get_clock();
+    RCLCPP_WARN_STREAM_THROTTLE(LOGGER, clock, ROS_LOG_THROTTLE_PERIOD,
```
jliukkonen, line 948: Nit: this too can be simplified by rewriting: `RCLCPP_WARN_STREAM_THROTTLE(LOGGER, *node_->get_clock(), ROS_LOG_THROTTLE_PERIOD, etc...`

AdamPettinger: This actually does not work. I don't understand exactly how ROS 2 clocks and timing works, but this has weird results:

It will compile, but some timestamps (seen in RCLCPP logging macros) turn into... not quite `nullptr` ? It is very strange

moveit_ros/moveit_servo/src/servo_calcs.cpp
AdamPettinger: There is a callback to change the robot command frame on the fly later. 

I added a note about the exception in the header file. However, we might consider a change so `start()` returns a bool and we can `return false` instead of throwing?

AdamPettinger:
> This is now causing some of the servo tests to fail. Please update the tests to make sure they are either not causing this to happen or test that this error handling works.

@tylerjw This change was indeed causing the tests to fail - because the original change skipped commands with an empty `frame_id`. So yay :tada: for enabled integration tests catching problems!

Fixed it, but I noticed just the servo tests (`colcon test --packages-select moveit_servo`) take 2.5 minutes on my not-wimpy machine. Haven't investigated if this PR causes that (but will), but that seems really excessive

tylerjw:
Rebased this change to see if it will pass CI. If it does, I suggest we merge this change, as it fixes a known bug. @AndyZe do you have any opinion on this?

tylerjw (approved):

AndyZe (approved):

PR #619, opened 2021-08-18
Update ccache action

vatanaksoytezer:
@tylerjw and I realised the CI was taking so much longer than before, and @tylerjw realized our ccache was not being loaded correctly. For eg: https://github.com/ros-planning/moveit2/runs/3233394484?check_suite_focus=true took 1 hour and has a failed `ccache cache` step. For reference https://github.com/ros-planning/moveit2/runs/3206185798?check_suite_focus=true is a normal run with correct `ccache cache` step.

This PR is an attempt to understand / fix our ccaches.

If works, this should be backported to Foxy, and all ros-planning repositories using this action including moveit.

### DIFF
--- a/.github/workflows/build_and_test_galactic.yaml
+++ b/.github/workflows/build_and_test_galactic.yaml
@@ -35,7 +35,7 @@ jobs:
       - uses: actions/checkout@v2
       # The released versions of all upstream repos are current
       - name: cache upstream_ws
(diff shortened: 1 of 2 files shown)


tylerjw (approved):
This worked... thank you for looking out and fixing this @vatanaksoytezer

henningkayser (approved):

PR #620, opened 2021-08-18
WIP: Enforce singularity threshold behavior even when moving away from a singularity

nbbrooks:
### Description

Currently the singularity threshold behavior is only enforced if the manipulator is moving further into singularity. If a different mode of operation is used (e.x. joint servoing) which puts the manipulator into/beyond the singularity region and then the user switches to Cartesian servoing, the velocity scaling/singularity halting behavior will not take effect if they are servoing away from the singularity. This can result in wild motion.

- Enforce singularity behavior regardless of whether they are moving into/away from singularity
- Publish singularity condition to topic servo_server/condition
- Add status code to distinguish between velocity scaling when moving towards/away from the singularity

This is WIP and may not be the final approach
- Should we allow the arm to move a bit faster (2x? Something less esoteric?) when inside the 2 thresholds but moving away from the singularity? 
- Should we add some way to servo very slowly out of a singularity when you are beyond the hard stop threshold? Right now that would require using joint servoing

### Checklist


AndyZe:
I like these things. Both very useful for debugging:

> Publish singularity condition to topic servo_server/condition
> Add status code to distinguish between velocity scaling when moving towards/away from the singularity

I don't like this. Several people have complained in the past when the robot could only move away from singularity slowly:

> Enforce singularity behavior regardless of whether they are moving into/away from singularity

Instead of that, I like the idea you brought up privately:

> Add a 3rd threshold that is inbetween the current two. This will be the "approaching_singularity_hard_stop". It won't let you move past that if you are moving into the singularity, but will let's you move away

AndyZe:
If you want to reduce the scope of this PR so it only does this two things, I'd approve:

> Publish singularity condition to topic servo_server/condition
> Add status code to distinguish between velocity scaling when moving towards/away from the singularity

nbbrooks:
@AndyZe I moved the singularity topic publishing to separate PR https://github.com/ros-planning/moveit2/pull/695

I did not include the new status code for servoing away from singularity to that PR, as currently we don't enforce singularity when moving away so that status code would never be used.

moveit_ros/moveit_servo/test/test_servo_singularity.cpp
```diff
@@ -51,8 +51,8 @@ TEST_F(ServoFixture, ReachSingular)
   ASSERT_TRUE(start());
   EXPECT_EQ(latest_status_, moveit_servo::StatusCode::NO_WARNING);
 
-  // Look for DECELERATE_FOR_SINGULARITY status
-  watchForStatus(moveit_servo::StatusCode::DECELERATE_FOR_SINGULARITY);
```
tylerjw, line 55: Would you mind adding a test that would check for the approach and hard stop conditions also?

nbbrooks: Lines 109 and lines 130

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -791,29 +793,29 @@ double ServoCalcs::velocityScalingFactorForSingularity(const Eigen::VectorXd& co
 
   // If this dot product is positive, we're moving toward singularity ==> decelerate
   double dot = vector_toward_singularity.dot(commanded_velocity);
-  if (dot > 0)
+  // When condition > lower_singularity_threshold and moving closer to singularity, scale between
+  // lower_singularity_threshold and approaching_stop_singularity_threshold
+  // When condition > lower_singularity_threshold and moving away from to singularity, scale more conservatively between
```
nbbrooks, line 798:
```suggestion
  // When condition > lower_singularity_threshold and moving away from singularity, scale more conservatively between
```

moveit_ros/moveit_servo/config/panda_simulated_config.yaml
```diff
@@ -53,6 +53,7 @@ num_outgoing_halt_msgs_to_publish: 4
 
 ## Configure handling of singularities and joint limits
 lower_singularity_threshold:  17.0  # Start decelerating when the condition number hits this (close to singularity)
+approaching_stop_singularity_threshold: 23.5 # Stop when the condition number hits this and we are continuing to approach singularity
 hard_stop_singularity_threshold: 30.0 # Stop when the condition number hits this
```
AndyZe, line 57:
```suggestion
hard_stop_singularity_threshold: 30.0 # Stop when the condition number hits this and we are moving away from singularity
```

moveit_ros/moveit_servo/config/panda_simulated_config_pose_tracking.yaml
```diff
@@ -46,6 +46,7 @@ num_outgoing_halt_msgs_to_publish: 4
 
 ## Configure handling of singularities and joint limits
 lower_singularity_threshold:  17.0  # Start decelerating when the condition number hits this (close to singularity)
+approaching_stop_singularity_threshold: 23.5 # Stop when the condition number hits this and we are continuing to approach singularity
 hard_stop_singularity_threshold: 30.0 # Stop when the condition number hits this
```
AndyZe, line 50:
```suggestion
hard_stop_singularity_threshold: 30.0 # Stop when the condition number hits this and we are moving away from singularity
```

moveit_ros/moveit_servo/config/panda_simulated_config.yaml
AndyZe: My understanding is, `hard_stop_singularity_threshold` should always be greater than `approaching _stop_singularity_threshold` or this strategy doesn't make sense. I'll make a suggestion below to verify that when the parameters are loaded

moveit_ros/moveit_servo/src/servo_parameters.cpp
```diff
@@ -230,6 +230,13 @@ void ServoParameters::declare(const std::string& ns,
   node_parameters->declare_parameter(
       ns + ".hard_stop_singularity_threshold", ParameterValue{ parameters.hard_stop_singularity_threshold },
       ParameterDescriptorBuilder{}.type(PARAMETER_DOUBLE).description("Stop when the condition number hits this"));
+  // Default approaching_stop_singularity_threshold to hard_stop_singularity_threshold
+  node_parameters->declare_parameter(
+      ns + ".approaching_stop_singularity_threshold",
+      ParameterValue{ parameters.approaching_stop_singularity_threshold },
+      ParameterDescriptorBuilder{}
+          .type(PARAMETER_DOUBLE)
+          .description("Stop when the condition number hits this and we are continuing to approach singularity"));
```
AndyZe, line 239: @henrygerardmoore  can you add this simple check around L355. I think you'll see what I mean:

```
if (parameters.hard_stop_singularity_threshold <= parameters.approaching_stop_singularity_threshold)
{
  ...
}
```

Because `hard_stop_singularity_threshold` should always be > `approaching_stop_singularity_threshold`

AndyZe (commented):
Here's a code review. I'll test it now and give a comment on that goes.

AndyZe:
The smoke test looks good!  :+1:

henrygerardmoore:
Thank you for the review @AndyZe, just made your changes!

AndyZe (approved):

nbbrooks:
I'd like to hold off merging this in, I had a discussion with Tyler about making the params less confusing and making the behavior "opt in"

> params less confusing
* Add 'leaving_singularity_threshold_multiplier' - rather than have people maintain the 'approaching_stop_singularity_threshold' and 'hard_stop_singularity_threshold'

> making the behavior "opt in"
* If 'leaving_singularity_threshold_multiplier' is not set, it uses the old behavior, which doesnt enforce slowdown when leaving. Warns the user.

nbbrooks:
> That sounds fine. I'll admit, the parameter names were a little confusing (because you were preserving backward compatability)

Totally agree. Here is my proposal for the new behavior in a graphic we can clean up and use in documentation. If people like it I'll update PR description
servo-singularity-threshold-behavior

henrygerardmoore:
@nbbrooks @AndyZe Just updated to the latest design, take a look when you have a chance and let me know if I should change anything (especially with the style of the deprecation warning and how I checked for the parameter's existence before getting it). 
Servo tests still pass locally, but should we add a test to make sure the scaling is working as expected? Mathematically it makes sense to me, I was just wondering if that might be warranted.
Also, is linking to a PR in a comment bad style? I could maybe link directly to the diagram or just try to write an explanation if that would be better.

moveit_ros/moveit_servo/config/panda_simulated_config.yaml
```diff
@@ -53,7 +53,8 @@ num_outgoing_halt_msgs_to_publish: 4
 
 ## Configure handling of singularities and joint limits
 lower_singularity_threshold:  17.0  # Start decelerating when the condition number hits this (close to singularity)
-hard_stop_singularity_threshold: 30.0 # Stop when the condition number hits this
+approaching_stop_singularity_threshold: 23.5 # Stop when the condition number hits this and we are continuing to approach singularity
```
AndyZe, line 56: @henrygerardmoore can't this be deleted? I think Nathan added it earlier on in this PR. It's not in the `main` branch, not in his graphic, and doesn't seem to be used for anything in the code

moveit_ros/moveit_servo/src/servo_parameters.cpp
```diff
@@ -345,6 +373,13 @@ std::optional<ServoParameters> ServoParameters::validate(ServoParameters paramet
     RCLCPP_WARN(LOGGER, "Parameter 'num_outgoing_halt_msgs_to_publish' should be greater than zero. Check yaml file.");
     return std::nullopt;
   }
+  if (parameters.hard_stop_singularity_threshold <= parameters.approaching_stop_singularity_threshold)
+  {
+    RCLCPP_WARN(LOGGER, "Parameter 'hard_stop_singularity_threshold' "
+                        "should be greater than 'approaching_stop_singularity_threshold.' "
+                        "Check yaml file.");
+    return std::nullopt;
+  }
```
AndyZe, line 382: This no longer applies.

```suggestion
```

moveit_ros/moveit_servo/src/servo_parameters.cpp
```diff
@@ -320,6 +334,20 @@ ServoParameters ServoParameters::get(const std::string& ns,
...
   parameters.joint_limit_margin = node_parameters->get_parameter(ns + ".joint_limit_margin").as_double();
 
+  if (node_parameters->has_parameter(ns + ".leaving_singularity_threshold_multiplier"))
+  {
+    parameters.leaving_singularity_threshold_multiplier =
+        node_parameters->get_parameter(ns + ".leaving_singularity_threshold_multiplier").as_double();
+  }
+  else
+  {
+    // if multiplier is 1.0, original behavior is preserved
```
AndyZe, line 344: This comment isn't quite true. I think we can just delete it.
```suggestion
```

moveit_ros/moveit_servo/src/servo_parameters.cpp
AndyZe: Can you replace it with a check that `parameters.leaving_singularity_threshold_multiplier > 0`?

AndyZe:
Almost there.

> Also, is [linking to a PR in a comment](https://github.com/nbbrooks/moveit2/blob/179b4b0d18165a0ac228a905c520d6ba5d3badae/moveit_ros/moveit_servo/src/servo_calcs.cpp#L910) bad style?

IMO that's fine

henrygerardmoore:
Thanks for the review, Andy! I think it should be good to go now. Made your changes and cleaned some stuff up as well as adding defaults for the panda `.yaml` files.

AndyZe (approved):

moveit_ros/moveit_servo/src/servo_parameters.cpp
```diff
@@ -320,6 +327,19 @@ ServoParameters ServoParameters::get(const std::string& ns,
...
+  {
+    parameters.leaving_singularity_threshold_multiplier =
+        node_parameters->get_parameter(ns + ".leaving_singularity_threshold_multiplier").as_double();
+  }
+  else
+  {
+    parameters.leaving_singularity_threshold_multiplier = 1.0;
+    RCLCPP_WARN(LOGGER, "Using default servo singularity handling; add parameter "
+                        "'leaving_singularity_threshold_multiplier' to define leaving singularity threshold. See "
+                        "https://github.com/ros-planning/moveit2/pull/620 for more information.");
```
tylerjw, line 340: Should this warning have the word `deprecated` in it to let users know that we might remove this behavior in the future?

tylerjw (approved):
I like the approach of keeping old configs working and using a warning to inform them of new/better behavior they can take advantage of through a config change.

I would like to see logic moved out of sevo_calcs and into free functions that can be tested independently of all of the state in servo_calcs but will not block this PR for that.

henrygerardmoore:
Is one of the new tests flaky? 52a77b7 failed CI even though the only change was the warning text :sweat:

AndyZe:
See https://github.com/ros-planning/moveit2/issues/1499

Feel free to debug that if you have time. That would be awesome.

PR #621, opened 2021-08-18
Add standalone executable for Servo server node

schornakj:
### Description

Previously the servo_server node was only available as a component node. This adds a new executable to run the servo_server as a standalone node.

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/launch/servo_example.launch.py
+++ b/moveit_ros/moveit_servo/launch/servo_example.launch.py
@@ -7,6 +7,7 @@ from launch.some_substitutions_type import SomeSubstitutionsType
 from launch.substitutions import LaunchConfiguration, PathJoinSubstitution
 from launch_ros.actions import ComposableNodeContainer, Node
 from launch_ros.descriptions import ComposableNode
+from launch.actions import ExecuteProcess
 from ament_index_python.packages import get_package_share_directory
 import xacro
 
@@ -39,27 +40,13 @@ def generate_launch_description():
     servo_yaml = load_yaml("moveit_servo", "config/panda_simulated_config.yaml")
     servo_params = {"moveit_servo": servo_yaml}
 
-    # Get URDF and SRDF
-    if start_position_path:
-        initial_positions_file = os.path.join(
-            os.path.dirname(__file__), start_position_path
+    robot_description_config = xacro.process_file(
+        os.path.join(
+            get_package_share_directory("moveit_resources_panda_moveit_config"),
+            "config",
+            "panda.urdf.xacro",
         )
-        robot_description_config = xacro.process_file(
-            os.path.join(
-                get_package_share_directory("moveit_resources_panda_moveit_config"),
-                "config",
-                "panda.urdf.xacro",
-            ),
-            mappings={"initial_positions_file": initial_positions_file},
-        )
-    else:
-        robot_description_config = xacro.process_file(
-            os.path.join(
-                get_package_share_directory("moveit_resources_panda_moveit_config"),
-                "config",
-                "panda.urdf.xacro",
-            )
-        )
+    )
 
     robot_description = {"robot_description": robot_description_config.toxml()}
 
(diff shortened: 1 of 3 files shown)


moveit_ros/moveit_servo/src/servo_server_node.cpp
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+#include <moveit_servo/servo_server.h>
+
+int main(int argc, char* argv[])
+{
+  rclcpp::init(argc, argv);
+
+  rclcpp::NodeOptions options;
+  options.automatically_declare_parameters_from_overrides(true);
```
tylerjw, line 47: In the future I hope we can get away from this by declaring all the parameters in a better way with nicer validation.  I'm working on that.

tylerjw (approved):
Looks really good and useful.  Thank you.  Only thing I'd add is an example launch file for it but it is so similar to the component node I'm not sure that's worth it.  I hope this is helpful in debugging the issues you've seen with component nodes.

moveit_ros/moveit_servo/src/servo_server_node.cpp
AndyZe: TBH I'm still fuzzy on the finer differences between component nodes and standalone nodes.

The PR looks good. I'll prob give it a thumbs up after testing.

AndyZe:
@schornakj do you have a launch file to test with? My simple modification of `moveit2_tutorials/docs/realtime_servo/launch/servo_teleop.launch.py` did not work. I'm guessing the parameters were loaded in the wrong namespace.

[servo_teleop.launch.py.txt](https://github.com/ros-planning/moveit2/files/7028574/servo_teleop.launch.py.txt)

schornakj:
> @schornakj do you have a launch file to test with? My simple modification of `moveit2_tutorials/docs/realtime_servo/launch/servo_teleop.launch.py` did not work. I'm guessing the parameters were loaded in the wrong namespace.
> 
> [servo_teleop.launch.py.txt](https://github.com/ros-planning/moveit2/files/7028574/servo_teleop.launch.py.txt)

I'll add a standalone launch file for this equivalent to the one we provide for the Servo component node.

henningkayser (approved):
Looks good to me. In general, would it be bad practice to add a main function to source files of component nodes directly? Say, add it to [servo_server.cpp](https://github.com/ros-planning/moveit2/blob/main/moveit_ros/moveit_servo/src/servo_server.cpp#L167) below registering the component node, or alternatively add the RCLCPP_COMPONENTS_REGISTER_NODE macro to `servo_server_node.cpp`? Also, the source file naming is really confusing, we have `servo.cpp` (main servo class), `servo_server.cpp` (node implementation) and `servo_server_node.cpp` which is the now added main function to run ServoServer as node executable. Out of scope of this PR, but we should clear that up.

tylerjw:
> Agree that the naming is starting to be confusing.

I created a PR because I feel the same way: https://github.com/ros-planning/moveit2/pull/649

AndyZe (changes requested):
blocking merge until I can test it

moveit_ros/moveit_servo/launch/servo_example.launch.py
```diff
@@ -0,0 +1,137 @@
...
+    for controller in ["panda_arm_controller", "joint_state_broadcaster"]:
+        load_controllers += [
+            ExecuteProcess(
+                cmd=["ros2 run controller_manager spawner.py {}".format(controller)],
+                shell=True,
+                output="screen",
+            )
+        ]
+
+    # Component nodes for tf and Servo
```
AndyZe:
```suggestion
    # Component nodes for tf
    # Servo can also be launched as a component but this launch file uses a standalone node
```

AndyZe:
@schornakj I updated the example launch file to:

- launch RViz
- include a commented example of launching Servo as a component, so that documentation is all in one place

I think the tutorial can now use this launch file rather than a completely separate one  :+1:

Let me know if that bugs you for some reason and we can revert the commits.

AndyZe (approved):

AndyZe (approved):

PR #629, opened 2021-08-19
CI: switch to ghcr.io docker images, simplify

rhaschke:
This is a follow-up to #594 to use docker images built via GHA.
Additionally, simplify, consolidate (rolling + galactic), and unify (with MoveIt1) CI configs.
See [here](https://github.com/ros-planning/moveit2/actions?query=branch%3Atest%2Fgha-docker++) for (hopefully succeeding) CI builds.

This should be merged w/o the last commit that just enables the builds on the test branch.
Please keep individual (because independent) commits.


moveit_core/collision_detection_fcl/src/collision_common.cpp
```diff
@@ -489,9 +489,9 @@ bool distanceCallback(fcl::CollisionObjectd* o1, fcl::CollisionObjectd* o2, void
 
   double dist_threshold = cdata->req->distance_threshold;
 
-  const std::pair<const std::string&, const std::string&> pc = cd1->getID() < cd2->getID() ?
-                                                                   std::make_pair(cd1->getID(), cd2->getID()) :
-                                                                   std::make_pair(cd2->getID(), cd1->getID());
+  const std::pair<const std::string&, const std::string&> pc =
+      cd1->getID() < cd2->getID() ? std::make_pair(std::cref(cd1->getID()), std::cref(cd2->getID())) :
+                                    std::make_pair(std::cref(cd2->getID()), std::cref(cd1->getID()));
```
vatanaksoytezer, line 494: Is this related with the CI changes?

moveit2.repos
```diff
@@ -1,28 +1,8 @@
...
-    url: https://github.com/ros-planning/srdfdom
-    version: ros2
-  ros2_control:
-    type: git
-    url: https://github.com/ros-controls/ros2_control
-    version: master
-  ros2_controllers:
-    type: git
-    url: https://github.com/ros-controls/ros2_controllers
-    version: master
```
vatanaksoytezer, line 25: We should not need the .repos files but as per idea of @tylerjw and discussion [here](https://github.com/ros-planning/moveit2/pull/565) we left them unused . If in a case some PR depends on one of these repositories. We could re-enable the .repos in CI.

.github/workflows/ci.yaml
```diff
@@ -1,47 +1,55 @@
...
       CCACHE_DIR: ${{ github.workspace }}/.ccache
       BASEDIR: ${{ github.workspace }}/.work
-      CACHE_PREFIX: ${{ matrix.env.ROS_DISTRO }}-${{ matrix.env.ROS_REPO }}
       CLANG_TIDY_BASE_REF: ${{ github.base_ref || github.ref }}
+      BEFORE_CLANG_TIDY_CHECKS: (cd $TARGET_REPO_PATH; clang-tidy --list-checks)
+      CC: ${{ matrix.env.CLANG_TIDY && 'clang' }}
+      CXX: ${{ matrix.env.CLANG_TIDY && 'clang++' }}
 
-    name: ${{ matrix.env.ROS_DISTRO }}-${{ matrix.env.ROS_REPO }}${{ contains(matrix.env.IKFAST_TEST, 'ikfast') && ' + ikfast' || ''}}${{ matrix.env.CCOV && ' + ccov' || ''}}${{ matrix.env.CLANG_TIDY && ' + clang-tidy' || '' }}
+    name: ${{ matrix.env.IMAGE }}${{ matrix.env.CCOV && ' + ccov' || ''}}${{ matrix.env.IKFAST_TEST && ' + ikfast' || ''}}${{ matrix.env.CLANG_TIDY && ' + clang-tidy' || '' }}
```
vatanaksoytezer, line 46: IKFast tests are not ported yet, so we don't need these in the naming for now.

.github/workflows/ci.yaml
```diff
@@ -1,47 +1,55 @@
...
     runs-on: ubuntu-latest
     steps:
       - name: "Free up disk space"
         if: ${{ matrix.env.CCOV }}
         run: |
-          sudo apt-get -qq purge build-essential ghc*
+          sudo apt-get -qq purge build-essential "ghc*"
```
vatanaksoytezer, line 52: Do we need the `" "`s?

.github/workflows/ci.yaml
```diff
@@ -50,41 +58,52 @@ jobs:
...
-        uses: 'ros-industrial/industrial_ci@master'
+            ${{ env.CACHE_PREFIX }}-${{ github.sha }}
+            ${{ env.CACHE_PREFIX }}
+        env:
+          CACHE_PREFIX: ccache-${{ matrix.env.IMAGE }}${{ matrix.env.CCOV && '-ccov' || '' }}
+
+      - name: generate ikfast packages
+        if: ${{ matrix.env.IKFAST_TEST }}
+        run: |
+          moveit_kinematics/test/test_ikfast_plugins.sh
```
vatanaksoytezer, line 97: Same with above, IKFast tests are not ported yet.

.github/workflows/docker.yaml
```diff
@@ -2,10 +2,9 @@ name: docker
 
 on:
   workflow_dispatch:
-  pull_request:
   push:
     branches:
-      - main
+      - test/gha-docker
```
vatanaksoytezer, line 7:
```suggestion
      - main
```

moveit2_rolling.repos
```diff
@@ -1,13 +0,0 @@
...
-    url: https://github.com/ros-planning/geometric_shapes
-    version: ros2
-  warehouse_ros:
-    type: git
-    url: https://github.com/ros-planning/warehouse_ros
-    version: ros2
-  warehouse_ros_mongo:
-    type: git
-    url: https://github.com/ros-planning/warehouse_ros_mongo
-    version: ros2
```
vatanaksoytezer, line 13: I think we should be able remove `rolling.repos` I am not %100 sure CI would be happy, but if it is then it is good to remove this.

.github/workflows/ci.yaml
```diff
@@ -1,47 +1,55 @@
...
             IKFAST_TEST: true
-            CLANG_TIDY: true
+            CLANG_TIDY: pedantic
+          - IMAGE: galactic-ci
+          - IMAGE: galactic-ci-testing
     env:
-      UPSTREAM_WORKSPACE: moveit2_rolling.repos
+      CXXFLAGS: "-Wall -Wextra -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls -Wno-deprecated-copy"
+      DOCKER_IMAGE: ghcr.io/ros-planning/moveit2:${{ matrix.env.IMAGE }}
+      UPSTREAM_WORKSPACE: moveit2.repos $(f="moveit2_$(sed 's/-.*$//' <<< "${{ matrix.env.IMAGE }}").repos"; test -r $f && echo $f)
```
vatanaksoytezer, line 30: We can comment out this if CI is happy in all distros, if not that's good, see comment below on .repos files.

tylerjw (approved):

vatanaksoytezer (commented):
@rhaschke Thanks a lot for all the reviews and helping me with the switch today. I always appreciate somebody super experienced , like yourself helping out with the reviews and PRs! I looked through the PR and suggested a few points along with some questions.

moveit_core/collision_detection_fcl/src/collision_common.cpp
rhaschke: This was required for switching to `clang` as the compiler for `clang-tidy` related jobs.

moveit2.repos
rhaschke: My idea in this PR is to list basic upstream dependencies in `moveit2.repos` and augment these with additional packages required for older distro, i.e. via `moveit2_galactic.repos`.

.github/workflows/ci.yaml
rhaschke: At least, they compile without issues ;-)

.github/workflows/ci.yaml
rhaschke: Yes, that was an important addition in MoveIt1: https://github.com/ros-planning/moveit/pull/2635

.github/workflows/docker.yaml
rhaschke: See my comment to skip the last commit when pushing ;-)

moveit2_rolling.repos
rhaschke: see above. `rolling.repos` will be removed.

PR #633, opened 2021-08-20
Cache dockers, use docker action to avoid untagged images in ghcr

vatanaksoytezer:
This adds docker caching which increases the speed of the docker builds. In a fully cached scenario (no changes in moveit source code), this takes less than 2 minutes to build all and push 8 docker images. Nevertheless we can still use some changes in our source Dockerfile to speed up caching.

Using the docker-action also don't generate (rather annoying) untagged images ghcr keeps with regular docker push. Appreciate if any maintainer can test this on the test branch before merging.

After this PR, if we are all happy with this: I will send follow up PRs to foxy branch and moveit.

For reference below is before this PR:
Screenshot from 2021-08-20 15-43-18
And after:
image


.github/workflows/docker.yaml
```diff
@@ -20,18 +20,26 @@ jobs:
...
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and Push
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          file: .docker/${{ github.job }}/Dockerfile
+          build-args: ROS_DISTRO=${{ matrix.ROS_DISTRO }}
+          pull: true
+          push: true
+          no-cache: false
```
rhaschke, line 39: What's the purpose of `pull` and `no-cache`?
I couldn't find any info about these attributes in the [docs](https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md).

rhaschke (commented):
Great improvement! Can we further simplify the config by dropping the args `pull` and `no-cache`?

.github/workflows/docker.yaml
vatanaksoytezer: It's on the main README: https://github.com/docker/build-push-action.

`pull`: always tries to pull a new image
`no-cache`: if true don't use cache. This defaults to false though, I can actually remove this, put it here for a little more verbosity on what I was doing.

rhaschke: Do you really need to pull the newest image if you are going to rebuild it anyway?
Yes, I suggest removing `no-cache`. Keep the config as simple as possible.

vatanaksoytezer:
> Great improvement! Can we further simplify the config by dropping the args `pull` and `no-cache`?

Thanks a ton for the review @rhaschke! I dropped no-cache since it defaults to false anyway. But I think we still need to keep pull: true to ensure we always use the latest image generated by CI.

rhaschke:
But we are going to rebuild this image! Why should we first fetch it? The examples in the cache doc don't use `pull` as well.

vatanaksoytezer:
> But we are going to rebuild this image! Why should we first fetch it? The examples in the cache doc don't use `pull` as well.

Thinking again, you are right. Removing it as well.

rhaschke:
> Thinking again, you are right. Removing it as well.

I'm just thinking aloud. Did you test before w/o that arg? Does it change behavior?
Could you link your GHA builds for closer inspection, please?

vatanaksoytezer:
> Did you test before w/o that arg? Does it change behavior?
> Could you link your GHA builds for closer inspection, please?

Just did a test.

Here is before I removed pull: https://github.com/vatanaksoytezer/moveit2/actions/runs/1151222115
Here is after I removed pull: https://github.com/vatanaksoytezer/moveit2/actions/runs/1151224783

For reference PR to my fork: https://github.com/vatanaksoytezer/moveit2/pull/15

Removing pull seem to break? some caching. See ci-testing build times. (Note: source is taking long time because I fetched main just before this PR, so it is building it again.)

vatanaksoytezer:
Update, now seems to work without any problems:

With pull: https://github.com/vatanaksoytezer/moveit2/actions/runs/1151353444
Without pull: https://github.com/vatanaksoytezer/moveit2/actions/runs/1151364845

.github/workflows/docker.yaml
```diff
@@ -20,18 +20,24 @@ jobs:
...
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Login to Github Containter Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and Push
+        id: docker_build
```
rhaschke, line 32: Great that it works w/o `pull` now. Probably you can also drop `id`. Sorry, that I didn't notice this before.

rhaschke (approved):
Generally approve. Please try removing `id` as well.

vatanaksoytezer:
As per to discuss https://github.com/ros-planning/moveit_task_constructor/pull/286#issuecomment-903550253 @rhaschke and I had a virtual conference. Below are the action points from this call to move on with ghcr migration. Anyone: feel free to chime in and discuss more about the migration.

- Stop docker images from autobuild in dockerhub (@rhaschke)
- Push to dockerhub and ghcr (@vatanaksoytezer)
- We would need to create a secret variable for GHA to acces dockerhub. (@henningkayser or @davetcoleman)
- Do scheduled builds weekly for apt updates. (@vatanaksoytezer)
- Clean untagged images (@vatanaksoytezer possibly via a Github action) 
- Do it for foxy and moveit (@vatanaksoytezer)

@rhaschke one thing @jspricke helped me noticed is that we need to at least build no-cache dockers either in every push to main or on scheduled intervals. The problem with caching is that event if there are upgrade-able packages via apt-get it won't update because there is a cached layer for apt-get, thus we would never be getting the latest packages.

Per this, I am updating this PR to use caching in push, but not use caching in scheduled builds. Let me know what you think.

rhaschke:
> I am updating this PR to use caching in push, but not use caching in scheduled builds.

I thought about this too. I agree to disable caching for `ci` and `ci-testing` images (`release` too?). For `source` we don't need to, as this image builds on `ci-testing`. Also, you should enable rebuilding conditionally based on available updates as done in `navigation2`.

.github/workflows/docker.yaml
```diff
@@ -28,6 +33,15 @@ jobs:
...
           password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Check apt updates
+        if: ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
+        run: |
+            apt-get update -o Dir::Etc::sourcelist="${SOURCELIST}"
+            apt-get --simulate upgrade -o Dir::Etc::sourcelist="${SOURCELIST}" \
+            | grep "^0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.$" \
+            && echo "::set-output name=trigger::false" \
+            || echo "::set-output name=trigger::true"
+            echo "::set-output name=no_cache::true"
```
rhaschke: @ruffsl, thinking deeper about this code snippet from nav2, I think it doesn't make sense: 
This code actually checks for system updates in the OS _running the job_, but not for updates _within_ the docker image.

My suggestion is to disable caching for scheduled jobs and thus fetch new updates once a week. See https://github.com/vatanaksoytezer/moveit2/pull/17

vatanaksoytezer: @rhaschke I think it actually runs on the previous CI image, see https://github.com/ros-planning/navigation2/blob/main/.github/workflows/update_ci_image.yaml#L38-L39.

rhaschke: Ah, you are right. I missed the reference to the container! However, you missed copying this from `nav2` :wink:

vatanaksoytezer: Yep you are right, I actually accidentally pushed that bit prematurely. I am working on something manageable to check the build triggers

vatanaksoytezer:
@rhaschke I was working a way of trigger weekly builds depending on whether there are available apt updates (like Nav2). I have found a few ways to do that, but adding more than 120 lines of CI (which will make the maintenance much harder)  for maybe 30-35 minutes time weekly, seemed not too worthy. I applied your suggestion and now we should be good to go, I would appreciate a last review.

rhaschke (approved):

ruffsl:
> But we are going to rebuild this image! Why should we first fetch it? The examples in the cache doc don't use pull as well.

The pull arg helps insure all the base/parent images your CI image builds upon will be pulled from the image registry, and thus up-to-date. If you know your base image already exists on the CI instance, say you already pulled it in a CI step prior, or intend to use a custom tag, etc then you could skip it.

PR #637, opened 2021-08-23
Throw instead of returning nullptr from RobotState and RobotModel

tylerjw:
### Description
This was started by realizing that servo's constructor will segfault if you specify an invalid joint model group name because getJointModelGroup is called and then the result is dereferenced.  We started throwing exceptions from these classes already when invalid input is given.  I think this is a much nicer solution than returning nullptr and relying on the user to check everywhere.  Instead, by throwing an exception you make it really clear why the program exited instead of a segfault.

### Checklist


AndyZe (approved):

AndyZe:
On second thought, wouldn't it be better to put this print inside `RobotState::getJointModelGroup()`. Then it will apply anywhere in the codebase

tylerjw:
> On second thought, wouldn't it be better to put this print inside `RobotState::getJointModelGroup()`. Then it will apply anywhere in the codebase

That is probably better.  I can put the printing of the more useful error there too.

tylerjw:
Note that I started checking parameters in one function and realized that the problem of potentially dereferencing nullptr is huge in this class.  I can remove that part of this fix it would be better to make a whole separate PR for checking input parameters.

I think throwing exceptions are generally much nicer than segfaults but there are other things we could do instead like introduce `StatusOr` from absl: `https://abseil.io/docs/cpp/guides/status`.  That solution would be much nicer if we want to clean this up to be more friendly to people building real-time systems where they want to compile without exceptions.  In the meantime (until we can decide if we are going to use something like StatusOr) I think this is s useful improvement.

tylerjw:
Another interesting thing to think about when looking at robot_model is how all of these are non-owning raw pointers.  The current pattern many users have (including us in places like servo) is to keep the raw pointer to the joint model group as a member variable.  The downside to that is the lifetime of the jmg is coupled to the lifetime of that pointer.  That is generally fine because the RobotModel is created during initialization and has the lifetime of the process with the PSM.  If we want to change the robot model representation during run-time we are going to have to think about the lifetime of this state.

AndyZe:
My opinion is, returning a nullptr may actually be useful in some cases. I think the warning should be printed but no error should be thrown.

Example:

Say you have a bunch of robots with two possible configurations, `Gripper` and `No Gripper`. You might want to use getEndEffector() to test what the configuration is for each robot.

I strongly support adding the prints to help with debugging, though.

tylerjw:
> Say you have a bunch of robots with two possible configurations, `Gripper` and `No Gripper`. You might want to use getEndEffector() to test what the configuration is for each robot.

In that case you could just use `RobotModel::hasEndEffector` to see what configuration you have.  The difference is it returns a bool for when it exists and `RobotModel::getEndEffector` returns a raw pointer to a JointModelGroup.  If you want to use `getEndEffector` to test for the existence and get it you can always use a try-catch block.

The other API option (StatusOr) I'm going to try today so we can debate which approach we like better.

henningkayser (commented):
I let this get stale somehow. @tylerjw What's the latest status? Do we still consider throwing exceptions or should we better go for something like #644 or [std::optional](https://en.cppreference.com/w/cpp/utility/optional)? I think I'd prefer an API change over exceptions so should we close this?

tylerjw:
We can close this.  I too would prefer a API change.  I see StatusOr and std::optional as the best options we have.  I created a test PR to show what StatusOr would be like.  I generally prefer the ergonomics of it but haven't used std::optional more than just reading the docs on it.  If you would mind putting together a test PR for it similar to mine for StatusOr we could discuss which we'd prefer.

As to the issue of yet another external dependency.  There will be a debian package for absl in 22.04, that means for Humble we can depend on that just like we depend on boost.  For rolling though in the mean time if we want to use absl we'll need to figure out how to either get the debian package for absl released into 20.04 or create a ros package wrapper for it (what I tried and have yet to succeed at).

PR #639, opened 2021-08-24
WIP: Remove const SharedPtrs, get main building

nbbrooks:
### Description

Something weird is going on with CI. When I build MoveIt using the latest ros:rolling-ros-base-focal base image I get a lot of complaints about messages and `note:   no known conversion for argument 1 from 'const`. I don't know if docker is caching a previous image, which will prevent it from pulling in an updated debian. It could be the case because on the base image, the newest timestamps on my rolling debians are ~Jul 16, whereas after I upgrade they are ~Aug 11.

This should replace all the const SharedPtr usages
`find . -type f -exec sed -i -E 's/([[:space:]\(])(const )([^,]*[:_]msg::.*SharedPtr)/\1\3/g' {} +`

Currently I'm attempting to get a successful build with source checkouts of moveit2 and the following message packages:
- control_msgs
- moveit_msgs
- octomap_msgs

It looks like there is an issue with Boost and sensor_msgs::msg::JointState still so something isn't quite right.

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/include/moveit_servo/collision_check.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/collision_check.h
@@ -85,7 +85,7 @@ private:
   planning_scene_monitor::LockedPlanningSceneRO getLockedPlanningSceneRO() const;
 
   /** \brief Callback for collision stopping time, from the thread that is aware of velocity and acceleration */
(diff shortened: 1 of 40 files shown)


jspricke:
I think this is wrong, will try to debug with Nathan once he is awake again.

PR #641, opened 2021-08-25
Add cleaning action

vatanaksoytezer:
The reasoning for this action is just to prune all the untagged images, Github saves when uploading a container with the same name. As of I am writing this we already have 100 untagged images. This will delete all the untagged images in every push. Requires a token that access to delete packages permission. 
image


tylerjw (approved):
This looks good, it just needs the token.

PR #649, opened 2021-08-30
Rename ServoServer to ServoNode

tylerjw:
### Description

Until we figure out how we want to change the API to prevent this sort of bug this makes servo error much nicer in the constructor of ServoCalcs.

This also includes a fix for a clang-tidy error I hadn't seen before.  I wonder if our clang-tidy changed since the last time someone made a code change to servo.

Updated tutorials: https://github.com/ros-planning/moveit2_tutorials/pull/109

### Checklist
- [x] Make PR to update tutorials.
- [x] Migration note

### DIFF
--- a/moveit_ros/moveit_servo/CMakeLists.txt
+++ b/moveit_ros/moveit_servo/CMakeLists.txt
@@ -16,11 +16,11 @@ set(SERVO_PARAM_LIB_NAME ${SERVO_LIB_NAME}_parameters)
 set(POSE_TRACKING pose_tracking)
 
 # Component Nodes (Shared libraries) ############################
-set(SERVO_COMPONENT_NODE servo_server)
+set(SERVO_COMPONENT_NODE servo_node)
 set(SERVO_CONTROLLER_INPUT servo_controller_input)
 
 # Executable Nodes ##############################################
-set(SERVO_SERVER_NODE_NAME servo_server_node)
+set(SERVO_NODE_MAIN_NAME servo_node_main)
 set(POSE_TRACKING_DEMO_NAME servo_pose_tracking_demo)
 set(FAKE_SERVO_CMDS_NAME fake_command_publisher)
 
@@ -96,10 +96,10 @@ target_link_libraries(${POSE_TRACKING} ${SERVO_LIB_NAME})
 #####################
 
 # Add and export library to run as a ROS node component, and receive commands via topics
-add_library(${SERVO_COMPONENT_NODE} SHARED src/servo_server.cpp)
+add_library(${SERVO_COMPONENT_NODE} SHARED src/servo_node.cpp)
 ament_target_dependencies(${SERVO_COMPONENT_NODE} ${THIS_PACKAGE_INCLUDE_DEPENDS})
 target_link_libraries(${SERVO_COMPONENT_NODE} ${SERVO_LIB_NAME})
-rclcpp_components_register_nodes(${SERVO_COMPONENT_NODE} "moveit_servo::ServoServer")
+rclcpp_components_register_nodes(${SERVO_COMPONENT_NODE} "moveit_servo::ServoNode")
 
 # Add executable for using a controller
 add_library(${SERVO_CONTROLLER_INPUT} SHARED src/teleop_demo/joystick_servo_example.cpp)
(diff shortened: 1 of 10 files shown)


moveit_ros/moveit_servo/src/teleop_demo/joystick_servo_example.cpp
```diff
@@ -49,8 +49,8 @@
 
 // We'll just set up parameters here
 const std::string JOY_TOPIC = "/joy";
-const std::string TWIST_TOPIC = "/servo_server/delta_twist_cmds";
-const std::string JOINT_TOPIC = "/servo_server/delta_joint_cmds";
+const std::string TWIST_TOPIC = "/servo_component/delta_twist_cmds";
```
henningkayser, line 52: Should we just use the `/servo/*` namespace?

moveit_ros/moveit_servo/include/moveit_servo/servo_component.h
```diff
@@ -45,10 +45,10 @@
 
 namespace moveit_servo
 {
-class ServoServer : public rclcpp::Node
+class ServoComponent : public rclcpp::Node
```
henningkayser, line 48: What are the implications of calling this class Component? Isn't it a node? So `ServoNode` makes most sense to me, which in turn can be a component node (exported, no class/file needed) or a node executable `ServoNodeExecutable`.

moveit_ros/moveit_servo/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@ set(SERVO_PARAM_LIB_NAME ${SERVO_LIB_NAME}_parameters)
 set(POSE_TRACKING pose_tracking)
 
 # Component Nodes (Shared libraries) ############################
-set(SERVO_COMPONENT_NODE servo_server)
+set(SERVO_COMPONENT_NODE servo_component)
```
AndyZe, line 19: I like this name a lot more

moveit_ros/moveit_servo/include/moveit_servo/servo_component.h
AndyZe: Changes here should be reviewed along with #621 (Add standalone executable for Servo server node). In fact, maybe this PR should be closed and we just make all of the naming changes as a separate commit in that PR.

I think `ServoNodeComponent` (class name) and `servo_component_node.cpp` make good sense for this class / file.

The executable that Joe added in #621 could be called `ServoMain` or `ServoStandaloneNode`. It just instantiates this class.

moveit_ros/moveit_servo/src/teleop_demo/joystick_servo_example.cpp
tylerjw: agree that probably makes sense or /moveit_servo ?

AndyZe: :+1: to /servo/*

moveit_ros/moveit_servo/include/moveit_servo/servo_component.h
schornakj: +1 to `ServoNode`. In my mind we create a node and then allow launching it in a component container.

tylerjw:
I renamed servo_component to servo_node and servo_server_node to servo_node_main.

henningkayser (approved):

henningkayser:
@tylerjw one thing before merging. Could you provide deprecated headers and typedefs for the old API?

tylerjw:
@AndyZe I believe this is ready for review.  I included a deprecated header so existing code should continue to build and produce a warning.  There is no way for me to deprecate the change in the name for the binaries (the component node plugin) so launch files that depend on the old ServoServer node name will stop working after this.

moveit_ros/moveit_servo/include/moveit_servo/servo_server.h
```diff
@@ -40,49 +40,12 @@
 
 #pragma once
 
-#include <moveit_servo/servo.h>
-#include <std_srvs/srv/trigger.hpp>
+#pragma message("Header `servo_server.h` is deprecated, please use `servo_node.h`")
```
JafarAbdi, line 43: Should we add a date for when this header and the alias will be removed, I like how drake does it https://github.com/RobotLocomotion/drake/blob/master/common/drake_deprecated.h#L59

JafarAbdi (approved):

moveit_ros/moveit_servo/include/moveit_servo/servo_server.h
henningkayser: I don't this is necessary. We'll remove it with the next major (=API breaing) release. No need to set a date for it

AndyZe (approved):
A bit of text still needs to be replaced. Approved assuming that happens.

```
moveit_ros/moveit_servo/launch/servo_example.launch.py:            #     name="servo_server",
moveit_ros/moveit_servo/launch/servo_example.launch.py:        executable="servo_server_node",
```

moveit_ros/moveit_servo/include/moveit_servo/servo_server.h
tylerjw: I think we should just plan on deleting all the deprecated APIs right before the next major release, Humble.  We can of course do it sooner if we think that is necessary for some other reason.  main/Rolling are not guaranteed to be API stable but we should attempt to not break it often if we can.

tylerjw:
> It looks like this PR along with some of the other recent changes (e.g. launch file shuffling) have broken the Servo tutorial. I'd like to put the brakes on more large Servo changes until the tutorial gets fixed.

The tutorial on the website or the tutorial in the main branch of the tutorial repo?  The website has changed to track a new foxy branch.  That website is based on a foxy install of moveit.

AndyZe:
main branch, built from source

AndyZe:
Yep, the plan for the tutorials makes sense. We should try to keep the `main` tutorials branch up to date with Rolling, though. Partly my fault for approving this PR without testing it.

PR #650, opened 2021-08-31
Forward port 2844, never use cache in source builds

vatanaksoytezer:
Forward ports https://github.com/ros-planning/moveit/commit/0ad09153cadfa92d74354598953d0ff853e52d47 in https://github.com/ros-planning/moveit/pull/2844.

Also I realized we are caching the same docker image in source builds. So although the source code changes the docker image does not get built again. This PR updates source dockers to never use cache. We can still use caching by changing the Dockerfiles to be more cache-able. But I prefer leaving it to another PR.

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -200,7 +200,6 @@ jobs:
           push: true
           cache-from: type=registry,ref=${{ env.GH_IMAGE }}
           cache-to: type=inline
-          no-cache: true
           tags: |
             ${{ env.GH_IMAGE }}
             ${{ env.DH_IMAGE }}
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -24,8 +24,18 @@ jobs:
       DH_IMAGE: moveit/moveit2:${{ matrix.ROS_DISTRO }}-${{ github.job }}
 
     steps:
+      - name: Check for apt updates
+        uses: addnab/docker-run-action@v3
+        id: apt
+        with:
+          image: ${{ env.GH_IMAGE }}
+          run: |
+            apt-get update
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yaml
```diff
@@ -167,6 +200,7 @@ jobs:
           push: true
           cache-from: type=registry,ref=${{ env.GH_IMAGE }}
           cache-to: type=inline
+          no-cache: true
```
rhaschke, line 203: This is not yet needed here, because we don't have a `.dockerignore` file in this repo (yet).

rhaschke (commented):
I generally approve these changes. 
Thanks for pointing out the source-build issue. Indeed, due to the [`.dockerignore` file](https://github.com/ros-planning/moveit/blob/master/.dockerignore), introduced in MoveIt 1 repo, there we currently don't automatically rebuild the `source` image. We need essentially an approach to pass different `.dockerignore` files to our various `docker build` commands. I shortly looked into this issue but didn't (quickly) find an option to pass a `dockerignore` file.

.github/workflows/docker.yaml
vatanaksoytezer: If we use caching here without the .dockerignore files aren't we always caching and not re-building source?

rhaschke: No. When building images, docker always considers the build _context_. If the context has changed, caching will be disabled automatically. With the .dockerignore file, I explicitly disable most of the (source) context to just focus on the package.xml files that are relevant for system packages to be installed. Usually, all source changes are considered.

vatanaksoytezer: I see your point, but looking at the docker jobs of GHA it seems source is not built again even though the repository has changes (see https://github.com/ros-planning/moveit2/actions/workflows/docker.yaml). Am I missing something?

rhaschke: Hm. This is indeed not expected!

rhaschke (approved):

Issue #646, opened 2021-08-27
Planning Scene is missing TF frames


schornakj:
### Description

I think there's a regression in #310 that prevents the planning scene from tracking frames that are published to TF but aren't associated with links in the URDF.

### Your environment
* Rolling
* Ubuntu 20.04
* Source build, `main` branch at [974a7a174](https://github.com/ros-planning/moveit2/commit/974a7a1749afa9f8d57254b052056f3fe0c241d5).

### Steps to reproduce
- `ros2 launch run_move_group run_move_group.launch.py`
- Create a node with a TF publisher and publish "world" to "test_frame".
- `ros2 service call /get_planning_scene moveit_msgs/srv/GetPlanningScene '{}'`

### Expected behaviour
Planning scene should contain info about a transform from "world" to "test_frame".

### Actual behaviour
After #310, the planning scene service response does not contain info about this transform. A practical result of this is that MoveIt Task Constructor's `GeneratePose` stages fail while finding IK solutions for named target poses published through TF.

jrgnicho:
I think I'm getting errors related to this issue.  I basically run the move_group node with the *occupancy_map_updater* expecting to see the octomap in rviz but no octomap is shown and in the terminal I get the following error messages:
```
Do not call canTransform or lookupTransform with a timeout unless you are using another thread for populating data. Without a dedicated thread it will always timeout.  If you have a separate thread servicing tf messages, call setUsingDedicatedThread(true) on your Buffer instance.
[move_group-1] [ERROR] [1630597665.236855007] [tf2_buffer]: Do not call canTransform or lookupTransform with a timeout unless you are using another thread for populating data. Without a dedicated thread it will always timeout.  If you have a separate thread servicing tf messages, call setUsingDedicatedThread(true) on your Buffer instance.
[move_group-1] [ERROR] [1630597665.236945719] [moveit_ros.planning_scene_monitor.planning_scene_monitor]: Transform error: "world_frame" passed to lookupTransform argument target_frame does not exist. 
[move_group-1] [ERROR] [1630597665.236971975] [moveit.ros.occupancy_map_updater]: Transform cache was not updated. Self-filtering may fail. If transforms were not available yet, consider setting robot_description_planning/shape_transform_cache_lookup_wait_time to wait longer for transforms
```

Issue #617, opened 2021-08-17
Refactor PSM & CSM to have private tf_buffer & tf_listener


JafarAbdi:
See https://github.com/ros-planning/moveit2/pull/310#pullrequestreview-731032219

PR #654, opened 2021-08-31
Add private buffer & tf listener to PSM

JafarAbdi:
### Description

Address https://github.com/ros-planning/moveit2/issues/617
Fix: https://github.com/ros-planning/moveit2/issues/646

### Checklist

### DIFF
--- a/moveit_ros/planning/moveit_cpp/src/moveit_cpp.cpp
+++ b/moveit_ros/planning/moveit_cpp/src/moveit_cpp.cpp
@@ -55,11 +55,8 @@ MoveItCpp::MoveItCpp(const rclcpp::Node::SharedPtr& node, const std::shared_ptr<
 
 MoveItCpp::MoveItCpp(const rclcpp::Node::SharedPtr& node, const Options& options,
                      const std::shared_ptr<tf2_ros::Buffer>& tf_buffer)
-  : node_(node), tf_buffer_(tf_buffer)
+  : node_(node)
 {
-  if (!tf_buffer_)
-    tf_buffer_ = std::make_shared<tf2_ros::Buffer>(node_->get_clock());
-
   // Configure planning scene monitor
   if (!loadPlanningSceneMonitor(options.planning_scene_monitor_options))
   {
@@ -100,7 +97,7 @@ MoveItCpp::~MoveItCpp()
 bool MoveItCpp::loadPlanningSceneMonitor(const PlanningSceneMonitorOptions& options)
 {
   planning_scene_monitor_.reset(
-      new planning_scene_monitor::PlanningSceneMonitor(node_, options.robot_description, tf_buffer_, options.name));
+      new planning_scene_monitor::PlanningSceneMonitor(node_, options.robot_description, options.name));
   // Allows us to sycronize to Rviz and also publish collision objects to ourselves
   RCLCPP_DEBUG(LOGGER, "Configuring Planning Scene Monitor");
   if (planning_scene_monitor_->getPlanningScene())
@@ -301,12 +298,11 @@ bool MoveItCpp::execute(const std::string& group_name, const robot_trajectory::R
 
 const std::shared_ptr<tf2_ros::Buffer>& MoveItCpp::getTFBuffer() const
 {
-  return tf_buffer_;
+  return planning_scene_monitor_->getTFClient();
 }
 
 void MoveItCpp::clearContents()
 {
-  tf_buffer_.reset();
   planning_scene_monitor_.reset();
   robot_model_.reset();
   planning_pipelines_.clear();
--- a/moveit_ros/planning/moveit_cpp/src/moveit_cpp.cpp
+++ b/moveit_ros/planning/moveit_cpp/src/moveit_cpp.cpp
@@ -48,14 +48,11 @@ namespace moveit_cpp
 static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit.ros_planning_interface.moveit_cpp");
 constexpr char PLANNING_PLUGIN_PARAM[] = "planning_plugin";
 
-MoveItCpp::MoveItCpp(const rclcpp::Node::SharedPtr& node, const std::shared_ptr<tf2_ros::Buffer>& tf_buffer)
-  : MoveItCpp(node, Options(node), tf_buffer)
+MoveItCpp::MoveItCpp(const rclcpp::Node::SharedPtr& node) : MoveItCpp(node, Options(node))
 {
 }
 
-MoveItCpp::MoveItCpp(const rclcpp::Node::SharedPtr& node, const Options& options,
-                     const std::shared_ptr<tf2_ros::Buffer>& tf_buffer)
-  : node_(node)
+MoveItCpp::MoveItCpp(const rclcpp::Node::SharedPtr& node, const Options& options) : node_(node)
 {
   // Configure planning scene monitor
   if (!loadPlanningSceneMonitor(options.planning_scene_monitor_options))
--- a/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
+++ b/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
@@ -113,7 +113,6 @@ public:
    *  @param name A name identifying this planning scene monitor
    */
   PlanningSceneMonitor(const rclcpp::Node::SharedPtr& node, const std::string& robot_description,
-                       const std::shared_ptr<tf2_ros::Buffer>& tf_buffer = std::shared_ptr<tf2_ros::Buffer>(),
                        const std::string& name = "");
 
   /** @brief Constructor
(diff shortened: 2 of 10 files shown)


schornakj:
This fixes the issues I was seeing related to #646. Awesome!

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
```diff
@@ -167,15 +167,6 @@ void CurrentStateMonitor::startStateMonitor(const std::string& joint_states_topi
...
     if (tf_buffer_ && !robot_model_->getMultiDOFJointModels().empty())
     {
-      // If a dedicated thread is enabled for the buffer this probably means the user is adding them either through
-      // tf2_ros::TransformListener or themselves, so we print a warning message warning that transformCallback is doing
-      // the same duplicate operation
-      if (tf_buffer_->isUsingDedicatedThread())
-      {
-        RCLCPP_WARN(LOGGER, "The tf2_ros::Buffer is attached to tf2_ros::TransformListener and the internal tf "
-                            "subscribers inside CurrentStateMonitor, you may want to remove the transform listener to "
-                            "avoid duplicate addition to the same transforms");
```
v4hn, line 177: This is still a very valid warning when CSM is used *outside* PSM.
There was probably a reason why this was added in the first place and I would be grateful for the feedback in such a case.

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
```diff
@@ -1552,4 +1549,15 @@ void PlanningSceneMonitor::configureDefaultPadding()
   RCLCPP_DEBUG_STREAM(LOGGER, "Loaded " << default_robot_link_padd_.size() << " default link paddings");
   RCLCPP_DEBUG_STREAM(LOGGER, "Loaded " << default_robot_link_scale_.size() << " default link scales");
 }
+
+void PlanningSceneMonitor::startTFListener()
+{
+  // No need to start the tf_listener if we have a current state monitor and multi-dof joints
```
v4hn, line 1555: Please at least add to the comment *why not* and not just *when not*.

v4hn (commented):
Having the transforms updated *either* through the PSM tf listener *or* through a CSM listener is quite ugly.

Why did you not just add a mechanism to start the tf listener in the CSM when it's used from the PSM - whether or not there are MultiDOF joints around?

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
```diff
@@ -1222,6 +1221,7 @@ void PlanningSceneMonitor::startStateMonitor(const std::string& joint_states_top
     {
       current_state_monitor_.reset(new CurrentStateMonitor(pnode_, getRobotModel(), tf_buffer_));
     }
+    startTFListener();
```
henningkayser, line 1224: This should be called from `stopStateMonitor()` as well

henningkayser (commented):
I think this is a good step forward, especially removing the buffer from the PSM constructor. 

> Having the transforms updated _either_ through the PSM tf listener _or_ through a CSM listener is quite ugly.

I totally agree. Ideally, we would be initializing a single listener and then simply registering the CSM callback dynamically. If buffer and listener are maintained by the PSM, this would be straight forward. If we want to be able to run CSM standalone, the buffer and listener need default values in the CSM constructor.  

> Why did you not just add a mechanism to start the tf listener in the CSM when it's used from the PSM - whether or not there are MultiDOF joints around?

Not sure if I follow you there. Are you basically describing what I wrote above?

moveit_ros/planning_interface/common_planning_interface_objects/src/common_objects.cpp
```diff
@@ -73,22 +73,6 @@ SharedStorage& getSharedStorage()
   return *storage;
 #endif
 }
-
```
henningkayser, line 76: Why is this being removed?

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
henningkayser: I agree, we should keep the warning

henningkayser (commented):
+1 after addressing the comments

moveit_ros/planning_interface/common_planning_interface_objects/src/common_objects.cpp
JafarAbdi: It's not used at all after this PR

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
JafarAbdi: I added this warning in the last PR to warn about the previous behavior of using an external tf listener, but since we have an internal one now it will always print this warning which may confuse users

moveit_ros/planning_interface/common_planning_interface_objects/src/common_objects.cpp
henningkayser: You mean that usage was removed in the past? I really had to look up where this was used at all, seemed completely unrelated. But I agree that it can be removed.

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
henningkayser: Hmm, so in both cases (PSM+CSM, CSM standalone) we use a dedicated thread for the tf_buffer, right? I guess there is no straight-forward way to check for external listeners anymore. It's fine to remove this warning then, IMO.
Thoughts for a follow-up: Since we don't want external listeners to run at the same time (only PSM is allowed), how about we make the current CSM constructors friend constructors for PSM and only provide public versions without tf_buffer?

henningkayser (approved):

schornakj (approved):
I tested the latest version of this PR in the context of #646 and it also fixes the problems I'd reported there.

moveit_ros/planning/moveit_cpp/src/moveit_cpp.cpp
```diff
@@ -55,11 +55,8 @@ MoveItCpp::MoveItCpp(const rclcpp::Node::SharedPtr& node, const std::shared_ptr<
 
 MoveItCpp::MoveItCpp(const rclcpp::Node::SharedPtr& node, const Options& options,
```
henningkayser, line 56: `tf_buffer` is not used anymore so we should provide a constructor without it. Technically, we should deprecate all constructors that used `tf_buffer` instead of removing them immediately, also.

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
```diff
@@ -113,7 +113,6 @@ class MOVEIT_PLANNING_SCENE_MONITOR_EXPORT PlanningSceneMonitor : private boost:
    *  @param name A name identifying this planning scene monitor
    */
   PlanningSceneMonitor(const rclcpp::Node::SharedPtr& node, const std::string& robot_description,
-                       const std::shared_ptr<tf2_ros::Buffer>& tf_buffer = std::shared_ptr<tf2_ros::Buffer>(),
                        const std::string& name = "");
```
tylerjw, line 116: Would you mind creating deprecated constructors that just call the new ones that we can delete after some time?

tylerjw (commented):
minor comment to help users transition to the new psm constructor

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
henningkayser: This (https://github.com/ros-planning/moveit2/pull/654#discussion_r719291506)

henningkayser (approved):
I think this has turned out pretty great! lgtm

schornakj:
One new thing I do see when testing with this branch is new error log messages from the `move_group` node:

```
[move_group-6] [ERROR] [1633475204.932107474] [tf2_buffer]: Do not call canTransform or lookupTransform with a timeout unless you are using another thread for populating data. Without a dedicated thread it will always timeout.  If you have a separate thread servicing tf messages, call setUsingDedicatedThread(true) on your Buffer instance.
```

This doesn't appear to have a functional effect, though.

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
```diff
@@ -116,6 +115,8 @@ PlanningSceneMonitor::PlanningSceneMonitor(const rclcpp::Node::SharedPtr& node,
                      std::to_string(reinterpret_cast<std::size_t>(this)));
   new_args.push_back("--");
   pnode_ = std::make_shared<rclcpp::Node>("_", node_->get_namespace(), rclcpp::NodeOptions().arguments(new_args));
+  tf_listener_ = std::make_shared<tf2_ros::TransformListener>(*tf_buffer_);
```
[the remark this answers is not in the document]
schornakj: I double-checked and I was on an out-of-date branch when I got this error. I made sure I was up to date and I no longer see this. Sorry for the false alarm.

schornakj (approved):
Explicit re-approval: I tested with the latest version of this PR and it works as expected.

henningkayser:
> Explicit re-approval: I tested with the latest version of this PR and it works as expected.

Thanks for looking into this again

PR #657, opened 2021-08-31
[main] Migrate to joint_state_broadcaster

vatanaksoytezer:
With ros-controls/ros2_controllers#230 ros2_control now removed joint_state_controller in favor of joint_state_broadcaster. This PR migrates our main branch to joint_state_broadcaster. Depends on ros-planning/moveit_resources#96.

### DIFF
--- a/moveit2_galactic.repos
+++ b/moveit2_galactic.repos
@@ -1,5 +1,9 @@
 # TODO(henningkayser): Remove once ros2_control is released (https://github.com/ros-controls/ros2_control/issues/421)
 repositories:
+  moveit_resources:
+    type: git
+    url: https://github.com/ros-planning/moveit_resources
+    version: ros2
   ros2_control:
     type: git
     url: https://github.com/ros-controls/ros2_control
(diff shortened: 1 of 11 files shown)


vatanaksoytezer:
Only tests are failing because  ros-planning/moveit_resources#96 is not merged.

vatanaksoytezer:
So, as this PR depends on moveit_resources. We need to add it to moveit.repos.

tylerjw:
> So, as this PR depends on moveit_resources. We need to add it to moveit.repos.

Can you add it to the repos file in this PR?

moveit2_galactic.repos
```diff
@@ -1,5 +1,9 @@
 # TODO(henningkayser): Remove once ros2_control is released (https://github.com/ros-controls/ros2_control/issues/421)
 repositories:
+  moveit_resources:
+    type: git
+    url: https://github.com/ros-planning/moveit_resources
+    version: ros2
```
tylerjw, line 6: I don't think you need this here if you put it in the main one.

vatanaksoytezer: CI takes that into account:

`UPSTREAM_WORKSPACE: moveit2.repos $(f="moveit2_$(sed 's/-.*$//' <<< "${{ matrix.env.IMAGE }}").repos"; test -r $f && echo $f)`

vatanaksoytezer: Thinking again you are right, I am removing it and see if it still passes.

tylerjw (approved):

Issue #651, opened 2021-08-31
Discussion: Should we make CCov CI only run in push?


vatanaksoytezer:
Possibly discussion for @tylerjw @rhaschke and @henningkayser. Also aynone please feel free to chime in.

CCov CI Jobs bottlenecks our CI time. Usually a nice cached CI job with Release builds takes around 8 to 20 minutes. But we still have to wait around an hour or so for CCov job to be completed (since it is built with RelWithDebugInfo).

For context example run from yesterday:
![image](https://user-images.githubusercontent.com/13888103/131490760-91269910-e460-4230-a2ac-8020a28fc212.png)

I propose we either run CCov with PR's that has a specific label, or only run it on push to save around 25 minutes every CI.

tylerjw:
Is there a way we could make it optional so it still runs with the PR but doesn't block the PR itself until it finishes?

vatanaksoytezer:
I think @henningkayser can do that

v4hn:
> Is there a way we could make it optional so it still runs with the PR but doesn't block the PR itself until it finishes?

Sure, it's a simple entry in the config for the branch. I'm not sure it's a good idea to remove it though. It's the only job testing `rolling-ci` and you probably want to test on rolling before merge.

Code coverage checks were originally proposed by @davetcoleman iirc.
Personally, I don't see the need to run this with each PR. I believe the original idea was to visualize how useful a PR is in terms of added tests. But  because we run with `RelWithDebInfo` even pull-requests that do *nothing* but add tests often are shown to *decrease* coverage.

> CCov CI Jobs bottlenecks our CI time

I'm not convinced 25 additional minutes are a bottleneck at all with MoveIt's development cycle. MoveIt2 lands a few more commits, but do you really feel that these 25 minutes limit your productivity? If you are concerned about reaching limits in GHA minutes, then it's probably a good idea to disable it altogether, but I would only take action here if we really observe paused jobs and can verify that's the reason.

--- Semi-related:
@vatanaksoytezer @rhaschke

Looking into the settings for the moveit2 repo, I noticed the clang-tidy job is currently allowed to fail - probably because it's based on `rolling-ci-testing` and it's not unlikely that `*-testing*` might fail sporadically for unrelated reasons.
Still, I would think `clang-tidy` should be enforced here as well. Not sure who currently manages the CI config for this repository.

vatanaksoytezer:
@v4hn thanks for chiming in and a detailed write-up!

[...]

PR #661, opened 2021-08-31
Move clang tidy under galactic-ci

vatanaksoytezer:
As per discussion in #651 I think it is a good idea to enforce clang-tidy and move it under galactic-ci job (which is enforced).

Update: @henningkayser or any admin needs to change the required jobs before merging this.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,8 +20,8 @@ jobs:
             CCOV: true
           - IMAGE: rolling-ci-testing
             IKFAST_TEST: true
-            CLANG_TIDY: pedantic
           - IMAGE: galactic-ci
+            CLANG_TIDY: pedantic
           - IMAGE: galactic-ci-testing
     env:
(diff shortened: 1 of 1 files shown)


v4hn:
Thanks for the quick response on a side note!
The relevant job fails for unrelated reasons though...

vatanaksoytezer:
> The relevant job fails for unrelated reasons though...

Yep unfortunately everything will fail in main branch until https://github.com/ros-planning/moveit2/pull/657 gets merged.

henningkayser:
This is coming in a little late, but why shouldn't we run this on rolling-ci?

vatanaksoytezer:
> This is coming in a little late, but why shouldn't we run this on rolling-ci?

rolling-ci is already slow with ccov (around 45 minutes each run). galactic-ci takes around 8 minutes and clang-tidy takes around 8 minutes itself. So we can run in parallel (53 + 8 vs 45 + 16).

henningkayser (approved):

PR #662, opened 2021-09-02
planning_scene_monitor: Fix warning about having two publisher with the same node

JafarAbdi:
### Description

Fixes the following warning when having multiple PSM for the same node:
```
rcl.logging_rosout: Publisher already registered for provided node name. If this is due to multiple nodes with the same name then all logs for that logger name will go out over the existing publisher. As soon as any node with that name is destructed it will unregister the publisher, preventing any further logs for that name from being published on the rosout topic.
```
### Checklist

### DIFF
--- a/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
+++ b/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
@@ -111,6 +111,7 @@ PlanningSceneMonitor::PlanningSceneMonitor(const rclcpp::Node::SharedPtr& node,
   std::vector<std::string> new_args = rclcpp::NodeOptions().arguments();
   new_args.push_back("--ros-args");
   new_args.push_back("-r");
+  // Add random ID to prevent warnings about multiple publishers within the same node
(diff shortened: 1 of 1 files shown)


moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
```diff
@@ -111,7 +111,8 @@ PlanningSceneMonitor::PlanningSceneMonitor(const rclcpp::Node::SharedPtr& node,
   std::vector<std::string> new_args = rclcpp::NodeOptions().arguments();
   new_args.push_back("--ros-args");
   new_args.push_back("-r");
-  new_args.push_back(std::string("__node:=") + node_->get_name() + "_private");
+  new_args.push_back(std::string("__node:=") + node_->get_name() + "_private_" +
```
davetcoleman, line 114:
```suggestion
  // Add random ID to prevent warnings about multiple publishers within the same node
  new_args.push_back(std::string("__node:=") + node_->get_name() + "_private_" +
```

davetcoleman (approved):

henningkayser (approved):

PR #664, opened 2021-09-02
Use sensor data QOS profile for sensor_msgs::Image and sensor_msgs::PointCloud

nbbrooks:
### Description

I have been working with a Realsense lately and suspect there are some network traffic issues. I'm starting to read up on how we should be using DDS effectively. This first PR will handle updating point clouds/depth images publishers to use BEST_EFFORT for instead of RELIABLE. Some explanation of the differences are available [here](https://docs.ros.org/en/foxy/Concepts/About-Quality-of-Service-Settings.html#qos-policies).

Currently the Realsense drivers use the system_default QOS, which has reliable settings, but the launch files take in parameters to change it for various topics. Testing this with a Realsense, you would need to launch it with the following parameters:
`pointcloud_qos:=SENSOR_DATA color_qos:=SENSOR_DATA depth_qos:=SENSOR_DATA infra_qos:=SENSOR_DATA`

Similarly, Rviz plugins for visualizing image/pointcloud data should use `Reliability Policy: Best Effort` - I have not yet tested if it will autodetect the QOS settings from the publisher and match them.

When testing this PR with applications using streaming images, users should run `ros2 doctor -r` and look for warnings about mismatched QOS settings between pub and sub.

More info about some of the defined QOS policies:

"sensor_data": what we should use for streaming large data constructs
```
static const rmw_qos_profile_t rmw_qos_profile_sensor_data =
{
  RMW_QOS_POLICY_HISTORY_KEEP_LAST,
  5,
  RMW_QOS_POLICY_RELIABILITY_BEST_EFFORT,
  RMW_QOS_POLICY_DURABILITY_VOLATILE,
  RMW_QOS_DEADLINE_DEFAULT,
  RMW_QOS_LIFESPAN_DEFAULT,
  RMW_QOS_POLICY_LIVELINESS_SYSTEM_DEFAULT,
  RMW_QOS_LIVELINESS_LEASE_DURATION_DEFAULT,
  false
};
```

"default" - what is currently used by MoveIt and by default by realsense
```
static const rmw_qos_profile_t rmw_qos_profile_default =
{
  RMW_QOS_POLICY_HISTORY_KEEP_LAST,
  10,
  RMW_QOS_POLICY_RELIABILITY_RELIABLE,
  RMW_QOS_POLICY_DURABILITY_VOLATILE,
  RMW_QOS_DEADLINE_DEFAULT,
  RMW_QOS_LIFESPAN_DEFAULT,
  RMW_QOS_POLICY_LIVELINESS_SYSTEM_DEFAULT,
  RMW_QOS_LIVELINESS_LEASE_DURATION_DEFAULT,
  false
};
```

"system_default" - this is different than "default" and is defined by the ros middleware. I'm not sure what these values are for cyclone. This is just a note for people to know that in the context of ROS2 DDS "default" != "system_default"
```
static const rmw_qos_profile_t rmw_qos_profile_system_default =
{
  RMW_QOS_POLICY_HISTORY_SYSTEM_DEFAULT,
  RMW_QOS_POLICY_DEPTH_SYSTEM_DEFAULT,
  RMW_QOS_POLICY_RELIABILITY_SYSTEM_DEFAULT,
  RMW_QOS_POLICY_DURABILITY_SYSTEM_DEFAULT,
  RMW_QOS_DEADLINE_DEFAULT,
  RMW_QOS_LIFESPAN_DEFAULT,
  RMW_QOS_POLICY_LIVELINESS_SYSTEM_DEFAULT,
  RMW_QOS_LIVELINESS_LEASE_DURATION_DEFAULT,
  false
};
```

> Best-effort settings reduce the amount of network traffic since the DDS implementation does not have to incur the overhead of reliable communications, where publishers require acknowledgements for messages sent to subscribers and must resend samples that have not been properly received.

Work to be done:
- [] Discuss adding the QOS to be used as a parameter versus just changing the hardcoded policy
- [] Perform some characterization of if/when this has a significant impact on network performance

### Checklist


moveit_ros/occupancy_map_monitor/src/occupancy_map_server.cpp
```diff
@@ -75,8 +75,8 @@ int main(int argc, char** argv)
 {
   rclcpp::init(argc, argv);
   auto node = rclcpp::Node::make_shared("occupancy_map_server");
-  auto octree_binary_pub =
-      node->create_publisher<octomap_msgs::msg::Octomap>("octomap_binary", RMW_QOS_POLICY_HISTORY_KEEP_LAST);
+  rclcpp::QoS qos(rclcpp::QoSInitialization::from_rmw(rmw_qos_profile_default));
```
henningkayser, line 78: Doesn't this have the same effect as leaving `qos` empty?

moveit_ros/perception/depth_image_octomap_updater/src/depth_image_octomap_updater.cpp
```diff
@@ -136,7 +136,7 @@ bool DepthImageOctomapUpdater::initialize(const rclcpp::Node::SharedPtr& node)
 
 void DepthImageOctomapUpdater::start()
 {
-  rmw_qos_profile_t custom_qos = rmw_qos_profile_system_default;
+  rmw_qos_profile_t custom_qos = rmw_qos_profile_sensor_data;
```
henningkayser, line 139: `custom_qos` is only used once when creating the subscription, shouldn't we pass `rmw_qos_profile_sensor_data` directly?

moveit_ros/perception/mesh_filter/src/depth_self_filter_nodelet.cpp
```diff
@@ -206,8 +206,8 @@ void mesh_filter::DepthSelfFiltering::connectCb()
   else if (!sub_depth_image_)
   {
     image_transport::TransportHints hints("raw", ros::TransportHints(), getPrivateNodeHandle());
-    sub_depth_image_ =
-        input_depth_transport_->subscribeCamera("depth", queue_size_, &DepthSelfFiltering::depthCb, this, hints);
+    sub_depth_image_ = input_depth_transport_->subscribeCamera("depth", queue_size_, &DepthSelfFiltering::depthCb, this,
```
henningkayser, line 209: `queue_size_` defaults to `1`, the profile `rmw_qos_profile_sensor_data` has a default history size of 5 which could be customized. I'm not sure if there are any best practices for matching these values.

moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
```diff
@@ -99,13 +99,14 @@ bool PointCloudOctomapUpdater::initialize(const rclcpp::Node::SharedPtr& node)
 
 void PointCloudOctomapUpdater::start()
 {
+  rclcpp::QoS qos(rclcpp::QoSInitialization::from_rmw(rmw_qos_profile_sensor_data));
   if (!filtered_cloud_topic_.empty())
-    filtered_cloud_publisher_ = node_->create_publisher<sensor_msgs::msg::PointCloud2>(filtered_cloud_topic_, 10);
-
+    filtered_cloud_publisher_ = node_->create_publisher<sensor_msgs::msg::PointCloud2>(filtered_cloud_topic_, qos);
```
henningkayser, line 104: hmm.. what effect did the `10` have before this change?

jliukkonen: This is speculation to some extent, but I believe it will end up calling the `rclcpp::QoS(size_t history_depth)` constructor.

Having said that, I suggest the code is written like this:
`filtered_cloud_publisher_ = node_->create_publisher<sensor_msgs::msg::PointCloud2>(filtered_cloud_topic_, rclcpp::SensorDataQoS()); `

moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
```diff
@@ -99,13 +99,14 @@ bool PointCloudOctomapUpdater::initialize(const rclcpp::Node::SharedPtr& node)
...
   if (point_cloud_subscriber_)
     return;
   /* subscribe to point cloud topic using tf filter*/
-  point_cloud_subscriber_ = new message_filters::Subscriber<sensor_msgs::msg::PointCloud2>(node_, point_cloud_topic_);
+  point_cloud_subscriber_ = new message_filters::Subscriber<sensor_msgs::msg::PointCloud2>(node_, point_cloud_topic_,
+                                                                                           rmw_qos_profile_sensor_data);
```
jliukkonen, line 109: This should be fine from the QoS point of view. Annoying how the constructor is so different from couple lines above... :(

moveit_ros/perception/mesh_filter/src/depth_self_filter_nodelet.cpp
jliukkonen: I also don't know which one takes precedence. To err on the side of caution, modify the profile history depth to match whatever value the `queue_` variable has?

moveit_ros/occupancy_map_monitor/src/occupancy_map_server.cpp
```diff
@@ -75,8 +75,8 @@ int main(int argc, char** argv)
...
+  auto octree_binary_pub = node->create_publisher<octomap_msgs::msg::Octomap>("octomap_binary", qos);
```
jliukkonen, line 79: Consider writing this like this:
`auto octree_binary_pub = node->create_publisher<octomap_msgs::msg::Octomap>("octomap_binary", rclcpp::SystemDefaultsQoS());`

... and removing the `qos` temporary variable.

jliukkonen (commented):
@nbbrooks can you make the requested changes so that we can merge this? It is a good set of changes IMO. However, I wonder how this will play together with e.g. Gazebo given how it prefers to use Reliable publishers in its camera plugin. Having said that, I still think we should aim to do the right thing and change the QoS settings here to reasonable constants, which, I believe the `rmw_qos_profile_sensor_data` is.

moveit_ros/occupancy_map_monitor/src/occupancy_map_server.cpp
nbbrooks: Dusting this very old PR off, I'm not sure why I made this change. Yes, it should have the same effect as leaving it empty, but I don't remember what my motivation was for changing this to default (which is reliable, instead of best effort). I think this was under the assumption that single octomap "snapshots" were going to be sent for what I was trying. But I don't think that is a good assumption to merge into main and I should revert this

henrygerardmoore:
@henningkayser @jliukkonen I made some changes based on what you said, give it a look and let me know if anything else needs changing. I wasn't quite sure about the changes in `depth_self_filter_nodelet` so let me know if those aren't what you meant.

henningkayser (approved):
Changes look good to me. Isn't it time that we get this merged? @jliukkonen what do you think?

tylerjw:
I rebased this, if it passes CI I'm going to merge this. Speak now if I should not do that.

vatanaksoytezer (approved):

PR #666, opened 2021-09-05
Small matrix calc speedup in collision_distance_field_types

AndyZe:
I thought it would be a good idea to scour the codebase for places where the fact that [(rotation matrix transpose) == (rot. matrix inverse)](https://www.google.com/search?channel=fs&client=ubuntu&q=transpose+of+rotation+matrix+is+inverse) could be used to speed up calculations. Since a transpose is cheaper than an inverse.

This is the only place I found where that applies.

Eigen has some internal optimizations for Isometry3d types, but I doubt it's smart enough to look ahead and realize it only needs to invert the 3x3 rotation matrix.

Reviewers will need a slightly advanced background in linear algebra, which is why I tagged @mamoll and @v4hn.

### DIFF
--- a/moveit_core/collision_distance_field/include/moveit/collision_distance_field/collision_distance_field_types.h
+++ b/moveit_core/collision_distance_field/include/moveit/collision_distance_field/collision_distance_field_types.h
@@ -160,7 +160,7 @@ public:
   double getDistanceGradient(double x, double y, double z, double& gradient_x, double& gradient_y, double& gradient_z,
                              bool& in_bounds) const
   {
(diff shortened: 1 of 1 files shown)


nbbrooks:
I think Eigen may already be optimizing this, as pose_ is an Isometry3d object.

For [Eigen::Transform](https://eigen.tuxfamily.org/dox/classEigen_1_1Transform.html):
> Isometry: same as Affine with the additional assumption that the linear part represents a rotation. This assumption is exploited to speed up some functions such as inverse() and rotation().

henningkayser (approved):

PR #667, opened 2021-09-07
Error if Octomap 'map_frame' is not provided

AndyZe:
Essentially just a copy of what is done a few lines down, if no "sensors" are provided.

### DIFF
--- a/moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor_middleware_handle.cpp
+++ b/moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor_middleware_handle.cpp
@@ -81,6 +81,10 @@ OccupancyMapMonitorMiddlewareHandle::OccupancyMapMonitorMiddlewareHandle(const r
   if (parameters_.map_frame.empty())
   {
     node_->get_parameter("octomap_frame", parameters_.map_frame);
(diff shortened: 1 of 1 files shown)


moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor_middleware_handle.cpp
```diff
@@ -81,6 +81,10 @@ OccupancyMapMonitorMiddlewareHandle::OccupancyMapMonitorMiddlewareHandle(const r
   if (parameters_.map_frame.empty())
   {
     node_->get_parameter("octomap_frame", parameters_.map_frame);
+    if (parameters_.map_frame.empty())
+    {
+      RCLCPP_ERROR(LOGGER, "No 'map_frame' defined for octomap updates");
```
JafarAbdi, line 86: Should be `octomap_frame` .?
```suggestion
      RCLCPP_ERROR(LOGGER, "No 'octomap_frame' parameter defined for octomap updates");
```

AndyZe: Yeah, good point. I guess that was changed since ROS1.

JafarAbdi (approved):

PR #673, opened 2021-09-09
moveit_servo: Fix ACM for collision checking & PSM's scene monitor topic

JafarAbdi:
### Description

This fixes two bugs:
1- The collision checking for servo is only using the initial ACM and doesn't reflect the updates PSM is receiving in the scene monitor topic
2- When using servo with move_group, servo's scene monitor should point to the move_group's one, otherwise we can't sync the planning scene between them

For the second point, I'm not sure whether we should have a parameter or always use `monitored_planning_scene` @AndyZe what do you think .? I think this depends on whether the PSM in servo is the primary one or not, right .?

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo_parameters.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo_parameters.h
@@ -83,6 +83,8 @@ struct ServoParameters
   std::string move_group_name;
   std::string planning_frame;
   std::string ee_frame_name;
+  bool is_primary_planning_scene_monitor;
+  std::string monitored_planning_scene_topic;
   // Stopping behaviour
   double incoming_command_timeout;
   int num_outgoing_halt_msgs_to_publish;
--- a/moveit_ros/moveit_servo/src/collision_check.cpp
+++ b/moveit_ros/moveit_servo/src/collision_check.cpp
@@ -79,7 +79,6 @@ CollisionCheck::CollisionCheck(rclcpp::Node::SharedPtr node, const ServoParamete
       std::bind(&CollisionCheck::worstCaseStopTimeCB, this, std::placeholders::_1));
 
   current_state_ = planning_scene_monitor_->getStateMonitor()->getCurrentState();
-  acm_ = getLockedPlanningSceneRO()->getAllowedCollisionMatrix();
 }
 
 planning_scene_monitor::LockedPlanningSceneRO CollisionCheck::getLockedPlanningSceneRO() const
@@ -114,8 +113,8 @@ void CollisionCheck::run()
 
   collision_result_.clear();
   // Self-collisions and scene collisions are checked separately so different thresholds can be used
-  getLockedPlanningSceneRO()->getCollisionEnvUnpadded()->checkSelfCollision(collision_request_, collision_result_,
-                                                                            *current_state_, acm_);
+  getLockedPlanningSceneRO()->getCollisionEnvUnpadded()->checkSelfCollision(
+      collision_request_, collision_result_, *current_state_, getLockedPlanningSceneRO()->getAllowedCollisionMatrix());
   self_collision_distance_ = collision_result_.distance;
   collision_detected_ |= collision_result_.collision;
   collision_result_.print();
--- a/moveit_ros/moveit_servo/src/servo_parameters.cpp
+++ b/moveit_ros/moveit_servo/src/servo_parameters.cpp
@@ -40,6 +40,7 @@
 */
 
 #include <moveit_servo/servo_parameters.h>
+#include <moveit/planning_scene_monitor/planning_scene_monitor.h>
 #include <rclcpp/rclcpp.hpp>
 #include <type_traits>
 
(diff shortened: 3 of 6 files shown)


moveit_ros/moveit_servo/src/collision_check.cpp
```diff
@@ -115,8 +114,8 @@ void CollisionCheck::run()
 
   collision_result_.clear();
   // Self-collisions and scene collisions are checked separately so different thresholds can be used
-  getLockedPlanningSceneRO()->getCollisionEnvUnpadded()->checkSelfCollision(collision_request_, collision_result_,
-                                                                            *current_state_, acm_);
+  getLockedPlanningSceneRO()->getCollisionEnvUnpadded()->checkSelfCollision(
+      collision_request_, collision_result_, *current_state_, getLockedPlanningSceneRO()->getAllowedCollisionMatrix());
```
AndyZe, line 118: I see that `getAllowedCollisionMatrix()` returns a reference to the ACM so I think there shouldn't be any hit to performance  :+1:

moveit_ros/moveit_servo/src/servo_server.cpp
```diff
@@ -96,7 +96,9 @@ bool ServoServer::init()
     planning_scene_monitor_->setPlanningScenePublishingFrequency(25);
     planning_scene_monitor_->startPublishingPlanningScene(planning_scene_monitor::PlanningSceneMonitor::UPDATE_SCENE,
                                                           "/moveit_servo/publish_planning_scene");
-    planning_scene_monitor_->startSceneMonitor();
+    planning_scene_monitor_->startSceneMonitor(
+        planning_scene_monitor::PlanningSceneMonitor::MONITORED_PLANNING_SCENE_TOPIC);
+    planning_scene_monitor_->requestPlanningSceneState();
```
AndyZe, line 101: I don't understand what this line does. Are you using it to wait for an update?

If so, please add a comment to that effect:

`// Request a planning scene update and wait for the service response`

AndyZe:
> For the second point, I'm not sure whether we should have a parameter or always use `monitored_planning_scene` @AndyZe what do you think .? I think this depends on whether the PSM in servo is the primary one or not, right .?

I'm not sure I know enough about how the PSM works to answer the question.

It looks like there can be multiple update topics that each PSM listens to, judging by [this function](https://github.com/ros-planning/moveit2/blob/5ff36ac4855411a8eb8eac636d00ef3838d9611e/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h#L367). So I would say, it would be very hard to handle all possible topic combinations.

I think we should leave it as you have it now. It's a good default. In the future, if somebody wants to do something wild like publish planning scene updates on 4 different topics, and maintain 3 separate planning scenes, we can figure it out then.

I'm open to opinions from others.

nbbrooks (commented):
I think using same default PSM topic is good default behavior. Supporting a separate PSM, and providing documentation, etc, on how to set that up (especially if they aren't starting at the same time and there might be missed collision object messages) sounds like a separate PR to me. I'm not aware of any MoveIt applications where separate PSMs are desired

moveit_ros/moveit_servo/include/moveit_servo/servo_parameters.h
```diff
@@ -85,6 +85,8 @@ struct ServoParameters
   std::string move_group_name;
   std::string planning_frame;
   std::string ee_frame_name;
+  bool is_primary_planning_scene_monitor;
```
nbbrooks, line 88: Should we add this to the sample servo config files in moveit2 repo and moveit2_tutorials repo?

nbbrooks (approved):

henningkayser:
https://github.com/ros-planning/moveit2/pull/713 should fix CI

henningkayser (approved):

moveit_ros/moveit_servo/src/servo_parameters.cpp
```diff
@@ -114,6 +115,11 @@ ServoParameters::SharedConstPtr ServoParameters::makeServoParameters(const rclcp
   declareOrGetParam<std::string>(parameters->move_group_name, ns + ".move_group_name", node, logger);
   declareOrGetParam<std::string>(parameters->planning_frame, ns + ".planning_frame", node, logger);
   declareOrGetParam<std::string>(parameters->ee_frame_name, ns + ".ee_frame_name", node, logger);
+  declareOrGetParam<bool>(parameters->is_primary_planning_scene_monitor, ns + ".is_primary_planning_scene_monitor",
+                          node, logger, true);
```
AndyZe, line 119: I think this should default to the same behavior we had previously. So, shouldn't this default to `false`?

AndyZe: @JafarAbdi can you answer this question? Then I'm good to merge this

tylerjw:
Please rebase this change as I recently made some changes that affect it.

moveit_ros/moveit_servo/src/servo_parameters.cpp
tylerjw: I think defaulting to true is reasonable because if you default to false you'll need to change it to true for every case where you want to run servo by itself without another psm running.

tylerjw (approved):

AndyZe:
I gave this a final test with the Rolling tutorial. LGTM

AndyZe (approved):

PR #674, opened 2021-09-10
Only run cache upstream ws if no .repos file changed in PRs

JafarAbdi:
### Description

One issue I had in this [PR](https://github.com/ros-planning/moveit2/pull/591) is a conflict between one of the cached upstream packages and the newly added hash for it which was causing the CI to always fail with the following error during `vcs pull $BASEDIR/upstream_ws/src`

```
  === /home/runner/work/moveit2/moveit2/.work/upstream_ws/src/moveit_resources (git) ===
  
  Already on 'pr-moveit_configs_utils'
  Your branch and 'origin/pr-moveit_configs_utils' have diverged,
  and have 1 and 3 different commits each, respectively.
    (use "git pull" to merge the remote branch into yours)
```

This will only affect PRs

### Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -60,7 +60,18 @@ jobs:
           sudo rm -rf /usr/local
           df -h
       - uses: actions/checkout@v2
+      - name: Check if upstream dependencies changed
+        id: upstream_ws_changed
+        shell: bash
+        run: |
+          sudo apt-get install -qq jq
+          URL="https://api.github.com/repos/${GITHUB_REPOSITORY}/pulls/${{ github.event.pull_request.number }}/files"
+          REPOS_CHANGED='false'
+          if [ -n "$(curl -s -X GET -G $URL | jq -r '.[] | .filename' | grep "moveit2*.repos")" ]; then REPOS_CHANGED='true';fi
+          echo set-output name=REPOS_CHANGED::${REPOS_CHANGED}
+          echo ::set-output name=REPOS_CHANGED::${REPOS_CHANGED}
       - name: cache upstream_ws
+        if: steps.upstream_ws_changed.outputs.REPOS_CHANGED == 'false'
         uses: pat-s/always-upload-cache@v2.1.5
         with:
           path: ${{ env.BASEDIR }}/upstream_ws


vatanaksoytezer (approved):
Thanks a ton for fixing that @JafarAbdi! I saw the issue but I can't think of a better solution right now. This looks clean and neat.

rhaschke (commented):
Hm. Looks like the `UPSTREAM_WORKSPACE` is not correctly considered in `CACHE_PREFIX`. I will file a simpler solution.

rhaschke:
I pushed a simpler fixup. However, unfortunately, the calculation of `CACHE_PREFIX` via [`hashFiles()`](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#hashfiles) doesn't work.
On my test branch https://github.com/rhaschke/moveit2/tree/pr-fix_ci I pushed some changes which should or should not invalidate the cache. However, it happens exactly the opposite of the expected:
When a file that is _not_ hashed has changed, the [`target_ws` cache was invalid](https://github.com/rhaschke/moveit2/runs/3607946377?check_suite_focus=true) (but `upstream_ws` cache was valid as expected).
When a file that _is_ hashed has changed, the [`upstream_ws` cache remains valid](https://github.com/rhaschke/moveit2/runs/3608515480?check_suite_focus=true).

@vatanaksoytezer, could you have a look at this in a smaller toy example?
Maybe GHA cannot handle the (output) variable being used in `hashFiles()`?

JafarAbdi:
> The currently proposed solution would disable the cache once a change is detected.

This is exactly what I was trying to do, the problem we have is when someone opens a PR and change one of the repo's branch in `.repos` to point to a temporary branch (a PR in one of the upstream repos) and force push to that temporary branch, it will cause `vcs pull` to fail in some cases

The ultimate solution would be to always force pull but that's not possible in vcs right now

rhaschke:
> The currently proposed solution would disable the cache once a change is detected.

The point is that the cache wouldn't be rewritten as well. So it stays disabled forever!
We really need to fix the cache keys.

JafarAbdi:
> The point is that the cache wouldn't be rewritten as well. So it stays disabled forever!

Yes, it will be disabled forever only for the PRs that have changes to `.repos`

> We really need to fix the cache keys.

I don't think this will help, in my case I pushed a conflicting change to the branch I used in the temporary repo way after changing the `.repos` file and it caused a problem once I pushed a new commit to the PR

vatanaksoytezer:
> @vatanaksoytezer, could you have a look at this in a smaller toy example?
> Maybe GHA cannot handle the (output) variable being used in `hashFiles()`?

Sure, I'll make a minimal example and see what I can do.

rhaschke:
> Yes, it will be disabled forever only for the PRs that have changes to `.repos`

It will also be disabled as soon as we want to push a change in the `.repos` files to the `main` branch!
I don't get why you think that fixing the cache-key mechanism will not work. It's exactly intended for this purpose.

JafarAbdi:
> It will also be disabled as soon as we want to push a change in the .repos files to the main branch!

I don't think this will be the case, this only work for pull-request, the expression inside the if-statement will evaluate to false for the branches, I even tried changing it on my fork [see](https://github.com/JafarAbdi/moveit2/tree/main)

> I don't get why you think that fixing the cache-key mechanism will not work. It's exactly intended for this purpose.

If we can get the cache-key solution to work it will be better than the original one, I'll meet with Vatan to work on fixing this

rhaschke:
I performed a series of tests in this [toy repo](https://github.com/rhaschke/test). Looks like `hashFiles` doesn't support any kind of computed input list. Thus falling back to static list with wildcards. That's a little bit too broad, but these files don't change that often...

PR #675, opened 2021-09-10
Fix hybrid planning includes

henningkayser:
@sjahr


sjahr (approved):
Works fine, thanks for fixing this!

PR #676, opened 2021-09-13
[Servo] Implement a smoothing plugin

AndyZe:
This keeps the same old low-pass filter functionality but wraps it in a plugin, in preparation for adding some other smoothing options in the future.

Smoothing functionality was moved to a new folder, `smoothing_plugin`.


moveit_ros/moveit_servo/include/moveit_servo/smoothing_plugin/smoothing_base_class.h
```diff
@@ -0,0 +1,77 @@
...
+
+  /**
+   * Reset to a given joint state
+   * @param joint_positions reset the filters to these joint positions
+   * @return True if reset was successful
+   */
+  virtual bool reset(const std::vector<double> joint_positions) = 0;
+
+protected:
+  SmoothingBaseClass(){};
```
griswaldbrooks, line 75: You don't need this as pure virtual classes cannot be instantiated

moveit_ros/moveit_servo/include/moveit_servo/smoothing_plugin/smoothing_base_class.h
```diff
@@ -0,0 +1,77 @@
...
+   * @return True if initialization was successful
+   */
+  virtual bool doSmoothing(Eigen::ArrayXd& delta_theta) = 0;
+
+  /**
+   * Reset to a given joint state
+   * @param joint_positions reset the filters to these joint positions
+   * @return True if reset was successful
+   */
+  virtual bool reset(const std::vector<double> joint_positions) = 0;
```
griswaldbrooks, line 72: Shouldn't this be `const&`?

moveit_ros/moveit_servo/include/moveit_servo/smoothing_plugin/low_pass_filter.h
```diff
@@ -0,0 +1,118 @@
...
+   */
+  LowpassFilterImpl(double low_pass_filter_coeff);
+
+  double filter(double new_measurement);
+
+  void reset(double data);
+
+private:
+  static constexpr std::size_t FILTER_LENGTH = 2;
+  double previous_measurements_[FILTER_LENGTH];
```
griswaldbrooks, line 72: consider `std::array`

moveit_ros/moveit_servo/CMakeLists.txt
```diff
@@ -30,34 +30,35 @@ find_package(moveit_common REQUIRED)
 moveit_package()
 
 find_package(ament_cmake REQUIRED)
+find_package(control_msgs REQUIRED)
```
AndyZe, line 33: Alphabetized

moveit_ros/moveit_servo/include/moveit_servo/smoothing_plugin/low_pass_filter.h
```diff
@@ -0,0 +1,118 @@
...
+
+  /**
+   * Initialize the smoothing algorithm
+   * @param node ROS node, used for parameter retrieval
+   * @param robot_model typically used to retrieve vel/accel/jerk limits
+   * @param num_joints number of actuated joints in the JointGroup Servo controls
+   * @param parameters access to all Servo parameters that have been parsed from yaml
+   * @return True if initialization was successful
+   */
+  bool initialize(rclcpp::Node::SharedPtr node, moveit::core::RobotModelConstPtr robot_model, const size_t num_joints,
```
griswaldbrooks, line 93: Avoid `const` value types in declaration (but keep them in definition) https://abseil.io/tips/109

moveit_ros/moveit_servo/include/moveit_servo/smoothing_plugin/low_pass_filter.h
```diff
@@ -0,0 +1,118 @@
...
+  bool doSmoothing(Eigen::ArrayXd& delta_theta);
+
+  /**
+   * Reset to a given joint state
+   * @param joint_positions reset the filters to these joint positions
+   * @return True if reset was successful
+   */
+  bool reset(const std::vector<double> joint_positions)
+  {
+    return true;
```
griswaldbrooks, line 110: Did you mean to reset the `position_filters_`?

moveit_ros/moveit_servo/src/smoothing_plugin/low_pass_filter.cpp
```diff
@@ -0,0 +1,85 @@
...
+  previous_measurements_[0] = data;
+  previous_measurements_[1] = data;
+
+  previous_filtered_measurement_ = data;
+}
+
+bool LowPassFilter::initialize(rclcpp::Node::SharedPtr node, moveit::core::RobotModelConstPtr robot_model,
+                               const size_t num_joints,
+                               const std::shared_ptr<const moveit_servo::ServoParameters>& parameters)
+{
```
griswaldbrooks, line 59: I think all of this can be done in the initializer list

moveit_ros/moveit_servo/src/smoothing_plugin/low_pass_filter.cpp
```diff
@@ -0,0 +1,85 @@
...
+    // Low-pass filters for the joint positions
+    position_filters_.emplace_back(parameters->low_pass_filter_coeff);
+  }
+
+  return true;
+};
+
+bool LowPassFilter::doSmoothing(Eigen::ArrayXd& delta_theta)
+{
+  for (uint i = 0; i < delta_theta.size(); ++i)
```
griswaldbrooks, line 74: Shouldn't `uint` be `Eigen::Index`?

AndyZe:
Thanks for a quick review and catching several issues @griswaldbrooks!

moveit_ros/moveit_servo/src/smoothing_plugin/low_pass_filter.cpp
AndyZe: AFAIK the plugin constructor cannot take any arguments, so it's actually not possible to do this in the constructor.

moveit_ros/moveit_servo/include/moveit_servo/smoothing_plugin/low_pass_filter.h
AndyZe: Meh, I'm not sure I agree with that. The `const` in the header file is part of documenting usage and it should match what's in the cpp file.

Let me know if I misinterpreted your comment.

moveit_ros/moveit_servo/src/smoothing_plugin/low_pass_filter.cpp
AndyZe: Here's a bit of documentation:

> With pluginlib, a constructor without parameters is required for classes so, if any parameters are required, we use the initialize method to initialize the object. 

http://wiki.ros.org/pluginlib/Tutorials/Writing%20and%20Using%20a%20Simple%20Plugin

griswaldbrooks: Sorry, I misread this and thought it was a ctor

moveit_ros/moveit_servo/include/moveit_servo/smoothing_plugin/low_pass_filter.h
griswaldbrooks: The point of the tip is that C++ cannot disambiguate between `const` value and non-`const` value passing so `void f(int x)` and `void f(const int x)` are the same. As abseil is not part of our style guide pipeline, it's your option. As far as usage, it's a copy so it doesn't affect the caller.

AndyZe: I did read the abseil thing. Basically, I think it says:

void F(const int);               //    <--  use const here
void F(const int) { /* ... */ }  // <-- do not use const here

Which is different from what you're saying. I agree with you though, so I'll change it

griswaldbrooks: The bottom of the page summarizes their guidance
> 1. Never use top-level const on function parameters in declarations that are not definitions (and be careful not to copy/paste a meaningless const). It is meaningless and ignored by the compiler, it is visual noise, and it could mislead readers.
> 2. Do use top-level const on function parameters in definitions at your (or your team’s) discretion. You might follow the same rationale as you would for when to declare a function-local variable const.

```
void F(const int); // Declaration
void F(const int) { /* ... */ } // Definition
```

moveit_ros/moveit_servo/include/moveit_servo/smoothing_plugin/low_pass_filter.h
```diff
@@ -0,0 +1,115 @@
...
+
+namespace moveit_servo
+{
+/**
+ * Class LowpassFilterImpl - Implementation of a signal filter to soften jerks.
+ * This is a first-order Butterworth low-pass filter. First-order was chosen for 2 reasons:
+ * - It doesn't overshoot
+ * - Computational efficiency
+ */
+class LowpassFilterImpl
```
henningkayser, line 55: Nit `Impl` is easily confused to be the implementation part for a pimpl pattern. Could you rename this to `SinglePositionLowpassFilter` or even `ButterworthLowPassFilter`?

moveit_ros/moveit_servo/src/smoothing_plugin/low_pass_filter.cpp
```diff
@@ -0,0 +1,93 @@
+#include <moveit_servo/smoothing_plugin/low_pass_filter.h>
```
henningkayser, line 1: nit: missing header

henningkayser (commented):
I like this idea a lot. Please consider making the plugin more generic and move it to `moveit_core`  to allow easier access for other parts of MoveIt.

moveit_ros/moveit_servo/test/config/servo_settings.yaml
```diff
@@ -27,6 +26,12 @@ publish_joint_positions: true
 publish_joint_velocities: false
 publish_joint_accelerations: false
 
+## Smoothing plugin
+# The default plugin is a Butterworth lowpass filter that requires this parameter.
+# It must be >1.0, typically in the range 1.0 < coefficient < 100.0.
+# If you use a different plugin, replace with the relevant parameters.
+low_pass_filter_coeff: 2.0  # Larger --> trust the filtered data more, trust the measurements less.
```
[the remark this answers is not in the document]
AndyZe: Yep, that was the plan for the followup PR

AndyZe:
@vatanaksoytezer @tylerjw Do you have any idea why this would pass Rolling CI but not Galactic?

vatanaksoytezer:
> @vatanaksoytezer @tylerjw Do you have any idea why this would pass Rolling CI but not Galactic?

It actually passes on galactic-ci-testing. I looked up the CI job and the error line looks like:

```
Error: mponent_container-2] [ERROR] [1631733238.022479483] [moveit_servo.servo_calcs]: Exception while loading the smoothing plugin 'smoothing_plugins::ButterworthFilterPlugin': 'MultiLibraryClassLoader: Could not create object of class type smoothing_plugins::ButterworthFilterPlugin as no factory exists for it. Make sure that the library exists and was explicitly loaded through MultiLibraryClassLoader::loadLibrary()'
Error: ROR] [component_container-2]: process has died [pid 11614, exit code 1, cmd '/opt/ros/galactic/lib/rclcpp_components/component_container --ros-args -r __node:=servo_server_container -r __ns:=/'].
```

Maybe the class loader or componenet was not acting nice in galactic-main but fixed in testing? Worth an investigation in those relevant repositories. Also can you run the tests in your local if you have a galactic setup? (Else I can do for you) I wonder if they also fail in local as well or is this a CI fault.

AndyZe:
> Maybe the class loader or component container was not acting nice in galactic-main but fixed in testing? Worth an investigation in those relevant repositories.

I'm not sure what you mean by "fixed in testing". Is there a separate version of Galactic that is used only for testing or something?

> Also can you run the tests in your local if you have a galactic setup? (Else I can do for you)

That would be great! I think you will do it a lot faster than me.

vatanaksoytezer:
> I'm not sure what you mean by "fixed in testing". Is there a separate version of Galactic that is used only for testing or something?

There is main repositories (which we all normally use) and testing repositories, which includes all the changes that are merged in https://github.com/ros/rosdistro but not yet synced. So your PR passes on testing repositories (which will become main with the next sync) but fails on current main repositories.

moveit_core/CMakeLists.txt
```diff
@@ -158,35 +157,35 @@ message(STATUS " *** Building MoveIt ${MOVEIT_VERSION} ***")
...
 add_subdirectory(backtrace)
+add_subdirectory(collision_distance_field)
+add_subdirectory(constraint_samplers)
+add_subdirectory(controller_manager)
+add_subdirectory(distance_field)
+add_subdirectory(dynamics_solver)
 add_subdirectory(exceptions)
+add_subdirectory(kinematics_base)
+add_subdirectory(kinematic_constraints)
+add_subdirectory(macros)
```
henningkayser, line 170: Alphabetized? macros should be below kinematic_metrics, collision_detection_* is still at the bottom

moveit_core/smoothing_plugins/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+  "${${PROJECT_NAME}_VERSION}"
+)
+target_link_libraries(moveit_butterworth_filter
+  moveit_robot_model
+)
+
+# Installation
+
+install(DIRECTORY include/ DESTINATION include)
+install(FILES ${CMAKE_CURRENT_BINARY_DIR}/moveit_butterworth_filter_export.h DESTINATION include)
```
henningkayser, line 21: Why is this extra install necessary? All headers should be installed with the line above

moveit_core/smoothing_plugins/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+install(TARGETS moveit_butterworth_filter EXPORT moveit_butterworth_filter)
+install(
```
henningkayser, line 23: The two install commands here are redundant, `moveit_butterworth_filter` is already installed with the `moveit_core` CMakeLists.txt.

moveit_core/smoothing_plugins/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
+# Plugin implementations
+
+add_library(moveit_butterworth_filter SHARED
```
henningkayser, line 3: Plugin base and implementation should be separate libraries or even packages (see `kinematics_base` for instance). Otherwise, you would have to depend on the ButterworthFilter to implement your own smoothing plugin which doesn't make a lot of sense.

moveit_core/smoothing_plugins/src/smoothing_base_class.cpp
```diff
@@ -0,0 +1 @@
+#include <moveit/smoothing_plugins/smoothing_base_class.h>
```
henningkayser, line 1: Could you at least provide some basic structure around this, like header and description on what this is? I also don't see this being used anywhere, but I guess it's required for building a standalone library for the virtual base class.

henningkayser (commented):
Almost there, imo. I think that SmoothingBaseClass should be a standalone lib, we can discuss later if we want to move plugin implementations into a separate package. There are still some clang-tidy fixes that need to be applied to make CI happy, could you fix those?

moveit_core/smoothing_plugins/CMakeLists.txt
AndyZe: I'm fairly sure that's not true. PR #682 adds another smoothing plugin that certainly does not depend on ButterworthFilter. See ruckig_filter.h [here](https://github.com/ros-planning/moveit2/pull/682/files#diff-36de96511bdf77e9c0095e1a897b4b6f2b56dd1a1ce7bfbeb996ccf073e68d47).

AndyZe: Also, if you read the [ROS2 plugin tutorial ](https://docs.ros.org/en/foxy/Tutorials/Pluginlib.html)carefully, you'll see that the base class (RegularPolygon) is only a header file. It's not compiled to a library.

moveit_core/CMakeLists.txt
AndyZe: If collision_detection_* is alphabetized, there is this compiler error:

CMake Error at collision_detection/CMakeLists.txt:42 (ament_add_gtest):
  Unknown CMake command "ament_add_gtest".

I don't know why but I don't intend to fix it in this PR. Will leave a TODO

moveit_core/smoothing_plugins/src/smoothing_base_class.cpp
AndyZe: Good catch, this cpp file actually isn't needed at all. I just forgot to delete it when messing around earlier. Deleted.

moveit_core/smoothing_plugins/src/butterworth_filter.cpp
```diff
@@ -0,0 +1,102 @@
...
+    position_filters_.emplace_back(1.5);
+  }
+  return true;
+};
+
+bool ButterworthFilterPlugin::doSmoothing(std::vector<double>& position_vector)
+{
+  if (position_vector.size() != position_filters_.size())
+  {
+    return false;
```
nbbrooks, line 76: should we add a throttled error message here?

moveit_core/smoothing_plugins/src/butterworth_filter.cpp
```diff
@@ -0,0 +1,102 @@
...
+    position_vector[i] = position_filters_.at(i).filter(position_vector[i]);
+  }
+  return true;
+};
+
+bool ButterworthFilterPlugin::reset(const std::vector<double>& joint_positions)
+{
+  if (joint_positions.size() != position_filters_.size())
+  {
+    return false;
```
nbbrooks, line 90: ditto

moveit_core/smoothing_plugins/include/moveit/smoothing_plugins/butterworth_filter.h
```diff
@@ -0,0 +1,113 @@
...
+   * @param robot_model typically used to retrieve vel/accel/jerk limits
+   * @param num_joints number of actuated joints in the JointGroup Servo controls
+   * @return True if initialization was successful
+   */
+  bool initialize(rclcpp::Node::SharedPtr node, moveit::core::RobotModelConstPtr robot_model,
+                  size_t num_joints) override;
+
+  /**
+   * Smooth the command signals for all DOF
+   * @param delta_theta array of incremental joint position commands
```
nbbrooks, line 96:
```suggestion
   * @param position_vector array of incremental joint position commands
```

moveit_core/smoothing_plugins/include/moveit/smoothing_plugins/smoothing_base_class.h
```diff
@@ -0,0 +1,72 @@
...
+   * @param robot_model typically used to retrieve vel/accel/jerk limits
+   * @param num_joints number of actuated joints in the JointGroup Servo controls
+   * @return True if initialization was successful
+   */
+  virtual bool initialize(rclcpp::Node::SharedPtr node, moveit::core::RobotModelConstPtr robot_model,
+                          size_t num_joints) = 0;
+
+  /**
+   * Smooth an array of joint position deltas
+   * @param delta_theta array of incremental joint position commands
```
nbbrooks, line 60:
```suggestion
   * @param position_vector array of incremental joint position commands
```

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
```diff
@@ -191,13 +192,13 @@ class ServoCalcs
...
   /** \brief Joint-wise update of a sensor_msgs::msg::JointState with given delta's
-   * Also calculates the previous velocity
+   * Also filters and calculates the previous velocity
    * @param delta_theta Eigen vector of joint delta's
    * @param joint_state The joint state msg being updated
    * @param previous_vel Eigen vector of previous velocities being updated
    * @return Returns false if there is a problem, true otherwise
    */
-  bool applyJointUpdate(const Eigen::ArrayXd& delta_theta, sensor_msgs::msg::JointState& joint_state,
+  bool applyJointUpdate(Eigen::ArrayXd& delta_theta, sensor_msgs::msg::JointState& joint_state,
```
nbbrooks, line 201: Why did we lose const here?

moveit_core/smoothing_plugins/src/butterworth_filter.cpp
```diff
@@ -0,0 +1,102 @@
...
+                                         size_t num_joints)
+{
+  node_ = node;
+  num_joints_ = num_joints;
+
+  for (std::size_t i = 0; i < num_joints_; ++i)
+  {
+    // Low-pass filters for the joint positions
+    // TODO(andyz): read the parameter from yaml
+    position_filters_.emplace_back(1.5);
```
nbbrooks, line 67: It looks like you are now reading this in but not using it here

nbbrooks: Nevermind ^ after looking at this again. But I think we'll be parsing this from the node not the yaml, correct?

moveit_ros/moveit_servo/test/config/servo_settings.yaml
nbbrooks, line 33: It doesn't look like servo_params allows for `smoothing_filter_plugin_name` to not be set? I would replace the low_pass_filter_coeff with the plugin name since we aren't parsing low_pass_filter_coeff right now.

moveit_core/smoothing_plugins/include/moveit/smoothing_plugins/butterworth_filter.h
AndyZe: Glad you caught that. I needed to delete `incremental` too

moveit_core/smoothing_plugins/include/moveit/smoothing_plugins/smoothing_base_class.h
AndyZe: I was wondering if people usually duplicate the documentation when they inherit from a base class. This is a good example where I had to make your change in 2 places. What do you think?

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
AndyZe: good catch, reverted

moveit_core/smoothing_plugins/src/butterworth_filter.cpp
AndyZe: Meh, I look at that as one and the same thing since yaml is usually converted to parameters. OK, updated it.

moveit_ros/moveit_servo/test/config/servo_settings.yaml
AndyZe: The items in this "test" folder just hadn't been updated yet. Done.

AndyZe:
I'm planning to rename the new pkg from `smoothing_plugins` to `single_waypt_smoothing_plugins`. I think that will help users distinguish between the algorithms in this folder and the time parameterization algorithms.

v4hn:
> I'm planning to rename the new pkg from `smoothing_plugins` to `single_waypt_smoothing_plugins`.
> I think that will help users distinguish between the algorithms in this folder and the time parameterization algorithms.

I agree that the previous name might be misleading, but there is nothing to smooth in a single waypoint (with all vowels :)).
How about "online_command_smoothing" or something along that line?

felixvd:
Also, short reminder that abbreviation by leaving out vowels is discouraged
by the Google style guidelines we use. Michael’s suggestion sounds good

moveit_core/smoothing_plugins/include/moveit/smoothing_plugins/smoothing_base_class.h
nbbrooks: I personally don't like having to dig my way up a interface chain to find documentation (granted, this is just 1 level). Maybe I could be using my IDE better but my preference would be to have the doc here as well.

nbbrooks: In particular I think some of the core ROS stuff is annoying to trace and up find the units being used (which is one reason why I am on the bandwagon of including the unit in the variable name)

moveit_core/online_signal_smoothing/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+)
+target_link_libraries(moveit_butterworth_filter
+  moveit_robot_model
+)
+
+# Installation
+
+install(DIRECTORY include/ DESTINATION include)
+install(FILES ${CMAKE_CURRENT_BINARY_DIR}/moveit_butterworth_filter_export.h DESTINATION include)
+install(TARGETS moveit_butterworth_filter EXPORT moveit_butterworth_filter)
```
henningkayser, line 22: No need to install the targets here. The top-level CMakeLists.txt in `moveit_core` already takes care of that.

moveit_core/online_signal_smoothing/src/butterworth_filter.cpp
```diff
@@ -0,0 +1,106 @@
...
+        "online_signal_smoothing::ButterworthFilter: Filter coefficient value resulted in feedback term of 0");
+}
+
+double ButterworthFilter::filter(double new_measurement)
+{
+  // Push in the new measurement
+  previous_measurements_[1] = previous_measurements_[0];
+  previous_measurements_[0] = new_measurement;
+
+  double new_filtered_measurement = scale_term_ * (previous_measurements_[1] + previous_measurements_[0] -
```
henningkayser, line 41: Directly store in `previous_filtered_measurement_` and return that one?

moveit_core/online_signal_smoothing/src/butterworth_filter.cpp
```diff
@@ -0,0 +1,106 @@
...
+                                         size_t num_joints)
+{
+  node_ = node;
+  num_joints_ = num_joints;
+
+  for (std::size_t i = 0; i < num_joints_; ++i)
+  {
+    // Low-pass filters for the joint positions
+    // TODO(andyz): read the parameter
+    position_filters_.emplace_back(1.5);
```
henningkayser, line 67: Or use the default positions provided with robot_model

moveit_core/online_signal_smoothing/src/butterworth_filter.cpp
```diff
@@ -0,0 +1,106 @@
...
+  }
+  return true;
+};
+
+bool ButterworthFilterPlugin::reset(const std::vector<double>& joint_positions)
+{
+  if (joint_positions.size() != position_filters_.size())
+  {
+    RCLCPP_ERROR_THROTTLE(node_->get_logger(), *node_->get_clock(), 1000,
+                          "Position vector to be smoothed does not have the right length.");
```
henningkayser, line 93:
```suggestion
                          "Position vector to be reset does not have the right length.");
```

moveit_ros/moveit_servo/config/panda_simulated_config.yaml
```diff
@@ -22,11 +22,11 @@ command_out_type: trajectory_msgs/JointTrajectory
 
 # What to publish? Can save some bandwidth as most robots only require positions or velocities
 publish_joint_positions: true
-publish_joint_velocities: false
+publish_joint_velocities: true
```
henningkayser, line 25: Publish both positions and velocities?

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
```diff
@@ -296,7 +300,8 @@ class ServoCalcs
   sensor_msgs::msg::JointState internal_joint_state_, original_joint_state_;
   std::map<std::string, std::size_t> joint_state_name_map_;
 
-  std::vector<LowPassFilter> position_filters_;
+  // Smoothing algorithm (loads a plugin)
+  std::shared_ptr<online_signal_smoothing::SmoothingBaseClass> smoother_;
```
henningkayser, line 304: unique_ptr

henningkayser (commented):
Only minor comments from my side

moveit_ros/moveit_servo/config/panda_simulated_config.yaml
AndyZe: yeah, it works just fine that way

moveit_core/online_signal_smoothing/src/butterworth_filter.cpp
AndyZe: ah, that's a nice little optimization. And here I thought this code was quite streamlined already.

moveit_core/online_signal_smoothing/src/butterworth_filter.cpp
AndyZe: This is actually a filter coefficient, related to cutoff frequency. Will add a comment about it

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
AndyZe: I think it has to be a shared_ptr to work as a plugin. unique_ptr causes this:

> error: no match for ‘operator=’ (operand types are ‘std::unique_ptr<online_signal_smoothing::SmoothingBaseClass>’ and ‘std::shared_ptr<online_signal_smoothing::SmoothingBaseClass>’)
>   179 |     smoother_ = smoothing_loader_.createSharedInstance(parameters_->smoothing_filter_plugin_name);

moveit_core/online_signal_smoothing/src/butterworth_filter.cpp
henningkayser: Sounds good!

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
henningkayser: There is `createUniqueInstance()` or `createUnmanagedInstance()` for this.

AndyZe: No luck.

createUniqueInstance:

```
error: no match for ‘operator=’ (operand types are ‘std::unique_ptr<online_signal_smoothing::SmoothingBaseClass>’ and ‘pluginlib::UniquePtr<online_signal_smoothing::SmoothingBaseClass>’ {aka ‘std::unique_ptr<online_signal_smoothing::SmoothingBaseClass, std::function<void(online_signal_smoothing::SmoothingBaseClass*)> >’})
  179 |     smoother_ = smoothing_loader_.createUniqueInstance(parameters_->smoothing_filter_plugin_name);
```

createUnmanagedInstance:

```
error: no match for ‘operator=’ (operand types are ‘std::unique_ptr<online_signal_smoothing::SmoothingBaseClass>’ and ‘online_signal_smoothing::SmoothingBaseClass*’)
  179 |     smoother_ = smoothing_loader_.createUnmanagedInstance(parameters_->smoothing_filter_plugin_name);
```

henningkayser: well, you would have to either store the UniquePtr type (createUniqueInstance) or reset the unique_ptr using the raw pointer returned by `createUnmanagedInstance`. But anyway, it's not a big deal.

henningkayser (approved):

JafarAbdi (approved):

moveit_core/CMakeLists.txt
```diff
@@ -158,35 +157,36 @@ message(STATUS " *** Building MoveIt ${MOVEIT_VERSION} ***")
...
-add_subdirectory(kinematics_metrics)
-add_subdirectory(dynamics_solver)
-add_subdirectory(collision_detection_bullet)
+add_subdirectory(online_signal_smoothing)
+add_subdirectory(trajectory_processing)
+add_subdirectory(transforms)
+add_subdirectory(utils)
+add_subdirectory(version)
 
-pluginlib_export_plugin_description_file(moveit_core collision_detector_bullet_description.xml)
```
henningkayser, line 189: I just noticed now that this was duplicate, thanks for cleaning that up!

moveit_core/CMakeLists.txt
```diff
@@ -158,35 +157,36 @@ message(STATUS " *** Building MoveIt ${MOVEIT_VERSION} ***")
...
+# TODO(andyz): why don't these compile when alphabetized with the rest?
```
henningkayser, line 186: Probably missing dependencies. For instance, `collision_detection` calls `ament_add_gtest` without calling `find_package(ament_cmake_gtest)`... I don't think it belongs into this PR, though.

moveit_core/package.xml
```diff
@@ -72,5 +72,6 @@
     <build_type>ament_cmake</build_type>
     <moveit_core plugin="${prefix}/collision_detector_fcl_description.xml"/>
     <moveit_core plugin="${prefix}/collision_detector_bullet_description.xml"/>
+    <moveit_core plugin="${prefix}/filter_plugin_butterworth.xml"/>
```
JafarAbdi, line 75: We don't need these exports in ROS2, it all happens in the CMake's pluginlib_export_plugin_description_file macro

JafarAbdi:
I'm working on fixing galactic

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -385,12 +404,16 @@ void ServoCalcs::calculateSingleIteration()
   }
 
   // If we should halt
-  if (!have_nonzero_command_)
+  if (!have_nonzero_command_ && !done_stopping_)
   {
-    suddenHalt(*joint_trajectory);
+    filteredHalt(*joint_trajectory);
     have_nonzero_twist_stamped_ = false;
     have_nonzero_joint_command_ = false;
```
nbbrooks, line 411:
```suggestion
```

ditto

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -385,12 +404,16 @@ void ServoCalcs::calculateSingleIteration()
   }
 
   // If we should halt
-  if (!have_nonzero_command_)
+  if (!have_nonzero_command_ && !done_stopping_)
   {
-    suddenHalt(*joint_trajectory);
+    filteredHalt(*joint_trajectory);
     have_nonzero_twist_stamped_ = false;
```
nbbrooks, line 410:
```suggestion
```

This should already be the case given line 354

```
  have_nonzero_command_ = have_nonzero_twist_stamped_ || have_nonzero_joint_command_;
```

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -385,12 +404,16 @@ void ServoCalcs::calculateSingleIteration()
...
     have_nonzero_joint_command_ = false;
   }
+  else
+  {
+    done_stopping_ = false;
```
nbbrooks, line 415: Wish there was a way to suggest changes to non-modified lines. Line 420 
```
  if (!have_nonzero_command_ && **done_stopping_** && (parameters_->num_outgoing_halt_msgs_to_publish != 0) &&
      (zero_velocity_count_ > parameters_->num_outgoing_halt_msgs_to_publish))
```

and line 442
```
  if (!have_nonzero_command_ && **done_stopping_**)
```

Need to check this var. Otherwise it can reach num_outgoing_halt_msgs_to_publish before filteredHalt actually completes and the last publishing point will be non-zero.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -829,40 +850,44 @@ ServoCalcs::enforcePositionLimits(sensor_msgs::msg::JointState& joint_state) con
...
+      {
+        done_stopping_ = false;
+      }
+    }
+    // If every joint is very close to stopped, round velocity to zero
+    if (done_stopping_)
+    {
+      std::fill(joint_trajectory.points[0].velocities.begin(), joint_trajectory.points[0].velocities.end(), 0);
+    }
   }
```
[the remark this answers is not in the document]
AndyZe: done!

henningkayser:
@AndyZe the commits I pushed worked locally for me using clang. However, I didn't get the tests to pass using the component node while components were no issue with the normal launch script. 8c6cbeb changes the tests to run with a standalone servo node (succeeded locally for me). I'm not sure what exactly causes the issues there though, might be worth investigating this further. I'd say if CI is happy now, you could squash the commit history a bit and do a rebase. Then lgtm.

henningkayser (approved):

Issue #536, opened 2021-07-02
Moving collision objects in RViz 


christianlandgraf:
### Description

Strange behaviour when loading and moving scene objects.

### Your environment
* ROS Distro: Foxy
* OS Version: Ubuntu 20.04
* Source build, master

### Steps to reproduce
When loading any collision object into the scene, moving it through interactive marker control or the position boxes does not work as expected... (see video below). The position of the interactive marker also changes . Do you have any idea why this is happening? I couldn't find the solution or a similar issue so far...

Thanks a lot!

https://user-images.githubusercontent.com/28567122/124273665-976fab80-db40-11eb-87e2-a2319758c9f5.mp4

JafarAbdi:
Hi @christianlandgraf, I can't reproduce this, do you mind sharing more info .? what commit hash are you using .?

christianlandgraf:
Hi @JafarAbdi  I'm on the latest commit b8813c34d7576ea32e6549b96a87072d2cff8930 and using the recommended setup from `moveit.repos`. It's quite a strange behaviour...

I am using a custom description file ([this one](https://github.com/PickNikRobotics/UR10e_welding_demo/blob/master/ipa_demo_cell_description/urdf/ipa_ur_demo.urdf.xacro)). If the robot is linked to some other frame, e.g. a table, it seems as if the interactive marker for controlling the object and the object itself are somehow working in different coordinate frames. I couldn't fully track it down yet, it's just a first suspicion.

It's also possible that I have to align the [srdf](https://github.com/PickNikRobotics/UR10e_welding_demo/blob/master/ipa_moveit_config/srdf/ur_macro.srdf.xacro) and the coresponding urdf.

christianlandgraf:
The error seem to appear when the fixed frame in Rviz and the origin of the layout in MoveIt are different. I'm not sure if this is a common bug or a mistake in my Rviz configuration.
![image](https://user-images.githubusercontent.com/28567122/129756091-42a632cd-b746-4c16-9c92-aba9ebb717c6.png)

JafarAbdi:
@christianlandgraf Could you please test https://github.com/ros-planning/moveit2/pull/680 ? 

Thanks!

christianlandgraf:
@JafarAbdi It works for me now, thanks a lot for fixing it!

PR #680, opened 2021-09-16
Fix object interactive marker in wrong pose after changing the fixed frame

JafarAbdi:
### Description

Fix: https://github.com/ros-planning/moveit2/issues/536

When we create the interactive marker displays we set the model frame as the reference frame, changing the fixed frame is causing the InteractiveMarkerFeedback poses to be specified w.r.t. the fixed frame so to fix it I converted the poses to be w.r.t. model frame

Working version:

https://user-images.githubusercontent.com/16278108/133537301-6dcaa09b-a80f-4475-9611-c636648c7ba1.mp4

Previous version:

https://user-images.githubusercontent.com/16278108/133537308-ed01893d-f596-40d0-af15-4a68c2a69815.mp4


christianlandgraf (approved):
I can approve that the PR fixes my issue #536

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp
```diff
@@ -1430,6 +1430,9 @@ void MotionPlanningDisplay::fixedFrameChanged()
   PlanningSceneDisplay::fixedFrameChanged();
   if (int_marker_display_)
     int_marker_display_->setFixedFrame(fixed_frame_);
+  // When the fixed frame changes we need to tell RViz to update the rendered interactive marker display
```
henningkayser, line 1433: Shouldn't this be part of `InteractiveMarkerDisplay::fixedFrameChanged()`? Btw, should we switch back to the [upstream implementation](https://github.com/ros2/rviz/blob/ros2/rviz_default_plugins/include/rviz_default_plugins/displays/interactive_markers/interactive_marker_display.hpp)?

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
```diff
@@ -432,21 +432,29 @@ void MotionPlanningFrame::collisionObjectChanged(QListWidgetItem* item)
 /* Receives feedback from the interactive marker and updates the shape pose in the world accordingly */
 void MotionPlanningFrame::imProcessFeedback(visualization_msgs::msg::InteractiveMarkerFeedback& feedback)
 {
+  if (!planning_display_->getPlanningSceneRO()->knowsFrameTransform(feedback.header.frame_id))
+  {
+    RCLCPP_ERROR_STREAM(LOGGER,
+                        "Frame `" << feedback.header.frame_id << "` unknown doesn't exists in the planning scene");
+  }
+  Eigen::Isometry3d fixed_frame_t_scene_marker;
```
henningkayser, line 440: hmm... what is the situation in ROS 1? Wouldn't this be needed there as well?

henningkayser (commented):
I'm fine with this fix. Just a couple question before getting this merged.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp
JafarAbdi: `scene_marker_`'s type is InteractiveMarker not InteractiveMarkerDisplay :D, it's already the case for `InteractiveMarkerDisplay::fixedFrameChanged()`

> Btw, should we switch back to the upstream implementation?

Yes, that's right, I'll work on pushing it upstream but that I think we need a multi thread executor for it to work https://github.com/ros2/rviz/blob/ros2/rviz_common/include/rviz_common/visualization_manager.hpp#L400

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
JafarAbdi: I tested it in ROS1, this bug doesn't exist there :confused:

henningkayser: How is the frame being handled in ROS 1 then? To me this looks like a bug that was introduced in a different place during conversion.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp
henningkayser: Stumbling over this again... since `requestPoseUpdate()` just feeds in the existing poses, I'd say that [updateReferencePose()](https://github.com/ros2/rviz/blob/cd1da34cfc01d9c6223c181ba6e7bf277ee4c1a8/rviz_default_plugins/src/rviz_default_plugins/displays/interactive_markers/interactive_marker.cpp#L600) is more appropriate here.

PR #683, opened 2021-09-17
Add jerk to the robot model

### LINKED ISSUE #574: Parse jerk limits from joint_limits.yaml and add jerk limits to Joint Model
Let's give MoveIt the functionality of processing jerk limits. They should be optional, just like acceleration limits are now.

AndyZe:
Now that we have some MoveIt capabilities that can process jerk limits, I think we should add jerk limits to the robot model. This simply adds the jerk parameters to joint_limits.yaml and adds the parsing code to robot_model_loader.cpp.

Fixes #574

### DIFF
--- a/moveit_ros/planning/robot_model_loader/CMakeLists.txt
+++ b/moveit_ros/planning/robot_model_loader/CMakeLists.txt
@@ -11,6 +11,7 @@ ament_target_dependencies(${MOVEIT_LIB_NAME}
   urdf
   Boost
   moveit_core
+  moveit_msgs
 )
 target_link_libraries(${MOVEIT_LIB_NAME}
   moveit_rdf_loader
--- a/moveit_core/robot_model/include/moveit/robot_model/joint_model.h
+++ b/moveit_core/robot_model/include/moveit/robot_model/joint_model.h
@@ -63,6 +63,9 @@ struct VariableBounds
     , min_acceleration_(0.0)
     , max_acceleration_(0.0)
     , acceleration_bounded_(false)
+    , min_jerk_(0.0)
+    , max_jerk_(0.0)
+    , jerk_bounded_(false)
   {
   }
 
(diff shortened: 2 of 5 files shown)


moveit_ros/planning/robot_model_loader/CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ ament_target_dependencies(${MOVEIT_LIB_NAME}
   urdf
   Boost
   moveit_core
+  moveit_msgs
```
nbbrooks, line 14: This CMakeLists is missing the `find_package` entries!

AndyZe: This is one of the many MoveIt packages that uses subdirectories, so the `find_package` is in the CMakeLists one level up.

nbbrooks (approved):
lgtm

JafarAbdi (commented):
I think we should also update [computeVariableBoundsMsg](https://github.com/ros-planning/moveit2/blob/main/moveit_core/robot_model/src/joint_model.cpp#L165) and [operator<<(std::ostream& out, const VariableBounds& b)](https://github.com/ros-planning/moveit2/blob/main/moveit_core/robot_model/src/joint_model.cpp#L221-L239) to include the new fields

I released moveit_msgs an hour ago but we need to add it to the `.repos` file until it gets synced to the Debians

henningkayser (approved):

henningkayser:
Changes look good to me. Let's hold off on merging this until the MoveIt release is done.

henningkayser (commented):
Could you add jerk to the `RobotTrajectory::print()` function (https://github.com/ros-planning/moveit2/pull/715) right [below efforts](https://github.com/ros-planning/moveit2/blob/main/moveit_core/robot_trajectory/src/robot_trajectory.cpp#L562)? LGTM after that

PR #684, opened 2021-09-17
Delete duplicate entry in Servo launch file

AndyZe:
This is a small mistake from #621. Probably mine.

### DIFF
--- a/moveit_ros/moveit_servo/launch/servo_example.launch.py
+++ b/moveit_ros/moveit_servo/launch/servo_example.launch.py
@@ -102,6 +102,7 @@ def generate_launch_description():
         executable="component_container",
         composable_node_descriptions=[
             # Example of launching Servo as a node component
+            # Assuming ROS2 intraprocess communications works well, this is a more efficient way.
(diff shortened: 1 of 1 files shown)


moveit_ros/moveit_servo/launch/servo_example.launch.py
```diff
@@ -125,17 +125,6 @@ def generate_launch_description():
                 name="static_tf2_broadcaster",
                 parameters=[{"child_frame_id": "/panda_link0", "frame_id": "/world"}],
             ),
-            ComposableNode(
```
AndyZe, line 128: Line 155 handles it

nbbrooks: Could we add a comment explaining why we are launching servo server as a separate node?

Also should we delete the comment block "# Example of launching Servo as a node component" ?

AndyZe: I added some better comments  :+1:

PR #689, opened 2021-09-20
Fix source versions of ros2_control, ros2_controllers

henningkayser:
Fixes CI by only installing released versions for Galactic. This should also prevent sync issues in the future.

### DIFF
--- a/moveit2_galactic.repos
+++ b/moveit2_galactic.repos
@@ -3,8 +3,8 @@ repositories:
   ros2_control:
     type: git
     url: https://github.com/ros-controls/ros2_control
-    version: master
+    version: 0.8.0
   ros2_controllers:
     type: git
     url: https://github.com/ros-controls/ros2_controllers
-    version: master
+    version: 0.5.0


tylerjw:
Failing for the galactic ruckig issue.

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -76,9 +76,9 @@ bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajecto
 
   // Instantiate the smoother
   double timestep = trajectory.getAverageSegmentDuration();
-  ruckig::Ruckig<0> ruckig{ num_dof, timestep };
-  ruckig::InputParameter<0> ruckig_input{ num_dof };
-  ruckig::OutputParameter<0> ruckig_output{ num_dof };
+  ruckig::Ruckig<0> ruckig(num_dof, timestep);
+  ruckig::InputParameter<0> ruckig_input(num_dof);
+  ruckig::OutputParameter<0> ruckig_output(num_dof);
```
tylerjw, line 81: Sweet, I hope this works.  I had no idea there would be a way to fix this without waiting for the next release.

tylerjw (approved):

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
henningkayser: nope, it doesn't. Did the last Ruckig release fail for Galactic?

tylerjw: I don't think so because it is in galactic-testing, I think the last release just missed the sync.

henningkayser: How did https://github.com/ros-planning/moveit2/pull/571 pass then? Also, there were a couple of main builds that succeeded with Galactic

henningkayser: [Ruckig was just synced](https://discourse.ros.org/t/new-packages-for-galactic-geochelone-2021-09-20/22374), so this should be fixed. I'm removing the changes from this PR.

JafarAbdi (approved):

henningkayser:
I just merged the first commit, the second warning fix went into #690

PR #690, opened 2021-09-21
Fix cmake warnings

henningkayser:
This fixes the build warnings discovered here https://github.com/ros-planning/moveit2/runs/3662305871?check_suite_focus=true#step:8:689. Should we enforce the build to be free of warnings?

### DIFF
--- a/moveit_kinematics/test/test_kinematics_plugin.cpp
+++ b/moveit_kinematics/test/test_kinematics_plugin.cpp
@@ -476,8 +476,8 @@ TEST_F(KinematicsTest, unitIK)
   };
 
   std::vector<double> ground_truth, pose_values;
-  constexpr char post_type_relative[] = "relative";
-  constexpr char post_type_absolute[] = "absolute";
+  constexpr char pose_type_relative[] = "relative";
+  constexpr char pose_type_absolute[] = "absolute";
 
   /* process tests definitions on parameter server of the form
      pose_1:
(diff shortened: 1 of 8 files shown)


destogl (approved):
This looks good. I suggest renaming the PR to something like "Fix build warnings". I found "cmake" in the title somewhat misleading.

There is also one more `clang-tidy` issue.

JafarAbdi (commented):
> Should we enforce the build to be free of warnings?

+1, the only problem with this is it may force us to branch off galactic since some features are deprecated in rolling and the alternatives don't exist in galactic

v4hn:
> Should we enforce the build to be free of warnings

We do that [for moveit](https://github.com/ros-planning/moveit/blob/master/.github/workflows/ci.yaml#L41), so I think it's a very good idea to enforce it for moveit2 as well!

> +1, the only problem with this is it may force us to branch off galactic since some features are deprecated in rolling and the alternatives don't exist in galactic

I would argue that using deprecated API should still be avoided in the newer distro, possibly using `#ifdef` approaches to support both alternatives.
That way, you can add clear TODOs to address when galactic support is dropped in the branch.

moveit_kinematics/test/test_kinematics_plugin.cpp
```diff
@@ -477,8 +476,8 @@ TEST_F(KinematicsTest, unitIK)
   };
 
   std::vector<double> ground_truth, pose_values;
-  constexpr char POSE_TYPE_RELATIVE[] = "relative";
-  constexpr char POSE_TYPE_ABSOLUTE[] = "absolute";
+  constexpr char post_type_relative[] = "relative";
```
DLu, line 479: Note: My PR was dumb and had a typo. Should be pose_type_relative (not post)

JafarAbdi (approved):

PR #692, opened 2021-09-21
Info -> Debug

vatanaksoytezer:
Revert to original log levels of some functions that was changed in https://github.com/ros-planning/moveit2/pull/598/files#diff-25706feb6b355123a5ae63dfa987a1cdd5754087392e9623960ed74a31b56d75R59.

### DIFF
--- a/moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
+++ b/moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
@@ -56,7 +56,7 @@ rclcpp::Parameter declare_parameter(const rclcpp::Node::SharedPtr& node, const s
     node->declare_parameter(parameter_name, ParamType);
   rclcpp::Parameter parameter;
   if (!node->get_parameter(parameter_name, parameter))
-    RCLCPP_INFO_STREAM(LOGGER, "Parameter `" << parameter_name << "` doesn't exists");
+    RCLCPP_DEBUG_STREAM(LOGGER, "Parameter `" << parameter_name << "` doesn't exists");
   return parameter;
 }
 class KinematicsPluginLoader::KinematicsLoaderImpl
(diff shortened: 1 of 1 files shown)


JafarAbdi (approved):

vatanaksoytezer:
> Should we hold off merging this in until we have fix for the different expected kinematics namespace this revealed?

That's a good question: Maybe @JafarAbdi or @henningkayser can chime in?

v4hn (approved):
> Should we hold off merging this in until we have fix for the different expected kinematics namespace this revealed?

The logs are not gone in this patch, so I don't see how the changed log-level does not reveal something anymore?
I expect you can configure the log-levels in ROS2 similarly to ROS's `ROSCONSOLE_CONFIG_FILE`? If involved developers do not use this approach I would say they should set it up in their workflows to avoid spamming logs for everyone.

tylerjw:
> I expect you can configure the log-levels in ROS2 similarly to ROS's ROSCONSOLE_CONFIG_FILE?

@v4hn last time I tried this was not yet a working feature in ROS 2.

nbbrooks:
> > I expect you can configure the log-levels in ROS2 similarly to ROS's ROSCONSOLE_CONFIG_FILE?
> 
> @v4hn last time I tried this was not yet a working feature in ROS 2.

Yeah I also don't think this is implemented yet. We can always change it back for debugging, I just think if we push this in it will be easy to ignore the loading issue (I'm unclear on how much of an issue the loading failure is though)

v4hn:
Excuse me? :confounded: 

It seems it's still only supported to set [the global log level](https://answers.ros.org/question/311471/selecting-log-level-in-ros2-launch-file/) on startup, but not the level of individual namespaces.
I really shouldn't be surprised about this, but I still am. MoveIt uses this feature extensively, so we (should) have a strong interest in implementing support for it upstream!

PR #148, opened 2019-12-20
Port occupancy_map_monitor to ROS2

RoboticsYY:
#112 is blocked by `occupancy_map_monitor`, please see https://github.com/ros-planning/moveit2/blob/fa81678b965eafcb1e8c72018ba4d99ab2593255/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h#L45

This work followed AcutronicRobotics' porting, but didn't use the original commits. Because `occupancy_map_monitor` was moved from `moveit_ros` to `moveit_ros_perception`([ref](https://github.com/AcutronicRobotics/moveit2/tree/master/moveit_ros/perception/occupancy_map_monitor)), the original commits are very difficult to directly use.

PR #697, opened 2021-09-22
Update Maintainers of MoveIt package

davetcoleman:
I'm getting a lot of noisy emails from Jenkins such as:

>Build failed in Jenkins: Fdev__moveit__ubuntu_focal_amd64 #148

But I do not have bandwidth for this anymore. I'm also guessing @mikeferguson @IanTheEngineer @130s are not looking at this either, since we have not seen them in a long time.

I'd like to instead replace us with @henningkayser @tylerjw who are very actively involved.

I've also added myself as an author of MoveIt, which is mostly an ego thing that I'm fine removing also if it is judged my involvement at Willow Garage since before MoveIt was released as beta is not enough. I'm cool either way but this is clearly my preference.


tylerjw (approved):

130s (approved):
Thanks.

If this would stop the notification from jenkins that'll be great too (my impression from the last I tried to do that a few years ago, updating package.xml may have not been enough though).

mikeferguson:
Yeah - that fetchrobotics email hasn't worked in about 4 years now...

mikeferguson:
> If this would stop the notification from jenkins that'll be great too (my impression from the last I tried to do that a few years ago, updating package.xml may have not been enough though).

If I recall correctly, you have to do a release as well to get the buildfarm to fully update

henningkayser:
Makes sense to me. Shouldn't this be changed for the other `package.xml` files as well?

v4hn (approved):
> Makes sense to me. Shouldn't this be changed for the other package.xml files as well?

Yes, but it depends on Dave's preferences of course. @davetcoleman Hope you got the cycles to clean these entries consistently across all packages. You probably want to have similar changes in the moveit repository as well?

It makes sense to add you as an author in the meta-package. You are listed as author of MSA already which is a part of the stack.

davetcoleman:
@v4hn can't tell if you're being sarcastic?

It would be great to have all the package.xml files updated. I did this edit through the Github webpage editor to address the specific issue of emails I was getting, it's a bit tougher for me to edit other instances as I don't have the code checked out.

v4hn:
> can't tell if you're being sarcastic?

I'm not. it's entirely your decision on where you want to be listed.
I'm surprised this is still not merged.

It would be great if you could go through the packages systematically and adapt the maintainers as you did here.
`git clone` + `grep` takes something like 5 minutes.
If you can't do it, I'm also ok with merging this standalone.

henningkayser:
I pushed a commit that
    * Removes Dave's maintainer entries with with Tyler,Henning
    * Adds Dave as author to moveit_core, moveit_ros
    * Adds missing authors to moveit_common, run_move_group, run_moeit_cpp
    * Removes entries of inactive maintainers

I would still like to make the entries more accurate from PickNik's side (add more of Jafar, Andy, hopefully soon Vatan), but we can still update these later.

henningkayser:
xmllint wasn't happy with the changes. so I disabled the xml schema for the "broken" files in 5904793 for better diff readability.
#779 adds the ROS format 3 schema and fixes it for all `package.xml` files.

tylerjw (approved):

PR #699, opened 2021-09-22
Switch back to using master branch of ros2_control repos

tylerjw:
### Description

This is a test to see if we can switch back to using the ros2_control repo master branches for Galactic.  I was told today in the WG meeting that it should now work again.

### DIFF
--- a/moveit2_galactic.repos
+++ b/moveit2_galactic.repos
@@ -3,8 +3,8 @@ repositories:
   ros2_control:
     type: git
     url: https://github.com/ros-controls/ros2_control
-    version: 0.8.0
+    version: master
   ros2_controllers:
     type: git
     url: https://github.com/ros-controls/ros2_controllers
-    version: 0.5.0
+    version: master


tylerjw:
@AndyZe when switching from the tagged versions of ros2_control to their master branch here the pose tracker test seems to fail.  Do you know where they posted their latest migration guides?  Is there a chance all I need to do to get this working is update a launch file or something like that?  The error is that the complete robot state is not yet known when servo is trying to start.

henningkayser:
It should work, but do we really want to risk having the same issue again and again? Imo we should depend on the released version and remove the repos once they are available as debians. I don't see the point in having to test upstream changes for other repos, especially if they are run dependencies.

vatanaksoytezer:
> It should work, but do we really want to risk having the same issue again and again? Imo we should depend on the released version and remove the repos once they are available as debians. I don't see the point in having to test upstream changes for other repos, especially if they are run dependencies.

I completely agree with this. We should not care about ros2-controls current master but rather depend on a release. But technically they can also break releases as well 🤔.

tylerjw:
I think for our own sanity I agree.  Once they release that is essentially what we will do anyway (depend on a release), even on rolling.

PR #704, opened 2021-09-24
Forward ports recent dockerfile changes on foxy

vatanaksoytezer:
Forward port of dockerfile changes in https://github.com/ros-planning/moveit2/pull/698. Unblocks https://github.com/ros-planning/moveit2_tutorials/pull/120

PR #705, opened 2021-09-24
Remove extra catkin variable in dockerfile

vatanaksoytezer:
Per https://github.com/ros-planning/moveit2/pull/704#discussion_r715644869

### DIFF
--- a/.docker/source/Dockerfile
+++ b/.docker/source/Dockerfile
@@ -9,8 +9,6 @@ MAINTAINER Robert Haschke rhaschke@techfak.uni-bielefeld.de
 
 # Export ROS_UNDERLAY for downstream docker containers
 ENV ROS_UNDERLAY /root/ws_moveit/install
-# Environment variable used in instructions on moveit.ros.org website for running clang-tidy
-ENV CATKIN_WS $(realpath $ROS_UNDERLAY/..)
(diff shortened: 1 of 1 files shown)


rhaschke (commented):
I'm actually not convinced to remove this, both here and in `main`.
We are using this environment variable in the clang-tidy section on this page:
https://moveit.ros.org/documentation/contributing/code

It doesn't really harm to keep it and the comment clearly explains why we have it here.
@vatanaksoytezer, @henningkayser

vatanaksoytezer:
> I'm actually not convinced to remove this, both here and in `main`.
> We are using this environment variable in the clang-tidy section on this page:
> https://moveit.ros.org/documentation/contributing/code

I think this bit explains that this variable should be replaced with your workspace in the moveit website:

> Format your source code if its in some directory such as the catkin_ws (feel free to change keywords catkin_ws):

I think making a variable named `CATKIN_WS` can be a bit confusing and unintuitive for future people looking at this file. I would rather clarify the documentation a little bit tbh.

tylerjw:
Closing as it was decided we'd do something other than what is in this PR. I created a new issue for the thing that we decided to do: https://github.com/ros-planning/moveit2/issues/1727

PR #682, opened 2021-09-17
Add a Ruckig jerk-limited smoothing plugin, mostly for Servo

AndyZe:
This builds off of PR #676, so please review that one first.

Add a Ruckig jerk-limited smoothing plugin. Right now, this can easily be used in Servo. In the future, it might be useful in hybrid planning.

(Sorry for the flood of smoothing-related PR's recently. This is being driven by client work.)

Here are videos of the Ruckig filter (top) and the Butterworth filter. They look pretty much the same to my eye.

https://user-images.githubusercontent.com/11284393/133804964-abc72f66-9c05-432a-a58f-34c28e5ea12e.mp4

https://user-images.githubusercontent.com/11284393/133804979-188f8027-b478-4512-8444-1533719151dc.mp4

@destogl

PR #706, opened 2021-09-24
Add a jerk-limited Ruckig smoothing plugin

AndyZe:
An update to #682.

Based on #676, so please review that one first.


AndyZe:
Closing because we've recently found out, Ruckig isn't really suited to this type of application:

https://github.com/pantor/ruckig/issues/74#issuecomment-929922333

PR #707, opened 2021-09-27
Skip publishing to Servo topics if input commands are stale

schornakj:
### Description

While the servo update loop currently includes a check to see if the latest messages on the joint and Cartesian command topics are up-to-date, this does not actually stop Servo from continuing to publish joint control messages even if the commands are stale. If both command topics are stale, the `joint_trajectory` message used to create the control messages is [populated with the last-commanded positions and zero velocities](https://github.com/ros-planning/moveit2/blob/d184714751b347a2389cf1e8742b1be94eed63b8/moveit_ros/moveit_servo/src/servo_calcs.cpp#L376-L384). This can cause problems if Servo is paused and another method is used to move the robot, since when Servo is unpaused it will publish a control message to the last-commanded position, which could be a significant distance from the robot's current state.

This PR adds an additional check to prevent publishing control messages if both command topics are stale. Control publishing will resume once a new joint or Carteisan command message is received.

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/src/servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/src/servo_calcs.cpp
@@ -383,14 +383,6 @@ void ServoCalcs::calculateSingleIteration()
     }
   }
 
-  // Print a warning to the user if both are stale
-  if (twist_command_is_stale_ && joint_command_is_stale_)
-  {
-    rclcpp::Clock& clock = *node_->get_clock();
-    RCLCPP_WARN_STREAM_THROTTLE(LOGGER, clock, ROS_LOG_THROTTLE_PERIOD,
-                                "Stale command. Try a larger 'incoming_command_timeout' parameter?");
-  }
-
   // If we should halt
   if (!have_nonzero_command_)
   {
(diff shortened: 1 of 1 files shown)


moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -408,6 +408,14 @@ void ServoCalcs::calculateSingleIteration()
     rclcpp::Clock& clock = *node_->get_clock();
     RCLCPP_DEBUG_STREAM_THROTTLE(LOGGER, clock, ROS_LOG_THROTTLE_PERIOD, "All-zero command. Doing nothing.");
   }
+  // Skip servoing publication if both types of commands are stale.
+  else if (twist_command_is_stale_ && joint_command_is_stale_)
+  {
+    ok_to_publish_ = false;
+    rclcpp::Clock& clock = *node_->get_clock();
+    RCLCPP_DEBUG_STREAM_THROTTLE(LOGGER, clock, ROS_LOG_THROTTLE_PERIOD,
```
AndyZe, line 416: Please delete the block of code at L386-392 since what you've added here does the same thing.

AndyZe:
Have tested -- this works well :+1:

AndyZe (approved):

MarqRazz:
stale_servo
We are still seeing `[component_container-12] [moveit_servo.servo_calcs]: Stale command. Try a larger 'incoming_command_timeout' parameter?`

AndyZe:
@MarqRazz are you sure you built it since checking out the correct commit? Maybe you have a debian installed that is being used instead?  (Try `sudo apt remove ros-rolling-moveit*`)

I don't see how that line could be printing since Joe deleted it

MarqRazz:
Fixed the build issue. The terminal output is now as expected.

MarqRazz (approved):

JafarAbdi (approved):

PR #710, opened 2021-09-29
moveit_ros_control_interface: Fix dangling reference

JafarAbdi:
### Checklist

### DIFF
--- a/moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
+++ b/moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
@@ -138,7 +138,8 @@ class MoveItControllerManager : public moveit_controller_manager::MoveItControll
     managed_controllers_.clear();
     active_controllers_.clear();
 
-    for (const controller_manager_msgs::msg::ControllerState& controller : result_future.get()->controller)
+    const auto& result = result_future.get();
+    for (const controller_manager_msgs::msg::ControllerState& controller : result->controller)
     {
       if (isActive(controller))
       {


henningkayser (commented):
Are you sure this is a dangling reference? From the code I don't see why this would happen (`result_future` doesn't go out of scope here). Not knowing your issue, I'd suspect that we are getting an [invalid future](https://en.cppreference.com/w/cpp/thread/future/valid) while the code only checks for timeout.
Side note: is `async_send_request()` guaranteed to return a valid future?

JafarAbdi:
> Are you sure this is a dangling reference? From the code I don't see why this would happen (`result_future` doesn't go out of scope here). Not knowing your issue, I'd suspect that we are getting an [invalid future](https://en.cppreference.com/w/cpp/thread/future/valid) while the code only checks for timeout.

Yes, I'm pretty sure, see [P2012R1](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2012r1.pdf) for why this's happening

henningkayser:
> > Are you sure this is a dangling reference? From the code I don't see why this would happen (`result_future` doesn't go out of scope here). Not knowing your issue, I'd suspect that we are getting an [invalid future](https://en.cppreference.com/w/cpp/thread/future/valid) while the code only checks for timeout.
> 
> Yes, I'm pretty sure, see [P2012R1](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2012r1.pdf) for why this's happening

I see, so `get()` returned a temporary copy instead of a reference so that the message ends up being a dangling reference...

I think it might still make sense to check for `valid()` in general

henningkayser (approved):

PR #719, opened 2021-10-03
Draft: Initialize servo in the constructor

tylerjw:
This PR is based on 
- Rename ServoServer to ServerNode

Please review the changes in the second commit in this PR.  I will rebase after the rename is merged.

### Description

This moves the initialization of the servo node into the constructor.  The reason it was in a start service previously is that we needed a shared_ptr to the Node object to initialize the various subsystems (psm, calcs, etc).  When reviewing the hybrid planner PR I found this nugget in the ros documentation: https://docs.ros2.org/foxy/api/rclcpp_components/register__node__macro_8hpp.html

### intra process comms

In the `servo_launch_test_common.py` intra process comms was not enabled as there was a typo.  `extra_arguments=[{'use_intra_process_comms' : True}]` was missing the s at the end of comms.  I tried enabling it, and moving the ServoNode into the same container, however it produced these errors on startup for each of the component nodes:
```
2: [component_container_mt-2] [ERROR] [1633289373.749679012] [test_servo_integration_container]: Component constructor threw an exception: intraprocess communication is not allowed with a zero qos history depth value
2: [ERROR] [launch_ros.actions.load_composable_nodes]: Failed to load node 'servo_node' of type 'moveit_servo::ServoNode' in container '/test_servo_integration_container': Component constructor threw an exception: intraprocess communication is not allowed with a zero qos history depth value
```
I don't understand what this means entirely but we should research it and figure out what policies we should have for communication and write some sort of guide as to what nodes can enable `use_intra_process_comms`.  Based on a thread on ros discourse recently I learned that it is a serious pessimisation to have a mix of enabled and disabled use_intra_process_comms and that it is much better to have it not enabled on any node as opposed to enabling it on the nodes you can.  It seems though that all three of the component nodes in the integration testing (`robot_state_publisher::RobotStatePublisher`, `tf2_ros::StaticTransformBroadcasterNode`, and `moveit_servo::ServoNode`) cannot enable it.  Being able to use this feature seems to be the key feature motivating component nodes.  It might be worth benchmarking without it to see if it is more performant to have separate processes.

### multi-threaded component container

I noticed we were still using the single-threaded component container.  As the multi-threaded executor has been greatly improved in Rolling we should transition to using the multi-threaded one.  I made that change here for the integration tests.

### servo status codes for when servo is not publishing output (early exit from the publish function)

I added some status codes for the various reasons servo might not be producing output.  This helped me debug an issue in one of the integration tests and I think it is a good chance but a reviewer should look at how we are using status to see if we have a consistent strategy for it.

### default QoS

I changed all the QoS settings to be the default for the middleware.  This change should make them all perform in a way that is most similar to ROS 1.  Here is the documentation on QoS for Rolling: https://docs.ros.org/en/rolling/Concepts/About-Quality-of-Service-Settings.html  A reviewer should verify this is a good approach.

### early exit for wait_for_servo_commands_ or paused_

I noticed that there was an opportunity that wait_for_servo_commands_ might be set false by an up-to-date message and we would wait for one more cycle before processing that input.  This would create a small delay between the first input and the first output of the period that servo is running at.  I made a change in the if statement for the early exit to make it continue if wait_for_servo_commands_ is cleared by an incoming message with a timestamp.

### tutorial updates and migration

I will add deprecation notices in the start/stop services and have them always fail as they are no longer needed.  I will also update the tutorial on this point.

### removing start/stop service

Please let me know if there is a reason we should not remove the start/stop service.  It seems that the pause/unpause interface should be able to accomplish enabling/disabling servo-based control in the most performant way and the primary reason for start/stop was initialization after construction.  Please let me know if there was another use case for this I don't know about.

### Checklist


tylerjw:
Here is the first one: https://github.com/ros-planning/moveit2/pull/721

moveit_ros/moveit_servo/include/moveit_servo/servo_node.h
```diff
@@ -45,18 +45,22 @@
 
 namespace moveit_servo
 {
-class ServoNode : public rclcpp::Node
+class ServoNode
```
henningkayser, line 48: I think it's a very reasonable step to not inherit from `rclcpp::Node` anymore, even if it's recommended in some of the tutorials. This makes the interface much cleaner and allows us to more easily implement DI or middleware wrappers in the future.

moveit_ros/moveit_servo/src/collision_check.cpp
```diff
@@ -72,10 +72,10 @@ CollisionCheck::CollisionCheck(rclcpp::Node::SharedPtr node, const ServoParamete
 
   // ROS pubs/subs
   collision_velocity_scale_pub_ =
-      node_->create_publisher<std_msgs::msg::Float64>("~/collision_velocity_scale", ROS_QUEUE_SIZE);
+      node_->create_publisher<std_msgs::msg::Float64>("~/collision_velocity_scale", rclcpp::SystemDefaultsQoS());
 
   worst_case_stop_time_sub_ = node_->create_subscription<std_msgs::msg::Float64>(
-      "~/worst_case_stop_time", ROS_QUEUE_SIZE,
+      "~/worst_case_stop_time", rclcpp::SystemDefaultsQoS(),
```
henningkayser, line 78: Did you squash #721 into this commit?

moveit_ros/moveit_servo/include/moveit_servo/servo_node.h
henningkayser: Would it make sense to provide a constructor that accepts a node as an argument `ServoNode()`? It seems a bit counter-intuitive to construct a node inside this class, but then depend on some external function spinning it.

PR #721, opened 2021-10-04
Use rclcpp::SystemDefaultsQoS in servo

tylerjw:
### Description

This standardizes the QoS used by servo to be the default that the ROS documentation recommends for pub/sub interfaces.

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo_parameters.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo_parameters.h
@@ -38,16 +38,14 @@
 
 #pragma once
 
-#include <thread>
 #include <mutex>
+#include <thread>
 #include <vector>
+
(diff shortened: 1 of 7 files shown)


henningkayser (commented):
This is so much cleaner than the magic numbers. What is the typical system default profile? With [rmw_qos_profile_default](https://github.com/ros2/rmw/blob/master/rmw/include/rmw/qos_profiles.h#L51) we would at least make sure the settings are reasonable. I don't have too much insight what to use here, though.

tylerjw:
I tried reading into the QoS settings and the code inside rclcpp for it and it seems this system default is what is recommended for most pub/sub.  I created this issue because for a while I was trying to add one for rmw_qos_profile_default: https://answers.ros.org/question/387911/should-there-be-a-defaultqos-or-default-constructor-for-qos/

henningkayser (approved):

PR #722, opened 2021-10-04
Disable use_intra_process_comms in servo launch files

tylerjw:
### Description

This is a fix for Rolling as enabling this causes the component container to die with this error:

```
[ERROR] [test_servo_integration_container]: Component constructor threw an exception: intraprocess communication is not allowed with a zero qos history depth value
```

Note there was also a typo in `servo_launch_test_common.py` (missing a s).

On the changes in `servo_example.launch.py`.  Even if we could enable it for the two components it is enabled for it is a major pessimization to enable it for only some of the nodes within a container.

### Checklist


vatanaksoytezer (commented):
~We should also change the ones in tests as well: https://github.com/ros-planning/moveit2/blob/main/moveit_ros/moveit_servo/test/launch/servo_launch_test_common.py#L150~ Sorry my bad.

henningkayser (commented):
You should probably remove [this warning](https://github.com/tylerjw/moveit2/blob/servo_disable_intra_process_comms/moveit_ros/moveit_servo/src/servo_node.cpp#L49), also. Do we even have a history depth of 0? Should we enforce something else in the profile and suggest enabling intra process comms on all components?

tylerjw:
Somewhere in moveit (I'm assuming) we do have a history depth of 0 for some topic and that is the cause of this.

henningkayser (approved):

PR #723, opened 2021-10-04
Use multi-threaded component container in servo"

tylerjw:
Based off:
* https://github.com/ros-planning/moveit2/pull/722

### Description

As the multi-threaded component container (and executor) are now usable in ROS Rolling we should switch to using them with servo and its tests.

### Checklist


AndyZe (commented):
I think this type of change should at least be tested with the tutorial to verify that publications to the controller have the right rate. (Which I would normally do but I guess it will have to wait a week.)

AndyZe:
I`m fine merging this and closing the other PR

Issue #726, opened 2021-10-07
Let the MoveItControllerManager manage more controllers


schornakj:
As currently implemented, the MoveItControllerManager plugin can only manage the activation and deactivation of controllers that have a MoveIt controller allocator plugin defined, due to [this bit of code](https://github.com/ros-planning/moveit2/blob/7292b9c20894dd65189d1ce72ee06981a520fd2f/moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp#L153-L163).

However, it would also be useful to allow the MoveItControllerManager to also manage controllers that are not associated with MoveIt trajectory execution. For example, MoveIt Servo is often configured to use a streaming position or velocity controller, so switching between Servo control and joint trajectory execution requires managing these controllers separately from the execution of trajectories through the TrajectoryExecutionManager. The TrajectoryExecutionManager already has a lot of information about the relationships between different available controllers and their potentially-overlapping associations with joint interfaces, and it would be great to be able to use that more flexibly.

It's fairly straightforward to allow the MoveItControllerManager to manage more controllers -- the lines of code I linked above would need to be modified to be like this so the controller is added to `managed_controllers_` regardless of if a plugin is available for it:

```
std::string absname = getAbsName(controller.name);
auto controller_it = managed_controllers_.insert(std::make_pair(absname, controller)).first;  // with namespace
auto& claimed_interfaces = controller_it->second.claimed_interfaces;
std::transform(claimed_interfaces.cbegin(), claimed_interfaces.cend(), claimed_interfaces.begin(),
               [](const std::string& claimed_interface) {
                 return parseJointNameFromResource(claimed_interface);
               });

if (loader_.isClassAvailable(controller.type))
{
  allocate(absname, controller_it->second);
}
```

One way to use this expanded functionality would be to create a MoveGroup capability that activates a specified controller through the TrajectoryExecutionManager.anager.

Does anyone have thoughts about this?

tylerjw:
My only thought is we probably should have a better name.

https://blog.codinghorror.com/i-shall-call-it-somethingmanager/

I like the idea though.

tylerjw:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MoveIt-61675936b391800012280f6d/issues/ros-planning/moveit2/726?planning-poker) with ZenHub @abake48 @Abishalini @AndyZe @DLu @henningkayser @JafarAbdi @vatanaksoytezer

[...]

PR #731, opened 2021-10-08
Allow MoveItControllerManager to manage more controllers

schornakj:
### Description

Implements #726.

This depends on [ros2_control #544](https://github.com/ros-controls/ros2_control/pull/544) to work correctly, since that PR allows the ros2_control controller manager to list joints associated with currently-inactive controllers.

- Add controllers to the `managed_controllers_` list even if they do not have a corresponding manager plugin defined. This allows the `TrajectoryExecutionManager::ensureActiveControllers` function to activate a wider variety of controllers, such as streaming position or velocity controllers.
- Add controller discovery to the initialization function of `MoveItControllerManager` to make sure that info about the available controllers is always available. Previously this was tied to the manager's introspection functions.
- Add some logging to help understand failure cases.

### Checklist

### DIFF
--- a/moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
+++ b/moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
@@ -157,17 +157,18 @@ class MoveItControllerManager : public moveit_controller_manager::MoveItControll
                        });
       }
 
+      // All controllers are added to the map of managed controllers, independent of whether they are currently active or not.
+      std::string absname = getAbsName(controller.name);
+      auto controller_it = managed_controllers_.insert(std::make_pair(absname, controller)).first;  // with namespace
+      // Get the names of the interfaces that would be claimed by this currently-inactive controller if it was activated.
+      auto& required_interfaces = controller_it->second.required_command_interfaces;
+      std::transform(required_interfaces.cbegin(), required_interfaces.cend(), required_interfaces.begin(),
+                     [](const std::string& required_interface) {
+                       return parseJointNameFromResource(required_interface);
+                     });
       // Instantiate a controller handle if one is available for this type of controller.
       if (loader_.isClassAvailable(controller.type))
       {
-        std::string absname = getAbsName(controller.name);
-        auto controller_it = managed_controllers_.insert(std::make_pair(absname, controller)).first;  // with namespace
-        // Get the names of the interfaces that would be claimed by this currently-inactive controller if it was activated.
-        auto& required_interfaces = controller_it->second.required_command_interfaces;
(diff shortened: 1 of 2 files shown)


moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -236,6 +236,9 @@ class MoveItControllerManager : public moveit_controller_manager::MoveItControll
         getAbsName("controller_manager/list_controllers"));
     switch_controller_service_ = node_->create_client<controller_manager_msgs::srv::SwitchController>(
         getAbsName("controller_manager/switch_controller"));
+
+    std::unique_lock<std::mutex> lock(controllers_mutex_);
```
tylerjw, line 240:
```suggestion
    std::lock_guard<std::mutex> lock(controllers_mutex_);
```

I don't see you manually unlocking this or using it with a condition variable so you should be able to use a unique_lock.

schornakj:
I split this up and made #785 for the changes that are just needed to enable MoveIt-managed controller switching. This PR now just contains the change to allow switching to a controller that doesn't have an available handle plugin.

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -360,15 +373,17 @@ class MoveItControllerManager : public moveit_controller_manager::MoveItControll
...
+    // For each controller to activate, find conflicts between the interfaces required by that controller and the
+    // interfaces claimed by currently-active controllers.
     for (const std::string& it : activate)
     {
       ControllersMap::iterator c = managed_controllers_.find(it);
       if (c != managed_controllers_.end())
       {  // controller belongs to this manager
         request->start_controllers.push_back(c->second.name);
-        for (const auto& claimed_resource : c->second.claimed_interfaces)
+        for (const auto& required_resource : c->second.required_command_interfaces)
```
AndyZe, line 384: What if Controller A requires the **position** command interface for Joint 1 and Controller B requires the **velocity** command interface for Joint1. Will that be seen as a conflict here? I think it should be.

(Refer to the [example here](https://github.com/ros-controls/ros2_control_demos/blob/37b3338b22ce99200791b191df647256d51f233e/ros2_control_demo_description/rrbot_description/ros2_control/rrbot_system_multi_interface.ros2_control.xacro#L30))

schornakj: This might be better to address as part of #785, since the commit that introduces the change you commented on is part of that PR.

schornakj:
I'm going to close this PR. It's replaced by #792, which provides a way to enable the functionality I'm looking for with fewer functional changes to the controller management logic. #785 brought in the other important updates that were originally here as well.

PR #733, opened 2021-10-11
Update version of pre-commit action

tylerjw:
### Description

We will need to back-port this change everywhere that uses pre-commit github action: https://github.com/pre-commit/action

The latest version stopped working this morning.


tylerjw:
> > The latest version stopped working this morning.
> 
> scream Do you have an example run for this?

False alarm, it stopped working when @Abishalini copied it into another repo that had some python file that caused problems.  Either way we should probably update this.

henningkayser:
Looks like pre-commit is deprecated. Should we switch to https://pre-commit.ci/?

Issue #755, opened 2021-10-25
Latched Strings for URDF and SRDF


DLu:
With the death of the one centralized parameter server to rule them all in ROS 2, it has become more cumbersome to have single instances of global parameters that everyone can use, i.e. often multiple nodes will need to individually load `/robot_description` into their parameter namespace. 

[`robot_state_publisher`](https://github.com/ros/robot_state_publisher/blob/361249090bdd8b089eb45318fd537034c6c42318/src/robot_state_publisher.cpp#L132) gets around that problem by publishing a `std_msgs::msg::String` message with the entire URDF file that other files can then use. 

Right now, [`rdf_loader`](https://github.com/ros-planning/moveit2/blob/df6dc7c340afc9e79c15f3b92a93db936a561ba9/moveit_ros/planning/rdf_loader/src/rdf_loader.cpp#L76) requires that a parameter with the robot description is set AND that a separate parameter is set with the SRDF is set. 

I find this most annoying when trying to plan from RViz, and you need to load the URDF/SRDF into the namespaces for both move_group and Rviz. 

There are three different things that I think would make things easier: 
 * `RDFLoader` allows for the parameter to be set for the URDF or reads it in from a topic.
 * `RDFLoader` publishes a `std_msgs::msg::String` with the SRDF
 * `RDFLoader` subscribes to the above String as an alternative to the parameter. 

There's two tricky parts that I foresee:
 * Defining the `RDFLoader` constructor in a sufficiently concise way that allows for parameter or topic. 
 * Making sure that a single instance `RDFLoader` doesn't subscribe to the topic it publishes.

tylerjw:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MoveIt-61675936b391800012280f6d/issues/ros-planning/moveit2/755?planning-poker) with ZenHub @abake48 @Abishalini @AndyZe @DLu @henningkayser @JafarAbdi @vatanaksoytezer

Issue #169, opened 2020-02-06
Add ability to load robot description of RobotState plugin from topic and file


JafarAbdi:
Copying Mike's comment from this [PR](https://github.com/ros-planning/moveit2/pull/168)

> I still think that the robot_description should be able to come from either a parameter, topic, or file like the RobotModelDisplay. Just using parameters doesn't make that much sense as it requires Rviz to be launched with special arguments

I think this's the rdf_loader responsibility, we should add a new constructor for rdf_loader to do so

henningkayser:
Continued in #755

PR #765, opened 2021-10-29
Latched Strings for URDF and SRDF

DLu:
### Description

Initial solution for #755 / #169

 * Adds the `rdf_loader::StringLoader` class which loads a string from the ROS environment, either via parameter or topic (could theoretically be extended for dynamic parameter later)
 * Uses the above functionality to load `robot_description` and `robot_description_semantic`
 * Adds a parameter to optionally publish a `std_msgs::msg::String` if read in from a parameter. 
 * You can also add callback hooks so that the class that uses `RDFLoader` can take appropriate actions when the description is updated. 

### Checklist

### DIFF
--- a/moveit_ros/planning/rdf_loader/include/moveit/rdf_loader/rdf_loader.h
+++ b/moveit_ros/planning/rdf_loader/include/moveit/rdf_loader/rdf_loader.h
@@ -37,6 +37,7 @@
 #pragma once
 
 #include <moveit/macros/class_forward.h>
+#include <moveit/rdf_loader/string_loader.h>
 #include <urdf/model.h>
 #include <srdfdom/model.h>
 #include <rclcpp/rclcpp.hpp>
@@ -46,17 +47,24 @@ namespace rdf_loader
 MOVEIT_CLASS_FORWARD(RDFLoader);  // Defines RDFLoaderPtr, ConstPtr, WeakPtr... etc
 
 /** @class RDFLoader
- *  @brief Default constructor
- *  @param robot_description The string name corresponding to the ROS param where the URDF is loaded*/
+ */
 class RDFLoader
 {
 public:
   /** @brief Default constructor
-   *  @param robot_description The string name corresponding to the ROS param where the URDF is loaded; the SRDF is
-   * assumed to be at the same param name + the "_semantic" suffix */
+   *
+   *  Loads the URDF from a parameter given by the string argument,
+   *  and the SRDF that has the same name + the "_semantic" suffix
+   *
+   *  If the parameter does not exist, attempt to subscribe to topics
+   *  with the same name and type std_msgs::msg::String.
+   *
+   *  @param node ROS interface for parameters / topics
+   *  @param robot_description The string name corresponding to the URDF
+   */
   RDFLoader(const std::shared_ptr<rclcpp::Node>& node, const std::string& robot_description = "robot_description");
 
-  /** \brief Initialize the robot model from a string representation of the URDF and SRDF documents */
+  /** @brief Initialize the robot model from a string representation of the URDF and SRDF documents */
   RDFLoader(const std::string& urdf_string, const std::string& srdf_string);
 
   /** @brief Get the resolved parameter name for the robot description */
@@ -75,6 +83,16 @@ public:
   const srdf::ModelSharedPtr& getSRDF() const
   {
     return srdf_;
+  }
+
+  void addDescriptionUpdateCallback(StringCallback cb)
+  {
+    external_description_update_cb_ = cb;
+  }
+
+  void addSemanticUpdateCallback(StringCallback cb)
+  {
+    external_semantic_update_cb_ = cb;
   }
 
   /** @brief determine if given path points to a xacro file */
@@ -97,7 +115,20 @@ public:
                                   const std::string& relative_path, const std::vector<std::string>& xacro_args);
 
 private:
+  bool loadURDFFromString(const std::string& content);
+  bool loadSRDFFromString(const std::string& content);
+
+  void descriptionUpdateCallback(const std::string& new_description);
+  void semanticUpdateCallback(const std::string& new_semantic);
+
+  StringCallback external_description_update_cb_;
+  StringCallback external_semantic_update_cb_;
+
   std::string robot_description_;
+
+  StringLoader description_loader_;
+  StringLoader semantic_loader_;
+
   srdf::ModelSharedPtr srdf_;
   urdf::ModelInterfaceSharedPtr urdf_;
 };
--- a/moveit_ros/planning/rdf_loader/include/moveit/rdf_loader/rdf_loader.h
+++ b/moveit_ros/planning/rdf_loader/include/moveit/rdf_loader/rdf_loader.h
@@ -61,10 +61,16 @@ public:
    *  If the parameter does not exist, attempt to subscribe to topics
    *  with the same name and type std_msgs::msg::String.
    *
+   *  (specifying default_continuous_value/default_timeout allows users
+   *   to specify values without setting ros parameters)
+   *
    *  @param node ROS interface for parameters / topics
    *  @param ros_name The string name corresponding to the URDF
+   *  @param default_continuous_value Default value for parameter with "_continuous" suffix.
+   *  @param default_timeout Default value for parameter with "_timeout" suffix.
    */
-  RDFLoader(const std::shared_ptr<rclcpp::Node>& node, const std::string& ros_name = "robot_description");
+  RDFLoader(const std::shared_ptr<rclcpp::Node>& node, const std::string& ros_name = "robot_description",
+            bool default_continuous_value = false, double default_timeout = 10.0);
 
   /** @brief Initialize the robot model from a string representation of the URDF and SRDF documents */
   RDFLoader(const std::string& urdf_string, const std::string& srdf_string);
--- a/moveit_ros/planning/rdf_loader/include/moveit/rdf_loader/rdf_loader.h
+++ b/moveit_ros/planning/rdf_loader/include/moveit/rdf_loader/rdf_loader.h
@@ -37,7 +37,7 @@
 #pragma once
 
 #include <moveit/macros/class_forward.h>
-#include <moveit/rdf_loader/string_loader.h>
+#include <moveit/rdf_loader/synchronized_string_parameter.h>
 #include <urdf/model.h>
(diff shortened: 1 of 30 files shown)


DLu:
I tested with [this branch of `moveit_resources`](https://github.com/ros-planning/moveit_resources/compare/ros2...DLu:magical_strings) and the command `ros2 launch moveit_resources_panda_moveit_config demo.launch.py`

moveit_ros/planning/rdf_loader/include/moveit/rdf_loader/rdf_loader.h
```diff
@@ -77,6 +85,16 @@ class RDFLoader
     return srdf_;
   }
 
+  void addDescriptionUpdateCallback(StringCallback cb)
```
henningkayser, line 88: Are these being used anywhere? It would be great to have an example. I assume the callbacks are only meant to be used for initialization and not or updates. Wouldn't separating srdf and urdf updates allow for some very nasty race conditions? If we allow this API, there should at least be a check if URDF and SRDF still fit together.

henningkayser: Also, can't the updates be registered with the `StringLoader` directly, so that we can get rid of the extra callback members?

moveit_ros/planning/rdf_loader/include/moveit/rdf_loader/string_loader.h
```diff
@@ -0,0 +1,83 @@
...
+ * If that fails, it subscribes to a std_msgs::String topic of the same name to get the value.
+ *
+ * If the parameter is loaded successfully, you can publish the value as a String msg if the publish_NAME param is true.
+ *
+ * You can specify how long to wait for a subscribed message with NAME_timeout (double in seconds)
+ *
+ * By default, the subscription will be killed after the first message is recieved.
+ * If the parameter NAME_continuous is true, then the parent_callback will be called on every subsequent message.
+ */
+class StringLoader
```
henningkayser, line 59: I find this class name to be a bit misleading. The class basically abstracts from the rclcpp parameter API and opts for the ability to receive and send values in case the parameter is not there. `SynchronizedStringParameter` or something similar would fit better imo.

moveit_ros/planning/rdf_loader/src/rdf_loader.cpp
```diff
@@ -115,22 +90,34 @@ RDFLoader::RDFLoader(const std::string& urdf_string, const std::string& srdf_str
...
-      RCLCPP_ERROR(LOGGER, "Unable to parse SRDF");
-      srdf_.reset();
-    }
+    loadSRDFFromString(srdf_string);
+  }
+}
+
+bool RDFLoader::loadURDFFromString(const std::string& content)
+{
+  std::unique_ptr<urdf::Model> urdf(new urdf::Model());
```
henningkayser, line 101: Can you change these to `std::make_unique`?

henningkayser (commented):
Implementation looks good to me so far. We should get a version of this feature merged to unblock MSA work, I would only like to at least make sure the API is compatible with switching to using synced parameters in the future. With this implementation, I don't see too many issues with this, though.

SRDF and URDF need to be checked for compatibility, otherwise the callbacks are risky to use. Possibly switch to using a single callback that triggers when both values are present.

Does this listen to a global namespace?

DLu:
One thing I'm pondering is how to specify parameters while maintaining maximum compatibility. The goal was to keep the `Node` based constructor the same, so lines like this didn't have to change:

    rdf_loader_.reset(new rdf_loader::RDFLoader(node_, robot_description_property_->getStdString()));

But if we want to have `robot_state_display` update when a new robot description is published, we need to set the `_continuous` parameter.... or add something to the constructor to allow bypassing/overriding the parameter check.

moveit_ros/planning/rdf_loader/include/moveit/rdf_loader/rdf_loader.h
DLu: A) I added an example using one of the RViz plugins. 
B) I unified it to one callback that only fires when the URDF and SRDF are compatible.

moveit_ros/planning/rdf_loader/src/synchronized_string_parameter.cpp
```diff
@@ -0,0 +1,154 @@
...
+  node_ = node;
+  name_ = name;
+  parent_callback_ = parent_callback;
+
+  if (getMainParameter())
+  {
+    if (shouldPublish())
+    {
+      // Transient local is similar to latching in ROS 1.
+      string_publisher_ = node_->create_publisher<std_msgs::msg::String>(name_, rclcpp::QoS(1).transient_local());
```
tylerjw, line 53: Should this be `ParameterQoS` or something else from here instead of `QoS(1)`: https://docs.ros2.org/galactic/api/rclcpp/qos_8hpp_source.html

moveit_ros/planning/rdf_loader/src/synchronized_string_parameter.cpp
```diff
@@ -0,0 +1,154 @@
...
+    node_->declare_parameter(publish_param, rclcpp::ParameterType::PARAMETER_BOOL);
+  }
+  node_->get_parameter_or(publish_param, publish_string, false);
+  return publish_string;
+}
+
+bool SynchronizedStringParameter::waitForMessage(const rclcpp::Duration timeout)
+{
+  string_subscriber_ = node_->create_subscription<std_msgs::msg::String>(
+      name_, rclcpp::QoS(1).transient_local().reliable(),
```
tylerjw, line 122: Same question here with the QoS.

tylerjw (commented):
I generally really like this.  I just have a small question about the base QoS you are using.

moveit_ros/planning/rdf_loader/src/synchronized_string_parameter.cpp
DLu: It's entirely possible that that does make more sense, but I was basing it on the [String publisher from robot_state_publisher](https://github.com/ros/robot_state_publisher/blob/37aff2034b58794b78f1682c8fab4d609f5d2e29/src/robot_state_publisher.cpp#L136)

moveit_ros/planning/rdf_loader/src/synchronized_string_parameter.cpp
DLu: If change the subscriber to use `rclcpp::ParametersQoS()`, the QoS does not match with the robot state publisher (linked in https://github.com/ros-planning/moveit2/pull/765#discussion_r759344079). 

RSP is publishing with `Durability: TRANSIENT_LOCAL` and `rclcpp::ParametersQoS()` uses `Durability: VOLATILE`

```
! moveit2/ > ros2 topic info /robot_description -v
Type: std_msgs/msg/String

Publisher count: 1

Node name: robot_state_publisher
Node namespace: /
Topic type: std_msgs/msg/String
Endpoint type: PUBLISHER
GID: 49.c3.10.01.fb.a5.9d.f1.1c.23.b6.ee.00.00.17.03.00.00.00.00.00.00.00.00
QoS profile:
  Reliability: RELIABLE
  Durability: TRANSIENT_LOCAL
  Lifespan: Infinite
  Deadline: Infinite
  Liveliness: AUTOMATIC
  Liveliness lease duration: Infinite

Subscription count: 2

Node name: move_group
Node namespace: /
Topic type: std_msgs/msg/String
Endpoint type: SUBSCRIPTION
GID: 29.5c.10.01.bf.0e.05.00.a0.49.dd.7e.00.00.15.04.00.00.00.00.00.00.00.00
QoS profile:
  Reliability: RELIABLE
  Durability: VOLATILE
  Lifespan: Infinite
  Deadline: Infinite
  Liveliness: AUTOMATIC
  Liveliness lease duration: Infinite
```


So I see some options:

  0. Add a QoS parameter to `rdf_loader` and `SynchronizedStringParameter`
  0. Leave it as-is, making `robot_description` and `robot_description_semantic` always have the same QoS (matching RSP)
  0. Try to change RSP to use `ParametersQoS`

I favor option 1 (leaving as-is) because its simpler and I don't care to learn the differences between the QoS settings. What do you think @tylerjw ?

tylerjw: I didn't realize that this QoS was already specified in robot_state_publisher.  In that case, I am in favor of leaving it as is (option 1) and matching the QoS from robot_state_publisher.

tylerjw (approved):

moveit_ros/planning/rdf_loader/src/rdf_loader.cpp
```diff
@@ -57,19 +57,23 @@ namespace rdf_loader
...
-  urdf_string_ = urdf_ssp_.loadInitialValue(node, ros_name,
-                                            std::bind(&RDFLoader::urdfUpdateCallback, this, std::placeholders::_1));
+  urdf_string_ =
+      urdf_ssp_.loadInitialValue(node, ros_name, std::bind(&RDFLoader::urdfUpdateCallback, this, std::placeholders::_1),
+                                 default_continuous_value, default_timeout);
 
   const std::string srdf_name = ros_name + "_semantic";
   srdf_string_ = srdf_ssp_.loadInitialValue(node, srdf_name,
-                                            std::bind(&RDFLoader::srdfUpdateCallback, this, std::placeholders::_1));
+                                            std::bind(&RDFLoader::srdfUpdateCallback, this, std::placeholders::_1),
```
tylerjw, line 75: Feel free to disregard this as we don't consistently follow this advice in the codebase but it would be nice to move away from bind if possible.  https://github.com/cpp-best-practices/cppbestpractices/blob/master/08-Considering_Performance.md#never-use-stdbind

```
[&](const auto& new_srdf_string) -> void { srdfUpdateCallback(new_srdf_string); }
```

DLu: I hate that syntax, so I probably will disregard. (Square brackets to me say array, not lambda function)

tylerjw: Ok, that is fair.  I think the reason for the advice is that most people find lambdas easier to read than the bind function and they tend to compile faster than bind.

I honestly don't have a problem with the bind syntax and it is used everywhere in moveit and ros already.  This was just a minor thing I thought I would point out if you would prefer it.

PR #769, opened 2021-11-01
WIP: Topic resampler for servo

tylerjw:
### Description

This is a work in progress, do not review yet.  

The whole point of this is to decouple the logic of calculating output from the resampling of the rate of the input.

Servo has 2 modes currently, one where the output is can be a different (presumably higher and maybe more consistant) rate than the input and a `low-latency` mode where it is input driven (input -> calcs -> output).

The idea of this is to separate that logic from the calculation of output.  So in the mode where the output is a fixed rate it would work like this:

input subscriber -> TopicResampler
TopicResampler -> calculate output -> publish output

And, in the case of the `low-latency` mode it would just execute like this:

input subscriber -> calculate output -> publish output

This would move the configuration of the mode into the constructor and simplify the logic in ServoCalcs as it no longer has to be concerned with resampling the output.

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/CMakeLists.txt
+++ b/moveit_ros/moveit_servo/CMakeLists.txt
@@ -245,10 +245,12 @@ if(BUILD_TESTING)
   add_ros_test(test/launch/test_basic_servo.test.py TIMEOUT 120 ARGS "test_binary_dir:=${CMAKE_CURRENT_BINARY_DIR}")
 
   # Enforce limits unit tests
-  ament_add_gtest(enforce_limits_tests
-    test/enforce_limits_tests.cpp
-  )
+  ament_add_gtest(enforce_limits_tests test/enforce_limits_tests.cpp)
   target_link_libraries(enforce_limits_tests ${SERVO_LIB_NAME})
+
+  # Topic resampler tests
+  ament_add_gtest(topic_resampler_tests test/topic_resampler_tests.cpp)
+  target_link_libraries(topic_resampler_tests ${SERVO_LIB_NAME})
 
 endif()
 
--- a/moveit_ros/moveit_servo/include/moveit_servo/topic_resampler.hpp
+++ b/moveit_ros/moveit_servo/include/moveit_servo/topic_resampler.hpp
@@ -1,0 +1,113 @@
+/*******************************************************************************
+ * BSD 3-Clause License
+ *
+ * Copyright (c) 2021, PickNik Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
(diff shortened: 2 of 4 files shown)


AndyZe:
This sounds very similar to how Servo originally worked. It had a "master" thread to dispatch messages. See slide 10 [here](https://roscon.ros.org/2019/talks/roscon2019_jog_arm_final.pdf).

Anyway, you yourself removed the third thread in the MoveIt1 refactoring, in order to "simplify". I think the way it is now is fine as long as the ServoCalcs loop spins very, very fast -- which it always does, from what I've seen.

tylerjw:
Please hold judgement of this until I can try to use it to replace logic in servo calcs.  I think it will be an improvement in readability and reduce CPU load because the control flow will be configured during initialization instead of with if statements on every loop cycle.

AndyZe:
Alright, I will wait to give it the thumbs down   ;)

Seriously though, it seems like a lot of effort is going into thrashing back and forth.

If we're talking about big refactorings, I would like to see the logic turned into a state machine to make it more maintainable.

tylerjw:
> logic turned into a state machine

This is essentially the first step to doing that.  My long-term plan is to try to decouple much of the logic inside servo_calcs from each-other so it makes it much easier to swap out various parts of the logic (like the smoothing step) like you would with a controller.

If the control flow can be configured during initialization then it will also enable us to enable dynamic reconfiguring of servo while it is running for tunning.

I am starting at the inputs and working towards the outputs.  These are the various sub-steps inside calcs that I think should be separate parts:

0. input source
1. TopicResampling (controls the rate of output by resampling input)
2. Input scaling (real units or unitless)
3. stale input controller (watches inputs and handles stale input condition)
4. current state source (one object to manage to get the current state for both collision checking and serving)
5. delta theta conversion (convert each of the input types to the delta_theta eigen vector)
6. enforce velocity limits (this is already separated / tested)
7. collision scaling
8. singularity scaling
9. enforce position limits (this is already separated / tested)
10. compose trajectory
11. smoothing controller
12. output message type conversion (if using std_msgs/Float64MultiArray)
13. publisher

If each of these were separate pieces of code executed by a state machine that is set up in a configuration step you could much more easily swap out one implementation for another or change parameters when a user does a dynamic reconfigure.

In short, I would really like to support dynamic reconfiguring for tuning and swapping out different types of control while executing.  I think this sort of refactoring will also make it much easier to change the various steps because they'll all have clearly defined inputs and outputs and that'll make them much easier to test.

Because I only have small amounts of time to work on this here and there I'm trying to do it in small steps.  The reason I think this change will make it nicer for you is it will be easier for you to change the behavior of individual steps or of the whole system because the logic of how each step is done will be separated from what step is done next.

tylerjw:
Closing in favor of https://github.com/ros-planning/moveit2/pull/825

Issue #770, opened 2021-11-01
Error with Servo launching


AndyZe:
Commit 7e3fdcffb seems to have caused an issue with Servo launching. The robot joint state is not initialized and the robot is upright rather than having a 90* bend at the elbow. I tested with this command on main branch, rosdistro Rolling:

`ros2 launch moveit_servo servo_example.launch.py`

No obvious errors in the terminal.

![wrong_robot_position](https://user-images.githubusercontent.com/11284393/139707844-2aaeecf5-ba91-4d4a-a2a0-19f90e4808b4.png)

AndyZe:
PR #771 partially fixed this but wait, there's more:

`Failed to call service get_planning_scene, have you launched move_group or called psm.providePlanningSceneService()?`

The joint angles look good but the robot still does not move.

tylerjw:
Locally testing I got it working by doing this:

Start servo with the above launch command:

```
ros2 launch moveit_servo servo_example.launch.py
```

Call the start service:

```
ros2 service call /servo_node/start_servo std_srvs/srv/Trigger
```

Change the topics in servo_keyboard_input.cpp from moveit2_tutorials from `servo_server` to `servo_node` namespace as the namespace of the topics changed.  Then launch it with:

```
ros2 run moveit2_tutorials servo_keyboard_input
```

I'll submit a PR to fix that in the tutorial.  Is there some other way you are running it where it is not moving?  The error about providePlanningService seems unrelated to me but I will also submit a PR to fix that here too (looking back that's been there for a while).

tylerjw:
Here is the update for the tutorials: https://github.com/ros-planning/moveit2_tutorials/pull/176

tylerjw:
Note that you still need to call the start service with something.  I hope to fix the racing condition that prevented me from removing the start service altogether.

AndyZe:
Yep, I forgot that service call is necessary in ROS2. [This PR](https://github.com/ros-planning/moveit2_tutorials/pull/109) updated the documentation.

Yay, let's move on.

PR #773, opened 2021-11-02
Call psm.providePlanningSceneService in servo init

tylerjw:
### Description

Fixes this error that you see on startup of servo:

```
Failed to call service get_planning_scene, have you launched move_group or called psm.providePlanningSceneService()?
```

As pointed out here: https://github.com/ros-planning/moveit2/issues/770#issuecomment-957030659

### Checklist


vatanaksoytezer (approved):
Tested locally, error is gone and looking all good. Thanks @tylerjw

JafarAbdi (commented):
This should be fixed with https://github.com/ros-planning/moveit2/pull/673, I think we should merge that one since it fixes other issues, too

tylerjw:
> This should be fixed with #673, I think we should merge that one since it fixes other issues, too

I'm good with that, I'll close this in favor of that one.  You need to rebase as I changed the name of one of the files you are editing.

PR #777, opened 2021-11-02
Fix Ruckig smoothing infinite loop

AndyZe:
For identical waypoints, Ruckig smoothing could get stuck in an infinite loop. The reason was, the `continue` went to the next iteration of `while`, instead of continuing to the next waypoint.

Also, I didn't have to copy the waypoint here since we just checked it was identical. We can leave it as it was.

`*next_waypoint = trajectory.getWayPoint(waypoint_idx);`

### DIFF
--- a/moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
+++ b/moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
@@ -129,6 +129,15 @@ bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajecto
 
       getNextRuckigInput(ruckig_output, next_waypoint, num_dof, idx, ruckig_input);
 
+      // Skip repeated waypoints with no change in position. Ruckig does not handle this well and there's really no
+      // need to smooth it. Simply leave it equal to the previous (identical) waypoint.
+      bool identical_waypoint =
+          checkForIdenticalWaypoints(*trajectory.getWayPointPtr(waypoint_idx), *next_waypoint, trajectory.getGroup());
+      if (identical_waypoint)
+      {
+        continue;
+      }
+
       // Run Ruckig
       ruckig_result = ruckig_ptr->update(ruckig_input, ruckig_output);
 
@@ -139,14 +148,6 @@ bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajecto
       double velocity_magnitude = getTargetVelocityMagnitude(ruckig_input, num_dof);
       while (backward_motion_detected && (velocity_magnitude > MINIMUM_VELOCITY_SEARCH_MAGNITUDE))
       {
-        // Skip repeated waypoints with no change in position. Ruckig does not handle this well and there's really no
-        // need to smooth it Simply set it equal to the previous (identical) waypoint.
-        if (checkForIdenticalWaypoints(*trajectory.getWayPointPtr(waypoint_idx), *next_waypoint, trajectory.getGroup()))
-        {
-          *next_waypoint = trajectory.getWayPoint(waypoint_idx);
-          continue;
-        }
-
         // decrease target velocity
         for (size_t joint = 0; joint < num_dof; ++joint)
         {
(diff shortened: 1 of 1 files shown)


moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -129,6 +129,15 @@ bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajecto
 
       getNextRuckigInput(ruckig_output, next_waypoint, num_dof, idx, ruckig_input);
 
+      // Skip repeated waypoints with no change in position. Ruckig does not handle this well and there's really no
+      // need to smooth it. Simply leave it equal to the previous (identical) waypoint.
+      bool identical_waypoint =
+          checkForIdenticalWaypoints(*trajectory.getWayPointPtr(waypoint_idx), *next_waypoint, trajectory.getGroup());
```
henningkayser, line 135: This really only compares the position. Further down in the loop Ruckig overwrites the position and dynamics in the same trajectory just to produce a new output that may have different positions and dynamics for the next waypoint. Wouldn't skipped waypoints lead to steps in the trajectory that are not smoothed at all? I think it would be safer to either make sure the robot is really halting (dynamics == 0) or remove duplicate waypoints if the time stamp is also the same.

henningkayser: Nit: I would probably skip the bool variable

AndyZe: I changed this to remove duplicated waypoints just once at the very beginning. That's more efficient. Good idea.

I think the waypoints should be removed even if (dynamics != 0) because they cause circular motions otherwise.

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -78,6 +78,33 @@ bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajecto
...
+        *trajectory.getWayPointPtr(waypoint_idx), *trajectory.getWayPointPtr(waypoint_idx + 1), trajectory.getGroup());
+    if (identical_waypoint)
+    {
+      continue;
+    }
+    else
+    {
+      requested_trajectory.addSuffixWayPoint(trajectory.getWayPoint(waypoint_idx),
+                                             trajectory.getWayPointDurationFromPrevious(waypoint_idx));
+    }
```
jliukkonen, line 97: Couple ideas. The `if/else` can be dropped and replaced with 
```
if(!identical_waypoint) {
 requested_trajectory.addSuffixWayPoint(trajectory.getWayPoint(waypoint_idx),
     trajectory.getWayPointDurationFromPrevious(waypoint_idx));
}
```
Logically it is the same of course. Also, the current waypoint, or the pointer to it, can be cached. Net benefit is probably negligible, but it would halve the number of calls to `getWayPointPtr`.

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -78,6 +78,33 @@ bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajecto
...
+  }
+  trajectory = requested_trajectory;
+
+  num_waypoints = trajectory.getWayPointCount();
+  if (num_waypoints < 2)
+  {
+    RCLCPP_ERROR(LOGGER, "Trajectory does not have enough points to smooth with Ruckig");
+    return false;
+  }
+
```
jliukkonen, line 107: Not really related to this PR, but something to keep in mind... The `applySmoothing` function is almost 200 lines long and contains complex logic. That's too long and should be, at some point, split into pieces. I'm thinking about just having helper functions that capture the phases of the smoothing correctly. 

The trajectory processing functions are also all `static` class member functions meaning they really don't have to be in a class but can instead just be simple functions in some suitable namespace.

jliukkonen (commented):
Couple code related comments. As for the idea itself, yup, makes sense.

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
AndyZe: Thanks!

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
AndyZe: Thanks @jliukkonen. I'm going to close this PR because some refactoring was done and it no longer applies. But I will try to do what you suggest in that PR.

AndyZe:
Deprecated by https://github.com/ros-planning/moveit2/pull/1076

PR #785, opened 2021-11-05
Update controller_manager_plugin to fix MoveIt-managed controller switching

schornakj:
### Description

This is a partial replacement for #731 that only includes the changes required to allow MoveIt-managed controller switching to work again while excluding changes to the behavior of the controller manager plugin.

- Discover available controllers when initializing the controller manager plugin.
- Account for changes to `controller_manager_msgs` in ros2_control v1.1.0. The most significant change is that resources that are required but not claimed by the controller (for example, if the controller is not active) are now listed in a separate message field than resources that are actively claimed by the controller.
- Use `scoped_lock` in place of `unique_lock`.
- Add more detailed comments and logging.

### Checklist

### DIFF
--- a/moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
+++ b/moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
@@ -100,7 +100,12 @@ class MoveItControllerManager : public moveit_controller_manager::MoveItControll
   HandleMap handles_;
 
   rclcpp::Time controllers_stamp_{ 0, 0, RCL_ROS_TIME };
+
+  /**
+   * @brief Protects access to managed_controllers_, active_controllers_, allocators_, handles_, and controllers_stamp.
+   */
   std::mutex controllers_mutex_;
+
   rclcpp::Node::SharedPtr node_;
   rclcpp::Client<controller_manager_msgs::srv::ListControllers>::SharedPtr list_controllers_service_;
   rclcpp::Client<controller_manager_msgs::srv::SwitchController>::SharedPtr switch_controller_service_;
--- a/moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
+++ b/moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
@@ -87,8 +87,12 @@ class MoveItControllerManager : public moveit_controller_manager::MoveItControll
   std::string ns_;
   pluginlib::ClassLoader<ControllerHandleAllocator> loader_;
   typedef std::map<std::string, controller_manager_msgs::msg::ControllerState> ControllersMap;
+
+  /** @brief Controllers that can be activated and deactivated by this plugin. */
   ControllersMap managed_controllers_;
+  /** @brief Controllers that are currently active. */
   ControllersMap active_controllers_;
+
   typedef std::map<std::string, ControllerHandleAllocatorPtr> AllocatorsMap;
   AllocatorsMap allocators_;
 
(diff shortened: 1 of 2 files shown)


moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -236,6 +245,9 @@ class MoveItControllerManager : public moveit_controller_manager::MoveItControll
         getAbsName("controller_manager/list_controllers"));
     switch_controller_service_ = node_->create_client<controller_manager_msgs::srv::SwitchController>(
         getAbsName("controller_manager/switch_controller"));
+
+    std::scoped_lock<std::mutex> lock(controllers_mutex_);
```
AndyZe, line 249: At L103, can you add a comment on what `controllers_mutex_` locks? I guess it is any of the member variables in this class.

AndyZe (commented):
Do you think all of this mutex locking could cause deadlocks? For example, if `getActiveControllers()` is called at the same time as `switchControllers()`?

We discussed on Wednesday about not using a MoveIt ControllerManager to find and actiave a suitable controller, but rather relying on the user's application to activate the correct controller. That seems like a safer and simpler way to go. ([Discussion here](https://github.com/ros-planning/moveit2/issues/726#issuecomment-951228453))

I do like your idea about being able to specify which controller to use from a [RobotTrajectory message](https://github.com/ros-planning/moveit_msgs/blob/ros2/msg/RobotTrajectory.msg). MoveIt could simply check if that controller is active.

AndyZe (changes requested):
LGTM

AndyZe (approved):

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
schornakj: I added a comment -- you're right that it's just protecting access to member variables.

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -360,15 +377,17 @@ class MoveItControllerManager : public moveit_controller_manager::MoveItControll
...
+    // For each controller to activate, find conflicts between the interfaces required by that controller and the
+    // interfaces claimed by currently-active controllers.
     for (const std::string& it : activate)
     {
       ControllersMap::iterator c = managed_controllers_.find(it);
       if (c != managed_controllers_.end())
       {  // controller belongs to this manager
         request->start_controllers.push_back(c->second.name);
-        for (const auto& claimed_resource : c->second.claimed_interfaces)
+        for (const auto& required_resource : c->second.required_command_interfaces)
```
AndyZe, line 388: What if Controller A requires the position command interface for Joint 1 and Controller B requires the velocity command interface for Joint1. Will that be seen as a conflict here? I think it should be.

(Refer to the [example here](https://github.com/ros-controls/ros2_control_demos/blob/37b3338b22ce99200791b191df647256d51f233e/ros2_control_demo_description/rrbot_description/ros2_control/rrbot_system_multi_interface.ros2_control.xacro#L30))

I think you will want to mark it as a conflict if both controllers have any command_interface for the same joint

AndyZe: Ok, that's fine with me. Thanks for the comment.

AndyZe (approved):

JafarAbdi (approved):

PR #789, opened 2021-11-09
Fix installation of moveit_ros_control_interface header files

schornakj:
### Description

The header files for the `moveit_ros_control_interface::ControllerHandleAllocator` class are not installed correctly, which prevents the creation of new controller handle plugins in packages other than the main `moveit_ros_control_interface` package (for example, if someone has created a custom controller and want to execute trajectories with it using the TrajectoryExecutionManager). This PR fixes that.

### Checklist


moveit_plugins/moveit_ros_control_interface/CMakeLists.txt
```diff
@@ -76,10 +76,11 @@ EXPORT export_${PROJECT_NAME}
 ARCHIVE DESTINATION lib
 LIBRARY DESTINATION lib
 RUNTIME DESTINATION bin
+INCLUDES DESTINATION include
```
JafarAbdi, line 79: Do we actually need this .?

JafarAbdi (approved):
LGTM

moveit_plugins/moveit_ros_control_interface/CMakeLists.txt
schornakj: I wasn't able to get the expected result without that, but I haven't dug into why exactly that is yet.

schornakj: I checked this again, and it looks like it's actually the other way around. The `install` function needs to have the `INCLUDES DESTINATION include` line for the headers to be found by other packages.

moveit_plugins/moveit_ros_control_interface/CMakeLists.txt
```diff
@@ -76,11 +76,7 @@ EXPORT export_${PROJECT_NAME}
 ARCHIVE DESTINATION lib
 LIBRARY DESTINATION lib
 RUNTIME DESTINATION bin
-)
-
-## Mark cpp header files for installation
-install(DIRECTORY include/${PROJECT_NAME}/
```
henningkayser, line 82: Usually, we would use `install(DIRECTORY include/ DESTINATION include)`. Otherwise, the headers will not be inside the projects path when you call `#include <project_name/header.h>`

moveit_plugins/moveit_ros_control_interface/CMakeLists.txt
henningkayser: afaik, this doesn't really install the headers but it only makes them available for the target. This should not be necessary, if the headers are actually installed in the right path, adding this is line is still correct.

moveit_plugins/moveit_ros_control_interface/CMakeLists.txt
schornakj: OK, I'll revert the change that deleted this line and push up again. I didn't do a clean build for this package when I was testing my change, which I realized means that changes to what was installed weren't reflected by the state of the built workspace.

henningkayser (approved):

tylerjw (approved):

vatanaksoytezer (approved):

PR #596, opened 2021-08-09
Clang-tidy fixes

henningkayser:
~~Apply https://github.com/ros-planning/moveit/pull/2792~~

I wanted to update this or get it closed. Of course, the changes to get clang enabled are already applied in the meantime. The fixes from the linked PR doesn't seem to be necessary, at least I didn't get any warnings locally using clang-10. However, I found other minor fixes that I added instead with this PR.

PR #792, opened 2021-11-10
Make controller management logic more tolerant of missing or late ros2_control nodes

schornakj:
### Description

Two related fixes to avoid some troublesome situations that happen if ros2_control's ROS interface isn't available when MoveIt is initialized (such as if the ros2_control nodes take longer to finish launching than the MoveIt nodes):

- Reload information about controllers whenever `TrajectoryExecutionManager::ensureActiveControllers` is called. Previously, reloading controller info only happened when initializing the TrajectoryExecutionManager and when executing a trajectory. This sometimes caused situations where using `ensureActiveControllers` outside the scope of executing a trajectory failed because the info about the controllers was out of date.

- Allow controller handles derived from `ActionBasedControllerHandle` to completely initialize even if the action server they expect to interact with is not initially available. I removed the blocking connection timeout check in the class constructor and changed `isConnected` to check for the existence/readiness of the expected action server.

edit: This also changes how parameter names are created within the `MoveItSimpleControllerManager`, originally to comply with clang-tidy recommendations. I added a new parameter name creation function that saves us from manually concatenating strings with specific separators.

### Checklist

### DIFF
--- a/moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
+++ b/moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
@@ -85,17 +85,6 @@ public:
     last_exec_ = moveit_controller_manager::ExecutionStatus::SUCCEEDED;
   }
 
-  bool isConnected() const
-  {
-    if (!controller_action_client_->action_server_is_ready())
-    {
-      RCLCPP_ERROR_STREAM(LOGGER, "Action client not connected: " << getActionName());
-      return false;
-    }
-
-    return true;
-  }
-
   bool cancelExecution() override
   {
     if (!controller_action_client_)
@@ -160,6 +149,11 @@ public:
   }
 
 protected:
+  bool isConnected() const
+  {
+    return controller_action_client_->action_server_is_ready();
+  }
+
   const rclcpp::Node::SharedPtr node_;
   std::string getActionName() const
   {
--- a/moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
+++ b/moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
@@ -82,39 +82,18 @@ public:
     : ActionBasedControllerHandleBase(name, logger_name), node_(node), done_(true), namespace_(ns)
   {
     controller_action_client_ = rclcpp_action::create_client<T>(node_, getActionName());
-
-    unsigned int attempts = 0;
-    double timeout;
-    node_->get_parameter_or("trajectory_execution.controller_connection_timeout", timeout, 15.0);
-
-    if (timeout == 0.0)
-    {
-      while (rclcpp::ok() && !controller_action_client_->wait_for_action_server(std::chrono::seconds(5)))
-      {
-        RCLCPP_WARN_STREAM(LOGGER, "Waiting for " << getActionName() << " to come up");
-      }
-    }
-    else
-    {
-      while (rclcpp::ok() &&
-             !controller_action_client_->wait_for_action_server(std::chrono::duration<double>(timeout / 3)) &&
-             ++attempts < 3)
-      {
-        RCLCPP_WARN_STREAM(LOGGER, "Waiting for " << getActionName() << " to come up");
-      }
-    }
-    if (!controller_action_client_->action_server_is_ready())
-    {
-      RCLCPP_ERROR_STREAM(LOGGER, "Action client not connected: " << getActionName());
-      controller_action_client_.reset();
-    }
-
     last_exec_ = moveit_controller_manager::ExecutionStatus::SUCCEEDED;
   }
 
   bool isConnected() const
   {
-    return static_cast<bool>(controller_action_client_);
+    if (!controller_action_client_->action_server_is_ready())
+    {
+      RCLCPP_ERROR_STREAM(LOGGER, "Action client not connected: " << getActionName());
+      return false;
+    }
+
+    return true;
   }
 
   bool cancelExecution() override
--- a/moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
+++ b/moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
@@ -79,9 +79,11 @@ class ActionBasedControllerHandle : public ActionBasedControllerHandleBase
 public:
   ActionBasedControllerHandle(const rclcpp::Node::SharedPtr& node, const std::string& name, const std::string& ns,
                               const std::string& logger_name)
-    : ActionBasedControllerHandleBase(name, logger_name), node_(node), done_(true), namespace_(ns)
+    : ActionBasedControllerHandleBase(name, logger_name), done_(true), namespace_(ns)
   {
-    controller_action_client_ = rclcpp_action::create_client<T>(node_, getActionName());
+    // Creating the action client does not ensure that the action server is actually running. Executing trajectories
+    // through the controller handle will fail if the server is not running when an action goal message is sent.
+    controller_action_client_ = rclcpp_action::create_client<T>(node, getActionName());
     last_exec_ = moveit_controller_manager::ExecutionStatus::SUCCEEDED;
   }
 
@@ -203,11 +205,6 @@ protected:
       last_exec_ = moveit_controller_manager::ExecutionStatus::FAILED;
     done_ = true;
   }
-
-  /**
-   * @brief Node used to create the action client.
-   */
-  const rclcpp::Node::SharedPtr node_;
 
   /**
    * @brief Status after last trajectory execution.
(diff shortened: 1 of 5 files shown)


moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
```diff
@@ -82,39 +82,18 @@ class ActionBasedControllerHandle : public ActionBasedControllerHandleBase
...
-    else
-    {
-      while (rclcpp::ok() &&
-             !controller_action_client_->wait_for_action_server(std::chrono::duration<double>(timeout / 3)) &&
-             ++attempts < 3)
-      {
-        RCLCPP_WARN_STREAM(LOGGER, "Waiting for " << getActionName() << " to come up");
-      }
-    }
+  bool isConnected() const
```
tylerjw, line 88: I'm sorry if this is obvious, but what is the user of this?  As we are no longer waiting for the connection in the constructor does the user need to call this before calling other methods on this?  Should each method that interacts with this check this before doing whatever it is going to do?

Would you mind adding some documentation (even just a comment) about the pre-conditions of each method here?

schornakj: After looking at how this function is used some more, I think this is a good thing to change here.

Right now `isConnected` is only used by the MoveItSimpleControllerManager, and it isn't part of the interface of the [MoveItControllerHandle base class](https://github.com/ros-planning/moveit2/blob/54c6e08ead8d5888198cb819cd933cb6229a3077/moveit_core/controller_manager/include/moveit/controller_manager/controller_manager.h#L105).

The expectation established in the base class is that `sendTrajectory` returns `false` if the controller cannot accept the trajectory, so it should not be a requirement to do a separate check before that to determine if the controller handle is ready to execute trajectories.

In the current ActionBasedControllerHandle implementation the class constructor leaves class members not fully initialized if the action server wasn't found within some timeout, which leaves it in sort of a limbo state. There isn't a way to have the handle retry connecting later if needed (other than knowing that you might have to create a new instance of the handle later). This doesn't seem like a good way to handle this.

I think the right way forward is to delete `isConnected` and modify MoveIt SimpleControllerManager to have it always create the controller handles regardless of the availability of the associated action servers.

schornakj:
It's out-of-scope for this PR, but I think the separate `ActionBasedControllerHandleBase` class can be removed entirely and `ActionBasedControllerHandle` can be changed to inherit directly from `moveit_controller_manager::MoveItControllerHandle`. As far as I can tell the `addJoint` and `getJoints` functions aren't used at all within MoveIt2.

schornakj:
CI is failing because clang-tidy objects to string concatenation in `moveit_simple_controller_manager.cpp`. How should I handle that? I can push up a commit to change that, but since I didn't originally edit the files that cause the failure it'll bloat the PR a bit.

tylerjw:
> CI is failing because clang-tidy objects to string concatenation in `moveit_simple_controller_manager.cpp`. How should I handle that? I can push up a commit to change that, but since I didn't originally edit the files that cause the failure it'll bloat the PR a bit.

The reason you are dealing with that is because of how we run clang-tidy.  clang-tidy takes a really really long time to execute so it is only run on packages that have changes in a given PR.  One side-effect of that is that when you make a PR against something that hasn't been changed in a long long time you run into this (as we changed clang-tidy versions and rules a little while ago).  If it is simple enough, please just fix whatever it is complaining about as part of your PR.  If you would prefer, you could also create a second PR for just that change that we can review/merge quickly, and then you could rebase on that one.

schornakj:
> The reason you are dealing with that is because of how we run clang-tidy. clang-tidy takes a really really long time to execute so it is only run on packages that have changes in a given PR. One side-effect of that is that when you make a PR against something that hasn't been changed in a long long time you run into this (as we changed clang-tidy versions and rules a little while ago). If it is simple enough, please just fix whatever it is complaining about as part of your PR. If you would prefer, you could also create a second PR for just that change that we can review/merge quickly, and then you could rebase on that one.

Understood. I added a commit to implement clang-tidy's suggested fix for string concatenation.

moveit_plugins/moveit_simple_controller_manager/src/moveit_simple_controller_manager.cpp
```diff
@@ -44,6 +44,23 @@
...
+ * @param param_name Specific name of the parameter.
+ * @return std::string concatenating the parameters separated by the `.` character. For example:
+ * base_namespace.controller_name.param_name
+ */
+std::string makeControllerParameterName(const std::string& base_namespace, const std::string& controller_name,
+                                        const std::string& param_name)
+{
+  return std::string(base_namespace).append(".").append(controller_name).append(".").append(param_name);
+}
+}  // namespace
```
tylerjw, line 62: Thank you for adding this and fixing the issue with strings and clang-tidy.  Long term I hope to make it easier to have subsystems of the node use parameters and that be really straightforward.  There are a couple of PRs to rclcpp you might find interesting relating to this:

https://github.com/ros2/rclcpp/pull/1824
https://github.com/ros2/rclcpp/pull/1796

tylerjw (approved):

moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
```diff
@@ -82,41 +82,13 @@ class ActionBasedControllerHandle : public ActionBasedControllerHandleBase
...
-      RCLCPP_ERROR_STREAM(LOGGER, "Action client not connected: " << getActionName());
-      controller_action_client_.reset();
-    }
-
     last_exec_ = moveit_controller_manager::ExecutionStatus::SUCCEEDED;
   }
 
-  bool isConnected() const
-  {
-    return static_cast<bool>(controller_action_client_);
```
AndyZe, line 117: The change you made here is a nice upgrade.

moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
```diff
@@ -82,41 +82,13 @@ class ActionBasedControllerHandle : public ActionBasedControllerHandleBase
     : ActionBasedControllerHandleBase(name, logger_name), node_(node), done_(true), namespace_(ns)
   {
     controller_action_client_ = rclcpp_action::create_client<T>(node_, getActionName());
```
AndyZe, line 84:
```suggestion
    // This does not ensure the action server is actually running.
    controller_action_client_ = rclcpp_action::create_client<T>(node_, getActionName());
```

moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
```diff
@@ -206,21 +204,40 @@ class ActionBasedControllerHandle : public ActionBasedControllerHandleBase
     done_ = true;
   }
 
-  /* execution status */
+  /**
+   * @brief Node used to create the action client.
+   */
+  const rclcpp::Node::SharedPtr node_;
```
AndyZe, line 210: It looks like `node_` is hardly used at all. Delete it

moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
AndyZe: ^I'm good with what you wrote, Joe, as long as `sendTrajectory` prints a nice error if the action server isn't running

AndyZe: ^What about you, @JafarAbdi ?

AndyZe (commented):
This looks good to me. I'll approve if you delete `isConnected()`, as you commented

schornakj:
> This looks good to me. I'll approve if you delete isConnected(), as you commented

I had changed `isConnected()` so it checks for the availability of the action server, which is used when determining if the controller handle is able to attempt executing a trajectory. I made it protected too so the controller handle implementations have an interface more like the base class. I think it makes sense to keep it around.

moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
schornakj: > as long as sendTrajectory prints a nice error if the action server isn't running

Since each controller handle has to provide its own implementation of `sendTrajectory`, it's the handle's responsibility to log an error if trajectory execution failed due to a missing action server. The existing handles in the `moveit_simple_controller_manager` package already do this.

moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
schornakj: Deleted!

schornakj:
There was a conflict introduced by #596, since that also touched the string concatenation strategy. While resolving the conflict I revised the parameter name creator function to be variadic, which makes it a lot more flexible and general-purpose.

AndyZe (approved):

PR #793, opened 2021-11-11
Use a steady clock for timeout for IK

galou:

### Description

The `searchPositionIK` can hang forever when using a simulated time because the time is stuck while waiting for this function to return. For this simple reason the function never times out. This was reported by @jmarsik, I'll let him write his full report here. Most of the work, i.e. isolating the cause, was done by him.

### Checklist

I don't know how to write a test for this. It doesn't appear to be worth it.

### DIFF
--- a/moveit_kinematics/kdl_kinematics_plugin/src/kdl_kinematics_plugin.cpp
+++ b/moveit_kinematics/kdl_kinematics_plugin/src/kdl_kinematics_plugin.cpp
@@ -49,6 +49,8 @@ namespace kdl_kinematics_plugin
 {
 static rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_kdl_kinematics_plugin.kdl_kinematics_plugin");
 
+rclcpp::Clock KDLKinematicsPlugin::steady_clock_{ RCL_STEADY_TIME };
+
 KDLKinematicsPlugin::KDLKinematicsPlugin() : initialized_(false)
 {
 }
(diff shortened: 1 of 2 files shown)


moveit_kinematics/kdl_kinematics_plugin/src/kdl_kinematics_plugin.cpp
```diff
@@ -268,7 +268,9 @@ bool KDLKinematicsPlugin::initialize(const rclcpp::Node::SharedPtr& node, const
 
 bool KDLKinematicsPlugin::timedOut(const rclcpp::Time& start_time, double duration) const
 {
-  return ((node_->now() - start_time).seconds() >= duration);
+  static rclcpp::Clock clock{ RCL_STEADY_TIME };
```
henningkayser, line 271: Can you make `clock` a class member `steady_clock_`?

henningkayser (commented):
@galou thanks for your PR! I agree, it makes perfect sense to use steady time here. I bet there are other cases in the codebase where we should fix this as well. Could you target `main` with your PR? We can backport it to foxy.

moveit_kinematics/kdl_kinematics_plugin/src/kdl_kinematics_plugin.cpp
tylerjw: `main` only builds for galactic or rolling, I presume that is why it does not build for you.  If it is something else please let us know and we are happy to help.  Again, thank you for this change.

jmarsik:
I have posted my report in the new PR ... see here https://github.com/ros-planning/moveit2/pull/795#issuecomment-966661217

PR #794, opened 2021-11-11
Add backward_ros to moveit_common

tylerjw:
### Description

This adds backward_ros to all moveit packages that use the moveit_common cmake function.  This is a really useful tool and building it in by default seems like a really nice feature.

https://github.com/pal-robotics/backward_ros/tree/foxy-devel

### DIFF
--- a/moveit_common/cmake/moveit_package.cmake
+++ b/moveit_common/cmake/moveit_package.cmake
@@ -32,6 +32,9 @@ macro(moveit_package)
   find_package(ament_cmake REQUIRED)
   ament_package_xml()
 
+  # Enable backward_ros on every moveit package
+  find_package(backward_ros REQUIRED)
+
   if(NOT "${CMAKE_CXX_STANDARD}")
     set(CMAKE_CXX_STANDARD 17)
   endif()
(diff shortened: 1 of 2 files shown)


tylerjw:
> This is great! Are we sure that this doesn't have any performance impacts? From what I've read it wouldn't change the binaries at all as it fully relies on the debug info.

This will result in slightly larger binaries (one more library linked in) but other than that I don't believe so.

vatanaksoytezer (approved):
I'm also in favor of this, this will help a lot!

v4hn:
O.O . This is literally a dream come true :angel:

v4hn (approved):

moveit_common/cmake/moveit_package.cmake
```diff
@@ -32,6 +32,9 @@ macro(moveit_package)
   find_package(ament_cmake REQUIRED)
   ament_package_xml()
 
+  # Enable backward_ros on every moveit package
+  find_package(backward_ros REQUIRED)
```
v4hn, line 36: It doesn't make much difference for a well-setup structure, but QUIET would be enough here I guess?

```suggestion
  find_package(backward_ros QUIET)
```

henningkayser: I agree

henningkayser (approved):

PR #795, opened 2021-11-11
Use a steady clock for timeout for IK

galou:

### Description

The `searchPositionIK` can hang forever when using a simulated time because the time is stuck while waiting for this function to return. For this simple reason the function never times out. This was reported by @jmarsik, I'll let him write his full report here. Most of the work, i.e. isolating the cause, was done by him.

### Checklist

I don't have Galactic or Rolling, I'll have to trust CI.
I don't know how to write a test for this. It doesn't appear to be worth it.

N.B.: follow-up of #793 .


tylerjw (approved):
Looks good to me, thank you!

jmarsik:
Hello guys,

I will just copy most of my report (the original resides in a private repo) here.
Right now we are using ROS2 Foxy.

> When moving the goal interactive marker in RViz2 the whole program often hangs and has to be killed. For a long time I did not see all the relevant conditions that has to be fulfilled for it to happen (or maybe I just did not have enough time to pay attention to them).
> 
> Today I did a full debug log of RViz2 node and noticed that it hangs on calculation of the IK solution which never ends. See log:
> 
> ```
> [DEBUG] [1636563254.040887548] [rviz]: Marker 'EE:goal_robot_tool0' is now at 0,804418, 0,000000, 2,600000
> [DEBUG] [1636563254.040945537] [rviz]: Calling default feedback callback for marker 'EE:goal_robot_tool0'
> [DEBUG] [1636563254.040973610] [moveit_ros_robot_interaction.robot_interaction]: Processing feedback from map for marker [EE:goal_robot_tool0]
> [DEBUG] [1636563254.041033483] [moveit_kinematics_base.kinematics_base]: searchPositionIK: Position request pose is 0.804418 0 2.6 0.707107 -4.32978e-17 0.707107 4.32978e-17
> [DEBUG] [1636563254.041045075] [moveit_kinematics_base.kinematics_base]: Input: [           0   -0.931892    -0.62646           0   -0.305432           0]
> [DEBUG] [1636563254.041056756] [moveit_kinematics_base.kinematics_base]: [  0] delta_twist: [   -0.248619,           0, 1.37123e-08,           0,           0,           0]
> [DEBUG] [1636563254.041070011] [moveit_kinematics_base.kinematics_base]: [  0] pos err: 0,248619  rot err: 0,000000  delta_q: 0,278722
> [DEBUG] [1636563254.041078517] [moveit_kinematics_base.kinematics_base]:       delta_q: [           0  -0.0681082  -0.0742861           0   -0.136328           0]
> [DEBUG] [1636563254.041085100] [moveit_kinematics_base.kinematics_base]:       q: [           0          -1   -0.700746           0    -0.44176           0]
> [DEBUG] [1636563254.041109786] [interactive_marker_display_93854378973952]: Pushing update with sequence number 26
> [DEBUG] [1636563254.041140865] [moveit_kinematics_base.kinematics_base]: [  1] delta_twist: [   -0.206502,           0,    0.105197,-1.90726e-25,   -0.142507,-2.67221e-24]
> [DEBUG] [1636563254.041159139] [moveit_kinematics_base.kinematics_base]: [  1] pos err: 0,231753  rot err: 0,142507  delta_q: 0,013814
> [DEBUG] [1636563254.041167435] [moveit_kinematics_base.kinematics_base]:       delta_q: [ 7.62809e-25           0  0.00208182 8.01646e-24  -0.0117324-7.54595e-24]
> [DEBUG] [1636563254.041173115] [moveit_kinematics_base.kinematics_base]:       q: [ 7.62809e-25          -1   -0.698664 8.01646e-24   -0.453492-7.54595e-24]
> [DEBUG] [1636563254.041180138] [moveit_kinematics_base.kinematics_base]: [  2] delta_twist: [   -0.204608, 2.05219e-26,    0.104368,-2.17532e-25,   -0.152157, 9.82378e-26]
> [DEBUG] [1636563254.041187242] [moveit_kinematics_base.kinematics_base]: [  2] pos err: 0,229689  rot err: 0,152157  delta_q: 0,000312
> [DEBUG] [1636563254.041194225] [moveit_kinematics_base.kinematics_base]:       delta_q: [  1.6109e-25           0 0.000285058-1.40926e-25-2.71725e-05 3.56224e-25]
> [DEBUG] [1636563254.041199395] [moveit_kinematics_base.kinematics_base]:       q: [ 9.23899e-25          -1   -0.698379 7.87553e-24    -0.45352-7.18972e-24]
> ```
> 
> I dived into MoveIt 2 sources and discovered that there is a loop in `searchPositionIK` method that should iteratively try to calculate the IK solution. See here https://github.com/ros-planning/moveit2/blob/fd9a562830bd5b3720905cd01dbf97e7591ca31e/moveit_kinematics/kdl_kinematics_plugin/src/kdl_kinematics_plugin.cpp#L381 ... This is the outer loop that is limited by timeout. There is also an inner loop in `CartToJnt` method that is limited by a maximum number of iterations. This number of iterations can be specified in configuration, but the main problem is the timeout.
> 
> See method `timedOut` here https://github.com/ros-planning/moveit2/blob/fd9a562830bd5b3720905cd01dbf97e7591ca31e/moveit_kinematics/kdl_kinematics_plugin/src/kdl_kinematics_plugin.cpp#L269 ... it simply calls `node_->now()` and the same call is also used to get the start time. All this runs inside RViz2 node (not move group node). RViz2 is single threaded application (probably because of Qt). Then it occured to me! We are using __simulation time__ (when running in simulation), in ROS2 the simulation time is done by publishing to `/clock` topic and using `use_sim_time` parameter on all the nodes. Each node then subscribes to this topic and updates its internal time in the subscriber callback. But this callback has to wait in the executor queue when it is handling some other task. That's why each call to `node_->now()` in this iterative IK solution finding process always returns the same value, therefore the timeout never happens!
> 
> Some related links:
> 
> * https://github.com/ros2/rclcpp/issues/1540
> * https://github.com/ros2/rclcpp/issues/1542
> * both solved by PR https://github.com/ros2/rclcpp/pull/1556 merged into ROS2 Galactic, so Foxy still has this problem
> * https://github.com/ros2/rclpy/issues/792 - the same problem, but for rclpy, not yet solved
> 
> This problem with RViz2 I can live with. But maybe we will get into similar problems when running in simulation when doing other things? What about transform listener buffer? See https://answers.ros.org/question/307682/how-to-set-up-tfbuffer-clock-in-ros-2-so-it-works-with-sim-time/ ... it's from 2018 and I did not check it in Foxy and also it is solvable by using multithreaded executor or spin in a thread, but similar constructs could be a problem (mainly in a code that we do not control).

So the problem occurs because we are using simulation time. It's probably already solved in Galactic+ because there the handling of the clock subscriber is different (see links above). So in Galactic and newer versions (main branch) this PR is probably not neccessary?

It would be very nice to have the changes from PR !1556 backported into Foxy. Because I think that there are other places with similar problem waiting to be discovered :( Just quick scan over moveit2 sources revealed at least dozen similar cases across robot state, inverse kinematics plugins, motion planning and others.

Also `rclpy` should be updated to behave in the same way.

galou:
This branch is now out-of-date, should I update it for a possible merge?

galou:
What are the next steps?

The testspace-analytics seemed to have succeeded though it's marked as queued here at the time of writing this.

vatanaksoytezer:
> The testspace-analytics seemed to have succeeded though it's marked as queued here at the time of writing this.

GH is acting a little weird today and throwing out http errors. I think testspace is still running, I'll re-run it if it fails / gets stuck. Otherwise I'll drop an official review and get it ready to merge.

henningkayser (approved):
Sorry for letting this get stale. It should have been merged and backported a long time ago. Even if the issue doesn't really occur in Galactic/Rolling, it does make sense to not use simulated time for tasks like this. LGTM

Issue #790, opened 2021-11-09
Use ros2 interfaces for parameters in moveit_servo


tylerjw:
Currently, moveit_servo uses a template function declareOrGetParam for specifying and getting parameters.  Here is the file where that is done: https://github.com/ros-planning/moveit2/blob/main/moveit_ros/moveit_servo/src/servo_parameters.cpp

I would like to use the ros2 parameter interfaces directly in order to specify descriptions and types of each parameter using the ParameterDescription.  An example of that is here: https://github.com/tylerjw/parameter_tools/blob/parameter_set/parameter_set/src/example_node/robot_parameters.cpp

Note that I created a ParaeterDescriptionBuilder.  Copy that into servo for now and eventually, we'll find a home for it.

### Steps

- [ ] Define defaults for all parameters in servo_paramets.h in the struct
- [ ] Copy ParaeterDescriptionBuilder into servo
- [ ] Create a static declare servo_pareters.cpp that uses the built in ros2 interfaces for parameters
- [ ] Use NodeParametersInterface::SharedPtr in the builder function for ServoParameters instead of the Node directly
- [ ] Create a get function that reads the parameters into the struct fields
- [ ] Modify the builder function for ServoParameters to use the get and declare functions you created

tylerjw:
@henningkayser I'll respond here to your comment in the PR: https://github.com/ros-planning/moveit2/pull/799#pullrequestreview-809715634

>There still is way too much boilerplate code and unnecessary duplication, inviting for user errors and frustrating debugging sessions (overstated, but still this this should be a lot simpler and shorter)

I'm not sure it is as bad as you seem to think it is.  It could be better but this doesn't seem like the largest evil in the parameter interface in ROS 2 to me.

> The descriptor constraints are somewhat useless if they don't provide the implementation for the validate function, additional_constraints are even more useless in this context

I don't agree that the additional_constraints string is useless.  I do think it would be a really nice upstream improvement for there to be an interface for declaring per-parameter validation functions.  I do think having a separate string from the description for constraints is a nice part of the interface.

> The application code is responsible for handling parameter callbacks while the config itself is not being updated
Intermixed runtime dependencies on rclcpp, this doesn't seem to go well with mock tests

This is true.  I do hope to fix this and I have a plan for how to work around this problem.  I didn't include that in what I asked for in this issue to limit the amount of work needed for this.

> No support for parameter namespaces, let alone group parameter constraints which are de facto implemented in the validate function.

[...]

PR #799, opened 2021-11-12
Change servo parameters to use ros2 interfaces for parameters

stephanie-eng:
### Description
Update servoParameters to use a new declare and get function that uses NodeParametersInterface to declare and get parameters. See #790

This PR  does the following:
- Sets default values for servo parameters based off of panda_simulated_config.yaml
- Replaces the getOrDeclareParam function with individual get and set functions
  - These 
- Adds parameter_descriptor_builder, which is used to create the messages used in setting up the descriptions for parameters

TODO:

- [x] When running the joystick servo example, declare will throw an error saying that the parameters are already declared. Not sure why this is happening yet, still working on this.
- [x] Build error with the set parameters callback

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/include/moveit_servo/parameter_descriptor_builder.hpp
+++ b/moveit_ros/moveit_servo/include/moveit_servo/parameter_descriptor_builder.hpp
@@ -1,0 +1,123 @@
+// Copyright 2021 PickNik Inc.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//    * Redistributions of source code must retain the above copyright
+//      notice, this list of conditions and the following disclaimer.
+//
+//    * Redistributions in binary form must reproduce the above copyright
+//      notice, this list of conditions and the following disclaimer in the
+//      documentation and/or other materials provided with the distribution.
+//
+//    * Neither the name of the PickNik Inc. nor the names of its
+//      contributors may be used to endorse or promote products derived from
+//      this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+#pragma once
+
+#include <rcl_interfaces/msg/floating_point_range.hpp>
+#include <rcl_interfaces/msg/integer_range.hpp>
+#include <rcl_interfaces/msg/parameter_descriptor.hpp>
+#include <rcl_interfaces/msg/parameter_type.hpp>
+#include <string>
+
+namespace moveit_servo
+{
+class ParameterDescriptorBuilder
+{
+  rcl_interfaces::msg::ParameterDescriptor msg_;
+
+public:
+  /**
+   * @brief      Rcl_interfaces::msg::parameterdescriptor conversion operator.
+   */
+  operator rcl_interfaces::msg::ParameterDescriptor() const
+  {
+    return msg_;
+  }
+
+  /**
+   * @brief      Set the type
+   *
+   * @param[in]  type  The type
+   *
+   * @return     Reference to this object
+   */
+  ParameterDescriptorBuilder& type(uint8_t type);
+
+  /**
+   * @brief      Set the description string
+   *
+   * @param[in]  description  The description
+   *
+   * @return     Reference to this object
+   */
+  ParameterDescriptorBuilder& description(std::string description);
+
+  /**
+   * @brief      Set the additional constraints string (a description of any additional constraints which cannot be
+   * expressed with the available parameter constraints)
+   *
+   * @param[in]  additional_constraints  The additional constraints
+   *
+   * @return     Reference to this object
+   */
+  ParameterDescriptorBuilder& additionalConstraints(std::string additional_constraints);
+
+  /**
+   * @brief      Sets the read only flag
+   *
+   * @param[in]  read_only  The read only flag
+   *
+   * @return     Reference to this object
+   */
+  ParameterDescriptorBuilder& readOnly(bool read_only);
+
+  /**
+   * @brief      Set the dynamic typing flag (rolling only)
+   *
+   * @param[in]  dynamic_typing  The dynamic typing flag
+   *
+   * @return     Reference to this object
+   */
+  ParameterDescriptorBuilder& dynamicTyping(bool dynamic_typing);
+
+  /**
+   * @brief      Set floating point range
+   *
+   * @param[in]  from_value  The from value
+   * @param[in]  to_value    To value
+   * @param[in]  step        The step
+   *
+   * @return     Reference to this object
+   */
+  ParameterDescriptorBuilder& floatingPointRange(double from_value = std::numeric_limits<double>::min(),
+                                                 double to_value = std::numeric_limits<double>::max(), double step = 0);
+
+  /**
+   * @brief      Set the integer range
+   *
+   * @param[in]  from_value  The from value
+   * @param[in]  to_value    To value
+   * @param[in]  step        The step
+   *
+   * @return     Reference to this object
+   */
+  ParameterDescriptorBuilder& integerRange(int64_t from_value = std::numeric_limits<int64_t>::min(),
+                                           int64_t to_value = std::numeric_limits<int64_t>::max(), int64_t step = 0);
+};
+
+}  // namespace moveit_servo
--- a/moveit_ros/moveit_servo/include/moveit_servo/parameter_descriptor_builder.hpp
+++ b/moveit_ros/moveit_servo/include/moveit_servo/parameter_descriptor_builder.hpp
@@ -26,6 +26,14 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
+/* Author    : Tyler Weaver
+   Desc      : Creates a parameter descriptor message used to describe parameters
+   Title     : parameter_descriptor_builder.hpp
+   Project   : moveit_servo
+*/
+
+// TODO(823): Move this into a separate message_builder package
+
 #pragma once
 
 #include <rcl_interfaces/msg/floating_point_range.hpp>
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo_parameters.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo_parameters.h
@@ -42,6 +42,7 @@
 #include <thread>
 #include <vector>
 
+#include <moveit/planning_scene_monitor/planning_scene_monitor.h>
 #include <rclcpp/rclcpp.hpp>
 
 namespace moveit_servo
@@ -59,47 +60,49 @@ struct ServoParameters
   // ROS Parameters
   // Note that all of these are effectively const because the only way to create one of these
   //  is as a shared_ptr to a constant struct.
-  bool use_gazebo;
-  std::string status_topic;
+  bool use_gazebo{ false };
+  std::string status_topic{ "~/status" };
   // Properties of incoming commands
-  std::string cartesian_command_in_topic;
-  std::string joint_command_in_topic;
-  std::string robot_link_command_frame;
-  std::string command_in_type;
-  double linear_scale;
-  double rotational_scale;
-  double joint_scale;
+  std::string cartesian_command_in_topic{ "~/delta_twist_cmds" };
+  std::string joint_command_in_topic{ "~/delta_joint_cmds" };
+  std::string robot_link_command_frame{ "panda_link0" };
+  std::string command_in_type{ "unitless" };
+  double linear_scale{ 0.4 };
+  double rotational_scale{ 0.8 };
+  double joint_scale{ 0.5 };
   // Properties of outgoing commands
-  std::string command_out_topic;
-  double publish_period;
-  std::string command_out_type;
-  bool publish_joint_positions;
-  bool publish_joint_velocities;
-  bool publish_joint_accelerations;
+  std::string command_out_topic{ "/panda_arm_controller/joint_trajectory" };
+  double publish_period{ 0.034 };
+  std::string command_out_type{ "trajectory_msgs/JointTrajectory" };
+  bool publish_joint_positions{ true };
+  bool publish_joint_velocities{ true };
+  bool publish_joint_accelerations{ false };
   // Plugins for smoothing outgoing commands
-  std::string joint_topic;
-  std::string smoothing_filter_plugin_name;
+  std::string joint_topic{ "/joint_states" };
+  std::string smoothing_filter_plugin_name{ "online_signal_smoothing::ButterworthFilterPlugin" };
   // MoveIt properties
-  std::string move_group_name;
-  std::string planning_frame;
-  std::string ee_frame_name;
-  bool is_primary_planning_scene_monitor;
-  std::string monitored_planning_scene_topic;
+  std::string move_group_name{ "panda_arm" };
+  std::string planning_frame{ "panda_link0" };
+  std::string ee_frame_name{ "panda_link8" };
+  bool is_primary_planning_scene_monitor = { true };
+  std::string monitored_planning_scene_topic{
+    planning_scene_monitor::PlanningSceneMonitor::DEFAULT_PLANNING_SCENE_TOPIC
+  };
   // Stopping behaviour
-  double incoming_command_timeout;
-  int num_outgoing_halt_msgs_to_publish;
-  bool halt_all_joints_in_joint_mode;
-  bool halt_all_joints_in_cartesian_mode;
+  double incoming_command_timeout{ 0.1 };
+  int num_outgoing_halt_msgs_to_publish{ 4 };
+  bool halt_all_joints_in_joint_mode{ true };
+  bool halt_all_joints_in_cartesian_mode{ true };
   // Configure handling of singularities and joint limits
-  double lower_singularity_threshold;
-  double hard_stop_singularity_threshold;
-  double joint_limit_margin;
-  bool low_latency_mode;
+  double lower_singularity_threshold{ 17.0 };
+  double hard_stop_singularity_threshold{ 30.0 };
+  double joint_limit_margin{ 0.1 };
+  bool low_latency_mode{ false };
   // Collision checking
-  bool check_collisions;
-  double collision_check_rate;
-  double self_collision_proximity_threshold;
-  double scene_collision_proximity_threshold;
+  bool check_collisions{ true };
+  double collision_check_rate{ 10.0 };
+  double self_collision_proximity_threshold{ 0.01 };
+  double scene_collision_proximity_threshold{ 0.02 };
 
   /**
    * Declares, reads, and validates parameters used for moveit_servo
@@ -113,8 +116,8 @@ struct ServoParameters
    * @param dynamic_parameters Enable dynamic parameter handling. (default: true)
    * @return std::shared_ptr<ServoParameters> if all parameters were loaded and verified successfully, nullptr otherwise
    */
-  static SharedConstPtr makeServoParameters(const rclcpp::Node::SharedPtr& node, const rclcpp::Logger& logger,
-                                            std::string ns = "moveit_servo", bool dynamic_parameters = true);
+  static SharedConstPtr makeServoParameters(const rclcpp::Node::SharedPtr& node, std::string ns = "moveit_servo",
+                                            bool dynamic_parameters = true);
 
   /**
    * Register a callback for a parameter set event.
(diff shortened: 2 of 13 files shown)


moveit_ros/moveit_servo/src/servo_parameters.cpp
```diff
@@ -236,6 +421,26 @@ ServoParameters::SharedConstPtr ServoParameters::makeServoParameters(const rclcp
...
+  auto node_parameters = node->get_node_parameters_interface();
+  // Get the parameters (organized same order as YAML file)
+  try
+  {
+    declare(ns, node_parameters, parameters);
+  }
+  catch (rclcpp::exceptions::ParameterAlreadyDeclaredException& e)
+  {
+    RCLCPP_WARN_STREAM(logger, "Parameters already declared.");
+  }
```
stephanie-eng, line 442: When running the joystick servo example, this will throw an exception that suggests the paramaters have already been declared before the declare function gets to do it. Catching this right now just as a test. Will try to fix later.

moveit_ros/moveit_servo/src/parameter_descriptor_builder.cpp
```diff
@@ -0,0 +1,89 @@
...
+  return *this;
+}
+
+// In rolling (not in Foxy)
+// ParameterDescriptorBuilder& ParameterDescriptorBuilder::dynamic_typing(bool
+// dynamic_typing)
+// {
+//   msg_.dynamic_typing = dynamic_typing;
+//   return *this;
+// }
```
tylerjw, line 63: You should be able to uncomment this.  Also, please look into the ParameterDescriptor to make sure there are not any new features that have been added since I did this.

moveit_ros/moveit_servo/src/servo_parameters.cpp
```diff
@@ -236,6 +422,19 @@ ServoParameters::SharedConstPtr ServoParameters::makeServoParameters(const rclcp
...
   }
+  return parameters;
+}
+
+ServoParameters::SharedConstPtr ServoParameters::makeServoParameters(const rclcpp::Node::SharedPtr& node,
+                                                                     const rclcpp::Logger& logger,
+                                                                     std::string ns, /* = "moveit_servo"*/
+                                                                     bool dynamic_parameters /* = true */)
+{
+  auto parameters = std::shared_ptr<ServoParameters>(new ServoParameters(logger, ns));
```
tylerjw, line 433: Use make_shared here.  Reference: https://stackoverflow.com/questions/20895648/difference-in-make-shared-and-normal-shared-ptr-in-c

moveit_ros/moveit_servo/src/servo_parameters.cpp
```diff
@@ -40,6 +40,7 @@
 */
 
 #include <moveit_servo/servo_parameters.h>
+#include "../include/moveit_servo/parameter_descriptor_builder.hpp"
```
tylerjw, line 43: This should be <moveit_servo/parameter_descriptor_builder.hpp>

moveit_ros/moveit_servo/src/servo_parameters.cpp
```diff
@@ -77,74 +78,259 @@ ServoParameters::setParametersCallback(std::vector<rclcpp::Parameter> parameters
...
-ServoParameters::SharedConstPtr ServoParameters::makeServoParameters(const rclcpp::Node::SharedPtr& node,
-                                                                     const rclcpp::Logger& logger,
-                                                                     std::string ns, /* = "moveit_servo"*/
-                                                                     bool dynamic_parameters /* = true */)
+void ServoParameters::declare(const std::string& ns,
+                              const rclcpp::node_interfaces::NodeParametersInterface::SharedPtr& node_parameters,
+                              std::shared_ptr<ServoParameters> parameters)
 {
-  auto parameters = std::shared_ptr<ServoParameters>(new ServoParameters(logger, ns));
+  // Properties of incoming commands
```
tylerjw, line 85: What does this mean?

moveit_ros/moveit_servo/src/servo_parameters.cpp
```diff
@@ -77,74 +78,259 @@ ServoParameters::setParametersCallback(std::vector<rclcpp::Parameter> parameters
...
+                                     parameter_set::ParameterDescriptorBuilder()
+                                         .type(rclcpp::ParameterType::PARAMETER_DOUBLE)
+                                         .description("Start decelerating when a scene collision is this far [m]"));
+}
 
+void ServoParameters::get(const std::string& ns,
+                          const rclcpp::node_interfaces::NodeParametersInterface::SharedPtr& node_parameters,
+                          std::shared_ptr<ServoParameters> parameters)
+{
   // Properties of incoming commands
```
tylerjw, line 270: This comment seems out of place.

moveit_ros/moveit_servo/src/servo_parameters.cpp
```diff
@@ -77,74 +78,259 @@ ServoParameters::setParametersCallback(std::vector<rclcpp::Parameter> parameters
...
+  node_parameters->declare_parameter(ns + ".scene_collision_proximity_threshold",
+                                     rclcpp::ParameterValue(parameters->scene_collision_proximity_threshold),
+                                     parameter_set::ParameterDescriptorBuilder()
+                                         .type(rclcpp::ParameterType::PARAMETER_DOUBLE)
+                                         .description("Start decelerating when a scene collision is this far [m]"));
+}
 
+void ServoParameters::get(const std::string& ns,
+                          const rclcpp::node_interfaces::NodeParametersInterface::SharedPtr& node_parameters,
+                          std::shared_ptr<ServoParameters> parameters)
```
tylerjw, line 268: Instead of using an output parameter here, this should just return a new ServoParameters object.  Something like this should work:

```suggestion
ServoParameters ServoParameters::get(const std::string& ns,
                          const rclcpp::node_interfaces::NodeParametersInterface::SharedPtr& node_parameters)
```

Then since this function is a member of the class it should be able to use the private constructor.

moveit_ros/moveit_servo/src/servo_parameters.cpp
```diff
@@ -77,74 +78,259 @@ ServoParameters::setParametersCallback(std::vector<rclcpp::Parameter> parameters
   return result;
 }
 
-ServoParameters::SharedConstPtr ServoParameters::makeServoParameters(const rclcpp::Node::SharedPtr& node,
-                                                                     const rclcpp::Logger& logger,
-                                                                     std::string ns, /* = "moveit_servo"*/
-                                                                     bool dynamic_parameters /* = true */)
+void ServoParameters::declare(const std::string& ns,
+                              const rclcpp::node_interfaces::NodeParametersInterface::SharedPtr& node_parameters,
+                              std::shared_ptr<ServoParameters> parameters)
```
tylerjw, line 83: Similar to the get function, this shouldn't need parameters shared ptr parameter.  Also, as declaring does not set any values (it only describes the parameter interface to the ros system) it shouldn't set values in a parameters object.  Internally you should just create a temporary one for the default values.  Something like this should work:

```suggestion
void ServoParameters::declare(const std::string& ns,
                              const rclcpp::node_interfaces::NodeParametersInterface::SharedPtr& node_parameters)
{
  auto parameters = ServoParameters{ns};  // for populating default values
```

Lastly, you should be able to get rid of the logger out of the constructor and just put one in this file's scope similar to what is done in other source files in moveit_servo.

tylerjw (commented):
I left a few comments, let me know if you have questions.

moveit_ros/moveit_servo/src/servo_parameters.cpp
stephanie-eng: See https://github.com/ros-planning/moveit2/blob/b426a1c6f957522f8261f3500ee11f82199cdef6/moveit_ros/moveit_servo/src/servo_parameters.cpp#L91

I believe is saying that the parameters below are properties of incoming commands (ex. command_in_topic)

moveit_ros/moveit_servo/src/servo_parameters.cpp
stephanie-eng: From here:
https://github.com/ros-planning/moveit2/blob/b426a1c6f957522f8261f3500ee11f82199cdef6/moveit_ros/moveit_servo/src/servo_parameters.cpp#L91-L99

I believe is saying that the parameters below are properties of incoming commands (ex. command_in_topic)

moveit_ros/moveit_servo/src/servo_parameters.cpp
```diff
@@ -77,175 +80,381 @@ ServoParameters::setParametersCallback(std::vector<rclcpp::Parameter> parameters
...
-    parameters->on_set_parameters_callback_handler_ =
-        node->add_on_set_parameters_callback(std::bind(&ServoParameters::setParametersCallback, parameters.get(), _1));
-  }
+    // register parameter change callback
+    if (dynamic_parameters)
+    {
+      // using std::placeholders::_1;
+      // parameters.value().on_set_parameters_callback_handler_ = node->add_on_set_parameters_callback(
+      //     std::bind(&ServoParameters::setParametersCallback, parameters.value(), _1));
+    }
```
stephanie-eng, line 453: TBD: Uncomment when this stops causing a build error

moveit_ros/moveit_servo/src/servo_parameters.cpp
```diff
@@ -77,175 +80,381 @@ ServoParameters::setParametersCallback(std::vector<rclcpp::Parameter> parameters
...
-    return nullptr;
+    return std::nullopt;
   }
+  return parameters;
+}
 
-  // register parameter change callback
-  if (dynamic_parameters)
+ServoParameters::SharedConstPtr ServoParameters::makeServoParameters(const rclcpp::Node::SharedPtr& node,
+                                                                     const rclcpp::Logger logger,  // remove?
```
stephanie-eng, line 435: TBD: Probably can remove this now that we don't need logger in the constructor. Removing this will mean we have to edit function calls in other files as well.

moveit_ros/moveit_servo/src/servo_parameters.cpp
AndyZe: ^yep, exactly

AndyZe:
In my first time testing this, it fails due to an undeclared parameter. It's nice that it failed and printed a useful error message but the error message could be formatted better:

```
[servo_node_main-3] terminate called after throwing an instance of 'rclcpp::exceptions::ParameterNotDeclaredException'
[servo_node_main-3]   what():  moveit_servocommand_out_topic
```

Can you please update this with a bit of a description of `moveit_servocommand_out_topic`? Otherwise it is easy to miss.

```
[servo_node_main-3] terminate called after throwing an instance of 'rclcpp::exceptions::ParameterNotDeclaredException'
[servo_node_main-3]   Parameter name:  moveit_servocommand_out_topic
```

moveit_ros/moveit_servo/include/moveit_servo/servo_parameters.h
```diff
@@ -43,6 +43,7 @@
 #include <vector>
 
 #include <rclcpp/rclcpp.hpp>
+#include <moveit/planning_scene_monitor/planning_scene_monitor.h>
```
AndyZe, line 46: Nitpick: please alphabetize. I think the order should be: system includes, then MoveIt includes.

```
#include <mutex>
#include <thread>
#include <vector>

#include <moveit/planning_scene_monitor/planning_scene_monitor.h>
#include <rclcpp/rclcpp.hpp>
```

AndyZe (commented):
@tylerjw and @stephanie-eng can you please add to the PR description -- **why** should we use parameter_descriptor_builder? I guess it has to do with setting parameters dynamically.

Without a better PR description, it kind of looks like 500 lines of code have been added without much change in functionality.

moveit_ros/moveit_servo/src/servo_parameters.cpp
AndyZe: ^Stephanie's right.  :+1:

stephanie-eng:
> In my first time testing this, it fails due to an undeclared parameter. It's nice that it failed and printed a useful error message but the error message could be formatted better:
> 
> ```
> [servo_node_main-3] terminate called after throwing an instance of 'rclcpp::exceptions::ParameterNotDeclaredException'
> [servo_node_main-3]   what():  moveit_servocommand_out_topic
> ```
> 
> Can you please update this with a bit of a description of `moveit_servocommand_out_topic`? Otherwise it is easy to miss.
> 
> ```
> [servo_node_main-3] terminate called after throwing an instance of 'rclcpp::exceptions::ParameterNotDeclaredException'
> [servo_node_main-3]   Parameter name:  moveit_servocommand_out_topic
> ```
> 
> Or incorporate it all in one line. That would be even better.
> 
> `[servo_node_main-3] terminate called after throwing an instance of 'rclcpp::exceptions::ParameterNotDeclaredException'. Parameter name: moveit_servocommand_out_topic`

This was due to me having a typo with some of the parameters. The parameter name causing this issue is hard-coded, so now that it's been corrected, is it still worth to have a specific error message?

tylerjw:
@AndyZe I'll respond to the parameter descriptor builder questions.  Long term I hope to put something like this in a much more general place than in servo.  It has to do with adding descriptions and types to the parameter declarations.  

ROS2 has this object (ParameterDescriptor) that it can take in the declare_parameter functions in order to provide a description and other details about the parameter that the ros2 system can use to help the user when interacting with the node.  Unfortunately, because this object is a ROS2 message the code is much longer and uglier if you don't have the ParameterDescriptorBuilder here.  An example of it in ROS2 is here: https://github.com/ros2/rclcpp/blob/1037822a63330495dcf0de5e8f20544375a5f116/rclcpp/include/rclcpp/detail/qos_parameters.hpp#L163  The ParameterDescriptorBuilder makes this much nicer when you have to create a bunch of these (because you have a bunch of parameters).

One large advantage of doing this is a user will be able to run `ros2 param dump <servo_node_name>` and it will output a yaml file of the parameters.  Another thing you can run is `ros2 param list <servo_node_name>` and it will output a list of the parameters and their descriptions.

This is also needed to support dynamic reconfiguring in the future.  As to the features of ParameterDescriptor, additional constraints as I understand it is a string where you can describe things like how the values set in that parameter affect other parameters.  You might put something like "must be a valid ros topic name" or "only considered if <other parameter> is true" or something like that.

AndyZe:
> This was due to me having a typo with some of the parameters. The parameter name causing this issue is hard-coded, so now that it's been corrected, is it still worth to have a specific error message?

No worries, I know this is still a draft. What I'm wondering is, can you somehow print the missing parameter name in the same line as the rest of the error message? In a general way, no matter what parameter it is. If not, it's not a huge deal. Like this:

`[servo_node_main-3] terminate called after throwing an instance of 'rclcpp::exceptions::ParameterNotDeclaredException'. Parameter name: whatever_my_parameter_name_is`

moveit_ros/moveit_servo/include/moveit_servo/parameter_descriptor_builder.hpp
```diff
@@ -0,0 +1,124 @@
...
+  // Rolling only
+
+  /**
+   * @brief      Set the dynamic typing flag (rolling only)
+   *
+   * @param[in]  dynamic_typing  The dynamic typing flag
+   *
+   * @return     Reference to this object
+   */
+  ParameterDescriptorBuilder& dynamic_typing(bool dynamic_typing);
```
henningkayser, line 97: Should this function be camel case like the others?

henningkayser (commented):
First off, I think this is a really good improvement as this approach provides much more consistency. I'd really appreciate if we could remove `automatically_declare_parameters_from_overrides` in all nodes. I don't have much to add to this PR as this is a step into the right direction.

However, I have plenty of feedback and questions that address the new helper API and this use case is a good opportunity to continue the design discussions we started half a year ago (again, this is not meant to block this PR, we should even get this merged imo):

* There still is way too much boilerplate code and unnecessary duplication, inviting for user errors and frustrating debugging sessions (overstated, but still this this should be a lot simpler and shorter)
* The descriptor constraints are somewhat useless if they don't provide the implementation for the `validate` function, `additional_constraints` are even more useless in this context
* The application code is responsible for handling parameter callbacks while the config itself is not being updated
* Intermixed runtime dependencies on rclcpp, this doesn't seem to go well with mock tests
* No support for parameter namespaces, let alone group parameter constraints which are de facto implemented in the validate function.

I'd really like to move towards an API that meets all of these requirements. While I agree with @tylerjw that we shouldn't abstract completely from rclcpp, it's unlikely that rclcpp will provide these any time soon, unless we implement something that works for us first and try to bring it upstream. The goal should really be something that is just as easy to use as `rosparam_shortcuts` while allowing us to work on the actually challenging problems like:
* Support for config profiles and generic parameter types (=parameter groups)
* Synchronization of parameter groups (e.g. transforms, profiles, robot model etc)
* Validating, versioning and generating of (yaml) config files

With that, I think it makes sense to continue the discussion in #790, https://github.com/PickNikRobotics/parameter_tools, or even https://github.com/PickNikRobotics/ros2_params. I'll put in more time to rework some of the interfaces and make suggestions on how to simplify things without abstracting too much.

henningkayser:
> > ```
> > * No support for parameter namespaces
> > ```
> 
> Does this work with namespaces? That's an absolute must. We have a dual-arm robot project starting up right now.

It does, but you have to stitch them together like ```ns + ".parameter"```

moveit_ros/moveit_servo/include/moveit_servo/parameter_descriptor_builder.hpp
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+   * @brief      Sets the read only flag
+   *
+   * @param[in]  read_only  The read only flag
+   *
+   * @return     Reference to this object
+   */
+  ParameterDescriptorBuilder& readOnly(bool read_only);
+  // Rolling only
```
tylerjw, line 88: Please remove this comment.

moveit_ros/moveit_servo/include/moveit_servo/servo_parameters.h
```diff
@@ -53,53 +54,76 @@ struct ServoParameters
 {
   using SharedConstPtr = std::shared_ptr<const ServoParameters>;
 
+  // Explicitly define copy constructor and assignment to avoid copying the mutex
+  // ServoParameters(const ServoParameters& s) = default;
+  // ServoParameters& operator=(const ServoParameters& s) : ServoParameters{ s }
```
tylerjw, line 59: This isn't needed anymore, so you should remove it.

moveit_ros/moveit_servo/include/moveit_servo/servo_parameters.h
```diff
@@ -53,53 +54,76 @@ struct ServoParameters
...
   // Collision checking
-  bool check_collisions;
-  double collision_check_rate;
-  double self_collision_proximity_threshold;
-  double scene_collision_proximity_threshold;
+  bool check_collisions = true;
+  double collision_check_rate = 10.0;
+  double self_collision_proximity_threshold = 0.01;
+  double scene_collision_proximity_threshold = 0.02;
+  // Pose Tracking parameters
```
tylerjw, line 109: Please don't put the pose tracking parameters in here.  Please create a separate pose tracking parameters object that uses the same pattern and declares/gets these.

moveit_ros/moveit_servo/src/parameter_descriptor_builder.cpp
```diff
@@ -0,0 +1,94 @@
...
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+/* Author    : Tyler Weaver
+   Desc      : Creates a parameter descriptor message used to describe parameters
```
tylerjw, line 30: @AndyZe here is the description of this.  I will think of a nicer place to put this than in servo.  I think having builders for some commonly used message types in moveit would be fine and so I can create a package for these.

moveit_ros/moveit_servo/src/servo_parameters.cpp
```diff
@@ -77,175 +83,386 @@ ServoParameters::setParametersCallback(std::vector<rclcpp::Parameter> parameters
...
 
-ServoParameters::SharedConstPtr ServoParameters::makeServoParameters(const rclcpp::Node::SharedPtr& node,
-                                                                     const rclcpp::Logger& logger,
-                                                                     std::string ns, /* = "moveit_servo"*/
-                                                                     bool dynamic_parameters /* = true */)
+void ServoParameters::declare(const std::string& ns,
+                              const rclcpp::node_interfaces::NodeParametersInterface::SharedPtr& node_parameters)
 {
-  auto parameters = std::shared_ptr<ServoParameters>(new ServoParameters(logger, ns));
+  auto parameters = ServoParameters();
```
tylerjw, line 89:
```suggestion
  auto parameters = ServoParameters{};
```

moveit_ros/moveit_servo/src/servo_parameters.cpp
```diff
@@ -77,175 +83,386 @@ ServoParameters::setParametersCallback(std::vector<rclcpp::Parameter> parameters
...
 
-  // Get the parameters (organized same order as YAML file)
-  declareOrGetParam<bool>(parameters->use_gazebo, ns + ".use_gazebo", node, logger);
-  declareOrGetParam<std::string>(parameters->status_topic, ns + ".status_topic", node, logger);
+  // ROS Parameters
+  node_parameters->declare_parameter(
+      ns + ".use_gazebo", rclcpp::ParameterValue(parameters.use_gazebo),
+      parameter_set::ParameterDescriptorBuilder()
```
tylerjw, line 94: ParameterDescriptorBuilder doesn't need to be in a separate `parameter_set` namespace.  It can just be in the moveit_servo namespace for now.

moveit_ros/moveit_servo/src/servo_parameters.cpp
```diff
@@ -77,175 +83,386 @@ ServoParameters::setParametersCallback(std::vector<rclcpp::Parameter> parameters
...
+          .type(rclcpp::ParameterType::PARAMETER_BOOL)
```
tylerjw, line 95: You can make all of this easier to read with a few using statements at the top of this function for the types.  For example:

```
using rclcpp::ParameterValue;
using rclcpp::ParameterType::PARAMETER_BOOL;
using rclcpp::ParameterType::PARAMETER_STRING;
```

tylerjw (commented):
A couple of small style things, but otherwise this looks really good.  I'll test this later today to verify it doesn't change the behavior.

Have you tried running the tutorial after these changes?

moveit_ros/moveit_servo/src/servo_parameters.cpp
```diff
@@ -77,175 +83,363 @@ ServoParameters::setParametersCallback(std::vector<rclcpp::Parameter> parameters
...
+  using rclcpp::ParameterType::PARAMETER_INTEGER;
+  using rclcpp::ParameterType::PARAMETER_STRING;
+  auto parameters = ServoParameters{};
 
-  // Get the parameters (organized same order as YAML file)
-  declareOrGetParam<bool>(parameters->use_gazebo, ns + ".use_gazebo", node, logger);
-  declareOrGetParam<std::string>(parameters->status_topic, ns + ".status_topic", node, logger);
+  // ROS Parameters
+  node_parameters->declare_parameter(
+      ns + ".use_gazebo", ParameterValue(parameters.use_gazebo),
```
tylerjw, line 98: I'm sorry to be nit-picky about this but we should use brace initialization by default.

```suggestion
      ns + ".use_gazebo", ParameterValue{parameters.use_gazebo},
```

moveit_ros/moveit_servo/src/servo_parameters.cpp
```diff
@@ -77,175 +83,363 @@ ServoParameters::setParametersCallback(std::vector<rclcpp::Parameter> parameters
...
+      ParameterDescriptorBuilder()
```
tylerjw, line 99:
```suggestion
      ParameterDescriptorBuilder{}
```

stephanie-eng:
> Have you tried running the tutorial after these changes?
I've been running the joystick and pose tracking examples from the launch folder as a sanity check. I don't have a joystick handy, so I can't move the arm for the teleop demo, but otherwise seems to work.

moveit_ros/moveit_servo/include/moveit_servo/servo_parameters.h
```diff
@@ -59,47 +60,48 @@ struct ServoParameters
...
-  std::string command_out_type;
-  bool publish_joint_positions;
-  bool publish_joint_velocities;
-  bool publish_joint_accelerations;
+  std::string command_out_topic = "/panda_arm_controller/joint_trajectory";
+  double publish_period = 0.034;
+  std::string command_out_type = "trajectory_msgs/JointTrajectory";
+  bool publish_joint_positions = true;
+  bool publish_joint_velocities = true;
+  bool publish_joint_accelerations = false;
```
tylerjw, line 79: Do you mind using uniform initialization for these too?

moveit_ros/moveit_servo/include/moveit_servo/servo_parameters.h
```diff
@@ -123,56 +125,43 @@ struct ServoParameters
...
    * @param callback function to call when parameter is changed
    */
-  void registerSetParameterCallback(const std::string name, SetParameterCallbackType callback) const
+  [[nodiscard]] bool registerSetParameterCallback(const std::string name, SetParameterCallbackType callback) const
   {
-    const std::lock_guard<std::mutex> guard(callback_mutex_);
-    set_parameter_callbacks_[name].push_back(callback);
+    if (callback_handler_)
+    {
+      const std::lock_guard<std::mutex> guard(callback_handler_->mutex_);
```
tylerjw, line 132:
```suggestion
      const std::lock_guard<std::mutex> guard{callback_handler_->mutex_};
```

moveit_ros/moveit_servo/src/servo_parameters.cpp
```diff
@@ -77,175 +83,363 @@ ServoParameters::setParametersCallback(std::vector<rclcpp::Parameter> parameters
...
+                                     ParameterValue(parameters.scene_collision_proximity_threshold),
+                                     ParameterDescriptorBuilder()
+                                         .type(PARAMETER_DOUBLE)
+                                         .description("Start decelerating when a scene collision is this far [m]"));
+}
+
+ServoParameters ServoParameters::get(const std::string& ns,
+                                     const rclcpp::node_interfaces::NodeParametersInterface::SharedPtr& node_parameters)
+{
+  auto parameters = ServoParameters();
```
tylerjw, line 253:
```suggestion
  auto parameters = ServoParameters{};
```

AndyZe:
>  I don't have a joystick handy

@stephanie-eng the tutorial has a way to test with a keyboard. (https://github.com/ros-planning/moveit2_tutorials/blob/main/doc/realtime_servo/realtime_servo_tutorial.rst)

ros2 launch moveit_servo servo_example.launch.py
ros2 service call /servo_node/start_servo std_srvs/srv/Trigger {}
ros2 run moveit2_tutorials servo_keyboard_input

stephanie-eng:
Thanks, Andy.
I've incorporated Tyler's suggestions and was able to run the demo without problem.

moveit_ros/moveit_servo/src/servo_parameters.cpp
```diff
@@ -77,175 +83,363 @@ ServoParameters::setParametersCallback(std::vector<rclcpp::Parameter> parameters
...
+      ns + ".use_gazebo", ParameterValue{ parameters.use_gazebo },
+      ParameterDescriptorBuilder{}
+          .type(PARAMETER_BOOL)
+          .description("Whether the robot is started in a Gazebo simulation environment"));
+  node_parameters->declare_parameter(
+      ns + ".status_topic", ParameterValue{ parameters.status_topic },
+      ParameterDescriptorBuilder{}.type(PARAMETER_STRING).description("Publish status to this topic"));
+  // Properties of incoming commands
+  node_parameters->declare_parameter(
+      ns + ".cartesian_command_in_topic", rclcpp::ParameterValue{ parameters.cartesian_command_in_topic },
```
tylerjw, line 107: You don't need the rclcpp namespace for ParameterValue here because of the using statement.

moveit_ros/moveit_servo/src/servo_parameters.cpp
```diff
@@ -77,175 +83,363 @@ ServoParameters::setParametersCallback(std::vector<rclcpp::Parameter> parameters
...
+      ParameterDescriptorBuilder{}.type(PARAMETER_STRING).description("Topic for incoming Cartesian twist commands"));
+  node_parameters->declare_parameter(
+      ns + ".joint_command_in_topic", rclcpp::ParameterValue{ parameters.joint_command_in_topic },
```
tylerjw, line 110: Here too

tylerjw (commented):
Only a couple of tiny things I found.  Otherwise this looks good.  I tested this locally and it doesn't break anything as far as I can tell.

stephanie-eng:
Just as an FYI: because this changes the function signature of makeServoParameters, a minor update to the tutorial is required for it to build. I've created a PR to make the tutorials build again for after this PR is merged.

tylerjw:
> Just as an FYI: because this changes the function signature of makeServoParameters, a minor update to the tutorial is required for it to build. I've created a PR to make the tutorials build again for after this PR is merged.

Thank you for doing that.  :)

AndyZe:
moveit_ros/moveit_servo/include/moveit_servo/parameter_descriptor_builder.hpp could still use a description at the top of the file. Something like:

```
/* Author: Stephanie Eng
   Description: blah blah blah
 */
```

AndyZe (approved):

PR #811, opened 2021-11-18
Hybrid Planning: Add review changes to feature branch and rebase

sjahr:
### Description

This PR reopens #718 with the intention to get the hybrid planning feature merged into main. 
I've rebased again on main (11/18/2021) and added the changes introduced by #763 (Thanks to @tylerjw and @AndyZe!). Furthermore, I've fixed some things to make the demo work again, although that is still work in progress right now. 
@AndyZe and @tylerjw I've sent you collaborator invitations to my moveit2 fork, so you can directly contribute to this PR to speed up the process.

### Checklist
- [X] **Required by CI**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)
- [ ] Adress review from #488 completely
- [ ] Fix segfault caused by invalid action goal handle in hybrid planning interface

[//]: # "You can expect a response from a maintainer within 7 days. If you haven't heard anything by then, feel free to ping the thread. Thank you!"

PR #813, opened 2021-11-18
Add hybrid planning

AndyZe:
This continues from #488. It addresses most of the comments there.

The big comment that's not addressed yet relates to the timer that is used in launching `hybrid_planning_manager/hybrid_planning_manager.cpp`. I made [an issue](https://github.com/ros-planning/moveit2/issues/814) for that. #811 and #763 tried to tackle that issue but they both crash. So I think it's something we can work on improving later.

I'll work on cleaning up the corresponding tutorial PR and adding some tests for this.

Suggestions on how to clean up the commit history are welcome. I'd be fine with squashing into five commits or so.


AndyZe:
CI won't pass until the next moveit_msgs release, I suppose.

AndyZe:
@sjahr I would appreciate it if you took a look at the most recent commit. The logic for the action result isn't working here because `reaction_result` always reflects a SUCCESS.

I'm not really sure what it's intended to do. These are the lines:

```
  // Add result callback to print the result
  local_goal_options.result_callback =
      [this](const rclcpp_action::ClientGoalHandle<moveit_msgs::action::LocalPlanner>::WrappedResult& action_result) {
        RCLCPP_ERROR(LOGGER, "local action callback");
        RCLCPP_ERROR_STREAM(LOGGER, (bool)(action_result.code == rclcpp_action::ResultCode::SUCCEEDED));
        // Reaction result from the latest event
        ReactionResult reaction_result =
            planner_logic_instance_->react(HybridPlanningEvent::LOCAL_PLANNING_ACTION_FINISHED);
        // TODO: This always thinks the action was successful!
        if (reaction_result.error_code.val != moveit_msgs::msg::MoveItErrorCodes::SUCCESS)
        {
          auto result = std::make_shared<moveit_msgs::action::HybridPlanner::Result>();
          result->error_code.val = reaction_result.error_code.val;
          result->error_message = reaction_result.error_message;
          hybrid_planning_goal_handle_->abort(result);
          RCLCPP_ERROR(LOGGER, "Hybrid Planning Manager failed to react to  '%s'", reaction_result.event.c_str());
        }
      };
```

If you run it, I think you'll easily be able to see the issue. `ros2 launch moveit_hybrid_planning hybrid_planning_demo.launch.py`

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/src/hybrid_planning_manager.cpp
```diff
@@ -184,14 +184,6 @@ bool HybridPlanningManager::sendGlobalPlannerAction()
           ReactionResult reaction_result =
               planner_logic_instance_->react(HybridPlanningEvent::GLOBAL_PLANNING_ACTION_FINISHED);
           auto result = std::make_shared<moveit_msgs::action::HybridPlanner::Result>();
-          if (reaction_result.error_code.val != moveit_msgs::msg::MoveItErrorCodes::SUCCESS)
```
AndyZe: This if() was never entered because the value of `reaction_result` is constant (see 2 lines up)

AndyZe:
@sjahr I believe that f8b0cbd fixed my comment above about `reaction_result` always reflecting SUCCESS. Take a look if you have time.

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/src/hybrid_planning_manager.cpp
sjahr: > @sjahr I believe that [f8b0cbd](https://github.com/ros-planning/moveit2/commit/f8b0cbd1bb541a2f11f8eb6f0cfeb17d8d19fe24) fixed my comment above about `reaction_result` always reflecting SUCCESS. Take a look if you have time.

The reaction result is not constant but depends on the planner logic used (because it defines the react() function):
```cpp
ReactionResult SinglePlanExecution::react(const HybridPlanningEvent& event)
{
  switch (event)
  {
    case HybridPlanningEvent::HYBRID_PLANNING_REQUEST_RECEIVED:
      if (!hybrid_planning_manager_->sendGlobalPlannerAction())  // Start global planning
      {
        hybrid_planning_manager_->sendHybridPlanningResponse(false);
      }
      break;
    case HybridPlanningEvent::GLOBAL_SOLUTION_AVAILABLE:
      std::call_once(LOCAL_PLANNER_STARTED, [this]() {            // ensure the local planner is not started twice
        if (!hybrid_planning_manager_->sendLocalPlannerAction())  // Start local planning
        {
          hybrid_planning_manager_->sendHybridPlanningResponse(false);
        }
      });
      break;
    case HybridPlanningEvent::LOCAL_PLANNING_ACTION_FINISHED:
      hybrid_planning_manager_->sendHybridPlanningResponse(true);
      break;
    default:
      break;
      // Do nothing
  }
  return ReactionResult(event, "", moveit_msgs::msg::MoveItErrorCodes::SUCCESS); // This never returns failure
}
```

The available planner logic plugins ([single_plan_execution](https://github.com/ros-planning/moveit2/pull/813/files#diff-0b0779bd232dff42badd15757460d5e7d4967d4eb6f42ac09d8076e80e970b85) or replan_invalidated_trajectory) do not return failure at any point but other logic plugins developed in the future might do that.

I think this fix is not correct because the logic should define how the architecture reacts to a failing global/local planner. Instead of aborting it might be desired to rerun the planner multiple times or adapt its constraints. Now it is not possible for the user-defined logic to handle this case because we abort hybrid planning before the react() function is called in case the planner fails. 
The previous implementation in the hybrid planning manager was correct but I guess the plugin logic code above needs a fix. I think the planner result needs to be saved in a variable of the hybrid planning manager and the evaluated in the plannger logic.

sjahr:
> @sjahr I believe that [f8b0cbd](https://github.com/ros-planning/moveit2/commit/f8b0cbd1bb541a2f11f8eb6f0cfeb17d8d19fe24) fixed my comment above about `reaction_result` always reflecting SUCCESS. Take a look if you have time.
@AndyZe I think this is not the correct way to fix it, sorry. I've answered in your review comment above.

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/src/hybrid_planning_manager.cpp
AndyZe: I'm not sure the `HybridPlanningEvent` enum contains enough information to ever communicate a failure. These are the only options:

```
  // Occurs when the hybrid planning manager receives a planning request
  HYBRID_PLANNING_REQUEST_RECEIVED,
  // Indicates that the global planning action finished (Does not indicate planning success! The action response
  // contains this information)
  GLOBAL_PLANNING_ACTION_FINISHED,
  // Indicates that the global planner found a solution (This solution is not necessarily the last or best solution)
  GLOBAL_SOLUTION_AVAILABLE,
  // Indicates that the local planning action finished (Does not indicate planning success! The action response contains this information)
  LOCAL_PLANNING_ACTION_FINISHED,
```

What would I put in case of a failure?

AndyZe: I think it's best if I wait for you to make a PR about this.

AndyZe: I think my confusion arose because I had put the "local planner stuck detection" functionality inside LocalPlannerComponent. That's not where it belongs -- it belongs in the local solver plugin, which is ForwardTrajectory. Having made that change, I don't have any issue communicating the failure.

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/include/moveit/hybrid_planning_manager/hybrid_planning_manager.h
```diff
@@ -0,0 +1,132 @@
...
+#include <moveit/hybrid_planning_manager/planner_logic_interface.h>
+
+#include <pluginlib/class_loader.hpp>
+
+namespace moveit::hybrid_planning
+{
+/**
+ * Class HybridPlanningManager - ROS 2 component node that implements the hybrid planning manager.
+ */
+class HybridPlanningManager : public rclcpp::Node
```
tylerjw, line 57: Can this be made similar to the other node objects and not inherit from rclcpp::Node and instead contain a node?

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/include/moveit/hybrid_planning_manager/hybrid_planning_manager.h
```diff
@@ -0,0 +1,132 @@
...
+  HybridPlanningManager(const rclcpp::NodeOptions& options);
+
+  /**
+   * Allows creation of a smart pointer that references to instances of this object
+   * @return shared pointer of the HybridPlanningManager instance that called the function
+   */
+  std::shared_ptr<HybridPlanningManager> shared_from_this()
+  {
+    return std::static_pointer_cast<HybridPlanningManager>(Node::shared_from_this());
+  }
```
tylerjw, line 70: If this doesn't inherit from Node (and instead contains a node) then we shouldn't need this.

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/include/moveit/hybrid_planning_manager/hybrid_planning_manager.h
```diff
@@ -0,0 +1,132 @@
...
+
+  /**
+   * Load and initialized planner logic plugin and ROS 2 action and topic interfaces
+   * @return Initialization successfull yes/no
+   */
+  bool initialize();
```
tylerjw, line 76: Can whatever is in this be moved into the constructor and this method eliminated?

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/include/moveit/hybrid_planning_manager/hybrid_planning_manager.h
```diff
@@ -0,0 +1,132 @@
...
+  std::unique_ptr<pluginlib::ClassLoader<PlannerLogicInterface>> planner_logic_plugin_loader_;
+
+  // Planner logic instance to implement reactive behavior
+  std::shared_ptr<PlannerLogicInterface> planner_logic_instance_;
+
+  // Timer to trigger events periodically
+  rclcpp::TimerBase::SharedPtr timer_;
+
+  // Flag that indicates whether the manager is initialized
+  bool initialized_;
```
tylerjw, line 114: If initialization can be done in the constructor this can probably be removed.

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/include/moveit/hybrid_planning_manager/planner_logic_interface.h
```diff
@@ -0,0 +1,132 @@
...
+#include <moveit_msgs/msg/move_it_error_codes.hpp>
+#include <moveit/hybrid_planning_manager/hybrid_planning_events.h>
+#include <moveit/hybrid_planning_manager/moveit_error_code_interface.h>
+
+namespace moveit::hybrid_planning
+{
+// Describes the outcome of a reaction to an event in the hybrid planning architecture
+struct ReactionResult
+{
+  ReactionResult(const HybridPlanningEvent& planning_event, const std::string& error_msg, const int& error_code)
```
tylerjw, line 51:
```suggestion
  ReactionResult(const HybridPlanningEvent& planning_event, const std::string& error_msg, int error_code)
```

Don't use const& for built-in types.

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/include/moveit/hybrid_planning_manager/planner_logic_interface.h
```diff
@@ -0,0 +1,132 @@
...
+        event = "Local planning action aborted";
+        break;
+      case HybridPlanningEvent::LOCAL_PLANNING_ACTION_CANCELED:
+        event = "Local planning action canceled";
+        break;
+      case HybridPlanningEvent::UNDEFINED:
+        event = "Undefined event";
+    }
+  };
+  ReactionResult(const std::string& event, const std::string& error_msg, const int& error_code)
```
tylerjw, line 84:
```suggestion
  ReactionResult(const std::string& event, const std::string& error_msg, int error_code)
```

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/src/hybrid_planning_manager.cpp
```diff
@@ -0,0 +1,335 @@
...
+    {
+      if (!this->initialize())
+      {
+        const std::string error = "Failed to initialize global planner";
+        timer_->cancel();
+        throw std::runtime_error(error);
+      }
+      initialized_ = true;
+    }
+  });
```
tylerjw, line 67: We shouldn't need this.  I will create a PR to remove this ( I thought I did this once before ).

tylerjw (commented):
I left some comments.  I'll create a PR to try to fix those things.

tylerjw:
I tried fixing this in a PR to andy's branch and failed to fix it easily.  I think it is important to acknowledge that one of the wonderful parts about software is that we can change it in the future and that nice software often takes many iterations to be really nice.  In this case I don't have the time to try to fix the issues I raised so I think we should just merge this as it is for now.  Sometime when one of has time to work on this we should consider rewriting how the initialization of plugins work and accept that the plugin interface will probably have to change as a result.

tylerjw:
I also expect all the "wait for the thing to happen then throw an exception or print a ros error and ignore when it hasn't yet" patterns will be flaky and need to be fixed, but I also don't have to try to fix that myself either.  I will not be surprised when the integration test in this goes flaky.

tylerjw:
I need to pick what I want to work on if I'm going to finish anything and it isn't fair for me to block other people's work because I don't have time to work on it.  Thank you @AndyZe for taking over the hybrid planner and pushing to get it merged.  It has been long enough and we should just merge this and try to learn from this experience and find ways to make so PRs don't live for so long.

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/include/moveit/hybrid_planning_manager/planner_logic_interface.h
tylerjw: @AndyZe can we accept these two small changes to the interface?  const& to a built-in type (like int) is a pessimization over passing by value.

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/src/hybrid_planning_manager.cpp
tylerjw: I'm closing this as future work.

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/include/moveit/hybrid_planning_manager/planner_logic_interface.h
AndyZe: definitely

tylerjw: you fixed the other one but this one seems to still be here

AndyZe: geez, that's embarassing

tylerjw (approved):
I re-read through all of this again.  I think I'm ok with it for now.  Eventually I want to figure out the initialization and lifetime of these plugins so there isn't a circular relationship.

henningkayser:
There are still some build warnings (check CI). They are not enforced yet, but it would be nice if they were fixed. Also, I noticed that the include folders are not installed locally when installing with clang, still looking into that.

sjahr:
> There are still some build warnings (check CI). They are not enforced yet, but it would be nice if they were fixed. Also, I noticed that the include folders are not installed locally. This is only apparent when depending on hybrid planning from another package. Here is an error log when compiling the `hybrid_planning_demo` from the [UR10e welding repo](https://github.com/PickNikRobotics/UR10e_welding_demo). Since this worked with an older version of hybrid planning, there must be some newer changes that broke this.
> 
> ```
> CMake Warning at /home/ubuntu/ws_ros2/install/moveit_hybrid_planning/share/moveit_hybrid_planning/cmake/ament_cmake_export_include_directories-extras.cmake:11 (message):
>   Package 'moveit_hybrid_planning' exports the include directory
>   '/home/ubuntu/ws_ros2/install/moveit_hybrid_planning/share/moveit_hybrid_planning/cmake/../../../global_planner/global_planner_component/include/'
>   which doesn't exist
> Call Stack (most recent call first):
>   /home/ubuntu/ws_ros2/install/moveit_hybrid_planning/share/moveit_hybrid_planning/cmake/moveit_hybrid_planningConfig.cmake:41 (include)
>   CMakeLists.txt:5 (find_package)
> 
> 
> CMake Warning at /home/ubuntu/ws_ros2/install/moveit_hybrid_planning/share/moveit_hybrid_planning/cmake/ament_cmake_export_include_directories-extras.cmake:11 (message):
>   Package 'moveit_hybrid_planning' exports the include directory
>   '/home/ubuntu/ws_ros2/install/moveit_hybrid_planning/share/moveit_hybrid_planning/cmake/../../../global_planner/global_planner_plugins/include/'
>   which doesn't exist
> Call Stack (most recent call first):
>   /home/ubuntu/ws_ros2/install/moveit_hybrid_planning/share/moveit_hybrid_planning/cmake/moveit_hybrid_planningConfig.cmake:41 (include)
>   CMakeLists.txt:5 (find_package)
> 
> 
> CMake Warning at /home/ubuntu/ws_ros2/install/moveit_hybrid_planning/share/moveit_hybrid_planning/cmake/ament_cmake_export_include_directories-extras.cmake:11 (message):
>   Package 'moveit_hybrid_planning' exports the include directory
>   '/home/ubuntu/ws_ros2/install/moveit_hybrid_planning/share/moveit_hybrid_planning/cmake/../../../hybrid_planning_manager/hybrid_planning_manager_component/include/'
>   which doesn't exist
> Call Stack (most recent call first):
>   /home/ubuntu/ws_ros2/install/moveit_hybrid_planning/share/moveit_hybrid_planning/cmake/moveit_hybrid_planningConfig.cmake:41 (include)
>   CMakeLists.txt:5 (find_package)
> 
> 
> CMake Warning at /home/ubuntu/ws_ros2/install/moveit_hybrid_planning/share/moveit_hybrid_planning/cmake/ament_cmake_export_include_directories-extras.cmake:11 (message):
>   Package 'moveit_hybrid_planning' exports the include directory
>   '/home/ubuntu/ws_ros2/install/moveit_hybrid_planning/share/moveit_hybrid_planning/cmake/../../../hybrid_planning_manager/planner_logic_plugins/include/'
>   which doesn't exist
> Call Stack (most recent call first):
>   /home/ubuntu/ws_ros2/install/moveit_hybrid_planning/share/moveit_hybrid_planning/cmake/moveit_hybrid_planningConfig.cmake:41 (include)
>   CMakeLists.txt:5 (find_package)
> 
> 
> CMake Warning at /home/ubuntu/ws_ros2/install/moveit_hybrid_planning/share/moveit_hybrid_planning/cmake/ament_cmake_export_include_directories-extras.cmake:11 (message):
>   Package 'moveit_hybrid_planning' exports the include directory
>   '/home/ubuntu/ws_ros2/install/moveit_hybrid_planning/share/moveit_hybrid_planning/cmake/../../../local_planner/local_planner_component/include/'
>   which doesn't exist
> Call Stack (most recent call first):
>   /home/ubuntu/ws_ros2/install/moveit_hybrid_planning/share/moveit_hybrid_planning/cmake/moveit_hybrid_planningConfig.cmake:41 (include)
>   CMakeLists.txt:5 (find_package)
> 
> 
> CMake Warning at /home/ubuntu/ws_ros2/install/moveit_hybrid_planning/share/moveit_hybrid_planning/cmake/ament_cmake_export_include_directories-extras.cmake:11 (message):
>   Package 'moveit_hybrid_planning' exports the include directory
>   '/home/ubuntu/ws_ros2/install/moveit_hybrid_planning/share/moveit_hybrid_planning/cmake/../../../local_planner/trajectory_operator_plugins/include/'
>   which doesn't exist
> Call Stack (most recent call first):
>   /home/ubuntu/ws_ros2/install/moveit_hybrid_planning/share/moveit_hybrid_planning/cmake/moveit_hybrid_planningConfig.cmake:41 (include)
>   CMakeLists.txt:5 (find_package)
> 
> 
> CMake Warning at /home/ubuntu/ws_ros2/install/moveit_hybrid_planning/share/moveit_hybrid_planning/cmake/ament_cmake_export_include_directories-extras.cmake:11 (message):
>   Package 'moveit_hybrid_planning' exports the include directory
>   '/home/ubuntu/ws_ros2/install/moveit_hybrid_planning/share/moveit_hybrid_planning/cmake/../../../local_planner/local_constraint_solver_plugins/include/'
>   which doesn't exist
> Call Stack (most recent call first):
>   /home/ubuntu/ws_ros2/install/moveit_hybrid_planning/share/moveit_hybrid_planning/cmake/moveit_hybrid_planningConfig.cmake:41 (include)
>   CMakeLists.txt:5 (find_package)
> 
> 
> CMake Error at CMakeLists.txt:37 (add_library):
>   Target "servo_local_solver" links to target
>   "moveit_hybrid_planning::hybrid_planning_demo_node" but the target was not
>   found.  Perhaps a find_package() call is missing for an IMPORTED target, or
>   an ALIAS target is missing?
> 
> 
> CMake Error at CMakeLists.txt:32 (add_library):
>   Target "global_mtc_planner_component" links to target
>   "moveit_hybrid_planning::hybrid_planning_demo_node" but the target was not
>   found.  Perhaps a find_package() call is missing for an IMPORTED target, or
>   an ALIAS target is missing?
> 
> 
> CMake Error at CMakeLists.txt:29 (add_executable):
>   Target "hybrid_planning_test_node" links to target
>   "moveit_hybrid_planning::hybrid_planning_demo_node" but the target was not
>   found.  Perhaps a find_package() call is missing for an IMPORTED target, or
>   an ALIAS target is missing?
> 
> 
> CMake Error at CMakeLists.txt:41 (add_library):
>   Target "servo_sampler" links to target
>   "moveit_hybrid_planning::hybrid_planning_demo_node" but the target was not
>   found.  Perhaps a find_package() call is missing for an IMPORTED target, or
>   an ALIAS target is missing?
> 
> 
> CMake Generate step failed.  Build files cannot be regenerated correctly.
> ---
> Failed   <<< hybrid_planning_demo [14.4s, exited with code 1]
> ```

I had the same issue and I've fixed it with [this PR](https://github.com/AndyZe/moveit2/pull/6)

sjahr:
> I re-read through all of this again. I think I'm ok with it for now. Eventually I want to figure out the initialization and lifetime of these plugins so there isn't a circular relationship.

Thanks! I think the key to solving this is to have a better solution for how hybrid_planning_manager and its logic plugin can share the same ros2 communication interfaces. When you have the time to work on it, I'd be happy to discuss the issue with you as I am struggling to find a good solution too

AndyZe:
I am now getting a parameter loading error. This started to happen after I rebased. I think it is related to the loading of `global_planner.yaml`:

```
[component_container_mt-1] [FATAL] [1638568924.298387221] [moveit_ros.trajectory_execution_manager]: Exception while loading controller manager '<undefined>': According to the loaded plugin descriptions the class <undefined> with base class type moveit_controller_manager::MoveItControllerManager does not exist. Declared types are  moveit_ros_control_interface/MoveItControllerManager moveit_ros_control_interface/MoveItMultiControllerManager moveit_simple_controller_manager/MoveItSimpleControllerManager
[component_container_mt-1] [ERROR] [1638568924.298412408] [moveit_ros.trajectory_execution_manager]: Failed to reload controllers: `controller_manager_` does not exist.
```

AndyZe:
Re. parameter loading:

`moveit_controller_manager` wasn't loaded because of a typo.

More importantly, all of these ompl parameters aren't declared or retrieved at all. No idea how this ever worked. I found this by comparing against the moveit_cpp demo. Maybe we can re-use the parameter loading code from moveit_cpp.

```
    ompl:
      panda_arm:
        planner_configs:
        - SBLkConfigDefault
        - ESTkConfigDefault
        - ...
```

sjahr:
> Re. parameter loading:
> 
> `moveit_controller_manager` wasn't loaded because of a typo.
> 
> More importantly, all of these ompl parameters aren't declared or retrieved at all. No idea how this ever worked. I found this by comparing against the moveit_cpp demo. Maybe we can re-use the parameter loading code from moveit_cpp.
> 
> ```
>     ompl:
>       panda_arm:
>         planner_configs:
>         - SBLkConfigDefault
>         - ESTkConfigDefault
>         - ...
> ```

There is a [default_planner_config within the ompl interface class](https://github.com/ros-planning/moveit2/blob/42afeea561949b1af197655c1cbb63b851e7dd8d/moveit_planners/ompl/ompl_interface/src/ompl_interface.cpp#L211) which is used if nothing else is declared. For the simple demos, this was sufficient but you are right, that we need to add the option to pass ompl parameters. The MoveIt cpp demo just uses   `node_options.automatically_declare_parameters_from_overrides(true);` which simply declares every parameter given to the node by yaml file but this is considered bad practice (although it could be an interim solution as the parameter handling best practice pattern is a MoveIt2 "global" issue). I think a good place to declare the ompl parameters would be in the ompl interface class because the MoveItPlanningPipeline global planner plugin should be agnostic to the MoveIt planner plugin used (i.e. the Fraunhofer demo uses the PILZ planner I think). @JafarAbdi and @tylerjw put some work into parameter handling, what do you think?

AndyZe:
@sjahr  I appreciate the thoughts!  `automatically_declare_parameters_from_overrides` is already set to true in **hybrid_planning_demo_node.cpp**. I think that's the same node that gets passed to all the classes. Not sure why it does not take effect.

AndyZe:
@sjahr I found the magic line that makes it all work again  :+1:

`node->declare_parameter<std::string>("ompl.planning_plugin", "ompl_interface/OMPLPlanner");`

I don't know why it was working previously without this.

Will make an issue about parameter loading. I think your idea to load the parameters from ompl is a good one. It's not just hybrid planning but many components of MoveIt that are having this issue.

tylerjw:
I fixed the merge conflict but it might be better to rebase this commit on main so my merge commit goes away (as we are going to merge this and not squash it).

PR #825, opened 2021-11-24
Draft: Servo pipeline

tylerjw:
# Servo Input Pipeline Design

## C++17 features

To understand this pattern let me introduce you to two really nice features from C++17, `variant` and `Visitors` that are used by the `visit` function.  To use these features you just need to include this:

```c++
#include <variant>
```

### std::variant

This design starts with using some new features from C++17.  The first is the variant.  `input_command.hpp` defines this:

```c++
using InputCommand = std::variant<geometry_msgs::msg::TwistStamped, control_msgs::msg::JointJog>;
```

When you send a command to servo it is one of these two types.  It then does a handful of things that are not specific to the type of message received.  Specifically, the timestamp it was received is recorded, it is stored locally, then the loop is started at a rate with whichever message was newest.  Lastly, the timestamp of the message is tested against the stale command timeout to see if it is stale and servo should halt instead of continuing to move the robot arm.

At many places, the existing code has to test which message was last received.  This is simplified by using a variant.  A variant is a type-safe version of the union from C.  It is a type that can store one of the two types and has interfaces for executing code based on the types.  Here is a reference for a really nice guide with small examples showing what variant is: https://www.cppstories.com/2018/06/variant/

### Visitors and the std::visit function

The second C++17 feature needed for this pattern is a Visitor.  A visitor is an object that you can call visit on with a variant.  To understand what that means here is a simple example:

```c++
class PrintVisitor
{
  void operator()(int value) { cout << "int value: " << value; }
  void operator()(double value) { cout << "double value: " << value; }
};

int main()
{
  using IntOrDouble = std::variant<int, double>;
  auto a = IntOrDouble{ 2.5f };
  auto b = IntOrDouble{ 7 };

  auto print_visitor = PrintVisitor{};
  std::visit(print_visitor, a);
  std::visit(print_visitor, b);
  return 0;
} 
```

There are a couple of important C++ pieces of this.  

The first is the callable class PrintVisitor.  `operator()(variables...)` is an operator that says you can treat objects of this type like a function that takes `(variables...)`.  This is an explicit way of defining a callable class.  Lambdas are anonymous versions of this sort of class as they are objects that can store state and have a callable interface.

The second thing is the call to `visit`.  Visit just says call the callable function for the type of the variant passed into the second argument.  The compiler will check that the first argument has all the interfaces for all the possible types of the variant.  This whole concept came from functional programming and is enabled by the rich type system in C++.

These tools enable us to write code that says "do **this operation** on **this thing**" and allow the **thing** to be of different types and have different implementations for the type.

## Software design motivations and decisions

Here are some of the decisions and motivations behind these changes.

### How do these features fit in servo?

The input side of servo can take one of two types.  We want to respond to the last message we received.  When we had two different variables for the two types there needed to be a bunch of code to test which one was newest.  With a variant, there can be just one variable (the variant) and you just update and use that variable.

The second part of this is the visitor pattern.  Before this PR there is a logic in ServoCalcs that has to do with the simple logic of running the calculations at a set rate and making sure we halt when we have stale commands.  Because this logic was tightly coupled with the rest of ServoCalcs we had a bunch of code that can't be tested on its own.  By breaking out the various operations taken on the input values into separate objects that just call the next step with visit they can be tested easily separated from each other.

### SRP (Single responsibility principle)

One important rule for software design is that by breaking a problem into small modular pieces you can test them separately from each other and they are easier to write and understand.  There was a talk at cppcon this year that really inspired me to try to take this much more seriously.  One real advantage to this approach is that it produces code that is easier to understand and modify because each piece is much smaller.  The only real downside is that there are many more pieces now and you have more total code to read.

https://digital-medium-co-uk.zoom.us/rec/play/JMkXvgiWQKX5EJNuaHL-1Rh0F8FiFUwKYomrJF_wEgAtRaDVTHc_p_j3ZkbZ7dsGsLIYU5v4lP0GEMLm.8PJcMpJSgWDT8NJy?continueMode=true&_x_zm_rtaid=HjhEGiY3SZCHZAI7zLBnzQ.1637779957973.3b927e82802ef1b90907a73d94e9aed7&_x_zm_rhtaid=567

### Testing and why I didn't use dependency injection

The tests I wrote for the various pieces don't have dependency injection and do use some ROS features directly.  This greatly simplified the tests and reduced the boilerplate I had to write.  The large benefits to dependency injection comes from cases where the ROS features are complex and inherently unreliable.  By making the compromise that some of the tests take a bit longer to execute I was able to greatly simplify how much boilerplate code I had to write to add tests.  Everything has tradeoffs and dependency injection gives you much more performant and potentially much more reliable testing with the drawback of a bunch of boilerplate code and much more complex testing code.

## Design

Here I will lay out the various pieces, what they do, and my understanding of how they should work

### `input_command.hpp`

This header defines the variant type `InputCommand` that is the type used by the input of servo.
Secondly, there is the interface class `InputVisitor` that each visitor class will inherit from and implement the interface for calling `visit` on.  By using an interface class each class can have a pointer to the next step without caring what the type is because this interface guarantees it implements the functions needed to call visit on it.

### `input_subscriber.hpp`

This object has one purpose.  It contains the two subscribers to the input topics and in the callbacks uses visit to call the next step with an InputCommand variant.

### `input_stale_command.hpp`

This header contains two classes: `TimestampNow` and `StaleCommandHalt`.  `TimestampNow` is a peice of fixing the bug where stale commands don't stop servo.  This visitor just sets the header timestamp in the message to the output of `node->now()` and calls the next step.  The second one, `StaleCommandHalt` checks for the stale command timeout.  If the message is stale it calls the halt function it was passed and if not it calls the next step with the message.

Here is one place where the implementation differs from servo as it is now.  Right now servo expects the message timestamp to be set by the user of servo.  The downside to this is if a user forgets to do this or is replaying from a bagfile.  In both of those cases, it is very easy for the user to send a timestamp in their message that is not current.  This change makes so what servo cares about when it comes to stale messages is the time since it received the message instead of the time since the timestamp in the header of the message that was sent to servo.

### `input_resampler.hpp`

This is probably the most complex of the Visitors.  This one receives a message and stores it internally.  It has a timer that calls visit on the next step with the newest message it has received.  

Here is another important change from what was before.  Before as soon as you called to start the loop would start executing (even if we still haven't received a message).  This time where the loop is executing before the user sends a command is used to initialize the smoothing algorithm with the current state of the robot.

If previously a message was available on the first run of the loop there could be wild jumps because the smoothing filter was not yet initialized.  This was really unlikely to happen before but became obvious when I started testing this because in this case no calls are made to ServoCalcs until the user sends the first input.  To fix this I have the constructor of ServoCalcs waiting for a complete state of the robot and initializing the smoother.

Another edge case this helped me find is if a user pauses servo, moves the robot some other way, and then resumes servo.  This can also cause wild jumps if the smoother was not updated to the new position before executing the first command since the resume (this is because the previous state of the robot in servo was where it was before it was paused).  To fix this I am now also updating the smoother with the current state at the top of the function for halting on stale commands when paused (presumably the command will go stale when the user pauses).

### `input_pipeline.hpp`

This is what puts together all the various steps based on the configuration.  In the simplest case of `low_latency_mode` we only need an InputSubscriber that calls ServoCalcs directly.  This may seem like a bad idea because there is no handling of stale commands in this mode.  In my local testing, I've not been able to get the robot to keep moving if I stop sending commands.  I presume this is because there is a 1:1 relationship between input commands and output trajectories.  Once the trajectory is complete there is no new trajectory because there is no new input command and the robot stops.

In the more complex normal mode, the InputPipeline is a ResamplingInputPipeline that contains all the steps listed above connected together and eventually connected to ServoCalcs.

### `servo_calcs`

Here is where we were able to remove some code and other code had to just be moved around.

The first thing to note is that much of what was in the `start` function is now at the end of the constructor.  One part of this I'm not sure I like the interface to is the last two lines:

```c++
updateState();
resetLowPassFilters(original_joint_state_);
```

It is important to know that `updateState` calls a function that initializes `original_joint_state_` and that `resetLowPassFilters` is needed to initialize the values within the smoothing filters.

Next, I moved the code to publish the status and reset its value to NO_WARNING into a function because I now need to call it from several places.

After that is the new `updateState` function that updates the joint positions, current robot state, and the two tf frames we use for math.

The last of the helper functions I created is `publishJointTrajectory`.  This function prepares and publishes the output command.

One thing I don't like about these functions is they do much more than their name would imply.  For example `publishJointTrajectory` sets some values, selects between two different types of outputs, and then eventually publishes.  I made the choice to not try to refactor much of this as I wanted to limit the number of changes and felt that this was an OK compromise for now.

Lastly there are the two `operator()` functions and the `halt()` function.  These have the logic that was in the main loop before but is specific to the type of input they receive and in the case of halt, just create and publish the messages to halt the robot.  There is some copy-paste between these and eventually, I'd like to greatly reduce that but for now, that was one more thing I resisted trying to refactor more.  A bit upside to these three separate functions that do what the one main loop did is that there is much less nesting because the type we receive and the halt function stop us from having to check a bunch of bools and do a bunch of compares.

## Testing

A vast majority of the LOC in this change is test code.  These tests use various Visitor object mocks I created to capture the output of the various steps in `test/input_visitors.hpp`.  Each test follows a very specific pattern of `GIVEN / WHEN / THEN` from TDD (test-driven development) and each test is very simple and fast.  This sort of testing where there are many tiny tests means that if something doesn't work it is really obvious what is not working.  With large tests that test many things when they fail, they become hard to debug.

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/include/moveit_servo/detail/input_check_valid.hpp
+++ b/moveit_ros/moveit_servo/include/moveit_servo/detail/input_check_valid.hpp
@@ -59,10 +59,11 @@ class InputCheckValid : public InputVisitor
 {
   rclcpp::Node::SharedPtr node_ = nullptr;
   std::string command_in_type_ = "unitless";
-  InputVisitor* next_ = nullptr;
+  std::shared_ptr<InputVisitor> next_ = nullptr;
 
 public:
-  InputCheckValid(rclcpp::Node::SharedPtr node, std::string_view command_in_type, InputVisitor* next)
+  InputCheckValid(const rclcpp::Node::SharedPtr& node, std::string_view command_in_type,
+                  const std::shared_ptr<InputVisitor>& next)
     : node_{ node }, command_in_type_{ command_in_type }, next_{ next }
   {
     assert(node_ != nullptr);
--- a/moveit_ros/moveit_servo/include/moveit_servo/detail/input_check_valid.hpp
+++ b/moveit_ros/moveit_servo/include/moveit_servo/detail/input_check_valid.hpp
@@ -1,0 +1,78 @@
+/*********************************************************************
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2021, PickNik Inc.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of PickNik Inc. nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/*      Title     : input_check_valid.hpp
+ *      Project   : moveit_servo
+ *      Created   : 11/06/2021
+ *      Author    : Tyler Weaver
+ */
+
+#pragma once
+
+#include <functional>
+#include <variant>
+#include <memory>
+#include <control_msgs/msg/joint_jog.hpp>
+#include <geometry_msgs/msg/twist_stamped.hpp>
+#include <moveit_servo/detail/input_command.hpp>
+#include <rclcpp/rclcpp.hpp>
+
+namespace moveit_servo
+{
+namespace detail
+{
+/**
+ * @brief      Visitor that checks if the input is valid, only calling next if that is true
+ */
+class InputCheckValid : public InputVisitor
+{
+  rclcpp::Node::SharedPtr node_ = nullptr;
+  std::string command_in_type_ = "unitless";
+  InputVisitor* next_ = nullptr;
+
+public:
+  InputCheckValid(rclcpp::Node::SharedPtr node, std::string_view command_in_type, InputVisitor* next)
+    : node_{ node }, command_in_type_{ command_in_type }, next_{ next }
+  {
+    assert(node_ != nullptr);
+    assert(next_ != nullptr);
+  }
+  ~InputCheckValid() override = default;
+
+  void operator()(geometry_msgs::msg::TwistStamped command) override;
+  void operator()(control_msgs::msg::JointJog command) override;
+};
+
+}  // namespace detail
+}  // namespace moveit_servo
--- a/moveit_ros/moveit_servo/include/moveit_servo/detail/input_check_valid.hpp
+++ b/moveit_ros/moveit_servo/include/moveit_servo/detail/input_check_valid.hpp
@@ -70,8 +70,8 @@ public:
   }
   ~InputCheckValid() override = default;
 
-  void operator()(geometry_msgs::msg::TwistStamped command) override;
-  void operator()(control_msgs::msg::JointJog command) override;
+  void operator()(const geometry_msgs::msg::TwistStamped& command) override;
+  void operator()(const control_msgs::msg::JointJog& command) override;
 };
 
 }  // namespace detail
--- a/moveit_ros/moveit_servo/src/servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/src/servo_calcs.cpp
@@ -347,7 +347,7 @@ bool ServoCalcs::pausedEarlyExit()
   return false;
 }
 
-void ServoCalcs::operator()(geometry_msgs::msg::TwistStamped input_command)
+void ServoCalcs::operator()(const geometry_msgs::msg::TwistStamped& input_command)
 {
   const std::lock_guard<std::mutex> lock(main_loop_mutex_);
   updateState();
@@ -389,7 +389,7 @@ void ServoCalcs::operator()(geometry_msgs::msg::TwistStamped input_command)
   }
 }
 
-void ServoCalcs::operator()(control_msgs::msg::JointJog input_command)
+void ServoCalcs::operator()(const control_msgs::msg::JointJog& input_command)
 {
   const std::lock_guard<std::mutex> lock(main_loop_mutex_);
   updateState();
@@ -465,7 +465,7 @@ rcl_interfaces::msg::SetParametersResult ServoCalcs::robotLinkCommandFrameCallba
 };
 
 // Perform the servoing calculations
-bool ServoCalcs::cartesianServoCalcs(geometry_msgs::msg::TwistStamped& cmd,
+bool ServoCalcs::cartesianServoCalcs(geometry_msgs::msg::TwistStamped cmd,
                                      trajectory_msgs::msg::JointTrajectory& joint_trajectory)
 {
   // Set uncontrolled dimensions to 0 in command frame
--- a/moveit_ros/moveit_servo/src/servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/src/servo_calcs.cpp
@@ -58,6 +58,8 @@ namespace moveit_servo
 {
 namespace
 {
+constexpr double ROBOT_STATE_WAIT_TIME = 10.0;  // seconds
+
 // Helper function for detecting zeroed message
 bool isNonZero(const geometry_msgs::msg::TwistStamped& msg)
 {
@@ -124,13 +126,6 @@ ServoCalcs::ServoCalcs(rclcpp::Node::SharedPtr node,
   // Subscribe to command topics
   using std::placeholders::_1;
   using std::placeholders::_2;
-  twist_stamped_sub_ =
-      node_->create_subscription<geometry_msgs::msg::TwistStamped>(parameters_->cartesian_command_in_topic,
-                                                                   rclcpp::SystemDefaultsQoS(),
-                                                                   std::bind(&ServoCalcs::twistStampedCB, this, _1));
-
-  joint_cmd_sub_ = node_->create_subscription<control_msgs::msg::JointJog>(
-      parameters_->joint_command_in_topic, rclcpp::SystemDefaultsQoS(), std::bind(&ServoCalcs::jointCmdCB, this, _1));
 
   // ROS Server for allowing drift in some dimensions
   drift_dimensions_server_ = node_->create_service<moveit_msgs::srv::ChangeDriftDimensions>(
@@ -224,17 +219,6 @@ ServoCalcs::ServoCalcs(rclcpp::Node::SharedPtr node,
   empty_matrix.setZero();
   tf_moveit_to_ee_frame_ = empty_matrix;
   tf_moveit_to_robot_cmd_frame_ = empty_matrix;
-}
-
-ServoCalcs::~ServoCalcs()
-{
-  stop();
-}
-
-void ServoCalcs::start()
-{
-  // Stop the thread if we are currently running
-  stop();
 
   // Set up the "last" published message, in case we need to send it first
   auto initial_joint_trajectory = std::make_unique<trajectory_msgs::msg::JointTrajectory>();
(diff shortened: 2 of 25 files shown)


AndyZe:
@tylerjw to be honest these PR's that massively increase lines of code are starting to bug me. Agree that it would be nice to test that functionality but not at the cost of much more complexity.

+900 lines is not worth it, in my opinion.

tylerjw:
I also don't like adding more lines.  Much of it is boilerplate because when you break up functionality to test things you have a bunch more of that.

If it makes you feel better 680 lines of this are just tests.  It is only really about 300 lines of actually new code.

tylerjw:
@AndyZe I updated the description of this with what I hope is a nice description of what this change is.  Please let me know if there are anythings you find confusing or think I missed or could have done better in some other way.

tylerjw:
I just pushed two changes to this.  

The first fixes the integration tests by changing how the test for if servo is started works.  It adds a function for waiting for a condition to be true and waits for the TwistStamped subscriber to exist.  Before we waited for a bunch of status messages to be published.  Because ServoCalcs doesn't publish status messages continuously until it is started now (only one on startup) this is not a good method of waiting for startup.  The new method of waiting for the subscriber is better because it waits until the last ros interface to setup is configured and therefore should be much more robust.

The second change is thanks to @facontidavide.  He graciously looked through this design and pointed out areas where I had copy-pasted code and suggested a way to refactor these sectoins.  This reduced the amount of code and is a really nice change.

moveit_ros/moveit_servo/include/moveit_servo/input_pipeline.hpp
```diff
@@ -0,0 +1,117 @@
...
+                                std::shared_ptr<const moveit_servo::ServoParameters> parameters,
+                                const std::function<void()>& halt, InputVisitor* next)
+{
+  assert(node != nullptr);
+  assert(parameters != nullptr);
+  assert(next != nullptr);
+
+  if (parameters->low_latency_mode)
+  {
+    return ReactiveInputPipeline{ node, parameters, next };
```
henningkayser, line 109: How is `halt` implemented for this pipeline?

moveit_ros/moveit_servo/include/moveit_servo/input_resampler.hpp
```diff
@@ -0,0 +1,115 @@
...
+#include <geometry_msgs/msg/twist_stamped.hpp>
+#include <moveit_servo/input_command.hpp>
+#include <rclcpp/rclcpp.hpp>
+
+namespace moveit_servo
+{
+/**
+ * @brief      Visitor that stores the the input command and then calls a visitor with a timer at a specified rate
+ */
+class InputResampler : public InputVisitor
```
henningkayser, line 56: I somewhat find the name of `InputVisitor` confusing, even after understanding the visitor concept.
Isn't the core idea of this visitor base class that it's a processing step in the pipeline which then forwards the output to the next step? It would be nice if the class names would somehow reference more that they are part of a pipeline.

moveit_ros/moveit_servo/include/moveit_servo/input_resampler.hpp
```diff
@@ -0,0 +1,115 @@
...
+
+  // Create a copy of the message so we don't have to lock durring the call to visit
+  InputCommand getCommandCopy()
+  {
+    const std::lock_guard<std::mutex> lock(mutex_);
+    return command_;
+  }
+
+public:
+  InputResampler(rclcpp::Node::SharedPtr node, const rclcpp::Duration& period, InputVisitor* next)
```
henningkayser, line 73: Is it really necessary that the inputs directly call the next visitor? This makes the architecture and pipeline initialization harder to understand than necessary, imo. Since this is based on functional programming a lot already, how about putting all of the pipeline visitors in a `std::vector` and use a fold expression for processing the inputs.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -261,80 +245,18 @@ void ServoCalcs::start()
...
-  {
-    thread_.join();
-  }
+  updateState();
+  resetLowPassFilters(original_joint_state_);
 }
 
-void ServoCalcs::mainCalcLoop()
-{
-  rclcpp::Rate rate(1.0 / parameters_->publish_period);
```
henningkayser, line 298: I must have missed this, how does servo keep the rate now?

moveit_ros/moveit_servo/include/moveit_servo/input_stale_command.hpp
```diff
@@ -0,0 +1,133 @@
...
+#include <geometry_msgs/msg/twist_stamped.hpp>
+#include <moveit_servo/input_command.hpp>
+#include <rclcpp/rclcpp.hpp>
+
+namespace moveit_servo
+{
+/**
+ * @brief      Visitor that sets the header.stamp to now and then calls the next visitor
+ */
+class TimestampNow : public InputVisitor
```
henningkayser, line 56: Is it really worth having an extra visitor for this? Looks like code bloat to me

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -377,44 +294,76 @@ void ServoCalcs::calculateSingleIteration()
   // Calculate this transform to ensure it is available via C++ API
   tf_moveit_to_ee_frame_ = current_state_->getGlobalLinkTransform(parameters_->planning_frame).inverse() *
                            current_state_->getGlobalLinkTransform(parameters_->ee_frame_name);
+}
 
-  have_nonzero_command_ = have_nonzero_twist_stamped_ || have_nonzero_joint_command_;
+void ServoCalcs::publishJointTrajectory(std::unique_ptr<trajectory_msgs::msg::JointTrajectory> joint_trajectory)
```
henningkayser, line 299: I guess an OutputPipeline comes next? ;)

henningkayser (commented):
I really like your idea here, having a clear pipeline architecture based on `std::variant` and `std::visit` is a huge improvement. Your description is awesome! I somewhat agree with Andy that this adds a little too much bloat for what it does, but I think I left some comments about it. Overall, I find that the pipeline structure is harder to understand than it should be. If the `operator()` functions would return their results, the internal calls to `next` could be replaced with an external fold expression. That would also allow using simple lambdas for trivial InputVisitors.

tylerjw:
The name "input" in this case is because these operations are done on the input messages before any math.

I'll draw a diagram of this and hopefully that'll clear up why I think visiting the next class is a nice pattern.  A big part of why is the InputResampler which is now what controls the rate in the resampling mode.

A visitor class is just something that can be visited (called with visit and the input variant).

tylerjw:
Servo InputPipeline

@henningkayser as illustrated by this diagram the various classes only share the interface of being able to be called (with the exception of the subscriber source).  How they handle their output can be broken into these different categories:

* source: external signal (message callback or timer) that triggers calling next
* sync: consumes a message but doesn't produce it on the same thread
* operation: the normal case where it is called, does something, and then calls the next on the same thread
* branch: can either call the next one or call some other function (error handling for halt)

To be clearer about the naming and what comes next.  For now, I'm using the term InputPipeline to signify that these are objects that perform operations on the input command.  The plan was for the input pipeline to end with a joint trajectory message and then there would be a servo pipeline that operates on the joint trajectory message eventually ending in the publisher sync step.  Writing this I admit I haven't thought much about the naming of objects in the whole scheme of servo and that InputPipeline was named because it does operations on the input messages before ServoCalcs.

Another important point about the motivation for this refactor.  I want to move the configuration of control flow into the construction phase of servo.  The reason for this is some of the control flow is defined by parameters.  What moving it into the construction will eventually allow is for the setup of the pipeline to be done in a configuration step that can be re-done if someone changes a parameter that would alter the control flow.

moveit_ros/moveit_servo/src/servo_calcs.cpp
tylerjw: Yes, this is a boundary (preparing and publishing the joint trajectory) I can hopefully break into a set of operations that are connected together.

moveit_ros/moveit_servo/include/moveit_servo/input_stale_command.hpp
tylerjw: I think it'll make more sense why I did this after you see the diagram I posted.

moveit_ros/moveit_servo/src/servo_calcs.cpp
tylerjw: InputResampler.  The diagram I posted should hopefully help clear this up.

moveit_ros/moveit_servo/include/moveit_servo/input_pipeline.hpp
tylerjw: It is not handled.  In my experimentation at least if you tie the output in a 1:1 relationship with input servo will not continue to move the robot.  This is because it sends small steps to the robot and once the robot achieves that step it doesn't continue.  This even seems to be true if you send a twist with a velocity.

moveit_ros/moveit_servo/include/moveit_servo/input_resampler.hpp
tylerjw: See the diagram for why I have this InputVisitor.  The shortest explanation is that most of these steps have some sort of logic around how they call next and it isn't a simple pipeline of things you execute one after the other.

tylerjw:
@henningkayser I renamed InputPipeline to just ServoPipeline and put the construction of it in a constructor (instead of a builder function).  Long term I hope to roll this just into the Servo object itself (removing ServoPipeline and just copying its constructor into Servo).  The reason I'd like to keep it separate for now is it makes it really clear what code has been refactored into this pipeline structure and what has not yet.  Doing it this way means that as I implement this I can test the ServoPipeline object before all of Servo is testable.

tylerjw:
### Guide for Reviewers

Here is the order I would suggest for reviewing this PR.  Each section has tests and they are worth looking at but I would focus on these first.  To hopefully put your mind at ease about how much code this is here is the breakdown of a few sections of this PR that should take almost no time to review:

* test code: 861 lines
* license agreement and title comment at the top of new files: 320 lines
* additions to cmake for new tests and files: 22 lines

1. src/detail/servo_pipeline.cpp (and include/moveit_servo/detail/servo_pipeline.hpp)
  a. Here you will find the grand scheme for how these steps in the pipeline are created.  For now, just realize that the next pointer passed in is normally ServoCalcs.
2. src/detail/input_check_valid.cpp ( and include/moveit_servo/detail/input_check_valid.hpp)
  a. This code is just moved out of ServoCalcs so it can be a separate testable step.  The important part here is that when it succeeds (input is valid, it continues onto the next step).
3. include/moveit_servo/detail/input_command.hpp
  a. This is where the InputCommand and InputVisitor types are defined.
4. include/moveit_servo/detail/input_subscriber.hpp
  a. This is a simple class that contains the two ROS subscribers for capturing the input and calling the next step.
5. include/moveit_servo/detail/input_resampler.hpp
  a. This class just stores the input when it is called and calls next on a timer.  This is how we control the output rate now.
6. include/moveit_servo/detail/input_stale_command.hpp
  a. This contains the two steps for setting the timestamp (done before the resampler) and checking for the stale timeout (done after the resampler).  Note that it doesn't care how halt is implemented, it just calls the function for halting the robot.
7. src/servo_calcs.cpp (and include/moveit_servo/servo_calcs.h)
  a. Carefully look at how I moved initialization into the constructor and out of the start function.
  b. Carefully look at the various helper functions and behavior in the two different callable functions.

### Testing

All of the new classes in the detail namespace come with 100 unit test coverage.  If you look at the tests for these you'll see how easy these changes made testing each step.  These tests should be really robust and are an easy pattern for copying.

Locally I've been testing these changes with the tutorial and demo launch files and have verified they both solve the bug with halting and create no new issues as far as I can tell.

moveit_ros/moveit_servo/include/moveit_servo/detail/input_check_valid.hpp
```diff
@@ -0,0 +1,78 @@
...
+namespace detail
+{
+/**
+ * @brief      Visitor that checks if the input is valid, only calling next if that is true
+ */
+class InputCheckValid : public InputVisitor
+{
+  rclcpp::Node::SharedPtr node_ = nullptr;
+  std::string command_in_type_ = "unitless";
+  InputVisitor* next_ = nullptr;
```
Abishalini, line 62: Why is this a raw pointer? Can it be a unique/shared pointer?

moveit_ros/moveit_servo/include/moveit_servo/detail/input_check_valid.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+public:
+  InputCheckValid(rclcpp::Node::SharedPtr node, const std::string& command_in_type, InputVisitor* next)
```
Abishalini, line 65:
```suggestion
  InputCheckValid(rclcpp::Node::SharedPtr node, std::string_view command_in_type, InputVisitor* next)
```

moveit_ros/moveit_servo/include/moveit_servo/detail/input_check_valid.hpp
tylerjw: They are raw pointers because all of these InputVisitor objects have their lifetime managed by ServoPipeline with a unique_ptr.  If I made them shared_ptr that would create a circular relationship:

If I made it a shared_ptr:
The lifetime of TimestampNow (next in the default case) would be tied to the lifetime of ServoPipeline 
The lifetime of TimestampNow would be tied to the lifetime of InputCheckValid
The lifetime of InputCheckValid would be tied to the lifetime of ServoPipeline

If I made it a unique_ptr:
The lifetime of TimestampNow would be tied to the lifetime of InputCheckValid
The lifetime of InputCheckValid would be tied to the lifetime of InputSubscriber
The lifetime of InputSubscriber would be tied to ServoPipeline (or whoever ended up controlling the InputSubscriber).

The second option would work, however, it makes testing more complex as I would need to move the unique_ptr<Visitor> into the constructor of each of these objects and retain a raw pointer to it.  Maybe that would be better?

In the current implementation, the lifetime of each of these is maintained by ServoPipeline (vector of unique_ptr<InputVisitor>) and raw pointers are handed to each one for the next one that it calls.

Do you think it would be better if I used unique_ptr instead of a non-owning raw pointer here?  Using the non-owning raw pointer does make the assumption that wherever the next object is in memory does not change and its lifetime is at least as long as the lifetime of the one we are constructing with the raw pointer.

tylerjw: I think I might be wrong about how shared_ptr behaves, but I still don't think it is necessary or should be used here.

tylerjw: I just tried turning these into unique_ptr and it breaks how I need to share access to ServoCalcs in servo.cpp.  I think these should stay raw pointers for now (potentially I could make them shared_ptr in the future) depending on how the rest of the ServoCalcs refactor goes.

tylerjw: I've thought about this more and I changed the implementation to use no raw pointers.  It now uses shared_ptr which does work.  I'm going to delete my ramblings that actually don't make any sense.

tylerjw:
Someday I'd really like to rely on the ROS Executor and timers instead of having to create our own thread.  The primary reason is that as the executor gets better we will benefit from it.  However, the conversation I heard in the latest ROS control meeting had me question converting this back to a timer.  Instead of profiling it and seeing if I could evaluate the performance of the current ROS executors I just put it back as a thread.  This change was relatively simple because the code that contains the thread contains almost nothing else.

I also got rid of all raw pointers for the visitors and convert them to using shared_ptr where I had to.  This is more modern even if I don't like it.

In addition, I made it really explicit when the input message was being copied because it had to be mutated and used const& everywhere else.

tylerjw:
If you read this PR please know that I am moving it back into a draft condition.  It appears the integration test is now flaky in this PR.  I will investigate that next.

PR #839, opened 2021-12-01
Remove unused dependency from online_signal_smoothing cmake

tylerjw:
### Description

This fixes this error:

```
CMake Error at online_signal_smoothing/CMakeLists.txt:49 (find_package):
  By not providing "Findros_testing.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "ros_testing", but CMake did not find one.

  Could not find a package configuration file provided by "ros_testing" with
  any of the following names:

    ros_testingConfig.cmake
    ros_testing-config.cmake

  Add the installation prefix of "ros_testing" to CMAKE_PREFIX_PATH or set
  "ros_testing_DIR" to a directory containing one of the above files.  If
  "ros_testing" provides a separate development package or SDK, be sure it
  has been installed.
```

### DIFF
--- a/moveit_core/online_signal_smoothing/CMakeLists.txt
+++ b/moveit_core/online_signal_smoothing/CMakeLists.txt
@@ -46,7 +46,6 @@ install(FILES ${CMAKE_CURRENT_BINARY_DIR}/${BUTTERWORTH_FILTER_LIB}_export.h DES
 if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
   find_package(ament_cmake_gtest REQUIRED)
-  find_package(ros_testing REQUIRED)
 
   # Lowpass filter unit test
   ament_add_gtest(test_butterworth_filter test/test_butterworth_filter.cpp)


vatanaksoytezer (commented):
Should we add this to `online_signal_smoothing` instead

tylerjw:
> Should we add this to `online_signal_smoothing` instead

`online_signal_smoothing` is just a subdirectory of moveit_core.

vatanaksoytezer (approved):

tylerjw:
I found this when working on #842

moveit_core/package.xml
```diff
@@ -68,6 +68,7 @@
   <test_depend>orocos_kdl</test_depend>
   <test_depend>ament_cmake_gtest</test_depend>
   <test_depend>ament_index_cpp</test_depend>
+  <test_depend>ros_testing</test_depend>
```
JafarAbdi, line 71: I don't see where it's being used in moveit_core, I think we should remove [this line](https://github.com/ros-planning/moveit2/blob/main/moveit_core/online_signal_smoothing/CMakeLists.txt#L49) rather than adding the dependency here
```suggestion
```

JafarAbdi (changes requested):

moveit_core/package.xml
tylerjw: Good catch, thank you.  I pushed this change instead.

JafarAbdi (approved):

PR #884, opened 2021-12-03
Revert debug warning

henningkayser:
This is a commit that I extracted form #452. The warning should be a debug message, it seems to have been merged as part of some verbose debugging attempt. Related note: Are we missing best practices on how to use and filter log messages properly? I've seen temporarily increased logger levels like this quite often and it should not be necessary.


JafarAbdi (approved):

JafarAbdi:
> Are we missing best practices on how to use and filter log messages properly? I've seen temporarily increased logger levels like this quite often and it should not be necessary.

As far as I know, there's no easy way to do it, a couple months ago I made a [PR](https://github.com/ros2/rcutils/pull/347) that hopefully will make our life easier. unfortunately, it still not merged yet

AndyZe (approved):

PR #886, opened 2021-12-03
Various Servo fixes, optional latency compensation parameter

### LINKED ISSUE #696: Consider a "feedback delay" parameter for Servo
Some manipulators have a significant delay in the feedback they provide. The Kinova comes to mind (although I'm not 100% sure that's the issue.)

The delay can lead to a drooping motion over time, because Servo is constantly updating its internal robot state from the feedback.

We might think about adding a delay parameter to get a better joint state estimate.

On the other hand, this might cause issues if the delay is not consistent.

@MarqRazz

AndyZe:
These are all relatively minor bug fixes or documentation improvement except "Add optional latency compensation parameter".

The latency comp parameter was very useful in improving a UR5e's path tracking accuracy. @MarqRazz can attest to that.

If the latency comp parameter is not specified, it doesn't change the behavior at all.

Closes #696

Should be merged with https://github.com/ros-planning/moveit2/pull/886 so that the changes in motion speed make sense with the tutorial.

I lumped the latency parameter into one number because there are quite a few sources of latency that can be hard to quantify:

>      - joint_states publication rate
>      - ROS message transmission. I believe there are 2 message hops
>      - network latency
>      - a little bit of processing time
>      - whatever happens in the ros_control interface
>      - low-level controller execution

@MarqRazz here is a diagram to help understand (one case) where this helps. If there is a net latency of 30 ms (from when the Servo message is published to when the low-level controller executes), it can reduce the path-tracking error quite a bit. Thus the robot's low-level PID output is small --> the robot does not move very fast. In the perfect world where the latency is 0, the path-tracking accuracy will be much larger and the robot would move with the expected speed.

The first green circle is when Servo publishes. The dashed green line represents 30ms of latency. Red is when the low-level controller starts executing. The final purple dot is the end of the control period.

You can see that when latency is present, the difference between (purple-red) is much less.

IMG_20211203_161358960

### DIFF
--- a/moveit_ros/moveit_servo/config/panda_simulated_config.yaml
+++ b/moveit_ros/moveit_servo/config/panda_simulated_config.yaml
@@ -11,10 +11,9 @@ scale:
   rotational:  0.8 # Max angular velocity. Unit is [rad/s]. Only used for Cartesian commands.
   # Max joint angular/linear velocity. Only used for joint commands on joint_command_in_topic.
   joint: 0.5
-# This is a fudge factor to account for any latency in the system, e.g. network latency or poor low-level
-# controller performance. It essentially increases the timestep when calculating the target pose, to move the target
-# pose farther away. [seconds]
-system_latency_compensation: 0.05
+# This is an optional factor to account for any latency or low-level controller ramp up/ramp down in the system.
+# It increases the timestep when calculating the target pose, to move the target pose farther away. [seconds]
+target_pose_lookahead_time: 0.05
 
 ## Properties of outgoing commands
 publish_period: 0.034  # 1/Nominal publish rate [seconds]
--- a/moveit_ros/moveit_servo/config/panda_simulated_config.yaml
+++ b/moveit_ros/moveit_servo/config/panda_simulated_config.yaml
@@ -7,9 +7,9 @@ use_gazebo: false # Whether the robot is started in a Gazebo simulation environm
 command_in_type: "unitless" # "unitless"> in the range [-1:1], as if from joystick. "speed_units"> cmds are in m/s and rad/s
 scale:
   # Scale parameters are only used if command_in_type=="unitless"
-  linear:  0.4  # Max linear velocity. Meters per publish_period. Unit is [m/s]. Only used for Cartesian commands.
-  rotational:  0.8 # Max angular velocity. Rads per publish_period. Unit is [rad/s]. Only used for Cartesian commands.
-  # Max joint angular/linear velocity. Rads or Meters per publish period. Only used for joint commands on joint_command_in_topic.
+  linear:  0.4  # Max linear velocity. Unit is [m/s]. Only used for Cartesian commands.
+  rotational:  0.8 # Max angular velocity. Unit is [rad/s]. Only used for Cartesian commands.
+  # Max joint angular/linear velocity. Only used for joint commands on joint_command_in_topic.
   joint: 0.5
 
 ## Properties of outgoing commands
--- a/moveit_ros/moveit_servo/config/panda_simulated_config.yaml
+++ b/moveit_ros/moveit_servo/config/panda_simulated_config.yaml
@@ -11,6 +11,10 @@ scale:
   rotational:  0.8 # Max angular velocity. Unit is [rad/s]. Only used for Cartesian commands.
   # Max joint angular/linear velocity. Only used for joint commands on joint_command_in_topic.
   joint: 0.5
+# This is a fudge factor to account for any latency in the system, e.g. network latency or poor low-level
+# controller performance. It essentially increases the timestep when calculating the target pose, to move the target
+# pose farther away. [seconds]
+system_latency_compensation: 0.05
 
 ## Properties of outgoing commands
 publish_period: 0.034  # 1/Nominal publish rate [seconds]
--- a/moveit_ros/moveit_servo/config/panda_simulated_config.yaml
+++ b/moveit_ros/moveit_servo/config/panda_simulated_config.yaml
@@ -11,9 +11,10 @@ scale:
   rotational:  0.8 # Max angular velocity. Unit is [rad/s]. Only used for Cartesian commands.
   # Max joint angular/linear velocity. Only used for joint commands on joint_command_in_topic.
   joint: 0.5
+
 # This is an optional factor to account for any latency or low-level controller ramp up/ramp down in the system.
 # It increases the timestep when calculating the target pose, to move the target pose farther away. [seconds]
-target_pose_lookahead_time: 0.05
+target_pose_lookahead_time: 0.0
 
 ## Properties of outgoing commands
 publish_period: 0.034  # 1/Nominal publish rate [seconds]
--- a/moveit_ros/moveit_servo/config/panda_simulated_config_pose_tracking.yaml
+++ b/moveit_ros/moveit_servo/config/panda_simulated_config_pose_tracking.yaml
@@ -7,9 +7,9 @@ use_gazebo: false # Whether the robot is started in a Gazebo simulation environm
 command_in_type: "speed_units" # "unitless"> in the range [-1:1], as if from joystick. "speed_units"> cmds are in m/s and rad/s
 scale:
   # Scale parameters are only used if command_in_type=="unitless"
-  linear:  0.6  # Max linear velocity. Meters per publish_period. Unit is [m/s]. Only used for Cartesian commands.
-  rotational:  0.3 # Max angular velocity. Rads per publish_period. Unit is [rad/s]. Only used for Cartesian commands.
-  # Max joint angular/linear velocity. Rads or Meters per publish period. Only used for joint commands on joint_command_in_topic.
+  linear:  0.6  # Max linear velocity. Unit is [m/s]. Only used for Cartesian commands.
+  rotational:  0.3 # Max angular velocity. Unit is [rad/s]. Only used for Cartesian commands.
+  # Max joint angular/linear velocity. Only used for joint commands on joint_command_in_topic.
   joint: 0.01
 
 ## Properties of outgoing commands
--- a/moveit_ros/moveit_servo/config/panda_simulated_config_pose_tracking.yaml
+++ b/moveit_ros/moveit_servo/config/panda_simulated_config_pose_tracking.yaml
@@ -11,9 +11,10 @@ scale:
   rotational:  0.3 # Max angular velocity. Unit is [rad/s]. Only used for Cartesian commands.
   # Max joint angular/linear velocity. Only used for joint commands on joint_command_in_topic.
   joint: 0.01
+
 # This is an optional factor to account for any latency or low-level controller ramp up/ramp down in the system.
 # It increases the timestep when calculating the target pose, to move the target pose farther away. [seconds]
-target_pose_lookahead_time: 0.05
+target_pose_lookahead_time: 0.0
 
 ## Properties of outgoing commands
 publish_period: 0.034  # 1/Nominal publish rate [seconds]
--- a/moveit_ros/moveit_servo/config/panda_simulated_config_pose_tracking.yaml
+++ b/moveit_ros/moveit_servo/config/panda_simulated_config_pose_tracking.yaml
@@ -11,6 +11,10 @@ scale:
   rotational:  0.3 # Max angular velocity. Unit is [rad/s]. Only used for Cartesian commands.
   # Max joint angular/linear velocity. Only used for joint commands on joint_command_in_topic.
   joint: 0.01
+# This is a fudge factor to account for any latency in the system, e.g. network latency or poor low-level
+# controller performance. It essentially increases the timestep when calculating the target pose, to move the target
+# pose farther away. [seconds]
+system_latency_compensation: 0.05
 
 ## Properties of outgoing commands
 publish_period: 0.034  # 1/Nominal publish rate [seconds]
--- a/moveit_ros/moveit_servo/config/panda_simulated_config_pose_tracking.yaml
+++ b/moveit_ros/moveit_servo/config/panda_simulated_config_pose_tracking.yaml
@@ -11,10 +11,9 @@ scale:
   rotational:  0.3 # Max angular velocity. Unit is [rad/s]. Only used for Cartesian commands.
   # Max joint angular/linear velocity. Only used for joint commands on joint_command_in_topic.
   joint: 0.01
-# This is a fudge factor to account for any latency in the system, e.g. network latency or poor low-level
-# controller performance. It essentially increases the timestep when calculating the target pose, to move the target
-# pose farther away. [seconds]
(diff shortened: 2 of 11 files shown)


moveit_ros/moveit_servo/config/panda_simulated_config.yaml
```diff
@@ -7,10 +7,14 @@ use_gazebo: false # Whether the robot is started in a Gazebo simulation environm
 command_in_type: "unitless" # "unitless"> in the range [-1:1], as if from joystick. "speed_units"> cmds are in m/s and rad/s
 scale:
   # Scale parameters are only used if command_in_type=="unitless"
-  linear:  0.4  # Max linear velocity. Meters per publish_period. Unit is [m/s]. Only used for Cartesian commands.
-  rotational:  0.8 # Max angular velocity. Rads per publish_period. Unit is [rad/s]. Only used for Cartesian commands.
-  # Max joint angular/linear velocity. Rads or Meters per publish period. Only used for joint commands on joint_command_in_topic.
```
AndyZe, line 12: The deleted sentence was inaccurate

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
```diff
@@ -148,13 +148,15 @@ class ServoCalcs
 
   /** \brief If incoming velocity commands are from a unitless joystick, scale them to physical units.
    * Also, multiply by timestep to calculate a position change.
+   * @return a vector of Cartesian deltas
    */
-  Eigen::VectorXd scaleCartesianCommand(const geometry_msgs::msg::TwistStamped& command);
+  Eigen::VectorXd scaleCartesianTwistToCartesianPositionDelta(const geometry_msgs::msg::TwistStamped& command);
```
AndyZe, line 153: Renamed this function to be a bit more specific about what it does

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -619,9 +614,6 @@ bool ServoCalcs::internalServoUpdate(Eigen::ArrayXd& delta_theta,
   // Set internal joint state from original
   internal_joint_state_ = original_joint_state_;
 
-  // Enforce SRDF Velocity, Acceleration limits
-  delta_theta = enforceVelocityLimits(joint_model_group_, parameters_->publish_period, delta_theta);
```
AndyZe, line 623: This is moved down so that velocity scaling doesn't happen so soon. Previously the order was:

- apply velocity limits
- apply velocity scaling for collisions
- low-pass filter
- apply position limits

The new order is:

- apply velocity scaling for collisions
- low-pass filter
- apply velocity limits
- apply position limits

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -703,9 +693,6 @@ bool ServoCalcs::applyJointUpdate(const Eigen::ArrayXd& delta_theta, sensor_msgs
     // Calculate joint velocity
     joint_state.velocity[i] =
```
MarqRazz, line 694: Can you please add some debug prints that I can use to prove the velocity calculation we have here is correct (i.e. print this calculation, the actual joint velocity, and the difference between the two)? This still looks wrong because the arm is not moving this entire distance commanded in the amount of time we are giving it. If it did then Servo would cause the arm to jerk from a start/stop state every update.

MarqRazz (changes requested):

moveit_ros/moveit_servo/config/panda_simulated_config.yaml
```diff
@@ -7,10 +7,14 @@ use_gazebo: false # Whether the robot is started in a Gazebo simulation environm
...
+  linear:  0.4  # Max linear velocity. Unit is [m/s]. Only used for Cartesian commands.
```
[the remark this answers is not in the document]
AndyZe: OK, will update it. Note that I was testing on a UR configuration, not here with the panda.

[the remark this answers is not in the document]
AndyZe: I think these values were so large because the tutorial was publishing joystick commands of 0.2.

Changed the joystick command to 1.0 and changed this to 0.08. That makes a little more sense.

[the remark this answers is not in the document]
AndyZe: New motion speed
new_motion_speed

moveit_ros/moveit_servo/config/panda_simulated_config_pose_tracking.yaml
```diff
@@ -7,10 +7,14 @@ use_gazebo: false # Whether the robot is started in a Gazebo simulation environm
...
-  rotational:  0.3 # Max angular velocity. Rads per publish_period. Unit is [rad/s]. Only used for Cartesian commands.
-  # Max joint angular/linear velocity. Rads or Meters per publish period. Only used for joint commands on joint_command_in_topic.
+  linear:  0.6  # Max linear velocity. Unit is [m/s]. Only used for Cartesian commands.
+  rotational:  0.3 # Max angular velocity. Unit is [rad/s]. Only used for Cartesian commands.
+  # Max joint angular/linear velocity. Only used for joint commands on joint_command_in_topic.
   joint: 0.01
+# This is a fudge factor to account for any latency in the system, e.g. network latency or poor low-level
+# controller performance. It essentially increases the timestep when calculating the target pose, to move the target
+# pose farther away. [seconds]
+system_latency_compensation: 0.05
```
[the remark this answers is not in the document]
AndyZe: I think it's a little more complicated than the publication rate of the arm's joint states, Marq.

Yes, I can probably make this adaptive.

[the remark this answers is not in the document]
AndyZe: The sources of latency I can think of are:

- joint_states publication rate
- ROS message transmission. I believe there are 2 message hops
- network latency
- a little bit of processing time

moveit_ros/moveit_servo/src/servo_calcs.cpp
AndyZe: please see the diagram in the PR description.

AndyZe: I'll work on actually tracking the Cartesian position and using that feedback to suggest a latency parameter.

tylerjw:
If you want to publish numbers simply for debugging would it make sense it add https://github.com/pal-robotics/pal_statistics to servo.  It has a way of sampling and publishing internal state in a way that should not affect the performance of servo and wouldn't require a separate node.  @facontidavide showed me it when I asked about how he would go about plotting something like this.

moveit_ros/moveit_servo/config/panda_simulated_config.yaml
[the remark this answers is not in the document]
MarqRazz: Is `joint4` the elbow on the Panda robot?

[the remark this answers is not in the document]
AndyZe: Yeah -- confirmed with tf.

elbow

MarqRazz:
> @MarqRazz here is a diagram to help understand why this works. If there is a net latency of 30 ms (from when the Servo message is published to when the low-level controller executes), it can reduce the path-tracking error quite a bit. Thus the robot's low-level PID output is small --> the robot does not move very fast. In the perfect world where the latency is 0, the path-tracking accuracy will be much larger and the robot would move with the expected speed.
> 
> The first green circle is when Servo publishes. The dashed green line represents 30ms of latency. Red is when the low-level controller starts executing. The final purple dot is the end of the control period.
> 
> You can see that when latency is present, the difference between (purple-red) is much less.

@AndyZe if the robot were to perform like you are showing with `latency = 0` then the robot would start and stop at each point which would equate to very jerky motion. The goal of Servo is to constantly keep the command in front of the robot (i.e. it should have following error) so that motion is smooth and moves in the direction commanded.

I assume the latency in this system is constant and is included in the update rate. That is if we are running this loop at 500 Hz then the are will send 500 commands per second to the robot, not 500 + some offset (i.e. `timestep = parameters_->publish_period + parameters_->system_latency_compensation`). That is if we keep the fudge factor then moving the carrot further ahead does not equate to the robot tracking the command better it just moves the goal further out causing the robot to move faster.

MarqRazz:
> If you want to publish numbers simply for debugging would it make sense it add https://github.com/pal-robotics/pal_statistics to servo. It has a way of sampling and publishing internal state in a way that should not affect the performance of servo and wouldn't require a separate node. @facontidavide showed me it when I asked about how he would go about plotting something like this.

@tylerjw I like this suggestion but was thinking it would only be a debugging print statement, meaning you only take the performance hit when you enable it without adding dependencies.

tylerjw:
> meaning you only take the performance hit when you enable it without adding dependencies.

I think you'd only take a performance hit from pal_statistics if you subscribe to its topic.  The real downside is adding yet another dependency.

AndyZe:
We don't need to publish any extra data, though. Everything we need is already published.

moveit_ros/moveit_servo/config/panda_simulated_config_pose_tracking.yaml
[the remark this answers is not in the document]
AndyZe: I changed this comment to be more professional and the parameter was renamed. Now it says:

```
# This is an optional factor to account for any latency or low-level controller ramp up/ramp down in the system.
# It increases the timestep when calculating the target pose, to move the target pose farther away. [seconds]
target_pose_lookahead_time: 0.05
```

Re. naming, I looked around to see if there is any name for this technique but I did not find an exact match. A Smith Predictor is similar but not exactly the same. That is strange there's not a better name for it already; I know the UR robots do something like this internally.

I have a branch where I tried to calculate the system latency automatically but TBH it did not work very well yet, so I won't put it in this PR.

moveit_ros/moveit_servo/src/servo_calcs.cpp
AndyZe: I started a branch called "latency monitor" here. It's not ready to merge yet, though.

https://github.com/AndyZe/moveit2/tree/andyz/latency_monitor

AndyZe:
@MarqRazz & @tylerjw I do want to get this merged soon. There are some good fixes in here re how the velocity limits are applied. The new "target pose lookahead time" parameter is completely optional.

tylerjw:
I'm mostly not convinced that adding some fixed value from the parameters is a good long-term solution.  I would like to find some way to measure and quantify what this delay is so instead of having a number you can tweak it to try to get the result closer to the velocity you expect you would actually set it from a measured value and achieve the velocities you are requesting.

moveit_ros/moveit_servo/config/panda_simulated_config.yaml
tylerjw: Did you update them in the parameter descriptions in servo_parameters.cpp?

moveit_ros/moveit_servo/src/servo_calcs.cpp
tylerjw: The low pass filter in a way is a velocity clamp too right?  It smooths large changes in position.  Have you measured how much velocity is being reduced by the low pass filter?

tylerjw: Are the commanded values for each of these steps published somewhere so we can plot them and measure how much each step is changing the result for a given input?

AndyZe:
> I would like to find some way to measure and quantify what this delay is so instead of having a number you can tweak it to try to get the result closer to the velocity you expect you would actually set it from a measured value and achieve the velocities you are requesting.

I attempted that here. It does not work very well yet and I don't have time/hardware to debug further at the moment. Let me know if you see any issues.

https://github.com/AndyZe/moveit2/tree/andyz/latency_monitor

At least it is only one parameter to adjust, right?

gavanderhoorn:
> I'm mostly not convinced that adding some fixed value from the parameters is a good long-term solution. 

Just an observation: many "tracking" systems provided by the big OEMs actually include exactly the kind of *twiddle factor* @AndyZe proposes here. Line-tracking systems or general IO interfacing with external "triggering devices" would be examples of this.

Sometimes even multiple (ie: one for "io delay", another for "processing delay" and yet another for "misc delay").

There can be many possible contributing factors to these kinds of delays, and their combination doesn't always make it possible to express them in a single unit, hence the "twiddle factor" naming.

Whether you'd want to implement this by scaling control outputs or in some other way is another discussion of course.

tylerjw:
Have you considered creating feedback from observing changes in joint states from previous commands adjusting the delay to achieve velocities closer to what the user requested?  Ideally, this doesn't change that quickly and you could observe the behavior of the robot to determine how much time offset to add in order to achieve the requested velocity (that is assuming it is actually modeled well enough by a single number).

AndyZe:
> Have you considered ...

Tyler, I already posted a link a couple message up where I tried something very similar to that on hardware. I only had about an hour to test on hardware but it was not working well at that time so I won't add it to this MR.

I think you have lost focus on the bigger picture here.

moveit_ros/moveit_servo/src/servo_calcs.cpp
AndyZe: These comments don't really seem to apply to this PR.

Yes, I've plotted the filter output many times in the past and it even has some unit tests.

tylerjw:
> I think you have lost focus on the bigger picture here.

The bigger picture is that I'd like to move moveit in the direction of fewer nobs and dials, not more.  It is already hard enough to use.  User interfaces should express the output intent, not expose nobs like this that have to be tuned.  When you expose a nob that has to be tuned your users end up depending on the specific behavior in your library and that makes it much harder to change the behavior for the better without breaking users who have optimized their setups around your nob.  It also makes it much more likely someone will misunderstand how to use your nob and create poorer results than if it didn't exist.

tylerjw:
I generally disagree with exposing a nob to tune in the parameters if at all possible.  At the same time, this PR is crashing and generating a stack trace in a test on rolling.

AndyZe:
> I generally disagree with exposing a nob to tune in the parameters if at all possible. At the same time, this PR is crashing and generating a stack trace in a test on rolling.

Sorry about the crash. It was a silly mistake that's fixed now.

You want an auto-tuner that runs in the background and just works perfectly. It sounds nice but here are some reasons that may be more difficult than you realized:

- 6 or 7-dof data has to be squashed down to a scalar.
- The robot needs to move over a trajectory that's free of singularities, joint limits, and collisions. Good luck coming up with a trajectory that generalizes to all robot types.
- There would be a delay at the beginning of the auto-tuning process.

I did set the parameter by zero to default and added validation, making it a bit harder for the user to screw up.

AndyZe:
Closing in favor of #956

PR #893, opened 2021-12-06
Get parameter on initialize

galou:
Get parameter `trajectory_execution.execution_duration_monitoring` in
initialize().

### Description

We have issues with a slow simulation and a velocity-limited robot where the trajectory execution fails because the robot goes more slowly than expected. I looked at the code and realized that the parameter `trajectory_execution.execution_duration_monitoring` is not set on startup but only in the callback for parameter change. I hope that this change correct the issue.

### Checklist


moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -181,6 +181,8 @@ void TrajectoryExecutionManager::initialize()
       EXECUTION_EVENT_TOPIC, 100, std::bind(&TrajectoryExecutionManager::receiveEvent, this, std::placeholders::_1),
       options);
 
+  controller_mgr_node_->get_parameter("trajectory_execution.execution_duration_monitoring",
+                                      execution_duration_monitoring_);
```
AndyZe, line 185: This generally makes sense to me! I would like to see L96-98 moved closer to these lines of code so it's more clear that they are the default and they might get overwritten.

```suggestion
  execution_duration_monitoring_ = true;
  controller_mgr_node_->get_parameter("trajectory_execution.execution_duration_monitoring",
                                      execution_duration_monitoring_);
  execution_velocity_scaling_ = 1.0;
  controller_mgr_node_->get_parameter("trajectory_execution.allowed_execution_duration_scaling",
                                      allowed_execution_duration_scaling_);
  controller_mgr_node_->get_parameter("trajectory_execution.allowed_goal_duration_margin",
  allowed_start_tolerance_ = 0.01;
  controller_mgr_node_->get_parameter("trajectory_execution.allowed_start_tolerance", allowed_start_tolerance_);
```

AndyZe: [get_parameter_or()](https://docs.ros2.org/bouncy/api/rclcpp/classrclcpp_1_1_node.html#ae5d51004e6e4fc32194251587eb8711a) is probably a better way to set a default value if none is defined.

tylerjw: I think the regular get_parameter will throw an exception if it is not declared first (or someone didn't use the setting that decades everything in your yaml file and it was in the yaml file).

AndyZe (approved):
This looks good enough for me. You could use get_parameter_or() but that's functionally the equivalent. Thanks!

vatanaksoytezer (approved):

JafarAbdi (changes requested):
Thanks @galou for the fix, please revert unrelated changes, I can't push to your branch to fix it

this PR should be only

```diff
diff --git a/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp b/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
index 86ae2a273..07f28fc21 100644
--- a/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
+++ b/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
@@ -181,6 +181,8 @@ void TrajectoryExecutionManager::initialize()
       EXECUTION_EVENT_TOPIC, 100, std::bind(&TrajectoryExecutionManager::receiveEvent, this, std::placeholders::_1),
       options);
 
+  controller_mgr_node_->get_parameter("trajectory_execution.execution_duration_monitoring",
+                                      execution_duration_monitoring_);
   controller_mgr_node_->get_parameter("trajectory_execution.allowed_execution_duration_scaling",
                                       allowed_execution_duration_scaling_);
   controller_mgr_node_->get_parameter("trajectory_execution.allowed_goal_duration_margin",
```

JafarAbdi:
Continued in https://github.com/ros-planning/moveit2/pull/996

PR #896, opened 2021-12-08
Remove temporary prbt related CI fixes

vatanaksoytezer:
Depends on https://github.com/ros-planning/moveit_resources/pull/105, will fail until that gets merged. After adding a build guard to prbt packages we wouldn't need the hacky fixes anymore. This should fix the regular CI builds as well as docker builds


henningkayser:
@vatanaksoytezer this can be closed, right?

vatanaksoytezer:
İdeally we should merge this if pilz is not a problem

PR #900, opened 2021-12-09
Fix boost linking errors

vatanaksoytezer:
There were some weird boost linking errors since yesterday, I was able to reproduce by running industrial_ci locally, and this PR should hopefully fix them. I have yet to know the root cause, so I'm happy to resolve this different ways as well if someone has any suggestion.

### DIFF
--- a/moveit_core/distance_field/CMakeLists.txt
+++ b/moveit_core/distance_field/CMakeLists.txt
@@ -16,6 +16,7 @@ ament_target_dependencies(${MOVEIT_LIB_NAME}
   tf2_eigen
   OCTOMAP
 )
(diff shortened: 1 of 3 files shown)


tylerjw (approved):

lilustga:
I noticed that moveit_core already includes Boost regex and iostreams in `ConfigExtras.cmake`. I think however, it is being overwritten by this change in geometric_shapes: https://github.com/ros-planning/geometric_shapes/pull/206

@tylerjw What do you think of instead moving `include(ConfigExtras.cmake)` to after find_package is called for geometric shapes?

PR #909, opened 2021-12-09
Pilz test_resources

### LINKED ISSUE #885: Fix circular dependency with moveit_resources (PILZ/PRBT)
Tracking issue for the MoveIt 2 side of https://github.com/ros-planning/moveit/issues/2947.

* The PILZ planner should use a supported robot rather than PRBT
* The circular dependency with moveit_resources should be fixed

tylerjw:
- Adding RPBT config
- prbt_support: drop all test code
- Port prbt packages to ROS 2
- Move PRBT into test_configs directory

### Description

This fixes the circular dependency with moveit_resources.  This is a temporary solution until we are able to test the Pilz planners with a different robot.

I preserved the commit history of these by cherry-picking from moveit_resources.

This depends on this PR: https://github.com/ros-planning/moveit_resources/pull/106

This fixes #885 

### Checklist


tylerjw:
Sadness... the IK solver that is generated code is not tested and would reduce our code coverage by 19.79%.  I could exclude it from the coverage report... meh

vatanaksoytezer (approved):
Made sense to me as a temporary change. @tylerjw can you also remove TODO I added with in #915.

henningkayser (approved):

tylerjw:
pre-release test here: https://github.com/tylerjw/moveit2/actions/runs/1612521573

v4hn:
> This is a temporary solution until we are able to test the Pilz planners with a different robot.

Did you really have such urgent issues to fix this way that you could not invest part of the time on actually transitioning the tests to Fanuc and including the opw ik plugin? :crying_cat_face: 
I saw this pr only now after it got merged, but this is exactly what we decided to avoid in the moveit repository.
Instead we just build the resources and main repo together wherever we expect to run tests.

Explicitly notice that this **does not**
> fixes #885

The issue - as well as the parallel moveit one - explicitly states:
> The PILZ planner should use a supported robot rather than PRBT

Me sad kitty...

tylerjw:
> Did you really have such urgent issues to fix this way that you could not invest part of the time on actually transitioning the tests to Fanuc and including the opw ik plugin?

Yes, actually I do have things that I think will provide much more value.  Here is a short list of things I'd like to do before spending any time on migrating the Pilz planner tests of the PRBT:

* Modernize memory allocation in moveit_core (no reason we should be using malloc in C++17)
* Enforce safe handling of pointers in API using some sort of railroad pattern
* Eliminate pointers (and dynamic memory) from the API of moveit_core where they are unnecessary 
* Enable sanitizers (and fix the issues they find) in CI
* Enable a much larger set of clang-tidy rules and fix the issues they find
* Port BioIK to ROS 2 and refactor it to have tests and use memory and threads in a safe way ( sanitizers in CI )
* Use a dependency injection pattern to add tests to much of the untested code in moveit_ros

This doesn't mean that I want to stand in the way of anyone fixing the issue with the Pilz test if they want to.

v4hn:
@tylerjw if you ever want to talk, just message me :smile_cat: . I understand your frustration about things to do and nobody doing them, but I don't see your point because you did not answer the question.

PR #911, opened 2021-12-09
Prefer lld linker over ld

### LINKED ISSUE #895: Long build times in CI after Pilz merge
### Description

CI is now taking over 3hrs to complete after the pilz merge.  We need to analyze why that is happening and find some fix to it.  3hr build times are awful.

tylerjw:
### Description

This might fix #895


tylerjw:
This gets us down to less than an hour with the clang tidy job.  This can be squashed or I can do that.  There are many commits here because I tried several things.  I would appreciate reviews from anyone else who's used other linkers or have any other ideas for his to fix this.

moveit_common/cmake/moveit_package.cmake
```diff
@@ -46,6 +46,16 @@ macro(moveit_package)
     add_compile_options(-Wall -Wextra
       -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls -Wcast-qual
       -Wno-unused-parameter -Wno-unused-function)
+
+    # Use lld linker if it is installed
+    execute_process(COMMAND ${CMAKE_CXX_COMPILER} -fuse-ld=lld -Wl,--version OUTPUT_VARIABLE lld_test ERROR_QUIET)
```
jspricke, line 51: This feels like too much automation, I would simply set the variables in the job.
Otherwise users with both linkers installed might wonder what changed or if they have a linker set on the command line it will get overwritten.

jspricke (changes requested):

moveit_common/cmake/moveit_package.cmake
tylerjw: I wasn't sure if anyone would care what linker ran.  And this way I give everyone (who has lld installed) faster builds without them doing anything.

I didn't know there was alternatives to ld until this week.  In my testing lld seems to perform slightly better than gold and both of them are much much faster than ld.

jspricke: Well, I guess most people don't care but we can't guarantee that it won't break moveit for some. If you really want to change the behavior for everyone, I would ask for more tests then this. Given that uses can change it easily in thair environment, I would stick with the default.

tylerjw: I opened another PR with the change only being made to the CI config.  Let me know if you think that is a better approach.

PR #913, opened 2021-12-10
Larger ccache

tylerjw:
- Log ccache stats
- Configure ccache to be larger

### Description

This depends on the PR that prints the ccache stats.  Github is compressing the ccache folder before storing it so we can have a larger one.  When I ran the one that prints the stats on the debug builds I see that it is deleting 402 files.  Hopefully this will help:

```
  $ ( ccache -s; )
  cache directory                     /home/runner/work/moveit2/moveit2/.ccache
  primary config                      /home/runner/work/moveit2/moveit2/.ccache/ccache.conf
  secondary config      (readonly)    /etc/ccache.conf
  stats updated                       Fri Dec 10 00:32:43 2021
  cache hit (direct)                  1185
  cache hit (preprocessed)             226
  cache miss                          5861
  cache hit rate                     19.40 %
  called for link                     4744
  called for preprocessing              54
  compile failed                        86
  cache file missing                     1
  cleanups performed                   402
  files in cache                      1182
  cache size                           4.2 GB
  max cache size                       5.0 GB
```


vatanaksoytezer (approved):

tylerjw:
Github increased it's cache size to 10gb / repo.  After compression, this cache goes from ~700mb to 1.2gb.  With this change the code cov run with debug is much faster.  The clang one is still really slow though.  Look at the other change (lld linker) for the fix to that.

PR #915, opened 2021-12-13
Docker - Temporarily move moveit_resources under target workspace due to #885

vatanaksoytezer:
This should fix docker CI fails in main. I would be happy if someone (that has write access) can test this in a branch before merging.


tylerjw (approved):

vatanaksoytezer:
@tylerjw tested locally, docker build completed without any issues.

Issue #875, opened 2021-12-01
Get rid of std::endl


tylerjw:
https://github.com/cpp-best-practices/cppbestpractices/blob/master/08-Considering_Performance.md#get-rid-of-stdendl

PR #918, opened 2021-12-14
Get rid of "std::endl"

ParthasarathyBana:
As described in PR #875, as a solution to the good first issue, this PR removes `std::endl` in all the code files in `moveit2` repo and replaces them with `'\n'`. As mentioned [here](https://github.com/cpp-best-practices/cppbestpractices/blob/master/08-Considering_Performance.md#get-rid-of-stdendl), removing `std::endl` is considered to be a better code practice as it improves performance. More information of when to use `std::endl` and when not to use, can be found [here](https://accu.org/journals/overload/27/149/sharpe_2619/)


tylerjw (approved):
This looks good.  We'll see if it passes CI first.

tylerjw (approved):

PR #932, opened 2021-12-20
Put hybrid planning actions under a common namespace

AndyZe:
This groups the hybrid planning action names together so they are easy to find, for example, with `ros2 action list`.

We're starting to use hybrid planning with a dual arm setup, so cleaning up the interface is becoming more important to me.

Test with the usual hybrid planning demo:

`ros2 launch moveit_hybrid_planning hybrid_planning_demo.launch.py`


moveit_ros/hybrid_planning/global_planner/global_planner_component/src/global_planner_component.cpp
```diff
@@ -63,7 +64,7 @@ bool GlobalPlannerComponent::initializeGlobalPlanner()
 {
   // Initialize global planning request action server
   global_planning_request_server_ = rclcpp_action::create_server<moveit_msgs::action::GlobalPlanner>(
-      node_, "global_planning_action",
+      node_, GLOBAL_PLANNING_ACTION_NAME,
```
henningkayser, line 67: Shouldn't these be private to the node in order to allow running multiple pipelines? If yes, prepending "~" should do that

henningkayser (approved):
Makes a lot of sense to use a common namespace here. Using private topics like suggested in the comment is really an option that we should consider, but not strictly necessary here

moveit_ros/hybrid_planning/global_planner/global_planner_component/src/global_planner_component.cpp
AndyZe: Great. I was thinking in the ROS1 mindset here.

JafarAbdi (commented):
`GLOBAL_PLANNING_ACTION_NAME`, `LOCAL_PLANNING_ACTION_NAME`, and `RUN_HYBRID_PLANNING_ACTION_NAME` are being duplicated in different files, should we move them to a new header file and include that file ?

henningkayser:
> `GLOBAL_PLANNING_ACTION_NAME`, `LOCAL_PLANNING_ACTION_NAME`, and `RUN_HYBRID_PLANNING_ACTION_NAME` are being duplicated in different files, should we move them to a new header file and include that file ?

that would make a lot of sense

AndyZe:
Where would you suggest putting that header file? It doesn't belong in any of the folders that exist right now: `global_planner`, `hybrid_planning_manager`, or `local_planner`

henningkayser:
> Where would you suggest putting that header file? It doesn't belong in any of the folders that exist right now: `global_planner`, `hybrid_planning_manager`, or `local_planner`

good question, the directory structure doesn't really allow common resources. Thinking about this more, it could make sense to move the directories into top-level `src` and `include` and move common headers into `include`. Here, the only alternative I see is to add a new subdirectory for the same purpose

AndyZe:
OK, I put the namespace stuff in a common header file. I find the `~` does not work, though. However, everything works if I delete the `~`.

```
namespace moveit::hybrid_planning
{
namespace
{
constexpr char GLOBAL_PLANNING_ACTION_NAME[] = "~/hybrid_planning/global_planning_action";
constexpr char LOCAL_PLANNING_ACTION_NAME[] = "~/hybrid_planning/local_planning_action";
constexpr char RUN_HYBRID_PLANNING_ACTION_NAME[] = "~/hybrid_planning/run_hybrid_planning";
}  // namespace
}  // namespace moveit::hybrid_planning
```

moveit_ros/hybrid_planning/hybrid_planning_common/include/moveit/hybrid_planning_common/hybrid_planning_common.h
```diff
@@ -0,0 +1,49 @@
...
+#pragma once
+
+namespace moveit::hybrid_planning
+{
+namespace
+{
+constexpr char GLOBAL_PLANNING_ACTION_NAME[] = "~/hybrid_planning/global_planning_action";
+constexpr char LOCAL_PLANNING_ACTION_NAME[] = "~/hybrid_planning/local_planning_action";
+constexpr char RUN_HYBRID_PLANNING_ACTION_NAME[] = "~/hybrid_planning/run_hybrid_planning";
+}  // namespace
```
JafarAbdi, line 48: To prevent ODR violation without the need for the anonymous namespace

```suggestion
inline constexpr char GLOBAL_PLANNING_ACTION_NAME[] = "~/hybrid_planning/global_planning_action";
inline constexpr char LOCAL_PLANNING_ACTION_NAME[] = "~/hybrid_planning/local_planning_action";
inline constexpr char RUN_HYBRID_PLANNING_ACTION_NAME[] = "~/hybrid_planning/run_hybrid_planning";
```

AndyZe: Thanks, done

AndyZe:
The `~` causes this:  (`ros2 action list`)

```
/global_planner/hybrid_planning/global_planning_action
/hybrid_planning_demo_node/hybrid_planning/run_hybrid_planning
/hybrid_planning_manager/hybrid_planning/local_planning_action
/local_planner/hybrid_planning/local_planning_action
```
So I'm going to suggest we pass an action_namespace parameter from the launch files.

AndyZe:
OK, I moved the action name parameters to a common config file. I think this is nice, it fits well with the other config files that already exist:

```
global_planner.yaml
local_planner.yaml
hybrid_planning_manager.yaml
common_hybrid_planning_params.yaml --> the new one
```

PR #942, opened 2021-12-28
Fix std vector insertion (collision distance field)

andreas-botbuilt:
### Description

Calling `getCollisionPoints` on `non_group_attached_body_decomposition` ([see here](https://github.com/ros-planning/moveit2/blob/main/moveit_core/collision_distance_field/src/collision_env_distance_field.cpp#L945-L946)) returns iterators to different vectors, thus causing undefined behavior, and resulting in the error: 

> terminate called after throwing an instance of 'std::length_error'
> what(): vector::_M_range_insert

This PR fixes this error.

### Checklist


moveit_core/collision_distance_field/src/collision_env_distance_field.cpp
```diff
@@ -942,8 +942,8 @@ DistanceFieldCacheEntryPtr CollisionEnvDistanceField::generateDistanceFieldCache
       for (collision_detection::PosedBodyPointDecompositionVectorPtr& non_group_attached_body_decomposition :
            non_group_attached_body_decompositions)
       {
-        all_points.insert(all_points.end(), non_group_attached_body_decomposition->getCollisionPoints().begin(),
```
JafarAbdi, line 945: > Calling getCollisionPoints on non_group_attached_body_decomposition (see here) returns iterators to different vectors

How are they different ? [getCollisionPoints](https://github.com/ros-planning/moveit2/blob/main/moveit_core/collision_distance_field/include/moveit/collision_distance_field/collision_distance_field_types.h#L358) returns a reference to a member variable

andreas-botbuilt: I guess my error description is not precise. The two calls to `getCollisionPoints` in the line referenced above provide two different copies of the collision points. So, when `begin()` and `end()` are called, it is the beginning and end of two **different** objects. Instead, it needs to be the beginning and end of the **same** object to make the insertion work. See also: http://www.cplusplus.com/forum/beginner/173704/#msg861903.

JafarAbdi: As you can see from the link `getVec` returns `std::vector<double>` whereas `getCollisionPoints` returns `const EigenSTL::vector_Vector3d&` they're not the same

JafarAbdi: Do you mind sharing the stacktrace ?

andreas-botbuilt: I think you were looking at PosedBodyPointDecomposition instead of PosedBodyPointDecompositionVector. The latter's [getCollisionPoints](https://github.com/ros-planning/moveit2/blob/main/moveit_core/collision_distance_field/include/moveit/collision_distance_field/collision_distance_field_types.h#L451-L459) does not return a reference to a member variable.

JafarAbdi: Oh, now I see, thanks!

JafarAbdi (approved):
This need to be backported to https://github.com/ros-planning/moveit

henningkayser (approved):

PR #948, opened 2021-12-30
WIP: [hybrid planning] Handle multiple waypoints, stream outgoing commands at the right rate

AndyZe:
- If an incoming request contains multiple waypoints, pass it straight to local planning. Do not re-plan a global trajectory.

The previous implementation of the default global planning plugin could only take one waypoint at a time (see https://github.com/ros-planning/moveit2/blob/c8d3b19da0b902c2ac2ab6791bcd5891bebcc505/moveit_ros/hybrid_planning/global_planner/global_planner_plugins/src/moveit_planning_pipeline.cpp#L101).

We want to take the output of MoveIt Task Constructor and pass an entire trajectory to hybrid planning for execution. This PR just passes an incoming request straight through to the local planner if it has multiple waypoints.

- Stream the outgoing commands at the proper rate.

Prior to this PR, the robot would jump immediately to the target position. With this PR, it follows the target waypoint timestamps.

Previously - jump instantly to the target pose
https://user-images.githubusercontent.com/11284393/148585450-c11676cb-acc4-4a8e-93e2-2fa2f9fd40ed.mp4

With this PR
https://user-images.githubusercontent.com/11284393/148585490-de3adede-a786-4d8c-9c78-f1de0ad4ec42.mp4

### DIFF
--- a/moveit_ros/hybrid_planning/global_planner/global_planner_plugins/src/moveit_planning_pipeline.cpp
+++ b/moveit_ros/hybrid_planning/global_planner/global_planner_plugins/src/moveit_planning_pipeline.cpp
@@ -179,6 +179,7 @@ moveit_msgs::msg::MotionPlanResponse MoveItPlanningPipeline::plan(
 
     // Copy goal constraint into planning component
     planning_components->setGoal(motion_plan_req.goal_constraints);
+    planning_components->setPathConstraints(motion_plan_req.path_constraints);
 
     // Plan motion
     auto plan_solution = planning_components->plan(plan_params);
--- a/moveit_ros/hybrid_planning/global_planner/global_planner_plugins/src/moveit_planning_pipeline.cpp
+++ b/moveit_ros/hybrid_planning/global_planner/global_planner_plugins/src/moveit_planning_pipeline.cpp
@@ -109,6 +109,15 @@ moveit_msgs::msg::MotionPlanResponse MoveItPlanningPipeline::plan(
     response.group_name = motion_plan_req.group_name;
     response.planning_time = motion_plan_req.allowed_planning_time;
     response.error_code.val = moveit_msgs::msg::MoveItErrorCodes::SUCCESS;
+    moveit::core::RobotStatePtr current_state;
+    if (!moveit_cpp_->getCurrentState(current_state, 0.5))
+    {
+      RCLCPP_ERROR(LOGGER, "Could not retrieve the current robot state.");
+      std::exit(EXIT_FAILURE);
+    }
+    moveit_msgs::msg::RobotState current_state_msg;
+    moveit::core::robotStateToRobotStateMsg(*current_state, current_state_msg);
+    response.trajectory_start = current_state_msg;
 
     moveit_msgs::msg::Constraints goal_constraints =
         global_goal_handle->get_goal()->motion_sequence.items.at(0).req.goal_constraints.at(0);
--- a/moveit_ros/hybrid_planning/global_planner/global_planner_plugins/src/moveit_planning_pipeline.cpp
+++ b/moveit_ros/hybrid_planning/global_planner/global_planner_plugins/src/moveit_planning_pipeline.cpp
@@ -97,45 +97,87 @@ moveit_msgs::msg::MotionPlanResponse MoveItPlanningPipeline::plan(
     return response;
   }
 
-  // Process goal
-  if ((global_goal_handle->get_goal())->motion_sequence.items.size() > 1)
+  // If multiple waypoints were requested, assume we have a fully planned trajectory already.
+  // Just forward to the local planner.
+  // This works only for a JointConstraint type.
+  size_t num_waypoints = global_goal_handle->get_goal()->motion_sequence.items.size();
+  if (num_waypoints > 1)
   {
-    RCLCPP_WARN(LOGGER, "Global planner received motion sequence request with more than one item but the "
-                        "'moveit_planning_pipeline' plugin only accepts one item. Just using the first item as global "
-                        "planning goal!");
-  }
-  auto motion_plan_req = (global_goal_handle->get_goal())->motion_sequence.items[0].req;
+    auto motion_plan_req = global_goal_handle->get_goal()->motion_sequence.items[0].req;
+    // Start state is the current state of the robot, by default
+    // Assume these parameters don't change for any waypoint, aka sequence item
+    response.group_name = motion_plan_req.group_name;
+    response.planning_time = motion_plan_req.allowed_planning_time;
+    response.error_code.val = moveit_msgs::msg::MoveItErrorCodes::SUCCESS;
 
-  // Set parameters required by the planning component
-  moveit_cpp::PlanningComponent::PlanRequestParameters plan_params;
-  plan_params.planner_id = motion_plan_req.planner_id;
-  plan_params.planning_pipeline = motion_plan_req.pipeline_id;
-  plan_params.planning_attempts = motion_plan_req.num_planning_attempts;
-  plan_params.planning_time = motion_plan_req.allowed_planning_time;
-  plan_params.max_velocity_scaling_factor = motion_plan_req.max_velocity_scaling_factor;
-  plan_params.max_acceleration_scaling_factor = motion_plan_req.max_acceleration_scaling_factor;
+    moveit_msgs::msg::Constraints goal_constraints =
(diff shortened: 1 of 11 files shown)


moveit_ros/hybrid_planning/global_planner/global_planner_plugins/src/moveit_planning_pipeline.cpp
```diff
@@ -97,44 +97,86 @@ moveit_msgs::msg::MotionPlanResponse MoveItPlanningPipeline::plan(
     return response;
   }
 
-  // Process goal
-  if ((global_goal_handle->get_goal())->motion_sequence.items.size() > 1)
+  // If multiple waypoints were requested, assume we have a fully planned trajectory already.
+  // Just forward to the local planner.
+  // This works only for a JointConstraint type.
+  size_t num_waypoints = global_goal_handle->get_goal()->motion_sequence.items.size();
+  if (num_waypoints > 1)
```
sjahr, line 104: I would prefer that the user actively needs to configure this option. Something like a simple "allow pass-through" [True/False] flag

moveit_ros/hybrid_planning/global_planner/global_planner_plugins/src/moveit_planning_pipeline.cpp
```diff
@@ -97,44 +97,86 @@ moveit_msgs::msg::MotionPlanResponse MoveItPlanningPipeline::plan(
...
   {
-    RCLCPP_WARN(LOGGER, "Global planner received motion sequence request with more than one item but the "
```
sjahr, line 103: In case this behavior is not configurable, I think it might be useful to notify the user that the global planning request was not processed by any planner

sjahr (commented):
I like the idea of giving the user more options for handling a global motion plan request with multiple waypoints with this plugin! In order to avoid confusion, I think it is preferable to keep the default behavior of this plugin as close to the moveit_cpp api as possible and I am not sure how multiple waypoint requests are currently handled by it.
The option to bypass the global planner by just forwarding a trajectory to the local planner is very usefull and might be even worth its own plugin which could be added to the default global planner plugins. Regarding the integration of MTC into the hybrid planning pipeline: The UR10_welding_demo avoided the problem of getting a global trajectory unchanged through the global planner by simly using [MTC as a global planner](https://github.com/PickNikRobotics/UR10e_welding_demo/blob/master/hybrid_planning_demo/src/global_mtc_planner.cpp).

AndyZe:
Thanks @sjahr. I generally agree with everything you wrote. I've just gotten this into a functional state so now I'll clean it up and add the `allow_pass_through` parameter (or something to that effect).

moveit_ros/hybrid_planning/global_planner/global_planner_plugins/src/moveit_planning_pipeline.cpp
AndyZe: It's configurable now.  :+1:

moveit_ros/hybrid_planning/local_planner/trajectory_operator_plugins/src/simple_sampler.cpp
```diff
@@ -68,11 +68,19 @@ SimpleSampler::addTrajectorySegment(const robot_trajectory::RobotTrajectory& new
...
   // Throw away old reference trajectory and use trajectory update
   reference_trajectory_ = std::make_shared<robot_trajectory::RobotTrajectory>(new_trajectory);
 
-  // Parametrize trajectory and calculate velocity and accelerations
-  time_parametrization_.computeTimeStamps(*reference_trajectory_);
+  // If there are very few waypoitns, time-parameterize the trajectory (calculate velocity and accelerations)
+  if (reference_trajectory_->getWayPointCount() <= 2 && !time_parametrization_.computeTimeStamps(*reference_trajectory_))
+  {
+    std::cout << "Time parameterization failed!" << std::endl;
+    feedback_.feedback = "Time parameterization failed.";
```
AndyZe, line 80: @sjahr Do you have any suggestion what to do here if time parameterization fails? Filling the feedback message is nice but that won't stop the local planner from trying to execute, will it?

sjahr: @AndyZe You're right, I think it is not possible right now to stop the local planner at this point except killing the node and thus failing the whole hybrid planning process. I've created an [issue](https://github.com/ros-planning/moveit2/issues/964) with a rough idea of how to add this. What do you think about it?

moveit_ros/hybrid_planning/local_planner/trajectory_operator_plugins/src/simple_sampler.cpp
```diff
@@ -41,21 +41,26 @@ namespace moveit::hybrid_planning
...
 const rclcpp::Logger LOGGER = rclcpp::get_logger("local_planner_component");
-constexpr double WAYPOINT_RADIAN_TOLERANCE = 0.2;  // rad: L1-norm sum for all joints
 }  // namespace
 
 bool SimpleSampler::initialize(const rclcpp::Node::SharedPtr& node, const moveit::core::RobotModelConstPtr& robot_model,
                                const std::string& group_name)
 {
   // Load parameter & initialize member variables
-  if (node->has_parameter("pass_through"))
+  node->declare_parameter("pass_through");
```
AndyZe, line 50: This parameter isn't related to the PR, but I was touching this same area of code and noticed it wasn't working properly.

moveit_ros/hybrid_planning/global_planner/global_planner_plugins/src/moveit_planning_pipeline.cpp
```diff
@@ -97,45 +100,101 @@ moveit_msgs::msg::MotionPlanResponse MoveItPlanningPipeline::plan(
...
+    plan_params.planning_attempts = motion_plan_req.num_planning_attempts;
+    plan_params.planning_time = motion_plan_req.allowed_planning_time;
+    plan_params.max_velocity_scaling_factor = motion_plan_req.max_velocity_scaling_factor;
+    plan_params.max_acceleration_scaling_factor = motion_plan_req.max_acceleration_scaling_factor;
+
+    // Create planning component
+    auto planning_components = std::make_shared<moveit_cpp::PlanningComponent>(motion_plan_req.group_name, moveit_cpp_);
+
+    // Copy goal constraint into planning component
+    planning_components->setGoal(motion_plan_req.goal_constraints);
```
sjahr, line 181: I just realized that this ignores path constraints, so I would suggest to add this:
```suggestion
    // Copy constraints into planning component
    planning_components->setGoal(motion_plan_req.goal_constraints);
    planning_components->setPathConstraints(motion_plan_req.path_constraints);
```

moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
```diff
@@ -279,6 +282,15 @@ void LocalPlannerComponent::executeIteration()
         local_planning_goal_handle_->publish_feedback(local_planner_feedback_);
       }
 
+      // Publish at the period given in the previous waypoint
+      // If initialized. Else, publish immediately.
+      if (prev_waypoint_duration_)
```
henningkayser, line 287: Better match `time_from_start` with the actual current time and sleep for the diff duration to rule out lagging behind.

moveit_ros/hybrid_planning/global_planner/global_planner_plugins/src/moveit_planning_pipeline.cpp
```diff
@@ -97,45 +100,102 @@ moveit_msgs::msg::MotionPlanResponse MoveItPlanningPipeline::plan(
...
+      std::exit(EXIT_FAILURE);
+    }
+
+    // Get joint names
+    for (const auto& constraint : goal_constraints.joint_constraints)
+    {
+      response.trajectory.joint_trajectory.joint_names.push_back(constraint.joint_name);
+    }
+
+    // Get joint positions
```
henningkayser, line 138: This is not really how the constraints are supposed to be used and would break some use cases of the normal planning pipeline. If you really want to forward a predefined trajectory, better adapt the planner logic and publish it to the local planner directly, no need to implement pass-through behavior

AndyZe:
I'm going to close this. It's about to diverge too much from something we could merge into `main`

Issue #783, opened 2021-11-04
CI job 'rolling-ci + ccov' broken for PRs (cached only)


henningkayser:
The CI job "rolling-ci + ccov" currently fails for pull requests with the following build error:

```
   --- stderr: moveit_ros_warehouse
  /usr/bin/ld: /home/runner/work/moveit2/moveit2/.work/upstream_ws/install/warehouse_ros/lib/libwarehouse_ros.so: undefined reference to `rclcpp::SerializedMessage::SerializedMessage(unsigned long, rcutils_allocator_t const&)'
  collect2: error: ld returned 1 exit status
  make[2]: *** [warehouse/CMakeFiles/moveit_warehouse_save_as_text.dir/build.make:318: warehouse/moveit_warehouse_save_as_text] Error 1
  make[1]: *** [CMakeFiles/Makefile2:162: warehouse/CMakeFiles/moveit_warehouse_save_as_text.dir/all] Error 2
  make[1]: *** Waiting for unfinished jobs....
  /usr/bin/ld: /home/runner/work/moveit2/moveit2/.work/upstream_ws/install/warehouse_ros/lib/libwarehouse_ros.so: undefined reference to `rclcpp::SerializedMessage::SerializedMessage(unsigned long, rcutils_allocator_t const&)'
  collect2: error: ld returned 1 exit status
  make[2]: *** [warehouse/CMakeFiles/moveit_warehouse_broadcast.dir/build.make:318: warehouse/moveit_warehouse_broadcast] Error 1
  make[1]: *** [CMakeFiles/Makefile2:189: warehouse/CMakeFiles/moveit_warehouse_broadcast.dir/all] Error 2
  make: *** [Makefile:141: all] Error 2
  ---
  Failed   <<< moveit_ros_warehouse [56.5s, exited with code 2]
```

The main branch doesn' t have this issue, also [disabling the cache fixes the job](https://github.com/ros-planning/moveit2/pull/673/commits/4a7b8402d2d4c0d804a6eb53a77627846bd04752).

tylerjw:
@JafarAbdi is going to add the release date to the hash of the upstream workspace to fix this.

tylerjw:
@JafarAbdi would you mind tackling this?  You created something for checking the date of the upstream (installed from debain) workspace already, correct?

Issue #947, opened 2021-12-30
Segfault in Servo "ExternalCollision" test


henningkayser:
### Description

Failing CI because of Segfault in Servo test.

See https://github.com/ros-planning/moveit2/runs/4665772418?check_suite_focus=true#step:9:12904
```
12904    [test_servo_collision-6] [INFO] [1640861996.843502159] [moveit_servo.servo_launch_test_common.hpp]: Setup time: 0.500997
12905    [servo_node_main-2] [WARN] [1640861997.022630541] [moveit_ros.planning_scene_monitor.planning_scene_monitor]: The complete state of the robot is not yet known.  Missing virtual_joint
12906    [servo_node_main-2] [WARN] [1640861998.024138407] [moveit_ros.planning_scene_monitor.planning_scene_monitor]: The complete state of the robot is not yet known.  Missing virtual_joint
12907    [servo_node_main-2] [WARN] [1640861999.035353542] [moveit_ros.planning_scene_monitor.planning_scene_monitor]: The complete state of the robot is not yet known.  Missing virtual_joint
12908    [servo_node_main-2] [WARN] [1640862000.046106274] [moveit_ros.planning_scene_monitor.planning_scene_monitor]: The complete state of the robot is not yet known.  Missing virtual_joint
12909    [servo_node_main-2] [WARN] [1640862001.049553558] [moveit_ros.planning_scene_monitor.planning_scene_monitor]: The complete state of the robot is not yet known.  Missing virtual_joint
12910    [servo_node_main-2] [WARN] [1640862002.081152181] [moveit_ros.planning_scene_monitor.planning_scene_monitor]: The complete state of the robot is not yet known.  Missing virtual_joint
12911    [servo_node_main-2] [WARN] [1640862003.081609293] [moveit_ros.planning_scene_monitor.planning_scene_monitor]: The complete state of the robot is not yet known.  Missing virtual_joint
12912    [servo_node_main-2] [WARN] [1640862004.082821410] [moveit_ros.planning_scene_monitor.planning_scene_monitor]: The complete state of the robot is not yet known.  Missing virtual_joint
12913    [servo_node_main-2] [WARN] [1640862005.085372836] [moveit_ros.planning_scene_monitor.planning_scene_monitor]: The complete state of the robot is not yet known.  Missing virtual_joint
[...]

PR #952, opened 2022-01-02
Add rosdistro version to upstream ws cache prefix

JafarAbdi:
### Description

Hopefully fix: https://github.com/ros-planning/moveit2/issues/783 and https://github.com/ros-planning/moveit2/issues/947

### Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -70,23 +70,11 @@ jobs:
           sudo rm -rf /usr/local
           df -h
       - uses: actions/checkout@v2
-      - name: Get rosdistro version for a given
-        id: ros_distro_version
-        shell: bash
-        env:
-          ROS_DISTRO: ${{ matrix.env.ROS_DISTRO }}
-        run: |
-          latest-release-date() {
-              echo "$(curl -sL "https://api.github.com/repos/ros/rosdistro/tags?per_page=100" \
-              | grep -o "$1/[0-9]\{4\}-[0-9]\{2\}-[0-9]\{2\}" \
-              | cut -d'/' -f2 \
-              | uniq \
-              | sort -t- -k1n -k2n -k3n -z | head -n 1)"
-          }
-          ROS_DISTRO_VERSION=`latest-release-date $ROS_DISTRO`
-          echo "The latest version for $ROS_DISTRO is $ROS_DISTRO_VERSION"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -66,14 +70,31 @@ jobs:
...
+              echo "$(curl -sL "https://api.github.com/repos/ros/rosdistro/tags?per_page=100" \
+              | grep -o "$1/[0-9]\{4\}-[0-9]\{2\}-[0-9]\{2\}" \
+              | cut -d'/' -f2 \
+              | uniq \
+              | sort -t- -k1n -k2n -k3n -z | head -n 1)"
+          }
+          ROS_DISTRO_VERSION=`latest-release-date $ROS_DISTRO`
+          echo "The latest version for $ROS_DISTRO is $ROS_DISTRO_VERSION"
+          echo set-output name=ROS_DISTRO_VERSION::${ROS_DISTRO_VERSION}
+          echo ::set-output name=ROS_DISTRO_VERSION::${ROS_DISTRO_VERSION}
```
tylerjw, line 89: What does the :: prefix do and why do we need both of these lines?

.github/workflows/ci.yaml
```diff
@@ -66,14 +70,31 @@ jobs:
...
+        env:
+          ROS_DISTRO: ${{ matrix.env.ROS_DISTRO }}
+        run: |
+          latest-release-date() {
+              echo "$(curl -sL "https://api.github.com/repos/ros/rosdistro/tags?per_page=100" \
+              | grep -o "$1/[0-9]\{4\}-[0-9]\{2\}-[0-9]\{2\}" \
+              | cut -d'/' -f2 \
+              | uniq \
+              | sort -t- -k1n -k2n -k3n -z | head -n 1)"
+          }
```
tylerjw, line 85: Should put this in another repo as @rhaschke did with the lcov one below?  Then we could easily reuse this in all our CI jobs.

.github/workflows/ci.yaml
JafarAbdi: The first one is just a debugging message, the second one for setting a step output value

tylerjw (approved):
I like this much better.  Thank you for moving the code into your own action.

vatanaksoytezer (approved):
That looks good to me, one quick question: how do we verify this resolves our problem? If we cannot it seems moveit2_tutorials Foxy CI seems broken the same way and we can use it to verify it fixes our issue.

PR #955, opened 2022-01-04
:test_tube: testspace

tylerjw:
### Description

This adds uploading test results to testspace.  Before this will work an owner of the github org will have to add the integration first.  Here is an example of the results in my bio_ik fork: https://tylerjw.testspace.com/spaces/159418/current


vatanaksoytezer (approved):
I think this is a very nice tool to visualize the tests, thanks Tyler.

PR #956, opened 2022-01-04
Servo: re-order velocity limit check & minor cleanup

AndyZe:
This is a simplified version of #886. It removes the latency compensation parameter that was controversial with reviewers for some reason (despite fixing some hardware issues and being backed as similar to what the robot manufacturers use  :/  ). By splitting it out, I hope the review is faster.

The various fixes here are:

- Re-order the checking of velocity limits. These limits should be checked **after** the other operations that affect velocity. Otherwise they aren't accurate.

- Delete a block of code that was accidentally duplicated.

- Clean up some comments.

- Delete a couple unused things.

You can test this PR with the tutorial:  https://moveit.picknik.ai/main/doc/realtime_servo/realtime_servo_tutorial.html

If you're familiar with the tutorial, you'll notice that motion is faster.

### DIFF
--- a/moveit_ros/moveit_servo/config/panda_simulated_config.yaml
+++ b/moveit_ros/moveit_servo/config/panda_simulated_config.yaml
@@ -7,9 +7,9 @@ use_gazebo: false # Whether the robot is started in a Gazebo simulation environm
 command_in_type: "unitless" # "unitless"> in the range [-1:1], as if from joystick. "speed_units"> cmds are in m/s and rad/s
 scale:
   # Scale parameters are only used if command_in_type=="unitless"
-  linear:  0.4  # Max linear velocity. Meters per publish_period. Unit is [m/s]. Only used for Cartesian commands.
-  rotational:  0.8 # Max angular velocity. Rads per publish_period. Unit is [rad/s]. Only used for Cartesian commands.
-  # Max joint angular/linear velocity. Rads or Meters per publish period. Only used for joint commands on joint_command_in_topic.
+  linear:  0.4  # Max linear velocity. Unit is [m/s]. Only used for Cartesian commands.
+  rotational:  0.8 # Max angular velocity. Unit is [rad/s]. Only used for Cartesian commands.
+  # Max joint angular/linear velocity. Only used for joint commands on joint_command_in_topic.
   joint: 0.5
 
 ## Properties of outgoing commands
--- a/moveit_ros/moveit_servo/include/moveit_servo/enforce_limits.hpp
+++ b/moveit_ros/moveit_servo/include/moveit_servo/enforce_limits.hpp
@@ -46,15 +46,11 @@
 namespace moveit_servo
 {
 /**
- * @brief      Scale the delta theta to match joint velocity/acceleration limits
- *
- * @param[in]  joint_model_group   The joint model group
- * @param[in]  publish_period      The publish period of servo (used for calculating joint velocity)
- * @param[in]  delta_theta         The delta theta input
- *
- * @return     The delta theta output
+ * @brief Decrease robot position change and velocity, if needed, to satisfy joint velocity limits
+ * @param joint_model_group Active joint group. Used to retrieve limits.
+ * @param joint_state The command that will go to the robot.
  */
-Eigen::ArrayXd enforceVelocityLimits(const moveit::core::JointModelGroup* joint_model_group, double publish_period,
-                                     const Eigen::ArrayXd& delta_theta);
+void enforceVelocityLimits(const moveit::core::JointModelGroup* joint_model_group, const double publish_period,
+                           sensor_msgs::msg::JointState& joint_state);
 
 }  // namespace moveit_servo
--- a/moveit_ros/moveit_servo/src/enforce_limits.cpp
+++ b/moveit_ros/moveit_servo/src/enforce_limits.cpp
@@ -47,7 +47,7 @@ namespace moveit_servo
 {
 namespace
 {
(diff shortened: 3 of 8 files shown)


moveit_ros/moveit_servo/config/panda_simulated_config.yaml
```diff
@@ -7,9 +7,9 @@ use_gazebo: false # Whether the robot is started in a Gazebo simulation environm
 command_in_type: "unitless" # "unitless"> in the range [-1:1], as if from joystick. "speed_units"> cmds are in m/s and rad/s
 scale:
   # Scale parameters are only used if command_in_type=="unitless"
-  linear:  0.4  # Max linear velocity. Meters per publish_period. Unit is [m/s]. Only used for Cartesian commands.
```
AndyZe, line 10: this simply wasn't accurate. It's left from the initial implementation years ago

moveit_ros/moveit_servo/include/moveit_servo/enforce_limits.hpp
```diff
@@ -46,15 +46,11 @@
...
- *
- * @return     The delta theta output
+ * @brief Decrease robot position change and velocity, if needed, to satisfy joint velocity limits
+ * @param joint_model_group Active joint group. Used to retrieve limits.
+ * @param joint_state The command that will go to the robot.
  */
-Eigen::ArrayXd enforceVelocityLimits(const moveit::core::JointModelGroup* joint_model_group, double publish_period,
-                                     const Eigen::ArrayXd& delta_theta);
+void enforceVelocityLimits(const moveit::core::JointModelGroup* joint_model_group, const double publish_period,
+                           sensor_msgs::msg::JointState& joint_state);
```
AndyZe, line 54: I had to refactor this because the order of velocity limit application was modified. No change in functionality.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -200,25 +195,6 @@ ServoCalcs::ServoCalcs(rclcpp::Node::SharedPtr node,
     std::exit(EXIT_FAILURE);
   }
 
-  // Load the smoothing plugin
```
AndyZe, line 203: This was duplicated in the block above

moveit_ros/moveit_servo/src/enforce_limits.cpp
```diff
@@ -69,17 +69,27 @@ double getVelocityScalingFactor(const moveit::core::JointModelGroup* joint_model
...
+  {
+    Eigen::VectorXd velocity_residuals = (1 - velocity_scaling_factor) * velocity;
+    Eigen::VectorXd positions =
+        Eigen::Map<Eigen::VectorXd, Eigen::Unaligned>(joint_state.position.data(), joint_state.position.size());
+    positions -= velocity_residuals * publish_period;
+
+    velocity *= velocity_scaling_factor;
+    // Back to sensor_msgs type
+    joint_state.velocity = std::vector<double>(velocity.data(), velocity.data() + velocity.size());
+    joint_state.position = std::vector<double>(positions.data(), positions.data() + positions.size());
```
henningkayser, line 91: It looks like the positions haven't been updated before your fixes. Is that accurate? Modifying the positions with the velocity residuals makes sense to me. I'm only wondering if the resulting position delta is really the product of the computed velocity, considering the smoothing step before velocity enforcement.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -638,13 +611,16 @@ bool ServoCalcs::internalServoUpdate(Eigen::ArrayXd& delta_theta,
   }
   delta_theta *= collision_scale;
 
-  // Loop through joints and update them, calculate velocities, and filter
-  if (!applyJointUpdate(delta_theta, internal_joint_state_, prev_joint_velocity_))
+  // Loop thru joints and update them, calculate velocities, and filter
```
henningkayser, line 614: til: thru is an actual word

henningkayser (commented):
I don't see anything wrong with these changes, will approve after my questions are answered. Could you give this PR a more meaningful name so that the current one doesn't end up as a commit message?

moveit_ros/moveit_servo/src/servo_calcs.cpp
AndyZe: Ha!

moveit_ros/moveit_servo/src/enforce_limits.cpp
AndyZe: The positions have been updated prior to this. You can see that in ServoCalcs::internalServoUpdate() --> applyJointUpdate().

The order is:

- apply velocity scaling for collisions or singularities (based on a position command)
- low-pass filter the position command in `applyJointUpdate()`
- calculate velocities in `applyJointUpdate()`
- apply velocity limits
- apply position limits  (I think this is a higher priority than velocity limits, so check it last)

So I think the answer to your question is, it's OK. All calculations and smoothing are done in the position domain, then velocities are calculated and checked at the end.

AndyZe: I'll add a comment about this OOP to the code.

henningkayser: thanks!

henningkayser (approved):

tylerjw (approved):
I read through these and they look like good changes.  Thank you!

PR #957, opened 2022-01-04
Fix boost linking errors for Windows

Ace314159:
### Description

For Windows, #900 isn't enough to fix the boost linking errors and similar fixes were needed in `background_processing` and `profiler`. This also needs to be backported to foxy.

### Checklist


vatanaksoytezer (approved):

JafarAbdi (changes requested):
We shouldn't need an explicit link against individual targets in Boost since we already have Boost in ament_target_dependencies

I'm on on my phone right now so I can't push the fix will do it later today

Ace314159:
The issue seems to arise because when doing `find_package(geometric_shapes)` it calls `find_package(Boost ...)` again with different and fewer components than what `moveit_core` initially calls it with, causing `Boost_LIBRARIES` to be overwritten. Then, `ament_target_dependencies` can't properly link all the Boost components.

I wasn't sure how to prevent `Boost_LIBRARIES` from being overwritten, so I chose to manually link. I was able to find [this thread](https://cmake.org/pipermail/cmake/2009-December/034099.html) detailing a similar issue, but according to them, just maintaining the internal Boost variables should be sufficient to avoid `Boost_LIBRARIES`, but that doesn't seem to be the case anymore from what I observed.

If you were were able to find a better fix, I'd be happy to use that instead.

JafarAbdi:
> The issue seems to arise because when doing `find_package(geometric_shapes)` it calls `find_package(Boost ...)` again with different and fewer components than what `moveit_core` initially calls it with, causing `Boost_LIBRARIES` to be overwritten. Then, `ament_target_dependencies` can't properly link all the Boost components.
> 
> I wasn't sure how to prevent `Boost_LIBRARIES` from being overwritten, so I chose to manually link. I was able to find [this thread](https://cmake.org/pipermail/cmake/2009-December/034099.html) detailing a similar issue, but according to them, just maintaining the internal Boost variables should be sufficient to avoid `Boost_LIBRARIES`, but that doesn't seem to be the case anymore from what I observed.
> 
> If you are able to find a better fix, I'd be happy to use that instead.

Thanks for the detailed explanation and sorry for the late response, we discussed this last moveit stand-up and we're not sure what the best solution is, we either need to 
1- run [include(ConfigExtras.cmake)](https://github.com/ros-planning/moveit2/blob/main/moveit_core/CMakeLists.txt#L13-L14) after calling all `find_packages(...)` and rely on ament_* to link (we need to make sure no call to `find_packages(...)` depend on Boost)
2- Remove Boost from `ament_target_dependencies` calls and manually specify each boost component like what's done in this PR but we have to do it throughout the whole codebase

Personally, I prefer the first solution since it takes advantage of the ament functionalities

Ace314159:
The first solution sounds good to me too. I'll make the necessary changes.

moveit_core/CMakeLists.txt
```diff
@@ -52,6 +49,9 @@ find_package(pluginlib REQUIRED)
 # TODO: Port python bindings
 # find_package(pybind11 REQUIRED)
 
+# Finds Boost Components
+include(ConfigExtras.cmake)
+
```
tylerjw, line 54: It looks like you just moved this line?

Ace314159: Sorry, I forgot to revert the previous commit as well. Is that what you were implying was missing?

tylerjw: Further up in this file you removed these two lines and then added them back here.  I'm confused about what difference that would make.

Ace314159: As per what Jafar said, the first solution involves moving the `include(ConfigExtras.cmake)` line after all the `find_pakage` calls. When we do this, the `include(ConfigExtras.cmake)` will be the last one to call `find_package(Boost ...)`, so `Boost_LIBRARIES` will not be overwritten by any dependencies, allowing us to use ament_* to link.

That's what I did when I removed those two lines from above and added them back there.

tylerjw (approved):

moveit_core/CMakeLists.txt
tylerjw: Thank you for the explination.  I misunderstood what this would do.

JafarAbdi:
@Ace314159 Super sorry for the delay, do you mind testing this PR ? I'll merge then immediately

moveit_ros/planning_interface/CMakeLists.txt
```diff
@@ -79,7 +79,6 @@ add_subdirectory(move_group_interface)
 if(BUILD_TESTING)
   find_package(ament_cmake_gtest REQUIRED)
   find_package(ros_testing REQUIRED)
-  find_package(Boost REQUIRED COMPONENTS filesystem)
```
JafarAbdi, line 82: This component is already in `ConfigExtras.cmake`

JafarAbdi (approved):

vatanaksoytezer:
If we are reverting #900. We should also revert #925 as well.

Ace314159:
@JafarAbdi This looks good on Windows.

tylerjw (approved):

Issue #841, opened 2021-12-01
Replace NULL with nullptr


tylerjw:
https://github.com/cpp-best-practices/cppbestpractices/blob/master/03-Style.md#use-nullptr

PR #961, opened 2022-01-04
Replaced NULL with nullptr

stephanie-eng:
### Description

Replace NULL with nullptr throughout the project as per #841


JafarAbdi (approved):

henningkayser (approved):

henningkayser:
@stephanie-eng I just noticed that there are still plenty of "NULL" mentions in the comments. Shouldn't those be updated as well for consistency?

PR #972, opened 2022-01-07
Remove geometric_shapes from repos file

tylerjw:
### Description

I don't think we need to build geometric_shapes from source so this PR removes it from the repos file.

### DIFF
--- a/moveit2.repos
+++ b/moveit2.repos
@@ -1,8 +1,4 @@
 repositories:
-  geometric_shapes:
-    type: git
-    url: https://github.com/ros-planning/geometric_shapes
-    version: 2.1.2
   moveit_msgs:
     type: git
     url: https://github.com/ros-planning/moveit_msgs


henningkayser:
are you sure that we should get rid of the repos files altogether? How do we test against upstream changes, then? I think there is value in knowing that the source versions of our repos work well together. How about we add a full source build but keep it optional or even disabled for PRs

tylerjw:
> are you sure that we should get rid of the repos files altogether? How do we test against upstream changes, then? I think there is value in knowing that the source versions of our repos work well together. How about we add a full source build but keep it optional or even disabled for PRs

When we make changes to those other repos that we then need to update moveit to depend on those changes we add it back.  I think maybe we should add moveit as a downstream repo to those other repos if we want to test that moveit builds with changes to them but I think we probably want to allow breaking merges into those other repos first (and then update moveit to point at the new versions of those repos).

henningkayser:
> > are you sure that we should get rid of the repos files altogether? How do we test against upstream changes, then? I think there is value in knowing that the source versions of our repos work well together. How about we add a full source build but keep it optional or even disabled for PRs
> 
> When we make changes to those other repos that we then need to update moveit to depend on those changes we add it back. I think maybe we should add moveit as a downstream repo to those other repos if we want to test that moveit builds with changes to them but I think we probably want to allow breaking merges into those other repos first (and then update moveit to point at the new versions of those repos).

I like the idea of doing downstream tests in the other repos, this should bring down the average CI time a lot. I just think it could cause a lot of confusion if the repos files are sometimes there and sometimes completely missing. Syncing changes across repos is a different issue but it probably doesn't play a role for the decision to temporarily remove the repos files

tylerjw:
> > > are you sure that we should get rid of the repos files altogether? How do we test against upstream changes, then? I think there is value in knowing that the source versions of our repos work well together. How about we add a full source build but keep it optional or even disabled for PRs
> > 
> > 
> > When we make changes to those other repos that we then need to update moveit to depend on those changes we add it back. I think maybe we should add moveit as a downstream repo to those other repos if we want to test that moveit builds with changes to them but I think we probably want to allow breaking merges into those other repos first (and then update moveit to point at the new versions of those repos).
> 
> I like the idea of doing downstream tests in the other repos, this should bring down the average CI time a lot. I just think it could cause a lot of confusion if the repos files are sometimes there and sometimes completely missing. Syncing changes across repos is a different issue but it probably doesn't play a role for the decision to temporarily remove the repos files

What I was saying is that we could either add downstream testing of moveit to those other repos or allow breaking changes in them (that have corresponding PRs to moveit that add those repos back to the upstream repos file here).  The second option of adding it back here after breaking changes is the right answer because users of these repos then have to check out less repos into their workspace normally to use moveit from the source and we build fewer things we don't have to in CI.

henningkayser:
redundant with #1026

PR #975, opened 2022-01-07
Update repos file on foxy

tylerjw:
### Description

The repos file is not needed for foxy.  All released versions are acceptable.  This PR removes the file and adapts the Dockerfiles to handle it not existing.


moveit2.repos
```diff
@@ -1,33 +0,0 @@
-repositories:
```
henningkayser, line 1: Instead of deleting, I think it's fine keeping an empty or commented file

tylerjw: I'll put it back all commented out.

tylerjw: Commenting out the whole file causes vcs to return non-zero and give this output:
```
Input data is not valid format: 'NoneType' object is not subscribable
```

The way I did it here it still returns 0, it just prints this warning:
```
List of repositories is empty
```

henningkayser: works for me

vatanaksoytezer: It also looks good to me. I agree with Henning that we might want to use .repos in the future, if we want to merge a change depending on our upstream.

vatanaksoytezer (approved):

henningkayser (approved):

PR #980, opened 2022-01-09
[hybrid planning] Add action abortion and test; improve the existing test

AndyZe:
- Add the ability to cancel a hybrid planning action request
- Remove the existing Python integration test. It was very basic; it only checked that certain nodes did not die
- Add a C++ integration test that checks for action completion
- Add a C++ integration test for the new cancelation feature
- Do not continue publishing the same trajectory over and over if the global planner needs to replan
- Fix the clearing of robot trajectory if the trajectory is invalid (use `.clear()` instead of `.empty()`)
- Ensure the initial state of the robot is set properly

@sjahr


AndyZe:
CI is failing due to something related to controller loading. Any ideas, @destogl or @jackcenter ?

```
 [component_container_mt-1] [FATAL] [1641768093.218988192] [moveit_ros.trajectory_execution_manager]: Exception while loading controller manager 'moveit_simple_controller_manager/MoveItSimpleControllerManager': According to the loaded plugin descriptions the class moveit_simple_controller_manager/MoveItSimpleControllerManager with base class type moveit_controller_manager::MoveItControllerManager does not exist. Declared types are 
Error: mponent_container_mt-1] [ERROR] [1641768093.219830987] [moveit_ros.trajectory_execution_manager]: Failed to reload controllers: `controller_manager_` does not exist.
```

tylerjw:
> 

My guess is you are missing a run_depend and it works locally because you have sourced your workspace you are running the test in (which adds all the workspace libraries to your path regardless of what is specified in the package.xml).

tylerjw:
> My guess is you are missing a run_depend and it works locally because you have sourced your workspace you are running the test in (which adds all the workspace libraries to your path regardless of what is specified in the package.xml).  The evidence is that plugin loader doesn't find any types that match that base type string.

AndyZe:
This PR now prevents behavior like this:

local_planner_stuck

moveit_ros/hybrid_planning/test/test_basic_integration.cpp
```diff
@@ -0,0 +1,272 @@
...
+    // Add new collision object as soon as global trajectory is available.
+    global_solution_subscriber_ = node_->create_subscription<moveit_msgs::msg::MotionPlanResponse>(
+        "global_trajectory", rclcpp::SystemDefaultsQoS(),
+        [this](const moveit_msgs::msg::MotionPlanResponse::SharedPtr msg) {});
+
+    RCLCPP_INFO(node_->get_logger(), "Initialize Planning Scene Monitor");
+    tf_buffer_ = std::make_shared<tf2_ros::Buffer>(node_->get_clock());
+
+    planning_scene_monitor_ = std::make_shared<planning_scene_monitor::PlanningSceneMonitor>(
+        node_, "robot_description", tf_buffer_, "planning_scene_monitor");
```
tylerjw, line 95: Passing tf2_ros::Buffer to PlanningSceneMonitor's constructor is deprecated

moveit_ros/hybrid_planning/test/test_basic_integration.cpp
```diff
@@ -0,0 +1,272 @@
...
+    if (!planning_scene_monitor_->getPlanningScene())
+    {
+      RCLCPP_ERROR(node_->get_logger(), "The planning scene was not retrieved!");
+      std::exit(EXIT_FAILURE);
+    }
+    else
+    {
+      planning_scene_monitor_->startStateMonitor();
+      planning_scene_monitor_->providePlanningSceneService();  // let RViz display query PlanningScene
```
tylerjw, line 104: is this needed for your test in some way?

moveit_ros/hybrid_planning/test/test_basic_integration.cpp
```diff
@@ -0,0 +1,272 @@
...
+    moveit_msgs::msg::MotionPlanRequest goal_motion_request;
+
+    moveit::core::robotStateToRobotStateMsg(*robot_state, goal_motion_request.start_state);
+    goal_motion_request.group_name = planning_group;
+    goal_motion_request.num_planning_attempts = 10;
+    goal_motion_request.max_velocity_scaling_factor = 0.1;
+    goal_motion_request.max_acceleration_scaling_factor = 0.1;
+    goal_motion_request.allowed_planning_time = 2.0;
+    goal_motion_request.planner_id = "ompl";
+    goal_motion_request.pipeline_id = "ompl";
```
tylerjw, line 145: use inline lambda initialization, there seems to be no reason this should be mutable

moveit_ros/hybrid_planning/test/test_basic_integration.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+    moveit::core::RobotState goal_state(robot_model);
+    std::vector<double> joint_values = { 0.0, 0.0, 0.0, 0.0, 0.0, 1.571, 0.785 };
+    goal_state.setJointGroupPositions(joint_model_group, joint_values);
```
tylerjw, line 149: again, lambda initalization

moveit_ros/hybrid_planning/test/test_basic_integration.cpp
```diff
@@ -0,0 +1,272 @@
...
+    send_goal_options_.feedback_callback =
+        [this](rclcpp_action::ClientGoalHandle<moveit_msgs::action::HybridPlanner>::SharedPtr /*unused*/,
+               const std::shared_ptr<const moveit_msgs::action::HybridPlanner::Feedback> feedback) {
+          RCLCPP_INFO(node_->get_logger(), feedback->feedback.c_str());
+        };
+  }
+
+  void TearDown() override
+  {
+  }
```
tylerjw, line 199: Does this need to be overriden?

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/include/moveit/hybrid_planning_manager/hybrid_planning_manager.h
```diff
@@ -113,6 +120,9 @@ class HybridPlanningManager : public rclcpp::Node
   // Flag that indicates whether the manager is initialized
   bool initialized_;
 
+  // Flag that indicates hybrid planning has been canceled
+  bool stop_hybrid_planning_;
```
tylerjw, line 124: Is this being used to communicate between concurrent sections of code?  If so, please use a lock-free queue or other message-passing technique instead of shared memory.

Here is a lock-free queue: https://www.boost.org/doc/libs/1_76_0/doc/html/boost/lockfree/spsc_queue.html

For reasoning read this blog post: http://henrikeichenhardt.blogspot.com/2013/06/why-shared-mutable-state-is-root-of-all.html

For examples from other languages see this section from the Rust Book: https://doc.rust-lang.org/book/ch16-02-message-passing.html and this section from the Golang documentation: https://go.dev/doc/effective_go#concurrency

For an alternate solution, you could use strong typing and have the code divided between behavior for executing and not executing and change a type that determines the action being taken.  Then you could use RAII for cleaning up and transitioning between your two states.

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/src/hybrid_planning_manager.cpp
```diff
@@ -45,7 +45,7 @@ namespace moveit::hybrid_planning
 using namespace std::chrono_literals;
 
 HybridPlanningManager::HybridPlanningManager(const rclcpp::NodeOptions& options)
```
tylerjw, line 47: I see we never got rid of the ManagerManager name or the inheriting from rclcpp::Node.   :disappointed:

moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
```diff
@@ -159,8 +159,9 @@ bool LocalPlannerComponent::initialize()
         RCLCPP_INFO(LOGGER, "Received local planning goal request");
         return rclcpp_action::GoalResponse::ACCEPT_AND_EXECUTE;
       },
-      [](const std::shared_ptr<rclcpp_action::ServerGoalHandle<moveit_msgs::action::LocalPlanner>>& /*unused*/) {
+      [this](const std::shared_ptr<rclcpp_action::ServerGoalHandle<moveit_msgs::action::LocalPlanner>>& /*unused*/) {
         RCLCPP_INFO(LOGGER, "Received request to cancel local planning goal");
+        state_ = LocalPlannerState::ABORT;
```
tylerjw, line 164: This looks like communication across threads with shared memory also.  See my previous comment on that.

moveit_ros/hybrid_planning/test/test_basic_integration.cpp
```diff
@@ -0,0 +1,272 @@
...
+namespace moveit_hybrid_planning
+{
+using namespace std::chrono_literals;
+
+class HybridPlanningFixture : public ::testing::Test
+{
+public:
+  void SetUp() override
+  {
+  }
```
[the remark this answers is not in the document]
AndyZe: I guess not. Done.

moveit_ros/hybrid_planning/test/test_basic_integration.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+  HybridPlanningFixture() : node_(std::make_shared<rclcpp::Node>("hybrid_planning_testing"))
```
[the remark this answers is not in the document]
AndyZe: I think you might have read that wrong? It's node subclassing a node.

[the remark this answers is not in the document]
tylerjw: you are right, I did read this wrong.

moveit_ros/hybrid_planning/test/test_basic_integration.cpp
```diff
@@ -55,7 +55,7 @@ using namespace std::chrono_literals;
 class HybridPlanningFixture : public ::testing::Test
 {
 public:
-  void SetUp() override
+  void SetUp()
   {
   }
```
tylerjw, line 60: I meant delete the empty function, not remove the override keyword.

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/src/hybrid_planning_manager.cpp
```diff
@@ -147,13 +147,46 @@ bool HybridPlanningManager::initialize()
...
+        stop_hybrid_planning_ = true;
+
+        // Cancel local action
+        local_planner_action_client_->async_cancel_all_goals();
+
+        // Cancel global action
+        global_planner_action_client_->async_cancel_all_goals();
+
+        RCLCPP_INFO(LOGGER, "Received request to cancel goal");
+        return rclcpp_action::CancelResponse::ACCEPT;
```
tylerjw, line 168: If you want to use the function instead of an inline lambda you can call the function from within the lambda and the compiler will optimize away the lambda.  You can also use `auto` for the type of parameter to the lambda and it should deduce the type if you use it to call another function.

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/src/hybrid_planning_manager.cpp
```diff
@@ -147,13 +147,46 @@ bool HybridPlanningManager::initialize()
...
+        hybrid_planning_goal_handle_ = std::move(goal_handle);
+
+        // react is defined in a hybrid_planning_manager plugin
+        ReactionResult reaction_result =
+            planner_logic_instance_->react(HybridPlanningEvent::HYBRID_PLANNING_REQUEST_RECEIVED);
+        if (reaction_result.error_code.val != moveit_msgs::msg::MoveItErrorCodes::SUCCESS)
+        {
+          auto result = std::make_shared<moveit_msgs::action::HybridPlanner::Result>();
+          result->error_code.val = reaction_result.error_code.val;
+          result->error_message = reaction_result.error_message;
```
tylerjw, line 185: This looks like another thing that could be const if initialized in a lambda.

reference: http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-lambda-init

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/src/hybrid_planning_manager.cpp
```diff
@@ -147,13 +147,46 @@ bool HybridPlanningManager::initialize()
...
+          hybrid_planning_goal_handle_->abort(result);
+          RCLCPP_ERROR(LOGGER, "Hybrid Planning Manager failed to react to  '%s'", reaction_result.event.c_str());
```
tylerjw, line 187: If possible use stream concatination instead of %s.  %s is a known security hole.

moveit_ros/hybrid_planning/test/test_basic_integration.cpp
AndyZe: nope, just copied from the demo

moveit_ros/hybrid_planning/test/test_basic_integration.cpp
```diff
@@ -92,7 +89,7 @@ class HybridPlanningFixture : public ::testing::Test
     tf_buffer_ = std::make_shared<tf2_ros::Buffer>(node_->get_clock());
 
     planning_scene_monitor_ = std::make_shared<planning_scene_monitor::PlanningSceneMonitor>(
-        node_, "robot_description", tf_buffer_, "planning_scene_monitor");
+        node_, "robot_description" /*, tf_buffer_*/, "planning_scene_monitor");
```
tylerjw, line 92: I'm confused, you don't need to pass this parameter.  Why are you just commenting it out?

moveit_ros/hybrid_planning/test/test_basic_integration.cpp
AndyZe: what's the point of this?

moveit_ros/hybrid_planning/test/test_basic_integration.cpp
AndyZe: Still in the middle of changes. Hang tight.

moveit_ros/hybrid_planning/test/test_basic_integration.cpp
AndyZe: ah, to make it const. I see

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/src/hybrid_planning_manager.cpp
AndyZe: that sounds complicated.

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/src/hybrid_planning_manager.cpp
AndyZe: We talked about it and decided it actually did make some sense in this case. I don't want to dig back through the different PR's to find the conversation right now, though

tylerjw: I think we decided we were going to fix this.  I don't remember it being good for any reason.  I think I had a PR that fixed it too.  That being said I won't block this change for fixing this.

tylerjw: I think it only made sense because we didn't have time to fix it at the time you were pushing to merge Hybrid Planner.  I don't remember any other reason but I do expect we will either fix this or document very clearly why this is necessary.

AndyZe: https://github.com/ros-planning/moveit2/pull/488#discussion_r720463753

AndyZe: What do you want to call it then? I'll do a quick find/replace.

tylerjw: The link to the comment didn't work for some reason.  It just opens that giant PR.

I forgot I mentioned the SomethingManager.  I care much less about that than the inheriting from Node.  I don't have a better suggestion for the name at this point but I do know I don't like inheriting from the node object.

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/include/moveit/hybrid_planning_manager/hybrid_planning_manager.h
AndyZe: Don't you need to know how many threads are talking to use spsc queue? I don't know that here. We could probably assume 2 but...

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/src/hybrid_planning_manager.cpp
tylerjw: https://github.com/ros-planning/moveit2/pull/488#discussion_r720408829

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/include/moveit/hybrid_planning_manager/hybrid_planning_manager.h
```diff
@@ -107,7 +107,7 @@ class HybridPlanningManager : public rclcpp::Node
   bool initialized_;
 
   // Flag that indicates hybrid planning has been canceled
-  bool stop_hybrid_planning_;
+  std::atomic<bool> stop_hybrid_planning_;
```
tylerjw, line 110: sadness... raw synchronization primitive and using mutable state

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/include/moveit/hybrid_planning_manager/hybrid_planning_manager.h
tylerjw: yes, you do... you need one producer and one consumer thread or it is not thread safe.  if you have multiple consumers of this you should use a multi-producer multi-consumer queue: https://www.boost.org/doc/libs/1_59_0/doc/html/boost/lockfree/queue.html

tylerjw: If you do have multiple consumers you'll need to know how many to make sure you notify each.  It might be better to have a stack of spsc queues that consumers can register themselves into.

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/include/moveit/hybrid_planning_manager/hybrid_planning_manager.h
tylerjw: You might find this interesting: https://en.wikipedia.org/wiki/Amdahl%27s_law

There are two reasons to remove shared memory between threads.  

The first is that it is a synchronization point (a point in the software that must be single-threaded) and that the more of those you add the worse our software can scale across multiple processors.  See above link.

The second reason is that they are mutable state.  See the C++ seasoning talk for more reasoning on that.

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/include/moveit/hybrid_planning_manager/hybrid_planning_manager.h
AndyZe: So how many producers and consumers do you think there are? I'm not trying to be difficult here, but that number needs to be known. I think it depends on whether a SingleThreadedExecutor or MultiThreadedExecutor is being used. MultiThreadedExecutor [defaults](https://github.com/ros2/rclcpp/issues/441) to match (num. hardware threads), it looks like. None of that stuff is determined in this header file.

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/include/moveit/hybrid_planning_manager/hybrid_planning_manager.h
AndyZe: My understanding is atomics are lock-free since the read or write happens in a single step. So the first bullet point doesn't apply. I'm not sure about the second one.

moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
AndyZe: Agree that this is a problem!

AndyZe: I can make this variable atomic. That's a lock-free synchronization primitive. I'll push this change; let me know if you can think of a better way (hopefully not ridiculously complicated).

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/src/hybrid_planning_manager.cpp
AndyZe: Made an issue re. inheriting from Node here:  https://github.com/ros-planning/moveit2/issues/1013

tylerjw:
Unrelated to my comments on how this should be implemented.  It is still failing in CI.  I'm assuming you don't intend for this test to fail in CI.

moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
tylerjw: Even lock-free synchronization primitives still cause points of synchronization (single-threaded sections of the code).  I would much prefer a method of communication that does not rely on synchronization of memory access between threads.

The primary reason for this is that if you make something an atomic it has to be stored in ram and access to that point in ram has to be synchronized.  This means that for the CPU to modify or read the value it has to go and get it from ram and then make sure it changed it while preventing no other CPU from accessing it at the same time.  lock-free things like atomic use built-in CPU instructions for this instead of mutexes so they are more performant but they are still creating points in the code that have to be executed in a single-threaded way.

tylerjw: My proposal would be to change our mental model for how we build multi-threaded applications to one that does not use mutable state shared between threads for communication.

AndyZe:
Sorry about that. It was passing at one point but seems to be failing due to a ros2_control change:

`[ros2_control_node-5] [WARN] [1642614151.127346724] [fake_generic_system]: The usage of initial_position has been deprecated. Please use 'initial_value' instead.`

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/include/moveit/hybrid_planning_manager/hybrid_planning_manager.h
tylerjw: To make it clear from our discussion today I want to find a better way to communicate between threads than raw synchronization primitives around memory.  That however is a much larger discussion and requires more research than this PR.  Please just make sure you are using some mechanism (atomic, mutex, etc) to synchronize access to any variables you are sharing between two concurrent sections.

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/include/moveit/hybrid_planning_manager/hybrid_planning_manager.h
tylerjw: I don't know what the right answer is in this case.  Another question I have is are these callbacks re-entrant and should we guard against that too.  Another way to ask that question is what would happen if on two separate threads ROS calls one of these callbacks twice at the same time?

For now, go ahead and use some sort of synchronization.  Long term we need a better model for how to approach this than shared memory if we want MoveIt to scale onto many threads well.

tylerjw:
> Sorry about that. It was passing at one point but seems to be failing due to a ros2_control change

You don't need to be sorry, we just need to figure this out before we can merge it.  If I have time I'll look into it.

moveit_ros/hybrid_planning/test/launch/test_basic_integration.test.py
```diff
@@ -0,0 +1,69 @@
...
+    get_robot_description_semantic,
+    load_file,
+    load_yaml,
+)
+
+
+def generate_test_description():
+    # generate_common_hybrid_launch_description() returns a list of nodes to launch
+    common_launch = generate_common_hybrid_launch_description()
+    robot_description = get_robot_description()
```
tylerjw, line 24: I wonder if Jafar's change would be useful here?

tylerjw (approved):
I read through these changes again and it all looks good.  We just need to figure out the ros control issue.

AndyZe:
This moveit_resources PR should fix it:

https://github.com/ros-planning/moveit_resources/pull/110

AndyZe:
After fixing some other issues, I think CI is failing now because we're using your branch of moveit_resources (this commit:  https://github.com/ros-planning/moveit2/commit/8a8249c681a871e4c1e27deb6480418f957cfdb2).

What do you think, @JafarAbdi? Can you at least rebase your fork on the main branch:  https://github.com/ros-planning/moveit_resources/tree/ros2

tylerjw:
I merged the PR of jafar's to fix that.  It looks like we are using the ros2 branch of moveit_resources now: https://github.com/ros-planning/moveit2/blob/b0e2cd0272ae2d99ffb6fd4df0f592caf930c327/moveit2.repos#L12

AndyZe:
Commit 2cc8146  here fixes the warnings in that CI run of main that you triggered, @tylerjw. But there is still another issue after that.

AndyZe:
Started a new PR here to focus on this issue:

https://github.com/ros-planning/moveit2/pull/1025

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/src/hybrid_planning_manager.cpp
```diff
@@ -147,16 +147,46 @@ bool HybridPlanningManager::initialize()
...
       [](const rclcpp_action::GoalUUID& /*unused*/,
          std::shared_ptr<const moveit_msgs::action::HybridPlanner::Goal> /*unused*/) {
         RCLCPP_INFO(LOGGER, "Received goal request");
         return rclcpp_action::GoalResponse::ACCEPT_AND_EXECUTE;
       },
-      [](const std::shared_ptr<rclcpp_action::ServerGoalHandle<moveit_msgs::action::HybridPlanner>>& /*unused*/) {
+      // Cancel callback
+      [this](std::shared_ptr<rclcpp_action::ServerGoalHandle<moveit_msgs::action::HybridPlanner>> /*unused*/) {
+        // Prevent any new global or local requests from going out
+        stop_hybrid_planning_ = true;
```
henningkayser, line 159: Just saw this now... Shouldn't this use the API of the ServerGoalHandle to flag the goal as canceled, so that the affected code can use `ServerGoalHandle::is_canceling()`? https://docs.ros2.org/latest/api/rclcpp_action/classrclcpp__action_1_1ServerGoalHandle.html

henningkayser (approved):
I think this is a very good addition, especially with the tests. I haven't tested this locally, but the logic seems very reasonable to me. I'm just not sure if we are fully using the rclcpp_action API as intended.

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/src/hybrid_planning_manager.cpp
AndyZe: If I understand, you would like me to add something like this, here in the cancel callback:

        `hybrid_planning_goal_handle->canceled();`
        
Looking at the [tutorial](https://docs.ros.org/en/foxy/Tutorials/Actions/Writing-a-Cpp-Action-Server-Client.html), they do not use canceled() within the cancel callback. My guess is, `goal->canceled()` would be redundant when used like this.

AndyZe: I can follow up with a new PR if I missed something here.

PR #981, opened 2022-01-10
:heavy_plus_sign: ament_cmake_gtest dependency

tylerjw:
### Description

I don't understand why the pre-release didn't catch this error.  Thank you @jspricke for helping us find this.

Here is the build-farm error this should fix: https://build.ros2.org/job/Gbin_uF64__pilz_industrial_motion_planner__ubuntu_focal_amd64__binary/9/console


jspricke (approved):

vatanaksoytezer (approved):

henningkayser:
> I don't understand why the pre-release didn't catch this error. Thank you @jspricke for helping us find this.

The prerelease job is not installing dependencies for each package in an isolated environment, right?

henningkayser (approved):

PR #983, opened 2022-01-10
Add ABI check to Foxy

vatanaksoytezer:


tylerjw (approved):

henningkayser:
@vatanaksoytezer should we really merge this? The build job took over 3 hours. Is there any way to reduce this by only checking the diff or so?

vatanaksoytezer:
> @vatanaksoytezer should we really merge this? The build job took over 3 hours. Is there any way to reduce this by only checking the diff or so?

I should investigate. Can we re-run the build, caching might help this significantly.

henningkayser:
@vatanaksoytezer should we port this to Galactic as well?

vatanaksoytezer:
> @vatanaksoytezer should we port this to Galactic as well? It's still 2.5h, btw

Yes, but it's still taking more than 2.5 hours, should we investigate or we are fine with this?

henningkayser:
I converted this to draft, we need to reduce the build time before getting this merged

tylerjw:
Created issue for tracking resuming this. Closing for now.

PR #988, opened 2022-01-13
Remove __has_includes from main, remove galactic CI, remove IKFast Testing

vatanaksoytezer:
Since we branched off for galactic, we can now remove __has_includes and remove galactic CI jobs. If we want to keep the CI jobs we can delay this PR. I also removed IKFast because we are not testing it and just installing it dependencies for no reason.


vatanaksoytezer:
As we discussed in the meeting I am going to make this draft, and we merge when we are ready to remove Galactic CI from main (possibly when there are more breakages on Rolling)

henningkayser:
@vatanaksoytezer since we removed the galactic jobs from main, could you update this PR please?

PR #989, opened 2022-01-13
Add a "target_pose_lookahead_time" parameter

### LINKED ISSUE #696: Consider a "feedback delay" parameter for Servo

AndyZe:
This work was started in #886. Closes #696.

The new parameter helps account for latency in the control loop. This can compensate for various issues or delays in the control loop that are hard to quantify, including:

 - joint_states publication rate
 - ROS message transmission. I believe there are 2 message hops
 - network latency
 - a little bit of processing time
 - whatever happens in the ros_control interface
 - low-level controller ramp up/ramp down time

### DIFF
--- a/moveit_ros/moveit_servo/src/servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/src/servo_calcs.cpp
@@ -989,6 +989,11 @@ Eigen::VectorXd ServoCalcs::scaleCartesianCommand(const geometry_msgs::msg::Twis
   Eigen::VectorXd result(6);
   result.setZero();  // Or the else case below leads to misery
 
+  // Add a user-defined, constant delay to the timestep.
+  // This can help if ramp-up / ramp-down of the low-level controllers cause jitter, or to account for network latency.
+  // Effectively it moves the target pose farther ahead.
+  double timestep = parameters_->publish_period + parameters_->target_pose_lookahead_time;
+
   // Apply user-defined scaling if inputs are unitless [-1:1]
   if (parameters_->command_in_type == "unitless")
   {
@@ -1022,6 +1027,11 @@ Eigen::VectorXd ServoCalcs::scaleJointCommand(const control_msgs::msg::JointJog&
 {
   Eigen::VectorXd result(num_joints_);
   result.setZero();
+
+  // Add a user-defined, constant delay to the timestep.
+  // This can help if ramp-up / ramp-down of the low-level controllers cause jitter, or to account for network latency.
+  // Effectively it moves the target pose farther ahead.
+  double timestep = parameters_->publish_period + parameters_->target_pose_lookahead_time;
 
   std::size_t c;
   for (std::size_t m = 0; m < command.joint_names.size(); ++m)
--- a/moveit_ros/moveit_servo/src/servo_parameters.cpp
+++ b/moveit_ros/moveit_servo/src/servo_parameters.cpp
@@ -134,6 +134,12 @@ void ServoParameters::declare(const std::string& ns,
                                          .type(PARAMETER_DOUBLE)
                                          .description("Max joint angular/linear velocity. Only used for joint "
                                                       "commands on joint_command_in_topic."));
+  node_parameters->declare_parameter(
+      ns + ".target_pose_lookahead_time", ParameterValue{ parameters.target_pose_lookahead_time },
+      ParameterDescriptorBuilder{}
+          .type(PARAMETER_DOUBLE)
+          .description("An optional parameter to smooth jitter due to latency in the system "
+                       "or low-level controller ramp up / ramp down"));
 
   // Properties of outgoing commands
   node_parameters->declare_parameter(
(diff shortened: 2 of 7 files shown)


AndyZe:
@livanov93

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -989,6 +989,11 @@ Eigen::VectorXd ServoCalcs::scaleCartesianCommand(const geometry_msgs::msg::Twis
   Eigen::VectorXd result(6);
   result.setZero();  // Or the else case below leads to misery
 
+  // Add a user-defined, constant delay to the timestep.
+  // This can help if ramp-up / ramp-down of the low-level controllers cause jitter, or to account for network latency.
+  // Effectively it moves the target pose farther ahead.
+  double timestep = parameters_->publish_period + parameters_->target_pose_lookahead_time;
```
jliukkonen, line 995: Where's the `timestep` being used? I can be marked as `const`.

moveit_ros/moveit_servo/src/servo_parameters.cpp
```diff
@@ -409,6 +417,12 @@ std::optional<ServoParameters> ServoParameters::validate(ServoParameters paramet
                         "greater than zero. Check yaml file.");
     return std::nullopt;
   }
+  if (parameters.target_pose_lookahead_time < 0 || parameters.target_pose_lookahead_time > 0.1)
+  {
+    RCLCPP_WARN(LOGGER,
+                "Parameter 'target_pose_lookahead_time' should be greater than zero and typically less than 0.1s.");
```
jliukkonen, line 423: I would consider putting this to the config files instead as a comment for the parameter. That way, if someone truly wants to use some high value, there wouldn't be warnings but in all likelihood the user still knows that it is not a great idea.

jliukkonen (commented):
Just to make sure I understand this correctly. The lookahead time is used to compensate for small delays that result in effectively a lower publish rate than what is requested of the system. Meaning that any non-zero lookahead time is effectively a tiny velocity boost?

moveit_ros/moveit_servo/src/servo_calcs.cpp
livanov93: I agree with @jliukkonen.
Instead of [this block](https://github.com/AndyZe/moveit2/blob/7fef7faefdac357da44ff01ceaa15838919a7ae7/moveit_ros/moveit_servo/src/servo_calcs.cpp#L1034-L1054):
Consider this:
```
  const double timestep = parameters_->publish_period + parameters_->target_pose_lookahead_time;

  std::size_t c;
  for (std::size_t m = 0; m < command.joint_names.size(); ++m)
  {
    try
    {
      c = joint_state_name_map_.at(command.joint_names[m]);
    }
    catch (const std::out_of_range& e)
    {
      rclcpp::Clock& clock = *node_->get_clock();
      RCLCPP_WARN_STREAM_THROTTLE(LOGGER, clock, ROS_LOG_THROTTLE_PERIOD, "Ignoring joint " << command.joint_names[m]);
      continue;
    }
    // Apply user-defined scaling if inputs are unitless [-1:1]
    if (parameters_->command_in_type == "unitless")
      result[c] = command.velocities[m] * parameters_->joint_scale * timestep;
    // Otherwise, commands are in m/s and rad/s
    else if (parameters_->command_in_type == "speed_units")
      result[c] = command.velocities[m] *timestep;
```

And instead of [this block](https://github.com/AndyZe/moveit2/blob/7fef7faefdac357da44ff01ceaa15838919a7ae7/moveit_ros/moveit_servo/src/servo_calcs.cpp#L995-L1015)

Consider using following:
```
  const double timestep = parameters_->publish_period + parameters_->target_pose_lookahead_time;

  // Apply user-defined scaling if inputs are unitless [-1:1]
  if (parameters_->command_in_type == "unitless")
  {
    result[0] = parameters_->linear_scale *timestep * command.twist.linear.x;
    result[1] = parameters_->linear_scale * timestep * command.twist.linear.y;
    result[2] = parameters_->linear_scale * timestep * command.twist.linear.z;
    result[3] = parameters_->rotational_scale * timestep * command.twist.angular.x;
    result[4] = parameters_->rotational_scale *timestep * command.twist.angular.y;
    result[5] = parameters_->rotational_scale * timestep * command.twist.angular.z;
  }
  // Otherwise, commands are in m/s and rad/s
  else if (parameters_->command_in_type == "speed_units")
  {
    result[0] = command.twist.linear.x * timestep;
    result[1] = command.twist.linear.y * timestep;
    result[2] = command.twist.linear.z *timestep;
    result[3] = command.twist.angular.x * timestep;
    result[4] = command.twist.angular.y * timestep;
    result[5] = command.twist.angular.z *timestep;

livanov93 (changes requested):
We need to update use of new `timestep`. Also the warning about the look ahead pararamter should be somehow evaluated using the servo publish period.

AndyZe:
It could still be useful. It's stuck somewhere between "potentially useful" and "useful enough that I want to merge it now".

tylerjw:
@AndyZe if you rebase this and it passes CI, I'll approve it. What do you think?

AndyZe:
I'll just close this since it's not clear if it would be useful. Let's remember this in the future in case Servo is laggy, though

PR #205, opened 2020-06-03
Port Ikfast kinematics solver

JafarAbdi:
### Description

eigen_conversion isn't ported yet to ROS2 so I copied the needed function and dropped its dependency

I tested it with panda robot steps are as follow [Source](https://ros-planning.github.io/moveit_tutorials/doc/ikfast/ikfast_tutorial.html)
1- clone https://github.com/JafarAbdi/panda_moveit_config/tree/ros2
2- cd moveit_resources/panda_description/urdf
3- ros2 run moveit_kinematics auto_create_ikfast_moveit_plugin.sh panda.urdf panda_arm panda_link0 panda_link8
4- now move the generated package to the main source path and recompile (the generated package name is panda_panda_arm_ikfast_plugin)

### Checklist
- [ ] **Required by CI**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)
- [ ] Extend the tutorials / documentation [reference](http://moveit.ros.org/documentation/contributing/)
- [ ] Document API changes relevant to the user in the MIGRATION.md notes
- [ ] Create tests, which fail without this PR [reference](https://ros-planning.github.io/moveit_tutorials/)
- [ ] Include a screenshot if changing a GUI
- [ ] While waiting for someone to review your request, please help review [another open pull request](https://github.com/ros-planning/moveit/pulls) to support the maintainers

[//]: # "You can expect a response from a maintainer within 7 days. If you haven't heard anything by then, feel free to ping the thread. Thank you!"

PR #991, opened 2022-01-14
Remove IKFast CI

vatanaksoytezer:
This is not even ported to ROS2 and we are using some extra time to build dependencies. This is also causing the current CI failures.


tylerjw (approved):

JafarAbdi (approved):

tylerjw:
@henningkayser will have to remove the ikfast test name as a required test before we can merge this.

henningkayser:
Wait a second. It has been ported here https://github.com/ros-planning/moveit2/pull/205. Afaik it has been functional, also. Shouldn't we just fix the test?

PR #998, opened 2022-01-17
Delete profiler

tylerjw:
### Description

MoveIt does not need a profiling tool.  With optimizations enabled from your compiler, it might produce misleading results and/or disable some possible optimizations.  For profiling use one of the many great existing profiling tools.  Here are some examples:

- gprof
- callgrind
- gperftools
- VTune
- TAU


henningkayser (approved):
Any thoughts about the evaluator tools? They don't seem to make a lot of sense without the profiler

tylerjw:
> Any thoughts about the evaluator tools? They don't seem to make a lot of sense without the profiler

I'm happy to remove them.  I do think it would be useful eventually to re-develop them with different tooling.

tylerjw:
I updated this by removing the two evaluator tools that depended on the profiler.

PR #1009, opened 2022-01-18
Output html to a better directory

vatanaksoytezer:
The previous directory was unintuitive and was outputting to one directory below what you are working. This is the default one is much nice to play with CI


tylerjw (approved):
I like this, it is more standard.

PR #1014, opened 2022-01-19
Set controller status before it is cheked on trajectory execution.

destogl:
### Description

This PR enalbes trajectory execution when using simulation.
The PR adds controllers' status before it is checked if all needed controller are activated.

This is neede to use MoveIt with Ignition Simulation (I will link UR example later today for this).
Though, the interesting part is, why trajectory execution is working without this changes when using "FakeSystem" or hardware.
It looks to me that the same checks should fail in any case.

I tested this with and without "moveit_manage_controllers" parameter set. And the changes are required in both cases.

I using "moveit_simple_controller_manager/MoveItSimpleControllerManager" in our setup. Is this maybe an issue?
Should another controller manager be used.

BTW, I see that MoveIs is not activelly checking which controllers are active and which not, although there is a management code for this implemented in `ensureActiveControllers` (line: 1718) of `trajectory_execution_manager.cpp`.

### Checklist

### DIFF
--- a/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
+++ b/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
@@ -569,6 +569,17 @@ void TrajectoryExecutionManager::reloadControllerInformation()
       known_controllers_[ci.name_] = ci;
     }
 
+    names.clear();
+    controller_manager_->getActiveControllers(names);
+    for (std::map<std::string, ControllerInformation>::iterator it = known_controllers_.begin();
+         it != known_controllers_.end(); ++it)
+    {
+      if (std::find(names.begin(), names.end(), it->first) != names.end())
+      {
+        it->second.state_.active_ = true;
+      }
+    }
+
     for (std::map<std::string, ControllerInformation>::iterator it = known_controllers_.begin();
          it != known_controllers_.end(); ++it)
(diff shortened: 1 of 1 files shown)


vatanaksoytezer:
@destogl clang-tidy seems unhappy of your changes, doo you mind re-basing and fixing clang-tidy suggestions?

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -567,6 +567,17 @@ void TrajectoryExecutionManager::reloadControllerInformation()
       known_controllers_[ci.name_] = ci;
     }
 
+    names.clear();
+    controller_manager_->getActiveControllers(names);
+    for (std::map<std::string, ControllerInformation>::iterator it = known_controllers_.begin();
```
henningkayser, line 572: I think with new code we should aim to use range-based loops or std algorithms. Also, is there a chance that `getActiveControllers()` returns names for which we don't have a known controller? If no, iterating names and accessing the map directly would make more sense to me

henningkayser (approved):

JafarAbdi (approved):

destogl:
can we merge this? I think it has all approvals and it is up-to-date

destogl:
@mergifyio backport galactic

PR #1018, opened 2022-01-21
Remove unused parameters.

corycrean:
- Remove unused parameters.
- Resolve additional clang-tidy warnings.

### Description

CI for https://github.com/ros-planning/moveit2/pull/999 was failing because of unused parameter warnings from clang-tidy.  This PR resolves those warnings.

### Checklist


JafarAbdi (approved):

moveit_ros/benchmarks/src/BenchmarkOptions.cpp
```diff
@@ -275,18 +275,22 @@ void BenchmarkOptions::readPlannerConfigs(const rclcpp::Node::SharedPtr& node)
   for (const std::string& pipeline : pipelines)
   {
     std::string pipeline_name;
-    if (!node->get_parameter(np + "." + pipeline + ".name", pipeline_name))
+    std::string pipeline_parameter_name(np);
+    pipeline_parameter_name.append(".");
+    pipeline_parameter_name.append(pipeline);
+    pipeline_parameter_name.append(".name");
```
tylerjw, line 281: Minor nit, this string could be a constant if you constructed it on one line.

```suggestion
    const std::string pipeline_parameter_name = np + '.' + pipeline + ".name";
```

tylerjw (approved):
One tiny nit.  Thank you for fixing these clang-tidy issues.

vatanaksoytezer (approved):

moveit_ros/benchmarks/src/BenchmarkOptions.cpp
corycrean: That's true, but then clang-tidy complains:
```
/root/target_ws/src/moveit2/moveit_ros/benchmarks/src/BenchmarkOptions.cpp:278:58: warning: string concatenation results in allocation of unnecessary temporary strings; consider using 'operator+=' or 'string::append()' instead [performance-inefficient-string-concatenation]
    const std::string pipeline_parameter_name = np + "." + pipeline + ".name";
                                                         ^
```

I could use a lambda to build the string the way I'm doing it here, and still have `pipeline_parameter_name` be const, but I don't know if that would give better performance than your solution (which seems to be clang-tidy's complaint).  What do you think?

tylerjw:
```suggestion
    constexpr auto pipeline_parameter_name = std::string(np).append('.').append(pipeline).append(".name");
```

Maybe that would quiet clang-tidy while still making the variable const without needing the lambda initialization pattern?

tylerjw: I don't actually know if you can make that constexpr or if it just can be const, but if you can try making it constexpr.

corycrean: That worked (though I couldn't make it constexpr, only const).

tylerjw (approved):

PR #1025, opened 2022-01-23
Fix CI by building ros2_control from source

AndyZe:
This moveit_resources PR has screwed up the parsing of initial_positions and initial_velocities for ros2_control.

https://github.com/ros-planning/moveit_resources/commit/fbc66ccd4b28b0ab740cf8572271c97471aef26e

### DIFF
--- a/moveit2.repos
+++ b/moveit2.repos
@@ -9,8 +9,8 @@ repositories:
     version: 2.2.0
   moveit_resources:
     type: git
-    url: https://github.com/ros-planning/moveit_resources
-    version: ros2
+    url: https://github.com/AndyZe/moveit_resources
+    version: andyz/no_initial_velocities
   launch_param_builder:
     type: git
     url: https://github.com/PickNikRobotics/launch_param_builder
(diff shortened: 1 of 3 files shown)


tylerjw:
If possible I'd like to avoid having to add this back to our repos file.  If we do this we can't release again until roscontrol releases.  If possible we should just figure out how to get our tests to work with their released version.

vatanaksoytezer:
Seems their latest release 2.1.0 (released for Rolling) is fine for us and our CI is passing in rolling-testing, but failing in others. We need to do one of the following:

1) Wait for them to backport `initial value `to galactic
2) Don't compile galactic anymore in main
3) Compile main using source ros2_control source

moveit2.repos
```diff
@@ -15,3 +15,7 @@ repositories:
     type: git
     url: https://github.com/PickNikRobotics/launch_param_builder
     version: main
+  ros2_control:
+    type: git
+    url: https://github.com/ros-controls/ros2_control.git
+    version: a362e2bfbe2308291f20cfc0109c4e1fbf11fb50
```
tylerjw, line 21:
```suggestion
    version: 2.1.0
```

tylerjw (approved):

vatanaksoytezer (approved):

PR #1026, opened 2022-01-24
Set fixed versions in repos file

tylerjw:
### Description

This is a test to see if CI will pass with these versions.

@vatanaksoytezer and @AndyZe https://github.com/ros-planning/moveit2/pull/1025#issuecomment-1020030546

To keep galactic working on main we have to build ros2_control from the source.  This is because a new feature was added and released only into rolling.

@henningkayser what do you think of dropping support for galactic on main instead?


AndyZe (approved):

henningkayser:
> @henningkayser what do you think of dropping support for galactic on main instead?

I don't see why we shouldn't...

JafarAbdi (approved):
+1 for dropping galactic support on main

vatanaksoytezer (approved):

vatanaksoytezer:
@henningkayser can we drop the galactic required CI jobs and add `rolling-ci-testing + ikfast + clang-tidy` job to required jobs

tylerjw:
Today I tried putting the old version of ruckig (what is in rolling-main) in our repos file to see if it would cause all tests to fail.  Weirdly this CI run timed out after 6 hours and was canceled by github.  Maybe something about compiling ruckig is not nice to the github runner?

Because I can't get this to complete I'm also trying to test with the latest version of ruckig that was just released and is in the testing repo.

tylerjw:
I updated this to just set the fixed version of the bloomed but not yet synced package from jafar and remove the released with no existing changes packages.

henningkayser (approved):
looks good to me

tylerjw:
This appears to be failing to build because we have some missing dependencies in moveit_core that are masked by building with geometric_shapes and moveit_msgs.  Even if we don't drop these dependencies from our repos file we should add all required dependencies to moveit so we don't have situations like this with transient dependencies.

PR #1029, opened 2022-01-25
Construct JointModelGroup with ranges-v3

tylerjw:
### Description

I know this probably won't pass for the same reason the other PRs aren't passing CI.  I don't understand the issue with the Ruckig test or what is causing it to fail.  While investigating that I discovered this constructor as I was trying to understand how some of these values are initialized.  I believe this is much-much easier to read than the nested for loops.  I should be able to move all of these into the initializer list and out of the body of the constructor too, I'm just done hacking on this tonight.

### DIFF
--- a/moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
+++ b/moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
@@ -428,7 +428,7 @@ public:
   }
 
   /** \brief Get the groups that are subsets of this one (in terms of joints set) */
-  void getSubgroups(std::vector<const JointModelGroup*>& sub_groups) const;
+  std::vector<const JointModelGroup*> getSubgroups() const;
 
   /** \brief Check if the joints of group \e group are a subset of the joints in this group */
   bool isSubgroup(const std::string& group) const
--- a/moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
+++ b/moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
@@ -699,10 +699,10 @@ protected:
   std::set<std::string> updated_link_model_with_geometry_name_set_;
 
   /** \brief The number of variables necessary to describe this group of joints */
-  unsigned int variable_count_;
+  size_t variable_count_;
 
   /** \brief The number of variables necessary to describe the active joints in this group of joints */
-  unsigned int active_variable_count_;
+  size_t active_variable_count_;
 
   /** \brief True if the state of this group is contiguous within the full robot state; this also means that
       the index values in variable_index_list_ are consecutive integers */
--- a/moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
+++ b/moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
@@ -282,7 +282,7 @@ public:
   }
 
   /** \brief Get the index locations in the complete robot state for all the variables in this group */
-  const std::vector<int>& getVariableIndexList() const
+  const std::vector<size_t>& getVariableIndexList() const
   {
     return variable_index_list_;
   }
@@ -644,11 +644,11 @@ protected:
 
   /** \brief The list of index values this group includes, with respect to a full robot state; this includes mimic
    * joints. */
-  std::vector<int> variable_index_list_;
+  std::vector<size_t> variable_index_list_;
 
   /** \brief For each active joint model in this group, hold the index at which the corresponding joint state starts in
       the group state */
-  std::vector<int> active_joint_model_start_index_;
+  std::vector<size_t> active_joint_model_start_index_;
 
   /** \brief The links that are on the direct lineage between joints
       and joint_roots_, as well as the children of the joint leafs.
--- a/moveit_core/robot_model/src/joint_model_group.cpp
+++ b/moveit_core/robot_model/src/joint_model_group.cpp
@@ -229,18 +229,18 @@ JointModelGroup::JointModelGroup(const std::string& group_name, const srdf::Mode
   is_contiguous_index_list_ = !variable_index_list_.empty()
     && all_of(variable_index_list_ | views::sliding(2), is_ascending);
 
-  // clang-format on
+  // when updating/sampling a group state only,
+  // only mimic joints that have their parent within the group get updated.
+  group_mimic_update_ = mimic_joints_
+    | views::filter([this](const auto& jm) { return hasJointModel(jm->getMimic()->getName()); })
+    | views::transform([this](const auto& jm) {
+      int src = joint_variables_index_map_[jm->getMimic()->getName()];
+      int dest = joint_variables_index_map_[jm->getName()];
+      return GroupMimicUpdate(src, dest, jm->getMimicFactor(), jm->getMimicOffset());
+    })
+    | to<std::vector>();
 
-  // when updating/sampling a group state only, only mimic joints that have their parent within the group get updated.
-  for (const JointModel* mimic_joint : mimic_joints_)
-    // if the joint we mimic is also in this group, we will need to do updates when sampling
-    if (hasJointModel(mimic_joint->getMimic()->getName()))
-    {
-      int src = joint_variables_index_map_[mimic_joint->getMimic()->getName()];
-      int dest = joint_variables_index_map_[mimic_joint->getName()];
-      GroupMimicUpdate mu(src, dest, mimic_joint->getMimicFactor(), mimic_joint->getMimicOffset());
-      group_mimic_update_.push_back(mu);
-    }
+  // clang-format on
 
   // now we need to make another pass for group links (we include the fixed joints here)
   std::set<const LinkModel*> group_links_set;
--- a/moveit_core/robot_model/src/joint_model_group.cpp
+++ b/moveit_core/robot_model/src/joint_model_group.cpp
@@ -100,6 +100,18 @@ bool jointPrecedes(const JointModel* a, const JointModel* b)
 
   return false;
 }
+
+const auto is_fixed_joint = [](const auto& jm) { return jm->getVariableCount() == 0; };
+const auto is_mimic_joint = [](const auto& jm) { return jm->getMimic() != nullptr; };
+const auto get_name = [](const auto& element) { return element->getName(); };
+const auto get_variable_count = [](const auto& jm) { return jm->getVariableCount(); };
+const auto get_variable_bounds_addr = [](const auto& jm) { return &jm->getVariableBounds(); };
+const auto get_first_variable_index = [](const auto& jm) { return jm->getFirstVariableIndex(); };
+const auto get_variable_names_view = [](const auto& jm) { return views::all(jm->getVariableNames()); };
+const auto get_variable_names_size = [](const auto& jm) { return jm->getVariableNames().size(); };
+const auto tuple_to_pair = [](const auto& element) {
+  return std::make_pair(std::get<0>(element), std::get<1>(element));
+};
 }  // namespace
 
 JointModelGroup::JointModelGroup(const std::string& group_name, const srdf::Model::Group& config,
@@ -115,53 +127,91 @@ JointModelGroup::JointModelGroup(const std::string& group_name, const srdf::Mode
   , is_single_dof_{ true }
   , config_{ config }
 {
-  joint_model_vector_ = unsorted_group_joints | to<std::vector>() | actions::sort(OrderJointsByIndex());
+  // Disable clang-format to keep nicer formatting of ranges
+  // clang-format off
+  joint_model_vector_ = unsorted_group_joints
+    | to<std::vector>()
+    | actions::sort(OrderJointsByIndex());
+  joint_model_name_vector_ = joint_model_vector_
+    | views::transform(get_name)
+    | to<std::vector>();
 
   // figure out active joints, mimic joints, fixed joints
   // construct index maps, list of variables
-  for (const JointModel* joint_model : joint_model_vector_)
-  {
-    joint_model_name_vector_.push_back(joint_model->getName());
-    joint_model_map_[joint_model->getName()] = joint_model;
-    unsigned int vc = joint_model->getVariableCount();
-    if (vc > 0)
-    {
-      if (vc > 1)
-        is_single_dof_ = false;
-      const std::vector<std::string>& name_order = joint_model->getVariableNames();
-      if (joint_model->getMimic() == nullptr)
-      {
-        active_joint_model_vector_.push_back(joint_model);
-        active_joint_model_name_vector_.push_back(joint_model->getName());
-        active_joint_model_start_index_.push_back(variable_count_);
-        active_joint_models_bounds_.push_back(&joint_model->getVariableBounds());
-        active_variable_count_ += vc;
-      }
-      else
-        mimic_joints_.push_back(joint_model);
-      for (const std::string& name : name_order)
-      {
-        variable_names_.push_back(name);
-        variable_names_set_.insert(name);
-      }
+  auto not_fixed_view = joint_model_vector_
+    | views::remove_if(is_fixed_joint);
+  auto start_index_view =
+    views::concat(views::single(0),
+      not_fixed_view
+        | views::transform(get_variable_count)
+        | views::partial_sum
+      )
+    | views::drop_last(1);
 
-      int first_index = joint_model->getFirstVariableIndex();
-      for (std::size_t j = 0; j < name_order.size(); ++j)
-      {
-        variable_index_list_.push_back(first_index + j);
-        joint_variables_index_map_[name_order[j]] = variable_count_ + j;
-      }
-      joint_variables_index_map_[joint_model->getName()] = variable_count_;
-
-      if (joint_model->getType() == JointModel::REVOLUTE &&
-          static_cast<const RevoluteJointModel*>(joint_model)->isContinuous())
-        continuous_joint_model_vector_.push_back(joint_model);
-
-      variable_count_ += vc;
-    }
-    else
-      fixed_joints_.push_back(joint_model);
-  }
+  active_joint_model_vector_ = not_fixed_view
+    | views::remove_if(is_mimic_joint)
+    | to<std::vector>();
+  active_joint_model_name_vector_ = active_joint_model_vector_
+    | views::transform(get_name)
+    | to<std::vector>();
+  fixed_joints_ = joint_model_vector_
+    | views::filter(is_fixed_joint)
+    | to<std::vector>();
+  mimic_joints_ = not_fixed_view
+    | views::filter(is_mimic_joint)
+    | to<std::vector>();
+  continuous_joint_model_vector_ = not_fixed_view
+    | views::filter([](const auto& jm) {
+        return jm->getType() == JointModel::REVOLUTE && static_cast<const RevoluteJointModel*>(jm)->isContinuous();
+    })
+    | to<std::vector>();
+  variable_names_ = not_fixed_view
+    | views::transform(get_variable_names_view)
+    | views::join
+    | to<std::vector>();
+  variable_names_set_ = variable_names_
+    | to<std::set>();
+  joint_model_map_ = joint_model_vector_
+    | views::transform([](const auto& jm) { return std::make_pair(jm->getName(), jm); })
+    | to<std::map>();
+  joint_variables_index_map_ =
+    views::concat(
+      views::zip(not_fixed_view | views::transform(get_variable_names_view),
+                 start_index_view)
+        | views::transform([](const auto& e) {
+          const auto& [names, start_index] = e;
+          return views::zip(names,
+                            views::iota(start_index, start_index + names.size()));
+        })
+        | views::join
+        | views::transform(tuple_to_pair)
+      ,
+      views::zip(not_fixed_view | views::transform(get_name),
+                 start_index_view)
+        | views::transform(tuple_to_pair)
+    )
+    | to<std::map>();
+  active_joint_models_bounds_ = active_joint_model_vector_
+    | views::transform(get_variable_bounds_addr)
+    | to<std::vector>();
+  variable_index_list_ =
+    views::zip(not_fixed_view | views::transform(get_first_variable_index),
+               not_fixed_view | views::transform(get_variable_names_size))
+    | views::transform([](const auto& e) {
+     const auto& [first_index, names_size] = e;
+     return views::iota(first_index, first_index + names_size);
+    })
+    | views::join
+    | to<std::vector>();
+  active_joint_model_start_index_ =
+    views::zip(start_index_view, not_fixed_view)
+    | views::remove_if([](const auto& e) { return is_mimic_joint(std::get<1>(e)); })
+    | views::transform([](const auto& e) { return std::get<0>(e); })
+    | to<std::vector>();
+  variable_count_ = accumulate(not_fixed_view | views::transform(get_variable_count), 0U);
+  active_variable_count_ = accumulate(not_fixed_view | views::transform(get_variable_count), 0U);
+  is_single_dof_ = none_of(not_fixed_view | views::transform(get_variable_count), [](const auto& e) { return e > 1; });
+  // clang-format on
 
   // now we need to find all the set of joints within this group
   // that root distinct subtrees
@@ -262,7 +312,7 @@ JointModelGroup::JointModelGroup(const std::string& group_name, const srdf::Mode
     if (chain)
       is_chain_ = true;
   }
-}
+}  // namespace core
 
 JointModelGroup::~JointModelGroup() = default;
 
@@ -767,5 +817,5 @@ bool JointModelGroup::isValidVelocityMove(const double* from_joint_pose, const d
 
   return true;
 }
-}  // end of namespace core
+}  // namespace core
 }  // end of namespace moveit
--- a/moveit_core/robot_model/src/joint_model_group.cpp
+++ b/moveit_core/robot_model/src/joint_model_group.cpp
@@ -103,6 +103,7 @@ bool jointPrecedes(const JointModel* a, const JointModel* b)
 
 const auto is_fixed_joint = [](const auto& jm) { return jm->getVariableCount() == 0; };
 const auto is_mimic_joint = [](const auto& jm) { return jm->getMimic() != nullptr; };
+const auto has_geometry = [](const auto& lm) { return !lm->getShapes().empty(); };
 const auto get_name = [](const auto& element) { return element->getName(); };
 const auto get_variable_count = [](const auto& jm) { return jm->getVariableCount(); };
 const auto get_variable_bounds_addr = [](const auto& jm) { return &jm->getVariableBounds(); };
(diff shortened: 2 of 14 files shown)


moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -113,154 +126,153 @@ JointModelGroup::JointModelGroup(const std::string& group_name, const srdf::Mode
...
-    if (hasJointModel(mimic_joint->getMimic()->getName()))
-    {
-      int src = joint_variables_index_map_[mimic_joint->getMimic()->getName()];
-      int dest = joint_variables_index_map_[mimic_joint->getName()];
-      GroupMimicUpdate mu(src, dest, mimic_joint->getMimicFactor(), mimic_joint->getMimicOffset());
-      group_mimic_update_.push_back(mu);
-    }
+  is_contiguous_index_list_ =
+      !variable_index_list_.empty() && none_of(variable_index_list_ | views::sliding(2), [](const auto& sub_view) {
+        const auto& values = sub_view | to<std::vector>();
```
griswaldbrooks, line 208: Feels like there should be a way around this type conversion

griswaldbrooks (commented):
I have more to review, but this is taking a while as I have to learn at the same time.

(The compiler errors make me feel like I'm in undergrad :facepalm:

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -113,154 +126,153 @@ JointModelGroup::JointModelGroup(const std::string& group_name, const srdf::Mode
   , config_(config)
 {
   // sort joints in Depth-First order
-  joint_model_vector_ = unsorted_group_joints;
-  std::sort(joint_model_vector_.begin(), joint_model_vector_.end(), OrderJointsByIndex());
-  joint_model_name_vector_.reserve(joint_model_vector_.size());
+  joint_model_vector_ = unsorted_group_joints | to<std::vector>() | actions::sort(OrderJointsByIndex());
```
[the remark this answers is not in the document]
tylerjw: I did not.  You have to convert it from a range to a vector before you can sort.  If you do not specify the conversion to vector at the end it prints a deprecation warning about the conversion.

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -113,154 +126,153 @@ JointModelGroup::JointModelGroup(const std::string& group_name, const srdf::Mode
...
-          static_cast<const RevoluteJointModel*>(joint_model)->isContinuous())
-        continuous_joint_model_vector_.push_back(joint_model);
+  fixed_joints_ = joint_model_vector_ | views::filter(is_fixed_joint) | to<std::vector>();
 
-      variable_count_ += vc;
-    }
-    else
-      fixed_joints_.push_back(joint_model);
-  }
+  auto not_fixed_joints_view = joint_model_vector_ | views::remove_if(is_fixed_joint);
```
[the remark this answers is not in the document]
tylerjw: This is a view.  You don't want to make const views.  Here is the rationale for that: https://ericniebler.github.io/range-v3/index.html#autotoc_md6

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -97,6 +100,16 @@ bool jointPrecedes(const JointModel* a, const JointModel* b)
 
   return false;
 }
+
+const auto is_fixed_joint = [](const auto& jm) { return jm->getVariableCount() == 0; };
+const auto is_mimic_joint = [](const auto& jm) { return jm->getMimic() != nullptr; };
+const auto has_geometry = [](const auto& lm) { return !lm->getShapes().empty(); };
+const auto get_name = [](const auto& element) { return element->getName(); };
+const auto get_variable_count = [](const auto& jm) { return jm->getVariableCount(); };
+const auto get_variable_bounds_addr = [](const auto& jm) { return &jm->getVariableBounds(); };
```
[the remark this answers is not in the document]
tylerjw: this returns the address of the getVariableBounds return value.  This is just continuing functionality that was there.  I agree this is weird.

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -113,154 +126,153 @@ JointModelGroup::JointModelGroup(const std::string& group_name, const srdf::Mode
...
+      views::drop_last(1);
+
+  active_joint_model_start_index_ = views::zip(not_fixed_start_index_view, not_fixed_joints_view) |
+                                    views::remove_if([](const auto& e) { return is_mimic_joint(std::get<1>(e)); }) |
+                                    views::transform([](const auto& e) { return std::get<0>(e); }) | to<std::vector>();
+
+  active_joint_models_bounds_ =
+      active_joint_model_vector_ | views::transform(get_variable_bounds_addr) | to<std::vector>();
+
+  active_variable_count_ = accumulate(active_joint_model_vector_ | views::transform(get_variable_count), 0);
```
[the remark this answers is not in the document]
tylerjw: good catch.  I didn't know about that.  That does work and I will include that in my next update.

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -113,154 +126,153 @@ JointModelGroup::JointModelGroup(const std::string& group_name, const srdf::Mode
...
+
+  is_single_dof_ =
+      none_of(not_fixed_joints_view | views::transform(get_variable_count), [](const auto& e) { return e > 1; });
+
+  mimic_joints_ = not_fixed_joints_view | views::filter(is_mimic_joint) | to<std::vector>();
+
+  variable_names_ = not_fixed_joints_view | views::transform(get_variable_names_view) | views::join | to<std::vector>();
+
+  variable_names_set_ = variable_names_ | views::unique | to<std::set>();
```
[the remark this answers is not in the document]
tylerjw: Good catch, I've tested and in this case I can drop the views::unique and it does the same thing.

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -41,9 +41,12 @@
 #include <moveit/exceptions/exceptions.h>
 #include <boost/lexical_cast.hpp>
 #include <algorithm>
+#include <range/v3/all.hpp>
```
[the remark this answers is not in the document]
tylerjw: From what I can tell this is standard.  Eventually all of this will probably be in one stl header <ranges>.

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -113,154 +126,153 @@ JointModelGroup::JointModelGroup(const std::string& group_name, const srdf::Mode
...
+
+  joint_variables_index_map_ =
+      views::concat(
+          views::zip(not_fixed_joints_view | views::transform(get_variable_names_view), not_fixed_start_index_view) |
+              views::transform([](const auto& e) {
+                const auto& [names, start_index] = e;
+                return views::zip(names, views::iota(start_index, start_index + names.size()));
+              }) |
+              views::join |
+              views::transform([](const auto& e) { return std::make_pair(std::get<0>(e), std::get<1>(e)); }),
```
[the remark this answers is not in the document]
tylerjw: I am, I've refactored this to make it less verbose and clear why I need this.

moveit_core/robot_model/src/joint_model_group.cpp
tylerjw: As far as I can tell this is the most straightforward way to be able to index the elements of a view.  I agree it is clunky and I feel like there should be something better.

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -113,154 +126,153 @@ JointModelGroup::JointModelGroup(const std::string& group_name, const srdf::Mode
...
-    else
-      fixed_joints_.push_back(joint_model);
-  }
+  auto not_fixed_joints_view = joint_model_vector_ | views::remove_if(is_fixed_joint);
+
+  active_joint_model_vector_ = not_fixed_joints_view | views::remove_if(is_mimic_joint) | to<std::vector>();
+
+  active_joint_model_name_vector_ = active_joint_model_vector_ | views::transform(get_name) | to<std::vector>();
+
+  auto not_fixed_start_index_view =
```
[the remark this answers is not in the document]
tylerjw: I refactored this to hopefully make it more clear.  Let me know what you think of the update.

tylerjw:
I broke this into separate small commits to hopefully make it easier to review.

tylerjw:
One thing I did that I'm not sure I like is manually formatted much of this code and disabled clang-format for the block.  I did this because I think it makes it much more readable than the way our clang-format is configured for these huge blocks of initialization via ranges.

tylerjw:
Before someone asks I did try putting this all in the initializer list and determined that it was basically unreadable.  The whole point of this change is to make it easier to reason about what is in these containers.  This is the same reason I disabled clang-format and manually formatted this block.

tylerjw:
One thing nice is that almost all of the code in this file is covered by existing tests.  While those tests might not be perfect they should give us more confidence that these changes do not change the behavior of this code.

griswaldbrooks:
> Before someone asks I did try putting this all in the initializer list and determined that it was basically unreadable. The whole point of this change is to make it easier to reason about what is in these containers. This is the same reason I disabled clang-format and manually formatted this block.

I would have asked about the initializer list but I really think that would require breaking this up into smaller objects which I realize is not the point of this PR.

moveit_core/robot_model/src/joint_model_group.cpp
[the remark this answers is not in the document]
griswaldbrooks: :+1:

moveit_core/robot_model/src/joint_model_group.cpp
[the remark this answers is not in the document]
griswaldbrooks: :+1:

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -97,188 +101,239 @@ bool jointPrecedes(const JointModel* a, const JointModel* b)
...
+  , is_contiguous_index_list_{ true }
+  , is_chain_{ false }
+  , is_single_dof_{ true }
+  , config_{ config }
+{
+  // Disable clang-format to keep nicer formatting of ranges
+  // clang-format off
+  joint_model_vector_ = unsorted_group_joints
+    | ranges::to<std::vector>()
+    | actions::sort(OrderJointsByIndex());
```
griswaldbrooks, line 142:
```suggestion
    | actions::sort(OrderJointsByIndex{});
```
Can we use uniform initialization?

moveit_core/robot_model/src/joint_model_group.cpp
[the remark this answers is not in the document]
griswaldbrooks: That's too bad. It'd be nice to have a touch more sugar.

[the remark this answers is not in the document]
tylerjw: agreed... it is a big step forward for sugar in the language but at the same time there are things like this that seem like obvious no-brainers but they are siding with more explicit code which is consistent with much of what ends up in the stl.

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -97,188 +101,239 @@ bool jointPrecedes(const JointModel* a, const JointModel* b)
...
   // now we need to find all the set of joints within this group
   // that root distinct subtrees
-  for (const JointModel* active_joint_model : active_joint_model_vector_)
-  {
-    // if we find that an ancestor is also in the group, then the joint is not a root
-    if (!includesParent(active_joint_model, this))
-      joint_roots_.push_back(active_joint_model);
-  }
+  joint_roots_ = active_joint_model_vector_
+    | views::filter([this](const auto& jm) {
```
griswaldbrooks, line 222: Do you think you should consider `remove_if` to avoid the negation of `includesParent` or does that result in something different?

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -97,188 +101,239 @@ bool jointPrecedes(const JointModel* a, const JointModel* b)
...
-  for (const JointModel* mimic_joint : mimic_joints_)
-    // if the joint we mimic is also in this group, we will need to do updates when sampling
-    if (hasJointModel(mimic_joint->getMimic()->getName()))
-    {
-      int src = joint_variables_index_map_[mimic_joint->getMimic()->getName()];
-      int dest = joint_variables_index_map_[mimic_joint->getName()];
-      GroupMimicUpdate mu(src, dest, mimic_joint->getMimicFactor(), mimic_joint->getMimicOffset());
-      group_mimic_update_.push_back(mu);
-    }
+  const auto is_ascending = [](const auto& sub_view) {
```
griswaldbrooks, line 229: Should this be renamed? This function implies that any value greater would qualify, but this check to see if it is *exactly* the next one

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -113,154 +126,153 @@ JointModelGroup::JointModelGroup(const std::string& group_name, const srdf::Mode
...
+  joint_model_name_vector_ = joint_model_vector_ | views::transform(get_name) | to<std::vector>();
 
-      int first_index = joint_model->getFirstVariableIndex();
-      for (std::size_t j = 0; j < name_order.size(); ++j)
-      {
-        variable_index_list_.push_back(first_index + j);
-        joint_variables_index_map_[name_order[j]] = variable_count_ + j;
-      }
-      joint_variables_index_map_[joint_model->getName()] = variable_count_;
+  joint_model_map_ = joint_model_vector_ |
```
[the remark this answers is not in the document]
tylerjw: I'm not sure about the type but that might be clearer.  I'll try that.

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -97,188 +101,231 @@ bool jointPrecedes(const JointModel* a, const JointModel* b)
...
+    && ranges::all_of(variable_index_list_ | views::sliding(2), is_ascending);
+
+  // when updating/sampling a group state only,
+  // only mimic joints that have their parent within the group get updated.
+  group_mimic_update_ = mimic_joints_
+    | views::filter([this](const auto& jm) { return hasJointModel(jm->getMimic()->getName()); })
+    | views::transform([this](const auto& jm) {
+      int src = joint_variables_index_map_[jm->getMimic()->getName()];
+      int dest = joint_variables_index_map_[jm->getName()];
+      return GroupMimicUpdate(src, dest, jm->getMimicFactor(), jm->getMimicOffset());
```
griswaldbrooks, line 243: Awww, I so wish this was a vector of lambdas, rather than structs :crying_cat_face:

moveit_core/robot_model/src/joint_model_group.cpp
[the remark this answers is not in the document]
tylerjw: through some testing, I've realized this was unnecessary and removed the tuple to pair transforms

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -97,188 +101,231 @@ bool jointPrecedes(const JointModel* a, const JointModel* b)
...
+    | ranges::to<std::vector>();
 
   // compute the common root of this group
   if (!joint_roots_.empty())
   {
-    common_root_ = joint_roots_[0];
-    for (std::size_t i = 1; i < joint_roots_.size(); ++i)
-      common_root_ = parent_model->getCommonRoot(joint_roots_[i], common_root_);
+    common_root_ = ranges::accumulate(joint_roots_, joint_roots_[0],
+      [this](const auto& left, const auto& right) {
```
griswaldbrooks, line 270: See, this. It feels like this should be possible for all the zipped views, instead of having to use `std::get` or unpack a parameter.

moveit_core/robot_model/src/joint_model_group.cpp
griswaldbrooks: Given so, can we
```suggestion
      return GroupMimicUpdate{src, dest, jm->getMimicFactor(), jm->getMimicOffset()};
```

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -97,188 +101,231 @@ bool jointPrecedes(const JointModel* a, const JointModel* b)
...
+  updated_link_model_set_ = joint_roots_
+    | views::transform([](const auto& jr) { return views::all(jr->getDescendantLinkModels()); })
+    | views::join
+    | ranges::to<std::set>();
+  updated_link_model_name_set_ = updated_link_model_set_
+    | views::transform(getElementName)
+    | ranges::to<std::set>();
+  updated_link_model_vector_ = updated_link_model_set_
+    | ranges::to<std::vector>()
+    | actions::sort(OrderLinksByIndex());
```
griswaldbrooks, line 286:
```suggestion
    | actions::sort(OrderLinksByIndex{});
```
?

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -97,188 +101,231 @@ bool jointPrecedes(const JointModel* a, const JointModel* b)
...
+  updated_link_model_with_geometry_vector_ = updated_link_model_set_
+    | views::filter(has_geometry)
+    | ranges::to<std::vector>()
+    | actions::sort(OrderLinksByIndex());
```
griswaldbrooks, line 290:
```suggestion
    | actions::sort(OrderLinksByIndex{});
```
?

griswaldbrooks (commented):
Not a fan of the new line length, but much of it does seem more readable.
At least I was able to get through the entire ctor this time.

I still don't really know what's going on with `joint_variables_index_map_` and some of the others in that part of the code (feels like some intermediate named variables or lambas could help).

griswaldbrooks (commented):
Also, feel free to resolve anything that I've thumb's upped, since I can't resolve things.

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -102,24 +102,9 @@ bool jointPrecedes(const JointModel* a, const JointModel* b)
...
-const auto get_first_variable_index = [](const auto& jm) { return jm->getFirstVariableIndex(); };
-const auto get_variable_names_view = [](const auto& jm) { return views::all(jm->getVariableNames()); };
-const auto get_variable_names_size = [](const auto& jm) { return jm->getVariableNames().size(); };
-
-// Helper to convert a tuple to a pair.  This is useful as a transform after
-// zipping two views inorder to initialize a map.
-const auto tuple_to_pair = [](const auto& element) {
-  return std::make_pair(std::get<0>(element), std::get<1>(element));
-};
+// These is a transform function for calling getName on a Joint or Link Model
```
griswaldbrooks, line 105: Grammar

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -102,24 +102,9 @@ bool jointPrecedes(const JointModel* a, const JointModel* b)
...
+constexpr auto getElementName = [](const auto& element) { return element->getName(); };
```
griswaldbrooks, line 106: Are we doing snake case or camel case?

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -216,7 +208,7 @@ JointModelGroup::JointModelGroup(const std::string& group_name, const srdf::Mode
     | ranges::to<std::vector>();
   active_joint_model_start_index_ =
     views::zip(start_index_view, not_fixed_view)
-    | views::remove_if([](const auto& e) { return is_mimic_joint(std::get<1>(e)); })
+    | views::remove_if([=](const auto& e) { return is_mimic_joint(std::get<1>(e)); })
```
griswaldbrooks, line 211: What is getting copied here?

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -97,188 +101,239 @@ bool jointPrecedes(const JointModel* a, const JointModel* b)
...
+    })
+    | views::join
+    | ranges::to<std::vector>();
+  active_joint_model_start_index_ =
+    views::zip(start_index_view, not_fixed_view)
+    | views::remove_if([](const auto& e) { return is_mimic_joint(std::get<1>(e)); })
+    | views::transform([](const auto& e) { return std::get<0>(e); })
+    | ranges::to<std::vector>();
+  variable_count_ = ranges::accumulate(not_fixed_view | views::transform(get_variable_count), 0U);
+  active_variable_count_ = ranges::accumulate(not_fixed_view | views::transform(get_variable_count), 0U);
```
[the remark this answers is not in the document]
tylerjw: They are and that's a bug.  I updated this fixing that.

moveit_core/robot_model/src/joint_model_group.cpp
tylerjw: good catch, changed

moveit_core/robot_model/src/joint_model_group.cpp
tylerjw: how about `is_contiguous` ?  I renamed it to that.

moveit_core/robot_model/src/joint_model_group.cpp
tylerjw: :+1:

moveit_core/robot_model/src/joint_model_group.cpp
tylerjw: I agree and I've tried but am unable to get it to compile.  accumulate specifically takes a binary function.

moveit_core/robot_model/src/joint_model_group.cpp
tylerjw: :+1:

moveit_core/robot_model/src/joint_model_group.cpp
tylerjw: :+1:

moveit_core/robot_model/src/joint_model_group.cpp
tylerjw: If it is in the global scope our clang-tidy rules say it has to be camel case.  If it is in the scope of a function it is snake case.  That rule drives me crazy.

moveit_core/robot_model/src/joint_model_group.cpp
tylerjw: :+1:

moveit_core/robot_model/src/joint_model_group.cpp
tylerjw: The is_mimic_joint lambda.  For some reason clang won't capture it by default where gcc will.

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -113,154 +126,153 @@ JointModelGroup::JointModelGroup(const std::string& group_name, const srdf::Mode
...
+
+  variable_index_list_ = views::zip(not_fixed_joints_view | views::transform(get_first_variable_index),
+                                    not_fixed_joints_view | views::transform(get_variable_names_size)) |
+                         views::transform([](const auto& e) {
+                           const auto& [first_index, names_size] = e;
+                           return views::iota(first_index, first_index + names_size);
+                         }) |
+                         views::join | to<std::vector>();
+
+  joint_variables_index_map_ =
```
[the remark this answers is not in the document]
tylerjw: I broke this up and added comments around it.  Let me know if that helps.

moveit_core/robot_model/src/joint_model_group.cpp
tylerjw: :+1:

tylerjw:
> I still don't really know what's going on with `joint_variables_index_map_` and some of the others in that part of the code (feels like some intermediate named variables or lambas could help).

That is a good point.  I updated by adding some intermediate named variables and adding comments.  Let me know if that makes it clearer.

moveit_core/robot_model/src/joint_model_group.cpp
tylerjw: The reason I want this one to be global scope is it is used by several functions after the compiler.

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -457,17 +536,24 @@ bool JointModelGroup::getVariableDefaultPositions(const std::string& name, std::
...
 
 void JointModelGroup::getVariableDefaultPositions(std::map<std::string, double>& values) const
 {
-  std::vector<double> tmp(variable_count_);
-  getVariableDefaultPositions(&tmp[0]);
-  for (std::size_t i = 0; i < variable_names_.size(); ++i)
-    values[variable_names_[i]] = tmp[i];
+  const auto default_positions = [=]() {
+    auto tmp = std::vector<double>(variable_count_);
+    getVariableDefaultPositions(&tmp[0]);
```
jliukkonen, line 550: I wonder what happens if the variable counts ends up being zero... The same problem is of course also present in the current code.

moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
```diff
@@ -282,7 +282,7 @@ class JointModelGroup
   }
 
   /** \brief Get the index locations in the complete robot state for all the variables in this group */
-  const std::vector<int>& getVariableIndexList() const
+  const std::vector<size_t>& getVariableIndexList() const
```
jliukkonen, line 285: I hope there is no code that does some "clever" negative value index tricks (like use -1 as a flag of a kind or something similar).

moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
```diff
@@ -699,10 +699,10 @@ class JointModelGroup
   std::set<std::string> updated_link_model_with_geometry_name_set_;
 
   /** \brief The number of variables necessary to describe this group of joints */
-  unsigned int variable_count_;
+  size_t variable_count_;
 
   /** \brief The number of variables necessary to describe the active joints in this group of joints */
-  unsigned int active_variable_count_;
+  size_t active_variable_count_;
```
jliukkonen, line 705: Nit: I don't know what the best practice here is, but since `size_t` is basically a type defined in the old C standard libraries as well as in the STL ported C++ versions of those libraries, shouldn't it be `std::size_t` everywhere?

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -97,188 +98,259 @@ bool jointPrecedes(const JointModel* a, const JointModel* b)
...
-    common_root_ = joint_roots_[0];
-    for (std::size_t i = 1; i < joint_roots_.size(); ++i)
-      common_root_ = parent_model->getCommonRoot(joint_roots_[i], common_root_);
-  }
+  link_model_vector_ = joint_model_vector_
+    | views::transform([](const auto& jm) { return jm->getChildLinkModel(); })
+    | views::unique
+    | ranges::to<std::vector>()
+    | actions::sort(OrderLinksByIndex{});
+  link_model_map_ = link_model_vector_
```
jliukkonen, line 288: I wonder if it would improve readability if there were small comments on top of these operations. 
Here it could say `// Create a link name / link model map.`

jliukkonen (commented):
Well, the code _looks_ correct to me but the change set is large and I'm not a Ranges expert so I can't be sure I didn't miss something.

moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
tylerjw: There was and I removed it with a previous PR.

moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
tylerjw: I think they are effectively the same thing and I like the shorter name.

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -55,28 +59,25 @@ static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_robot_model.join
 // check if a parent or ancestor of joint is included in this group
 bool includesParent(const JointModel* joint, const JointModelGroup* group)
 {
-  bool found = false;
+  const auto test = [group](const auto& jm) {
+    return group->hasJointModel(jm->getName()) && jm->getVariableCount() > 0 && jm->getMimic() == nullptr;
+  };
```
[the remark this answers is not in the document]
jliukkonen: I'm fine with `is_parent`. It is more descriptive than `test`.

jliukkonen (commented):
There are couple comments of mine that should either be marked as "wont do" or then addressed. I'm happy with the changes in general.

moveit_core/robot_model/src/joint_model_group.cpp
tylerjw: I added some comments where I thought the ranges expression was particularly complex.  Are there any others you think could be improved with comments?

moveit_core/robot_model/src/joint_model_group.cpp
tylerjw: I don't know... that's a good question.  I wasn't trying to change the behavior of existing code though with this PR.

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -113,154 +126,153 @@ JointModelGroup::JointModelGroup(const std::string& group_name, const srdf::Mode
...
+  is_single_dof_ =
+      none_of(not_fixed_joints_view | views::transform(get_variable_count), [](const auto& e) { return e > 1; });
+
+  mimic_joints_ = not_fixed_joints_view | views::filter(is_mimic_joint) | to<std::vector>();
+
+  variable_names_ = not_fixed_joints_view | views::transform(get_variable_names_view) | views::join | to<std::vector>();
+
+  variable_names_set_ = variable_names_ | views::unique | to<std::set>();
+
+  variable_index_list_ = views::zip(not_fixed_joints_view | views::transform(get_first_variable_index),
```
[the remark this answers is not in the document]
tylerjw: I reworked this to try to make it clearer.

tylerjw:
Closing as this is abandoned work.

PR #1030, opened 2022-01-26
Split release and testing dependencies

henningkayser:


moveit2.repos
```diff
@@ -1,17 +1 @@
 repositories:
```
henningkayser, line 1: I'm inclined to either not use this file for CI at all, but still list our dependencies so that they are included when building from source

.github/workflows/ci.yaml
```diff
@@ -22,11 +22,6 @@ jobs:
           - IMAGE: rolling-ci-testing
             IKFAST_TEST: true
             ROS_DISTRO: rolling
-          - IMAGE: galactic-ci
```
henningkayser, line 25: We'll need to drop this anyway

AndyZe (approved):
This makes sense to me

moveit2_rolling-ci-testing.repos
```diff
@@ -0,0 +1,17 @@
+repositories:
+  geometric_shapes:
+    type: git
+    url: https://github.com/ros-planning/geometric_shapes
+    version: ros2
+  moveit_msgs:
+    type: git
+    url: https://github.com/ros-planning/moveit_msgs
+    version: ros2
```
tylerjw, line 9: Would it make sense to remove these from this file because technically the released versions are exactly the same (there are no changes).  The same applies to the other repos file.

henningkayser: Not really, the point is that these can be ahead of the released version like moveit_resources and keeping them in here allows us to catch syncing issues early. The testing job should really run anything that is not released yet. The only alternative for keeping these in here would probably be downstream testing in the other repos. But still, we would have to add them again for the prerelease test, so why not just keep them in here?
Maybe we can remove the entries for the release build, because those should really match the version of the debians.

moveit2.repos
henningkayser: that works as well

henningkayser:
I guess the PILZ planner will still fail with this...

moveit2_rolling-ci-testing.repos
tylerjw: I understand having them here when there are changes that are not yet released.  My point was that specifically in the case of these two there is no difference between the ros2 branch and the release.  Your point is that by leaving them here it would catch if there is a change.  As we control those repos though we will know when there is a change.

moveit2_rolling-ci.repos
```diff
@@ -0,0 +1,17 @@
...
+    url: https://github.com/ros-planning/moveit_msgs
+    version: 2.2.0
+  moveit_resources:
+    type: git
+    url: https://github.com/ros-planning/moveit_resources
+    version: 2.0.3
+  launch_param_builder:
+    type: git
+    url: https://github.com/PickNikRobotics/launch_param_builder
+    version: main
```
tylerjw, line 17: There is a tagged released version of this.  (it just hasn't been synced yet)

Secondly, don't you need to add ros2_control into this second ci one?

tylerjw (commented):
The more I think about it the more I like this change.  One really large advantage to having the branches (as opposed to tags) in the repos file is if you use the repos file for development locally.

That being said I think what is needed here is to add the ros2_control repo temporarily like was done on main.

moveit2_rolling-ci.repos
henningkayser: ros2_control should not be required since the fix in moveit_resources is the reason we were failing in the first place. With moveit_resources set to the released version, the released version of ros2_control still works.

moveit2_rolling-ci-testing.repos
henningkayser: I don't trust us enough to change this in time, I guess ;)

henningkayser:
> The more I think about it the more I like this change. One really large advantage to having the branches (as opposed to tags) in the repos file is if you use the repos file for development locally.
> 
> That being said I think what is needed here is to add the ros2_control repo temporarily like was done on main.

Like mentioned above, ros2_control should not be required for the release build since the released versions of moveit_resources and ros2_control work well with each other. We only need the moveit_resources fix for the testing build, and that's it. Adding ros2_control was only required because the fix in moveit_resources is not backwards compatible.

tylerjw:
> > The more I think about it the more I like this change. One really large advantage to having the branches (as opposed to tags) in the repos file is if you use the repos file for development locally.
> > That being said I think what is needed here is to add the ros2_control repo temporarily like was done on main.
> 
> Like mentioned above, ros2_control should not be required for the release build since the released versions of moveit_resources and ros2_control work well with each other. We only need the moveit_resources fix for the testing build, and that's it. Adding ros2_control was only required because the fix in moveit_resources is not backwards compatible.

I didn't realize that.  Another thing I discovered is that by dropping moveit_msgs and geometric_shapes from the repos file in my other PR we fail to build because we are depending on transient dependencies from them.  I'll make a PR to add those dependencies as it should be possible to build without them even if they are in our repos file.

tylerjw:
@henningkayser I can't tell where we ended with this change. What should we do, how do you think we should handle these dependencies?

tylerjw:
As this PR has conflicts and hasn't been worked on for over a year, I'm closing it. We can revisit this work and re-open this if you'd like.

PR #1039, opened 2022-01-27
Fix race condition in SynchronizedStringParameter

corycrean:
- Fix race condition in SynchronizedStringParameter.
- Remove redundant callbacks.

### Description

This PR addresses the issue in https://github.com/ros-planning/moveit2_tutorials/pull/262 where the robot model can't be loaded.  Currently, the subscription created in `SynchronizedStringParameter::waitForMessage` is added to two `WaitSet` objects (one belonging to the executor, and one belonging to the SynchronizedStringParameter), and it's unpredictable which `WaitSet` will end up processing the messages received by that subscription.  This PR fixes the problem by removing the `WaitSet` from `SynchronizedStringParameter`.

### Checklist


moveit_ros/planning/rdf_loader/include/moveit/rdf_loader/synchronized_string_parameter.h
```diff
@@ -80,5 +84,7 @@ class SynchronizedStringParameter
   rclcpp::Publisher<std_msgs::msg::String>::SharedPtr string_publisher_;
 
   std::string content_;
+  boost::lockfree::spsc_queue<std::string> queue_;
+  std::mutex queue_mutex_;
```
JafarAbdi, line 88: Is there a need for a mutex? I thought lock-free mean no mutex needed

moveit_ros/planning/rdf_loader/include/moveit/rdf_loader/synchronized_string_parameter.h
```diff
@@ -80,5 +84,7 @@ class SynchronizedStringParameter
   rclcpp::Publisher<std_msgs::msg::String>::SharedPtr string_publisher_;
 
   std::string content_;
+  boost::lockfree::spsc_queue<std::string> queue_;
```
JafarAbdi, line 87: I would vote for a solution that doesn't require Boost since we're planning to get rid of it as a dependency

moveit_ros/planning/rdf_loader/src/rdf_loader.cpp
```diff
@@ -66,14 +66,10 @@ RDFLoader::RDFLoader(const std::shared_ptr<rclcpp::Node>& node, const std::strin
 
   auto start = node->now();
 
-  urdf_string_ =
-      urdf_ssp_.loadInitialValue(node, ros_name, std::bind(&RDFLoader::urdfUpdateCallback, this, std::placeholders::_1),
```
DLu, line 70: I'm missing the context a little, but why did you remove the callbacks?

moveit_ros/planning/rdf_loader/include/moveit/rdf_loader/synchronized_string_parameter.h
tylerjw: There is no non-boost replacement for the spsc queue.  The plan to remove boost is to remove the parts of boost that have been migrated into the stl.  The spsc_queue is not one of those things.

moveit_ros/planning/rdf_loader/include/moveit/rdf_loader/synchronized_string_parameter.h
tylerjw: The reason for the lock is that pushing into a spsc queue is not re-entrant safe.  Because there are many threads that can be run by a ROS executor and there is the possibility of the callback being called on two threads at once you need to protect pushing into the spsc_queue.

moveit_ros/planning/rdf_loader/src/synchronized_string_parameter.cpp
```diff
@@ -150,6 +152,8 @@ void SynchronizedStringParameter::stringCallback(const std_msgs::msg::String::Sh
   {
     parent_callback_(msg->data);
   }
-  content_ = msg->data;
+  queue_mutex_.lock();
+  queue_.push(msg->data);
+  queue_mutex_.unlock();
```
tylerjw, line 157: @DLu here is where the callback for the message is.  The problem before is there was a race condition between the two places that were subscribing to the topic.

tylerjw: @corycrean would you mind using a lock_guard here instead of the manual lock and unlock functions?

moveit_ros/planning/rdf_loader/src/synchronized_string_parameter.cpp
```diff
@@ -123,19 +127,17 @@ bool SynchronizedStringParameter::waitForMessage(const rclcpp::Duration timeout)
       name_, rclcpp::QoS(1).transient_local().reliable(),
       std::bind(&SynchronizedStringParameter::stringCallback, this, std::placeholders::_1));
 
-  rclcpp::WaitSet wait_set;
-  wait_set.add_subscription(string_subscriber_);
+  auto timeout_ns =
+      std::chrono::duration_cast<std::chrono::nanoseconds>(timeout.to_chrono<std::chrono::duration<double>>());
```
tylerjw, line 131: Can this be simplified to just this:

```suggestion
  auto timeout_ns =
      timeout.to_chrono<std::chrono::duration<std::chrono::nanoseconds>>();
```

tylerjw: Or maybe even simpler (idk if this works):

```suggestion
  auto timeout_ns = timeout.to_chrono<std::chrono::nanoseconds>();
```

moveit_ros/planning/rdf_loader/src/rdf_loader.cpp
corycrean: It looked to me like the callback was responsible for:
1. Saving the `urdf_string_`
2. Calling `loadFromStrings`

but those things are already done in the constructor, so I thought the callbacks could be safely removed.  I'm noticing now that the callbacks also call `new_model_cb_()` -- should this be added to the constructor as well?

Also, calling `loadFromStrings` from `urdfUpdateCallback`, before `srdf_string_` was loaded, produced an error message https://github.com/ros-planning/moveit2/blob/2ee83c31e107ff71f4469fe0d048eb18606c8e93/moveit_ros/planning/rdf_loader/src/rdf_loader.cpp#L106 which was harmless but a little confusing, and removing the callbacks got rid of that error message as well.

moveit_ros/planning/rdf_loader/src/synchronized_string_parameter.cpp
```diff
@@ -123,19 +127,17 @@ bool SynchronizedStringParameter::waitForMessage(const rclcpp::Duration timeout)
...
   {
-    std_msgs::msg::String msg;
-    rclcpp::MessageInfo info;
-    if (string_subscriber_->take(msg, info))
+    if (queue_.pop(content_))
     {
-      content_ = msg.data;
       return true;
     }
+    std::this_thread::sleep_for(0.01 * timeout_ns);
```
tylerjw, line 140: No magic numbers.  Make this a constant at the top of the file like this:

```suggestion
    // fraction of timeout used for period of polling loop 
    constexpr auto POLL_RATE = 0.01;
    ... 
    std::this_thread::sleep_for(POLL_RATE * timeout_ns);
```

moveit_ros/planning/rdf_loader/src/synchronized_string_parameter.cpp
tylerjw: On line 146 the shared data in content_ is accessed.  To avoid a separate race condition I would move that test to the code that pops from the queue.  The whole point of using a queue like this is to keep access to the variable content_ on a single thread.

corycrean:
It looks like my implementation here fails when `node_` isn't spinning when the `RDFLoader` is created.  (The original implementation worked fine in that case, but had a race condition when `node_` _was_ spinning.)  I think there's a simpler way to handle both cases, which I implemented in https://github.com/ros-planning/moveit2/pull/1050, so I'm closing this PR.

PR #1047, opened 2022-01-29
Don't cache docker builds

tylerjw:
### Description

@henningkayser I think this is what is causing CI to fail.  The docker image caching is not being disabled and the ros2_control packages are still in /opt/ros even though they are in the repos file.  I think maybe we should just not cache our docker builds.

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -55,7 +55,7 @@ jobs:
           file: .docker/${{ github.job }}/Dockerfile
           build-args: ROS_DISTRO=${{ matrix.ROS_DISTRO }}
           push: true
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yaml
```diff
@@ -5,6 +5,7 @@ on:
     # 5 PM UTC every Sunday
     - cron:  '0 17 * * 6'
   workflow_dispatch:
+  pull_request:
```
vatanaksoytezer, line 8: Don't forget to remove this, when we are merging.

vatanaksoytezer (commented):
So this should push dockers to GH and DH, then we should re-run the CI so it passes. You can also remove https://github.com/ros-planning/moveit2/blob/main/.github/workflows/docker.yaml#L27-L36 bits and anything with `steps.apt.outputs.no_cache`

PR #1048, opened 2022-01-29
Don't cache docker builds

tylerjw:
### Description

Disable cache in docker builds.  Pushed into the main repo to rebuild docker image with secrets.

I believe this is the reason the hybrid planner test is failing on main.  We were caching the docker builds and the ros2_control packages were still in /opt/ros in the docker images even though we added them to our repos file.


vatanaksoytezer (approved):
This PR will make docker built without caching always. I am not too worried for that honestly as it only saves a few minutes in an action that only runs on schedule and pushes to main.

PR #1050, opened 2022-01-31
Fix race condition in SynchronizedStringParameter::waitForMessage

corycrean:
- Add a test for RDFLoader where the node is spinning in a different thread.
- Detach the thread.
- Create a new node in SynchronizedStringParameter and use it to create subscriptions.

### Description

This PR addresses the issue in https://github.com/ros-planning/moveit2_tutorials/pull/262 where the robot model can't be loaded.  It also represents an improvement on https://github.com/ros-planning/moveit2/pull/1039.  The implementation in https://github.com/ros-planning/moveit2/pull/1039 only worked in the case where `node_` was spinning in a separate thread, and failed when `node_` was not spinning.  The original implementation worked well when `node_` was not spinning, but had a race condition when it was spinning.

`SynchronizedStringParameter` has no control over whether or not `node_` is spinning, so in this PR I created a new node in `SynchronizedStringParamater::waitForMessage`, and used it to create a subscription.  This way, the race condition is eliminated, and we can be sure that `temp_node` isn't spinning.

### Checklist

### DIFF
--- a/moveit_ros/planning/rdf_loader/src/synchronized_string_parameter.cpp
+++ b/moveit_ros/planning/rdf_loader/src/synchronized_string_parameter.cpp
@@ -119,7 +119,8 @@ bool SynchronizedStringParameter::shouldPublish()
 
 bool SynchronizedStringParameter::waitForMessage(const rclcpp::Duration timeout)
 {
-  string_subscriber_ = node_->create_subscription<std_msgs::msg::String>(
+  auto temp_node = std::make_shared<rclcpp::Node>("synchronized_string_parameter");
+  string_subscriber_ = temp_node->create_subscription<std_msgs::msg::String>(
       name_, rclcpp::QoS(1).transient_local().reliable(),
       std::bind(&SynchronizedStringParameter::stringCallback, this, std::placeholders::_1));
 
--- a/moveit_ros/planning/rdf_loader/src/synchronized_string_parameter.cpp
+++ b/moveit_ros/planning/rdf_loader/src/synchronized_string_parameter.cpp
@@ -119,7 +119,7 @@ bool SynchronizedStringParameter::shouldPublish()
 
 bool SynchronizedStringParameter::waitForMessage(const rclcpp::Duration timeout)
 {
-  auto temp_node = std::make_shared<rclcpp::Node>("synchronized_string_parameter");
+  auto const temp_node = std::make_shared<rclcpp::Node>("synchronized_string_parameter");
   string_subscriber_ = temp_node->create_subscription<std_msgs::msg::String>(
       name_, rclcpp::QoS(1).transient_local().reliable(),
       std::bind(&SynchronizedStringParameter::stringCallback, this, std::placeholders::_1));
--- a/moveit_ros/planning/rdf_loader/src/synchronized_string_parameter.cpp
+++ b/moveit_ros/planning/rdf_loader/src/synchronized_string_parameter.cpp
@@ -119,7 +119,7 @@ bool SynchronizedStringParameter::shouldPublish()
 
 bool SynchronizedStringParameter::waitForMessage(const rclcpp::Duration timeout)
 {
-  std::string const nd_name = std::string(node_->get_name()).append("_ssp_").append(name_);
+  auto const nd_name = std::string(node_->get_name()).append("_ssp_").append(name_);
   auto const temp_node = std::make_shared<rclcpp::Node>(nd_name);
   string_subscriber_ = temp_node->create_subscription<std_msgs::msg::String>(
       name_, rclcpp::QoS(1).transient_local().reliable(),
--- a/moveit_ros/planning/rdf_loader/src/synchronized_string_parameter.cpp
+++ b/moveit_ros/planning/rdf_loader/src/synchronized_string_parameter.cpp
@@ -119,7 +119,8 @@ bool SynchronizedStringParameter::shouldPublish()
 
 bool SynchronizedStringParameter::waitForMessage(const rclcpp::Duration timeout)
 {
-  auto const temp_node = std::make_shared<rclcpp::Node>("synchronized_string_parameter");
+  std::string const nd_name = std::string(node_->get_name()).append("_ssp_").append(name_);
+  auto const temp_node = std::make_shared<rclcpp::Node>(nd_name);
   string_subscriber_ = temp_node->create_subscription<std_msgs::msg::String>(
       name_, rclcpp::QoS(1).transient_local().reliable(),
       std::bind(&SynchronizedStringParameter::stringCallback, this, std::placeholders::_1));
--- a/moveit_ros/planning/rdf_loader/test/test_rdf_integration.cpp
+++ b/moveit_ros/planning/rdf_loader/test/test_rdf_integration.cpp
@@ -66,6 +66,29 @@ TEST(RDFIntegration, topic_based)
   EXPECT_EQ("gonzo", loader.getSRDF()->getName());
 }
 
+TEST(RDFIntegration, executor)
+{
+  rclcpp::Node::SharedPtr node = std::make_shared<rclcpp::Node>("executor");
+
+  // Create a thread to spin an Executor.
+  std::promise<void> promise;
+  auto shared_future = promise.get_future().share();
+  auto thread = std::thread([node, shared_future]() {
+    rclcpp::executors::SingleThreadedExecutor executor;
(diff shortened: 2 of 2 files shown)


moveit_ros/planning/rdf_loader/src/synchronized_string_parameter.cpp
```diff
@@ -119,7 +119,8 @@ bool SynchronizedStringParameter::shouldPublish()
 
 bool SynchronizedStringParameter::waitForMessage(const rclcpp::Duration timeout)
 {
-  string_subscriber_ = node_->create_subscription<std_msgs::msg::String>(
+  auto temp_node = std::make_shared<rclcpp::Node>("synchronized_string_parameter");
+  string_subscriber_ = temp_node->create_subscription<std_msgs::msg::String>(
```
tylerjw, line 123: Two things, this should probably be a const variable.  Secondly, I am worried about name collisions with that name.  I tried to find documentation on if there is some way to get a unique random name in ros with something like what you can do with topics.  I couldn't find any documentation on that.

```suggestion
  auto const temp_node = std::make_shared<rclcpp::Node>("synchronized_string_parameter");
  string_subscriber_ = temp_node->create_subscription<std_msgs::msg::String>(
```

tylerjw (commented):
This generally looks like a much nicer solution than the other one as it fixes the race condition by removing the precondition on the passed-in node.  Should the node be removed from the interface for RDFLoader or is it used for something else?

moveit_ros/planning/rdf_loader/src/synchronized_string_parameter.cpp
corycrean: ROS 1 had a way to create an "anonymous node," but I don't see a way to do it in ROS 2.  But I changed the name to something more unique.

corycrean:
The SynchronizedStringParameter still needs the node in order to check if the string it's looking for was specified as a parameter, and to (optionally) publish it on a new topic.

moveit_ros/planning/rdf_loader/src/synchronized_string_parameter.cpp
```diff
@@ -119,7 +119,9 @@ bool SynchronizedStringParameter::shouldPublish()
 
 bool SynchronizedStringParameter::waitForMessage(const rclcpp::Duration timeout)
 {
-  string_subscriber_ = node_->create_subscription<std_msgs::msg::String>(
+  std::string const nd_name = std::string(node_->get_name()).append("_ssp_").append(name_);
```
tylerjw, line 122: Use auto, you don't need to specify that this is a string twice on the same line.

```suggestion
  auto const nd_name = std::string(node_->get_name()).append("_ssp_").append(name_);
```

moveit_ros/planning/rdf_loader/test/test_rdf_integration.cpp
```diff
@@ -66,6 +66,28 @@ TEST(RDFIntegration, topic_based)
...
+  rclcpp::Node::SharedPtr node = std::make_shared<rclcpp::Node>("executor");
+
+  // Create a thread to spin an Executor.
+  std::promise<void> promise;
+  auto shared_future = promise.get_future().share();
+  std::thread([node, shared_future]() {
+    rclcpp::executors::SingleThreadedExecutor executor;
+    executor.add_node(node);
+    executor.spin_until_future_complete(shared_future, std::chrono::seconds(10));
+  }).detach();
```
tylerjw, line 80: Is there a reason you detach this thread instead of joining with it after you set the value of the promise?  Also, you can move the initialization of the shared future into the capture if you'd like as you don't need it outside of the lambda.

```suggestion
  std::thread([node, shared_future = promise.get_future().share()]() {
    rclcpp::executors::SingleThreadedExecutor executor;
    executor.add_node(node);
    executor.spin_until_future_complete(shared_future, std::chrono::seconds(10));
  }).detach();
```

tylerjw: After reading more about it, using detach is just fine.  There really is no difference and using it here simplifies the test which is nice.

moveit_ros/planning/rdf_loader/test/test_rdf_integration.cpp
```diff
@@ -66,6 +66,28 @@ TEST(RDFIntegration, topic_based)
   EXPECT_EQ("gonzo", loader.getSRDF()->getName());
 }
 
+TEST(RDFIntegration, executor)
+{
+  rclcpp::Node::SharedPtr node = std::make_shared<rclcpp::Node>("executor");
```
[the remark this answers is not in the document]
corycrean: The promise and future should have been removed.  I was switching from `spin_until_future_complete` to `detach` and I didn't finish cleaning up.

moveit_ros/planning/rdf_loader/test/test_rdf_integration.cpp
```diff
@@ -66,6 +66,25 @@ TEST(RDFIntegration, topic_based)
...
+    executor.add_node(node);
+    executor.spin();
+  }).detach();
+
+  rdf_loader::RDFLoader loader(node, "topic_description");
+
+  ASSERT_NE(nullptr, loader.getURDF());
+  EXPECT_EQ("gonzo", loader.getURDF()->name_);
+  ASSERT_NE(nullptr, loader.getSRDF());
+  EXPECT_EQ("gonzo", loader.getSRDF()->getName());
```
tylerjw, line 85: Would you mind adding some comments in this format:

```
// RDFLoader should <describe behavior at a high level>
// GIVEN <precondition>
// WHEN <the thing you do in the test>
// THEN <what you expect to be true>
```

moveit_ros/planning/rdf_loader/test/test_rdf_integration.cpp
```diff
@@ -66,6 +66,30 @@ TEST(RDFIntegration, topic_based)
   EXPECT_EQ("gonzo", loader.getSRDF()->getName());
 }
 
+// RDFLoader should successfully load URDF and SRDF strings from a ROS topic when the node that is passed to it is
+// spinning.
+// GIVEN a node that has been added to an executor that is spinning on another thread
+// WHEN the RDFLoader is created
+// THEN the RDFLoader should return non-null values for the URDF and SRDF model.
```
tylerjw, line 73: I'm sorry to keep asking for changes but can you put these comments in the body of the test.  The first two lines go at the beginning and the WHEN and THEN lines go right before the respective parts.

moveit_ros/planning/rdf_loader/test/test_rdf_integration.cpp
```diff
@@ -66,6 +66,30 @@ TEST(RDFIntegration, topic_based)
...
+    executor.spin();
+  }).detach();
+
+  rdf_loader::RDFLoader loader(node, "topic_description");
+
+  ASSERT_NE(nullptr, loader.getURDF());
+  EXPECT_EQ("gonzo", loader.getURDF()->name_);
+  ASSERT_NE(nullptr, loader.getSRDF());
+  EXPECT_EQ("gonzo", loader.getSRDF()->getName());
+}
```
tylerjw, line 91:
```suggestion
TEST(RDFIntegration, executor)
{
  // RDFLoader should successfully load URDF and SRDF strings from a ROS topic when the node that is 
  // passed to it is spinning.
  // GIVEN a node that has been added to an executor that is spinning on another thread
  rclcpp::Node::SharedPtr node = std::make_shared<rclcpp::Node>("executor");

  // Create a thread to spin an Executor.
  std::thread([node]() {
    rclcpp::executors::SingleThreadedExecutor executor;
    executor.add_node(node);
    executor.spin();
  }).detach();

  // WHEN the RDFLoader is created
  rdf_loader::RDFLoader loader(node, "topic_description");

  // THEN the RDFLoader should return non-null values for the URDF and SRDF model.
  ASSERT_NE(nullptr, loader.getURDF());
  EXPECT_EQ("gonzo", loader.getURDF()->name_);
  ASSERT_NE(nullptr, loader.getSRDF());
  EXPECT_EQ("gonzo", loader.getSRDF()->getName());
}
```

tylerjw: note that I did not run the formatted on this... I don't know if I got the line lengths right.  I was just trying to make it more clear where I would like the comments.

tylerjw (approved):

PR #1051, opened 2022-01-31
Disable hybrid planning test, don't cache ci docker at all and don't cache upstream_ws if repos file is changed

vatanaksoytezer:
Just testing out if CI failures are due to upstream workspace caching.

Edit: So if we disable docker and upstream caching together CI issues seem to resolve well. I've written an action that gets the latest edit timestamp of the repos file and generates a new cache if the repos has been changes. By utilizing that we should be able to achieve upstream caching and not worry about broken CIs every time the .repos file change.

Edit 2: It turns out we have flaky behaviour (likely due to ros2_control), I am going to disable hybrid planning test for now and let the CI pass to unblock all the PRs waiting for this. I am going to create a high priority issue to resolve this as soon as possible. I am suspecting a race condition because the failures are not consistent and happens ~50% of the time.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -79,6 +79,9 @@ jobs:
         uses: JafarAbdi/latest-rosdistro-release-date-action@main
         with:
           rosdistro: ${{ matrix.env.ROS_DISTRO }}
+      - name: Check directory
+        run: |
+          ls -la
       - name: Get latest timestamp repos file has been edited
         id: repos_timestamp
         uses: vatanaksoytezer/latest-file-edit-timestamp-action@main
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -79,14 +79,14 @@ jobs:
         uses: JafarAbdi/latest-rosdistro-release-date-action@main
         with:
           rosdistro: ${{ matrix.env.ROS_DISTRO }}
-      - name: Cache upstream workspace
-        uses: pat-s/always-upload-cache@v2.1.5
-        with:
-          path: ${{ env.BASEDIR }}/upstream_ws
-          key: ${{ env.CACHE_PREFIX }}-${{ github.run_id }}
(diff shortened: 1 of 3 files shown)


AndyZe:
It's great to see CI passing again!

.github/workflows/ci.yaml
```diff
@@ -79,14 +79,19 @@ jobs:
...
+          file: moveit2.repos
       - name: Cache upstream workspace
         uses: pat-s/always-upload-cache@v2.1.5
         with:
           path: ${{ env.BASEDIR }}/upstream_ws
           key: ${{ env.CACHE_PREFIX }}-${{ github.run_id }}
           restore-keys: ${{ env.CACHE_PREFIX }}
         env:
-          CACHE_PREFIX: ${{ steps.rosdistro_release_date.outputs.date }}-upstream_ws-${{ matrix.env.IMAGE }}-${{ hashFiles('moveit2*.repos', '.github/workflows/ci.yaml') }}
+          CACHE_PREFIX: ${{ steps.rosdistro_release_date.outputs.date }}-upstream_ws-${{ steps.repos_timestamp.outputs.timestamp }}-${{ matrix.env.IMAGE }}-${{ hashFiles('moveit2*.repos', '.github/workflows/ci.yaml') }}
```
AndyZe, line 94: A comment about what this does would be great.

AndyZe (approved):

.github/workflows/ci.yaml
vatanaksoytezer: It's in the name of the job: https://github.com/ros-planning/moveit2/blob/vatan/fix_ci/.github/workflows/ci.yaml#L82. I don't mind adding more comment if you think is needed though.

vatanaksoytezer: I updated the name to be slightly more verbose: `repos_timestamp` to `repos_edit_timestamp`

AndyZe: I see. That works for me.

PR #1052, opened 2022-01-31
Ensure global/local planner planning scene topics are configurable

AndyZe:
We are trying to use hybrid planning with a dual-arm system and have noticed that planning scene parameters weren't handled well. They just took the default values. This should take care of that.

This should not change how the demo works in any way. (https://moveit.picknik.ai/main/doc/examples/hybrid_planning/hybrid_planning_tutorial.html)

### DIFF
--- a/moveit_ros/hybrid_planning/global_planner/global_planner_plugins/src/moveit_planning_pipeline.cpp
+++ b/moveit_ros/hybrid_planning/global_planner/global_planner_plugins/src/moveit_planning_pipeline.cpp
@@ -67,6 +67,15 @@ bool MoveItPlanningPipeline::initialize(const rclcpp::Node::SharedPtr& node)
   node->declare_parameter<double>(PLAN_REQUEST_PARAM_NS + "max_acceleration_scaling_factor", 1.0);
   node->declare_parameter<std::string>("ompl.planning_plugin", "ompl_interface/OMPLPlanner");
 
+  // Planning Scene options
+  node->declare_parameter<std::string>(PLANNING_SCENE_MONITOR_NS + "name", UNDEFINED);
+  node->declare_parameter<std::string>(PLANNING_SCENE_MONITOR_NS + "robot_description", UNDEFINED);
+  node->declare_parameter<std::string>(PLANNING_SCENE_MONITOR_NS + "joint_state_topic", UNDEFINED);
+  node->declare_parameter<std::string>(PLANNING_SCENE_MONITOR_NS + "attached_collision_object_topic", UNDEFINED);
+  node->declare_parameter<std::string>(PLANNING_SCENE_MONITOR_NS + "publish_planning_scene_topic", UNDEFINED);
+  node->declare_parameter<std::string>(PLANNING_SCENE_MONITOR_NS + "monitored_planning_scene_topic", UNDEFINED);
+  node->declare_parameter<double>(PLANNING_SCENE_MONITOR_NS + "wait_for_initial_state_timeout", 10.0);
+
   // Trajectory Execution Functionality (required by the MoveItPlanningPipeline but not used within hybrid planning)
   node->declare_parameter<std::string>("moveit_controller_manager", UNDEFINED);
 
--- a/moveit_ros/hybrid_planning/local_planner/local_planner_component/include/moveit/local_planner/local_planner_component.h
+++ b/moveit_ros/hybrid_planning/local_planner/local_planner_component/include/moveit/local_planner/local_planner_component.h
@@ -118,6 +118,10 @@ public:
       declareOrGetParam<std::string>("local_solution_topic_type", local_solution_topic_type, undefined, node);
       declareOrGetParam<bool>("publish_joint_positions", publish_joint_positions, false, node);
       declareOrGetParam<bool>("publish_joint_velocities", publish_joint_velocities, false, node);
+      // Planning scene monitor options
+      declareOrGetParam<std::string>("monitored_planning_scene", monitored_planning_scene_topic, undefined, node);
+      declareOrGetParam<std::string>("collision_object_topic", collision_object_topic, undefined, node);
+      declareOrGetParam<std::string>("joint_states_topic", joint_states_topic, undefined, node);
     }
 
     std::string group_name;
@@ -133,6 +137,9 @@ public:
     bool publish_joint_positions;
     bool publish_joint_velocities;
     double local_planning_frequency;
+    std::string monitored_planning_scene_topic;
+    std::string collision_object_topic;
+    std::string joint_states_topic;
   };
 
   /** \brief Constructor */
--- a/moveit_ros/hybrid_planning/test/config/global_planner.yaml
+++ b/moveit_ros/hybrid_planning/test/config/global_planner.yaml
@@ -7,9 +7,9 @@ planning_scene_monitor_options:
   joint_state_topic: "/joint_states"
   attached_collision_object_topic: "/moveit_cpp/planning_scene_monitor"
   # Subscribe to this topic (The name comes from the perspective of moveit_cpp)
-  publish_planning_scene_topic: "/moveit_cpp/publish_planning_scene"
+  publish_planning_scene_topic: "/planning_scene"
   # Publish this topic, e.g. to visualize with RViz
-  monitored_planning_scene_topic: "/moveit_cpp/monitored_planning_scene"
+  monitored_planning_scene_topic: "/global_planner/planning_scene"
   wait_for_initial_state_timeout: 10.0
 planning_pipelines:
   #namespace: "moveit_cpp"  # optional, default is ~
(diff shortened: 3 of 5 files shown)


moveit_ros/hybrid_planning/global_planner/global_planner_plugins/src/moveit_planning_pipeline.cpp
```diff
@@ -67,6 +67,15 @@ bool MoveItPlanningPipeline::initialize(const rclcpp::Node::SharedPtr& node)
   node->declare_parameter<double>(PLAN_REQUEST_PARAM_NS + "max_acceleration_scaling_factor", 1.0);
   node->declare_parameter<std::string>("ompl.planning_plugin", "ompl_interface/OMPLPlanner");
 
+  // Planning Scene options
```
AndyZe, line 70: These parameters were already present in `test/config/global_planner.yaml`, they just weren't being used anywhere.

moveit_ros/hybrid_planning/test/config/global_planner.yaml
```diff
@@ -6,8 +6,10 @@ planning_scene_monitor_options:
   robot_description: "robot_description"
   joint_state_topic: "/joint_states"
   attached_collision_object_topic: "/moveit_cpp/planning_scene_monitor"
-  publish_planning_scene_topic: "/moveit_cpp/publish_planning_scene"
-  monitored_planning_scene_topic: "/moveit_cpp/monitored_planning_scene"
+  # Subscribe to this topic (The name comes from the perspective of moveit_cpp)
+  publish_planning_scene_topic: "/planning_scene"
+  # Publish this topic, e.g. to visualize with RViz
+  monitored_planning_scene_topic: "/global_planner/planning_scene"
```
AndyZe, line 12: The naming of these parameters is atrocious from the perspective of the hybrid planner. I'm not sure what can be done about it since the parameters get passed to moveit_cpp.

AndyZe:
@henningkayser pinging for review

moveit_ros/hybrid_planning/global_planner/global_planner_plugins/src/moveit_planning_pipeline.cpp
tylerjw: What if these were already declared?  should this check to see if the parameters exist before declaring them?

Should we give them defaults here that are not UNDEFINED?

moveit_ros/hybrid_planning/local_planner/local_planner_component/include/moveit/local_planner/local_planner_component.h
```diff
@@ -118,6 +118,10 @@ class LocalPlannerComponent
       declareOrGetParam<std::string>("local_solution_topic_type", local_solution_topic_type, undefined, node);
       declareOrGetParam<bool>("publish_joint_positions", publish_joint_positions, false, node);
       declareOrGetParam<bool>("publish_joint_velocities", publish_joint_velocities, false, node);
+      // Planning scene monitor options
+      declareOrGetParam<std::string>("monitored_planning_scene", monitored_planning_scene_topic, undefined, node);
+      declareOrGetParam<std::string>("collision_object_topic", collision_object_topic, undefined, node);
+      declareOrGetParam<std::string>("joint_states_topic", joint_states_topic, undefined, node);
```
tylerjw, line 124: Same question here about defaults.  Should the defaults for all these be defined somewhere (maybe a header file with a struct that has default values).

tylerjw (commented):
I left a couple of small comments.  As to the name of monitored_planning_scene_topic, do you have a different name you'd propose?

Here is the document we have with a graph that shows what the various topics are that PSM uses: https://moveit.picknik.ai/galactic/doc/concepts/planning_scene_monitor.html

moveit_ros/hybrid_planning/global_planner/global_planner_plugins/src/moveit_planning_pipeline.cpp
henningkayser: Can't you just use the PlanningSceneMonitorOptions? https://github.com/ros-planning/moveit2/blob/main/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h#L57

moveit_ros/hybrid_planning/test/config/global_planner.yaml
henningkayser: I think it would make sense to use the MoveItCpp options directly and load them with a moveit_cpp.yaml. But you are right, those parameters are confusing

AndyZe: Moving the parameters to a separate `moveit_cpp.yaml` makes a lot of sense as well as reducing LOC. Done.

Yep, those parameters will still look confusing if somebody does `ros2 param list`.

AndyZe: Actually I'm not sure that will work. I have to test if the parameters are read properly or if it just uses defaults.

moveit_ros/hybrid_planning/global_planner/global_planner_plugins/src/moveit_planning_pipeline.cpp
AndyZe: PlanningSceneMonitorOptions uses get_parameter_or(). Is that even valid without declaring the parameters first? Or is it going to use the default every time?

https://github.com/ros-planning/moveit2/blob/77e43f339cb6ee06679a424e21c72301797920d4/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h#L62

moveit_ros/hybrid_planning/test/config/global_planner.yaml
AndyZe: @henningkayser I tried this in commit [d2cf1b3](https://github.com/ros-planning/moveit2/pull/1052/commits/d2cf1b321fcafbba85208730c9c3824104f3c9f7). You can quickly verify that it doesn't work:

- Modify one of the topics in global_planner_moveit_cpp.yaml, for example, like this:  `monitored_planning_scene_topic: "test/global_planner/planning_scene"`
- Launch the hybrid planning demo:  `ros2 launch moveit_hybrid_planning hybrid_planning_demo.launch.py`
- List the subscriptions of the global planner:  `ros2 node info /global_planner_private_`

You will see that it does not subscribe to the right topic.

I'm going to revert that commit

moveit_ros/hybrid_planning/local_planner/local_planner_component/include/moveit/local_planner/local_planner_component.h
AndyZe: MoveIt already does have that header file. I don't think we should use it, though -- I think we should enforce correct yaml.

(Using defaults leads to issues. See Henning's previous comment about using MoveItCppOptions and my response.)

tylerjw: You don't have to provide a default when you call declare_parameter.

moveit_ros/hybrid_planning/test/config/global_planner.yaml
tylerjw: I'm going to try to write some small tests to try to understand how declaring and overriding parameters work.  I think the ros2 interface is confusing and I don't know what the right answer is.

moveit_ros/hybrid_planning/local_planner/local_planner_component/include/moveit/local_planner/local_planner_component.h
AndyZe: I tried something like this. It seems nice for several reasons but it crashes.

```
void DeclareParam(const std::string& param_name, const rclcpp::Node::SharedPtr& node)
{
  try
  {
    if (node->has_parameter(param_name))
    {
      rcl_interfaces::msg::ParameterDescriptor descriptor;
      descriptor.dynamic_typing = true;
      node->declare_parameter(param_name, rclcpp::ParameterValue{}, descriptor);
    }
    else
      RCLCPP_ERROR_STREAM(node->get_logger(),
                          "Error getting parameter '" << param_name << "', check parameter type in YAML file");
  }
  catch (const rclcpp::exceptions::InvalidParameterTypeException& e)
  {
    RCLCPP_ERROR_STREAM(node->get_logger(),
                        "Error getting parameter '" << param_name << "', check parameter type in YAML file");
    throw e;
  }
}

std::string param_name = PLANNING_PIPELINES_NS + "pipeline_names";
DeclareParam(param_name, node);  --> Crashes because no default value is provided, I suppose
```

moveit_ros/hybrid_planning/test/config/global_planner.yaml
AndyZe: Can we get this merged and make an issue in the meantime? It's better than the previous state because at least these parameters are being used.

moveit_ros/hybrid_planning/local_planner/local_planner_component/include/moveit/local_planner/local_planner_component.h
tylerjw: has_parameter returns true if you have already declared the parameter, if you then call declare_parameter then it will throw an exception because it is already declared.

AndyZe: That's a good point, maybe it will work without that. I'll try again.

PR #1064, opened 2022-02-07
[hybrid planning] Add a way for local constraint solver to pause waypoint retrieval

AndyZe:
**Background on how hybrid planning works:**

- simple_sampler.cpp selects the next target waypoint. The only criteria on this is a tolerance on distance:
```

  if (next_desired_goal_state.distance(current_state, joint_group_) <= WAYPOINT_RADIAN_TOLERANCE)
  {
    // Update index (and thus desired robot state)
    next_waypoint_index_ = std::min(next_waypoint_index_ + 1, reference_trajectory_->getWayPointCount() - 1);
  }
```

- forward_trajectory.cpp checks if the local trajectory is valid, including collision:

`is_path_valid = locked_planning_scene->isPathValid(local_trajectory, local_trajectory.getGroupName(), false);`

**The issue**

The problem here is, `simple_sampler.cpp` only regards the radian tolerance. It doesn't know anything about the collision. So, it is possible for simple_sampler to continue selecting future waypoints even if they would be in collision. If the collision ever clears up, this can result in a sudden jump of the robot.

This PR adds a way for `forward_trajectory` to communicate back to `simple_sampler` that it should stop moving forward to a new waypoint.

### DIFF
--- a/moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
+++ b/moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
@@ -304,11 +304,18 @@ void LocalPlannerComponent::executeIteration()
       *local_planner_feedback_ = local_constraint_solver_instance_->solve(
           local_trajectory, local_planning_goal_handle_->get_goal(), local_solution);
 
-      // Feedback is only send when the hybrid planning architecture should react to a discrete event
+      // Feedback is only sent when the hybrid planning architecture should react to a discrete event
       if (!local_planner_feedback_->feedback.empty())
       {
+        // Prevent the SimpleSampler from moving to the next waypoint, e.g. if a collision is ahead
+        trajectory_operator_instance_->preventForwardProgress();
+
         local_planning_goal_handle_->publish_feedback(local_planner_feedback_);
         return;
+      }
+      else
+      {
+        trajectory_operator_instance_->allowForwardProgress();
       }
 
       // Use a configurable message interface like MoveIt servo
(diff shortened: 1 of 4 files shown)


moveit_ros/hybrid_planning/local_planner/trajectory_operator_plugins/include/moveit/trajectory_operator_plugins/simple_sampler.h
```diff
@@ -51,19 +51,28 @@ class SimpleSampler : public TrajectoryOperatorInterface
...
+
+  void allowForwardProgress() override;
+
 private:
   std::size_t
       next_waypoint_index_;  // Indicates which reference trajectory waypoint is the current local goal constrained
   moveit_msgs::action::LocalPlanner::Feedback feedback_;  // Empty feedback
   trajectory_processing::TimeOptimalTrajectoryGeneration time_parametrization_;
   const moveit::core::JointModelGroup* joint_group_;
+  std::atomic<bool> prevent_forward_progress_;  // Flag to prevent moving to the next waypoint
```
henningkayser, line 76: I don't see atomic features being used at all, a normal bool would suffice here imo.

moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
```diff
@@ -277,12 +277,19 @@ void LocalPlannerComponent::executeIteration()
       *local_planner_feedback_ = local_constraint_solver_instance_->solve(
           local_trajectory, local_planning_goal_handle_->get_goal(), local_solution);
 
-      // Feedback is only send when the hybrid planning architecture should react to a discrete event
+      // Feedback is only sent when the hybrid planning architecture should react to a discrete event
       if (!local_planner_feedback_->feedback.empty())
```
henningkayser, line 281: This looks like a pretty broad assumption. Should the feedback be handled in any way?

henningkayser (commented):
Sorry for letting this get stale. I don't see anything wrong with this addition. Besides this change, I'm wondering if halting/proceeding should be handled more explicitly.

tylerjw:
@henningkayser @AndyZe what's the status of this?

tylerjw:
This work seems to be un-finished and no one is currently working on it. I'll create an issue to track finishing this work.

PR #1065, opened 2022-02-07
[hybrid planning] Use a map of expected feedback codes

AndyZe:
Get rid of hard-coded string constants

### DIFF
--- a/moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/forward_trajectory.cpp
+++ b/moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/forward_trajectory.cpp
@@ -118,10 +118,10 @@ ForwardTrajectory::solve(const robot_trajectory::RobotTrajectory& local_trajecto
     {
       if (!path_invalidation_event_send_)
       {  // Send feedback only once
-        feedback_result.feedback = LOCAL_FEEDBACK_MAP.at(LocalFeedbackEnum::COLLISION_AHEAD);
+        feedback_result.feedback = FeedbackEnumToString(LocalFeedbackEnum::COLLISION_AHEAD);
         path_invalidation_event_send_ = true;  // Set feedback flag
       }
-      RCLCPP_INFO(LOGGER, "Collision ahead, hold current position");
+      RCLCPP_INFO(LOGGER, "Collision ahead, holding current position");
       // Keep current position
       moveit::core::RobotState current_state_command(*current_state);
       if (current_state_command.hasVelocities())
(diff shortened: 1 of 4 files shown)


tylerjw (approved):
This is straightforward and looks good.

One alternate approach you might consider to the map which has to be evaluated at runtime is a constexpr function that takes the enum type and returns the string, that way you can get the performance of the string being in the code with also defining it in only one place.

AndyZe:
> One alternate approach you might consider to the map which has to be evaluated at runtime is a constexpr function that takes the enum type and returns the string, that way you can get the performance of the string being in the code with also defining it in only one place.

Would that function go in the new `feedback_types.h` header file?

tylerjw:
> > One alternate approach you might consider to the map which has to be evaluated at runtime is a constexpr function that takes the enum type and returns the string, that way you can get the performance of the string being in the code with also defining it in only one place.
> 
> Would that function go in the new `feedback_types.h` header file?

Yes.  Here is an example of me doing nearly the same thing: https://github.com/tylerjw/fp/blob/85ee0e5e6d5445082bfcf464e474a5600f06f2e9/include/fp/result.hpp#L140

tylerjw:
I'm sorry I didn't provide an example before.  What that function does is take an enum and return a string_view.  You don't need to do the string view and could instead return a normal string if that works better for your situation.  string_view is an optimization that is tiny in this case as the strings are so short they are going to be allocated on the stack anyway.

tylerjw (approved):

moveit_ros/hybrid_planning/local_planner/local_planner_component/include/moveit/local_planner/feedback_types.h
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * \brief Expected feedback types
+ */
+enum LocalFeedbackEnum
+{
+  COLLISION_AHEAD = 1,
+  LOCAL_PLANNER_STUCK = 2
+};
+
+[[nodiscard]] constexpr std::string_view FeedbackEnumToString(const LocalFeedbackEnum& code)
```
tylerjw, line 57: one tiny thing you could do is change the function name to just toString because that won't collide with any other functions because this is the only toString that will take a FeedbackEnum type as its parameter.

Then you don't have to type FeedbackEnum twice when you use it.

moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/forward_trajectory.cpp
```diff
@@ -150,7 +151,7 @@ ForwardTrajectory::solve(const robot_trajectory::RobotTrajectory& local_trajecto
         {
           num_iterations_stuck_ = 0;
           prev_waypoint_target_ = nullptr;
-          feedback_result.feedback = "local_planner_stuck";
+          feedback_result.feedback = FeedbackEnumToString(LocalFeedbackEnum::LOCAL_PLANNER_STUCK);
```
tylerjw, line 154: if you change the name of the function this just becomes:

```suggestion
          feedback_result.feedback = toString(LocalFeedbackEnum::LOCAL_PLANNER_STUCK);
```

tylerjw (approved):

PR #1074, opened 2022-02-14
ros_planning: Get kinematic params from move_group node

jbohren:
### Description

This makes it so that the kinematics loader, and therefore the MoveIt RViz plugin can pull the appropriate kinematics parameters from the `move_group` node, so that they don't need to be replicated in RViz's parameter space.

If this is a pattern the maintainers like, I'll clean up formatting and add further documentation. If not, feel free to close.

The main functional change is here:
https://github.com/ros-planning/moveit2/pull/1074/files#diff-25706feb6b355123a5ae63dfa987a1cdd5754087392e9623960ed74a31b56d75R409-R428

### Checklist

### DIFF
--- a/moveit_kinematics/kdl_kinematics_plugin/CMakeLists.txt
+++ b/moveit_kinematics/kdl_kinematics_plugin/CMakeLists.txt
@@ -1,8 +1,15 @@
 set(MOVEIT_LIB_NAME moveit_kdl_kinematics_plugin)
 
+find_package(orocos_kdl REQUIRED)
+
+include_directories(${orocos_kdl_INCLUDE_DIRS})
+link_directories(${orocos_kdl_LIBRARY_DIRS})
+
 add_library(${MOVEIT_LIB_NAME} SHARED
   src/kdl_kinematics_plugin.cpp
   src/chainiksolver_vel_mimic_svd.cpp)
+
+target_link_libraries(${MOVEIT_LIB_NAME} orocos-kdl)
 
 ament_target_dependencies(${MOVEIT_LIB_NAME}
   rclcpp
--- a/moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
+++ b/moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
@@ -38,6 +38,7 @@
 #include <moveit/rdf_loader/rdf_loader.h>
 #include <pluginlib/class_loader.hpp>
 #include <boost/thread/mutex.hpp>
+#include <chrono>
 #include <sstream>
 #include <vector>
 #include <map>
@@ -56,7 +57,7 @@ rclcpp::Parameter declare_parameter(const rclcpp::Node::SharedPtr& node, const s
     node->declare_parameter(parameter_name, ParamType);
   rclcpp::Parameter parameter;
   if (!node->get_parameter(parameter_name, parameter))
-    RCLCPP_DEBUG_STREAM(LOGGER, "Parameter `" << parameter_name << "` doesn't exists");
+    RCLCPP_DEBUG_STREAM(LOGGER, "Parameter `" << parameter_name << "` doesn't exist");
   return parameter;
 }
 class KinematicsPluginLoader::KinematicsLoaderImpl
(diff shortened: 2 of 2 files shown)


jbohren:
@tylerjw what do you think, should this design pattern be encouraged?

moveit_kinematics/kdl_kinematics_plugin/CMakeLists.txt
```diff
@@ -1,9 +1,16 @@
 set(MOVEIT_LIB_NAME moveit_kdl_kinematics_plugin)
 
+find_package(orocos_kdl REQUIRED)
```
henningkayser, line 3: orocos_kdl is already required by the top-level CMakeLists.txt [here](https://github.com/ros-planning/moveit2/blob/main/moveit_kinematics/CMakeLists.txt#L10). I also think that the other changes in this file are redundant, `ament_target_dependencies()` already handles include directories and linked libraries.

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -303,36 +346,100 @@ moveit::core::SolverAllocatorFn KinematicsPluginLoader::getLoaderFunction(const
 
       if (default_solver_plugin_.empty())
       {
-        RCLCPP_DEBUG(LOGGER, "Loading settings for kinematics solvers from the ROS param server ...");
+        // construct a parameter client to get parameters from the move_group node
```
henningkayser, line 349: Wouldn't it make sense to move this into step 3?

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -303,36 +346,100 @@ moveit::core::SolverAllocatorFn KinematicsPluginLoader::getLoaderFunction(const
...
+        using namespace std::chrono_literals;
+        auto parameters_client = std::make_shared<rclcpp::AsyncParametersClient>(node_, "move_group");
+        while (!parameters_client->wait_for_service(1s)) {
```
henningkayser, line 352: I'm not a fan of this kind of busy waiting and shutting down on failure, but I see that the current API doesn't really provide other options. @tylerjw here is another use case for improved error handling ;). For now, I would definitely return the function after calling shutdown.

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -303,36 +346,100 @@ moveit::core::SolverAllocatorFn KinematicsPluginLoader::getLoaderFunction(const
...
+                      });
+
+              RCLCPP_DEBUG(LOGGER, "Waiting for param request future...");
+
+              std::future_status status;
+              do {
+                  status = future.wait_for(1s);
+              } while (status != std::future_status::ready);
+
+              ksolver_param          = future.get()[0];
```
henningkayser, line 425: Better check if the future is valid fist. Also, I think it would be cleaner to check for vector size and use `at()`.

henningkayser (commented):
@jbohren thanks for your PR! In general, I would really like to make more of the parameters support proper syncing, especially the RViz plugins would profit a lot from this. It's definitely a good starting point to discuss design patterns. My only concern with this approach is that MoveItCpp is not supported, just as non-default MoveGroup namespaces. That's not a real blocker, though, some kind of node namespace registration can be added later if needed. From the looks of it, your implementation is also backwards-compatible, right?

**Side note**: I updated your branch to see if CI succeeds (we had some issues with Rolling). Currently clang-format fails, you can fix that using [pre-commit](https://github.com/ros-planning/moveit2/runs/5252745522?check_suite_focus=true#step:5:72).

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
jbohren: This was to avoid creating it in the loop.

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
jbohren: Yeah, ideally this would just defer loading the plugins until the parameters could be found, but all of the control flow up to this point is synchronous and would likely require a much larger re-architecture effort.

jbohren:
> @jbohren thanks for your PR! In general, I would really like to make more of the parameters support proper syncing, especially the RViz plugins would profit a lot from this. 

Yeah, I really just did this for myself so I didn't have to keep restarting it, and I thought I'd share.

> My only concern with this approach is that MoveItCpp is not supported, just as non-default MoveGroup namespaces. That's not a real blocker, though, some kind of node namespace registration can be added later if needed. From the looks of it, your implementation is also backwards-compatible, right?

Yeah, this assumes that this plugin loader is instantiated in a node in the same namespace as the `move_group` node **AND** the `move_group` node hasn't been renamed.

It's probably better to move the parameter client _out_ of this loader and into whatever owns the loader. Then the owner (hopefully the `node`, itself) can address the `move_group` namespace/name and sync the parameters locally. Then the plugin loader doesn't have to communicate externally.

> **Side note**: I updated your branch to see if CI succeeds (we had some issues with Rolling). Currently clang-format fails, you can fix that using [pre-commit](https://github.com/ros-planning/moveit2/runs/5252745522?check_suite_focus=true#step:5:72).

Yeah I saw this when I committed, I thought it would be good enough to discuss the design pattern. It doesn't quite feel like the right implementation though, but this would be fixed before I'd expect anything to be merged.

jbohren:
Funny thing happens when you make it so you don't have to restart rviz2...

2022-02-19-090230_2377x146_scrot

tylerjw:
Rebasing, I'll investigate this and see if it is the approach we'd like to take. I want to simplify the configuration space of MoveIt and make it less complicated for people to use MoveIt with RViz. Here is a discussion I started around configs: https://github.com/ros-planning/moveit2/discussions/1733

tylerjw:
I read through these changes again, and I don't think we want to apply them as is. There are a couple of reasons for that:

- We are not backporting changes to galactic
- We are going through a refactor of how all these parameters are loaded on main with generate_parameter_library

However, I'll add this to my issue for discussing ways to reduce the extra complexity of configs used by MoveIt2 and the rviz plugin. This does address one of the major pain points in that you have to send these parameters to both requiring you to restart rviz when you change parameters and it would be much nicer if you could just leave rviz up.

PR #1077, opened 2022-02-15
Temporarily use galactic in main CI

vatanaksoytezer:
Should unblock us from merging PRs until rolling is migrated to jammy.


henningkayser:
Not sure why #1078 is not working as intended (locally, this seemed to do the trick for me). Anyway, let's stick to Galactic for now until Jammy is fixed. Thanks for preparing this

henningkayser (approved):

PR #1078, opened 2022-02-15
Use focal in industrial_ci

vatanaksoytezer:


.github/workflows/ci.yaml
```diff
@@ -33,6 +37,8 @@ jobs:
       AFTER_SETUP_DOWNSTREAM_WORKSPACE: vcs pull $BASEDIR/downstream_ws/src
       # Clear the ccache stats before and log the stats after the build
       BEFORE_BUILD_UPSTREAM_WORKSPACE: ccache -z
+      BEFORE_INSTALL_UPSTREAM_DEPENDENCIES: export ROSDISTRO_INDEX_URL=https://raw.githubusercontent.com/ros/rosdistro/rolling/2022-01-28/index-v4.yaml
```
henningkayser, line 40: @vatanaksoytezer I think you need to add this before the setup_rosdep job https://github.com/ros-planning/moveit2/runs/5202810205?check_suite_focus=true#step:12:137

.github/workflows/ci.yaml
```diff
@@ -19,10 +19,14 @@ jobs:
           - IMAGE: rolling-ci
             CCOV: true
             ROS_DISTRO: rolling
+            OS_NAME: ubuntu
```
henningkayser, line 22: OS_NAME is redundant here, no?

henningkayser:
@vatanaksoytezer can we close this?

PR #1079, opened 2022-02-17
Off by one in getAverageSegmentDuration

stephanie-eng:
### Description

Corrects an off-by-one error in getAverageSegmentDuration

If we have **n** waypoints, we have **n - 1** timesteps

Assume a trivial case of durations = [1, 2, 3] with a timestamp of 1 and a size of 3

getDuration returns:
3 - 1 = 2

The current return is:
2 / 3 = 0.67

The expected return should be: 
2 / (3 - 1)  = 1

### Checklist

### DIFF
--- a/moveit_core/robot_trajectory/src/robot_trajectory.cpp
+++ b/moveit_core/robot_trajectory/src/robot_trajectory.cpp
@@ -102,7 +102,7 @@ double RobotTrajectory::getAverageSegmentDuration() const
   {
     if (duration_from_previous_.size() <= 1)
     {
-      RCLCPP_WARN(rclcpp::get_logger("RobotTrajectory"), "Too few waypoints to calculate a duration. Returning 0.");
+      RCLCPP_WARN(rclcpp::get_logger("RobotTrajectory"), "First and only waypoint has a duration of 0.");
       return 0.0;
     }
     else
--- a/moveit_core/robot_trajectory/src/robot_trajectory.cpp
+++ b/moveit_core/robot_trajectory/src/robot_trajectory.cpp
@@ -94,7 +94,7 @@ double RobotTrajectory::getAverageSegmentDuration() const
   if (duration_from_previous_.empty())
     return 0.0;
   else
-    return getDuration() / static_cast<double>(duration_from_previous_.size());
+    return getDuration() / static_cast<double>(duration_from_previous_.size() - 1);
 }
 
 void RobotTrajectory::swap(RobotTrajectory& other)
--- a/moveit_core/robot_trajectory/src/robot_trajectory.cpp
+++ b/moveit_core/robot_trajectory/src/robot_trajectory.cpp
@@ -91,6 +91,12 @@ double RobotTrajectory::getDuration() const
 
 double RobotTrajectory::getAverageSegmentDuration() const
 {
+  if (duration_from_previous_.empty())
+  {
+    RCLCPP_WARN(rclcpp::get_logger("RobotTrajectory"), "Too few waypoints to calculate a duration. Returning 0.");
+    return 0.0;
+  }
+
   // If the initial segment has a duration of 0, exclude it from the average calculation
   if (duration_from_previous_[0] == 0)
   {
--- a/moveit_core/robot_trajectory/src/robot_trajectory.cpp
+++ b/moveit_core/robot_trajectory/src/robot_trajectory.cpp
@@ -91,13 +91,19 @@ double RobotTrajectory::getDuration() const
 
 double RobotTrajectory::getAverageSegmentDuration() const
 {
-  if (duration_from_previous_.size() <= 1)
+  // If the initial segment has a duration of 0, exclude it from the average calculation
+  if (duration_from_previous_[0] == 0)
   {
-    RCLCPP_WARN(rclcpp::get_logger("RobotTrajectory"), "Too few waypoints to calculate a duration. Returning 0.");
-    return 0.0;
+    if (duration_from_previous_.size() <= 1)
+    {
+      RCLCPP_WARN(rclcpp::get_logger("RobotTrajectory"), "Too few waypoints to calculate a duration. Returning 0.");
+      return 0.0;
+    }
+    else
+      return getDuration() / static_cast<double>(duration_from_previous_.size() - 1);
   }
   else
-    return getDuration() / static_cast<double>(duration_from_previous_.size() - 1);
+    return getDuration() / static_cast<double>(duration_from_previous_.size());
 }
 
 void RobotTrajectory::swap(RobotTrajectory& other)
--- a/moveit_core/robot_trajectory/src/robot_trajectory.cpp
+++ b/moveit_core/robot_trajectory/src/robot_trajectory.cpp
@@ -93,6 +93,7 @@ double RobotTrajectory::getAverageSegmentDuration() const
 {
   if (duration_from_previous_.size() <= 1)
   {
+    RCLCPP_WARN(rclcpp::get_logger("RobotTrajectory"), "Too few waypoints to calculate a duration. Returning 0.");
     return 0.0;
   }
   else
--- a/moveit_core/robot_trajectory/src/robot_trajectory.cpp
+++ b/moveit_core/robot_trajectory/src/robot_trajectory.cpp
@@ -91,8 +91,10 @@ double RobotTrajectory::getDuration() const
 
 double RobotTrajectory::getAverageSegmentDuration() const
 {
-  if (duration_from_previous_.empty())
+  if (duration_from_previous_.size() <= 1)
+  {
     return 0.0;
+  }
   else
     return getDuration() / static_cast<double>(duration_from_previous_.size() - 1);
 }


moveit_core/robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -94,7 +94,7 @@ double RobotTrajectory::getAverageSegmentDuration() const
   if (duration_from_previous_.empty())
     return 0.0;
   else
-    return getDuration() / static_cast<double>(duration_from_previous_.size());
+    return getDuration() / static_cast<double>(duration_from_previous_.size() - 1);
```
AndyZe, line 97: Good catch!

What if there is only one waypoint? Will add a suggestion to catch that.

moveit_core/robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -94,7 +94,7 @@ double RobotTrajectory::getAverageSegmentDuration() const
   if (duration_from_previous_.empty())
     return 0.0;
```
AndyZe, line 95:
```suggestion
  if (duration_from_previous_.size() <= 1)
  {
    return 0.0;
  }
```

AndyZe: It's too bad we can't print a warning here, but it looks like there is no Node available. I don't want to break the API just for a warning.

AndyZe (approved):

moveit_core/robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -91,10 +91,12 @@ double RobotTrajectory::getDuration() const
 
 double RobotTrajectory::getAverageSegmentDuration() const
 {
-  if (duration_from_previous_.empty())
+  if (duration_from_previous_.size() <= 1)
+  {
```
AndyZe, line 95: Sorry, but I just found this option for logging with no Node:  https://answers.ros.org/question/355985/global-logger-for-logging-without-a-node/?answer=355986#post-id-355986

```suggestion
  {
    RCLCPP_WARN(rclcpp::get_logger("my_logger"), "Too few waypoints to calculate a duration. Returning 0.");
```

v4hn (commented):
I'm sorry, I do not understand your explanation for the claimed error @stephanie-eng and stared at it longer than I would admit. :)

> If we have n waypoints, we have n - 1 timesteps

Unless I'm horribly misreading something here this is wrong. Each waypoint of course has a *timestamp*. But even if you mean duration_from_previous when you write "timesteps", this is not what `RobotTrajectory` does. The class keeps one duration_from_previous value per waypoint - unless these values are not set at all.

> Assume a trivial case of durations = [1, 2, 3] with a timestamp of 1 and a size of 3

I assume you mean 3 waypoints with *time_from_start* 1, 2, and 3, respectively.
Thus the `duration_from_previous_` list is [1,1,1] for this case.

> getDuration returns:
> 3 - 1 = 2

This is not what `getDuration()` computes, right?  It just accumulates the `duration_from_previous_` list, so you will get 3.

> The expected return should be:
> 2 / (3 - 1) = 1

I would expect 3 / 3 = 1, which should be the case with the current code.

This is a matter of definition, but as @AndyZe agrees with you, we might want to state this explicitly somewhere or redefine it (consistently across the class): A `RobotTrajectory` always represents the trajectory from the time point hinted by the duration_from_previous value of the first waypoint and thus might start *before* the first waypoint.
This is most visible in that a single waypoint with a duration_from_previous value set to non-zero is a well-defined trajectory.
Sending such a trajectory to a controller will generate the path of that trajectory only at runtime, when the real state of the system is known upon receiving the trajectory.
That definition is somewhat ugly for MoveIt, but it's how ros_control acts upon receiving such a trajectory and why would you even have a duration_from_previous value that is non-zero in the first waypoint if you don't want that behavior?

AndyZe:
@v4hn I think you're right, we didn't consider that a nonzero `duration_from_previous` might actually make sense at the first waypoint.

However, it was also confusing from our perspective that we had 21 waypoints, all of them having a `duration_from_previous` of 0.02 seconds (except the first waypoint had a `duration_from_previous` of 0). It was intuitive in our case that the average segment duration should have been 0.02.

Well, I guess it will be confusing either way so perhaps we should not merge it and leave things as they are.

v4hn:
> it was also confusing from our perspective that we had 21 waypoints, all of them having a `duration_from_previous` of 0.02 seconds (except the first waypoint had a `duration_from_previous` of 0). It was intuitive in our case that the average segment duration should have been 0.02.

I see where you are coming from. You might actually claim that an initial segment with a duration_from_previous of 0.0 does not actually exist and thus should not affect the average.  Would it be appropriate for you to add such a special case here (with an inline comment explaining the reasoning)?

moveit_core/robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -91,8 +91,17 @@ double RobotTrajectory::getDuration() const
 
 double RobotTrajectory::getAverageSegmentDuration() const
 {
-  if (duration_from_previous_.empty())
-    return 0.0;
+  // If the initial segment has a timstep of 0, exclude it grom teh average calculation
```
AndyZe, line 94:
```suggestion
  // If the initial segment has a duration of 0, exclude it from the average calculation
```

moveit_core/robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -91,8 +91,17 @@ double RobotTrajectory::getDuration() const
...
+  if (duration_from_previous_[0] == 0)
+  {
+    if (duration_from_previous_.size() <= 1)
+    {
+      RCLCPP_WARN(rclcpp::get_logger("RobotTrajectory"), "Too few waypoints to calculate a duration. Returning 0.");
+      return 0.0;
+    }
```
AndyZe, line 101: I think this error checking should be moved outside of `if (duration_from_previous_[0] == 0)` since it applies to either case

stephanie-eng: I was thinking about this: if you have one waypoint with a non-zero `duration_from_previous`, wouldn't moving this check return zero rather than the value?

AndyZe: OK, good point. Well, I think the `else` case here needs a bit of error checking too like it originally had. Will make a suggestion.

moveit_core/robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -91,8 +91,17 @@ double RobotTrajectory::getDuration() const
 
 double RobotTrajectory::getAverageSegmentDuration() const
 {
-  if (duration_from_previous_.empty())
-    return 0.0;
```
AndyZe, line 95: How about restoring these two lines as an initial error check?

```
if (duration_from_previous_.empty())
{
  RCLCPP_WARN(rclcpp::get_logger("RobotTrajectory"), "Too few waypoints to calculate a duration. Returning 0.");
  return 0.0;
}
```

AndyZe (approved):

moveit_core/robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -92,7 +92,22 @@ double RobotTrajectory::getDuration() const
 double RobotTrajectory::getAverageSegmentDuration() const
 {
   if (duration_from_previous_.empty())
```
v4hn, line 94:
```suggestion
  if (duration_from_previous_.empty() || (duration_from_previous_.size() == 1 && !has_initial_segment))
```

moveit_core/robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -92,7 +92,22 @@ double RobotTrajectory::getDuration() const
...
+    if (duration_from_previous_.size() <= 1)
+    {
+      RCLCPP_WARN(rclcpp::get_logger("RobotTrajectory"), "Too few waypoints to calculate a duration. Returning 0.");
+      return 0.0;
+    }
+    else
+      return getDuration() / static_cast<double>(duration_from_previous_.size() - 1);
+  }
   else
     return getDuration() / static_cast<double>(duration_from_previous_.size());
```
v4hn, line 112:
```suggestion
  if (has_initial_segment)
    return getDuration() / static_cast<double>(duration_from_previous_.size());
  else
    return getDuration() / static_cast<double>(duration_from_previous_.size() - 1);
```

moveit_core/robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -92,7 +92,22 @@ double RobotTrajectory::getDuration() const
 double RobotTrajectory::getAverageSegmentDuration() const
 {
   if (duration_from_previous_.empty())
+  {
+    RCLCPP_WARN(rclcpp::get_logger("RobotTrajectory"), "Too few waypoints to calculate a duration. Returning 0.");
```
v4hn, line 96:
```suggestion
    RCLCPP_WARN(rclcpp::get_logger("RobotTrajectory"), "No motion segments in trajectory. Returning 0 average segment duration.");
```

v4hn (commented):
It's bad practice to trigger the same log message from different lines of code.
Either change the messages to be sufficiently specific or you can go with something like my suggestion to simplify the code a bit, but loose the distinction in the logs.

PR #1087, opened 2022-02-21
[hybrid planning] Adjust planning scene locking

AndyZe:
We have a dual-arm project and found there was a deadlock here where the planning scene locks. This change fixes the deadlock and I think it's nice in a couple other ways, too:

- Make the robot state const.
- Eliminate an intermediate variable.

I'm all-ears in case there is a better way that I could test.

### DIFF
--- a/moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
+++ b/moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
@@ -235,14 +235,17 @@ void LocalPlannerComponent::executeIteration()
     // If the planner received an action request and a global solution it starts to plan locally
     case LocalPlannerState::LOCAL_PLANNING_ACTIVE:
     {
-      // Read current robot state
       planning_scene_monitor_->updateSceneWithCurrentState();
-      planning_scene_monitor_->lockSceneRead();  // LOCK planning scene
-      const auto current_robot_state = planning_scene_monitor_->getPlanningScene()->getCurrentState();
-      planning_scene_monitor_->unlockSceneRead();  // UNLOCK planning scene
+
+      // Read current robot state
+      moveit::core::RobotStatePtr current_robot_state;
+      {
+        planning_scene_monitor::LockedPlanningSceneRO ls(planning_scene_monitor_);
+        current_robot_state = std::make_shared<moveit::core::RobotState>(ls->getCurrentState());
+      }
 
       // Check if the global goal is reached
-      if (trajectory_operator_instance_->getTrajectoryProgress(current_robot_state) > PROGRESS_THRESHOLD)
+      if (trajectory_operator_instance_->getTrajectoryProgress(*current_robot_state) > PROGRESS_THRESHOLD)
       {
         local_planning_goal_handle_->succeed(result);
         reset();
@@ -253,7 +256,7 @@ void LocalPlannerComponent::executeIteration()
       robot_trajectory::RobotTrajectory local_trajectory =
           robot_trajectory::RobotTrajectory(planning_scene_monitor_->getRobotModel(), config_.group_name);
       *local_planner_feedback_ =
-          trajectory_operator_instance_->getLocalTrajectory(current_robot_state, local_trajectory);
+          trajectory_operator_instance_->getLocalTrajectory(*current_robot_state, local_trajectory);
 
       // Feedback is only sent when the hybrid planning architecture should react to a discrete event that occurred
       // during the identification of the local planning problem
--- a/moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
+++ b/moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
@@ -96,6 +96,12 @@ bool LocalPlannerComponent::initialize()
   planning_scene_monitor_->startWorldGeometryMonitor(config_.collision_object_topic);
   planning_scene_monitor_->startStateMonitor(config_.joint_states_topic);
 
+  // Create a copy of the planning scene
+  planning_scene_monitor_->lockSceneRead();  // LOCK planning scene
+  planning_scene::PlanningScenePtr planning_scene = planning_scene_monitor_->getPlanningScene();
+  planning_scene->decoupleParent();
+  planning_scene_monitor_->unlockSceneRead();  // UNLOCK planning scene
+
   // Load trajectory operator plugin
   try
   {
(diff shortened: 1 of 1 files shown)


moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
```diff
@@ -96,6 +96,12 @@ bool LocalPlannerComponent::initialize()
   planning_scene_monitor_->startWorldGeometryMonitor(config_.collision_object_topic);
   planning_scene_monitor_->startStateMonitor(config_.joint_states_topic);
 
+  // Create a copy of the planning scene
+  planning_scene_monitor_->lockSceneRead();  // LOCK planning scene
```
v4hn, line 100: use `LockedPlanningSceneRO` over the `lock/unlock` methods. What happens if you hit an exception in between? :/

moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
```diff
@@ -96,6 +96,12 @@ bool LocalPlannerComponent::initialize()
...
+  planning_scene::PlanningScenePtr planning_scene = planning_scene_monitor_->getPlanningScene();
+  planning_scene->decoupleParent();
```
v4hn, line 102: The planning_scene_monitor already maintains a decoupled scene I guess.
You are trying to decouple the planning scene that is modified inside the monitor here.
Why? I expect that these lines are a no-op.

moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
```diff
@@ -235,15 +241,12 @@ void LocalPlannerComponent::executeIteration()
...
+      // Read current robot state
       planning_scene_monitor_->updateSceneWithCurrentState();
       planning_scene_monitor_->lockSceneRead();  // LOCK planning scene
-      planning_scene::PlanningScenePtr planning_scene = planning_scene_monitor_->getPlanningScene();
+      const auto current_robot_state = planning_scene_monitor_->getPlanningScene()->getCurrentState();
       planning_scene_monitor_->unlockSceneRead();  // UNLOCK planning scene
 
-      // Get current state
-      auto current_robot_state = planning_scene->getCurrentStateNonConst();
-
```
v4hn, line 246: This was a clear bug in the previous code and the fix looks good :+1: 
Still `LockedPlanningSceneRO` should be strongly preferred over the `lock`/`unlock` methods...

AndyZe: Thanks. Maybe `lockSceneRead() / lockSceneWrite()` should be private, then?  (in a follow-up PR, of course)

moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
```diff
@@ -235,17 +235,17 @@ void LocalPlannerComponent::executeIteration()
...
-      planning_scene_monitor_->unlockSceneRead();  // UNLOCK planning scene
 
-      // Get current state
-      auto current_robot_state = planning_scene->getCurrentStateNonConst();
+      // Read current robot state
+      moveit::core::RobotStatePtr current_robot_state;
+      {
+        planning_scene_monitor::LockedPlanningSceneRO ls(planning_scene_monitor_);
+        current_robot_state = std::make_shared<moveit::core::RobotState>(ls->getCurrentState());
+      }
```
tylerjw, line 245: Is there a reason this needs to be a new shared printer and you couldn't just do this since you are making a copy:
```suggestion
      moveit::core::RobotState const current_robot_state = [&]()
      {
        planning_scene_monitor::LockedPlanningSceneRO ls(planning_scene_monitor_);
        return ls->getCurrentState();
      }
```

This is a bit of optimization but I do think it is good not to use less fancy language features (like shared_prt) when you don't need them.  Then later in the code, you don't need to add the dereference operator `*` to the calls to getTrajectoryProgress and getLocalTrajectory.

v4hn: You say "not to use less fancy features" and propose to use IIFE in the same comment. :rocket: :eyes:  :bouquet: 
I agree, though (1) `const` in second place in the type is very unusual for our coding style (I've never seen it anywhere), and (2) `[&](){` simplifies to `[&]{`.

It would be even better though to provide the method with that interface directly with the `PlanningSceneMonitor` class to begin with!

moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
v4hn: YES. Though you will have to add friends to the locker to get there.
Locking the PSM can most likely be refactored quite a bit more with modern coding practice. the whole `LockedPlanningSceneR*` structure seems rather messy..

moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
tylerjw: I updated my suggestion.  IFEE is an advanced technique but it uses a much simpler and less complex language feature (lambdas) than shared_ptr which is basically adding a global and an extra call to malloc where it isn't needed.  I'm just here to continue my crusade against shared_ptr and for making more things const.  :shrug: :fire:

tylerjw: On east const.  I also updated my suggestion to remove it.  I know we don't use that style in MoveIt.  It isn't a new thing even if you haven't seen it: https://www.youtube.com/watch?v=fv--IKZFVO8

AndyZe: I'm OK with the lambda function. I don't think it's tooo complicated. Had to tweak your suggestion to capture _this_ instead of `planning_scene_monitor_`, though.

`error: capture of non-variable ‘moveit::hybrid_planning::LocalPlannerComponent::planning_scene_monitor_’`

tylerjw (approved):

v4hn (approved):

moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
v4hn: > I don't really care that much except that @v4hn's statement about not seeing east const before

I've seen it before, just not in MoveIt or even the ROS community. I'm absolutely ok with a clang-tidy option to use it everywhere if there are strong opinions among multiple maintainers. I don't think clang-tidy merged support for it yet though...
I also think we have more urgent issues and you are exaggerating syntactic details.
Let's continue this some time in a pub. :beers:

PR #1090, opened 2022-02-23
[hybrid planning] improve planning scene monitoring

AndyZe:
I don't have time to get this cleaned up to MoveIt standards but I'm pushing it as a record for anybody else who wants to use hybrid planning (especially for a dual-arm system). Prior to this, we found that the planning scene only updated occasionally and it was reporting phantom collisions.

What does this PR do?

- Use a different collision detection type (to match what's done in moveit_servo)
- Turn on planning scene diff monitoring
- Spawn a new PSM in the local constraint solver, with clearly-specified settings.

### DIFF
--- a/moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/forward_trajectory.cpp
+++ b/moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/forward_trajectory.cpp
@@ -43,7 +43,6 @@ const rclcpp::Logger LOGGER = rclcpp::get_logger("local_planner_component");
 // If stuck for this many iterations or more, abort the local planning action
 constexpr size_t STUCK_ITERATIONS_THRESHOLD = 5;
 constexpr double STUCK_THRESHOLD_RAD = 1e-4;  // L1-norm sum across all joints
-constexpr double COLLISION_THRESHOLD = 0.01;  // Stop if closer than this [meters]
 }  // namespace
 
 namespace moveit::hybrid_planning
@@ -66,9 +65,6 @@ bool ForwardTrajectory::initialize(const rclcpp::Node::SharedPtr& node,
   num_iterations_stuck_ = 0;
 
   planning_scene_monitor_ = planning_scene_monitor;
-
-  collision_request_.distance = true;   // enable distance-based collision checking
-  collision_request_.contacts = false;  // Record the names of collision pairs
 
   return true;
 }
@@ -115,9 +111,7 @@ ForwardTrajectory::solve(const robot_trajectory::RobotTrajectory& local_trajecto
       planning_scene_monitor_->updateSceneWithCurrentState();
       planning_scene_monitor::LockedPlanningSceneRO locked_planning_scene(planning_scene_monitor_);
       current_state = std::make_shared<moveit::core::RobotState>(locked_planning_scene->getCurrentState());
-      collision_result_.clear();
-      locked_planning_scene->checkCollision(collision_request_, collision_result_);
-      is_path_valid = (!(collision_result_.collision || (collision_result_.distance < COLLISION_THRESHOLD)));
+      is_path_valid = locked_planning_scene->isPathValid(local_trajectory, local_trajectory.getGroupName(), false);
     }
 
     // Check if path is valid
--- a/moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/forward_trajectory.cpp
+++ b/moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/forward_trajectory.cpp
@@ -112,13 +112,11 @@ ForwardTrajectory::solve(const robot_trajectory::RobotTrajectory& local_trajecto
     bool is_path_valid = false;
     // Lock the planning scene as briefly as possible
     {
+      planning_scene_monitor_->updateSceneWithCurrentState();
       planning_scene_monitor::LockedPlanningSceneRO locked_planning_scene(planning_scene_monitor_);
       current_state = std::make_shared<moveit::core::RobotState>(locked_planning_scene->getCurrentState());
-      current_state->updateCollisionBodyTransforms();
       collision_result_.clear();
-      collision_request_.group_name = local_trajectory.getGroupName();
-      locked_planning_scene->getCollisionEnv()->checkRobotCollision(collision_request_, collision_result_,
-                                                                    *current_state);
+      locked_planning_scene->checkCollision(collision_request_, collision_result_);
       is_path_valid = (!(collision_result_.collision || (collision_result_.distance < COLLISION_THRESHOLD)));
     }
 
--- a/moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/forward_trajectory.cpp
+++ b/moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/forward_trajectory.cpp
@@ -49,7 +49,7 @@ constexpr double COLLISION_THRESHOLD = 0.01;  // Stop if closer than this [meter
 namespace moveit::hybrid_planning
 {
 bool ForwardTrajectory::initialize(const rclcpp::Node::SharedPtr& node,
-                                   const planning_scene_monitor::PlanningSceneMonitorConstPtr& planning_scene_monitor,
+                                   const planning_scene_monitor::PlanningSceneMonitorPtr& planning_scene_monitor,
                                    const std::string& /* unused */)
 {
   // Load parameter & initialize member variables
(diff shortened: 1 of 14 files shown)


tylerjw:
@henningkayser do you have thoughts on these changes?

moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/forward_trajectory.cpp
```diff
@@ -60,10 +61,15 @@ bool ForwardTrajectory::initialize(const rclcpp::Node::SharedPtr& node,
   {
     stop_before_collision_ = node->declare_parameter<bool>("stop_before_collision", false);
   }
-  planning_scene_monitor_ = planning_scene_monitor;
   node_ = node;
   path_invalidation_event_send_ = false;
   num_iterations_stuck_ = 0;
+
+  planning_scene_monitor_ = planning_scene_monitor;
```
sjahr, line 68: @AndyZe I re-enabled passing the a configured planning scene monitor down from the local_planning component

moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
```diff
@@ -93,9 +93,13 @@ bool LocalPlannerComponent::initialize()
   }
 
   // Start state and scene monitors
+  // Set frequency faster than local planner frequency (If we don't call this it would default to 100Hz)
+  planning_scene_monitor_->setStateUpdateFrequency(2.0 * config_.local_planning_frequency);
```
sjahr, line 97: Based on the documentation:
```cpp
  /** @brief Update the scene using the monitored state. This function is automatically called when an update to the
     current state is received (if startStateMonitor() has been called).
      The updates are throttled to a maximum update frequency however, which is set by setStateUpdateFrequency(). */
  void updateSceneWithCurrentState();
```
I decided instead of calling `updateSceneWithCurrentState()` every time the local planning component gets invoked, I increased the state update frequency.

AndyZe: Why a factor of 2? How about a factor of 1.1 so the updates are just barely faster than required.

AndyZe:
I tested with the tutorial. It's always been flaky afaik and still is, but at least this PR doesn't cause a regression. I'll approve after that factor of 2 is changed.

sjahr:
@AndyZe I decreased the factor to 1.1

sjahr:
> I tested with the tutorial. It's always been flaky afaik and still is, but at least this PR doesn't cause a regression. I'll approve after that factor of 2 is changed.

Yes, I think we need to revisit the hybrid planning tutorial. It is flaky for me too :/

moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
henningkayser: the documentation is misleading. The automated updates are throttled, but calling `updateSceneWithCurrentState()` will always immediately apply the latest incoming state of the CurrentStateMonitor to the scene. I think calling `updateSceneWithCurrentState()` is therefore better (more stable and more up-to-date) than running the throttled updates at an async rate. Having a factor of 1.1 (or even higher) may cause an artificial undersampling of the actual sensor input.

moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/forward_trajectory.cpp
```diff
@@ -60,10 +61,15 @@ bool ForwardTrajectory::initialize(const rclcpp::Node::SharedPtr& node,
...
+
+  collision_request_.distance = true;   // enable distance-based collision checking
```
henningkayser, line 70: good move to enable distance checking by default here :+1:

moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/forward_trajectory.cpp
```diff
@@ -108,7 +114,12 @@ ForwardTrajectory::solve(const robot_trajectory::RobotTrajectory& local_trajecto
     {
       planning_scene_monitor::LockedPlanningSceneRO locked_planning_scene(planning_scene_monitor_);
       current_state = std::make_shared<moveit::core::RobotState>(locked_planning_scene->getCurrentState());
-      is_path_valid = locked_planning_scene->isPathValid(local_trajectory, local_trajectory.getGroupName(), false);
+      current_state->updateCollisionBodyTransforms();
+      collision_result_.clear();
+      collision_request_.group_name = local_trajectory.getGroupName();
+      locked_planning_scene->getCollisionEnv()->checkRobotCollision(collision_request_, collision_result_,
```
henningkayser, line 120: So instead of checking the local trajectory, we are only checking the current state now?

moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/forward_trajectory.cpp
```diff
@@ -108,7 +114,12 @@ ForwardTrajectory::solve(const robot_trajectory::RobotTrajectory& local_trajecto
     {
       planning_scene_monitor::LockedPlanningSceneRO locked_planning_scene(planning_scene_monitor_);
```
henningkayser, line 115: we should probably update the current state before checking using `updateSceneWithCurrentState()` as well, that also takes care about updating transforms. After that we can directly call [PlanningScene::checkCollision()](https://moveit.picknik.ai/humble/api/html/classplanning__scene_1_1PlanningScene.html#a4ebcc755346e443319104e8b964f8faa) So that we don't have to copy the current state. It would be even better if we could extend the [distanceToCollision()](https://moveit.picknik.ai/humble/api/html/classplanning__scene_1_1PlanningScene.html#a838cbad712afd5bc66374f4bb46402fa) API to support groups and current state

moveit_ros/hybrid_planning/local_planner/local_planner_component/src/local_planner_component.cpp
```diff
@@ -93,9 +93,13 @@ bool LocalPlannerComponent::initialize()
   }
 
   // Start state and scene monitors
+  // Set frequency faster than local planner frequency (If we don't call this it would default to 100Hz)
+  planning_scene_monitor_->setStateUpdateFrequency(1.1 * config_.local_planning_frequency);
   planning_scene_monitor_->startSceneMonitor(config_.monitored_planning_scene_topic);
   planning_scene_monitor_->startWorldGeometryMonitor(config_.collision_object_topic);
-  planning_scene_monitor_->startStateMonitor(config_.joint_states_topic);
+  planning_scene_monitor_->startStateMonitor(config_.joint_states_topic, "/attached_collision_object");
```
PeterMitrano, line 100: Is having this hard-coded a good idea?

PeterMitrano (commented):
If this is changing behavior in a meaningful way, perhaps there are docs or tutorials that should get updated? I'm just throwing stuff out there since y'all asked for a reviewer -- I'm not familiar with hybrid planning.

moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/forward_trajectory.cpp
```diff
@@ -43,6 +43,7 @@ const rclcpp::Logger LOGGER = rclcpp::get_logger("local_planner_component");
 // If stuck for this many iterations or more, abort the local planning action
 constexpr size_t STUCK_ITERATIONS_THRESHOLD = 5;
 constexpr double STUCK_THRESHOLD_RAD = 1e-4;  // L1-norm sum across all joints
+constexpr double COLLISION_THRESHOLD = 0.01;  // Stop if closer than this [meters]
```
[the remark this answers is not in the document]
AndyZe: it's not :+1: but due to not having a client to actively fund this, it will probably have to stay that way for now  :(

Feel free to open a PR, of course.

moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/forward_trajectory.cpp
AndyZe: To be honest, this PR is over a year old. It seems there was a bug with `isPathValid()` but I don't remember clearly any longer. I guess I'll revert this change since it would be nice to check the entire path.

AndyZe: Ah, but the downside of `isPathValid()` is that there's no distance checking.

moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/forward_trajectory.cpp
AndyZe: done

AndyZe:
I can't approve my own PR but this looks OK to me now

moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/forward_trajectory.cpp
sjahr: Do you mean we don't get the distance to collision objects? Does this matter since we only care about a binary decision of whether the local trajectory is collision-free or not?

AndyZe: I reverted this change since we don't currently have a good way to test it.

AndyZe:
@sjahr @henningkayser what about a final review?

henningkayser (approved):

sjahr (approved):

PR #1095, opened 2022-03-01
Compilation fixes for Jammy and bring back Rolling CI

vatanaksoytezer:
Also updating clang-format to 12 since 10 is not available in Jammy, and is pretty old.

### DIFF
--- a/moveit2.repos
+++ b/moveit2.repos
@@ -2,7 +2,7 @@ repositories:
   geometric_shapes:
     type: git
     url: https://github.com/ros-planning/geometric_shapes
-    version: 2.1.2
+    version: ros2
   moveit_msgs:
     type: git
     url: https://github.com/ros-planning/moveit_msgs
@@ -19,3 +19,7 @@ repositories:
     type: git
     url: https://github.com/ros-controls/ros2_control.git
     version: 2.1.0
+  ompl:
+    type: git
+    url: https://github.com/ompl/ompl.git
+    version: main
--- a/moveit_ros/moveit_servo/test/servo_launch_test_common.hpp
+++ b/moveit_ros/moveit_servo/test/servo_launch_test_common.hpp
@@ -113,7 +113,7 @@ public:
     // Otherwise the Servo is still running when another test starts...
     if (!client_servo_stop_)
     {
-      client_servo_stop_->async_send_request(std::make_shared<std_srvs::srv::Trigger::Request>());
+      stop();
     }
     executor_->cancel();
     if (executor_thread_.joinable())
--- a/moveit_ros/visualization/package.xml
+++ b/moveit_ros/visualization/package.xml
@@ -25,7 +25,6 @@
 
   <build_depend>class_loader</build_depend>
   <build_depend>eigen</build_depend>
-  <build_depend>libogre-dev</build_depend>
   <build_depend>libqt5-opengl-dev</build_depend>
   <build_depend>qtbase5-dev</build_depend>
 
--- a/moveit_ros/warehouse/warehouse/src/broadcast.cpp
+++ b/moveit_ros/warehouse/warehouse/src/broadcast.cpp
@@ -68,6 +68,7 @@ int main(int argc, char** argv)
   // time to wait in between publishing messages
   double delay = 0.001;
 
+  // clang-format off
   boost::program_options::options_description desc;
   desc.add_options()("help", "Show help message")("host", boost::program_options::value<std::string>(),
                                                   "Host for the "
@@ -80,7 +81,7 @@ int main(int argc, char** argv)
       "state", boost::program_options::value<std::string>(),
       "Name of the robot state to publish.")("delay", boost::program_options::value<double>()->default_value(delay),
                                              "Time to wait in between publishing messages (s)");
-
+  // clang-format on
   boost::program_options::variables_map vm;
   boost::program_options::store(boost::program_options::parse_command_line(argc, argv, desc), vm);
   boost::program_options::notify(vm);
--- a/moveit_ros/warehouse/warehouse/src/import_from_text.cpp
+++ b/moveit_ros/warehouse/warehouse/src/import_from_text.cpp
@@ -216,12 +216,14 @@ int main(int argc, char** argv)
   node_options.automatically_declare_parameters_from_overrides(true);
   rclcpp::Node::SharedPtr node = rclcpp::Node::make_shared("import_from_text_to_warehouse", node_options);
 
+  // clang-format off
   boost::program_options::options_description desc;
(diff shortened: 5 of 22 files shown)


moveit_ros/warehouse/warehouse/src/broadcast.cpp
```diff
@@ -69,17 +69,41 @@ int main(int argc, char** argv)
...
+                                                                                                                  "publ"
+                                                                                                                  "ish"
+                                                                                                                  ".")("delay",
+                                                                                                                       boost::program_options::value<
+                                                                                                                           double>()
+                                                                                                                           ->default_value(
+                                                                                                                               delay),
+                                                                                                                       "Time to wait in between publishing messages (s)");
```
vatanaksoytezer, line 106: This clang-format fix is pretty weird. I am open to suggestions

moveit_ros/warehouse/warehouse/src/import_from_text.cpp
```diff
@@ -217,11 +217,18 @@ int main(int argc, char** argv)
...
+      "Name of file containing motion planning queries.")("scene", boost::program_options::value<std::string>(),
+                                                          "Name of file containing motion "
+                                                          "planning scene.")("host",
+                                                                             boost::program_options::value<std::string>(),
+                                                                             "Host for the "
+                                                                             "DB.")("port",
+                                                                                    boost::program_options::value<
+                                                                                        std::size_t>(),
+                                                                                    "Port for "
+                                                                                    "the DB.");
```
vatanaksoytezer, line 231: Also this.

moveit_setup_assistant/src/collisions_updater.cpp
```diff
@@ -128,19 +128,56 @@ int main(int argc, char* argv[])
...
+                                                                                                                                                                                                             "number of trials for searching never colliding pairs")("min-collision-fraction",
+                                                                                                                                                                                                                                                                     po::value(
+                                                                                                                                                                                                                                                                         &min_collision_fraction),
+                                                                                                                                                                                                                                                                     "fraction of small sample size to determine links that are always colliding");
```
vatanaksoytezer, line 180: And this

vatanaksoytezer:
https://github.com/pal-robotics/backward_ros/pull/11 would help with `backward_ros` warnings.

vatanaksoytezer:
There are bunch of warnings, I would prefer to deal with them later if possible.

moveit_ros/warehouse/warehouse/src/broadcast.cpp
```diff
@@ -68,6 +68,7 @@ int main(int argc, char** argv)
   // time to wait in between publishing messages
   double delay = 0.001;
 
+  // clang-format off
```
vatanaksoytezer, line 71: I'm not a huge fan of putting this here, but there was insane indentation. clang-format-12 really acts weird on 20.04 and clang-format-10 is not supported in Ubuntu 22.04. In fact I am guessing this weirdness is because our format jobs are running in 20.04 images and I'm formatting in 22.04 - which seemed reasonable. Unfortunately GH ubuntu-latest is still 20.04 and there is no option for 22.04, I believe until May. I might migrate our pre-commit to run on an 22.04 image to avoid weird pre-commit problems.

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -1287,7 +1287,7 @@ template <typename T>
 bool parse(const YAML::Node& node, const std::string& key, T& storage, const T& default_value = T())
 {
   const YAML::Node& n = node[key];
-  bool valid = n;
+  bool valid = n.IsDefined();
```
vatanaksoytezer, line 1290: Fixing a small operator here

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -1207,7 +1207,7 @@ class SortableDisabledCollision
     : dc_(dc), key_(dc.link1_ < dc.link2_ ? (dc.link1_ + "|" + dc.link2_) : (dc.link2_ + "|" + dc.link1_))
   {
   }
-  operator const srdf::Model::DisabledCollision &() const
+  operator const srdf::Model::DisabledCollision&() const
```
vatanaksoytezer, line 1210: pre-commit did this.

moveit_ros/moveit_servo/test/servo_launch_test_common.hpp
```diff
@@ -113,7 +113,7 @@ class ServoFixture : public ::testing::Test
     // Otherwise the Servo is still running when another test starts...
     if (!client_servo_stop_)
     {
-      client_servo_stop_->async_send_request(std::make_shared<std_srvs::srv::Trigger::Request>());
+      stop();
```
vatanaksoytezer, line 116: I'm not exactly sure why the previous command was causing null pointers, but this seems to compile and passes tests.

moveit2.repos
```diff
@@ -19,3 +19,7 @@ repositories:
     type: git
     url: https://github.com/ros-controls/ros2_control.git
     version: 2.1.0
+  ompl:
```
vatanaksoytezer, line 22: ompl has been released but not synced, we should keep this until the next sync.

moveit2.repos
```diff
@@ -2,7 +2,7 @@ repositories:
   geometric_shapes:
     type: git
     url: https://github.com/ros-planning/geometric_shapes
-    version: 2.1.2
+    version: ros2
```
vatanaksoytezer, line 5: We should make a new release for geometric shapes to fix buildfarm errors and use the tag again.

vatanaksoytezer:
@henningkayser I would appreciate a review on this one. If looks good can you remove the required galactic CI job and make the rolling one required?

moveit_ros/visualization/package.xml
```diff
@@ -25,7 +25,6 @@
 
   <build_depend>class_loader</build_depend>
   <build_depend>eigen</build_depend>
-  <build_depend>libogre-dev</build_depend>
```
vatanaksoytezer, line 28: I think this isn't used

tylerjw: There are some headers that include it.

moveit_ros/moveit_servo/test/servo_launch_test_common.hpp
tylerjw: Was there a path where `client_servo_stop_` wasn't valid?

tylerjw (approved):

henningkayser (approved):

PR #1096, opened 2022-03-02
[moveit_cpp] Fix config of multiple pipelines

galou:

### Description

The issue I had was that MoveItCpp was working with one pipeline but when I wanted to configure a second pipeline, I couldn't use the first one.

### Checklist

Do I really need to write a test for such a change?

Edit: Added PR description

### DIFF
--- a/moveit_ros/planning/moveit_cpp/src/moveit_cpp.cpp
+++ b/moveit_ros/planning/moveit_cpp/src/moveit_cpp.cpp
@@ -165,7 +165,6 @@ bool MoveItCpp::loadPlanningPipelines(const PlanningPipelineOptions& options)
   {
     for (const auto& group_name : group_names)
     {
-      groups_pipelines_map_[group_name] = {};
       const auto& pipeline = pipeline_entry.second;
       for (const auto& planner_configuration : pipeline->getPlannerManager()->getPlannerConfigurations())
       {


v4hn (approved):
> Testing this change would require launching a node with the configuration of two pipelines. Then I don't know how I could check the internal state of the node (group_pipelines_map_).

Well, apparently you hit a bug or you would not have created this patch called "Fix". You can definitely check for whether the buggy behavior is fixed. :)
Looking at this further the whole code is a mess. The only place where this map is used in `getPlanningPipelineNames` and there it's even used as a lookup to *verify* whether the planner exists for this group instead of just taking the set of pipelines from there... It's clearly not faster that way, as the comment here claims...

I would much prefer a proper cleanup of these methods instead of this one-line patch, but the patch itself is correct.

galou:
@v4hn I can't agree more with you. The issue I had was that MoveItCpp was working with one pipeline but when I wanted to configure a second pipeline, I couldn't use the first one. I took me two days to figure out that the reason was not my configuration but the code itself.

v4hn:
> The issue I had was that MoveItCpp was working with one pipeline but when I wanted to configure a second pipeline, I couldn't use the first one. I took me two days to figure out that the reason was not my configuration but the code itself.

It might have been a good idea to have these sentences in the PR description already ;)

henningkayser (approved):

v4hn:
> henningkayser approved this pull request.

Please merge if you want. I rather approved above, but I still think someone who wrote these methods (you? ;)) should rewrite them instead.

PR #1098, opened 2022-03-04
[moveit_cpp] Add setLastPlanSolution

galou:

### Description

Without this function, it is not possible to execute another plan that the last computed one. I don't like the name (but I don't like the idea of having `execute()`  not taking a plan as input either), maybe `setNextPlanToExecute()` could be an alternative. So, it's more to start a discussion.

### Checklist

### DIFF
--- a/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
+++ b/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
@@ -177,9 +177,8 @@ private:
   planning_scene_monitor::PlanningSceneMonitorPtr planning_scene_monitor_;
 
   // Planning
-  std::map<std::string, planning_pipeline::PlanningPipelinePtr> planning_pipelines_;
-  std::map<std::string, std::set<std::string>> groups_pipelines_map_;
-  std::map<std::string, std::set<std::string>> groups_algorithms_map_;
+  std::map<std::string, planning_pipeline::PlanningPipelinePtr> planning_pipelines_;  // pipeline name: pipeline pointer.
+  std::map<std::string, std::set<std::string>> groups_pipelines_map_;                 // group name: set{pipeline name}.
 
   // Execution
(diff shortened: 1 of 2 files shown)


moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
```diff
@@ -177,9 +177,8 @@ class MoveItCpp
   planning_scene_monitor::PlanningSceneMonitorPtr planning_scene_monitor_;
 
   // Planning
-  std::map<std::string, planning_pipeline::PlanningPipelinePtr> planning_pipelines_;
-  std::map<std::string, std::set<std::string>> groups_pipelines_map_;
-  std::map<std::string, std::set<std::string>> groups_algorithms_map_;
```
henningkayser, line 182: Is this change related in any way?

henningkayser (commented):
@galou I'm not convinced that this addition is necessary. You can execute arbitrary trajectories using `MoveItCpp::execute()`. In fact, you can even access the `TrajectoryExecutionManager` and mess with it during execution. The `PlanningComponent` is meant to be a convenience wrapper for planning requests and not really an execution interface. I would rather remove the `PlanningComponent::execute()`, probably. 

Side note: `getLastPlanSolution()` returns a pointer to non-const (which is not really ideal). That would also allow you to change the last solution trajectory.

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
galou: Thanks, `MoveItCpp::execute()` is the answer I was looking for.

Wouldn't it be also better to remove `getLastPlanSolution()`? I don't see the use case.

henningkayser: I think being able to introspect or process the solution makes sense, but maybe it should really be non-const.

PR #1099, opened 2022-03-07
Remove include of OgrePrerequisites header

AndyZe:
It fixes this specific build error on our project's CI setup:

```
  --- stderr: moveit_task_constructor_visualization
  In file included from /opt/ros/galactic/include/moveit/rviz_plugin_render_tools/render_shapes.h:39,
                   from /opt/ros/galactic/include/moveit/rviz_plugin_render_tools/planning_scene_render.h:41,
                   from /home/runner/work/lab0_harvester/lab0_harvester/.work/upstream_ws/src/moveit_task_constructor/visualization/visualization_tools/src/task_solution_visualization.cpp:42:
  /opt/ros/galactic/include/moveit/rviz_plugin_render_tools/octomap_render.h:45:10: fatal error: OGRE/OgrePrerequisites.h: No such file or directory
     45 | #include <OGRE/OgrePrerequisites.h>
```

@jliukkonen and @tylerjw know the context and more details on the CI situation. I don't know so much about it.

### DIFF
--- a/moveit_ros/visualization/rviz_plugin_render_tools/include/moveit/rviz_plugin_render_tools/octomap_render.h
+++ b/moveit_ros/visualization/rviz_plugin_render_tools/include/moveit/rviz_plugin_render_tools/octomap_render.h
@@ -42,7 +42,6 @@
 #include <rviz_default_plugins/displays/pointcloud/point_cloud_helpers.hpp>
 #include <rviz_common/properties/color_property.hpp>
 #include <moveit/rviz_plugin_render_tools/octomap_render.h>
-#include <OGRE/OgrePrerequisites.h>
 
 namespace octomap
 {
--- a/moveit_ros/visualization/rviz_plugin_render_tools/include/moveit/rviz_plugin_render_tools/octomap_render.h
+++ b/moveit_ros/visualization/rviz_plugin_render_tools/include/moveit/rviz_plugin_render_tools/octomap_render.h
@@ -41,7 +41,6 @@
 #include <octomap/octomap.h>
 #include <rviz_default_plugins/displays/pointcloud/point_cloud_helpers.hpp>
 #include <rviz_common/properties/color_property.hpp>
-#include <moveit/rviz_plugin_render_tools/octomap_render.h>
 
 namespace octomap
 {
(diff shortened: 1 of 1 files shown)


tylerjw (approved):
Looks like something about ros control is failing to build in jammy.

JafarAbdi (approved):

henningkayser (changes requested):
Where do types like `Ogre::SceneNode` come from, if not from this header? Removing the include doesn't make any sense to me, as it must at least be included somewhere else in order to find the used types. Please clarify what's the issue here.

AndyZe:
> Please clarify what's the issue here.

I added the specific build error to the PR description. Maybe @tylerjw can provide more info on why this would be failing for our project's CI setup.

Clearly the header files are found _somewhere_ since CI passed.

AndyZe:
Note that this is the ONLY MoveIt2 PR passing CI right now. What the heck...

tylerjw:
@AndyZe it is bad to depend on transient includes (that something you include includes something you need).  The reason is that it makes so your code doesn't break when the things you include change what they include.

The way the pre-processor (the step of the compiler that processes lines that start with #) handles includes is it literally goes and finds the file you included and pastes it in place of the include line.  It does this recursively which means that if this PR works something you include is including the thing with the symbols (I'm assuming you are testing building something that links this file this PR and it compiles).  That means that there is some other file with the symbols needed here in the transient includes that can be found.  It would be much better if you went and figured out what that file is and include it here in place of this include instead of removing this one.

There is a second solution that would also get you the affect you want, and that is to change the cmake export set of this package. To do that you'll need to figure out what package exports this header and add it to the list of packages that cmake exports as a public dependency of this package. This is the apporach I was suggesting earlier but after thinking about his more I would prefer the first approach because it doesn't add another dependency to this package and keeps the set of dependencies one smaller.

AndyZe:
@henningkayser if you wouldn't mind giving it another review, I found a simple solution that doesn't involve deleting the header file  ;)   I was inspired by looking at [this file](https://github.com/ros-planning/moveit2/blob/be33e822fedb0b2fa106e996b59705975d34b343/moveit_ros/visualization/rviz_plugin_render_tools/include/moveit/rviz_plugin_render_tools/octomap_render.h), which you can see includes `OgrePrerequisites.h`, not `Ogre/OgrePrerequisites.h`. I guess something about Ogre installation must have changed recently.

tylerjw (approved):
This looks really good, thank you for digging into that.

PR #1100, opened 2022-03-07
Make lockSceneRead() and lockSceneWrite() protected member functions

AndyZe:
### Description

Previously there were 2 ways to lock the planning scene:

- `lockSceneRead()` or `lockSceneWrite()`
- Create a `LockedPlanningSceneRO` or `LockedPlanningSceneRW` object. Internally, this uses `lockSceneRead()` or `lockSceneWrite()`

The [tutorials recommend](https://moveit.picknik.ai/galactic/doc/examples/planning_scene_monitor/planning_scene_monitor_tutorial.html#planningscenemonitor) the LockedPlanningSceneRO approach. @v4hn and I [thought](https://github.com/ros-planning/moveit2/pull/1087#discussion_r812298593) it would be a good idea to make `lockSceneRead()` and `lockSceneWrite()` private so they don't get used by mistake.

`lockSceneRead()` and `lockSceneWrite()` don't get used very much in the codebase so this was actually a pretty simple change.

So far, I've tested with the Move Group C++ interface tutorial.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -3,6 +3,11 @@
 API changes in MoveIt releases
 
 ## ROS Rolling
+- `lockSceneRead()` and `lockSceneWrite()` are now protected member functions, for internal use only. To lock the planning scene, use LockedPlanningSceneRO or LockedPlanningSceneRW:
+```
+      planning_scene_monitor::LockedPlanningSceneRO ls(planning_scene_monitor);
+      moveit::core::RobotModelConstPtr model = ls->getRobotModel();
+```
 - ServoServer was renamed to ServoNode
 - `CollisionObject` messages are now defined with a `Pose`, and shapes and subframes are defined relative to the object's pose. This makes it easier to place objects with subframes and multiple shapes in the scene. This causes several changes:
     - `getFrameTransform()` now returns this pose instead of the first shape's pose.
--- a/moveit_ros/planning/moveit_cpp/src/planning_component.cpp
+++ b/moveit_ros/planning/moveit_cpp/src/planning_component.cpp
@@ -127,11 +127,11 @@ PlanningComponent::PlanSolution PlanningComponent::plan(const PlanRequestParamet
   planning_scene_monitor::PlanningSceneMonitorPtr planning_scene_monitor =
       moveit_cpp_->getPlanningSceneMonitorNonConst();
   planning_scene_monitor->updateFrameTransforms();
-  planning_scene_monitor->lockSceneRead();  // LOCK planning scene
-  planning_scene::PlanningScenePtr planning_scene =
-      planning_scene::PlanningScene::clone(planning_scene_monitor->getPlanningScene());
-  planning_scene_monitor->unlockSceneRead();  // UNLOCK planning scene
-  planning_scene_monitor.reset();             // release this pointer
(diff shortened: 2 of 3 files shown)


henningkayser (commented):
I think I'm fine with this change in general, even though deprecating them would be a sensible first step. Maybe I should revisit the attempt to improve the locking API in https://github.com/ros-planning/moveit/pull/2794.

AndyZe:
> though deprecating them would be a sensible first step.

I'll open a new PR for that. Then we can revisit this one in a few months or so.

> Maybe I should revisit the attempt to improve the locking API in [ros-planning/moveit#2794](https://github.com/ros-planning/moveit/pull/2794).

I think that's a good idea. It feels like there should be a better way but I'm not sure that PR is it. It would take some research.

AndyZe:
@henningkayser FYI an issue with marking this `deprecated` is we will get a lot of compiler warnings whenever the function is used internally. Like this:

```
/home/andy/ws_ros2/src/moveit2/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h:686:49: warning: ‘void planning_scene_monitor::PlanningSceneMonitor::lockSceneWrite()’ is deprecated [-Wdeprecated-declarations]
  686 |         planning_scene_monitor_->lockSceneWrite();
```

I doubt this is going to pass CI

v4hn (approved):
> I think that's a good idea. It feels like there should be a better way but I'm not sure that PR is it. It would take some research.

I feel the same way.

Assuming we keep the current locking interfaces for now, the change here looks good to me browsing through the patch in my browser.

> an issue with marking this deprecated is we will get a lot of compiler warnings whenever the function is used internally. Like this:

You could solve that that by renaming the internal method and providing a new deprecated method with the matching signature.
That being said I would not deprecate this interface before removal and ROS2 rolling should not need a deprecation for "minor" changes like that. Just add an `MIGRATIONS.md` entry with a replacement to copy&paste

Additionally I would propose to deprecate the `getPlanningScene()` method that goes along with these interfaces.
It's just ludicrous to have a method with its documentation saying [DON'T USE ME](https://github.com/ros-planning/moveit2/blob/main/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h#L189-L190).

It would also help usability to have a simple `getScene()` method in the locking classes instead of the [awful cast operator](https://github.com/ros-planning/moveit2/blob/main/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h#L651-L654).

AndyZe:
> Additionally I would propose to deprecate the getPlanningScene() method that goes along with these interfaces.
> It's just ludicrous to have a method with its documentation saying [DON'T USE ME](https://github.com/ros-planning/moveit2/blob/main/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h#L189-L190).
> 
> It would also help usability to have a simple getScene() method in the locking classes instead of the [awful cast operator](https://github.com/ros-planning/moveit2/blob/main/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h#L651-L654).

Agree! Will make an issue because I don't have time just now.

PR #1102, opened 2022-03-07
[MSA] Simplify loading of new SRDF

DLu:
### Description
Undoes https://github.com/ros-planning/moveit2/commit/78896a129a9e9c0bd12601fa45537a17f37edbbc to simplify the loading of new SRDF files, and then some minor other cleanup. 

### Checklist

### DIFF
--- a/moveit_setup_assistant/moveit_setup_core_plugins/src/start_screen.cpp
+++ b/moveit_setup_assistant/moveit_setup_core_plugins/src/start_screen.cpp
@@ -48,16 +48,7 @@ void StartScreen::onInit()
 void StartScreen::loadURDFFile(const std::string& urdf_file_path, const std::string& xacro_args)
 {
   urdf_config_->loadFromPath(urdf_file_path, xacro_args);
-}
-
-void StartScreen::setSRDFFile(const std::string& srdf_string)
-{
-  srdf_config_->loadSRDFString(srdf_string);
-}
-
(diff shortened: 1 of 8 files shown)


moveit_setup_assistant/moveit_setup_core_plugins/src/start_screen.cpp
```diff
@@ -48,16 +48,7 @@ void StartScreen::onInit()
...
-
-void StartScreen::setSRDFFile(const std::string& srdf_string)
-{
-  srdf_config_->loadSRDFString(srdf_string);
-}
-
-std::string StartScreen::getURDFName()
-{
-  return urdf_config_->getName();
+  srdf_config_->updateRobotModel();
```
henningkayser, line 51: what does this do? At this point, there is not really a RobotModel yet, right?

DLu: It [creates the shared RobotModel object](https://github.com/ros-planning/moveit2/blob/93be6385cdb294a7b82470c29fb98dde93e319a3/moveit_setup_assistant/moveit_setup_framework/src/srdf_config.cpp#L147), which is basically empty, but once the URDF is loaded, the robot name can be set. It's equivalent to the removed loading of the blank SRDF file.

vatanaksoytezer:
I tested this, works well and looks good to me.

PR #1103, opened 2022-03-07
[moveit_configs_utils] Minor fixes

DLu:
### Description

 * Fix a minor grammatical error
 * Change the ordering in package.xml
    * According to the [schema](http://download.ros.org/schema/package_format3.xsd) `author` comes after `maintainer` and `license`. 
    * While not a strict requirement, `test_depend`s  come after `exec_depend`

### Checklist

### DIFF
--- a/moveit_configs_utils/moveit_configs_utils/moveit_configs_builder.py
+++ b/moveit_configs_utils/moveit_configs_utils/moveit_configs_builder.py
@@ -34,7 +34,7 @@ Example:
     # Or to get all the parameters as a dictionary
     moveit_configs.to_dict()
 
-Each function in MoveItConfigsBuilder have a file_path as an argument which is used to override the default
+Each function in MoveItConfigsBuilder has a file_path as an argument which is used to override the default
 path for the file
 
 Example:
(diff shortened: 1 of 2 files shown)


JafarAbdi (approved):
Thanks! @DLu

henningkayser (approved):
Why wouldn't xmllint catch this? It doesn't even look like it runs for this package, even though it's listed as a test plugin https://github.com/ros-planning/moveit2/runs/4900042271?check_suite_focus=true#step:10:485

DLu:
The problem is that this is a Python package without [a CMake that calls `ament_lint_auto_find_test_dependencies`](https://github.com/ros-planning/moveit2/blob/419db3b3b0245e9857e212ef86e441e60a4b10e4/moveit_common/CMakeLists.txt#L8). However, the problem popped up in my editor, so I fixed it. 

Options:
 * Leave as is
 * Make this a hybrid CMake/Python package just so we can call the tests
 * Figure out how to get [ament_lint_auto](https://github.com/ament/ament_lint/blob/master/ament_lint_auto/doc/index.rst) to generate Python tests

tylerjw (approved):

PR #1104, opened 2022-03-08
Drop outdated/unnecessary workspace dependencies

henningkayser:
Drop OMPL and ros2_control when both are available as debians.


mikeferguson:
moveit2.repos file is already empty now, so this is implemented

PR #1107, opened 2022-03-08
Temporarily add galactic CI

vatanaksoytezer:
Rolling, ROS and almost everything in Jammy is broken due to [transition of Python version from 3.9 to 3.10](https://discourse.ros.org/t/preparing-for-rolling-sync-delayed-by-python-3-10-transition/24521/9). I am commenting out Rolling and adding Galactic back so we can at least merge PRs.


henningkayser (approved):
Thanks! I'm looking forward to some more stable times :/

JafarAbdi (approved):

PR #1111, opened 2022-03-10
Ruckig smoothing cleanup

### LINKED ISSUE #767: With the ruckig_traj_smoothing plugin, motion does not always end exactly at the desired robot position
I think what happens is, the smoother passes over every waypoint and modifies the positions enough that it does not end exactly at the final desired position. It "can't catch up."

Possible solution:

While iterating and stretching the trajectory, I think we should also check if the final waypoint matches the target. Keep stretching until it does.

Easy to reproduce if `velocity_scaling` is set very high, `acceleration_scaling` is set very low.

AndyZe:
### Description

Fixes #767 

This fixes some bugs in the Ruckig smoothing plugin and simplifies it. I tried to comment it better and simplify so it will be self-explanatory.

### Testing

You can test this easily with the MoveIt tutorials.

- Add Ruckig to the list of OMPL adapters. In `panda_moveit_config/config/ompl_planning.yaml`, add `AddRuckigTrajectorySmoothing` to the top of the list like this:

```
request_adapters: >-
  default_planner_request_adapters/AddRuckigTrajectorySmoothing
```

- Then launch the Panda with move_group:  `ros2 launch moveit_resources_panda_moveit_config demo.launch.py`

- You will know the Ruckig plugin loaded if you see this in terminal:

`Using planning request adapter 'Add Ruckig trajectory smoothing.`

### Data plots

Representative data from some very large trajectories (1000+ waypoints). We feed 10 waypoints at a time into this smoothing algorithm. Notice jerk decreased from 4000 rad/s^3 to 10 rad/s^3.

joint_1_positions
joint_1_vel_accel_jerk
joint_2_positions
joint_2_vel_accel_jerk

### DIFF
--- a/moveit_core/trajectory_processing/include/moveit/trajectory_processing/ruckig_traj_smoothing.h
+++ b/moveit_core/trajectory_processing/include/moveit/trajectory_processing/ruckig_traj_smoothing.h
@@ -31,7 +31,7 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  *******************************************************************************/
 
-/* Author: Jack Center, Wyatt Rees, Andy Zelenak */
+/* Author: Jack Center, Wyatt Rees, Andy Zelenak, Stephanie Eng */
 
 #pragma once
 
@@ -61,35 +61,6 @@ private:
                                  const moveit::core::RobotStatePtr& next_waypoint,
                                  const moveit::core::JointModelGroup* joint_group,
                                  ruckig::InputParameter<0>& ruckig_input);
-
-  /**
-   * \brief Check for lagging motion of any joint at a waypoint.
-   * \param joint_group     The MoveIt JointModelGroup of interest
-   * \param ruckig_input  Input parameters to Ruckig
-   * \param ruckig_output Output parameters from Ruckig
-   * \return              true if lagging motion is detected on any joint
-   */
-  static bool checkForLaggingMotion(const moveit::core::JointModelGroup* joint_group,
-                                    const ruckig::InputParameter<0>& ruckig_input,
-                                    const ruckig::OutputParameter<0>& ruckig_output);
-
-  /**
-   * \brief Return L2-norm of velocity, taking all joints into account.
-   * \param ruckig_input  Input parameters to Ruckig
-   * \param joint_group   The MoveIt JointModelGroup of interest
-   */
-  static double getTargetVelocityMagnitude(const ruckig::InputParameter<0>& ruckig_input,
-                                           const moveit::core::JointModelGroup* joint_group);
-
-  /**
-   * \brief Check if the joint positions of two waypoints are very similar.
-   * \param prev_waypoint State at waypoint i-1
-   * \param prev_waypoint State at waypoint i
-   * \joint_group         The MoveIt JointModelGroup of interest
-   */
-  static bool checkForIdenticalWaypoints(const moveit::core::RobotState& prev_waypoint,
-                                         const moveit::core::RobotState& next_waypoint,
-                                         const moveit::core::JointModelGroup* joint_group);
 
   /**
    * \brief Initialize Ruckig position/vel/accel. This initializes ruckig_input and ruckig_output to the same values
--- a/moveit_core/trajectory_processing/include/moveit/trajectory_processing/ruckig_traj_smoothing.h
+++ b/moveit_core/trajectory_processing/include/moveit/trajectory_processing/ruckig_traj_smoothing.h
@@ -46,8 +46,8 @@ class RuckigSmoothing
 {
 public:
   static bool applySmoothing(robot_trajectory::RobotTrajectory& trajectory,
-                             const double max_velocity_scaling_factor = 1.0,
-                             const double max_acceleration_scaling_factor = 1.0);
+                             double const max_velocity_scaling_factor = 1.0,
+                             double const max_acceleration_scaling_factor = 1.0);
 
 private:
   /**
--- a/moveit_core/trajectory_processing/include/moveit/trajectory_processing/ruckig_traj_smoothing.h
+++ b/moveit_core/trajectory_processing/include/moveit/trajectory_processing/ruckig_traj_smoothing.h
@@ -46,8 +46,8 @@ class RuckigSmoothing
 {
 public:
   static bool applySmoothing(robot_trajectory::RobotTrajectory& trajectory,
-                             double const max_velocity_scaling_factor = 1.0,
-                             double const max_acceleration_scaling_factor = 1.0);
+                             const double max_velocity_scaling_factor = 1.0,
+                             const double max_acceleration_scaling_factor = 1.0);
 
 private:
   /**
--- a/moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
+++ b/moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
@@ -45,12 +45,12 @@ namespace trajectory_processing
 {
 namespace
 {
-rclcpp::Logger const LOGGER = rclcpp::get_logger("moveit_trajectory_processing.ruckig_traj_smoothing");
-double constexpr DEFAULT_MAX_VELOCITY = 5;       // rad/s
-double constexpr DEFAULT_MAX_ACCELERATION = 10;  // rad/s^2
-double constexpr DEFAULT_MAX_JERK = 200;         // rad/s^3
-double constexpr MAX_DURATION_EXTENSION_FACTOR = 10.0;
-double constexpr DURATION_EXTENSION_FRACTION = 1.1;
+const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_trajectory_processing.ruckig_traj_smoothing");
+constexpr double DEFAULT_MAX_VELOCITY = 5;       // rad/s
+constexpr double DEFAULT_MAX_ACCELERATION = 10;  // rad/s^2
+constexpr double DEFAULT_MAX_JERK = 200;         // rad/s^3
+constexpr double MAX_DURATION_EXTENSION_FACTOR = 10.0;
+constexpr double DURATION_EXTENSION_FRACTION = 1.1;
 }  // namespace
 
 bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajectory,
@@ -64,7 +64,7 @@ bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajecto
     return false;
   }
 
-  size_t const num_waypoints = trajectory.getWayPointCount();
+  const size_t num_waypoints = trajectory.getWayPointCount();
   if (num_waypoints < 2)
   {
     RCLCPP_WARN(LOGGER,
@@ -76,7 +76,7 @@ bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajecto
   robot_trajectory::RobotTrajectory original_trajectory =
       robot_trajectory::RobotTrajectory(trajectory, true /* deep copy */);
 
-  size_t const num_dof = group->getVariableCount();
+  const size_t num_dof = group->getVariableCount();
 
   // This lib does not actually work properly when angles wrap around, so we need to unwind the path first
   trajectory.unwind();
@@ -89,15 +89,14 @@ bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajecto
   ruckig::OutputParameter<ruckig::DynamicDOFs> ruckig_output{ num_dof };
 
   // Initialize the smoother
-  const std::vector<int>& idx = group->getVariableIndexList();
   initializeRuckigState(ruckig_input, ruckig_output, *trajectory.getFirstWayPointPtr(), group);
 
   // Kinematic limits (vel/accel/jerk)
-  std::vector<std::string> const& vars = group->getVariableNames();
-  moveit::core::RobotModel const& rmodel = group->getParentModel();
+  const std::vector<std::string>& vars = group->getVariableNames();
+  const moveit::core::RobotModel& rmodel = group->getParentModel();
   for (size_t i = 0; i < num_dof; ++i)
   {
-    moveit::core::VariableBounds const& bounds = rmodel.getVariableBounds(vars.at(i));
+    const moveit::core::VariableBounds& bounds = rmodel.getVariableBounds(vars.at(i));
 
     // This assumes min/max bounds are symmetric
     if (bounds.velocity_bounded_)
@@ -159,7 +158,7 @@ bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajecto
               duration_extension_factor * original_trajectory.getWayPointDurationFromPrevious(time_stretch_idx));
           // re-calculate waypoint velocity and acceleration
           auto target_state = trajectory.getWayPointPtr(time_stretch_idx);
-          auto const prev_state = trajectory.getWayPointPtr(time_stretch_idx - 1);
+          const auto prev_state = trajectory.getWayPointPtr(time_stretch_idx - 1);
           timestep = trajectory.getAverageSegmentDuration();
           for (size_t joint = 0; joint < num_dof; ++joint)
           {
--- a/moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
+++ b/moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
@@ -45,26 +45,26 @@ namespace trajectory_processing
 {
 namespace
 {
-const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_trajectory_processing.ruckig_traj_smoothing");
-constexpr double DEFAULT_MAX_VELOCITY = 5;       // rad/s
-constexpr double DEFAULT_MAX_ACCELERATION = 10;  // rad/s^2
-constexpr double DEFAULT_MAX_JERK = 200;         // rad/s^3
-constexpr double MAX_DURATION_EXTENSION_FACTOR = 10.0;
-constexpr double DURATION_EXTENSION_FRACTION = 1.1;
+rclcpp::Logger const LOGGER = rclcpp::get_logger("moveit_trajectory_processing.ruckig_traj_smoothing");
+double constexpr DEFAULT_MAX_VELOCITY = 5;       // rad/s
+double constexpr DEFAULT_MAX_ACCELERATION = 10;  // rad/s^2
+double constexpr DEFAULT_MAX_JERK = 200;         // rad/s^3
+double constexpr MAX_DURATION_EXTENSION_FACTOR = 10.0;
+double constexpr DURATION_EXTENSION_FRACTION = 1.1;
 }  // namespace
 
 bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajectory,
                                      const double max_velocity_scaling_factor,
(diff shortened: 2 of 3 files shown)


moveit_core/trajectory_processing/include/moveit/trajectory_processing/ruckig_traj_smoothing.h
```diff
@@ -42,6 +42,11 @@
 
 namespace trajectory_processing
 {
+namespace
+{
+constexpr size_t RUCKIG_DYNAMIC_DOF = 0;  // Put Ruckig in "dynamic DOF" mode
```
pantor, line 47: Hey Andy, just saw this PR and wanted to say that there is also the `ruckig::DynamicDOFs` constant inside the Ruckig namespace that might be cleaner to use here.

pantor, line 47: Hey Andy, just saw this PR and wanted to say that there is also the `ruckig::DynamicDOFs` constant inside the Ruckig namespace that might be cleaner to use here.

AndyZe: Thanks - done

AndyZe:
I had an epiphany - I don't actually need to use Ruckig to smooth the waypoints in this application. I can use it as a pass/fail check for whether the trajectory duration is acceptable, and extend the duration if needed. This is a lot simpler and works great for a small number of waypoints.

For a large number of waypoints, this algorithm can be called repeatedly with small chunks of waypoints.

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -119,89 +121,61 @@ bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajecto
...
+        trajectory = robot_trajectory::RobotTrajectory(original_trajectory, true /* deep copy */);
+        for (size_t time_stretch_idx = 1; time_stretch_idx < num_waypoints; ++time_stretch_idx)
+        {
+          trajectory.setWayPointDurationFromPrevious(
+              time_stretch_idx,
+              duration_extension_factor * original_trajectory.getWayPointDurationFromPrevious(time_stretch_idx));
+          // re-calculate waypoint velocity and acceleration
+          auto target_state = trajectory.getWayPointPtr(time_stretch_idx);
+          auto const prev_state = trajectory.getWayPointPtr(time_stretch_idx - 1);
+          timestep = trajectory.getWayPointDurationFromStart(num_waypoints - 1) / (trajectory.getWayPointCount() - 1);
```
stephanie-eng, line 157: As above: consider using getAverageSegmentDuration?

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -119,89 +121,61 @@ bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajecto
...
+                                                  target_state->getVariableVelocity(move_group_idx.at(joint)));
+
+            double prev_velocity = prev_state->getVariableVelocity(move_group_idx.at(joint));
+            double curr_velocity = target_state->getVariableVelocity(move_group_idx.at(joint));
+            target_state->setVariableAcceleration(move_group_idx.at(joint), (curr_velocity - prev_velocity) / timestep);
+          }
+          target_state->update();
+        }
+        ruckig_ptr = std::make_unique<ruckig::Ruckig<ruckig::DynamicDOFs>>(num_dof, timestep);
+        waypoint_idx = 0;
```
stephanie-eng, line 171: Is this necessary? I'm trying to follow the code here: we break out of the for loop, and go back to the while loop, then restart the for loop where we reinitialize waypoint_idx to be 0.

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -69,25 +67,29 @@ bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajecto
   const size_t num_waypoints = trajectory.getWayPointCount();
```
stephanie-eng, line 67: Probably not a big deal - but I noticed you started using right const in your updates. We still have left consts from previously; can we be consistent with the placement of const throughout the file?

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -69,25 +67,29 @@ bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajecto
...
-    return false;
+    RCLCPP_WARN(LOGGER,
+                "Trajectory does not have enough points to smooth with Ruckig. Returning an unmodified trajectory.");
+    return true;
   }
 
+  // Cache the trajectory in case we need to reset it
+  robot_trajectory::RobotTrajectory original_trajectory = trajectory;
+
   const size_t num_dof = group->getVariableCount();
```
stephanie-eng, line 78: Right const?

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -69,25 +67,29 @@ bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajecto
...
+  double timestep = trajectory.getWayPointDurationFromStart(num_waypoints - 1) / (trajectory.getWayPointCount() - 1);
+  std::unique_ptr<ruckig::Ruckig<ruckig::DynamicDOFs>> ruckig_ptr;
+  ruckig_ptr = std::make_unique<ruckig::Ruckig<ruckig::DynamicDOFs>>(num_dof, timestep);
+  ruckig::InputParameter<ruckig::DynamicDOFs> ruckig_input{ num_dof };
+  ruckig::OutputParameter<ruckig::DynamicDOFs> ruckig_output{ num_dof };
 
   // Initialize the smoother
-  const std::vector<int>& idx = group->getVariableIndexList();
-  initializeRuckigState(ruckig_input, ruckig_output, *trajectory.getFirstWayPointPtr(), num_dof, idx);
+  const std::vector<int>& move_group_idx = group->getVariableIndexList();
```
stephanie-eng, line 91: Right const?

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -69,25 +67,29 @@ bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajecto
...
+  initializeRuckigState(ruckig_input, ruckig_output, *trajectory.getFirstWayPointPtr(), num_dof, move_group_idx);
 
   // Kinematic limits (vel/accel/jerk)
   const std::vector<std::string>& vars = group->getVariableNames();
```
stephanie-eng, line 95: Right const?

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -46,19 +46,21 @@ namespace trajectory_processing
...
-constexpr double MINIMUM_VELOCITY_SEARCH_MAGNITUDE = 0.01;  // rad/s. Stop searching when velocity drops below this
+constexpr double LAG_TOLERANCE = 0.005;  // unitless fraction, used in lead/lag detection
 }  // namespace
 
 bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajectory,
                                      const double max_velocity_scaling_factor,
                                      const double max_acceleration_scaling_factor)
 {
+  // Cache the trajectory in case we need to reset it
+  robot_trajectory::RobotTrajectory original_trajectory = trajectory;
```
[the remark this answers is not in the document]
AndyZe: that's a really good catch. Yes, absolutely.

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
AndyZe: Good point. Yep, not necessary.

AndyZe: I'll add a comment, too

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
AndyZe: Good point, I'll make all the things right-const. It really is a little easier to read, I think

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -69,25 +67,29 @@ bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajecto
...
   // This lib does not actually work properly when angles wrap around, so we need to unwind the path first
   trajectory.unwind();
 
   // Instantiate the smoother
-  double timestep = trajectory.getAverageSegmentDuration();
-  std::unique_ptr<ruckig::Ruckig<0>> ruckig_ptr;
-  ruckig_ptr = std::make_unique<ruckig::Ruckig<0>>(num_dof, timestep);
-  ruckig::InputParameter<0> ruckig_input{ num_dof };
-  ruckig::OutputParameter<0> ruckig_output{ num_dof };
+  double timestep = trajectory.getWayPointDurationFromStart(num_waypoints - 1) / (trajectory.getWayPointCount() - 1);
```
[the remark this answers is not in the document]
AndyZe: @stephanie-eng  actually this change caused CI to fail. I'm not sure why but I trust one of the CI tests a lot now, since it compares against an analytical solution. (The duration was off by a factor of 2.) I'll look into it more tomorrow.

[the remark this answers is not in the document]
AndyZe: Resolved -- the first waypoint in the test had a nonzero duration. So it was not an issue with `getAverageSegmentDuration()`

trajectory_->addSuffixWayPoint(robot_state, DEFAULT_TIMESTEP);

AndyZe:
CI fails because this PR is not available on Galactic yet. I think I'll backport it.  https://github.com/ros-planning/moveit2/pull/1079/files

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -45,59 +45,59 @@ namespace trajectory_processing
...
-constexpr double MINIMUM_VELOCITY_SEARCH_MAGNITUDE = 0.01;  // rad/s. Stop searching when velocity drops below this
+rclcpp::Logger const LOGGER = rclcpp::get_logger("moveit_trajectory_processing.ruckig_traj_smoothing");
+double constexpr DEFAULT_MAX_VELOCITY = 5;       // rad/s
+double constexpr DEFAULT_MAX_ACCELERATION = 10;  // rad/s^2
+double constexpr DEFAULT_MAX_JERK = 200;         // rad/s^3
+double constexpr MAX_DURATION_EXTENSION_FACTOR = 10.0;
+double constexpr DURATION_EXTENSION_FRACTION = 1.1;
 }  // namespace
 
 bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajectory,
```
stephanie-eng, line 56: Smallest of nitpicks: can we right const these too?

stephanie-eng (approved):

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -116,92 +116,72 @@ bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajecto
...
+      // Extend the trajectory duration if Ruckig could not reach the waypoint successfully
+      if (ruckig_result != ruckig::Result::Finished)
       {
-        trajectory.setWayPointDurationFromPrevious(
-            waypoint_idx, DURATION_EXTENSION_FRACTION * trajectory.getWayPointDurationFromPrevious(waypoint_idx));
-        // TODO(andyz): re-calculate waypoint velocity and acceleration here?
+        duration_extension_factor *= DURATION_EXTENSION_FRACTION;
+        // Reset the trajectory
+        trajectory = robot_trajectory::RobotTrajectory(original_trajectory, true /* deep copy */);
+        for (size_t time_stretch_idx = 1; time_stretch_idx < num_waypoints; ++time_stretch_idx)
```
henningkayser, line 155: wouldn't it make sense to only stretch until the current `waypoint_idx`?

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -45,59 +45,59 @@ namespace trajectory_processing
...
 {
-const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_trajectory_processing.ruckig_traj_smoothing");
-constexpr double DEFAULT_MAX_VELOCITY = 5;           // rad/s
-constexpr double DEFAULT_MAX_ACCELERATION = 10;      // rad/s^2
-constexpr double DEFAULT_MAX_JERK = 20;              // rad/s^3
-constexpr double IDENTICAL_POSITION_EPSILON = 1e-3;  // rad
-constexpr double MAX_DURATION_EXTENSION_FACTOR = 5.0;
-constexpr double DURATION_EXTENSION_FRACTION = 1.1;
-constexpr double MINIMUM_VELOCITY_SEARCH_MAGNITUDE = 0.01;  // rad/s. Stop searching when velocity drops below this
+rclcpp::Logger const LOGGER = rclcpp::get_logger("moveit_trajectory_processing.ruckig_traj_smoothing");
```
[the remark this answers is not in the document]
henningkayser: Also, should we really have magic default values in here? I'd argue that it would probably be safer to just require configured limits

henningkayser (commented):
Changes look reasonable to me in general. Please consider switching back to west const for consistency, otherwise just minor comments.

moveit_core/trajectory_processing/include/moveit/trajectory_processing/ruckig_traj_smoothing.h
```diff
@@ -46,8 +46,8 @@ class RuckigSmoothing
 {
 public:
   static bool applySmoothing(robot_trajectory::RobotTrajectory& trajectory,
-                             const double max_velocity_scaling_factor = 1.0,
-                             const double max_acceleration_scaling_factor = 1.0);
+                             double const max_velocity_scaling_factor = 1.0,
```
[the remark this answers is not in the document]
AndyZe: Tyler has put right-const in my head. I'll revert it. It doesn't matter much to me either way alway I slightly prefer right-const.

moveit_core/trajectory_processing/include/moveit/trajectory_processing/ruckig_traj_smoothing.h
```diff
@@ -58,36 +58,10 @@ class RuckigSmoothing
...
-  /**
-   * \brief Check if the joint positions of two waypoints are very similar.
-   * \param prev_waypoint State at waypoint i-1
-   * \param prev_waypoint State at waypoint i
-   * \joint_group         The MoveIt JointModelGroup of interest
-   */
-  static bool checkForIdenticalWaypoints(const moveit::core::RobotState& prev_waypoint,
-                                         const moveit::core::RobotState& next_waypoint,
-                                         const moveit::core::JointModelGroup* joint_group);
+  static void getNextRuckigInput(ruckig::OutputParameter<ruckig::DynamicDOFs> const& ruckig_output,
```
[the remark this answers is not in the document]
AndyZe: I'll revert it everywhere except (I would argue) right here, where it really makes sense:

`moveit::core::JointModelGroup const* const group = trajectory.getGroup();`

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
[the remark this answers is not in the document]
AndyZe: I think the default (vel/accel/jerk) values do make sense. There are a ton of robots out there that don't even have manufacturer-specified acceleration limits and I don't think any of them have jerk limits specified. It's nice to have these safe default values that just work nicely and prevent hardware damage without the user needing to do anything.

If we wanted to move these parameters to a yaml file, where would they even go? I guess they would belong in `ompl_planning.yaml` because that's where the Ruckig plugin is specified.

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -116,92 +116,72 @@ bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajecto
...
       ruckig_input.max_acceleration.at(i) = max_acceleration_scaling_factor * DEFAULT_MAX_ACCELERATION;
     }
+    if (bounds.jerk_bounded_)
+    {
+      ruckig_input.max_jerk.at(i) = bounds.max_jerk_;
+    }
+    else
+    {
+      ruckig_input.max_jerk.at(i) = DEFAULT_MAX_JERK;
+    }
```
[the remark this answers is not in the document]
AndyZe: nice.

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
AndyZe: I think that's good logic and it should work but it complicates the code for a couple reasons:

- You have to worry about the index zero edge case. (`waypoint_idx == 0`)
- The tracking of `duration_extension_factor` gets tricky

Overall I don't think it's worth the complication. Usually MoveIt trajectories are pretty small, like 100 waypoints or less. If you're working with very large trajectories like the ~7,000 waypoints I showed above, you should feed 10 at a time or so into this algorithm.

Anyway, I did try this replacement but it often fails:

```
        // Reset the trajectory
        trajectory = robot_trajectory::RobotTrajectory(original_trajectory, true /* deep copy */);

        // If failure occurred at index 0, stretch up to index 1.
        // (Stretching index 0 is meaningless)
        // Else, stretch up to (current waypoint)
        size_t stretch_to_idx = (waypoint_idx == 0) ? 1 : waypoint_idx + 1;

        for (size_t time_stretch_idx = 1; time_stretch_idx <= stretch_to_idx; ++time_stretch_idx)
        {
          // Do stretching here...
```

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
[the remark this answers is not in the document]
gavanderhoorn: If there are standard magic values, it could be nice to info / warn users about those being used.

There have been many questions on ROS Answers along the lines of "why is my robot so slow when using MoveIt?", which almost always turned out to be due to the time-parameterisation algorithms using the default `1.0m/s` and/or `1.0m/s^2` limits hard-coded in them (without any mention of this to the user).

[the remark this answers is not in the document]
AndyZe: OK, I'll add a helpful hint if no vel/accel/jerk limit is defined.

AndyZe:
The CI failure is unrelated

henningkayser (approved):

AndyZe:
> What do you think about having it inherited from [TimeParameterization](https://github.com/ros-planning/moveit2/blob/main/moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_parameterization.h) similar to the other paramterization algorithms? not required here we could add it as a first-time issue

I'm not sure about that. It's not really a time-parameterization algorithm. It just does smoothing.

Issue #1038, opened 2022-01-27
Remove outdated and unused code


henningkayser:
This epic is meant to coordinate efforts that make MoveIt leaner which helps us with keeping up with maintenance and and with making bigger changes in the future. It is also a first step for disentangling the code base.
Here is a list of things that can either be removed or better be replaced with other libraries.

**moveit_core**
- ~~background_processing: only used for visualization, also there are probably better options for this. Move out of moveit_core or replace with better third-party library~~
- ~~backtrace: not used or needed, remove~~
- dynamics_solver: remove, this should be a property of an IK solver if needed, not a library in moveit_core, only used in MotionPlanningPanel for showing dynamic metrics. If we want to keep it, it doesn’t belong into moveit_core
- exceptions: remove, we don’t use exceptions (https://github.com/ros-planning/moveit2/pull/1117)
- kinematics_metrics: remove, only used in MotionPlanningPanel. If we want to keep it, it doesn’t belong into moveit_core
- macros: move out of moveit_core or remove entirely? we might not want to use these in moveit_core, but in general I don’t care too much about these
- ~~profiler: remove and replace with external tool like ros2_tracing?~~
- python: move into separate package
- ~~sensor_manager: remove, only used in plan_with_sensing which should be removed as well.~~
    - ~~I think there would be a need for a “SensorManager” which really only maintains the available sensors (camera, laser scanner etc) and provides interfaces for these, but this would be a separate effort~~
- transforms: not sure about this one, it’s used very often but it seems like this should be an external lib like geometry2
- utils: different package? at least split into separate test_utils

**moveit_ros**
- ~~planning/plan_execution/plan_with_sensing~~
- ~~manipulation~~

**moveit_planners**
- ~~sbpl~~

v4hn:
I just saw this issue, so let me give my two cents for the discussed modules

* ROS manipulation

yes, the interfaces in MGI should target the MTC capability instead (my fault for not reviewing/reworking the plugin there, I believe it needs more work)

* controller_manager should also be removed

Yes and no. It should be *severely* simplified at least and doesn't need the plugin interface. I do see a lot of merit in being able to specify which controllers should be used to actuate a given trajectory inside MoveIt. But to do that we could simply forward the requested controllers to the typical ros_control service API (assuming it exists) before execution. The obscure logic to decide the controller to use before execution should just be dropped.  My main point about this years ago was that ros_control does it better anyway, so why even give yet another plugin API in MoveIt.

[...]

PR #1117, opened 2022-03-11
Remove MoveIt exceptions

henningkayser:
This is as part of #1038. I think using something like std::optional or std::expected is much nicer than throwing exceptions, especially with constructors. There are still some other core classes that use exceptions, with this I wanted to continue the discussion about API changes like this.

### DIFF
--- a/moveit/scripts/create_readme_table.py
+++ b/moveit/scripts/create_readme_table.py
@@ -1,7 +1,5 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-# Usage: python moveit/moveit/scripts/create_readme_table.py > /tmp/table.md
-# First update supported_distro_ubuntu_dict below!
 
 # Copyright 2021 PickNik Inc.
 #
@@ -30,6 +28,9 @@
 # CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
 # ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
+
+# Usage: python moveit/moveit/scripts/create_readme_table.py > /tmp/table.md
+# First update supported_distro_ubuntu_dict below!
 
 
 from __future__ import print_function
--- a/moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
+++ b/moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
@@ -55,6 +55,7 @@
 #include <boost/function.hpp>
 #include <boost/concept_check.hpp>
 #include <memory>
+#include <optional>
 #include "rclcpp/rclcpp.hpp"
 
 #include "moveit_planning_scene_export.h"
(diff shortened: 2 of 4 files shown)


v4hn:
Yes to getting rid of `boost::noncopyable` (hell, that was in there? :confused: ) and also bye bye `createRobotModel` (redundant method).

I would actually propose to just remove the `PlanningScene(urdf_model, srdf_model)` constructor right away instead of deprecating it. There is not much use in constructing a `PlanningScene` that way, because a `RobotModel` holds the same information and is the much more appropriate datastructure here.
Removing this constructor already suffices to get rid of the exceptions though and I don't think it's useful to move to a monadic factory method.

In theory I agree with the idea of monadic returns. I also agree with the idea of factories when constructors would need to "do some work". But in this case there is not much going on in the constructor and I do think
```
planning_scene::PlanningScene scene{ robot_model }
```
or even
```
auto scene{ planning_scene::PlanningScene{ robot_model } }
```
reads a lot better than
```
auto scene{ planning_scene::PlanningScene::create(robot_model).value() }
```

AndyZe:
> I do think
> 
> ```
> planning_scene::PlanningScene scene{ robot_model }
> ```
> 
> or even
> 
> ```
> auto scene{ planning_scene::PlanningScene{ robot_model } }
> ```
> 
> reads a lot better than
> 
> ```
> auto scene{ planning_scene::PlanningScene::create(robot_model).value() }
> ```

Agree with this!  I'm not a fan of "always auto"

henningkayser:
> I would actually propose to just remove the PlanningScene(urdf_model, srdf_model) constructor right away instead of deprecating it.

Sure thing, will do!

The point of the `create()` function is that it's allowed to fail gracefully. With the constructor, we have no other choice but throwing an exception or flagging the instance invalid. I agree, however, that a plain constructor would be nice.

v4hn:
> The point of the `create()` function is that it's allowed to fail gracefully. With the constructor, we have no other choice but throwing an exception or flagging the instance invalid.

 I'm aware of that, but there are simply no things left in the constructor now that need graceful failure handling. 🙂
and adding the additional `.value()` everywhere is only worth the hassle when there are failure cases.

If you want to propose consistent factory methods for all classes as a standard interface, that's different (and a lot of rewriting). but then we should check how much that would blow up the caller side using moveit interfaces.

henningkayser:
> > The point of the `create()` function is that it's allowed to fail gracefully. With the constructor, we have no other choice but throwing an exception or flagging the instance invalid.
> 
> I'm aware of that, but there are simply no things left in the constructor now that need graceful failure handling. slightly_smiling_face and adding the additional `.value()` everywhere is only worth the hassle when there are failure cases.

RobotModel could be null

v4hn:
> RobotModel could be null

And you are telling me because the user can pass a nullptr to the constructor I need to call `.value()` on each method call?

How would you handle the case where your create method returns no value? That error condition gives you nothing you can handle logically without knowing where the model came from. Why should I consider that case in my code if I can also consider a case right before the call for my input parameter to be null? This should probably be checked already wherever you construct your robot model.

If we want to check for nullptr parameter in the constructor (which still makes sense) I would recommend `assert`, or possibly an assert that is active even with `-DNDEBUG`.

tylerjw:
Sadly this work seems to be abandoned. I'll open an issue to track continuing this work.

PR #1131, opened 2022-03-18
move_group launch for moveit_configs_utils

DLu:
### Description

Building off of #1113, this PR makes a launch file for move group. 

Small unrelated change: The first commit just renames some of the generation functions to make things easier in MSA. 

Two key changes to `moveit_configs_builder.py`
 * Calling `planning_pipelines` with no arguments would previously result in only OMPL being loaded. Now we use `pathlib` to look for files matching `*_planning.yaml` to find multiple different pipelines to be loaded. 
   * If there are no `*_planning.yaml` parameters in the `moveit_config` package, then by default, the ones defined in this package are loaded, i.e. 
     * `chomp_planning.yaml`
     * `ompl_planning.yaml`
     * `trajopt_planning.yaml` (*Is this still a thing?*)
     * `pilz_industrial_motion_planner_planning.yaml`
   * As an added bonus, if there aren't any `planner_configs` defined by the OMPL config, the values are loaded from `ompl_defaults.yaml`. 
 * Similarly, the behavior for calling `trajectory_execution` with no arguments is also updated. Previously, it would attempt to load `f"{self.__robot_name}_controllers.yaml"`. Now we search for `*_controllers.yaml` in the `moveit_config`. If only one is found, we use that. Otherwise, it gets complicated, but it tries a couple other common names for the default parameter file, including using the robot name. 

Details about `generate_move_group_launch`:
 * There's a little hacky bit in `add_debuggable_node` to semi-neatly get `--debug` passed into `move_group` as an argument. 
 * There still needs to be something to load the perception config in. 

### Checklist


AndyZe:
I just want to check you're aware that we will want the option to load `joint_limits.yaml` soon. It's for a client project so it will be important to get merged at some point. From grepping, it looks like `moveit_configs_builder.py` has functionality to load `joint_limits.yaml` but it's not mentioned in any of the launch files yet.

I made a [PR to moveit_resources](https://github.com/ros-planning/moveit_resources/pull/123) that might be useful for inspiration. It would be sweet if you can add it to this PR but I don't mind it being a followup either.

moveit_configs_utils/config/chomp_planning.yaml
```diff
@@ -0,0 +1,11 @@
+planning_plugin: chomp_interface/CHOMPPlanner
```
AndyZe, line 1: Suggest renaming this folder to `default_config`

moveit_configs_utils/config/trajopt_planning.yaml
```diff
@@ -0,0 +1,8 @@
+planning_plugin: trajopt_interface/TrajOptPlanner
```
AndyZe, line 1: I'm not sure what the state of traj_opt is but I don't think it's used enough to justify being here.

moveit_configs_utils/default_configs/pilz_industrial_motion_planner_planning.yaml
```diff
@@ -0,0 +1,5 @@
+planning_plugin: pilz_industrial_motion_planner::CommandPlanner
```
henningkayser, line 1: Better use the specified plugin name. We've seen weird issues when mixing plugin names with class names (even though it shouldn't make a difference).
```suggestion
planning_plugin: pilz_industrial_motion_planner/CommandPlanner
```

moveit_configs_utils/moveit_configs_utils/launches.py
```diff
@@ -132,3 +133,56 @@ def generate_warehouse_launch(moveit_config):
...
+    add_debuggable_node(
+        ld,
+        package="moveit_ros_move_group",
+        executable="move_group",
+        commands_file=str(moveit_config.package_path / "launch" / "gdb_settings.gdb"),
+        output="screen",
+        parameters=move_group_params,
+        extra_debug_args=["--debug"],
+        # Set the display variable, in case OpenGL code is used internally
+        additional_env={"DISPLAY": ":0"},
```
henningkayser, line 186: is this really necessary? I don't think it's our responsibility to override DISPLAY

moveit_configs_utils/moveit_configs_utils/moveit_configs_builder.py
```diff
@@ -53,11 +53,36 @@
...
+import re
 from ament_index_python.packages import get_package_share_directory
 
 from launch_param_builder import ParameterBuilder, load_yaml, load_xacro
 
 
+moveit_configs_utils_path = Path(get_package_share_directory("moveit_configs_utils"))
+
+
+def get_pattern_matches(folder, pattern):
```
henningkayser, line 65: Why not use [glob](https://docs.python.org/3/library/glob.html) here?

moveit_configs_utils/default_configs/ompl_defaults.yaml
```diff
@@ -0,0 +1,130 @@
+move_group:
```
henningkayser, line 1: why is this in the move_group namespace?

henningkayser (commented):
Looks good to me overall. @JafarAbdi should have a closer look at this

moveit_configs_utils/default_configs/pilz_industrial_motion_planner_planning.yaml
DLu: Copied from [here](https://github.com/ros-planning/moveit2/blob/9e30bdc77d0a9386dfdc50766e782de8be155010/moveit_setup_assistant/templates/moveit_config_pkg_template/launch/pilz_industrial_motion_planner_planning_pipeline.launch.xml#L6)

moveit_configs_utils/moveit_configs_utils/launches.py
DLu: Copied from [here](https://github.com/ros-planning/moveit2/blob/9e30bdc77d0a9386dfdc50766e782de8be155010/moveit_setup_assistant/templates/moveit_config_pkg_template/launch/move_group.launch#L85)

DLu:
In trying to figure out the best way to get the tests to pass, I'm waiting for 
 * https://github.com/ros-planning/moveit_resources/pull/127
 * https://github.com/ros-planning/moveit_resources/pull/120

moveit_configs_utils/default_configs/chomp_planning.yaml
```diff
@@ -0,0 +1,11 @@
+planning_plugin: chomp_interface/CHOMPPlanner
+enable_failure_recovery: true
+jiggle_fraction: 0.05
+request_adapters: "default_planner_request_adapters/AddTimeParameterization
+                   default_planner_request_adapters/ResolveConstraintFrames
+                   default_planner_request_adapters/FixWorkspaceBounds
+                   default_planner_request_adapters/FixStartStateBounds
+                   default_planner_request_adapters/FixStartStateCollision
+                   default_planner_request_adapters/FixStartStatePathConstraints"
```
JafarAbdi, line 9: Super nit: [existing `*_planning.yaml`](https://github.com/ros-planning/moveit_resources/blob/ros2/panda_moveit_config/config/ompl_planning.yaml#L2-L8) and https://github.com/ros-planning/moveit_resources/pull/120/files uses `>-` for multiline string, could you please unify them for `*_planning.yaml` files?

moveit_configs_utils/default_configs/ompl_planning.yaml
```diff
@@ -0,0 +1,9 @@
+planning_plugin: ompl_interface/OMPLPlanner
```
JafarAbdi, line 1: Any reason to split this from the other ompl parameters file?

moveit_configs_utils/moveit_configs_utils/moveit_configs_builder.py
```diff
@@ -345,15 +370,43 @@ def trajectory_execution(
...
                 )
-            )
-        )
+            else:
+                chosen_name = None
+                if len(possible_names) == 1:
+                    chosen_name = possible_names[0]
+                else:
+                    # Try a couple other common names, in order of precedence
+                    for name in [self.__robot_name, "ros", "ros2"]:
```
JafarAbdi, line 390: `ros/ros2_controllers.yaml` have to be loaded to ros2 controllers not moveit's trajectory execution manager, you renamed the file in https://github.com/ros-planning/moveit_resources/pull/127/files to moveit_controllers.yaml, what do you think about using that as a possible name?

```suggestion
                    for name in [self.__robot_name, "moveit"]:
```

moveit_configs_utils/moveit_configs_utils/moveit_configs_builder.py
```diff
@@ -345,15 +370,43 @@ def trajectory_execution(
...
+                        raise RuntimeError(
+                            "trajectory_execution: "
+                            f"Unable to guess which parameter file to load. Options:\n - {option_str}"
+                        )
+                file_path = config_folder / (chosen_name + "_controllers.yaml")
+
+        else:
+            file_path = Path(file_path)
+            if not file_path.is_absolute():
+                file_path = self._package_path / file_path
```
JafarAbdi, line 407: Is this needed? as far as I know, having `absolute path #1` / `absolute path #2` gives `absolute path #2` so I think using `self._package_path / file_path` directly below should be fine

JafarAbdi (commented):
Looks good to me, I lift a few small comments

Are you ok with me pushing a commit to this branch? I'm going to add a logging mechanism so a user knows what files were used for launching

moveit_configs_utils/moveit_configs_utils/moveit_configs_builder.py
DLu: Because this method returns the proper group (i.e. the part before `_controllers.yaml`). So either I have this custom method OR I use glob, then have a separate regex for grabbing the group. It's messy.

moveit_configs_utils/default_configs/ompl_planning.yaml
DLu: So if there's one file, the options are
 * Use the unified `default_configs/ompl_planning.yaml` options
 * Specify your own `ompl_planning.yaml`. 

Each of these parameters will have the standard parameters (`planning_plugin` and `request_adapters`) and all the `planner_configs`. 

By splitting the `planner_configs` into their own yaml that is loaded via special rule, the options are more flexible. 
 * Use all the ompl parameters from `default_configs`
 * Specify our own `ompl_planning.yaml` with changes to `request_adapters` etc. 
 * Specify our own `ompl_planning.yaml` that also includes `planner_configs`. 

Hence, if we just wanted to add a `request_adapter` or change some other base parameter (outside the `planner_configs`), we don't have to manually copy all the `planner_configs` to our new parameter file.

JafarAbdi (approved):

DLu:
https://getyarn.io/yarn-clip/b02dd583-32dc-4836-bce8-e193078c76e7

vatanaksoytezer (approved):

PR #1139, opened 2022-03-25
[pilz_industrial_motion_planner] Add includes

galou:

### Description

Add missing `#include`.

### Checklist


henningkayser (commented):
I don't really get why these would be needed. Did you run into build issues? At least we shouldn't have to include them inside both header and source file.

galou:
I just add all includes that I cite in the given file, i.e. possibly repeating an include in the source and associated header files. Just close the PR if you don't want this.

tylerjw:
@ChrisThrasher what do you think of this PR? It doesn't seem to harm anything but also I'm not sure it is necessary. Should we decide on a policy of when you should include a header and where?

ChrisThrasher:
Are these additional headers necessary to fix something?

galou:
As stated, these includes are not strictly necessary.

tylerjw:
@galou thank you for the clarification. I'm going to close this for now. We should write a policy around what and when we include various things. There are many opinions about how we should use includes and we lack consistency.

PR #1141, opened 2022-03-25
[pilz_fix_config] Fix planners' configuration

galou:

### Description

I could not load the PILZ pipeline for MoveItCpp (maybe it works without this with only one pipeline but I needed two). To solve the issue I copied some code from the OMPL plugin.

The code is prepare to take future parameters into account, though it may look awkward to have this now if we don't have any parameter. The problem is that I don't know what I can remove to have MoveItCpp work with Pilz.

### Checklist


henningkayser:
@galou could you give some details as to why you weren't able to load the planner for multiple groups? I don't see how your changes would be related. I agree that it's a little odd to copy config loader code even though it's not really used yet. I would prefer a more direct fix of your issue.

galou:
I guess this is related to `CommandPlanner` missing a call to `setPlannerConfigurations()` in its constructor. I'll try to find out how I can reduce the amount of not-yet-used code.

galou:
I tried to simplify `setPlannerConfigurations()` to keep only what is necessary. Without this, I cannot use Pilz Industrial Motion Planner with MoveItCpp (maybe it'd work if I had only the Pilz pipeline but I have multiple pipelines).

tylerjw:
@sea-bass you are currently using the Pilz planner. Would you mind reviewing this change?

moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/pilz_industrial_motion_planner.h
```diff
@@ -134,6 +134,11 @@ class CommandPlanner : public planning_interface::PlannerManager
   pilz_industrial_motion_planner::JointLimitsContainer aggregated_limit_active_joints_;
 
   /// cartesian limit
+  pilz_industrial_motion_planner::CartesianLimit cartesian_limit_;
```
sea-bass, line 137: I think this line is causing build failures, and seems we're not using it. Try removing it.

sea-bass (commented):
What is interesting to me is that basically this same code was actually already in `pilz_industrial_motion_planner.cpp` and was just removed 3 days ago in this PR : https://github.com/ros-planning/moveit2/pull/1710.

As that PR hints, since that is merged we then have to revert https://github.com/ros-planning/moveit2/pull/1522 -- and this PR is the Pilz part of this, but we also need to do this for CHOMP.

... so should we just revert https://github.com/ros-planning/moveit2/pull/1522 and close this?

@sjahr FYI

sjahr:
I think this PR is outdated because @rhaschke thankfully cleaned up the pipeline lookup in moveit_cpp (see https://github.com/ros-planning/moveit2/pull/1710/files#diff-e586c154422c4e479f62d2450db5d7f68adac4fb022f1fb916e26de8316d9061). With his PR merged, it is not required anymore by a planning pipeline to have a PlannerConfig in order to be found by moveit_cpp. Before that PR it was fixed in a similar way as this PR does but now there is no point to add an empty config now. @galou Can you confirm that PILZ with moveit_cpp on the latest version of main works for you?

PR #1144, opened 2022-03-28
[moveit_cpp] Fix return value of execute

galou:
Previously, `MoveItCpp::execute()` would return true only if
`waitForExecution()` would return `UNKNOWN`.
This sort of break the API because users that did use the return value
of `execute` had to use an inverted logic.
These users will quickly notice the change.

### Description

Previously, `MoveItCpp::execute()` would return true only if
`waitForExecution()` would return `UNKNOWN`.
This sort of break the API because users that did use the return value
of `execute` had to use an inverted logic.
These users will quickly notice the change.

The returned value is not documented in the [API documentation](https://moveit.picknik.ai/galactic/api/html/classmoveit__cpp_1_1MoveItCpp.html#a5ca934bc472fc16cb8ca62c5263448cd).

### Checklist

### DIFF
--- a/moveit_ros/planning/moveit_cpp/src/moveit_cpp.cpp
+++ b/moveit_ros/planning/moveit_cpp/src/moveit_cpp.cpp
@@ -43,6 +43,8 @@
 #endif
 #include <geometry_msgs/msg/quaternion_stamped.hpp>
 
+#include <moveit/controller_manager/controller_manager.h>
+
 namespace moveit_cpp
 {
 static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit.ros_planning_interface.moveit_cpp");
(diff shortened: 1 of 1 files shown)


galou:
I actually think that it would be better to return a `moveit_controller_manager::ExecutionStatus::Value` to better inform the caller, i.e. `trajectory_execution_manager_->waitForExecution()` when blocking and `moveit_controller_manager::ExecutionStatus::SUCCESS` otherwise. This would break API compatibility.

v4hn:
1. I agree it makes sense to return an `ExecutionStatus` instead of the boolean. Why would you return `Value` directly? `ExecutionStatus` has a `bool()` operator, so that would not even break API.

But that brings me to my second question 2. `ExecutionStatus::operator bool()` performs the same comparison that you propose to add here? Why does that not already work the way you want it to?
I see an API that could be improved, but I do not yet see the bug you describe. :confused:

galou:
Oh, I didn't notice that `ExecutionStatus` has a `bool()` operator, so in this case, of course, it's better to return it rather than its `Value`.
It's possible that I made a mistake when testing my code, sorry if it's the case (I'll double-check). Could it be that the `bool()` operator will only work in combination with `static_cast`: `return static_cast<bool>(trajectory_execution_manager_->waitForExecution());` in `MoveItCpp::execute`?

galou:
I just double-checked. I don't know what happened when I tested it but the result is what is actually expected. Sorry about it. A `static_cast` would be more explicit but the result would not change.
I close the PR.

v4hn:
> I close the PR.

I'm in favor of adapting the return value of `execute` though as discussed. Could you create a request for that instead?

Issue #1127, opened 2022-03-18
`execute()` and `stop()` both call `spin_some()`, thus crash


galou:
### Description

Despite #503 and #506, you cannot call `MoveGroupInterface::stop()` in a separate thread while `MoveGroupInterface::execute()` is running. This is related to thread-safety but I couldn't find information about thread-safety of `MoveGroupInterface`.

### Your environment
* ROS Distro: Galactic
* OS Version: Ubuntu 20.04
* Source build, branch `main` (419db3b3b0).

### Steps to reproduce
Call `move_group->execute()`, then, before `execute` returns, call `move_group->stop()` in another thread.

### Expected behaviour
The executable should not crash and the trajectory execution should stop.

### Actual behaviour
The executable crashes with
```
terminate called after throwing an instance of 'std::runtime_error'
what():  Node has already been added to an executor.
```

### Backtrace or Console output

[Backtrace](https://gist.github.com/galou/c3fd463fd4794bbdfc31908443bc9fbb).

galou:
Please note that I'm willing to fix this myself (in the range of my knowledge of MoveIt) but I don't know exactly which strategy to use. I even think that this should be solved in `rclcpp` to allow calling `spin_some()` concurrently.

henningkayser:
@galou I think the MoveGroupInterface was never designed to be thread-safe, and I don't think there is a good way to make the full API work reliably across multiple threads. What you can do is to use `MoveGroupInterface::asyncExecute()` for your use case. However, I agree that spinning should not be implemented explicitly inside MGI, but removing all spin calls is a somewhat bigger effort. Would you be interested in working on that?

galou:
I can work on it but I have a question first. Why is there a call to `spin_some()` in `MoveGroupInterface::stop()`? Are there some callbacks involved? `stop()` just contains a call to `publish()`.

My suggestion would be to add a `std::atomic_bool is_spinning_` member variable and use it to conditionally call `spin_some()`. Would it be OK?

henningkayser:
@galou #1305 should fix this bug. Could you verify and close this issue?

henningkayser:
spin_some has been removed a while ago, closing

PR #1157, opened 2022-03-31
[moveit_ros_benchmarks] Add missing dependency

galou:

### Description

`moveit_ros_benchmarks` doesn't compile without `merge-install` without this.

### Checklist


henningkayser:
> Maybe someone can explain me what's happening there. The stderr is
> 
> ```
> --- stderr: moveit_ros_benchmarks                                                                                                                                                                                   
> /home/libor/robocon_ros2_monows/src/moveit2/moveit2/moveit_ros/benchmarks/src/BenchmarkExecutor.cpp: In member function ‘virtual void moveit_ros_benchmarks::BenchmarkExecutor::writeOutput(const moveit_ros_benchma
> rks::BenchmarkExecutor::BenchmarkRequest&, const string&, double)’:                                                                                                                                                 
> /home/libor/robocon_ros2_monows/src/moveit2/moveit2/moveit_ros/benchmarks/src/BenchmarkExecutor.cpp:1114:31: error: ‘MOVEIT_VERSION_STR’ was not declared in this scope; did you mean ‘MOVEIT_VERSION_EXTRA’?       
>  1114 |   out << "MoveIt version " << MOVEIT_VERSION_STR << '\n';                                                                                                                                                   
>       |                               ^~~~~~~~~~~~~~~~~~                                                                                                                                                            
>       |                               MOVEIT_VERSION_EXTRA                                                                                                                                                          
> make[2]: *** [CMakeFiles/moveit_ros_benchmarks.dir/build.make:76: CMakeFiles/moveit_ros_benchmarks.dir/src/BenchmarkExecutor.cpp.o] Error 1                                                                         
> make[1]: *** [CMakeFiles/Makefile2:109: CMakeFiles/moveit_ros_benchmarks.dir/all] Error 2                                                                                                                           
> make: *** [Makefile:141: all] Error 2
> ```
> 
> As you can see, the compiler doesn't complain about the fact that it didn't find `moveit/version.h`. So, how is it possible that the macro is not defined if the file was found?

This probably happens because of conflicts with an already installed debian version of `moveit_core`, or possibly an underlay (do you have any of those?). The `MOVEIT_VERSION_STR` was just added recently and will not be present in older header files. Btw, adding `moveit_core` is not wrong, but it should also not be necessary as it's already provided and exported with `moveit_ros_planning` which is already included.

galou:
I do have `moveit_core` system package installed and I need it for another workspace, where I'd rather not pull the whole moveit source. Thanks for the explanation!

I'd rather include `moveit_core` as this avoids `moveit_ros_benchmarks` making the assumption that another package includes it.

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
```diff
@@ -1012,7 +1012,7 @@ class MoveGroupInterface::MoveGroupInterfaceImpl
       std_msgs::msg::String event;
       event.data = "stop";
       trajectory_event_publisher_->publish(event);
-      rclcpp::spin_some(pnode_);
+      // rclcpp::spin_some(pnode_);
```
JafarAbdi, line 1015: I agree that there's no need to spin for publishing, please either remove this line or revert commenting this line

JafarAbdi (commented):
Just one request other than that LGTM, multiple headers are being included from moveit_core without having it as a dependency. thanks for the fix

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
galou: Ups, the change was not supposed to be committed.
@JafarAbdi this issue with spinning was discussed [there](#1127). I'd like to make a change related to this but I don't know which strategy to choose.

JafarAbdi (approved):

henningkayser (approved):

PR #1163, opened 2022-04-02
Update panda controller filenames

AndyZe:
This PR should be merged along with these PR's in other repos:

moveit_resources:  https://github.com/ros-planning/moveit_resources/pull/127
moveit_tutorials:  https://github.com/ros-planning/moveit2_tutorials/pull/343#issuecomment-1086283003

@DLu is leading the charge on this, but I think the idea is to give the arm and the gripper a standard name that makes sense and can be used by Jafar's simplified launch module. Previously both MoveIt and ros_control config files were often named `controller.yaml`.


moveit_ros/moveit_servo/test/launch/test_servo_pose_tracking.test.py
```diff
@@ -34,7 +34,7 @@ def generate_servo_test_description(*args, gtest_name: SomeSubstitutionsType):
     ros2_controllers_path = os.path.join(
         get_package_share_directory("moveit_resources_panda_moveit_config"),
         "config",
-        "panda_ros2_controllers.yaml",
+        "ros2_controllers.yaml.yaml",
```
DLu, line 37:
```suggestion
        "ros2_controllers.yaml",
```

moveit_ros/planning_interface/test/launch/move_group_launch_test_common.py
```diff
@@ -55,7 +55,7 @@ def generate_move_group_test_description(*args, gtest_name: SomeSubstitutionsTyp
     ros2_controllers_path = os.path.join(
         get_package_share_directory("moveit_resources_panda_moveit_config"),
         "config",
-        "panda_ros2_controllers.yaml",
+        "ros2_controllers.yaml.yaml",
```
DLu, line 58:
```suggestion
        "ros2_controllers.yaml",
```

DLu (approved):

vatanaksoytezer:
Seems something wrong with rmw implementation in testing. Will investigate

```
    [unittest_planning_context-1] terminate called after throwing an instance of 'rclcpp::exceptions::RCLError'
    [unittest_planning_context-1]   what():  failed to initialize rcl init options: failed to load any RMW implementations, at ./src/functions.cpp:125, at ./src/rcl/init_options.c:75
    ```

vatanaksoytezer:
Until this PR get merged, everything will be blocked since we merged the moveit_resources PR, which is breaking. If we can't resolve the CI issue in testing, I will add a specific version of moveit_resources.

DLu:
I could not replicate this test failure on focal.

vatanaksoytezer:
Working on some fixes here: https://github.com/vatanaksoytezer/moveit2/pull/26

vatanaksoytezer:
Merged with #1170.

PR #1168, opened 2022-04-06
Revert parameter initialization part of #1138

vatanaksoytezer:
As spoken here https://github.com/ros-planning/moveit2/pull/1138#discussion_r839611238 this causes problems.


vatanaksoytezer:
Merged with #1170.

PR #1169, opened 2022-04-06
[Servo] override velocity scaling with a constant

MarqRazz:
### Description

This adds an option to MoveIt Servo to override the velocity scaling calculation with a constant value. The default behavior has not been changed, but if the user adds a parameter to their servo.yaml that includes `override_velocity_scaling_factor: 0.5` or any other number between 0 and 1 it will use that value for velocity scaling.

### DIFF
--- a/moveit2.repos
+++ b/moveit2.repos
@@ -11,4 +11,3 @@ repositories:
     type: git
     url: https://github.com/ros-planning/srdfdom.git
     version: ros2
-
--- a/moveit2.repos
+++ b/moveit2.repos
@@ -11,3 +11,4 @@ repositories:
     type: git
     url: https://github.com/ros-planning/srdfdom.git
     version: ros2
+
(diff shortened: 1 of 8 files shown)


AndyZe (commented):
The idea looks good to me but I haven't tested it.  :+1:

MarqRazz:
FYI I tested this on a robot in the office. If you don't have the value specified in your `servo.yaml` then the internal calculations for velocity limiting are used, but if you add the override then you can tune the speed as desired.

AndyZe:
Can you add the parameter to the yaml files (but comment it) so it's documented? Then I'll approve.

```
scale:
  linear:  0.4  # Max linear velocity. Unit is [m/s]. Only used for Cartesian commands.
  rotational:  0.8 # Max angular velocity. Unit is [rad/s]. Only used for Cartesian commands.
  # Max joint angular/linear velocity. Only used for joint commands on joint_command_in_topic.
  joint: 0.5
  
# Optionally override Servo's internal velocity scaling when near singularity or collision
# override_velocity_scaling_factor = 0.5
```

AndyZe (approved):

JafarAbdi (approved):

PR #1179, opened 2022-04-13
Delete an unused variable and a redundant log message

AndyZe:
### Description

- joint_model_group_ was completely unused, so delete it
- The log statement was redundant. If you check [here](https://github.com/ros-planning/moveit2/blob/a6ae014e06de20ce5190645f63779a4440054704/moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp#L63), all the return paths already have log messages that are more specific than this one


JafarAbdi (approved):
CI faliure isn't related to this PR

vatanaksoytezer (approved):

vatanaksoytezer:
We can merge as long as main is passing, can safely ignore the failure in testing for this PR.

PR #1184, opened 2022-04-15
Dynamic joint velocity/acceleration limits

### LINKED ISSUE #1180: Feedback wanted: how to update joint limits dynamically
I have a use case where the robot is picking a heavy payload, and it should move more slowly while carrying the payload. We are hoping to implement runtime configurable velocity & acceleration limits for this. The vel & accel (VA) limits are currently stored in the planning scene (`moveit::core::RobotModelConstPtr robot_model_`).  Note, it's `const`  :(

The RobotModel stores the PVAJ limits in a [JointBoundsVector](https://github.com/ros-planning/moveit2/blob/a6ae014e06de20ce5190645f63779a4440054704/moveit_core/robot_model/include/moveit/robot_model/robot_model.h#L574). Or you can [get the bounds](https://github.com/ros-planning/moveit2/blob/a6ae014e06de20ce5190645f63779a4440054704/moveit_core/robot_model/include/moveit/robot_model/robot_model.h#L432) for any individual joint.

I have about a week to implement this, not oodles and oodles of time.

** Where are VA limits used? **

Grepping through the codebase, it looks like these are the pieces that actually use VA limits. It's not so many places.

```
    Pilz joint_limits_aggregator
    time-parameterization algorithms in moveit_core/trajectory_processing
    RobotState isValidVelocityMove() --> Probably could delete. Not used internally
    Servo
```

** How would (Servo / MTC / time-parameterization plugins) work if the VA limits are updated? **

- MTC already has setRobotModel() and loadRobotModel() functions so I don't think it would have to change
- Time-parameterization plugins: won't need any change because they get a new copy of the RobotModel every time they're called
- Servo uses the RobotModel in the constructor. I could probably add a `reinitialize` service call that would stop motion then re-initialize a few things.
- Pilz: I think this would be similar to Servo that a few things need to be re-initialized if the RobotModel changes.

Here are some ideas I'm hoping to get feedback on.

** Proposal 1 - non-const RobotModel pointer **

Just make the RobotModel pointer in the Planning Scene non-const so it can be modified. Add a `setLimits()` method to the RobotModel. Add a MoveGroup service call to trigger this.

** Proposal 2 - Multiple planning groups **

Don't change the MoveIt codebase. Just use multiple planning groups, each with their own joint_limits.yaml

For us, that would still require a new `reinitialize` service in Servo.

AndyZe:
### Description

Update velocity/accel/jerk limits dynamically, like this:

```
  std::vector<double> new_velocity_bounds{ 10, 10, 10, 10, 10, 10, 10 };
  rmodel.updateVelocityBounds(new_velocity_bounds);
```
Fixes #1180 

### Testing

Requires moveit_resources.

`ros2 launch moveit_resources_panda_moveit_config demo launch.py`

### DIFF
--- a/moveit_core/robot_model/include/moveit/robot_model/joint_model.h
+++ b/moveit_core/robot_model/include/moveit/robot_model/joint_model.h
@@ -73,17 +73,17 @@ struct VariableBounds
   double max_position_;
   bool position_bounded_;
 
-  double min_velocity_;
-  double max_velocity_;
-  bool velocity_bounded_;
+  mutable double min_velocity_;
+  mutable double max_velocity_;
+  mutable bool velocity_bounded_;
 
-  double min_acceleration_;
-  double max_acceleration_;
-  bool acceleration_bounded_;
+  mutable double min_acceleration_;
+  mutable double max_acceleration_;
+  mutable bool acceleration_bounded_;
 
-  double min_jerk_;
-  double max_jerk_;
-  bool jerk_bounded_;
+  mutable double min_jerk_;
+  mutable double max_jerk_;
+  mutable bool jerk_bounded_;
 };
 
 class LinkModel;
--- a/moveit_core/robot_model/include/moveit/robot_model/robot_model.h
+++ b/moveit_core/robot_model/include/moveit/robot_model/robot_model.h
@@ -456,8 +456,17 @@ public:
     return joint_model_vector_[common_joint_roots_[a->getJointIndex() * joint_model_vector_.size() + b->getJointIndex()]];
   }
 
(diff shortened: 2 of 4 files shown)


rhaschke (commented):
As I pointed out in https://github.com/ros-planning/moveit2/issues/1180#issuecomment-1103899372, the limits in RobotModel are considered maximal limits. If you want smaller values for a particular motion segment, you can already do so, specifying scaling factors for the time-parameterization plugins, can't you?
I vote against making RobotModel members mutable. If scaling factors are not sufficient for you, I suggest extending the API of the time-parameterization plugins to accept limit maps as well.

PR #1191, opened 2022-04-21
Remove Iterative Spline and Iterative Parabola time-param algorithms

AndyZe:
We've been using TOTG by default for awhile now. The only drawback of TOTG I'm aware of is that it can't handle nonzero initial or final velocities/accelerations. So I was thinking that would be a reason to keep the older algorithms (ISTP and IPTP). But, I added a little test and it looks like neither of those algorithms handles it either. Even worse -- they just fail without throwing any error. So I think we should delete them.

3 commits:

- The first commit adds a simple test for IPTP. The test passes but the output is wrong. (It didn't do anything.)

```
20: [ RUN      ] TestTimeParameterization.NonzeroInitialVelAccel
20: Trajectory has 2 points over 0.000 seconds
20:   waypoint   0 time 0.000 pos  0.000 vel  0.000 acc  0.000
20:   waypoint   1 time 0.000 pos  0.000 vel  0.000 acc  0.000
20: [       OK ] TestTimeParameterization.NonzeroInitialVelAccel (0 ms)
```

- The second commit adds a simple test for ISTP. Again, the test passes but the output is wrong. The state at the final waypoint should be (position, vel, accel) == (0.01, 0.1, 0.2).

```
20: [ RUN      ] TestTimeParameterization.NonzeroInitialVelAccel
20: Trajectory has 4 points over 0.343 seconds
20:   waypoint   0 time 0.000 pos  0.000 vel  0.000 acc  0.742
20:   waypoint   1 time 0.118 pos  0.003 vel  0.021 acc -0.394
20:   waypoint   2 time 0.253 pos  0.005 vel  0.029 acc  0.526
20:   waypoint   3 time 0.343 pos  0.010 vel  0.100 acc  1.036
20: [       OK ] TestTimeParameterization.NonzeroInitialVelAccel (0 ms)
```

- The third commit deletes both of the plugins.


v4hn:
What do we gain from removing these plugins?

Changing defaults (if that didn't happen yet) is a different matter.

AndyZe:
> What do we gain

- Nothing unexpected happens if somebody tries to execute one of these plugins on a real robot
- The API changes we've been talking about don't need to be added to these older plugins --> less maintenance
  - https://github.com/ros-planning/moveit2/issues/1180#issuecomment-1103899372
  - https://github.com/ros-planning/moveit2/pull/1186
- Slightly faster CI run times and build times for MoveIt users

v4hn:
I just noticed by feedback by mail never made it here! :angry: Here's a copy:

Thank you for answering the question. :+1: I know I'm being annoying by
opposing various remove-that-code initiatives lately.

> - Nothing unexpected happens if somebody tries to execute one of these
plugins on a real robot

If someone explicitly uses a non-default they should be looking for different
behavior explicitly.
Please note that TOTG is also (by far) not the perfect dynamics
parameterization you make it sound like.
We had some long discussions on that [here](https://github.com/ros-planning/moveit/issues/2183) and [here](https://github.com/ros-planning/moveit/issues/2110).

> The API changes we've been talking about don't need to be added to these
older plugins --> less maintenance
> https://github.com/ros-planning/moveit2/issues/1180#issuecomment-1103899372 https://github.com/ros-planning/moveit2/pull/1186

True, though I believe it would be perfectly justified to add new arguments to
the callback and ignore them or warn if they are set.
And that takes 5 minutes to implement and test. :)

> - Slightly faster CI run times and build times for MoveIt users

I'm not sure this is relevant. From
https://github.com/ros-planning/moveit/issues/2081 it looks like the
longest-compiling method in this code part took 300ms to compile. So you might
gain a second.

v4hn:
Thank you for answering the question. :+1: I know I'm being annoying by opposing various remove-that-code initiatives lately.

> - Nothing unexpected happens if somebody tries to execute one of these plugins on a real robot

If someone explicitly uses a non-default they should be looking for different behavior explicitly.
Please note that TOTG is also (by far) not the perfect dynamics parameterization you make it sound like.
We had some long discussions on that [here](https://github.com/ros-planning/moveit/issues/2183) and [here](https://github.com/ros-planning/moveit/issues/2110).

> - The API changes we've been talking about don't need to be added to these older plugins --> less maintenance
>   - https://github.com/ros-planning/moveit2/issues/1180#issuecomment-1103899372
>   - https://github.com/ros-planning/moveit2/pull/1186

True, though I believe it would be perfectly justified to add new arguments to the callback and ignore them or warn if they are set.
And that takes 5 minutes to implement and test. :)

> - Slightly faster CI run times and build times for MoveIt users

I'm not sure this is relevant. From https://github.com/ros-planning/moveit/issues/2081 it looks like the longest-compiling method in this code part took 300ms to compile. So you might gain a second.

PR #1195, opened 2022-04-25
Allow custom velocity/acceleration limits for TOTG time-parameterization algorithm

AndyZe:
### Description

Add an overloaded version of `computeTimeStamps()` which will allow individual joints to move faster or slower than the default.

```
  bool computeTimeStamps(robot_trajectory::RobotTrajectory& trajectory,
                         const std::map<std::string, double>& velocity_limits,
                         const std::map<std::string, double>& acceleration_limits) const;
```

This work is sponsored by RE2 Robotics.


moveit_core/robot_model/include/moveit/robot_model/robot_model.h
```diff
@@ -428,7 +428,7 @@ class RobotModel
     return variable_names_;
   }
 
-  /** \brief Get the bounds for a specific variable. Throw an exception of variable is not found. */
```
AndyZe, line 431: I happened to dig through the code and did not see a place where an exception is thrown

moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
```diff
@@ -169,7 +169,15 @@ class TimeOptimalTrajectoryGeneration : public TimeParameterization
   bool computeTimeStamps(robot_trajectory::RobotTrajectory& trajectory, const double max_velocity_scaling_factor = 1.0,
                          const double max_acceleration_scaling_factor = 1.0) const override;
 
+  bool computeTimeStamps(robot_trajectory::RobotTrajectory& trajectory,
+                         const std::map<std::string, double>& velocity_limits,
+                         const std::map<std::string, double>& acceleration_limits) const;
+
```
kylc, line 175: Is there a reason why the velocity and acceleration limits need to maintain their ordering, or is it worth converting these to `std::unordered_map`s?

moveit_core/robot_model/include/moveit/robot_model/robot_model.h
kylc: If I were reading the docs I'd be curious to know what does happen if the variable is not found. It can't return a nullptr due to the ref return type, so does it just return garbage or worse segfault?

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -969,6 +963,122 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
...
+    bool set_velocity_limit = false;
+    for (const auto& [joint_name, velocity_limit] : velocity_limits)
+    {
+      if (joint_name.compare(vars[j]))
+      {
+        max_velocity[j] = velocity_limit;
+        set_velocity_limit = true;
+        break;
+      }
+    }
```
kylc, line 1004: Could this iteration be replaced by a simple lookup like this, or am I missing something?
```
auto it = velocity_limits.find(vars[j]);
if (it != velocity_limits.end()) {
  max_velocity[j] = it->second;
  set_velocity_limit = true;
}
```

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -969,6 +963,122 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
...
+
+    if (bounds.velocity_bounded_ && !set_velocity_limit)
+    {
+      // Set the default velocity limit, from robot model
+      if (bounds.max_velocity_ < std::numeric_limits<double>::epsilon())
```
kylc, line 1009: Is this functionally different than `if (bounds.max_velocity_ <= 0.0)`?

moveit_core/robot_model/include/moveit/robot_model/robot_model.h
AndyZe: OK, I was wrong here. It does throw but you have to dig several layers deep to find it. Will revert this change.

https://github.com/ros-planning/moveit2/blob/002dcbf82d0b95a28ac17048c16b8f00679d0e66/moveit_core/robot_model/src/joint_model.cpp#L89

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
AndyZe: I copied this code from elsewhere but I think you're right, there's no difference.

moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
```diff
@@ -169,7 +169,15 @@ class TimeOptimalTrajectoryGeneration : public TimeParameterization
   bool computeTimeStamps(robot_trajectory::RobotTrajectory& trajectory, const double max_velocity_scaling_factor = 1.0,
                          const double max_acceleration_scaling_factor = 1.0) const override;
 
+  bool computeTimeStamps(robot_trajectory::RobotTrajectory& trajectory,
+                         const std::unordered_map<std::string, double>& velocity_limits,
+                         const std::unordered_map<std::string, double>& acceleration_limits) const;
+
 private:
+  bool doTimeParameterizationCalculations(robot_trajectory::RobotTrajectory& trajectory,
```
AndyZe, line 177: This is a common function used by both versions of `computeTimeStamps`.

moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
```diff
@@ -169,7 +169,15 @@ class TimeOptimalTrajectoryGeneration : public TimeParameterization
   bool computeTimeStamps(robot_trajectory::RobotTrajectory& trajectory, const double max_velocity_scaling_factor = 1.0,
                          const double max_acceleration_scaling_factor = 1.0) const override;
 
+  bool computeTimeStamps(robot_trajectory::RobotTrajectory& trajectory,
```
JafarAbdi, line 172: Would it make sense to add this overload to the base class?

JafarAbdi (approved):

moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
AndyZe: @JafarAbdi I went ahead and added it to the base class. Can you give the new commit a review?  Thanks

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -911,17 +911,11 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
                 max_acceleration_scaling_factor, acceleration_scaling_factor);
   }
 
-  // This lib does not actually work properly when angles wrap around, so we need to unwind the path first
-  trajectory.unwind();
```
v4hn, line 915: Did you actually test the code with continuous joints and a trajectory that wraps angles?
Perusing the commits I can't see any explanation related to removing this... :cat2:

AndyZe: It was just moved down to L1063  :+1:

PR #1199, opened 2022-04-28
[Fix Rolling CI] Update moveit_configs_utils dependencies

AndyZe:
### Description

Add `moveit_configs_utils` dependencies. Should fix CI failures like this:

```
build/moveit_kinematics/test_results/moveit_kinematics/test_launch_fanuc-kdl-singular.test.py.xunit.xml: 1 test, 1 error, 0 failures, 0 skipped
- moveit_kinematics test_launch_fanuc-kdl-singular.test.py.xunit.missing_result
  <<< error message
...
    ModuleNotFoundError: No module named 'moveit_configs_utils'
```

### DIFF
--- a/moveit_ros/moveit_servo/package.xml
+++ b/moveit_ros/moveit_servo/package.xml
@@ -26,8 +26,9 @@
   <depend>control_msgs</depend>
   <depend>control_toolbox</depend>
   <depend>geometry_msgs</depend>
-  <depend>moveit_msgs</depend>
+  <depend>moveit_configs_utils</depend>
   <depend>moveit_core</depend>
+  <depend>moveit_msgs</depend>
   <depend>moveit_ros_planning_interface</depend>
   <depend>pluginlib</depend>
   <depend>sensor_msgs</depend>
--- a/moveit_kinematics/CMakeLists.txt
+++ b/moveit_kinematics/CMakeLists.txt
@@ -6,17 +6,18 @@ find_package(moveit_common REQUIRED)
 moveit_package()
 
 find_package(ament_cmake REQUIRED)
-find_package(Eigen3 REQUIRED)
-find_package(orocos_kdl REQUIRED)
-find_package(tf2_kdl REQUIRED)
-find_package(kdl_parser REQUIRED)
-find_package(rclcpp REQUIRED)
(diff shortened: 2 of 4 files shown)


moveit_ros/moveit_servo/package.xml
```diff
@@ -26,8 +26,9 @@
   <depend>control_msgs</depend>
   <depend>control_toolbox</depend>
   <depend>geometry_msgs</depend>
-  <depend>moveit_msgs</depend>
+  <depend>moveit_configs_utils</depend>
```
JafarAbdi, line 29: This should be a test_depend

JafarAbdi (commented):
This seems to fix the main issue, but now we have another issue, not sure what is related to

moveit_ros/moveit_servo/package.xml
AndyZe: i don't think so, these files are used when running the tutorials too

JafarAbdi: Sorry, I meant for the moveit_kinematics not servo

AndyZe: OK, that already is <test_depend>. Marking this as resolved

AndyZe:
@henningkayser @vatanaksoytezer @tylerjw  Do you have any idea how to fix this CI issue?

```
  $ ( source /opt/ros/rolling/setup.bash && rosdep install -q --from-paths /home/runner/work/moveit2/moveit2/.work/upstream_ws/src --ignore-src -y | grep -E '(executing command)|(Setting up)' ; )
  '( source /opt/ros/rolling/setup.bash && rosdep install -q --from-paths /home/runner/work/moveit2/moveit2/.work/upstream_ws/src --ignore-src -y | grep -E '(executing command)|(Setting up)' ; )' returned with 1
'install_upstream_dependencies' returned with code '1' after 0 min 0 sec
```

AndyZe:
Ah, it looks like industrial_ci is having the same issue:  [https://github.com/ros-industrial/industrial_ci](https://github.com/ros-industrial/industrial_ci)

AndyZe:
Well, this seems like a pretty small change but CI is not having it. OK, will close the PR.

PR #1202, opened 2022-04-29
Clamp inputs to Ruckig. Use current waypoint as input for next iteration

AndyZe:
### Description

This PR does 2 things:

- TOTG usually runs before this algorithm. If the output of TOTG is just a _tiny_ bit above velocity/acceleration limits due to numerical error, Ruckig fails. So, this PR clamps the inputs into Ruckig.
- Use the current waypoint state as input for the next iteration. For our client, this resulted in velocity that tracked the desired velocity much more closely. (Although it seems like it should be very similar either way, to me.)

### Verification

Clone the MoveIt2 tutorials and add Ruckig to `ompl_planning.yaml`:

```
request_adapters: >-
  default_planner_request_adapters/AddRuckigTrajectorySmoothing
  default_planner_request_adapters/AddTimeOptimalParameterization
```

Then run one of the tutorials. It should work like normal.

`ros2 launch moveit2_tutorials demo.launch.py rviz_tutorial:=true`

This work is sponsored by RE2 Robotics.

### DIFF
--- a/moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
+++ b/moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
@@ -220,9 +220,10 @@ void RuckigSmoothing::initializeRuckigState(ruckig::InputParameter<0>& ruckig_in
     current_velocities_vector.at(i) = first_waypoint.getVariableVelocity(idx.at(i));
     current_accelerations_vector.at(i) = first_waypoint.getVariableAcceleration(idx.at(i));
     // Clamp velocities/accelerations in case they exceed the limit due to small numerical errors
-    std::clamp(current_velocities_vector.at(i), -ruckig_input.max_velocity.at(i), ruckig_input.max_velocity.at(i));
-    std::clamp(current_accelerations_vector.at(i), -ruckig_input.max_acceleration.at(i),
-               ruckig_input.max_acceleration.at(i));
+    current_velocities_vector.at(i) =
+        std::clamp(current_velocities_vector.at(i), -ruckig_input.max_velocity.at(i), ruckig_input.max_velocity.at(i));
+    current_accelerations_vector.at(i) = std::clamp(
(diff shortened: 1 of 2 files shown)


moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -138,6 +135,9 @@ bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajecto
     }
   }
 
+  // Initialize the smoother
+  initializeRuckigState(ruckig_input, ruckig_output, *trajectory.getFirstWayPointPtr(), group);
```
AndyZe, line 139: Moved initialization to after the limits are known so the input can be clamped

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -229,28 +233,34 @@ void RuckigSmoothing::initializeRuckigState(ruckig::InputParameter<0>& ruckig_in
...
+
+    // Clamp velocities/accelerations in case they exceed the limit due to small numerical errors
+    std::clamp(ruckig_input.current_velocity.at(joint), -ruckig_input.max_velocity.at(joint),
+               ruckig_input.max_velocity.at(joint));
+    std::clamp(ruckig_input.current_acceleration.at(joint), -ruckig_input.max_acceleration.at(joint),
+               ruckig_input.max_acceleration.at(joint));
+    std::clamp(ruckig_input.target_velocity.at(joint), -ruckig_input.max_velocity.at(joint),
+               ruckig_input.max_velocity.at(joint));
+    std::clamp(ruckig_input.target_acceleration.at(joint), -ruckig_input.max_acceleration.at(joint),
+               ruckig_input.max_acceleration.at(joint));
```
kylc, line 263: `std::clamp` takes its value parameter by const ref, so I don't think these statements will have any effect.

Issue #872, opened 2021-12-01
Never Use std::bind


tylerjw:
https://github.com/cpp-best-practices/cppbestpractices/blob/master/08-Considering_Performance.md#never-use-stdbind

peterdavidfagan:
Hi @tylerjw, 

I'm happy to also pick up this issue as part of my GSoC application so feel free to assign this one to me as well. 

FYI @henningkayser.

v4hn:
I implemented that for moveit last week over the course of some hours: https://github.com/ros-planning/moveit/pull/3106
it's basically clang-tidy + assigning proper variable names.
However I also used the time to simplify and improve some code throughout the project.
With a million expected merge conflicts I would propose you go through the same steps again instead of meeting, but still take my pr as a reference and put all non-trivial commits please. :+1:

peterdavidfagan:
Awesome, thanks @v4hn it's great to have your pull request as a reference. I will go through the same steps as you have outlined above. 

I don't have the privileges to assign reviewers but it would be great to get your review of my pull request given your familiarity with the changes.

henningkayser:
@peterdavidfagan thanks for taking this on. Did you end up working on this already? Let us know if you need any help with this.

peterdavidfagan:
Hi @henningkayser, I have started working on this issue. I am aiming to open a pull request by this Friday/Saturday.

henningkayser:
> Hi @henningkayser, I have started working on this issue. I am aiming to open a pull request by this Friday/Saturday.

Awesome, thanks!

peterdavidfagan:
I just wished to provide a quick update on this issue, I am about 70% there with the changes. 

I am taking my time to go through this task as it is also a way for me to read through the codebase. I hope to submit a pr for this shortly, you may find the current status of changes on my fork [here](https://github.com/peterdavidfagan/moveit2/tree/remove_standard_bind). I also have a lot of local changes that still need to be pushed for the [banish bind commit](https://github.com/ros-planning/moveit/pull/3106/commits/a2911c80c28958c1fce8fb52333d770248c4ec05).

PR #1204, opened 2022-04-30
Remove bind

peterdavidfagan:
Fixes https://github.com/ros-planning/moveit2/issues/872

This pull request removes std::bind from the MoveIt2 codebase and includes some minor rework of pointer usage. 

The changes in this pull request are based on [these changes](https://github.com/ros-planning/moveit/pull/3106) that were completed within MoveIt (thanks to @v4hn for this).

This is my first pull request on the MoveIt2 project, I hope I didn't overlook anything in the process. I am happy to address any feedback items to ensure this code can be merged into the main project.


peterdavidfagan:
Hi @henningkayser and @v4hn, 

It looks like I need a maintainer to approve the execution of GHA workflows since I am a first-time contributor. I would appreciate the approval of this workflow run.

I can look to run industrial_ci locally if usage of GHA minutes is an important constraint.

peterdavidfagan:
Thanks for running CI, it looks like one of the tests is failing (`test_ruckig_traj_smoothing`), I will take a look into this to understand why it has occurred.

AndyZe:
Hmm, that's a weird failure. It looks like the test should pass. The requirement is:

(actual_duration < (1.3 * ideal_duration), actual: 0.285312 vs 0.273325

So it's definitely within 30%. I'm confused why that fails

AndyZe:
I'll try running the test again.

AndyZe:
Also, note that the failing test is on `rolling-ci-testing`. That is the upcoming changes that haven't been released yet. The required tests are `colling-ci` and `Formatting` -- both of which pass. So I think you are A-OK  :+1:

henningkayser:
Thank you @peterdavidfagan! The changes look all good at first sight. I would have expected cherry-picked commits from the original PR to keep the original authors of this effort, but I guess we haven't really required that. I'll leave it to @v4hn, @rhaschke to decide if they are ok with the new commits.

peterdavidfagan:
Thanks @henningkayser I appreciate this and I agree it would have been more optimal to preserve the original authors through cherry-picking commits. I would be happy to resubmit this pull request in this format if it is preferred, I would in this case separate any of my changes that are distinct from the original commits.

rhaschke:
Keeping the original authors would have been great indeed. If you don't mind, @peterdavidfagan, please adapt. You can easily change authors via manual rebase and `git commit --amend --author="..."`.

peterdavidfagan:
> Keeping the original authors would have been great indeed. If you don't mind, @peterdavidfagan, please adapt. You can easily change authors via manual rebase and `git commit --amend --author="..."`.

Will do @rhaschke I have set aside time for this at the end of my current work session so I will perform the rebase and push the changes then.

peterdavidfagan:
The above force push resets the authors to those who authored the original commits on the MoveIt repository. For [changes](https://github.com/ros-planning/moveit2/pull/1204/commits/26b93c9fd602ee5e43c4f9ed2eb5a432cd14f1f5) that were made to the planning context manager I assigned @rhaschke as the author since he made the last edits. In this case the commits made to this file have been squashed hence the two sources in the commit description. 

@rhaschke would you prefer for this commit to not be squashed in order to preserve the original history? Happy to also make this change if needed.

rhaschke:
Thanks. Looks good to me.

moveit_core/planning_scene/test/test_multi_threaded.cpp
```diff
@@ -117,23 +117,24 @@ TEST_P(CollisionDetectorTests, Threaded)
...
+    state.setToRandomPositions();
+    state.update();
+    collisions.push_back(runCollisionDetection(0, 1, *planning_scene_, state));
+    states.push_back(std::move(state));
   }
 
   for (unsigned int i = 0; i < THREADS; ++i)
-    threads.push_back(new std::thread(
-        std::bind(&runCollisionDetectionAssert, i, TRIALS, planning_scene_.get(), states[i].get(), collisions[i])));
+    threads.emplace_back(std::thread([i, &scene = *planning_scene_, &state = states[i], expected = collisions[i]] {
```
henningkayser, line 129: I've noticed that there are several occasions where pointers are dereferenced like here. Is there a particular reason instead of just dereferencing them in the function call? Just curious @v4hn

henningkayser (approved):
Ok, diff looks all good to me! CI fails because of an unrelated issue (fixed [here](https://github.com/ros-planning/moveit2/pull/1207)).

moveit_core/planning_scene/test/test_multi_threaded.cpp
rhaschke: I thinks, it's just a matter of what is actually captured, e.g. the full `states` array or a single state (reference).

henningkayser: Yeah, with states it makes sense to me, with pointers like `planning_scene_` I'm not sure. At least I haven't seen this with local shared pointers afaik, this was my only concern at first.

v4hn: > Yeah, with states it makes sense to me, with pointers like planning_scene_ I'm not sure. At least I haven't seen this with local shared pointers afaik which was my first concern.

I did that for multiple reasons:
- Restrict scope as Robert explained above
- clang-tidy does it automatically :smiley: 
- Never share ownership with methods that don't need it. It's a rather poisonous pattern in the whole code-base to pass shared pointers for everything, even if no ownership is transferred. It's one of the various things @tylerjw complains about a lot :cat2: 
- Dereferencing pointers when they are captured instead of used makes sure you actually get a traceable error at point of definition and not just at point of use. That pattern can help avoid multiple hours of debugging when `std::function`s are copied around.

PR #1206, opened 2022-05-02
Fix CI permission errors

AndyZe:
### Description

Attempt to fix CI permission issues like this:

```
  Could not determine remotes: fatal: unsafe repository ('/home/runner/work/moveit2/moveit2/.work/upstream_ws/src/geometric_shapes' is owned by someone else)
  To add an exception for this directory, call:
  
  	git config --global --add safe.directory /home/runner/work/moveit2/moveit2/.work/upstream_ws/src/geometric_shapes
```

### DIFF
--- a/.docker/ci/Dockerfile
+++ b/.docker/ci/Dockerfile
@@ -29,6 +29,13 @@ RUN \
         ccache && \
     #
     # Fetch all dependencies from moveit2.repos
+    # Fetch required upstream sources for building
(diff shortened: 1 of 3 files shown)


tylerjw (approved):
It looks like this worked. I'm ok with this solution for now but eventually it would be nice to have the docker file iterate through the directories in the upstream_ws/src directory and apply this command because we might change what is in the upstream.repos file, and when we do we'll also need to update this.

PR #1207, opened 2022-05-04
Use orocos_kdl_vendor package

henningkayser:
### Description

[orocos_kdl doesn't exist anymore,](https://github.com/ros/rosdistro/pull/33028) there is a [vendor package](https://github.com/ros2/orocos_kdl_vendor) instead.

Related CI failure: https://github.com/ros-planning/moveit2/runs/6294353126?check_suite_focus=true

### DIFF
--- a/moveit_core/CMakeLists.txt
+++ b/moveit_core/CMakeLists.txt
@@ -218,7 +218,7 @@ install(
 )
 
 ament_export_targets(export_${PROJECT_NAME} HAS_LIBRARY_TARGET)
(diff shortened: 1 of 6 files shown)


AndyZe (commented):
If you grep:

`grep -r orocos_kdl`

There are a few more replacements to make

henningkayser:
> 

Only the package.xml entries need to be replaced. The package name itself is provided by the vendor package. I pretty much copied the changes from [here](https://github.com/ros2/geometry2/pull/473).

rhaschke (approved):
Seems to work.

PR #1210, opened 2022-05-05
Docker cleanup: single "safe" directory for vcs

AndyZe:
### Description

Attempt modifying the "safe directory" to reduce a few lines and include everything in one line.


henningkayser (approved):

AndyZe:
Replaced by #1215

PR #1215, opened 2022-05-06
CI: More generically fix permission errors

rhaschke:
Generalize #1206, disabling git security check for any folder.
We are running our self-generated docker container on GHA. Chances are low that this gets compromised.
This should also solve https://github.com/ros-planning/moveit2_tutorials/pull/355.

### DIFF
--- a/.docker/ci/Dockerfile
+++ b/.docker/ci/Dockerfile
@@ -28,13 +28,11 @@ RUN \
         clang clang-format-12 clang-tidy clang-tools \
         ccache && \
     #
-    # Fetch all dependencies from moveit2.repos
-    # As of 5/2/2022, permissions need to be set explicitly.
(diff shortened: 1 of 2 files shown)


rhaschke:
I think the failing unit test (`unittest_trajectory_generator_circ.test.py`) is unrelated. Is it possible that this test is flaky?
I have seen it fail in https://github.com/ros-planning/moveit2/pull/1208 as well, which is an unrelated changes too.

AndyZe (approved):

AndyZe:
@jschleicher got any idea about the failing test?

I'd be OK with disabling it. To get CI passing again seems more important.

tylerjw (approved):

JafarAbdi (approved):

Issue #1070, opened 2022-02-08
When an Attached Collision Object is removed, it should actually be removed


AndyZe:
From issue #1069 , I learned that specifying moveit_msgs::AttachedCollisionObject::REMOVE of an Attached Collision Object does not actually remove it. It just converts it to a non-attached collision object.

I think specifying REMOVE should actually make the object disappear.

Maybe an additional DETACH message field would be a good idea, if detachment is the goal.

v4hn:
This is the failed design in MoveIt where it was much too late to change it...
It's perfectly justified to finally clean this up in MoveIt2 and cleanly provide `ATTACH`/`DETACH`!

Changing the meaning of `REMOVE` in attached objects without indication is a severe issue though and users will wonder where their objects went after detach with their existing code. It might be a very good idea to change the command names/values altogether to make sure user code that relies on this behavior breaks.

Issue #1069, opened 2022-02-08
Attached Collision Object is not removed from the scene


AndyZe:
### Description

The attached Python script simply publishes an attached collision object on the `/attached_collision_object` topic. Then it removes the same object.

Drop the attached Python file into install/moveit2_tutorials/lib/moveit2_tutorials/. Make it executable with `chmod +x simple_collision_object_pub.py`. 

#### Steps to reproduce

Run the Move Group tutorial (https://moveit.picknik.ai/main/doc/examples/move_group_interface/move_group_interface_tutorial.html) then publish the attached collision object.

```
ros2 launch moveit2_tutorials move_group.launch.py

ros2 launch moveit2_tutorials move_group_interface_tutorial.launch.py

ros2 run moveit2_tutorials simple_collision_object_pub.py
```

Press Enter to publish the collision object.

Press Next in the RViz GUI to plan a motion. It fails because there is a collision object in the way. All good so far.

Now press Enter again to remove the attached collision object.

Press Next in the RViz GUI to plan another motion. It still fails. This is the bug.

### Your environment
ROS Rolling, built from source

[simple_collision_object_pub.py.txt](https://github.com/ros-planning/moveit2/files/8025552/simple_collision_object_pub.py.txt)

AndyZe:
This video shows the issue. Planning fails even after the ACO has been removed.

https://user-images.githubusercontent.com/11284393/153030252-3f4933ac-0c2b-4346-a0d0-849fe4ac6e13.mp4

AndyZe:
Looking at the code, I think I have a conceptual misunderstanding here -- specifying to REMOVE an ACO only detaches it. The object still exists in the world.

AndyZe:
Here's a Python script to remove then detach the ACO.
[detach_then_remove.py.txt](https://github.com/ros-planning/moveit2/files/8026129/detach_then_remove.py.txt)

v4hn:
To remove an attached object in one processing step (at least from the outside) you should actually send* a PlanningScene message with the robot state filled in to remove the attached object and the world object filled in to remove the (then detached) object from the world. This way you need to send only one message at least.

-- 
* apply scene changes via service, not via message topics! MoveIt2 should really discourage using these topics for one-step updates to the scene! They do not run synchronously and might even drop!

PR #1223, opened 2022-05-06
removing AttachedCollisionObjects actually removes them

jeoseo:
### Description

This is for: https://github.com/ros-planning/moveit2/issues/1070. While the original PR suggests a new DETACH message field, from reading the code and this issue (https://github.com/ros-planning/moveit/issues/1581), it seems like the MOVE message field is unused. I made it such that REMOVE removes the object completely, where MOVE moves the object from the list of attached objects to the list of world objects; this way moveit_msgs can stay unchanged. 

### Checklist

No formal tests, but running the steps to reproduce the original issue in https://github.com/ros-planning/moveit2/issues/1069 with this slightly modified simple_collision_object_pub.py should show that the object being attached, detached, attached, then removed, and planning should fail until the final removal happens.
```
#!/usr/bin/env python3

# Publish an attached object. Can be used for collision testing.
# Author: Andy Zelenak

from moveit_msgs.msg import AttachedCollisionObject
from moveit_msgs.msg import CollisionObject
import rclpy
from rclpy.node import Node
from shape_msgs.msg import SolidPrimitive


class AttachedCollisionObjectClient(Node):

    def __init__(self):
        super().__init__('attached_collision_object_client')
        self.publisher = self.create_publisher(AttachedCollisionObject,
                                               '/attached_collision_object', 1)
        self.object_msg = AttachedCollisionObject()
        self.object_msg.link_name = 'panda_link0'
        self.object_msg.object.header.frame_id = 'panda_link0'
        # self.object_msg.object.pose.position.x = 1.0
        self.object_msg.object.id = 'Box_0'
        self.is_diff = True
        box = SolidPrimitive()
        box.type = 1
        box.dimensions = [0.5, 0.5, 0.5]
        self.object_msg.object.primitives = [box]

    def publish_attach(self):
        self.object_msg.object.operation = CollisionObject.ADD
        self.object_msg.object.header.stamp = self.get_clock().now().to_msg()
        self.publisher.publish(self.object_msg)

    def publish_detach(self):
        self.object_msg.object.operation = CollisionObject.MOVE
        self.object_msg.object.header.stamp = self.get_clock().now().to_msg()
        self.publisher.publish(self.object_msg)

    def publish_remove(self):
        self.object_msg.object.operation = CollisionObject.REMOVE
        self.object_msg.object.header.stamp = self.get_clock().now().to_msg()
        self.publisher.publish(self.object_msg)


def main(args=None):
    rclpy.init(args=args)

    collision_publisher = AttachedCollisionObjectClient()

    input('Press Enter to publish a collision object')
    collision_publisher.publish_attach()

    input('Press Enter to detach the collision object')
    collision_publisher.publish_detach()

    input('Press Enter to attach the collision object, should not produce a duplicate')
    collision_publisher.publish_attach()

    input('Press Enter to remove the object')
    collision_publisher.publish_remove()


if __name__ == '__main__':
    main()
```


v4hn:
1. I just made a case in the mentioned issue for *explicitly breaking* message compatibility so that user code explicitly breaks instead of changing fundamental behavior.
2.
> it seems like the MOVE message field is unused

It is not **implemented**, but the semantics are reserved for in-hand manipulation or in-hand object pose estimation! Changing these semantics for the keyword is not acceptable.

AndyZe:
Well, I really appreciate your effort in digging into this issue anyway! It is important. I tend to agree with @v4hn that breaking the API explicitly is the way to go

tylerjw:
@AndyZe if you do not think this is the right approach to solving this problem would you mind creating an issue to describe what the desired behavior is?

tylerjw:
This work seems to have been abandoned, closing for now.

PR #1227, opened 2022-05-09
Declare the default_planning_pipeline parameter

stephanie-eng:
### Description

The motion planning Rviz plugin doesn't seem to respect the `default_planning_pipeline` parameter as discussed in https://github.com/ros-planning/moveit_resources/issues/134

The node doesn't seem to declare this parameter, so this PR adds a line to declare it.

### Testing
Test with:
`ros2 launch moveit_resources_panda_moveit_config demo.launch.py`
Add in the motion planning panel, navigate to the context tab, and check the planning library. This PR should change the default selection from CHOMP to OMPL.


AndyZe (commented):
I still see `chomp` come up as the default in the GUI, though. I guess `default_planning_pipeline` needs to be specidied in a launch file somewhere. Maybe add it to your [move_resources 135](https://github.com/ros-planning/moveit_resources/pull/135/files) PR?

AndyZe:
Please ignore my previous comment -- I wasn't sourced correctly when testing. This works  :+1:

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
```diff
@@ -612,13 +612,20 @@ void MotionPlanningFrame::initFromMoveGroupNS()
...
+
   // Set initial velocity and acceleration scaling factors from ROS parameters
   double factor;
   node_->get_parameter_or("robot_description_planning.default_velocity_scaling_factor", factor, 0.1);
   ui_->velocity_scaling_factor->setValue(factor);
   node_->get_parameter_or("robot_description_planning.default_acceleration_scaling_factor", factor, 0.1);
   ui_->acceleration_scaling_factor->setValue(factor);
 
+  // Query default planning pipeline id
+  node_->get_parameter(planning_display_->getMoveGroupNS() + "default_planning_pipeline", default_planning_pipeline_);
```
AndyZe, line 627: I still think this should get moved up to about L618 so it is together with the `default_planning_pipeline` parameter declaration. Is there a reason not to do that?

stephanie-eng: Ah, I misunderstood your previous comment - I thought you were asking to put it with the other get_parameter lines. Moved the declare and the get to be more close.

AndyZe (approved):

Issue #1239, opened 2022-05-11
Plan and Execute in MoveGroupInterface return 'false' instead of MoveItErrorCode


tbastiaens-riwo:
### Description
Plan, Execute and Move methods in MoveGroupInterface erroneously return 'false' instead of MoveItErrorCode 

### Your environment
* ROS Distro: Foxy
* OS Version: Ubuntu 20.04
* Source or Binary build? Source, foxy

### Steps to reproduce
- Call movegroupinterface.plan(Plan& plan), movegroupinterface.move(bool wait) or movegroupinterface.execute(const moveit_msgs::msg::RobotTrajectory& trajectory, bool wait) with a path that cannot succesfully be planned or executed, for whatever reason.

### Expected behaviour
Expected behaviour is to return a [MoveItErrorCode](https://github.com/ros-planning/moveit_msgs/blob/152f6f6ac1ec83e0028c806831a4fbca32008f5e/msg/MoveItErrorCodes.msg)

### Actual behaviour
The methods return 'false' if the error code is not 'rclcpp_action::ResultCode::SUCCEEDED'. This happens because of the following lines: [plan](https://github.com/ros-planning/moveit2/blob/4f83466a6a0de56ed8dcd670b46ecb7cdeb82f5c/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp#L789), [move](https://github.com/ros-planning/moveit2/blob/4f83466a6a0de56ed8dcd670b46ecb7cdeb82f5c/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp#L869), [execute](https://github.com/ros-planning/moveit2/blob/4f83466a6a0de56ed8dcd670b46ecb7cdeb82f5c/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp#L939). This results in the MoveItErrorCode being interpreted as a value of '0', which is undefined according to [MoveItErrorCode](https://github.com/ros-planning/moveit_msgs/blob/152f6f6ac1ec83e0028c806831a4fbca32008f5e/msg/MoveItErrorCodes.msg)

### Suggested fix
Remove the 3 lines mentioned above such that 'return false' does not happen and instead the MoveItErrorCode is correctly returned.

JafarAbdi:
Thanks @tbastiaens-riwo for the bug report, do you mind opening a PR that fixes this issue?

tbastiaens-riwo:
[...]

PR #1266, opened 2022-05-20
Changed 'return false' in plan, move and execute such that MoveItErrorCode is returned

tbastiaens-riwo:
### Description
Provided a fix for [issue  #1239](https://github.com/ros-planning/moveit2/issues/1239#issue-1232732923).

### DIFF
--- a/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
+++ b/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
@@ -794,7 +794,7 @@ public:
     if (code != rclcpp_action::ResultCode::SUCCEEDED)
     {
       RCLCPP_ERROR_STREAM(LOGGER, "MoveGroupInterface::plan() failed or timeout reached");
-      return false;
+      return res->error_code;
     }
 
     plan.trajectory_ = res->planned_trajectory;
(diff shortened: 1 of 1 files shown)


tylerjw (approved):

henningkayser (approved):

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
```diff
@@ -806,7 +806,7 @@ class MoveGroupInterface::MoveGroupInterfaceImpl
     if (code != rclcpp_action::ResultCode::SUCCEEDED)
     {
       RCLCPP_ERROR_STREAM(LOGGER, "MoveGroupInterface::plan() failed or timeout reached");
-      return false;
+      return res->error_code;
```
AndyZe, line 809: Really minor nitpick: why not use `code` here since somebody took the trouble to define it?

```suggestion
      return code;
```

AndyZe (approved):

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
AndyZe: Ah, I see why now. Will revert my change  :+1:

AndyZe:
Yeah, I screwed it up. @tbastiaens-riwo can you revert the last commit with `git reset --hard HEAD~`? Then force-push it. Sorry.

henningkayser (approved):

vatanaksoytezer (approved):

PR #1274, opened 2022-05-25
Fix a bug when checking a pose is empty and TOTG corner case

JafarAbdi:
### Description

Fix: https://github.com/ros-planning/moveit_task_constructor/issues/350

The first commit fix `isEmpty` for ROS2 since `w == 1` 
The second commit fix a corner case for TOTG where it might not use the last input waypoint when parameterizing the trajectory

To test run:

```
ros2 launch moveit_task_constructor_demo demo.launch.py
ros2 launch moveit_task_constructor_demo pickplace.launch.py
```

### DIFF
--- a/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
+++ b/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
@@ -1096,6 +1096,10 @@ bool TimeOptimalTrajectoryGeneration::doTimeParameterizationCalculations(robot_t
 
     if (diverse_point)
       points.push_back(new_point);
+    // If the last point is not a diverse_point we replace the last added point with it to make sure to always have the
+    // input end point as the last point
+    else if (p == num_points - 1)
+      points.back() = new_point;
   }
 
   // Return trajectory with only the first waypoint if there are not multiple diverse points
--- a/moveit_core/utils/src/message_checks.cpp
+++ b/moveit_core/utils/src/message_checks.cpp
@@ -71,7 +71,7 @@ bool isEmpty(const moveit_msgs::msg::Constraints& constr)
 
 bool isEmpty(const geometry_msgs::msg::Quaternion& msg)
 {
-  return msg.x == 0.0 && msg.y == 0.0 && msg.z == 0.0 && msg.w == 0.0;
+  return msg.x == 0.0 && msg.y == 0.0 && msg.z == 0.0 && msg.w == 1.0;
 }
 
 bool isEmpty(const geometry_msgs::msg::Pose& msg)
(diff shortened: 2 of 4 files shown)


moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -1096,6 +1096,10 @@ bool TimeOptimalTrajectoryGeneration::doTimeParameterizationCalculations(robot_t
 
     if (diverse_point)
       points.push_back(new_point);
+    // If the last point is not a diverse_point we replace the last added point with it to make sure to always have the
```
jliukkonen, line 1099: I don't have anything against the fix. But why is the variable called `diverse_point` when it seems to mean, based on the comments, simply "unique waypoint", which would be easier to understand?

moveit_core/utils/src/message_checks.cpp
```diff
@@ -71,7 +71,7 @@ bool isEmpty(const moveit_msgs::msg::Constraints& constr)
 
 bool isEmpty(const geometry_msgs::msg::Quaternion& msg)
 {
-  return msg.x == 0.0 && msg.y == 0.0 && msg.z == 0.0 && msg.w == 0.0;
+  return msg.x == 0.0 && msg.y == 0.0 && msg.z == 0.0 && msg.w == 1.0;
```
AndyZe, line 74: I think this should cover both cases. Whether `w` is initialized or not initialized.

```suggestion
  return (msg.x == 0.0 && msg.y == 0.0 && msg.z == 0.0 && msg.w == 1.0) || (msg.x == 0.0 && msg.y == 0.0 && msg.z == 0.0 && msg.w == 0.0);
```

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
JafarAbdi: > I don't really understand the fix. If it's not a diverse_point, that means it's identical to the previous point, right?

It's identical with some tolerance `min_angle_change_`, it was causing [MTC to fail](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/src/stages/connect.cpp#L127-L132) because the last waypoint after parameterizing the trajectory isn't the expected goal waypoint

moveit_core/utils/src/message_checks.cpp
JafarAbdi: I don't think we should handle the other case since it no longer exist in ROS2

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
AndyZe: OK, that makes sense

AndyZe (approved):

PR #1275, opened 2022-05-25
[MSA] Workaround to launch files without controllers

vatanaksoytezer:
- Enable compilation of controllers
- Add user warning for hand generating controllers
- Issue a warning instead of raising an exception if the controllers haven't been found

### DIFF
--- a/moveit_configs_utils/moveit_configs_utils/moveit_configs_builder.py
+++ b/moveit_configs_utils/moveit_configs_utils/moveit_configs_builder.py
@@ -415,6 +415,7 @@ class MoveItConfigsBuilder(ParameterBuilder):
         else:
             file_path = self._package_path / file_path
 
+        if file_path:
             self.__moveit_configs.trajectory_execution.update(load_yaml(file_path))
         return self
 
--- a/moveit_configs_utils/moveit_configs_utils/moveit_configs_builder.py
+++ b/moveit_configs_utils/moveit_configs_utils/moveit_configs_builder.py
@@ -387,9 +387,10 @@ class MoveItConfigsBuilder(ParameterBuilder):
             controller_pattern = re.compile("^(.*)_controllers.yaml$")
             possible_names = get_pattern_matches(config_folder, controller_pattern)
             if not possible_names:
-                raise RuntimeError(
-                    "trajectory_execution: `Parameter file_path is undefined "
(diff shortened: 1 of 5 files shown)


henningkayser (approved):

moveit_configs_utils/moveit_configs_utils/moveit_configs_builder.py
```diff
@@ -414,7 +415,7 @@ def trajectory_execution(
         else:
             file_path = self._package_path / file_path
 
-        self.__moveit_configs.trajectory_execution.update(load_yaml(file_path))
+            self.__moveit_configs.trajectory_execution.update(load_yaml(file_path))
```
JafarAbdi, line 418:
```suggestion
        if file_path:
            self.__moveit_configs.trajectory_execution.update(load_yaml(file_path))
```

JafarAbdi (commented):
This will not use the file_path set from the other if statement

moveit_configs_utils/moveit_configs_utils/moveit_configs_builder.py
```diff
@@ -387,9 +387,10 @@ def trajectory_execution(
             controller_pattern = re.compile("^(.*)_controllers.yaml$")
             possible_names = get_pattern_matches(config_folder, controller_pattern)
             if not possible_names:
-                raise RuntimeError(
-                    "trajectory_execution: `Parameter file_path is undefined "
-                    f"and no matches for {config_folder}/*_controllers.yaml"
+                # Warn the user instead of raising exception
+                logging.warning(
+                    "\x1b[33;20mtrajectory_execution: `Parameter file_path is undefined "
+                    f"and no matches for {config_folder}/*_controllers.yaml\x1b[0m"
```
AndyZe, line 393:
```suggestion
                    "and no matches for {config_folder}/*_controllers.yaml\x1b[0m"
```

JafarAbdi: I'm pretty sure this's needed, why removing it? Printing the value of `config_folder` should make the warning more helpful.

AndyZe: is the typo `f` needed?  (Unless that's some crazy Python syntax I'm not aware of)

AndyZe (approved):

PR #1273, opened 2022-05-24
Port OMPL orientation constraints to MoveIt2

stephanie-eng:
### Description

Port of the OMPL orientation constraints. Update to the tutorial also incoming.

PR #1300, opened 2022-06-01
Increase priority for constrained planning state space

stephanie-eng:
### Description

Depends on #1273

Increases the priority for using the constrained planning state space when there is exactly one position or orientation constraint. The current behaviour sets the priority to -2 always, making it so constrained planning will never be used unless `enforce_constrained_state_space` is `True`.

### Checklist


tylerjw:
@stephanie-eng is this still a draft? Would you like a review of this? If so, who do you think would best be able to review this change?

stephanie-eng:
It's still a draft as I need to rework some tests, and haven't gotten around to it. I'll take another look at this and try to finish it off soon, and tag you for review when ready :smile:

stephanie-eng:
Sorry about the premature ping, Tyler - I fixed the failing test, but it looks like another one is failing and I'll need to look at that.

sjahr (approved):

PR #1312, opened 2022-06-02
Use CallbackGroup in PlanningSceneMonitor

henningkayser:
Similar to #1305, this replaces the private node with a CallbackGroup. There are still some issues with the synced parameters that I need to look into before this is really ready. Also, the CallbackGroup should be applied to the CSM as well.


tylerjw (approved):

henningkayser:
I switched this back to DRAFT because there are still runtime issues with parameter updates I need to look into first.

tylerjw:
@henningkayser do you have some documentation you can link to I can read to better understand what the issues are here?

Is there some test I can run to see the issues you are experiencing with this change?

tylerjw:
There is a small conflict but it seems like we abandoned this work. I'll create a issue to track it.

PR #1265, opened 2022-05-19
publish robot description config

tylerjw:
- Add options to config for publishing description
- Fix clang-format-12 errors

This fix adds support to moveit_config_utils for the two parameters that determine if the robot description is published on a latched topic. This is useful for reducing the amount of boilerplate in launch files and in some cases making so some launch files don't need to be written at all.

This change is necesary for keeping the tutorials I am writing simple.

I also included two changes from running clang-format-12 over moveit in this. I've tested this change on 22.04 on Rolling.

### Description

Please explain the changes you made, including a reference to the related issue if applicable

### Checklist
- [x] **Required by CI**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)
- [x] Extend the tutorials / documentation [reference](http://moveit.ros.org/documentation/contributing/)

PR #1315, opened 2022-06-03
Fix moveit_configs_utils parameter ordering

DLu:
Because https://github.com/ros-planning/moveit2/pull/1265 added a value for `publish_robot_description_semantic`, the value specified in the [move_group launch](https://github.com/ros-planning/moveit2/blob/7427a8ead23955f69bf9e530bbd06e93f8dcfae0/moveit_configs_utils/moveit_configs_utils/launches.py#L205) is overwritten. 

Either we should fix this by changing the order the parameters are loaded (as in this PR) OR the default value of `publish_robot_description_semantic` should be true. 

### Checklist

### DIFF
--- a/moveit_configs_utils/moveit_configs_utils/launches.py
+++ b/moveit_configs_utils/moveit_configs_utils/launches.py
@@ -219,8 +219,8 @@ def generate_move_group_launch(moveit_config):
     }
 
     move_group_params = [
-        move_group_configuration,
         moveit_config.to_dict(),
+        move_group_configuration,
     ]
 
     add_debuggable_node(


moveit_configs_utils/moveit_configs_utils/launches.py
```diff
@@ -219,8 +219,8 @@ def generate_move_group_launch(moveit_config):
     }
 
     move_group_params = [
-        move_group_configuration,
         moveit_config.to_dict(),
+        move_group_configuration,
```
JafarAbdi, line 223: Why is the reordering necessary? Is it because we have the parameter in both variables and the second one will be used?

DLu: I believe so.

JafarAbdi: Should we remove https://github.com/ros-planning/moveit2/blob/7427a8ead23955f69bf9e530bbd06e93f8dcfae0/moveit_configs_utils/moveit_configs_utils/launches.py#L205 then?

DLu: No, I want that parameter to be set to true.

AndyZe (approved):

PR #1327, opened 2022-06-07
Remove repeated checking of same value

rcywongaa:
### Description

While the repeated checking of the same value seems likely wrong, I wonder if this function is doing what it claims.  Seems to be checking whether a scalar is within bounds of a vector[3].

### Checklist

### DIFF
--- a/moveit_core/robot_model/src/planar_joint_model.cpp
+++ b/moveit_core/robot_model/src/planar_joint_model.cpp
@@ -270,10 +270,10 @@ double PlanarJointModel::distance(const double* values1, const double* values2) 
 
 bool PlanarJointModel::satisfiesPositionBounds(const double* values, const Bounds& bounds, double margin) const
 {
-  for (unsigned int i = 0; i < 3; ++i)
-    if (values[0] < bounds[0].min_position_ - margin || values[0] > bounds[0].max_position_ + margin)
-      return false;
-  return true;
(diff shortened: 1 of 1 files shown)


moveit_core/robot_model/src/planar_joint_model.cpp
```diff
@@ -270,10 +270,10 @@ double PlanarJointModel::distance(const double* values1, const double* values2)
...
 bool PlanarJointModel::satisfiesPositionBounds(const double* values, const Bounds& bounds, double margin) const
 {
-  for (unsigned int i = 0; i < 3; ++i)
-    if (values[0] < bounds[0].min_position_ - margin || values[0] > bounds[0].max_position_ + margin)
-      return false;
-  return true;
+  if (values[0] < bounds[0].min_position_ - margin || values[0] > bounds[0].max_position_ + margin)
+    return false;
+  else
+    return true;
```
tylerjw, line 276: Is there any guarantee that` values` and `bounds` have at least one value? This looks like a segfault in waiting.

tylerjw: Does it also seem like it is more likely that the intent was to use the index `i` somehow than to check the first element 3 times?

rcywongaa:
I'll close this for now until I clarify what this function is supposed to do...

PR #1329, opened 2022-06-09
Enable moveit2 kinematics framework within ros2 control loop

pac48:
### Description

Previously, the RDFLoader and RobotModelLoader classes required a rclcpp::Node node in order to load a robot from the ROS network. Since the ros2_control controllers now run within a rclcpp_lifecycle/LifecycleNode node, there is no way to use Moveit2's kinematic framework.

However, the rclcpp::Node and LifecycleNode nodes do provide getter methods for common interfaces, e.g. get_node_base_interface(), get_node_parameters_interface(), and get_node_topics_interface(). My approach to solve the issue was to refactor the RobotModelLoader and RDFLoader classes to use the provided common interfaces, rather than a rclcpp::Node. Since these classes pass a rclcpp::Node reference to KinematicsPluginLoader, which depends on KinematicsBase, I has to modify both of those classes as well.       

### Checklist


moveit_ros/planning/rdf_loader/src/rdf_loader.cpp
```diff
@@ -56,27 +56,47 @@ namespace rdf_loader
...
-RDFLoader::RDFLoader(const std::shared_ptr<rclcpp::Node>& node, const std::string& ros_name,
-                     bool default_continuous_value, double default_timeout)
-  : ros_name_(ros_name)
+void RDFLoader::loadRobot(const NodeInterfaceSharedPtr& node_interface,
+                          const TopicsInterfaceSharedPtr& topics_interface,
+                          const ParametersInterfaceSharedPtr& parameters_interface, const std::string& ros_name,
+                          bool default_continuous_value, double default_timeout)
 {
-  auto start = node->now();
+  rclcpp::Clock clock;
```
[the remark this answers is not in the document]
pac48: This clock is only used to determine the amount of time it takes to load robot description. A time difference is not affected by the clock offset. I can change it is you still think its important.

moveit_setup_assistant/src/tools/compute_default_collisions.cpp
```diff
@@ -52,8 +52,8 @@ const boost::unordered_map<DisabledReason, std::string> REASONS_TO_STRING = boos
     DEFAULT, "Default")(ADJACENT, "Adjacent")(ALWAYS, "Always")(USER, "User")(NOT_DISABLED, "Not Disabled");
 
 const boost::unordered_map<std::string, DisabledReason> REASONS_FROM_STRING =
-    boost::assign::map_list_of("Never", NEVER)("Default", DEFAULT)(
-        "Adjacent", ADJACENT)("Always", ALWAYS)("User", USER)("Not Disabled", NOT_DISABLED);
+    boost::assign::map_list_of("Never", NEVER)("Default", DEFAULT)("Adjacent", ADJACENT)("Always", ALWAYS)(
```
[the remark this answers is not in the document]
pac48: Correct, clang changed this.

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
```diff
@@ -651,7 +651,7 @@ class LockedPlanningSceneRO
     return planning_scene_monitor_ && planning_scene_monitor_->getPlanningScene();
   }
 
-  operator const planning_scene::PlanningSceneConstPtr&() const
+  operator const planning_scene::PlanningSceneConstPtr &() const
```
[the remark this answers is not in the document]
pac48: Correct, clang changed this.

moveit_ros/planning/rdf_loader/include/moveit/rdf_loader/rdf_loader.h
```diff
@@ -41,6 +41,7 @@
 #include <urdf/model.h>
 #include <srdfdom/model.h>
 #include <rclcpp/rclcpp.hpp>
+//#include <rclcpp/node_interfaces/node_parameters_interface.hpp>
```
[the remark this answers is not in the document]
pac48: This should be removed. Good catch.

moveit_core/package.xml
```diff
@@ -2,7 +2,7 @@
 <?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="3">
   <name>moveit_core</name>
-  <version>2.5.1</version>
```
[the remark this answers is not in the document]
pac48: Done

moveit_core/kinematic_constraints/src/kinematic_constraint.cpp
```diff
@@ -429,7 +429,7 @@ bool PositionConstraint::configure(const moveit_msgs::msg::PositionConstraint& p
 
   if (pc.weight <= std::numeric_limits<double>::epsilon())
   {
-    RCLCPP_WARN(LOGGER, "The weight on position constraint for link '%s' is near zero.  Setting to 1.0.",
+    RCLCPP_WARN(LOGGER, "The weight on orientation constraint for link '%s' is near zero.  Setting to 1.0.",
```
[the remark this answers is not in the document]
pac48: I am not sure how that got changed. I will change it back to "position".

moveit_ros/planning/kinematics_plugin_loader/include/moveit/kinematics_plugin_loader/kinematics_plugin_loader.h
```diff
@@ -57,7 +60,23 @@ class KinematicsPluginLoader
...
-    : node_(node), robot_description_(robot_description), default_search_resolution_(default_search_resolution)
+    : node_interface_(node->get_node_base_interface())
+    , parameters_interface_(node->get_node_parameters_interface())
+    , topics_interface_(node->get_node_topics_interface())
+    , robot_description_(robot_description)
+    , default_search_resolution_(default_search_resolution)
+  {
+  }
+
+  KinematicsPluginLoader(const NodeInterfaceSharedPtr& node_interface, const TopicsInterfaceSharedPtr& topics_interface,
```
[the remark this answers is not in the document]
pac48: Where should that struct be located? Since these changes are across multiple packages, would it make sense to create a utility package in /moveit_ros that define a wrapper struct?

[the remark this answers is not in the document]
henningkayser: That's the question... We've discussed abstracting from the RCLCPP interfaces in the past, and came up with some DI approach for increased testability. https://github.com/ros-planning/moveit2/pull/562 and https://github.com/ros-planning/moveit2/pull/569 are examples for classes that provide ROS features without direct API usage. However, these interfaces were very specific to their classes. It would be great, if we could come up with a more generic approach for the node interfaces, which technically could be used all over the codebase. I'm not saying that this needs to be done with this PR, but if we could reduce it to one type, refactoring would be much easier in the future.

> Since these changes are across multiple packages, would it make sense to create a utility package in /moveit_ros that define a wrapper struct?

Yes, that would make sense to me in this case.

Tagging @tylerjw @griswaldbrooks for their opinion.

moveit_core/kinematics_base/src/kinematics_base.cpp
```diff
@@ -92,6 +92,17 @@ bool KinematicsBase::initialize(const rclcpp::Node::SharedPtr& /*node*/,
   return false;
 }
 
+bool KinematicsBase::initialize(const moveit::core::RobotModel& /*robot_model*/, const std::string& group_name,
+                                const std::string& /*base_frame*/, const std::vector<std::string>& /*tip_frames*/,
+                                double /*search_discretization*/)
+{
+  RCLCPP_ERROR(LOGGER,
+               "IK plugin for group '%s' relies on deprecated API. "
```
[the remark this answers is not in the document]
pac48: For some reason kinematics_plugin_loader.cpp is calling this deprecated code in on of its methods. I don't see why at the moment. The reason I added the method is because it the existing KinematicsBase::initialize needed a node as an input. Now that you point it out, I can just remove the node input instead of overloading it.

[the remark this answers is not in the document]
pac48: Actually I take that back. KinematicsBase::initialize is a virtual method. I'm guessing it needs a default implementation because it is a plugin. The error message will occur if the user forgets to define the initialize message. So its is saying that their code  is deprecated, not ours.

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -69,19 +69,17 @@ class KinematicsPluginLoader::KinematicsLoaderImpl
    * \param search_res
    * \param iksolver_to_tip_links - a map between each ik solver and a vector of custom-specified tip link(s)
    */
-  KinematicsLoaderImpl(const NodeInterfaceSharedPtr& node_interface, const TopicsInterfaceSharedPtr& topics_interface,
-                       const ParametersInterfaceSharedPtr& parameters_interface, const std::string& robot_description,
+  KinematicsLoaderImpl(const node_interface::NodeInterfaceSharedPtr& node_interface,
```
tylerjw, line 72: Tiny bit of feedback is that if you are going to make a copy of a shared_ptr (to increment the reference count, which you are doing here) it is generally good to make it obvious to your caller that you are doing that through your function signature. The way to do that is to not take a reference but instead take it by value, and then move that value into your local state. Here is a minimal example of what I mean:

```Foo(std::shared_ptr<Node> node) : node_{std::move(node)} {}```

File: moveit_ros/planning/rdf_loader/include/moveit/rdf_loader/synchronized_string_parameter.h [hunk 10]
```diff
@@ -72,12 +78,14 @@ class SynchronizedStringParameter
...
-  std::shared_ptr<rclcpp::Node> node_;
+  NodeInterfaceSharedPtr node_interface_ = nullptr;
+  TopicsInterfaceSharedPtr topics_interface_ = nullptr;
+  ParametersInterfaceSharedPtr parameters_interface_ = nullptr;
   std::string name_;
   StringCallback parent_callback_;
 
   rclcpp::Subscription<std_msgs::msg::String>::SharedPtr string_subscriber_;
-  rclcpp::Publisher<std_msgs::msg::String>::SharedPtr string_publisher_;
+  std::shared_ptr<rclcpp::Publisher<std_msgs::msg::String>> string_publisher_;
```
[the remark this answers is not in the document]
henningkayser: I somewhat agree, but then we should also care about consistency. As a consequence, we should also drop MOVEIT_CLASS_FORWARD and update the style guide, but I also know that people really like it for some reason. I just found it odd that this line was changed while the line above still has SharedPtr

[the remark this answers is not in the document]
pac48: I changed it back for now.

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
pac48: Good point. I made the change.

moveit_ros/moveit_node_interface/src/node_interface.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+    rclcpp::node_interfaces::NodeParametersInterface::SharedPtr NodeInterface::get_node_parameters_interface() const {
+        return node_handle_->get_node_parameters_interface();
+    }
+
+    std::shared_ptr<rclcpp::Node> NodeInterface::get_rcl_node() const {
+        if (!has_lifecycle_node_) {
+            return node_handle_->get_rcl_node();
+        }
+        return nullptr;
```
tylerjw, line 57: This should probably return a `std::option<std::shared_ptr<rclcpp::Node>>` instead if a nullptr here.

The reason is that if you expose this and the user depends on this then it is easy for them to make mistakes and cause a segfault.

My understanding is that you want to use IK solvers in a ros2_control plugin which has a LifecycleNode. If any IK solvers use this function and depend on it returning a Node, you can't use it in your ros2_control controller.

If you can't just remove this function you should use option because that will create much nicer errors when a user makes a mistake and assumes this works. If you return a nullptr and they dereference it is at best going to segfault.

moveit_ros/planning/CHANGELOG.rst
```diff
@@ -2,9 +2,6 @@
 Changelog for package moveit_ros_planning
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-2.5.1 (2022-05-31)
-------------------
-
```
tylerjw, line 7: I don't think you want to change this. This is probably just a rebase mistake?

moveit_ros/planning/kinematics_plugin_loader/include/moveit/kinematics_plugin_loader/kinematics_plugin_loader.h
```diff
@@ -57,8 +58,17 @@ class KinematicsPluginLoader
   KinematicsPluginLoader(const rclcpp::Node::SharedPtr& node,
                          const std::string& robot_description = "robot_description",
                          double default_search_resolution = 0.0)
-    : node_(node), robot_description_(robot_description), default_search_resolution_(default_search_resolution)
+    : robot_description_(robot_description), default_search_resolution_(default_search_resolution)
   {
+    node_interface_ = std::make_shared<node_interface::NodeInterface>(node_interface::NodeInterface(node));
+  }
```
tylerjw, line 64: Put the node interface in the initalizer list with the rest of members.

moveit_ros/planning/kinematics_plugin_loader/include/moveit/kinematics_plugin_loader/kinematics_plugin_loader.h
```diff
@@ -57,8 +58,17 @@ class KinematicsPluginLoader
...
+
+  KinematicsPluginLoader(const node_interface::NodeInterfaceSharedPtr& node_interface,
+                         const std::string& robot_description = "robot_description",
+                         double default_search_resolution = 0.0)
+    : robot_description_(robot_description), default_search_resolution_(default_search_resolution)
+  {
+    node_interface_ = node_interface;
```
tylerjw, line 71: Same comment as above, use the initializer list like the reset of the member variables.

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -68,16 +69,17 @@ class KinematicsPluginLoader::KinematicsLoaderImpl
    * \param search_res
    * \param iksolver_to_tip_links - a map between each ik solver and a vector of custom-specified tip link(s)
    */
-  KinematicsLoaderImpl(const rclcpp::Node::SharedPtr& node, const std::string& robot_description,
+  KinematicsLoaderImpl(node_interface::NodeInterfaceSharedPtr node_interface, const std::string& robot_description,
                        const std::map<std::string, std::vector<std::string>>& possible_kinematics_solvers,
                        const std::map<std::string, std::vector<double>>& search_res,
                        const std::map<std::string, std::vector<std::string>>& iksolver_to_tip_links)
-    : node_(node)
+    : node_interface_{ std::move(node_interface) }
```
tylerjw, line 76: Please use this interface style in all the other constructors that have shared_ptr parameters.

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -68,16 +69,17 @@ class KinematicsPluginLoader::KinematicsLoaderImpl
...
     , robot_description_(robot_description)
     , possible_kinematics_solvers_(possible_kinematics_solvers)
     , search_res_(search_res)
     , iksolver_to_tip_links_(iksolver_to_tip_links)
   {
+//    node_interface_ = node_interface;
```
tylerjw, line 82: remove commented out code, with the std::move this would be UB if you uncommented it

moveit_ros/moveit_node_interface/include/node_interface/node_interface.h
```diff
@@ -0,0 +1,108 @@
...
+        {
+            virtual rclcpp::node_interfaces::NodeBaseInterface::SharedPtr get_node_base_interface() const = 0;
+
+            virtual rclcpp::node_interfaces::NodeTopicsInterface::SharedPtr get_node_topics_interface() const = 0;
+
+            virtual  rclcpp::node_interfaces::NodeParametersInterface::SharedPtr get_node_parameters_interface() const = 0;
+
+            virtual std::shared_ptr<rclcpp::Node> get_rcl_node() const = 0;
+
+            virtual ~NodeBase() {}
```
tylerjw, line 60: Since you specified a Destructor you need to specify the copy and move constructors and assiment operators. This is often called the 'Rule of 5': https://en.cppreference.com/w/cpp/language/rule_of_three

moveit_core/kinematics_base/src/kinematics_base.cpp
```diff
@@ -92,6 +92,26 @@ bool KinematicsBase::initialize(const rclcpp::Node::SharedPtr& /*node*/,
...
+
+    if (node->has_lifecycle_node_){
+        RCLCPP_ERROR(LOGGER,
+                     "IK plugin for group '%s' relies on deprecated API. "
+                     "Please implement initialize(rclcpp::Node::SharedPtr, RobotModel, ...).",
+                     group_name.c_str());
+    } else{
+        // try rclcpp node interface
+        return initialize(node->get_rcl_node(), robot_model, group_name, base_frame, tip_frames,
+                          search_discretization);
```
tylerjw, line 109: I'm confused by this, if the node interface is the lifecycle one, it won't initalize?

tylerjw (commented):
I really like the approach of this. This seems like a really straightforward way to abstract the two types of Nodes in ros2 for the IK interface.

moveit_ros/moveit_node_interface/src/node_interface.cpp
pac48: I agree, it should be an optional.

moveit_ros/moveit_node_interface/include/node_interface/node_interface.h
pac48: I think the destruct can actually be removed.

moveit_core/kinematics_base/src/kinematics_base.cpp
pac48: Becuase I chnaged KinematicsPluginLoader::KinematicsLoaderImpl to call KinematicsBase::initialize with NodeInterfaceSharedPtr instead of rclcpp::Node::SharedPtr, IK plugins that use rclcpp::Node::SharedPt would now never get loaded. If this ocde is hit, it means the user did not load a plugin with the NodeInterfaceSharedPtr interface. So I need to call KinematicsBase::initialize with rclcpp::Node::SharedPtr in case that interface is being used.

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
pac48: Removed

moveit_ros/planning/kinematics_plugin_loader/include/moveit/kinematics_plugin_loader/kinematics_plugin_loader.h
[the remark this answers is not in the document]
pac48: I made a package called moveit_node_interface that wraps the lifecycle and regular node.

moveit_ros/planning/CHANGELOG.rst
pac48: Yeah, that was not meant to be changed.

moveit_ros/planning/kinematics_plugin_loader/include/moveit/kinematics_plugin_loader/kinematics_plugin_loader.h
pac48: I made the fix.

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -39,7 +39,7 @@
 #include <geometry_msgs/msg/pose.hpp>
 #include <moveit_msgs/msg/move_it_error_codes.hpp>
 #include <moveit/macros/class_forward.h>
-#include "rclcpp/rclcpp.hpp"
+#include "node_interface/node_interface.h"
```
sjahr, line 42: You should rename this to moveit_node_interface.h to make clear, that this is not part of rclcpp

moveit_core/kinematics_base/src/kinematics_base.cpp
```diff
@@ -92,6 +92,23 @@ bool KinematicsBase::initialize(const rclcpp::Node::SharedPtr& /*node*/,
   return false;
 }
 
+bool KinematicsBase::initialize(const node_interface::NodeInterfaceSharedPtr& node_interface, const moveit::core::RobotModel& robot_model, const std::string& group_name,
+                                const std::string& base_frame, const std::vector<std::string>& tip_frames,
+                                double search_discretization)
+{
+    if (auto node = node_interface->get_rcl_node()){
```
sjahr, line 99: Is this correct? From my understanding an error is now returned if `node != nullptr`. Shouldn't this be something like
```
auto node = node_interface->get_rcl_node();
if(node == nullptr){
...
return false;
} else { ***Do something with node?***}
```

moveit_ros/moveit_node_interface/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
+cmake_minimum_required(VERSION 3.10.2)
+project(moveit_node_interface)
+
+find_package(ament_cmake REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(rclcpp_lifecycle REQUIRED)
+
+set(MOVEIT_LIB_NAME moveit_node_interface)
```
sjahr, line 8: Nit: I would remove this line and use `${PROJECT_NAME} `instead of `${MOVEIT_LIB_NAME}` since it is the same

moveit_ros/moveit_node_interface/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+install(DIRECTORY include/ DESTINATION include)
+
+if(BUILD_TESTING)
+  find_package(ament_lint_auto REQUIRED)
+  ament_lint_auto_find_test_dependencies()
+endif()
+
+ament_export_targets(export_${PROJECT_NAME} HAS_LIBRARY_TARGET)
+ament_export_dependencies(rclcpp rclcpp_lifecycle)
+
```
sjahr, line 38:
```suggestion
ament_export_include_directories(include)

```

moveit_ros/moveit_node_interface/include/node_interface/node_interface.h
```diff
@@ -0,0 +1,107 @@
...
+
+/* Author: Paul Gesel */
+
+#pragma once
+
+#include <rclcpp/rclcpp.hpp>
+#include <rclcpp_lifecycle/lifecycle_node.hpp>
+#include <memory>
+
+// The NodeInterface struct is a wrapper for rclcpp::Node and rclcpp_lifecycle::LifecycleNode types.
```
sjahr, line 43: Nit: Move this before the #includes
```
/* Author: Paul Gesel
   Description: ...
 */
 ```

File: moveit_ros/moveit_node_interface/include/node_interface/node_interface.h [hunk 24]
```diff
@@ -0,0 +1,107 @@
...
+
+namespace node_interface
```
sjahr, line 45: `moveit::node_interface` for consistency and to make clear this is not part of any ROS client lib

moveit_ros/moveit_node_interface/include/node_interface/node_interface.h
```diff
@@ -0,0 +1,107 @@
...
+{
+    struct NodeInterface
```
sjahr, line 47: Please add a description for this struct and it's members + functions

moveit_ros/moveit_node_interface/include/node_interface/node_interface.h
```diff
@@ -0,0 +1,107 @@
...
+        rclcpp::node_interfaces::NodeParametersInterface::SharedPtr get_node_parameters_interface();
+
+        std::shared_ptr<rclcpp::Node> get_rcl_node() const;
+
+        std::shared_ptr<NodeBase> node_handle_;
+
+        std::optional<std::shared_ptr<rclcpp_lifecycle::LifecycleNode>> get_lifecycle_node() const;
+    };
+
+    using NodeInterfaceSharedPtr = std::shared_ptr<NodeInterface>;
```
sjahr, line 105: Nit: Move this on top of the file so it is less likely to be over read

moveit_ros/moveit_node_interface/include/node_interface/node_interface.h
```diff
@@ -0,0 +1,107 @@
+/*********************************************************************
```
sjahr, line 1: You should make this file a .hpp file

moveit_ros/planning/rdf_loader/include/moveit/rdf_loader/rdf_loader.h
```diff
@@ -40,7 +40,7 @@
 #include <moveit/rdf_loader/synchronized_string_parameter.h>
 #include <urdf/model.h>
 #include <srdfdom/model.h>
-#include <rclcpp/rclcpp.hpp>
+#include "node_interface/node_interface.h"
```
sjahr, line 43:
```suggestion
#include <node_interface/node_interface.h>
```

sjahr (commented):
I like this improvement of the API

moveit_core/kinematics_base/src/kinematics_base.cpp
pac48: Good catch! The if/else was flipped.

moveit_ros/moveit_node_interface/CMakeLists.txt
pac48: I made the change

moveit_ros/moveit_node_interface/CMakeLists.txt
pac48: Added!

moveit_ros/moveit_node_interface/include/node_interface/node_interface.h
pac48: Done

moveit_ros/moveit_node_interface/include/node_interface/node_interface.h
```diff
@@ -0,0 +1,107 @@
...
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Paul Gesel */
+
+#pragma once
+
+#include <rclcpp/rclcpp.hpp>
```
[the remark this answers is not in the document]
pac48: Replaced with #include <rclcpp/node.hpp>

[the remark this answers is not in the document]
henningkayser: How about forward-declaring the node types and only including the interface headers here?  This should speed up build times even more.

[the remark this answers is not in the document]
pac48: Do you know if I can forward declare node like this:
class Node : public std::enable_shared_from_this<Node>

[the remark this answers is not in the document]
ChrisThrasher: Unless build times are a problem, I don't want us to start forward declaring all over the place in the name of compile times. I'd rather we first aggressively remove unnecessary headers (or include more granular headers as #1333 does) before we start resorting to clever tricks.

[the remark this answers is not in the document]
tylerjw: I saw this in my email and got worried. I see you aren't doing it here, that was just a trick for forward declaration.

Some of the most confusing interface designs I've seen started with enable_shared_from_this. Please avoid putting that on any class you create.

moveit_ros/moveit_node_interface/include/moveit_node_interface/moveit_node_interface.hpp
```diff
@@ -0,0 +1,113 @@
...
+          "NodeInterface can only wrap std::shared_ptr<rclcpp::Node> and rclcpp_lifecycle::LifecycleNode type");
+    }
+
+    rclcpp::node_interfaces::NodeBaseInterface::SharedPtr get_node_base_interface() override;
+
+    rclcpp::node_interfaces::NodeTopicsInterface::SharedPtr get_node_topics_interface() override;
+
+    rclcpp::node_interfaces::NodeParametersInterface::SharedPtr get_node_parameters_interface() override;
+
+    std::optional<std::shared_ptr<rclcpp::Node>> get_rcl_node() const override;
```
henningkayser, line 80: Why would you expose the nodes at all? Also, since this is already a template, I would say you could make this a single function `get_node()` and return T (with T being the node type, not the shared_ptr type)

tylerjw:
@destogl do you know if this change is still needed for something in ros2_control or should I close this?

@pac48 do you remember where we ended up with this change?

pac48:
> @destogl do you know if this change is still needed for something in ros2_control or should I close this?
> 
> @pac48 do you remember where we ended up with this change?

This change originally was meant to allow MoveIt's kinematics plugins to be used by lifecycle nodes. This was a requirement for ros2_control because controller are implemented using lifecycle nodes. Since ros2_control now has its own bare bones kinematics interface, there is currently no need to introduce MoveIt's kinematics to ros2_control. 

However, this change provides a consistent node interface for MoveIt. It has some value on its own. I don't have a strong opinion either way.

tylerjw:
@pac48 thank you for responding. I'm going to close it for now and open an issue referencing this as a possible implementation for when someone gets time to revisit the issue of having a generic interface to the two types of nodes.

PR #1330, opened 2022-06-09
Add a debug print for OMPL

AndyZe:
### Description

Add an OMPL print at the DEBUG level. This would have helped us in tracking down a recent bug.

### PR Verification

I modified `moveit_resources_panda_moveit_config/launch/demo.launch.py` to print at the DEBUG level:

```
.
.
.
    move_group_node = Node(
        package="moveit_ros_move_group",
        executable="move_group",
        output="screen",
        parameters=[moveit_config.to_dict()],
        arguments=["--ros-args", "--log-level", "debug"],
    )
.
.
.
```

Run the demo to see the printout. You will see the printout upon clicking Plan:

`ros2 launch moveit_resources_panda_moveit_config demo.launch.py`

The printout from OMPL looks like this:

```
[move_group-4]   - signature: 2 0 7 
[move_group-4]   - dimension: 7
[move_group-4]   - extent: 33.5633
[move_group-4]   - sanity checks for state space passed
[move_group-4]   - probability of valid states: 0.88
[move_group-4]   - average length of a valid motion: 10.622
[move_group-4]   - average number of samples drawn per second: sampleUniform()=2.26639e+07 sampleUniformNear()=1.64018e+07 sampleGaussian()=7.21038e+06
[move_group-4] Settings for the state space 'panda_arm_JointModel'
[move_group-4]   - state validity check resolution: 1%
[move_group-4]   - valid segment count factor: 1
[move_group-4]   - state space:
[move_group-4] ModelBasedStateSpace 'panda_arm_JointModel' at 0x7ff444003530
[move_group-4] 
[move_group-4] Declared parameters:
[move_group-4] longest_valid_segment_fraction = 0.01
[move_group-4] tag_snap_to_segment = 0.95
[move_group-4] valid_segment_count_factor = 1
[move_group-4] Valid state sampler named uniform with parameters:
[move_group-4] nr_attempts = 100
[move_group-4] Planner panda_arm/panda_arm specs:
[move_group-4] Multithreaded:                 No
[move_group-4] Reports approximate solutions: No
[move_group-4] Can optimize solutions:        No
[move_group-4] Aware of the following parameters: intermediate_states range
[move_group-4] Declared parameters for planner panda_arm/panda_arm:
[move_group-4] intermediate_states = 0
[move_group-4] range = 6.71266
[move_group-4] Start states:
[move_group-4] panda_joint1 = 0 
[move_group-4] panda_joint2 = -0.785 
[move_group-4] panda_joint3 = 0 
[move_group-4] panda_joint4 = -2.356 
[move_group-4] panda_joint5 = 0 
[move_group-4] panda_joint6 = 1.571 
[move_group-4] panda_joint7 = 0.785 
[move_group-4] * valid state 
[move_group-4] Tag: -1
[move_group-4] 10 goal states, threshold = 2.22045e-16, memory address = 0x7ff44400df10
[move_group-4] panda_joint1 = -0.0435932 
[move_group-4] panda_joint2 = -0.755479 
[move_group-4] panda_joint3 = -0.105317 
[move_group-4] panda_joint4 = -2.45022 
[move_group-4] panda_joint5 = -0.0726383 
[move_group-4] panda_joint6 = 1.69726 
[move_group-4] panda_joint7 = 0.673878 
[move_group-4] * valid state 
[move_group-4] Tag: -1

```

### DIFF
--- a/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
+++ b/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
@@ -750,6 +750,11 @@ void ompl_interface::ModelBasedPlanningContext::postSolve()
   {
     RCLCPP_WARN(LOGGER, "Computed solution is approximate");
   }
+
+  // Debug OMPL setup and solution
+  std::stringstream debug_out;
+  ompl_simple_setup_->print(debug_out);
+  RCLCPP_DEBUG(LOGGER, "%s", rclcpp::get_c_string(debug_out.str()));
 }
 
 bool ompl_interface::ModelBasedPlanningContext::solve(planning_interface::MotionPlanResponse& res)


moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
```diff
@@ -279,8 +279,14 @@ void ompl_interface::ModelBasedPlanningContext::useConfig()
     return;
   std::map<std::string, std::string> cfg = config;
 
+  auto it = cfg.find("debug_mode");
+  if (it != cfg.end() && !(it->second.compare("1")))
```
AndyZe, line 283: This is a little tricky because it->second is a string (see L280). Happy to hear a better way to do this!

moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
```diff
@@ -279,8 +279,14 @@ void ompl_interface::ModelBasedPlanningContext::useConfig()
...
+  {
+    ompl_simple_setup_->print();
```
henningkayser, line 285:
```suggestion
  std::stringstream debug_out;
  ompl_simple_setup_->print(debug_out);
  RCLCPP_DEBUG(LOGGER, "%s", rclcpp::get_c_string(debug_out.str()));
```

henningkayser (commented):
I really like the idea of printing this content. I'd even consider dropping the flag and simply print to RCLCPP_DEBUG with the stringstream like suggested. Why did you pick this function for printing the space? Would it maybe make more sense to print it after setup(), or even after solving?

moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
AndyZe: This is probably my ignorance, but I don't see any output from this. Am I doing something wrong?

`ros2 launch moveit_resources_panda_moveit_config demo.launch.py --debug`

mamoll: It makes sense to call print right after `SimpleSetup::solve` is called (maybe in [ompl_interface::ModelBasedPlanningContext::postSolve](https://github.com/ros-planning/moveit2/blob/main/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp#L742-L753)?). I modified a basic demo program print the SimpleSetup object right after it's created and before it's configured. In that case the output would look something like this:
```
Properties of the state space 'SE2CompoundSpace0'
  - signature: 6 4 3 1 2 2 1 
  - dimension: 3
  - extent: 4.39922
Call setup() before to get more information
Settings for the state space 'SE2CompoundSpace0'
  - state validity check resolution: 1%
  - valid segment count factor: 1
  - state space:
Compound state space 'SE2CompoundSpace0' of dimension 3 (locked) [
Real vector state space 'RealVectorSpace1' of dimension 2 with bounds: 
  - min: -1 -1 
  - max: 1 1 
 of weight 1
SO2 state space 'SO2Space2'
 of weight 0.5
]
No registered projections

Declared parameters:
longest_valid_segment_fraction = 0.01
valid_segment_count_factor = 1
Valid state sampler named uniform with parameters:
nr_attempts = 100
Start states:
Goal = nullptr
OptimizationObjective = nullptr
There are 0 solutions
```
If I call SimpleSetup::print right after planning, the output would look like this:
```
Properties of the state space 'SE2CompoundSpace0'
  - signature: 6 4 3 1 2 2 1 
  - dimension: 3
  - extent: 4.39922
  - sanity checks for state space passed
  - probability of valid states: 0.808
  - average length of a valid motion: 1.71476
  - average number of samples drawn per second: sampleUniform()=1.81818e+07 sampleUniformNear()=2.38095e+07 sampleGaussian()=7.0922e+06
Settings for the state space 'SE2CompoundSpace0'
  - state validity check resolution: 1%
  - valid segment count factor: 1
  - state space:
Compound state space 'SE2CompoundSpace0' of dimension 3 (locked) [
Real vector state space 'RealVectorSpace1' of dimension 2 with bounds: 
  - min: -1 -1 
  - max: 1 1 
 of weight 1
SO2 state space 'SO2Space2'
 of weight 0.5
]
Registered projections:
  - <default>
Projection of dimension 2
Cell sizes (computed defaults): [0.1 0.1]

Declared parameters:
longest_valid_segment_fraction = 0.01
projection.cellsize.0 = 0.1
projection.cellsize.1 = 0.1
projection.cellsize_factor = 0
valid_segment_count_factor = 1
Valid state sampler named uniform with parameters:
nr_attempts = 100
Planner LBKPIECE1 specs:
Multithreaded:                 No
Reports approximate solutions: No
Can optimize solutions:        No
Aware of the following parameters: border_fraction min_valid_path_fraction range
Declared parameters for planner LBKPIECE1:
border_fraction = 0.9
min_valid_path_fraction = 0.5
range = 0.879845
Start states:
Compound state [
RealVectorState [0.5 -0.35144]
SO2State [-1.25508]
]
Goal state, threshold = 2.22045e-16, memory address = 0x600001cad3c8, state = 
Compound state [
RealVectorState [-0.5 0.389726]
SO2State [-1.62618]
]
OptimizationObjective = nullptr
There are 1 solutions
```
The print method is defined [here](https://github.com/ompl/ompl/blob/main/src/ompl/geometric/src/SimpleSetup.cpp#L226-L240). I find it useful to see the exact planner that was being used, including its parameters, as well as the problem definition (which includes the start and goal states, any optimization objectives, and how many solutions were found). It's a useful sanity check to make sure parameters were read correctly and planning is performed as expected.

henningkayser: > This is probably my ignorance, but I don't see any output from this. Am I doing something wrong?
> 
> `ros2 launch moveit_resources_panda_moveit_config demo.launch.py --debug`

You need to set the log level for the node: https://docs.ros.org/en/rolling/Tutorials/Logging-and-logger-configuration.html#logger-level-configuration-command-line. Note, that the example is a run command, for launch files you'd have to add the `--log-level` arg as a node argument. There also is a service for setting logger levels dynamically for a node namespace, I haven't tried that one yet, though. https://docs.ros.org/en/rolling/Tutorials/Logging-and-logger-configuration.html#logger-level-configuration-externally.

@mamoll thanks for the examples. We should definitely print after configuration or solve, the earlier prints don't even show the planner alg.

AndyZe: OK, I changed the PR to print at the DEBUG level and moved the print to `postSolve()`.

henningkayser (approved):

PR #1339, opened 2022-06-11
Update pre-commit run on Ubuntu 22.04

vatanaksoytezer:
Although our main branch is meant to run on Ubuntu 22.04 our pre-commit was running on ubuntu-latest which still targets Ubuntu 20.04. This PR should fix this


tylerjw (approved):

henningkayser (approved):
Huh, I was already wondering about these showing up, but didn't investigate it further. Thanks!

PR #1342, opened 2022-06-11
Add dependabot for docker and gh actions

vatanaksoytezer:
Add dependabot to keep up to date with changing gh actions and docker action versions.

Inspired by what navigation2 does.


henningkayser (approved):
This should be backported to the other branches, right?

JafarAbdi (approved):

henningkayser:
> This should be backported to the other branches, right?

@vatanaksoytezer @JafarAbdi

PR #1343, opened 2022-06-11
Organize docker and testspace event conditions 

vatanaksoytezer:
Some QoL changes to docker and testspace conditions so CIs are happy in forks and docker's are built in PRs that changes docker related files, without being pushed.


.github/workflows/docker.yaml
```diff
@@ -8,6 +8,10 @@ on:
   push:
     branches:
       - main
+  pull_request:
+    paths:
```
henningkayser, line 12: nice

henningkayser: maybe add the repos file as well

JafarAbdi (approved):

PR #1331, opened 2022-06-09
Removing some boost usage

henrygerardmoore:
### Description

Taking over from #1118, this is part 1 of the work that will be done. More boost usage can be (somewhat) easily removed, but the PR is already quite large so that will take place in another PR. The changes included in this PR are

- `boost::array` -> `std::array`
- `boost::function` -> `std::function`
- `boost::mutex` and similar objects (`boost::shared_lock` etc) -> `std::mutex`
- `boost::starts_with(name, object.first)` -> `name.rfind(object.first, 0) == 0`
- `boost::math::tools::epsilon<double>()` -> `std::numeric_limits<double>::min()`
- `boost::filesystem` -> `std::filesystem`
- `boost::thread` -> `std::thread`
- Adding now-necessary includes
- Changing some calls that differ between `boost` and the `std` equivalent

In a future PR, the following will be done:

- replace `boost::regex`
- replace `boost::noncopyable`
- replace `boost::lexical_cast`
- replace `boost::optional`
- replace `boost::thread_group`

When MoveIt is updated to C++20, any use of `boost::math::constants` can be replaced with `std::numbers` constants

### Checklist
- [x] **Required by CI**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)
- [x] Extend the tutorials / documentation [reference](http://moveit.ros.org/documentation/contributing/)
- [x] Document API changes relevant to the user in the [MIGRATION.md](https://github.com/ros-planning/moveit/blob/master/MIGRATION.md) notes
- [x] Create tests, which fail without this PR [reference](https://ros-planning.github.io/moveit_tutorials/doc/tests/tests_tutorial.html)
- [x] Include a screenshot if changing a GUI
- [x] While waiting for someone to review your request, please help review [another open pull request](https://github.com/ros-planning/moveit/pulls) to support the maintainers

[//]: # "You can expect a response from a maintainer within 7 days. If you haven't heard anything by then, feel free to ping the thread. Thank you!"

PR #1372, opened 2022-06-16
Removing more boost usage

henrygerardmoore:
### Description

As a folllow-on to #1331, this PR was made to break up the work to avoid necessitating too much review. This PR includes
replace boost::regex
replace boost::noncopyable
replace boost::lexical_cast
replace boost::optional
replace boost::thread_group

- Replacing `boost::regex`
- Replacing `boost::noncopyable`
- Replacing most usage of `boost::lexical_cast`
- Replacing `boost::optional`
- Replacing `boost::thread_group`

### Checklist


moveit_core/robot_state/src/conversions.cpp
```diff
@@ -45,6 +45,8 @@
 #include <tf2_eigen/tf2_eigen.h>
 #endif
 #include <boost/lexical_cast.hpp>
+#include <string>
+#include "rclcpp/rclcpp.hpp"
```
henningkayser, line 49: This has just been removed, it shouldn't be needed.
```suggestion
```

moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
```diff
@@ -58,6 +58,7 @@
 #include <tf2_ros/buffer.h>
 
 #include <boost/python.hpp>
+#include <boost/noncopyable.hpp>
```
henningkayser, line 61: remove?

henningkayser: But why would it have to be added back? we don't use noncopyable, do we?

henrygerardmoore: Later in this file there is `bp::class_<MoveGroupInterfaceWrapper, boost::noncopyable> move_group_interface_class(
      "MoveGroupInterface", bp::init<std::string, std::string, bp::optional<std::string, double>>());`

henningkayser (approved):
I went through all the changes and they look really good to me. Great work!

henningkayser (approved):

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -52,6 +52,7 @@
 #include <moveit/macros/console_colors.h>
 #include <functional>
 #include <moveit/robot_model/aabb.h>
+#include <cassert>
```
AndyZe, line 55: Nit: sort / alphabetize these includes

moveit_core/utils/src/robot_model_test_utils.cpp
```diff
@@ -34,7 +34,6 @@
 
 /* Author: Bryce Willey */
 
-#include <boost/algorithm/string_regex.hpp>
 #include <boost/math/constants/constants.hpp>
```
AndyZe, line 37: You could get rid of this because PI is available from `math.h`. I don't care too much about this suggestion, though

moveit_core/utils/src/robot_model_test_utils.cpp
```diff
@@ -117,8 +117,11 @@ RobotModelBuilder::RobotModelBuilder(const std::string& name, const std::string&
 void RobotModelBuilder::addChain(const std::string& section, const std::string& type,
                                  const std::vector<geometry_msgs::msg::Pose>& joint_origins, urdf::Vector3 joint_axis)
 {
-  std::vector<std::string> link_names;
-  boost::split_regex(link_names, section, boost::regex("->"));
+  // https://stackoverflow.com/questions/53849/how-do-i-tokenize-a-string-in-c
+  std::regex re("->");
+  auto const link_names = std::vector<std::string>(std::sregex_token_iterator{ begin(section), end(section), re, -1 },
+                                                   std::sregex_token_iterator{});
```
AndyZe, line 123: In this block of code, it seems to me that the boost version is more readable and nicer than the alternative. I'd actually prefer to keep boost.

moveit_planners/chomp/chomp_motion_planner/include/chomp_motion_planner/multivariate_gaussian.h
```diff
@@ -63,32 +61,26 @@ class MultivariateGaussian
...
 template <typename Derived1, typename Derived2>
 MultivariateGaussian::MultivariateGaussian(const Eigen::MatrixBase<Derived1>& mean,
                                            const Eigen::MatrixBase<Derived2>& covariance)
-  : mean_(mean)
-  , covariance_(covariance)
-  , covariance_cholesky_(covariance_.llt().matrixL())
-  , rng_()
-  , normal_dist_(0.0, 1.0)
-  , gaussian_(rng_, normal_dist_)
+  : mean_(mean), covariance_(covariance), covariance_cholesky_(covariance_.llt().matrixL()), gaussian_(0.0, 1.0)
```
AndyZe, line 73: It looks like you are no longer initializing `normal_dist_`. Are you sure that's OK? Or maybe `normal_dist_` is not needed and could be removed.

moveit_ros/benchmarks/benchmarks/src/benchmark_execution.cpp
```diff
@@ -61,6 +60,9 @@
 
 #include <fstream>
 #include <memory>
+#include <string>
+#include <stdexcept>
+#include <regex>
```
AndyZe, line 65: Nit: sort alphabetically

moveit_ros/moveit_servo/include/moveit_servo/pose_tracking.h
```diff
@@ -39,7 +39,7 @@
 #pragma once
 
 #include <atomic>
-#include <boost/optional/optional.hpp>
+#include <optional>
```
AndyZe, line 42: sort alphabetically

moveit_planners/chomp/chomp_motion_planner/include/chomp_motion_planner/multivariate_gaussian.h
henrygerardmoore: Yeah I did remove `normal_dist_` from the rest of the function here. `std` doesn't need it, you can sample directly from the `std::normal_distribution` object by using the Mersenne twister object.

moveit_core/utils/src/robot_model_test_utils.cpp
henrygerardmoore: Should I use `#define _USE_MATH_DEFINES` before including `cmath` or `math.h`, or use `add_definitions(-D_USE_MATH_DEFINES)`? Or should we just leave it as using boost? We might be able to remove boost from some `CMakeList.txt`s or `package.xml`s if we change.

AndyZe: If this stackexchange answer is right, you can just do `#include <math.h>`

https://stackoverflow.com/a/1727896

No need to #define anything or add_definitions as far as I can tell since we're targeting newer systems. I could be wrong, though.

henrygerardmoore: Ok great sounds good, I'll switch to that then.

henrygerardmoore:
Closes #1373

AndyZe (approved):

PR #1380, opened 2022-06-19
Launch file cleanup

AndyZe:
### Description

Clean up several launch files:

- Update ros2_control launching. You can compare this usage against the example here:  https://github.com/ros-controls/ros2_control_demos/blob/master/ros2_control_demo_bringup/launch/rrbot.launch.py#L79
- Clean up several unused args in `prbt_moveit_config/demo.launch.py`
- Delete deprecated launch files in `prbt_moveit_config` (those having a `.launch` prefix rather than `.launch.py`


vatanaksoytezer (approved):
Looks great, thank you for doing this @AndyZe!

PR #1384, opened 2022-06-21
fix regression from #1331

mikeferguson:
### Description

The noted PR swapped boost::thread to std::thread - however
std::thread will cause the program to crash with a note that
TERMINATE CALLED WITHOUT ACTIVE EXCEPTION if the thread object
goes out of scope while not joinable. Detaching the thread
restores the original workflow prior to the PR and allows
users to send more than one trajectory via the motion planning
plugin in RVIZ (otherwise, it crashes on the first trajectory).

### Checklist


tylerjw (approved):
Thank you for this. I'm the future when we have jthread we could use that to fix this that way too.

mikeferguson:
@tylerjw yeah - I'm not sure this is the *best* way to fix this - but it's definitely the simplest way to get things back to working

tylerjw:
> @tylerjw yeah - I'm not sure this is the _best_ way to fix this - but it's definitely the simplest way to get things back to working

I don't understand how detached threads don't create a use after scope but it seems they are fairly common in ros code. I'm ok with it until someone has time to clean it up, definitely better than the bug you experienced.

PR #1386, opened 2022-06-21
Add generic cost function to KinematicsBase, CartesianInterpolator, Robot State

wyattrees:
### Description

This adds a generic cost function to be utilized by kinematics solvers, `kinematics::KinematicsBase::IKCostFn`. An instance of such may be passed to `KinematicsBase::searchPositionIK`. Additionally, it may be passed to `CartesianInterpolator::computeCartesianPath` and `RobotState::setFromIK`, where it is simply passed through to the kinematics solver.

### Checklist

### DIFF
--- a/moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
+++ b/moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
@@ -154,6 +154,10 @@ public:
   using IKCallbackFn = std::function<void(const geometry_msgs::msg::Pose&, const std::vector<double>&,
                                           moveit_msgs::msg::MoveItErrorCodes&)>;
 
+  /** @brief Signature for a cost function used to evaluate IK solutions. */
+  using IKCostFn = std::function<double(const geometry_msgs::msg::Pose&, const moveit::core::RobotState&,
+                                        moveit::core::JointModelGroup*, const std::vector<double>&)>;
+
   /**
    * @brief Given a desired pose of the end-effector, compute the joint angles to reach it
    *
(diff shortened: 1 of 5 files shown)


moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -154,6 +154,9 @@ class MOVEIT_KINEMATICS_BASE_EXPORT KinematicsBase
   using IKCallbackFn = std::function<void(const geometry_msgs::msg::Pose&, const std::vector<double>&,
                                           moveit_msgs::msg::MoveItErrorCodes&)>;
 
+  /** @brief Signature for a cost function used to evaluate IK solutions. */
+  using IKCostFn = std::function<double(const geometry_msgs::msg::Pose&, const moveit::core::RobotState&)>;
```
henningkayser, line 158: Will this be enough to cover our use cases? I thought we might want to have the seed state in here as well

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -322,6 +325,46 @@ class MOVEIT_KINEMATICS_BASE_EXPORT KinematicsBase
...
+   * @param error_code an error code that encodes the reason for failure or success
+   * @param options container for other IK options. See definition of KinematicsQueryOptions for details.
+   * @param context_state (optional) the context in which this request is being made.
+   *        The position values corresponding to joints in the current group may not match those in ik_seed_state.
+   *        The values in ik_seed_state are the ones to use. The state is passed to provide the \em other joint values,
+   *        in case they are needed for context, like with an IK solver that computes a balanced result for a biped.
+   * @return True if a valid solution was found, false otherwise
+   */
+  virtual bool
+  searchPositionIK(const std::vector<geometry_msgs::msg::Pose>& ik_poses, const std::vector<double>& ik_seed_state,
```
henningkayser, line 351: This function signature is huge, just like the others. It should be simplified if possible, cleaning up return types and so on. I think most of the function arguments should be moved into KinematicsQueryOptions or something similar at some point, maybe using a builder pattern. I won't hold up the PR over this, though.

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
wyattrees: I think the seed state is a good idea to put in here as well. For our use case, there's actually [a lot more](https://github.com/PickNikRobotics/bio_ik/blob/ad75b2e8085338311a2f6e7224f43b9b1f35a79c/include/bio_ik/goal.h#L49-L95) that can go into evaluating a cost function. While including all of this here is probably superfluous, seed state will allow us to do things like a minimal displacement cost function

wyattrees: I also added a `JointModelGroup*` parameter, since a lot of what we can do with `RobotState` requires it

wyattrees:
Testing can be done with [this new tutorial](https://github.com/ros-planning/moveit2_tutorials/pull/462) in moveit2_tutorials

moveit_core/robot_state/include/moveit/robot_state/cartesian_interpolator.h
```diff
@@ -169,42 +169,44 @@ class CartesianInterpolator
...
+     the returned path is truncated up to just before the jump.
+
+     Kinematics solvers may use cost functions to prioritize certain solutions, which may be specified with \e cost_function. */
+  static Distance computeCartesianPath(
+      RobotState* start_state, const JointModelGroup* group, std::vector<std::shared_ptr<RobotState>>& traj,
+      const LinkModel* link, const Eigen::Vector3d& direction, bool global_reference_frame, double distance,
+      const MaxEEFStep& max_step, const JumpThreshold& jump_threshold,
+      const GroupStateValidityCallbackFn& validCallback = GroupStateValidityCallbackFn(),
+      const kinematics::KinematicsQueryOptions& options = kinematics::KinematicsQueryOptions(),
+      const kinematics::KinematicsBase::IKCostFn& cost_function = kinematics::KinematicsBase::IKCostFn());
```
tylerjw, line 181: `std::function` objects should usually be passed by value because it is more optimal when the implementation saves this function object somewhere and nearly the same otherwise.

Here is an explanation: https://stackoverflow.com/questions/18365532/should-i-pass-an-stdfunction-by-const-reference

tylerjw (commented):
One small comment; otherwise looks really good.

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
tylerjw: One thing to note is that because this is a `std::function` if you have the state externally you want to pass in there is always the option of using a lambda and capturing those values.

moveit_core/robot_state/include/moveit/robot_state/cartesian_interpolator.h
wyattrees: Gotcha, I've changed them all to be passed by value

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -322,6 +326,46 @@ class MOVEIT_KINEMATICS_BASE_EXPORT KinematicsBase
...
+                   double timeout, const std::vector<double>& consistency_limits, std::vector<double>& solution,
+                   const IKCallbackFn& solution_callback, const IKCostFn cost_function,
```
tylerjw, line 354: You don't want to mark the IKCostFn const here. Value parameters marked const in the declaration has no meaning as they can't modify the original version as they are passed by value.

tylerjw (commented):
One more tiny thing. When you pass by the value you shouldn't mark the value const in the declaration (the code that goes in the header file describing the interface). This is because making it const has no meaning in this case.

You can mark them const in the source files as that does have meaning. The compiler will ignore the const word in this case in the header as the declaration of a function or method only describes the contract with the user of that function (not if internally it modifies the value or not).

tylerjw (approved):

Issue #1388, opened 2022-06-24
[MSA] Sensor configuration is incorrect


mikeferguson:
The sensor configuration in MSA was ported directly from ROS1, however that style doesn't work in ROS2 (like, literally, you can't create parameters in that form with ROS2, so the loader in ROS2 has been updated). The changes are:

 * "sensors" is expected to be a vector of strings (sensor names)
 * the sensor config then needs to be namespaced under the sensor name provided
 * all of the distance/padding parameters need to be floating point - not integer or string.
 * the readme should be updated to remove the notes about issues with the PerceptionWidget

An example of a working config versus what is currently exported is shown here: https://github.com/mikeferguson/ubr_reloaded/commit/45b9e52ac6ef04c9a0fd94f8a2d18956739a278f

mikeferguson:
I'm working on this in the background - should have a PR shortly

PR #1398, opened 2022-06-28
update format of msa sensors_3d config

mikeferguson:
### Description

This mostly handles https://github.com/ros-planning/moveit2/issues/1388 - it is still possible to enter an integer where a float is required (or to leave a field blank and end up with a string), but it's a big improvement over what we currently have

### Checklist

### DIFF
--- a/moveit_setup_assistant/moveit_setup_app_plugins/src/perception_config.cpp
+++ b/moveit_setup_assistant/moveit_setup_app_plugins/src/perception_config.cpp
@@ -89,19 +89,17 @@ std::vector<SensorParameters> PerceptionConfig::load3DSensorsYAML(const std::fil
     if (sensors_node && sensors_node.IsSequence())
     {
       // Loop over the sensors available in the file
-      for (const YAML::Node& sensor : sensors_node)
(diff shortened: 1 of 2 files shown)


moveit_setup_assistant/moveit_setup_app_plugins/src/perception_config.cpp
```diff
@@ -149,7 +147,10 @@ bool PerceptionConfig::GeneratedSensorConfig::writeYaml(YAML::Emitter& emitter)
 {
   emitter << YAML::BeginMap;
   emitter << YAML::Key << "sensors";
-  emitter << YAML::Value << YAML::BeginSeq;
+  emitter << YAML::BeginSeq;
+  emitter << YAML::Value << "default_sensor";
```
mikeferguson, line 151: this is not quite right - it doesn't work if no sensors are defined...

mikeferguson:
@DLu @vatanaksoytezer this isn't perfect - but at least it exports a working configuration (and more importantly, once you have a working configuration it preserves it and won't break the configuration if you update your stuff, as long as you only have a single source).

henningkayser (approved):
The changes look all good to me.

PR #1400, opened 2022-06-28
Static variable to manage creation of multiple nodes

SalahSoliman:
### Description

The problem is when creating multiple MoveGroupInterface objects, the node gets created with the same name, so since this a shared library, whenever an object's (any where) created a counter gets incremented an give the node a name based on the order of creation.
### Checklist


henningkayser (commented):
@SalahSoliman Thanks for your fix. The private node was really not the most ideal implementation in the first place. Therefore, we're replacing it with proper callback groups in [this PR](https://github.com/ros-planning/moveit2/pull/1305). Node namespaces should then not be an issue anymore. Could you check if the other PR also fixes your problem?

tylerjw:
Does the PR from Henning solve the problem this one was addressing? If it does, do you mind closing this as we merged the other change?

henningkayser:
> Does the PR from Henning solve the problem this one was addressing? If it does, do you mind closing this as we merged the other change?

It does, as there aren't multiple nodes anymore that may conflict.

PR #1406, opened 2022-06-29
msa: fix duplicated ros_control joints

### LINKED ISSUE #1390: [MSA] joints used by multiple groups are duplicated in ros2_control.xacro
### Description

Some robots have multiple groups with similar joints (for instance: arm, arm_with_torso). When this is configured in the MSA, the joints that are present in both groups show up twice in the ros2_control.xacro

### Your environment
* ROS Distro: humble
* OS Version: Ubuntu 22.04
* Source build of feature/msa

mikeferguson:
### Description

When more than one controller can control a joint, that joint would appear multiple times. Fixes #1390

### Checklist

### DIFF
--- a/moveit_setup_assistant/moveit_setup_controllers/src/control_xacro_config.cpp
+++ b/moveit_setup_assistant/moveit_setup_controllers/src/control_xacro_config.cpp
@@ -125,7 +125,11 @@ void ControlXacroConfig::loadFromDescription()
   {
     for (const std::string& joint_name : srdf_config->getJointNames(group_name, true, false))  // exclude passive
     {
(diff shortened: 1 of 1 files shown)


henningkayser (commented):
Shouldn't `joint_names_` be changed to a `std:set` instead?

DLu:
> Shouldn't `joint_names_` be changed to a `std:set` instead?

It could be, but then you lose the ordering.

mikeferguson:
Yeah, I'm not sure how important the ordering is - I presumed the original author knew  better than I and they said it needed to preserve order

henningkayser (approved):
I think it's fine here, lgtm

PR #1408, opened 2022-06-29
[MSA] Add additional parameters to Controller definitions

### LINKED ISSUE #1392: MSA: Add configuration for action_ns in moveit controllers
### Description

I'm not even sure if this existed in the ROS1 MSA (hence labeling it as an enhancement). Currently the action_ns parameter gets set automatically to "follow_joint_trajectory" and the gripper is missing the parameter (ticketed at https://github.com/ros-planning/moveit2/issues/1390). Ideally, one would be able to edit this within the MSA, rather than patching the files manually afterwards.

### Your environment
* ROS Distro: humble
* OS Version: Ubuntu 22.04
* Source build of feature/msa

DLu:
### Description

Fixes #1392 

Maybe a bit overengineered, but adds the `action_ns` and `default` parameters to `moveit_controllers.yaml`. 

Also, changed one thing about outputting the joints. I couldn't find any code to support reading a single value for the `joint` parameter so I removed that. I also deleted the `gains` parameter output. 

### Checklist

### DIFF
--- a/moveit_setup_assistant/moveit_setup_controllers/include/moveit_setup_controllers/controller_edit_widget.hpp
+++ b/moveit_setup_assistant/moveit_setup_controllers/include/moveit_setup_controllers/controller_edit_widget.hpp
@@ -56,7 +56,7 @@ public:
   // ******************************************************************************************
 
   /// Constructor
-  ControllerEditWidget(QWidget* parent);
+  ControllerEditWidget(QWidget* parent, const FieldPointers& additional_fields);
 
   /// Set the previous data
   void setSelected(const std::string& controller_name, const ControllerInfo* info);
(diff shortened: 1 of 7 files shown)


mikeferguson:
Tested locally: one issue I found (but I can't quite see how it is happening):
 * Change the action_ns and save your files (in my case, I edit: gripper_cmd -> command)
 * If I restart MSA and reload my config, and export again (without looking at the controller config), the value is preserved (I am sure of this, because I can delete the moveit_controllers.yaml while MSA is running and hit "generate package" and I get my custom action_ns) - so the "loading from yaml" looks to work right.
 * HOWEVER, if I go back to "edit" the controller, it still shows the default gripper_cmd in the textbox

mikeferguson (approved):

mikeferguson:
Works great now - thanks!

PR #1418, opened 2022-07-07
pointcloud_octomap_updater: Don't return false when not finding optional parameter

JafarAbdi:
### Description

It was causing the function to return false even if we are missing an optinal parameter

### DIFF
--- a/moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
+++ b/moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
@@ -72,6 +72,7 @@ PointCloudOctomapUpdater::~PointCloudOctomapUpdater()
 
 bool PointCloudOctomapUpdater::setParams(const std::string& name_space)
 {
+  // This parameter is optional
   node_->get_parameter_or(name_space + ".ns", ns_, std::string());
   return node_->get_parameter(name_space + ".point_cloud_topic", point_cloud_topic_) &&
          node_->get_parameter(name_space + ".max_range", max_range_) &&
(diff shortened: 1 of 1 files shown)


moveit/package.xml
```diff
@@ -2,7 +2,7 @@
 <?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="3">
   <name>moveit</name>
-  <version>2.5.1</version>
```
tylerjw, line 5: ?

tylerjw (changes requested):
I think you must have accidentally included a bunch of version bumps in this?

MarqRazz (approved):
This fixes errors on my system running Humble.

moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
```diff
@@ -72,14 +72,14 @@ PointCloudOctomapUpdater::~PointCloudOctomapUpdater()
 
 bool PointCloudOctomapUpdater::setParams(const std::string& name_space)
 {
+  node_->get_parameter_or(name_space + ".ns", ns_, std::string());
```
AndyZe, line 75: Please add a comment here so we can follow this logic in the future.

```suggestion
  // This parameter is optional
  node_->get_parameter_or(name_space + ".ns", ns_, std::string());
```

AndyZe (approved):

vatanaksoytezer (approved):

tylerjw (approved):

PR #1419, opened 2022-07-08
Add reference to MoveIt for ROS one back

v4hn:
https://github.com/ros-planning/moveit2/pull/1307 removed the link and I would like to add it back as reference.
I will also add a more prominent link to the moveit/README.md in a moment.


AndyZe (approved):
LGTM

Issue #1200, opened 2022-04-28
Better path quality when using MoveItCpp


mamoll:
**Is your feature request related to a problem? Please describe.**
As a non-motion planning expert, I want predictable performance on easy to moderately-hard motion planning problems, where performance is broadly defined to include: fast planning times, "intuitive" motions, high repeatability. The default planner in the default planning pipeline (RRTConnect from the OMPL pipeline) is typically fast, requires minimal to no parameter tuning, but can produce wacky solutions on seemingly easy problems.

**Describe the solution you'd like**
I am looking at how `moveitcpp::PlanningComponent::plan` plans a path. I see it looks up the planning pipeline by some parameter, calls `generatePlan` for the selected pipeline and returns the solution if it finds one. I would like to propose to augment the semantics of the `parameters.planning_pipeline` parameter: instead of selecting one pipeline, you can select multiple, e.g., by setting its value to “`ompl|stomp|pilz`”. The only change would then be in `moveitcpp::PlanningComponent::plan`: it spawns interruptible threads for each planning pipeline, picks the first found solution, terminates the other threads. All existing code will behave the same way. This seems simple enough to try and might result in faster, more deterministic behavior in many cases. With some extra code, we can keep some stats around inside `moveitcpp::PlanningComponent` that tracks how often each planner “wins” and how much time it needs. These stats could be printed/logged upon destruction. (You could also further augment the parameter syntax to allow pipelines to run sequentially until one succeeds, e.g., using `>` as a separator: “`pilz>stomp>ompl`”.) 

**Describe alternatives you've considered**
* Alternative planners in the OMPL planner pipeline have more parameters that non-experts find hard to tune or are simply not as performant in general.
* There are several alternative planning pipelines that each have their own weaknesses:
   * The Pilz planner pipeline is a fast, deterministic planner, but is incomplete (i.e., it won't always find a solution if one exists.
   *  STOMP and CHOMP are typically fast and usually produce nice-looking paths, but are also incomplete.
[...]

PR #1420, opened 2022-07-08
Parallel planning pipelines

sjahr:
## Description

This PR adds a first implementation of @mamoll proposal in #1200 to run multiple planning pipelines in parallel to improve the path quality through creating multiple solutions and choose the most suitable one. I extended the MoveItCpp interface with an additional function that allows processing multiple planning requests in parallel in separate threads. Additionally,  I've refactored the kinematics_metrics package in moveit_core to a more general package called metrics which contains several metric functions for robot trajectories. These implementation were mostly extracted out of the benchmarks package to become more visible and usable at runtime outside of the benchmark context. I'll probably separate this change in a second PR once this is not WIP anymore.

### WIP tutorial + demo code
- https://github.com/ros-planning/moveit2_tutorials/pull/493

### Usage example:
config.yaml
```
one:
  plan_request_params:
    planning_attempts: 1
    planning_pipeline: ompl
    planner_id: "RRTConnectkConfigDefault"
    max_velocity_scaling_factor: 1.0
    max_acceleration_scaling_factor: 1.0

two:
  plan_request_params:
    planning_attempts: 1
    planning_pipeline: ompl
    planner_id: "PRMstarkConfigDefault"
    max_velocity_scaling_factor: 1.0
    max_acceleration_scaling_factor: 1.0

three:
  plan_request_params:
    planning_attempts: 1
    planning_pipeline: pilz_industrial_motion_planner
    planner_id: "LIN"
    max_velocity_scaling_factor: 1.0
    max_acceleration_scaling_factor: 1.0
```

demo_node.cpp
```
    planning_component->setGoal(some_goal_state);
    planning_component->setStartStateToCurrentState();

    moveit_cpp::PlanningComponent::MultiPipelinePlanRequestParameters multi_pipeline_plan_request;
    multi_pipeline_plan_request.load(node, {"one", "two", "three"});

    auto plan_solution = planning_components->plan(multi_pipeline_plan_request);
```
### TODO
- [x] Write unittests for robot_trajectory_metrics
- [X] Add default stopping criteria --> Finish when all pipelines finished
- [x] Make plan() thread safe

### Checklist


moveit_core/metrics/include/moveit/metrics/robot_trajectory_metrics.hpp
```diff
@@ -0,0 +1,71 @@
...
+namespace moveit
+{
+namespace core
+{
+/// \brief Calculate the path length of a given trajectory based on the accumulated robot state distances. The distance
+/// between two robot states is calculated based on the sum of active joint distances between the two states (L1 norm).
+/// This function is adapted from
+/// https://github.com/ros-planning/moveit2/blob/main/moveit_ros/benchmarks/src/BenchmarkExecutor.cpp#L872 \param[in]
+/// trajectory Given robot trajectory \return Length of the robot trajectory [rad]
+[[nodiscard]] double calculateTrajectoryLength(robot_trajectory::RobotTrajectory const& trajectory);
```
mamoll, line 53: I'd remove `calculate` from all the function names as that is implied.

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
```diff
@@ -73,6 +75,33 @@ class PlanningComponent
...
+  {
+  public:
+    /// \brief Constructor
+    PlanSolutions(const size_t expected_size = 0)
+    {
+      solutions_.reserve(expected_size);
+    }
+
+    /// \brief Thread safe method to add PlanSolutions to this data structure
+    void pushBack(PlanSolution plan_solution)
```
mamoll, line 88: You may want to change this later to `insert`, where the `solutions_` container is kept in sorted order according to some solution ranking mechanism. See [OMPL](https://github.com/ompl/ompl/blob/main/src/ompl/base/src/ProblemDefinition.cpp#L54-L161) for an example. It be neat if the ranking can be user-specified. Maybe that's overkill and maybe the logic for extracting the best solution can be combined with the logic for terminating planning when a "good enough" solution has been found.

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
```diff
@@ -83,15 +112,73 @@ class PlanningComponent
...
-      node->get_parameter_or(ns + "planning_pipeline", planning_pipeline, std::string(""));
-      node->get_parameter_or(ns + "planning_time", planning_time, 1.0);
-      node->get_parameter_or(ns + "planning_attempts", planning_attempts, 5);
-      node->get_parameter_or(ns + "max_velocity_scaling_factor", max_velocity_scaling_factor, 1.0);
-      node->get_parameter_or(ns + "max_acceleration_scaling_factor", max_acceleration_scaling_factor, 1.0);
+      if (!param_namespace.empty())
+      {
+        ns = param_namespace + ".plan_request_params.";
+      }
+      if (!node->get_parameter_or(ns + "planner_id", planner_id, std::string("")))
```
mamoll, line 122: Could this be simplified with [rosparam_shortcuts](https://github.com/PickNikRobotics/rosparam_shortcuts/tree/ros2)?

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
```diff
@@ -171,11 +266,13 @@ class PlanningComponent
 private:
   // Core properties and instances
   rclcpp::Node::SharedPtr node_;
-  MoveItCppPtr moveit_cpp_;
+  MoveItCppPtr moveit_cpp_;  // TODO Make thread safe!
```
mamoll, line 269: I'd be interested in understanding what is currently not thread-safe.

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
```diff
@@ -195,7 +199,70 @@ PlanningComponent::PlanSolution PlanningComponent::plan(const PlanRequestParamet
...
+  PlanningComponent::PlanSolutions planning_solutions{ parameters.multi_plan_request_parameters.size() };
+  std::vector<std::thread> planning_threads;
+  planning_threads.reserve(parameters.multi_plan_request_parameters.size());
+
+  // Launch planning threads
+  for (auto const& plan_request_parameter : parameters.multi_plan_request_parameters)
+  {
+    auto planning_thread = std::thread([&]() {
+      auto solution = plan(plan_request_parameter);
+      planning_solutions.pushBack(solution);
```
mamoll, line 223: some planners in OMPL (asymptotically optimal ones) can produce more than solution before exiting. These could also be retrieved and added to the `planning_solutions` in case the criteria the callback uses for selecting the best solution are different than the optimization objective used by planner.

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
```diff
@@ -195,7 +199,70 @@ PlanningComponent::PlanSolution PlanningComponent::plan(const PlanRequestParamet
...
   //}
-  return *last_plan_solution_;
+  return *last_plan_solution_;  // TODO: Make better
+}
+
+PlanningComponent::PlanSolution
+PlanningComponent::plan(const PlanningComponent::MultiPipelinePlanRequestParameters& parameters,
+                        std::function<bool(std::vector<std::thread> planning_threads,
+                                           PlanningComponent::MultiPipelinePlanRequestParameters& parameters)>
+                            stop_criterion_callback,
```
mamoll, line 209: I'm guessing the idea is that this callback gets called periodically in a separate thread that monitors `planning_solutions` for changes?

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
sjahr: As far as I know it is not used in the moveit2 codebase at this point and I have not used it yet. Uniform and good parameter handling is still and open issue. I've simplified the code for now by adding a template to remove the copy&paste code similar to e.g. moveit_servo

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
sjahr: That's a good idea but I feel like this should not block this PR. I've added a TODO for now

moveit_core/metrics/include/moveit/metrics/robot_trajectory_metrics.hpp
sjahr: Which name would you propose instead? I feel like a function requires a verb in its name because it does something. Would you replace `calculate` with `get`?

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
sjahr: I still have to figure out a good way how it is possible to end a thread from outside. The way this is currently implemented the execution of `pipeline->generatePlan(planning_scene, req, res)`  takes the most time and should be interrupt-able. But to do this I'd need to check e.g. an atomic flag within generatePlan on a regular basis. From my understanding the cpp std::thread API does not allow killing/interruption a thread from outside without leaking memory :/ As you previously pointed out, maybe boost thread is a solution for this.

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
sjahr: Do you know if multiple solutions are accessible through the interface that is provided by
```
PlanningComponent::PlanSolution PlanningComponent::plan(const PlanRequestParameters& parameters)
```

sjahr: Do you know if multiple solutions are accessible through the interface that is provided by
```
PlanningComponent::PlanSolution PlanningComponent::plan(const PlanRequestParameters& parameters)
```

moveit_core/metrics/include/moveit/metrics/robot_trajectory_metrics.hpp
mamoll: I'd use nothing at all (i.e., just `doube trajectoryLength(robot_trajectory::RobotTrajectory const& trajectory);`). The `get`/`set` terminology seems more appropriate for member functions, not pure functions.

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
mamoll: I don't know. My comment isn't that important. The way to get to these solutions is first check that `dynamic_cast<OMPLPlanningPipelinePtr>(planning_pipeline_ptr> != nullptr` (or something like that). Next, you have to get the ModelBasedPlanningContextPtr. Next, you loop over each solution in  `getOMPLSimpleSetup()->getProblemDefinition()->getSolutions()`. This can be done at some later point.

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
mamoll: Yes.

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
henningkayser: @sjahr where did you get the segfaults? Any stacktrace?

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
```diff
@@ -83,15 +115,54 @@ class PlanningComponent
...
     {
+      // Try to get parameter or use default
+      if (!node->get_parameter_or(param_name, output_value, default_value))
+      {
+        RCLCPP_WARN(node->get_logger(),
+                    "Parameter \'%s\' not found in config use default value instead, check parameter type and "
+                    "namespace in YAML file",
+                    (param_name).c_str());
+      }
+    }
```
tylerjw, line 130: How is this different from what was here before? Also, I don't see anything in this function that `declares` the parameter, only a call to `get_parameter_or` which depends on the parameter being declared already with some other mechanism.

tylerjw (commented):
Here are some general comments on this. Do you have a minimal example of using this somewhere so it can be tested?

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
sjahr: I added this to add a warning when a default value is used. I felt this might be helpful since it is very easy to mess up namespace and parameter declarations

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
sjahr: The issue was likely not related to the planning pipeline. I've been calling plan() in multiple threads and each tried to write to the last_plan_solution variable. I fixed that

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
```diff
@@ -189,7 +216,127 @@ PlanningComponent::PlanSolution PlanningComponent::plan(const PlanRequestParamet
...
+}
+
+PlanningComponent::PlanSolution PlanningComponent::plan(const MultiPipelinePlanRequestParameters& parameters,
+                                                        SolutionCallbackFunction solution_selection_callback,
+                                                        StoppingCriterionFunction stopping_criterion_callback)
+{
+  // Create solutions container
+  PlanningComponent::PlanSolutions planning_solutions{ parameters.multi_plan_request_parameters.size() };
+  std::vector<std::thread> planning_threads;
+  planning_threads.reserve(parameters.multi_plan_request_parameters.size());
```
henningkayser, line 234: Wouldn't it make sense to sanity-check the requested number of threads, especially since `planning_attempts` is also handled as thread count by OMPL?

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
```diff
@@ -189,7 +216,127 @@ PlanningComponent::PlanSolution PlanningComponent::plan(const PlanRequestParamet
...
+      try
+      {
+        auto solution = plan(plan_request_parameter, false);
+        planning_solutions.pushBack(solution);
+      }
+      catch (const std::exception& e)
+      {
+        RCLCPP_ERROR_STREAM(LOGGER, "Planning pipeline '" << plan_request_parameter.planning_pipeline.c_str()
+                                                          << "' threw exception '" << e.what() << "'");
+        auto plan_solution = std::make_shared<PlanSolution>();
```
henningkayser, line 249: This temporary plan_solution is not being added to the solutions list and doesn't seem to be used anywhere else

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
```diff
@@ -189,7 +216,127 @@ PlanningComponent::PlanSolution PlanningComponent::plan(const PlanRequestParamet
...
+    });
+    planning_threads.push_back(std::move(planning_thread));
+  }
+
+  // Launch monitor thread if a stopping criterion is provided as function arguments. Otherwise,
+  // the parallel planning process takes as long as the longest planner takes.
+  if (stopping_criterion_callback != nullptr)
+  {
+    auto monitor_thread = std::thread([&]() {
+      while (planning_solutions.getSolutions().size() < planning_threads.size())
```
henningkayser, line 261: This will not terminate if an exception is thrown, see comment above.

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
```diff
@@ -189,7 +216,127 @@ PlanningComponent::PlanSolution PlanningComponent::plan(const PlanRequestParamet
...
+      {
+        if (stopping_criterion_callback(planning_solutions, parameters))
```
henningkayser, line 263: It wouldn't be too bad to ensure that the stopping criterion is thread-safe. Also, if the callback is dependent on solutions, I would simply check it when a solution is returned in the planning thread. A time-out can be handled explicitly instead of this busy-wait loop

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
```diff
@@ -189,7 +216,127 @@ PlanningComponent::PlanSolution PlanningComponent::plan(const PlanRequestParamet
...
+        {
+          // Terminate planning pipelines
+          for (auto const& plan_request_parameter : parameters.multi_plan_request_parameters)
```
henningkayser, line 266: A `terminatePipelines()` function would fit well with the MoveItCpp API

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
```diff
@@ -189,7 +216,127 @@ PlanningComponent::PlanSolution PlanningComponent::plan(const PlanRequestParamet
...
+  {
+    if (planning_thread.joinable())
+    {
+      planning_thread.join();
+    }
+  }
+
+  // Select solution
+  // If no custom solution selection callback is provided, choose the shortest path available
+  if (solution_selection_callback == nullptr)
```
henningkayser, line 296: This block could itself be a SolutionSelectionCallback which you could set as a default argument.

moveit_ros/planning/planning_pipeline/include/moveit/planning_pipeline/planning_pipeline.h
```diff
@@ -173,9 +173,18 @@ class MOVEIT_PLANNING_PIPELINE_EXPORT PlanningPipeline
...
+  bool isActive() const
+  {
+    return active_;
+  }
+
 private:
   void configure();
 
+  // Flag that indicates whether or not the planning pipeline is currently solving a planning problem
+  std::atomic<bool> active_;
```
henningkayser, line 186: Why is this atomic? I don't see any atomic operations

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
mamoll: sanity check against `std::thread::hardware_concurrency()` you mean? That should at most produce a warning. Even with a single core there might be value in running multiple threads concurrently. Also, `std::thread::hardware_concurrency()` can (apparently) return 0 on some platforms.

AndyZe:
I think a good default planner setup would be:

Priority 1. Pilz
Priority 2. PRM-something
Priority 3. RRTConnect

Select Pilz by default, if it succeeded. Because it follows straight lines that are nice and predictable

RRTConnect is kind of a last-ditch method. Nobody wants that solution, unless the other planners fail

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
sjahr: Thanks!

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
sjahr: I've added a warning for it.

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
sjahr: That is a good idea. This way I can completely remove the monitor thread :+1: I removed the load() function from the MultiPipelinePlanRequestParameters struct so the StoppingCriterionFunction should be thread safe now:
```cpp
  typedef std::function<bool(PlanSolutions const& solutions,
                             MultiPipelinePlanRequestParameters const& plan_request_parameters)>
      StoppingCriterionFunction;
```

moveit_ros/planning/planning_pipeline/include/moveit/planning_pipeline/planning_pipeline.h
sjahr: I made it atomic since I assumed it might be possible that a thread tries to read this variable through `isActive()` while the another thread is calling `generatePlan(...)` which writes to it. Maybe I am misunderstanding something here: Is it possible to have a race condition between two member functions which access the same member variable?

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
```diff
@@ -43,34 +43,49 @@
...
 #include <moveit/utils/moveit_error_code.h>
+#include <moveit/planning_interface/planning_response.h>
 
 namespace moveit_cpp
 {
 MOVEIT_CLASS_FORWARD(PlanningComponent);  // Defines PlanningComponentPtr, ConstPtr, WeakPtr... etc
 
+/// \brief A function to choose the solution with the shortest path from a vector of solutions
+planning_interface::MotionPlanResponse
+getShortestSolution(std::vector<planning_interface::MotionPlanResponse> const& solutions);
```
henningkayser, line 54: This looks like a good opportunity to start a new header which defines PlanSolutions and free functions related to it.

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
```diff
@@ -83,18 +98,67 @@ class PlanningComponent
...
+        multi_plan_request_parameters.push_back(parameters);
+      }
+    }
+    // Plan request parameters for the individual planning pipelines which run concurrently
+    std::vector<PlanRequestParameters> multi_plan_request_parameters;
+  };
+
+  /// \brief A solution callback function type for the parallel planning API of planning component
+  typedef std::function<planning_interface::MotionPlanResponse(
+      std::vector<planning_interface::MotionPlanResponse> const& solutions)>
```
henningkayser, line 155: I don't think that we should mix east const into here, better be consistent with the rest. I'm open for discussions to change the style overall, though.

moveit_ros/planning/moveit_cpp/src/moveit_cpp.cpp
```diff
@@ -276,6 +276,25 @@ MoveItCpp::execute(const std::string& group_name, const robot_trajectory::RobotT
...
 }
 
+bool MoveItCpp::terminatePlanningPipeline(std::string const& pipeline_name)
+{
+  try
+  {
+    auto const& planning_pipeline = planning_pipelines_.at(pipeline_name);
+    if (planning_pipeline->isActive())
+    {
+      planning_pipeline->terminate();
```
henningkayser, line 286: `terminate()` returns true, even if the planner is not running. https://github.com/ros-planning/moveit2/blob/main/moveit_core/planning_interface/include/moveit/planning_interface/planning_interface.h#L128
We should check for `false`, though.

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
```diff
@@ -107,19 +108,31 @@ bool PlanningComponent::setPathConstraints(const moveit_msgs::msg::Constraints&
...
-PlanningComponent::PlanSolution PlanningComponent::plan(const PlanRequestParameters& parameters)
+bool PlanningComponent::setTrajectoryConstraints(const moveit_msgs::msg::TrajectoryConstraints& trajectory_constraints)
 {
-  last_plan_solution_ = std::make_shared<PlanSolution>();
+  current_trajectory_constraints_ = trajectory_constraints;
+  return true;
+}
+
+planning_interface::MotionPlanResponse PlanningComponent::plan(const PlanRequestParameters& parameters,
+                                                               const bool update_last_solution)
```
henningkayser, line 118:
```suggestion
                                                               const bool store_solution)
```
just a suggestion

moveit_ros/planning/planning_pipeline/include/moveit/planning_pipeline/planning_pipeline.h
henningkayser: It looked like it would only be used by the same thread to me. If it's accessed by multiple threads, this makes sense.

henningkayser (approved):
Sorry for letting this get stale for so long. I'm generally approving this implementation with some smaller requests.
Long-term, I would really like to iterate on this feature to make it more of a default. Some ideas:
* All plan requests (even single ones) should run in a thread with futures for synchronization
* Multi-plan-requests would simply run multiple plan requests and keep track of futures
* Things like `terminate()` shouldn't require another pipeline lookup. The pipeline or terminate function should be stored with the requested plan instance/thread.
* If possible, building of requests and parsing of responses should be in separate (free?) functions for improved readability

moveit_core/planning_interface/include/moveit/planning_interface/planning_response.h
```diff
@@ -37,23 +37,39 @@
...
 #include <moveit_msgs/msg/motion_plan_response.hpp>
 #include <moveit_msgs/msg/motion_plan_detailed_response.hpp>
 
 namespace planning_interface
 {
+/// \brief Response to a planning query
 struct MotionPlanResponse
 {
-  MotionPlanResponse() : planning_time_(0.0)
+  // Constructor
```
AndyZe, line 50:
```suggestion
  /** \brief Constructor */
```

moveit_core/planning_interface/include/moveit/planning_interface/planning_response.h
```diff
@@ -37,23 +37,39 @@
...
+  // Time to plan the respond to the planning query
   double planning_time_;
-  moveit_msgs::msg::MoveItErrorCodes error_code_;
+  // Result status of the query
+  moveit::core::MoveItErrorCode error_code_;
+  /// The full starting state used for planning
+  moveit_msgs::msg::RobotState start_state_;
+  std::string planner_id_;
+
+  // Enable checking of query success or failure, for example if(response) ...
```
AndyZe, line 68:
```suggestion
  // \brief Enable checking of query success or failure, for example if(response) ...
```

moveit_ros/hybrid_planning/global_planner/global_planner_plugins/src/moveit_planning_pipeline.cpp
```diff
@@ -132,17 +132,17 @@ moveit_msgs::msg::MotionPlanResponse MoveItPlanningPipeline::plan(
 
   // Plan motion
   auto plan_solution = planning_components->plan(plan_params);
-  if (plan_solution.error_code != moveit_msgs::msg::MoveItErrorCodes::SUCCESS)
+  if (plan_solution.error_code_ != moveit_msgs::msg::MoveItErrorCodes::SUCCESS)
```
AndyZe, line 135:
```suggestion
  if (!bool(plan_solution.error_code_))
```

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
```diff
@@ -39,40 +39,25 @@
 
 #include <rclcpp/rclcpp.hpp>
 #include <moveit/moveit_cpp/moveit_cpp.h>
+#include <moveit/moveit_cpp/plan_solutions.h>
 #include <moveit/robot_state/robot_state.h>
 #include <geometry_msgs/msg/pose_stamped.h>
 #include <moveit/robot_state/conversions.h>
 #include <moveit/utils/moveit_error_code.h>
+#include <moveit/planning_interface/planning_response.h>
```
AndyZe, line 47: Nit: alphabetize this whole list.

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
```diff
@@ -83,18 +68,67 @@ class PlanningComponent
...
+        PlanRequestParameters parameters;
+        parameters.load(node, planning_pipeline_name);
+        multi_plan_request_parameters.push_back(parameters);
+      }
     }
+    // Plan request parameters for the individual planning pipelines which run concurrently
+    std::vector<PlanRequestParameters> multi_plan_request_parameters;
   };
 
+  /// \brief A solution callback function type for the parallel planning API of planning component
```
AndyZe, line 123:
```suggestion
  /** \brief A solution callback function type for the parallel planning API of planning component */
```

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
```diff
@@ -83,18 +68,67 @@ class PlanningComponent
...
+  typedef std::function<planning_interface::MotionPlanResponse(
+      const std::vector<planning_interface::MotionPlanResponse>& solutions)>
+      SolutionCallbackFunction;
+  /// \brief A stopping criterion callback function for the parallel planning API of planning component
```
AndyZe, line 127:
```suggestion
  /** \brief A stopping criterion callback function for the parallel planning API of planning component */
```

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
```diff
@@ -154,19 +188,30 @@ class PlanningComponent
   /** \brief Set the path constraints generated from a moveit msg Constraints */
   bool setPathConstraints(const moveit_msgs::msg::Constraints& path_constraints);
 
+  /** \brief Set the trajectory constraints generated from a moveit msg Constraints */
+  bool setTrajectoryConstraints(const moveit_msgs::msg::TrajectoryConstraints& trajectory_constraints);
```
AndyZe, line 192: What the heck is `trajectory_constraints` vs `path_constraints`? The only difference is timing and vel/accel information, but the motion planners aren't concerned with that. So I don't get what this function does that `setPathConstraints()` can't do.

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
```diff
@@ -176,16 +221,19 @@ class PlanningComponent
   // The robot_model_ member variable of MoveItCpp class will manually free the joint_model_group_ resources
   const moveit::core::JointModelGroup* joint_model_group_;
 
+  std::mutex plan_mutex_;
```
AndyZe, line 224: I did a search for this variable but I don't see it used anywhere?

moveit_ros/planning/planning_pipeline/include/moveit/planning_pipeline/planning_pipeline.h
```diff
@@ -44,6 +44,7 @@
 #include <visualization_msgs/msg/marker_array.hpp>
 
 #include <memory>
+#include <atomic>
```
AndyZe, line 47: nit: alphabetize

AndyZe (commented):
Here are some mostly-nitpicky comments, apologies for that. I'll give it a test now.

We've been using this in ROS1 and it's AWESOME!

moveit_core/planning_interface/include/moveit/planning_interface/planning_response.h
tylerjw: You can also just drop this comment entirely. A comment telling us the next line is a constructor is not that helpful

moveit_core/planning_interface/include/moveit/planning_interface/planning_response.h
```diff
@@ -37,23 +37,39 @@
...
+  // Construct a ROS message from struct data
   void getMessage(moveit_msgs::msg::MotionPlanResponse& msg) const;
 
+  // Trajectory generated by the queried planner
   robot_trajectory::RobotTrajectoryPtr trajectory_;
+  // Time to plan the respond to the planning query
   double planning_time_;
-  moveit_msgs::msg::MoveItErrorCodes error_code_;
+  // Result status of the query
+  moveit::core::MoveItErrorCode error_code_;
```
tylerjw, line 63: As this is a struct (all public members), I would remove the _ suffix. That is a convention for private member variables of classes.

moveit_core/planning_interface/include/moveit/planning_interface/planning_response.h
```diff
@@ -64,6 +80,7 @@ struct MotionPlanDetailedResponse
   std::vector<std::string> description_;
   std::vector<double> processing_time_;
   moveit_msgs::msg::MoveItErrorCodes error_code_;
+  std::string planner_id_;
```
tylerjw, line 83: Same comment here about this being a struct and not needing _ suffix.

moveit_ros/planning/moveit_cpp/src/moveit_cpp.cpp
sjahr: Good point, is it enough to just check the activity of the planner instance itself or do we need to make the whole pipeline more interrupt-able, for example how long can `planning_scene->isPathValid(...)`? The isActive() function checks on the pipeline level while terminate() only checks on the planner level.

moveit_core/planning_interface/include/moveit/planning_interface/planning_response.h
sjahr: :thinking: ... you're right :sweat_smile:

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
sjahr: I simply tried to make the interface consistent since the MotionPlanRequest differs between path and trajectory constraints. 

I'd be happy to simplify the interface but I think we should do this in a follow up PR.

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
sjahr: That's a leftover, thank you! Initially I thought certain function segments of the planning component are not thread safe. I'll remove it

moveit_core/planning_interface/include/moveit/planning_interface/planning_response.h
sjahr: Sounds good, but this is requires changes in a lot of files so I'd rather prefer to make a follow up PR for this

moveit_core/planning_interface/include/moveit/planning_interface/planning_response.h
sjahr: See comment below

moveit_core/planning_interface/include/moveit/planning_interface/planning_response.h
sjahr: I opened an issue https://github.com/ros-planning/moveit2/issues/1689

AndyZe (approved):
I gave it a final test with the WIP tutorial. LGTM!

https://github.com/ros-planning/moveit2_tutorials/pull/493

Issue #1334, opened 2022-06-10
CI: RDF loader test is flaky


henrygerardmoore:
### Description

The RDF Loader appears to fail in CI for no reason on [one commit](https://github.com/ros-planning/moveit2/pull/1331/commits/58f3828224f051b40ce0b66d217ae4335703a5d2) and then succeeds on [the other](https://github.com/ros-planning/moveit2/pull/1331/commits/190aa61aa48bd504b569767e27410f907339eeca) where only a comment was removed.

### Your environment
* ROS Distro: Humble
* OS Version: Ubuntu 22.04
* Built from source on the remove_boost branch of my fork (gh pr checkout 1331 to see)

### Steps to reproduce
Only encountered this once, so it is difficult to say how to reproduce it. Possibly just repeated runs of CI will do it.

### Expected behaviour
Test should pass consistently

### Actual behaviour
Test appears to fail inconsistently

### Backtrace or Console output

[CI for failing test](https://github.com/ros-planning/moveit2/runs/6821795888?check_suite_focus=true)
[CI for passing test](https://github.com/ros-planning/moveit2/runs/6834209568?check_suite_focus=true)

mikeferguson:
I notice in the logs "terminate called without an active exception" - which usually means that a std::thread went out of scope while not joined/stopped - I notice we have a std::thread in this test - but I don't actually see how it could go out of scope...

I fixed an instance of this in https://github.com/ros-planning/moveit2/pull/1384

henningkayser:
I haven't seen this one lately. Are we still running into this? Maybe it really was an unrelated threading issue somewhere...

vatanaksoytezer:
I think I saw this like a few months ago, but haven't seen lately.

henningkayser:
closing for now

PR #1422, opened 2022-07-08
Rename cartesian_limits.yaml

### LINKED ISSUE #1421: Rename `cartesian_limits.yaml` --> `pilz_cartesian_limits.yaml`
Since this file is only used by Pilz, I think it makes sense to rename it.

Motivated by this question:  https://answers.ros.org/question/403542/set-cartesian-limits-with-moveit/?answer=403545#post-id-403545

AndyZe:
I think it makes sense to specify this is only for Pilz. [Users were getting confused](https://answers.ros.org/question/403542/set-cartesian-limits-with-moveit/).

Merge with https://github.com/ros-planning/moveit_resources/pull/146.

Fixes #1421

### DIFF
--- a/moveit_planners/test_configs/prbt_moveit_config/launch/demo.launch.py
+++ b/moveit_planners/test_configs/prbt_moveit_config/launch/demo.launch.py
@@ -59,13 +59,13 @@ def generate_launch_description():
     joint_limits_yaml = load_yaml(
         "moveit_resources_prbt_moveit_config", "config/joint_limits.yaml"
     )
-    cartesian_limits_yaml = load_yaml(
-        "moveit_resources_prbt_moveit_config", "config/cartesian_limits.yaml"
+    pilz_cartesian_limits_yaml = load_yaml(
+        "moveit_resources_prbt_moveit_config", "config/pilz_cartesian_limits.yaml"
     )
     robot_description_planning = {
         "robot_description_planning": {
             **joint_limits_yaml,
-            **cartesian_limits_yaml,
+            **pilz_cartesian_limits_yaml,
         }
     }
     kinematics_yaml = load_yaml(
--- a/moveit2.repos
+++ b/moveit2.repos
@@ -1,1 +1,10 @@
 repositories:
+  # https://github.com/ros-planning/moveit_resources/pull/147
+  moveit_resources:
+    type: git
+    url: https://github.com/ros-planning/moveit_resources.git
+    version: ros2
+  ros2_control:
+    type: git
+    url: https://github.com/ros-controls/ros2_control.git
+    version: 2.12.0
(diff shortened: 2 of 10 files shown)


moveit_planners/test_configs/prbt_moveit_config/launch/demo.launch.py
```diff
@@ -59,13 +59,13 @@ def generate_launch_description():
     joint_limits_yaml = load_yaml(
         "moveit_resources_prbt_moveit_config", "config/joint_limits.yaml"
     )
-    cartesian_limits_yaml = load_yaml(
-        "moveit_resources_prbt_moveit_config", "config/cartesian_limits.yaml"
+    pilz_cartesian_limits_yaml = load_yaml(
```
mikeferguson, line 62: should we do anything to try and migrate the existing file to the new filename for users who are updating their configuration? (it's possible there are very few users who fall in this bucket - we haven't even cut a release with MSA yet AFAIK).

mikeferguson (approved):

moveit_planners/test_configs/prbt_moveit_config/launch/demo.launch.py
AndyZe: Probably should print a warning from `moveit_configs_builder.py`, huh?

AndyZe: Added a print statement that gives output like this:

```
[INFO] [launch]: All log files can be found below /home/andy/.ros/log/2022-07-08-13-02-04-616594-andy-G7-7500-225832
[INFO] [launch]: Default logging verbosity is set to INFO
WARNING:root:cartesian_limits.yaml is deprecated, please rename to pilz_cartesian_limits.yaml
[ERROR] [launch]: Caught exception in launch (see debug for traceback): Caught exception when trying to load file of format [py]: "File /home/andy/ws_ros2/install/moveit_resources_panda_moveit_config/share/moveit_resources_panda_moveit_config/config/pilz_cartesian_limits.yaml doesn't exist"
```

AndyZe:
I think CI is failing because the moveit_resources change hasn't propagated yet (https://github.com/ros-planning/moveit_resources/pull/146).

`"File /opt/ros/rolling/share/moveit_resources_panda_moveit_config/config/pilz_cartesian_limits.yaml doesn't exist"`

Somebody with higher-level permissions will probably have to merge it (or just wait). @henningkayser @mamoll @nbbrooks

mikeferguson:
@AndyZe I think we'll actually need a new release of moveit_resources for this to pass - looks like @vatanaksoytezer has been the one releasing moveit_resources recently?

vatanaksoytezer (approved):

tylerjw:
I pushed my change here to update the repos file for moveit_resources and add the fix for the breaking change in ros2_control that was just released. That way I can merge both of these changes (there was no way to make either pass CI without each-other).

mikeferguson:
@tylerjw do we actually want to be merging things that won't build without people adding packages to their local workspace? Or should we be holding until things propagate through proper release/build?

tylerjw:
We do plan on cutting a release of moveit_resources quickly to make this very temporary. Then once that and the ros2_control change has propagated to the main ros apt repos we can remove the entries from the repos file I added here.

tylerjw:
I reran the humble-ci-testing job because the failure there is the flaky test in the rdf loader: https://github.com/ros-planning/moveit2/issues/1334

PR #1424, opened 2022-07-11
WIP: Hybrid planning manager doesn't inherit from node

### LINKED ISSUE #1013: HybridPlanningManager should not inherit from Node
See `moveit_ros/hybrid_planning/.../hybrid_planning_manager.h`

Discussion here.  https://github.com/ros-planning/moveit2/pull/980#discussion_r787230230

This likely causes some initialization issues.

henrygerardmoore:
### Description

Changes hybrid planning manager from a node child class to having a node. Closes #1013 

### Checklist


henrygerardmoore:
@AndyZe are there tests to run that can demonstrate if this change works correctly? I noticed that the `test_basic_integration` test in `moveit_hybrid_planning` is disabled at the moment.

AndyZe:
You should check that the r still runs:. https://moveit.picknik.ai/main/doc/examples/hybrid_planning/hybrid_planning_tutorial.html

I don't remember why the integration test is connected but it would be cool if you try it again, too

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/include/moveit/hybrid_planning_manager/hybrid_planning_manager.h
```diff
@@ -76,7 +76,12 @@ class HybridPlanningManager : public rclcpp::Node
    */
   std::shared_ptr<HybridPlanningManager> shared_from_this()
   {
-    return std::static_pointer_cast<HybridPlanningManager>(Node::shared_from_this());
+    return std::static_pointer_cast<HybridPlanningManager>(shared_from_this());
+  }
```
Abishalini, line 80: Since now the class inherits from `std::enable_shared_from_this<HybridPlanningManager>`, do we need this function?

moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/include/moveit/hybrid_planning_manager/hybrid_planning_manager.h
```diff
@@ -54,14 +54,14 @@ namespace moveit::hybrid_planning
 /**
  * Class HybridPlanningManager - ROS 2 component node that implements the hybrid planning manager.
  */
-class HybridPlanningManager : public rclcpp::Node
```
tylerjw, line 57: Inheriting from `enable_shared_from_this` might be worse than inheriting from `Node`. Why do you need to do this?

tylerjw: There is a way to get around that clang-tidy check: https://github.com/ros-planning/moveit2/blob/2de48a7aafacd45d4f9d4c667ff7337b1da31ddd/moveit_ros/moveit_servo/include/moveit_servo/servo_node.h#L53

```
if (!planner_logic_instance_->initialize(HybridPlanningManager::shared_from_this()))
```
This is a part of the Hybrid planner design I'd like to find a way to get rid of. This circular relationship between the manager and the planners seems unnecessary to me.

PR #1434, opened 2022-07-19
Use kinematics plugin instead of inverse Jacobian for servo IK

wyattrees:
### Description

This changes servo to utilize a kinematics plugin if one is defined. If one is not present or doesn't support the group, default to inverse Jacobian

Also updated servo_example launch file to use MoveItConfigsBuilder.

To test, the current moveit_servo tutorial can be used as is. The modification of the servo_example launch file in this repository also includes a reference to kinematics.yaml for the servo node.

The kinematics plugin used by the servo node is defined via ROS parameters, which may be loaded from a yaml file. If using `MoveItConfigsBuilder` in your servo launch file (as is done in `servo_example.launch.py` in this PR), you can pass the necessary parameters as such:

```python
moveit_config = (
        MoveItConfigsBuilder("moveit_resources_panda")
        .robot_description(file_path="config/panda.urdf.xacro")
        .to_moveit_configs()
    )

servo_node = Node(
        package="moveit_servo",
        executable="servo_node_main",
        parameters=[
            servo_params,
            moveit_config.robot_description_kinematics, # here is where kinematics plugin parameters are passed
        ],
```

In the above example, the `kinematics.yaml` file is taken from the `moveit_resources` repository in the workspace, specifically `moveit_resources/panda_moveit_config/config/kinematics.yaml`. By default, this yaml file defines the KDL plugin to be used. To use `bio_ik` as the solver, first clone the `ros2` branch of [the repository](https://github.com/PickNikRobotics/bio_ik) into your workspace. Then edit the above yaml file to contain the following:

```yaml
panda_arm:
  kinematics_solver: bio_ik/BioIKKinematicsPlugin
  kinematics_solver_search_resolution: 0.005
  kinematics_solver_timeout: 0.05 # this timeout parameter is actually ignored by servo
  kinematics_solver_attempts: 1.0
  mode: gd_c # see bio_ik for other modes, but gd_c (gradient descent) has the best performance
```
The timeout parameter is ignored for servo, as it uses a fraction of the publish period instead.

The actual ROS parameter names that get passed by loading the yaml file are of the form `robot_description_kinematics.<group_name>.<param_name>`, e.g. `robot_description_kinematics.panda_arm.kinematics_solver`

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/launch/servo_example.launch.py
+++ b/moveit_ros/moveit_servo/launch/servo_example.launch.py
@@ -7,6 +7,7 @@ from launch_ros.actions import ComposableNodeContainer
 from launch_ros.descriptions import ComposableNode
 from launch.actions import ExecuteProcess
 import xacro
+from moveit_configs_utils import MoveItConfigsBuilder
 
 
 def load_file(package_name, file_path):
@@ -32,26 +33,15 @@ def load_yaml(package_name, file_path):
 
 
 def generate_launch_description():
+    moveit_config = (
+        MoveItConfigsBuilder("moveit_resources_panda")
+        .robot_description(file_path="config/panda.urdf.xacro")
+        .to_moveit_configs()
+    )
+
     # Get parameters for the Servo node
     servo_yaml = load_yaml("moveit_servo", "config/panda_simulated_config.yaml")
     servo_params = {"moveit_servo": servo_yaml}
-
-    # Get URDF and SRDF
-    robot_description_config = xacro.process_file(
-        os.path.join(
-            get_package_share_directory("moveit_resources_panda_moveit_config"),
-            "config",
-            "panda.urdf.xacro",
-        )
-    )
-    robot_description = {"robot_description": robot_description_config.toxml()}
-
-    robot_description_semantic_config = load_file(
-        "moveit_resources_panda_moveit_config", "config/panda.srdf"
-    )
-    robot_description_semantic = {
-        "robot_description_semantic": robot_description_semantic_config
-    }
 
     # RViz
     rviz_config_file = (
@@ -63,7 +53,10 @@ def generate_launch_description():
         name="rviz2",
         output="log",
         arguments=["-d", rviz_config_file],
-        parameters=[robot_description, robot_description_semantic],
+        parameters=[
+            moveit_config.robot_description,
+            moveit_config.robot_description_semantic,
+        ],
     )
 
     # ros2_control using FakeSystem as hardware
@@ -75,7 +68,7 @@ def generate_launch_description():
     ros2_control_node = Node(
         package="controller_manager",
         executable="ros2_control_node",
-        parameters=[robot_description, ros2_controllers_path],
+        parameters=[moveit_config.robot_description, ros2_controllers_path],
         output="screen",
     )
 
@@ -112,15 +105,15 @@ def generate_launch_description():
             #     name="servo_server",
             #     parameters=[
             #         servo_params,
-            #         robot_description,
-            #         robot_description_semantic,
+            #         moveit_config.robot_description,
+            #         moveit_config.robot_description_semantic,
             #     ],
             # ),
             ComposableNode(
                 package="robot_state_publisher",
                 plugin="robot_state_publisher::RobotStatePublisher",
                 name="robot_state_publisher",
-                parameters=[robot_description],
+                parameters=[moveit_config.robot_description],
             ),
             ComposableNode(
                 package="tf2_ros",
@@ -148,8 +141,9 @@ def generate_launch_description():
         executable="servo_node_main",
         parameters=[
             servo_params,
-            robot_description,
-            robot_description_semantic,
+            moveit_config.robot_description,
+            moveit_config.robot_description_semantic,
+            moveit_config.robot_description_kinematics,
         ],
         output="screen",
     )
--- a/moveit_ros/moveit_servo/src/servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/src/servo_calcs.cpp
@@ -271,6 +271,11 @@ void ServoCalcs::start()
                            current_state_->getGlobalLinkTransform(parameters_->ee_frame_name);
   tf_moveit_to_robot_cmd_frame_ = current_state_->getGlobalLinkTransform(parameters_->planning_frame).inverse() *
                                   current_state_->getGlobalLinkTransform(robot_link_command_frame_);
+  if (!use_inv_jacobian_)
+  {
+    ik_base_to_tip_frame_ = current_state_->getGlobalLinkTransform(ik_solver_->getBaseFrame()).inverse() *
+                            current_state_->getGlobalLinkTransform(ik_solver_->getTipFrame());
+  }
 
   stop_requested_ = false;
   thread_ = std::thread([this] { mainCalcLoop(); });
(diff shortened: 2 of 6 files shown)


wyattrees:
@AndyZe I've added more to the description showing how to change the kinematics plugin.

moveit_ros/moveit_servo/launch/servo_example.launch.py
```diff
@@ -112,15 +105,15 @@ def generate_launch_description():
             #     name="servo_server",
             #     parameters=[
             #         servo_params,
-            #         robot_description,
-            #         robot_description_semantic,
+            #         moveit_config.robot_description,
+            #         moveit_config.robot_description_semantic,
```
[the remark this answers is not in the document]
wyattrees: Sure thing -- is it commented out because, as of now, it doesn't work as a `ComposeableNode`? Should it just be removed (in a separate PR)?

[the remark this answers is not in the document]
AndyZe: I think somebody should test if it works as a `ComposableNode` yet, then delete it if not. Doesn't have to be you right now.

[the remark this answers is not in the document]
AndyZe: issue #1441

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -576,7 +595,53 @@ bool ServoCalcs::cartesianServoCalcs(geometry_msgs::msg::TwistStamped& cmd,
...
   Eigen::MatrixXd pseudo_inverse = svd.matrixV() * matrix_s.inverse() * svd.matrixU().transpose();
 
-  delta_theta_ = pseudo_inverse * delta_x;
+  // Convert from cartesian commands to joint commands
+  // Use an IK solver plugin if we have one, otherwise use inverse Jacobian.
+  if (!use_inv_jacobian_)
+  {
+    // get a transformation matrix with the desired position change &
+    // get a transformation matrix with desired orientation change
+    Eigen::Isometry3d tf_pos_delta(Eigen::Isometry3d::Identity());
```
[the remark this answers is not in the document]
AndyZe: We discussed this and `static` doesn't fit here because it needs to be initialized to the identity matrix with every iteration.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -576,7 +595,53 @@ bool ServoCalcs::cartesianServoCalcs(geometry_msgs::msg::TwistStamped& cmd,
...
+    // Poses passed to IK solvers are assumed to be in the tip link (EE) reference frame
+    // So, apply the base->end effector transform to the delta transform
+    // apply rotation after so that it's local to EE
+    auto full_tf = tf_pos_delta * tf_moveit_to_ee_frame_ * tf_rot_delta;
+    geometry_msgs::msg::Pose next_pose = tf2::toMsg(full_tf);
+
+    // setup for IK call
+    static std::vector<double> solution(num_joints_, 0.0);
+    static moveit_msgs::msg::MoveItErrorCodes err;
+    static kinematics::KinematicsQueryOptions opts;
```
AndyZe, line 622: Sorry to backtrack. From reading here, `static` does not really help in terms of realtime performance. Please remove `static`.  :P

https://stackoverflow.com/a/3835902

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -576,7 +595,53 @@ bool ServoCalcs::cartesianServoCalcs(geometry_msgs::msg::TwistStamped& cmd,
...
+    opts.return_approximate_solution = true;
+    if (ik_solver_->searchPositionIK(next_pose, internal_joint_state_.position, parameters_->publish_period / 2.0,
+                                     solution, err, opts))
+    {
+      // find the difference in joint positions that will get us to the desired pose
+      for (size_t i = 0; i < num_joints_; i++)
+      {
+        delta_theta_[i] = solution[i] - internal_joint_state_.position[i];
```
AndyZe, line 630:
```suggestion
        delta_theta_.at(i) = solution.at(i) - internal_joint_state_.position.at(i);
```

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -576,7 +595,53 @@ bool ServoCalcs::cartesianServoCalcs(geometry_msgs::msg::TwistStamped& cmd,
...
+    // setup for IK call
+    static std::vector<double> solution(num_joints_, 0.0);
+    static moveit_msgs::msg::MoveItErrorCodes err;
+    static kinematics::KinematicsQueryOptions opts;
+    opts.return_approximate_solution = true;
+    if (ik_solver_->searchPositionIK(next_pose, internal_joint_state_.position, parameters_->publish_period / 2.0,
+                                     solution, err, opts))
+    {
+      // find the difference in joint positions that will get us to the desired pose
+      for (size_t i = 0; i < num_joints_; i++)
```
AndyZe, line 628:
```suggestion
      for (size_t i = 0; i < num_joints_; ++i)
```

AndyZe (commented):
Sorry for these very nitpicky changes. LGTM after this.

AndyZe (approved):

wyattrees:
Found a bug in how rotations are applied while doing some singularity avoidance testing, we should hold off on merging for now.

wyattrees:
@AndyZe I've fixed an issue where rotations were always being applied in the same frame--would you mind taking a second look at the recent changes (lines 613-633)?

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -576,7 +595,70 @@ bool ServoCalcs::cartesianServoCalcs(geometry_msgs::msg::TwistStamped& cmd,
...
+    Eigen::Isometry3d tf_pos_delta(Eigen::Isometry3d::Identity());
+    tf_pos_delta.translate(Eigen::Vector3d(delta_x[0], delta_x[1], delta_x[2]));
+
+    Eigen::Isometry3d tf_rot_delta(Eigen::Isometry3d::Identity());
+    Eigen::Quaterniond q = Eigen::AngleAxisd(delta_x[3], Eigen::Vector3d::UnitX()) *
+                           Eigen::AngleAxisd(delta_x[4], Eigen::Vector3d::UnitY()) *
+                           Eigen::AngleAxisd(delta_x[5], Eigen::Vector3d::UnitZ());
+    tf_rot_delta.rotate(q);
+
+    // Poses passed to IK solvers are assumed to be in the tip link (EE) reference frame
```
AndyZe, line 613: I think these comments ought to be more generic. It's not necessarily the tip link reference frame, but rather whatever the IK reference frame is set to, right?

Otherwise the feature of switching from World to EE frame as the user's frame of reference (mentioned in [the tutorial](https://moveit.picknik.ai/humble/doc/examples/realtime_servo/realtime_servo_tutorial.html)) would not work.

wyattrees: Yeah, it really should be more generic -- but it's actually more than just the comment. Really, we shouldn't just blindly use `tf_moveit_to_ee_frame_`, but rather a transform from what the IK solver knows as the base frame -> what the IK solver knows as the tip frame (which is likely the same as what servo knows as moveit -> ee frame, but not necessarily). 

Switching of the reference frame for the incoming twist commands is already handled above in the same function, so either way it should work for any incoming frame of reference.

See the latest commit for my solution to this

AndyZe (approved):

Issue #1442, opened 2022-07-21
Error while installing ros-humble-pilz-industrial-motion-planner binary


Yadunund:
### Description

Installing moveit2 from binaries on `humble` + `Ubuntu 22.04` seems to throw unmet dependency error. Specifically due to `ros-humble-pilz-industrial-motion-planner`

### Your environment
* ROS Distro: Humble
* OS Version: Ubuntu 22.04
* Binary install
* `2.5.1-1jammy.20220625.084800_amd64`

### Steps to reproduce
Tell us how to reproduce this issue. Attempt to provide a working demo, perhaps using Docker.
```
sudo apt update && sudo apt upgrade -y
sudo apt install ros-humble-moveit
```

### Expected behaviour
All binaries install without any errors

### Actual behaviour
```
You might want to run 'apt --fix-broken install' to correct these.
The following packages have unmet dependencies:
 ros-humble-moveit-planners : Depends: ros-humble-pilz-industrial-motion-planner but it is not going to be installed
E: Unmet dependencies. Try 'apt --fix-broken install' with no packages (or specify a solution).
ERROR: the following rosdeps failed to install
  apt: command [sudo -H apt-get install -y ros-humble-moveit] failed
  apt: command [sudo -H apt-get install -y ros-humble-moveit-resources-panda-moveit-config] failed
  apt: command [sudo -H apt-get install -y ros-humble-moveit-configs-utils] failed
  apt: command [sudo -H apt-get install -y ros-humble-control-msgs] failed
  apt: command [sudo -H apt-get install -y ros-humble-moveit-core] failed
  apt: command [sudo -H apt-get install -y ros-humble-moveit-ros-perception] failed
  apt: command [sudo -H apt-get install -y ros-humble-moveit-ros-planning-interface] failed
  apt: command [sudo -H apt-get install -y ros-humble-moveit-servo] failed
  apt: command [sudo -H apt-get install -y ros-humble-moveit-hybrid-planning] failed
  apt: command [sudo -H apt-get install -y ros-humble-moveit-visual-tools] failed
  apt: command [sudo -H apt-get install -y ros-humble-rviz-visual-tools] failed
  apt: command [sudo -H apt-get install -y ros-humble-geometric-shapes] failed
  apt: command [sudo -H apt-get install -y ros-humble-moveit-common] failed
  apt: command [sudo -H apt-get install -y ros-humble-moveit-ros-planning] failed
[...]

PR #1443, opened 2022-07-21
rename header files so debs are installable

mikeferguson:
### Description

These files were copied in, now causing installation issues for our Debians: https://github.com/ros-planning/moveit2/issues/1442

### Checklist


mikeferguson:
So it turns out this is more involved than I thought - the files we copied in are not what eventually got merged - in particular support for "namespace" has been lost - so our options are either:

 A) Remove the namespace usage within our codebase
 B) Keep the copied files, but rename them so we don't collide with the Debian
 C) Get an updated version into ros2_control, get that released, and then release this change

I'm not actually sure what option A even entails at this moment

tylerjw (approved):

mikeferguson:
This now implements option B

vatanaksoytezer:
@mikeferguson are we planning to eventually switch to option C, or keep it this way?

mikeferguson:
@vatanaksoytezer I think we should eventually do option C (and I'll make sure there is a ticket for it) - but that is going to require at least 2 release cycles and somebody who is actually active on ros2_control (which I'm not). My goal with this PR is simply to get our debs to properly install - we can clean up the tech debt later

vatanaksoytezer (approved):

vatanaksoytezer:
I'll try to cut a new release today.

mikeferguson:
Ticketed Option C here: https://github.com/ros-planning/moveit2/issues/1451

Issue #1279, opened 2022-05-26
GSoC: Create Python bindings for MoveIt 2 


peterdavidfagan:
Hi there, 

I am currently working on creating Python bindings for MoveIt 2 as part of my [GSoC project](https://summerofcode.withgoogle.com/programs/2022/projects/AICLMf2p).

The goals of my project are to:

1. Redesign and implement a Python API for MoveIt 2
2. Create a tutorial demonstrating the usage of this API for learning-based approaches
3. Implement feature requests that come from the MoveIt community

I will be adding further details to this issue page and my project page over time. Feel free to comment below or link feature requests that you feel may be relevant to this project. 

Have a great rest of the day.

peterdavidfagan:
Hi @tlpss, 

Thank you for these points. I have taken note of these and will look to incorporate the outlined functionalities into the Python API that gets created.  It would be great to have you test the initial library. 

Currently we are deciding what to bind and how the API should be structured. At the moment it looks like the functionality of the `moveit_cpp` library will be binded. The overall API design is still not finalized. I hope to provide demos using Isaac Sim.

tlpss:
> It would be great to have you test the initial library that results from this project.

Feel free to ping me, I would be interested to take a look. 

> Currently we are deciding what to bind and how the API should be structured. At the moment it looks like the functionality of the moveit_cpp library will be binded. The overall API design is still not finalized. I hope to provide demos using Isaac Sim so maybe this will be helpful for your third point.

Sounds great,in addition to ISAAC sim, Unity might also be an interesting choice as it is completely open source and does not require a beefy GPU (ISAAC sim requires an RTX GPU if I'm not mistaken). 

Best of luck with the project!

peterdavidfagan:
Hi everyone, 

The development of a library that is called `moveit_py` has begun; initially this library will consist of binding functionalities of `moveit_cpp` and creating useful python interfaces using these functionalities. As `moveit_cpp` continues to develop this library will develop alongside it. For those who are interested in tracking the development work for this library the [moveit_py](https://github.com/peterdavidfagan/moveit2/tree/moveit_py) branch on my fork should remain up to date.

I also have another repository [moveit_py_example](https://github.com/peterdavidfagan/moveit_py_example
) which will demonstrate examples of using the library. There's only one example there for the moment but I will be adding many more in the coming weeks (including isaac-sim examples which I have from work undertaken while onboarding). 

[...]

PR #1446, opened 2022-07-24
Refactor and open the Python interface of planning_interface module

larrylu0426:
Abstract:
Replace the Python binding method based on boost_python in the existed code with pybind11. 

Main Idea:
To keep the existed function no changed, and use the inheritance method(seen in move_group_interface.cpp) to wrap the existed C++ APIs.

Contributions:
1. Replace boost python with pybind11, and keep the name and style of existed APIs as stable as possible;
2. Adapt the old ROS1-related codes;
3. Open both C++  and Python APIs about robot interface;
4. Open and refactor the Python APIs about rclcpp interface/move_group interface/planning_scene interface.


larrylu0426:
I'm working on refactoring MoveIt-commander. To keep the git's log more clear, I will update them together :)

vatanaksoytezer:
Hey @larrylu0426 thanks a lot for all the work! Have you checked @peterdavidfagan 's work on https://github.com/ros-planning/moveit2/issues/1279 it might be similar / redundant. So if you can sync up and check with him I think that would be very beneficial.

larrylu0426:
Hi @vatanaksoytezer , I have seen Peter's repo. These two works are for the different goals. This work is to adapt existed functions based on the remained codes. I want to keep the original project architecture. I also have seen that there are some usages of pybind11 in the moveit-core module, so I just want to follow this progress for providing more complete python adaptions. Peter is doing a great change for the whole moveit2. While I don't know when it will be released. So if an available python adaption is offered, it's good for some Python developers.

PR #1454, opened 2022-07-28
Organize package xml of MSA

### LINKED ISSUE #1452: MSA package xmls
When I install moveit_setup_assistant via apt, it doesn't install all the dependencies namely `setup-controller`, `setup-app-plugins` and `setup-core-plugins`. We need add to change package.xml of MSA to include these.

vatanaksoytezer:
Resolves #1452


tylerjw (approved):
Looks good to me.

PR #1464, opened 2022-07-31
Servo: check for and enable a realtime kernel

AndyZe:
### Description

Servo automatically checks if a realtime kernel is installed and sets a higher thread priority if one is found. This is basically a port of this `ros2_control` PR:  https://github.com/ros-controls/ros2_control/pull/748

I started a tutorial PR to document this here:

https://github.com/ros-planning/moveit2_tutorials/pull/478


moveit_ros/moveit_servo/src/realtime.cpp
```diff
@@ -0,0 +1,43 @@
...
+  return has_realtime;
+}
+
+bool configure_sched_fifo(int priority)
+{
+  struct sched_param schedp;
+  memset(&schedp, 0, sizeof(schedp));
+  schedp.sched_priority = priority;
+  return !sched_setscheduler(0, SCHED_FIFO, &schedp);
+}
```
tylerjw, line 41: Please don't copy-paste this. I put these in functions so you can just call them from other places. It is just fine to depend on and link to ros2_control::controller_manager from here.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -88,6 +85,15 @@ geometry_msgs::msg::TransformStamped convertIsometryToTransform(const Eigen::Iso
...
 }
+
+static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_servo.servo_calcs");
+constexpr auto ROS_LOG_THROTTLE_PERIOD = std::chrono::milliseconds(3000).count();
+static constexpr double STOPPED_VELOCITY_EPS = 1e-4;  // rad/s
+
+// This value is used when configuring the main loop to use SCHED_FIFO scheduling
+// We use a midpoint RT priority to allow maximum flexibility to users
+// Reference: https://man7.org/linux/man-pages/man2/sched_setparam.2.html
+int const THREAD_PRIORITY = 50;
```
tylerjw, line 96: Suggest picking a lower number than the one we chose for ros2_control to avoid higher variance in the jitter that will be caused by the schedule randomly choosing this or controller_manager's thread to run if both are runnable. I would probably choose 40 to have lower number but leave some numbers between this and controller_manger.

moveit_ros/moveit_servo/src/realtime.cpp
```diff
@@ -0,0 +1,43 @@
+// Copyright 2022 PickNik Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
```
tylerjw, line 3: My other comment pointed out that this shouldn't be copied. But if you have to copy it, you'll need to update the license file and copyright notices (at the project's root) because this is Apache, and the rest of MoveIt is BSD. The are compatible license so both can exist in a project, but you will need to make sure they are both called out at the root of the project correctly. One thing you must do to comply with is place a copy of the license file itself in the repo, not just this notice.

tylerjw:
Please don't copy-paste code when you could just link against controller_manager and call these functions.

moveit_ros/moveit_servo/src/servo_calcs.cpp
AndyZe: done

moveit_ros/moveit_servo/src/realtime.cpp
AndyZe: Good idea. Do we want to link against the `controller_manager` library when we just need this little bit of functionality, though? It would be nice if `realtime` were in its own library.

Anyway, done

moveit_ros/moveit_servo/CMakeLists.txt
```diff
@@ -72,15 +73,14 @@ ament_target_dependencies(${SERVO_PARAM_LIB_NAME} ${THIS_PACKAGE_INCLUDE_DEPENDS
...
-  # These files are used to produce differential motion
   src/collision_check.cpp
   src/enforce_limits.cpp
   src/servo.cpp
   src/servo_calcs.cpp
 )
 set_target_properties(${SERVO_LIB_NAME} PROPERTIES VERSION "${${PROJECT_NAME}_VERSION}")
 ament_target_dependencies(${SERVO_LIB_NAME} ${THIS_PACKAGE_INCLUDE_DEPENDS})
-target_link_libraries(${SERVO_LIB_NAME} ${SERVO_PARAM_LIB_NAME})
+target_link_libraries(${SERVO_LIB_NAME} ${SERVO_PARAM_LIB_NAME} )
```
tylerjw, line 83: It looks like it is failing the formatting check for cmake in CI for this extra space here.

tylerjw (approved):
Looks good other than failing CI for a formatting issue introduced into the cmake file.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -241,6 +247,19 @@ void ServoCalcs::start()
   // Stop the thread if we are currently running
   stop();
 
+  // Check if a realtime kernel is installed. Set a higher thread priority, if so
+  if (controller_manager::has_realtime_kernel())
+  {
+    if (!controller_manager::configure_sched_fifo(THREAD_PRIORITY))
```
wyattrees, line 253: Shouldn't setting the thread priority happen inside the thread where all the servo calculations are happening (i.e. [here](https://github.com/ros-planning/moveit2/blob/13e8bc31b4e4495d372ab28c26f38038592df702/moveit_ros/moveit_servo/src/servo_calcs.cpp#L300))? Currently this would set the priority for whatever thread called `ServoCalcs::start`

tylerjw: Good catch, it should be done there. Please update the PR to change which thread is being configured to be the one that runs the main control loop of servo.

Issue #1455, opened 2022-07-28
Clean up PlanningComponent API


tylerjw:
This class: https://github.com/ros-planning/moveit2/blob/main/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h

This class builds a `::planning_interface::MotionPlanRequest` through a bunch of setters that modify the internal state of `PlanningComponent` and then the `plan()` function combines those into the request message and sends it and handles the response.

The API for building a `MotionPlanRequest` should be more composable and flexible to make it more extensible and usable and not tied to the action of sending the request and handing the response.

Goals of this project:
* Remove redundant overloads
* Support python bindings for manipulating and reasoning about the `MotionPlanRequest`
* Separate code that performs actions that mutate the global state of the system (like requesting a plan from another process) and code that reasons about the construction of the data in the request
* Leaner interfaces that remove the tight coupling
* Enable concurrent planning and other more advanced implementations of how this message is used

As an API breakage, this change only targets Rolling and could be backported to Humble only afterward. We should consider tick-tocking the old API.

tylerjw:
These are my **DRAFT** notes about this topic. I will update this as I have more time to read and think about how this interface should be designed.

## Actions that result in Data that can be made a separate independent thing
- Planning pipeline parameter loading should be separate from the constructor of PlanningComponent and passed in as a parameter.

## Methods that could just be functions
- getNamedTargetStates uses group_name from constructor to get joint_model_group from RobotModel (read from moveitcpp) and then calls joint_model_group_->getDefaultStateNames and returns the result.

```cpp
std::vector<std::string> get_named_target_states(auto robot_model, auto group_name) {
  if (auto jmg = robot_model->getJointModelGroup(group_name);
      jmg != nullptr) {
    return jmg->getDefaultStateNames();
  }
  return {};
}
```

This should maybe return a std::optional as there is a difference between a group_name existing and having no named states and a group_name not existing.

```cpp
std::optional<std::vector<std::string>>> get_named_target_states(
  auto robot_model, auto group_name) {
  if (auto jmg = robot_model->getJointModelGroup(group_name);
      jmg != nullptr) {
    return jmg->getDefaultStateNames();
  }
  return std::nullopt;
}
```

Another way to look at it is this calls a function on joint_model_group and returns the result if joint_model_group is not a nullptr. Pointers are a type of monads as they can point to memory or nullptr.

```cpp
template<typename T, typename F, typename Ret>
[...]

PR #1470, opened 2022-08-02
Default destructor for PlanningComponent

tylerjw:
Signed-off-by: Tyler Weaver <tyler@picknik.ai>

### Description

While working on https://github.com/ros-planning/moveit2/issues/1455 I noticed two things that seem wrong here.

1. The move constructor should be deleted, not set to default.
2. The destructor implementation is completely extra, default destruction will work just fine and reduces code.

### Checklist


moveit_ros/planning/moveit_cpp/src/planning_component.cpp
```diff
@@ -81,12 +81,6 @@ PlanningComponent::PlanningComponent(const std::string& group_name, const rclcpp
   planning_pipeline_names_ = moveit_cpp_->getPlanningPipelineNames(group_name);
 }
 
-PlanningComponent::~PlanningComponent()
-{
-  RCLCPP_INFO(LOGGER, "Deleting PlanningComponent '%s'", group_name_.c_str());
-  clearContents();
```
henningkayser, line 87: Would you mind removing the clearContents() function as well?

tylerjw: Done, pushed that change. Calling that function could break all sorts of things in this class.

tylerjw: Done, pushed that change. Calling that function could break all sorts of things in this class.

tylerjw: Done, pushed that change. Calling that function could break all sorts of things in this class.

vatanaksoytezer (approved):

henningkayser (approved):

Issue #1478, opened 2022-08-05
Floating joint interpolation takes the long way around


werner291:
Hi, I believe I found a bug in the joint value interpolation code for floating joints.

See the snippet below:

```
const double jv1[] = {0,0,0, 0, 0, 0.862044, 0.506833};
const double jv2[] = { 0,0,0, 0, 0, -0.969825, 0.243803 };

double intp[7];

moveit::core::FloatingJointModel fjm("joint", 0, 0);

fjm.interpolate(jv1, jv2, 0.5, intp);

Eigen::Isometry3d tf;

fjm.computeTransform(jv1, tf);
Eigen::AngleAxisd rotA(tf.rotation());

fjm.computeTransform(jv2, tf);
Eigen::AngleAxisd rotB(tf.rotation());

fjm.computeTransform(intp, tf);
Eigen::AngleAxisd rotC(tf.rotation());

std::cout << "AA: " << rotA.angle() << " " << rotA.axis().transpose() << std::endl; // 2.07858 0 0 1
std::cout << "BB: " << rotB.angle() << " " << rotB.axis().transpose() << std::endl; // 2.64902 -0 -0 -1
std::cout << "CC: " << rotC.angle() << " " << rotC.axis().transpose() << std::endl; // 0.285223  0  0 -1
```

Note how the interpolated angle has value 0.285: it's taking the long way around, with the interpolation going through 0 instead of Pi/-Pi radians.

werner291:
I believe this line (https://github.com/ros-planning/moveit2/blob/main/moveit_core/robot_model/src/floating_joint_model.cpp#L146)  is supposed to prevent exactly this case, but evidently it does not appear to be working quite right.

werner291:
Wait a minute... how is `dq < 0` ever going to happen if you take the absolute value earlier?

werner291:
I think the formula got copied wrong... I'mma try moving that `fabs` around to see what happens.

werner291:
PR should fix the issue, I threw in a test as well.

henningkayser:
PR is merged, so I'm closing this

PR #1479, opened 2022-08-05
Fixed fabs() use in quaternion interpolation

werner291:
### Description

See issue #1478, it appears that the formula got copied wrong from Wikipedia, as applying the absolute-value function in the place it was at prevented detection of the long-way-around case.

The new code seems to work for the case mentioned in the issue, I'll see if I can test it a bit more thoroughly.

### Checklist


moveit_core/robot_model/test/test.cpp
```diff
@@ -128,6 +128,50 @@ TEST(SiblingAssociateLinks, SimpleYRobot)
...
+{
+  // Create a simple floating joint model with some dummy parameters (these are not used by the test)
+  moveit::core::FloatingJointModel fjm("joint", 0, 0);
+
+  double jv1[7];
+  double jv2[7];
+  double intp[7];
+  random_numbers::RandomNumberGenerator rng;
+
+  for (size_t i = 0; i < 1000; i++)
```
AndyZe, line 141:
```suggestion
  for (size_t i = 0; i < 1000; ++i)
```

Pre-increment is slightly preferred for speed (although the compiler _probably_ takes care of it). Reference:  https://betterprogramming.pub/stop-using-i-in-your-loops-1f906520d548

moveit_core/robot_model/test/test.cpp
```diff
@@ -128,6 +128,50 @@ TEST(SiblingAssociateLinks, SimpleYRobot)
...
+    // We set some bounds where the joint position's translation component is bounded between -1 and 1 in all
+    // dimensions. This is necessary, otherwise we just get (0,0,0) translations.
+    moveit::core::JointModel::Bounds bounds;
+    bounds = fjm.getVariableBounds();
+    bounds[0].min_position_ = -1.0;
+    bounds[0].max_position_ = 1.0;
+    bounds[1].min_position_ = -1.0;
+    bounds[1].max_position_ = 1.0;
+    bounds[2].min_position_ = -1.0;
+    bounds[2].max_position_ = 1.0;
```
AndyZe, line 152: I think you could move setting of the bounds outside of the loop

AndyZe (commented):
Some quick comments on the code. I'll come back to do my best to check the quaternion math

moveit_core/robot_model/src/floating_joint_model.cpp
```diff
@@ -136,8 +136,9 @@ void FloatingJointModel::interpolate(const double* from, const double* to, const
   state[1] = from[1] + (to[1] - from[1]) * t;
   state[2] = from[2] + (to[2] - from[2]) * t;
 
-  double dq = fabs(from[3] * to[3] + from[4] * to[4] + from[5] * to[5] + from[6] * to[6]);
-  double theta = (dq + std::numeric_limits<double>::epsilon() >= 1.0) ? 0.0 : acos(dq);
+  double dq = from[3] * to[3] + from[4] * to[4] + from[5] * to[5] + from[6] * to[6];
+  double theta = (fabs(dq) + std::numeric_limits<double>::epsilon() >= 1.0) ? 0.0 : acos(fabs(dq));
```
AndyZe, line 140: Isn't there a factor of 2 missing here?

[This website](https://www.euclideanspace.com/maths/algebra/realNormedAlgebra/quaternions/slerp/index.htm) gives this formula, note the 2:

`cos(theta/2) = qa.w*qb.w + qa.x*qb.x + qa.y*qb.y+ qa.z*qb.z`

henningkayser: I'm not sure if it's really missing, the implementation matches the example code on the website you linked. `theta` could probably be renamed to `halfTheta`, though.

moveit_core/robot_model/test/test.cpp
```diff
@@ -128,6 +128,50 @@ TEST(SiblingAssociateLinks, SimpleYRobot)
...
 
+TEST(FloatingJointTest, interpolation_test)
+{
+  // Create a simple floating joint model with some dummy parameters (these are not used by the test)
+  moveit::core::FloatingJointModel fjm("joint", 0, 0);
+
+  double jv1[7];
+  double jv2[7];
+  double intp[7];
+  random_numbers::RandomNumberGenerator rng;
```
henningkayser, line 139: There is currently some ongoing refactoring around RNG, see #1267. @sjahr any thoughts if we should merge this and refactor later?

henningkayser (commented):
Thank you for the fix, it looks correct to me. However, I somewhat agree with Andy that it would be cleaner to use another library for that. How about simply calling [Eigen::Quaternion::slerp()](https://eigen.tuxfamily.org/dox/classEigen_1_1QuaternionBase.html#ac840bde67d22f2deca330561c65d144e)?

moveit_core/robot_model/src/floating_joint_model.cpp
werner291: The factor 2 comes from converting between axis-angle and quaternion. It's because quaternions are weird: you don't just multiply them with the thing you want to transform (though that's how most linalg libraries typically have their API set up) you pre-multiply with the quaternion, then post-multiply with the inverse quaternion.

Also, I don't see how the test could possible be passing if I was missing something as big as a factor 2 in there.

moveit_core/robot_model/test/test.cpp
werner291: Looks good to me. I put it up near the joint model allocation, fits in better anyway.

moveit_core/robot_model/test/test.cpp
werner291: It's not half-theta at all, that really is the angle between the two quaternions. The halving only shows up when you when you convert between Quaternions and axis-angle stuff to my knowledge

werner291: 100% agree that just calling into Eigen would be way easier than DIY-ing the math ourselves; I just didn't feel confident suggesting a larger rewrite than was necessary to fix the bug.

AndyZe: I think we can merge this now b/c it's an obvious improvement. Thanks! especially for the new test

I'm still not sure I agree on the factor of 2 though. We can make an issue and figure it out later.

AndyZe (approved):

moveit_core/robot_model/test/test.cpp
werner291: Oh I think I misread your website. Lemme take a look at what they're doing, Quaternions are tricky

werner291: Ok, find a source here that features the *2 as well https://nl.mathworks.com/matlabcentral/answers/415936-angle-between-2-quaternions

werner291: Here's a *fantastic* introduction to quaternions that only *slightly* breaks my brain: https://eater.net/quaternions

werner291: https://math.stackexchange.com/questions/4053346/the-angle-between-two-quaternion-vectors-is-equal-to-half-the-angle-between-thei

> that given two quaternions q1,q2∈R4, the angle between the 4D quaternion vectors is equal to half the angle between their corresponding 3D orientations.

Oh my god the angle is half the angle... I think this is a definitions issue. Do we want the angle between rotations, or do we want the angle between the quaternions?

On the referenced Wikipedia article: "Compute Ω as the angle [subtended](https://en.wikipedia.org/wiki/Subtended_angle) by the arc, so that cos Ω = p0 ∙ p1, the n-dimensional [dot product](https://en.wikipedia.org/wiki/Dot_product) of the unit vectors from the origin to the ends."

So (if I read this correctly) it's the angle between the vectors (quaternions are basically 4D vectors), not the angle between the rotations that the quaternions represent. Therefore, it's theta and not half-theta.

This also reinforced my belief that we really should just let Eigen handle this...

werner291: In light of that explanation, this code here (https://github.com/ros-planning/moveit2/blob/main/moveit_core/robot_model/src/floating_joint_model.cpp#L122) is also wrong (the method name explicitly mentions "rotation" and not "quaternion angle"), and is missing the factor 2.

(Doesn't change anything for our case since the slerp operation explicitly requires the angle between quaternions, not rotations)

werner291:
So I wrote a version that uses Eigen... Feel free to reject these if the changes are a bit too radical. Test shows no difference, though, and I'm a strong supporter of using Eigen instead of DIY math wherever possible.

werner291:
Oh, this line is a significant change ```
if (abs(from[3] - to[3]) + abs(from[4] - to[4]) + abs(from[5] - to[5]) + abs(from[6] - to[6]) >
      std::numeric_limits<double>::epsilon())
```

Though really, all it does is skip the quaternion math if the rotation parts are not significantly different. The original code tested whether theta was significantly > 0, so I'd argue same difference.

File: moveit_core/robot_model/src/floating_joint_model.cpp [hunk 5]
```diff
@@ -136,19 +136,19 @@ void FloatingJointModel::interpolate(const double* from, const double* to, const
   state[1] = from[1] + (to[1] - from[1]) * t;
   state[2] = from[2] + (to[2] - from[2]) * t;
 
-  double dq = fabs(from[3] * to[3] + from[4] * to[4] + from[5] * to[5] + from[6] * to[6]);
-  double theta = (dq + std::numeric_limits<double>::epsilon() >= 1.0) ? 0.0 : acos(dq);
-  if (theta > std::numeric_limits<double>::epsilon())
+  if (abs(from[3] - to[3]) + abs(from[4] - to[4]) + abs(from[5] - to[5]) + abs(from[6] - to[6]) >
```
AndyZe, line 139:
```suggestion
  // Check if the quaternions are significantly different
  if (abs(from[3] - to[3]) + abs(from[4] - to[4]) + abs(from[5] - to[5]) + abs(from[6] - to[6]) >
```

AndyZe (approved):
Thank you! I have one small documentation suggest. LGTM besides that

henningkayser (approved):

PR #1482, opened 2022-08-07
Support chained controllers

pac48:
This PR adds support for ros2_control's chained controllers. Only simple chains are supported currently. That is, controller can only chain to at most one other controller. I also added logic to start and stop the chained controller in the correct order.

I created a demo repository under ros2_control with that uses the chnages made in this PR. Here is a link to the [README](https://github.com/pac48/ros2_control_demos/blob/add-admittance-controller/using_admittance_controller.md)


v4hn:
> Only simple chains are supported currently. That is, controller can only chain to at most one other controller

Why? It does not sound like it would be too difficult to extend it?
That being said I do wonder whether it isn't ros_control's responsibility to tell you which hardware it is responsible to on top of which other controller it feeds into...

pac48:
> > Only simple chains are supported currently. That is, controller can only chain to at most one other controller
> 
> Why? It does not sound like it would be too difficult to extend it? That being said I do wonder whether it isn't ros_control's responsibility to tell you which hardware it is responsible to on top of which other controller it feeds into...

The response from the ros2_control list controllers service does indicate which controllers are chained to which. The logic in this PR is really about taking that information and converting it to a format that MoveIt understands. 

Regarding chaining multiple controllers, MoveIt's controller manager plugin assumes there exists a direct mapping from controllers to controllable interfaces. If the controllers are chained with only one connection, then it is straight forward to propagate the controllable interfaces to the beginning of the chain, which ultimately controls those interfaces. If branching occurs in the chain, then propagating the  interfaces is ambiguous.

Lastly, I believe the ros2_control maintainers are working towards a simplified service interface to handle chained controllers. When that happens, handling the general case will be more straightforward.

v4hn:
I'll leave the actual review here to people who work with ros2_control.

AndyZe:
@pac48 it would help a lot if you could add test instructions to the PR description. I think one of your ros2_control PR's added a 6-dof example... maybe we can test this PR with that.

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -109,6 +110,8 @@ class MoveItControllerManager : public moveit_controller_manager::MoveItControll
   rclcpp::Client<controller_manager_msgs::srv::ListControllers>::SharedPtr list_controllers_service_;
   rclcpp::Client<controller_manager_msgs::srv::SwitchController>::SharedPtr switch_controller_service_;
 
+  std::unordered_map<std::string, std::vector<std::string>> dependency_map_;
```
AndyZe, line 113: Does this only apply to chained controllers? May want to rename it `chained_dependencies_map_` or something similar, if so. And add comments.

```suggestion
  // Chained controllers have dependencies (other controllers which must be running)
  std::unordered_map<std::string /* controller name */, std::vector<std::string> /* dependencies */> dependency_map_;
```

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -419,6 +452,38 @@ class MoveItControllerManager : public moveit_controller_manager::MoveItControll
...
+   * are propagated back to the first controller with a non-chained input
+   */
+  bool fixChainedControllers(std::shared_ptr<controller_manager_msgs::srv::ListControllers::Response>& result)
+  {
+    std::unordered_map<std::string, controller_manager_msgs::msg::ControllerState*> controller_name_map;
+    for (auto& c : result->controller)
+    {
+      controller_name_map[c.name] = &c;
+    }
+    for (auto& c : result->controller)
```
AndyZe, line 467:
```suggestion
    for (const auto& c : result->controller)
```

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -419,6 +452,38 @@ class MoveItControllerManager : public moveit_controller_manager::MoveItControll
...
     return true;
   }
+  /**
+   * \brief fixChainedControllers modifies ListControllers service response if it contains chained controllers.
+   * Since chained controllers cannot be written to directly, they are removed from the response and their interfaces
+   * are propagated back to the first controller with a non-chained input
+   */
+  bool fixChainedControllers(std::shared_ptr<controller_manager_msgs::srv::ListControllers::Response>& result)
+  {
+    std::unordered_map<std::string, controller_manager_msgs::msg::ControllerState*> controller_name_map;
```
AndyZe, line 462: Prob should make this a unique_ptr for safety's sake.
```suggestion
    std::unordered_map<std::string, std::unique_ptr<controller_manager_msgs::msg::ControllerState>> controller_name_map;
```

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
pac48: It does only apply to chained controllers for now, so the name should be changed.

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -350,8 +358,29 @@ class MoveItControllerManager : public moveit_controller_manager::MoveItControll
...
+        controller.erase(0, 1);
+        queue.pop_back();
+        for (const auto& dependency : dependency_map_[controller])
+        {
+          queue.push_back(dependency);
+          controllers->push_back("/" + dependency);
+        }
+      }
+    }
+    std::reverse(activate.begin(), activate.end());
```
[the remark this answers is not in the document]
pac48: Unfortunately, the order in which the chained controllers is started matters. Dependencies must be started first, but they are processed last, so the order needs to be flipped to start everything in the correct order.

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -419,6 +454,40 @@ class MoveItControllerManager : public moveit_controller_manager::MoveItControll
...
   }
+  /**
+   * \brief fixChainedControllers modifies ListControllers service response if it contains chained controllers.
+   * Since chained controllers cannot be written to directly, they are removed from the response and their interfaces
+   * are propagated back to the first controller with a non-chained input
+   */
+  bool fixChainedControllers(std::shared_ptr<controller_manager_msgs::srv::ListControllers::Response>& result)
+  {
+    std::unordered_map<std::string, size_t> controller_name_map;
+    for (size_t i = 0; i < result->controller.size(); i++)
```
tylerjw, line 465:
```suggestion
    for (size_t i = 0; i < result->controller.size(); ++i)
```

tylerjw (approved):
One tiny nit. Otherwise, this looks good.

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -350,8 +359,30 @@ class MoveItControllerManager : public moveit_controller_manager::MoveItControll
...
+        controller.erase(0, 1);
+        queue.pop_back();
+        for (const auto& dependency : dependency_map_[controller])
+        {
+          queue.push_back(dependency);
+          controllers->push_back("/" + dependency);
+        }
+      }
+    }
+    // activation dependencies must be started first, but they are processed last, so the of needs to be flipped
```
AndyZe, line 383:
```suggestion
    // activation dependencies must be started first, but they are processed last, so the order needs to be flipped
```

moveit_plugins/moveit_ros_control_interface/src/empty_controller_plugin.cpp
```diff
@@ -0,0 +1,64 @@
...
+/* Author: Paul Gesel */
+
+#include <pluginlib/class_list_macros.hpp>
+#include <moveit_simple_controller_manager/empty_controller_handle.h>
+#include <rclcpp/node.hpp>
+
+namespace moveit_ros_control_interface
+{
+/**
+ * \brief This class allows MoveIt's controller manager to start and stop a controller that do not have a
```
AndyZe, line 44:
```suggestion
 * \brief This class allows MoveIt's controller manager to start and stop a controller that does not have a
```

moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/empty_controller_handle.h
```diff
@@ -0,0 +1,90 @@
...
+#pragma once
+
+#include <moveit_ros_control_interface/ControllerHandle.h>
+#include <rclcpp/rclcpp.hpp>
+
+namespace moveit_simple_controller_manager
+{
+/*
+ * This is an interface for a gripper using control_msgs/GripperCommandAction
+ * action interface (single DOF).
```
AndyZe, line 46:
```suggestion
 * An interface for controllers that have no handle, e.g. chained controllers like an Admittance controller
```

AndyZe (approved):
I've tested it now, and also tested that it doesn't break the regular Joint Trajectory Controller. Once you rebase, we should be clear to merge it.   :)

PR #1498, opened 2022-08-09
WIP: Joint model group pointer

### LINKED ISSUE #1430: JointModelGroup should not be returned as raw pointer
Currently, the robot model class returns a raw pointer to the JointModelGroup [robot_model.cpp#L473](https://github.com/ros-planning/moveit2/blob/2de48a7aafacd45d4f9d4c667ff7337b1da31ddd/moveit_core/robot_model/src/robot_model.cpp#L473). I think it would be much safer to return a shared_ptr or a reference instead. This is issue is probably related to #860 

### TODO:
- [ ] Refactor `RobotModel::getJointModelGroup` to return something safer than a raw pointer

henrygerardmoore:
### Description

Closes #1430. Will be reviewing with @sjahr to decide best methods to achieve goals of this PR

### Checklist


moveit_planners/pilz_industrial_motion_planner/test/unit_tests/src/unittest_get_solver_tip_frame.cpp
```diff
@@ -111,7 +111,9 @@ TEST_F(GetSolverTipFrameTest, TestExceptionMoreThanOneTipFrame)
 
   EXPECT_CALL(solver_mock_, getTipFrames()).Times(AtLeast(1)).WillRepeatedly(ReturnRef(tip_frames));
 
-  EXPECT_THROW(getSolverTipFrame(&jmg_mock_), MoreThanOneTipFrameException);
+  EXPECT_THROW(
+      getSolverTipFrame(std::shared_ptr<moveit::core::JointModelGroup>((moveit::core::JointModelGroup*)&jmg_mock_)),
```
henrygerardmoore, line 115: We need to figure out how to change this obviously

v4hn:
The only real thing moving from `*` to `shared_ptr` does here is to add a reference counter to each use of a JMG (thus adding a tiny bit of overhead).
`nullptr` still needs to be checked everywhere (and often isn't). Ownership of the JMG is held (and imho should be held) by the `RobotModel`, mostly because the model also holds ownership of the IK plugin that is associated with the group.

The only real alternative to pointers I see here are references `&`.
I know that some maintainers are wary of `vector<Object&>` et al. though.
And just to avoid the likely followup, `std::optional<Object&>` is *not* better in my opinion, because it makes the API more annoying to use and people will still call `value()` without handling the empty case.

tylerjw:
>  and people will still call value() without handling the empty case
At least when they do this they'll get an exception instead of a segfault.

I think this raw pointer is a symptom of a bad software design where there are two classes (RobotModel and JointModelGroup) that have a circular lifetime dependency. Each JointModelGroup has a raw pointer to the parent RobotModel. That is a sign these are not actually separate objects and are instead one thing being split in two with an unsafe interface between them.

Issue #1431, opened 2022-07-14
Use clang-format's `QualifierAlignment` upon upgrade to clang-format-14


ChrisThrasher:
https://releases.llvm.org/14.0.0/tools/clang/docs/ClangFormatStyleOptions.html

clang-format-14 adds a new feature called `QualifierAlignment` which among other benefits, lets us automate the decision to use `east const` or `const west`. Once we can safely start requiring clang-format-14, enable this feature to enforce whatever style is already more common in MoveIt. Decisions about potentially changing our style will be saved for a later discussion once the tooling is in place to enforce a particular style.

![image](https://user-images.githubusercontent.com/39244355/179063465-47ddacc2-946e-44bb-a76a-8be49f4448a9.png)

tylerjw:
We do use clang-format-14 for this project on the main branch. Do we want to enable this rule?

ChrisThrasher:
I've started using it [elsewhere](https://github.com/SFML/SFML/pull/2608) and quite like it. It removes yet another detail of code style that would otherwise have to get caught and fixed in code review.

Shobuj-Paul:
Are we proceeding with this? 
If yes, should I make the change?

PR #1504, opened 2022-08-11
rclcpp::Duration in RobotTrajectory

tylerjw:
Signed-off-by: Tyler Weaver <maybe@tylerjw.dev>

### Description

Use the strong type `rclcpp::Duration` in `RobotTrajectory` and opportunistically convert member functions to free functions.

### DIFF
--- a/moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
+++ b/moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
@@ -150,7 +150,7 @@ public:
     return duration_from_previous_;
   }
 
-  [[deprecated("getWayPointDurations is deprecated, replaed by getDurationFromPreviousDeque")]] std::deque<double>
+  [[deprecated("getWayPointDurations is deprecated, replaced by getDurationFromPreviousDeque")]] std::deque<double>
   getWayPointDurations() const
   {
     std::deque<double> ret;
--- a/moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
+++ b/moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
@@ -49,6 +49,7 @@
 #include "rclcpp/rclcpp.hpp"
 #include "rclcpp/time.hpp"
 #include "rclcpp/utilities.hpp"
+#include "rclcpp/duration.hpp"
 
 namespace robot_trajectory
 {
@@ -116,7 +117,7 @@ public:
 
   const moveit::core::RobotState& getWayPoint(std::size_t index) const
   {
-    return *waypoints_[index];
+    return *waypoints_.at(index);
   }
 
   const moveit::core::RobotState& getLastWayPoint() const
@@ -131,7 +132,7 @@ public:
 
   moveit::core::RobotStatePtr& getWayPointPtr(std::size_t index)
   {
-    return waypoints_[index];
+    return waypoints_.at(index);
   }
 
   moveit::core::RobotStatePtr& getLastWayPointPtr()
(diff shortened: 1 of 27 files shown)


tylerjw:
Ideally I'd like to make the robot trajectory a composed type inside a single deque instead of two parallel deques and then move as much as possible out into free functions. I am only getting started with this PR cleaning this type up.

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -220,7 +220,7 @@ bool RuckigSmoothing::runRuckig(robot_trajectory::RobotTrajectory& trajectory,
   trajectory.unwind();
 
   // Initialize the smoother
-  double timestep = trajectory.getAverageSegmentDuration();
+  double timestep = robot_trajectory::average_segment_duration(trajectory).seconds();
```
henningkayser, line 223: Why not use the provided member function instead? We shouldn't mix these two different paradigms. If you are planning to make RobotTrajectory a pure data type with free functions, we can talk about that. Here this change is unnecessarily confusing.

henningkayser (commented):
I really like the move away from using double for time durations. We should probably do this for all of the codebase at some point. I can relate to your idea of making RobotTrajectory a composed type, but then this should happen in one refactoring, or at least we should have an API ready for discussion before merging these changes. IMO, it would make more sense to implement the composed type and then deprecate the old class altogether.

moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
```diff
@@ -144,32 +145,57 @@ class RobotTrajectory
     return waypoints_.front();
   }
 
-  const std::deque<double>& getWayPointDurations() const
+  std::deque<rclcpp::Duration> const& getDurationFromPreviousDeque() const
```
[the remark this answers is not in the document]
tylerjw: The problem is that you can't overload return types to deprecate a function where you want to change the return type.

moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
```diff
@@ -144,32 +145,57 @@ class RobotTrajectory
...
+                   [](auto const& e) { return e.seconds(); });
+    return ret;
+  }
+
   /** @brief  Returns the duration after start that a waypoint will be reached.
    *  @param  The waypoint index.
    *  @return The duration from start; returns overall duration if index is out of range.
    */
-  double getWayPointDurationFromStart(std::size_t index) const;
+  rclcpp::Duration getWayPointDurationFromStartAt(std::size_t index) const;
```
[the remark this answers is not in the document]
tylerjw: Long term I'd like to make all these free functions. 

Maybe I should just design that and we can discuss that plan instead of this incremental PR.

The reason I made this PR was to separate the idea of going from double -> rclcpp::Duration from a larger data-structure change.

moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
```diff
@@ -380,11 +464,17 @@ class RobotTrajectory
...
-  std::deque<double> duration_from_previous_;
+  std::deque<rclcpp::Duration> duration_from_previous_;
   rclcpp::Clock clock_ros_;
 };
 
 /** @brief Operator overload for printing trajectory to a stream */
 std::ostream& operator<<(std::ostream& out, const RobotTrajectory& trajectory);
 
+/// \brief Calculate the total duration of a trajectory
+rclcpp::Duration total_duration(RobotTrajectory const& trajectory);
```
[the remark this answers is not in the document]
tylerjw: I'd like to change the norm to be free functions and not Classes with state. Free functions are easier to reason about being correct, easier to test, easier to compose, and easier to make generic and DRY.

Changing the whole code-base to free functions that operate on structs at once is an impossible task, however I could reasonably do it one object at a time.

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
tylerjw: The reason I picked these two functions and changed them to functions is that you can't deprecate a return type of a function with an overload of the same name, similar to the places where I added an `At` suffix or `Deque` suffix.

moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
```diff
@@ -116,7 +117,7 @@ class RobotTrajectory
 
   const moveit::core::RobotState& getWayPoint(std::size_t index) const
   {
-    return *waypoints_[index];
+    return *waypoints_.at(index);
```
ChrisThrasher, line 120: This change seems unrelated to using `rclcpp::Duration` and deserves its own conversation in a separate issue or pull request. Once we widen the API in this way, we cannot narrow it again without breaking the API so we must be very careful about such changes.

moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
ChrisThrasher, line 148:
```suggestion
  const std::deque<rclcpp::Duration>& getDurationFromPreviousDeque() const
```
There's no need to sprinkle in stylistic changes. Lets keep style consistent until we can enforce const style with clang-format-14. Same goes for a few other places in this PR. See https://github.com/ros-planning/moveit2/issues/1431

moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
```diff
@@ -183,50 +209,86 @@ class RobotTrajectory
...
+    return addSuffixWayPoint(state, rclcpp::Duration::from_seconds(dt));
+  }
+
   /**
    * \brief Add a point to the trajectory
    * \param state - current robot state
    * \param dt - duration from previous
    */
-  RobotTrajectory& addSuffixWayPoint(const moveit::core::RobotStatePtr& state, double dt)
+  RobotTrajectory& addSuffixWayPoint(const moveit::core::RobotStatePtr& state, rclcpp::Duration dt)
```
ChrisThrasher, line 228: Is `rclcpp::Duration` the same size as `double`? We ought to be sure about that before we continue to pass it around by value.

moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
```diff
@@ -237,9 +299,16 @@ class RobotTrajectory
...
-  RobotTrajectory& append(const RobotTrajectory& source, double dt, size_t start_index = 0,
+  RobotTrajectory& append(const RobotTrajectory& source, rclcpp::Duration dt, size_t start_index = 0,
                           size_t end_index = std::numeric_limits<std::size_t>::max());
 
+  [[deprecated("Use of double for duration is deprecated, please pass in rclcpp::Duration")]] RobotTrajectory&
+  append(const RobotTrajectory& source, double dt, size_t start_index = 0,
+         size_t end_index = std::numeric_limits<std::size_t>::max())
+  {
+    return append(source, rclcpp::Duration::from_seconds(dt), start_index, end_index);
+  }
```
ChrisThrasher, line 310: Why is there a mix between some functions being defined in this header and some functions merely being declared in it?

ChrisThrasher (changes requested):

moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
tylerjw: Internally it is an int64_t representing nanoseconds. I think that's the same size as `double`.

moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
tylerjw: That is a problem that already exists here and deserves its own cleanup.

moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
tylerjw: I'll go make sure I clean these out. I have a pattern of doing this without thinking about it because I've been writing all my new code this way recently.

tylerjw:
Closing as this is abandoned unfinished work. I created an issue here to track this if we decide to come back and do this: https://github.com/ros-planning/moveit2/issues/1728

PR #1515, opened 2022-08-15
MSA: write the default controller namespace

### LINKED ISSUE #1514: MSA: default `moveit_controllers.yaml` is missing a few lines
As reported in issue #1506, the default `moveit_controllers.yaml` created by MSA is missing a few lines:

```
    action_ns: follow_joint_trajectory
    default: true
```

AndyZe:
### Description

MoveIt Setup Assistant should write the controller namespace to the yaml file, like this:

```
  arm_controller:
    type: FollowJointTrajectory
    action_ns: follow_joint_trajectory
    joints:
      - J1
...
      - J6
```

Fixes #1514

### DIFF
--- a/moveit_setup_assistant/moveit_setup_controllers/src/moveit_controllers_config.cpp
+++ b/moveit_setup_assistant/moveit_setup_controllers/src/moveit_controllers_config.cpp
@@ -190,12 +190,7 @@ bool MoveItControllersConfig::GeneratedControllersConfig::writeYaml(YAML::Emitte
           if (controller.type_ == "FollowJointTrajectory")
           {
             emitter << YAML::Key << "action_ns" << YAML::Value << "follow_joint_trajectory";
-          }
-          else
-          {
-            RCLCPP_ERROR_STREAM(*logger_, "Only a FollowJointTrajectory type of controller is supported, for now. "
-                                          "Update moveit_controllers.yaml.");
-            return false;
+            emitter << YAML::Key << "default" << YAML::Value << "true";
           }
 
           // Write joints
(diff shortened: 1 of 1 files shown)


JafarAbdi (approved):

moveit_setup_assistant/moveit_setup_controllers/src/moveit_controllers_config.cpp
```diff
@@ -187,6 +187,17 @@ bool MoveItControllersConfig::GeneratedControllersConfig::writeYaml(YAML::Emitte
...
           emitter << YAML::Key << "type" << YAML::Value << controller.type_;
+          if (controller.type_ == "FollowJointTrajectory")
+          {
+            emitter << YAML::Key << "action_ns" << YAML::Value << "follow_joint_trajectory";
+          }
+          else
+          {
+            RCLCPP_ERROR_STREAM(*logger_, "Only a FollowJointTrajectory type of controller is supported, for now. "
+                                          "Update moveit_controllers.yaml.");
+            return false;
```
mikeferguson, line 198: this will actually break the ability to re-export if you have other controllers defined/updated (for instance, a gripper controller would cause this to fail to export).

mikeferguson (commented):
It seems like there must have been a recent regression - my generated config a few weeks ago (June 29) has these defined properly - https://github.com/mikeferguson/ubr_reloaded/blob/ros2/ubr1_moveit/config/moveit_controllers.yaml - aren't these actually "controller.parameters_" (which get exported about ten lines below?)

mikeferguson:
This PR: https://github.com/ros-planning/moveit2/commit/8ee767a49912476497a45917057654a3351dc465 is relevant - it's where these became generic "parameters_"

moveit_setup_assistant/moveit_setup_controllers/src/moveit_controllers_config.cpp
AndyZe: Thanks for mentioning that -- I forgot about grippers.

I simplified this to be identical the state before https://github.com/ros-planning/moveit2/commit/8ee767a49912476497a45917057654a3351dc465

Issue #1328, opened 2022-06-08
`moveit_configs_utils` tries to load `cartesian_limits.yaml` even if Pilz planner is not loaded


AndyZe:
### Description

See discussion here:  https://github.com/ros-planning/moveit2_tutorials/pull/391#issuecomment-1150096437

henningkayser:
Is there anything to do about this now? It's clearly not a bug since nothing is broken because of this. My thought is that technically we could use cartesian limits outside of PILZ in the future.

henningkayser:
Fixed by #1519

PR #1519, opened 2022-08-16
Only require Cartesian limits if Pilz is used

stephanie-eng:
### Description

Attempt to solve https://github.com/ros-planning/moveit2/issues/1328 by checking if Pilz is being used before trying to load pilz_cartesian_limits.yaml. If Pilz is not being used, there shouldn't be a need to have this file.


AndyZe (approved):
I see that `pilz_industrial_motion_planner` is hard-coded as the name of the plugin in moveit_planners/pilz/plugins/pilz_industrial_motion_planner_plugin_description.xml. So I think this is fine  :+1:

tylerjw (approved):

JafarAbdi (approved):

Issue #1435, opened 2022-07-19
support HPP-FCL as collision checker


v4hn:
See https://github.com/humanoid-path-planner/hpp-fcl

Authors of this heavily-changed version of FCL published [a paper at RSS this year](https://arxiv.org/abs/2205.09663) about speedups of factor 2 for convex collision checking over standard GJK.

Together with providing support to mark meshes as convex in general (!), it could accelerate our current pipelines *a lot*.

vatanaksoytezer:
> Authors of this heavily-changed version of FCL published [a paper at RSS this year](https://arxiv.org/abs/2205.09663) about speedups of factor 2 for convex collision checking over standard GJK.

This sounds amazing, definitely, +1 to adding this to be MoveIt should be one of our short-term goals.

simonschmeisser:
Please see (and comment on ...) https://github.com/ros-planning/moveit/pull/3167 as well for improving on the broad phase part of the collision checking.

sea-bass:
+1 this would be such a great enhancement. I've been trying out Pinocchio, which uses HPP-FCL, and it's quite nice. Also has Python bindings which plays well with our goals with MoveIt.

jcarpent:
Hi all,

You may look at our latest benchmarks (see https://github.com/humanoid-path-planner/hpp-fcl?tab=readme-ov-file#a-high-performance-library) done by @lmontaut to see the speed of hpp-fcl compared to alternative open-source solutions:

![image](https://github.com/ros-planning/moveit2/assets/1567317/d1ec6af3-f23a-482b-9707-09fd7327b68d)

jcarpent:
@simonschmeisser Do you have a set of scenes you would like to test?
@lmontaut Interested in incorporating the broad phases into your benchmarks?

jcarpent:
By the way, [here](https://hal.science/hal-04070039/) is a follow-up of our R:SS conference with several accelerated variants of GJK.

PR #1521, opened 2022-08-16
WIP: HPP FCL Plugin

vatanaksoytezer:
Adresses #1435 by creating a new hpp-fcl plugin for MoveIt. Binaries of hpp-fcl are due to be available on tomorrow's sync so I would wait for that at least.


simonschmeisser:
Very interesting, just a short general comment: When copying and modifying files it's nice to do this in separate commits so it's easier to see in github what changes were necessary besides renaming

vatanaksoytezer:
> Very interesting, just a short general comment: When copying and modifying files it's nice to do this in separate commits so it's easier to see in github what changes were necessary besides renaming

Thank you for the suggestion, I'll revise the commit history as soon as this is ready for reviews. I need to get some non-moveit work this week but I'm hoping I can get things going next week.

simonschmeisser:
For MoveIt 1 HPP FCL is not extremely impressive unfortunately
grafik

tylerjw:
It looks like this PR is no longer being worked on as it was determined it did not offer the hoped for performance benefits.

simonschmeisser:
Just for completeness: hpp fcl does offer a newer api for collision checks and this might speed up things in fact. Also hpp fcl might be faster at the narrow phase, especially mesh to mesh, which my benchmarks do not test that well. I still hope to finish the MoveIt 1 PR and then this could be ported

PR #1539, opened 2022-08-26
Benchmark parallel planning pipelines

sjahr:
### Description

Fixes, cleanup and updates benchmarks package + enable benchmarking of parallel planing interface of moveit_cpp. This PR includes only an initial effort for general cleanups, refactorings and improvements of the benchmarks package, since I originally planned to "just add" benchmarking capabilities for parallel planning. Once it is merged, benchmarking is functional again and we can put further effort into it.

### Testing
Run the benchmarks with
```bash
ros2 launch moveit_ros_benchmarks run_benchmark.launch.py
```
Once all benchmarks run through, terminate the window.

Afterwards, based a database can be created in the directory where you run the following command:
```bash
ros2 run moveit_ros_benchmarks moveit_benchmark_statistics.py /tmp/moveit_benchmarks/*
```
The results can be visualized with https://plannerarena.org/

If you want to introspect the test database, you use the moveit2_tutorials and run:

```
ros2 launch moveit2_tutorials persistent_scenes_demo.launch.py
```
In the `Context` tab of the RVIZ MotionPlanning window, you can change the path to the host database to the one in the benchmarks package and  the scenes can be visualized in the `Stored Scenes` tab (when the db is successfully loaded).
### Checklist


moveit_ros/benchmarks/launch/run_benchmark.launch.py
```diff
@@ -0,0 +1,94 @@
+import os
```
henningkayser, line 1: as discussed, this should probably be moved into moveit_resources along with the db file.

moveit_ros/benchmarks/src/BenchmarkExecutor.cpp
```diff
@@ -95,148 +98,154 @@ static std::string getHostname()
...
+[[nodiscard]] bool BenchmarkExecutor::initialize(const std::vector<std::string>& planning_pipeline_names)
 {
-  planning_pipelines_.clear();
+  // Initialize moveit_cpp
+  moveit_cpp_ = std::make_shared<moveit_cpp::MoveItCpp>(node_);
+  moveit_cpp_->getPlanningSceneMonitor()->providePlanningSceneService();
 
-  // ros::NodeHandle pnh("~");
-  std::string parent_node_name = node_->get_name();
+  // std::string parent_node_name = node_->get_name();
```
henningkayser, line 128:
```suggestion
```

moveit_ros/benchmarks/src/BenchmarkExecutor.cpp
```diff
@@ -95,148 +98,154 @@ static std::string getHostname()
...
+         moveit_cpp_->getPlanningPipelines())
       RCLCPP_INFO_STREAM(LOGGER, "Pipeline: " << entry.first << ", Planner: " << entry.second->getPlannerPluginName());
   }
+  return true;
 }
 
 void BenchmarkExecutor::clear()
 {
-  if (pss_)
+  if (planning_scene_storage_)
```
henningkayser, line 168: good readability improvement :+1:

moveit_ros/benchmarks/src/BenchmarkOptions.cpp
```diff
@@ -260,40 +271,109 @@ void BenchmarkOptions::readWorkspaceParameters(const rclcpp::Node::SharedPtr& no
...
+    RCLCPP_ERROR(LOGGER, "Fail to get the parameter in `%s` namespace.", (ns + ".parallel_pipelines").c_str());
+    return false;
+  }
 
-    std::vector<std::string> planners;
-    if (!node->get_parameter(pipeline_parameter_name, planners))
-    {
-      RCLCPP_ERROR(LOGGER, "Fail to get the parameter in `%s` namespace.", pipeline_parameter_name.c_str());
-      return;
+  for (const std::string& parallel_pipeline : parallel_pipelines)
```
henningkayser, line 327: Wouldn't it be simpler to just use the pipelines that are configured with MoveItCpp, and only reference the to-be-benchmarked pipeline ids?

moveit_ros/benchmarks/src/RunBenchmark.cpp
```diff
@@ -56,17 +56,65 @@ int main(int argc, char** argv)
...
-  // Running benchmarks
-  if (!server.runBenchmarks(opts))
-    RCLCPP_ERROR(LOGGER, "Failed to run all benchmarks");
+  if (options.scene_name_.empty())
+  {
+    std::vector<std::string> scene_names;
+    try
+    {
+      warehouse_ros::DatabaseLoader db_loader(node);
+      warehouse_ros::DatabaseConnection::Ptr warehouse_connection = db_loader.loadDatabase();
```
henningkayser, line 78: There doesn't seem to be any resource handling for `warehouse_connection` or `planning_scene_storage`. Should this be a unique_ptr instead? Can `planning_scene_storage` be an object instead of a pointer?

moveit_ros/benchmarks/src/simple_benchmarks/CombinePredefinedPosesBenchmark.cpp
```diff
@@ -1,156 +0,0 @@
-/*********************************************************************
```
henningkayser, line 1: Do we have a replacement for this benchmark?

moveit_ros/benchmarks/include/moveit/benchmarks/BenchmarkExecutor.h
```diff
@@ -191,36 +193,35 @@ class BenchmarkExecutor
...
-                    const std::map<std::string, std::vector<std::string>>& planners, int runs);
+  void runBenchmark(moveit_msgs::msg::MotionPlanRequest request);
 
-  planning_scene_monitor::PlanningSceneMonitor* psm_;
-  moveit_warehouse::PlanningSceneStorage* pss_;
-  moveit_warehouse::PlanningSceneWorldStorage* psws_;
-  moveit_warehouse::RobotStateStorage* rs_;
-  moveit_warehouse::ConstraintsStorage* cs_;
-  moveit_warehouse::TrajectoryConstraintsStorage* tcs_;
+  planning_scene_monitor::PlanningSceneMonitor* planning_scene_monitor_;
```
henningkayser, line 203: Can we change these to smart pointers, so that we don't have to manage lifetime?

moveit_ros/benchmarks/src/RunBenchmark.cpp
sjahr: No, because `boost::shared_ptr<DatabaseConnection> Ptr` we need to  clean this up first :/

moveit_ros/benchmarks/src/simple_benchmarks/CombinePredefinedPosesBenchmark.cpp
sjahr: I'd like to move this into the tutorial repository. What do you think about that?

moveit_ros/benchmarks/include/moveit/benchmarks/BenchmarkOptions.h
```diff
@@ -100,12 +132,11 @@ class BenchmarkOptions
   /* \brief Get the parameter set of the planning workspace */
   const moveit_msgs::msg::WorkspaceParameters& getWorkspaceParameters() const;
 
-protected:
```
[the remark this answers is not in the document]
sjahr: I refactored it in a way, where I hope the struct datatype makes more sense. I didn't like the previous version of the configuration containter

moveit_ros/benchmarks/src/BenchmarkOptions.cpp
sjahr: Not sure if I understand your proposal correctly but I think it is more useful to give both pieces of information for example if you want to use the same planner_id in two different pipelines that have a different set of plan_request_adapters or trajectory generation algorithms. For example, two OMPL pipelines that both use RRTConnect but one uses ruckig and the other one just TOTG

moveit_ros/benchmarks/include/moveit/benchmarks/BenchmarkOptions.h
```diff
@@ -44,9 +44,38 @@
 
 namespace moveit_ros_benchmarks
 {
-class BenchmarkOptions
+/// \brief Options to configure a benchmark experiment. The configuration is provided via ROS2 parameters
```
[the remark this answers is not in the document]
sjahr: Yes, I've added a TODO

moveit_ros/benchmarks/launch/run_benchmark.launch.py
henningkayser: shouldn't this file be removed if it's part of resources or the tutorial?

henningkayser (approved):

PR #1541, opened 2022-08-28
Delete unported `moveit_joy` visualization demo

AndyZe:
### Description

I was poking around, looking into build warnings about deprecated `setup.py`. While I didn't find a complete solution, I think we can safely delete all of these scripts since they still refer to `catkin`.


moveit_commander/setup.py
```diff
@@ -1,8 +0,0 @@
-from setuptools import setup
```
sjahr, line 1: This is part of a package that is COLCON_IGNOREd right now, should we keep so it can be removed, when the package is ported/refactored/removed?

moveit_core/setup.py
```diff
@@ -1,10 +0,0 @@
-# ! DO NOT MANUALLY INVOKE THIS setup.py, USE CATKIN INSTEAD
-
-from catkin_pkg.python_setup import generate_distutils_setup
-from setuptools import setup, find_packages
-
-packages = find_packages("python/src/")
```
sjahr, line 6: What about the `/python` subdirectory? Isn't this related to this file?

sjahr (commented):
Good idea to address the setup.py deprecation warnings! I fear that removing these files might lead to dead code, for example the \python in moveit_core. @peterdavidfagan what do you think about this? Will the deprecation warning be addressed by your work on MoveIt2 python bindings?

AndyZe:
Small misconception here:

> I fear that removing these files might lead to dead code

All of this code is already dead!

You're right, the Python code in `moveit_core` and all of `moveit_commander` should probably be removed all at once. I'll remove them from this PR and tweak the other packages as well.

peterdavidfagan:
> 

Hi @sjahr, 

Thanks for looping me into this conversation. 

> I fear that removing these files might lead to dead code, for example the \python in moveit_core

I have actually removed this code in my current working branch of moveit_py so it would be ok to remove in my mind as @AndyZe has suggested. 

> Will the deprecation warning be addressed by your work on MoveIt2 python bindings?

I am happy to address this as part of my GSoC project. For any of the code related to the library I am planning to release I would expect to ensure any warning including those related to deprecations to be addressed where they possibly can be. If existing code that is being maintained contains such deprecation warning I would be happy to fix these too.

AndyZe:
Thanks @peterdavidfagan.

My bit of googling into this type of error suggested that it's a ROS2 build system issue and we need to wait for Open-Robotics to fix it. Nothing to be done on our end.

```
--- stderr: moveit_configs_utils
/usr/lib/python3/dist-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(
```

sjahr (approved):

PR #1542, opened 2022-08-29
Add setTrajectoryConstraints() to PlanningComponent and add planning time to PlanningSolution

sjahr:
### Description

Two small additions for planning_component.cpp

### Checklist


moveit_ros/planning/moveit_cpp/src/planning_component.cpp
```diff
@@ -177,6 +185,8 @@ PlanningComponent::PlanSolution PlanningComponent::plan(const PlanRequestParamet
   }
   last_plan_solution_->start_state = req.start_state;
   last_plan_solution_->trajectory = res.trajectory_;
```
henningkayser, line 187: Since it's pretty much the exact same data anyway, could you replace the `PlanSolution` struct with MotionPlanResponse` from [here](https://github.com/ros-planning/moveit2/blob/main/moveit_core/planning_interface/include/moveit/planning_interface/planning_response.h#L46), and maybe add the bool conversion to it? The start_state is not needed, really.

henningkayser (commented):
I'm fine with these changes. Keep in mind that we are planning to simplify this class a lot, as most of it would be better implemented by a state-less [MotionPlanRequest builder](https://github.com/ros-planning/moveit2/issues/1455)

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
sjahr: Good idea

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
```diff
@@ -150,33 +156,37 @@ PlanningComponent::PlanSolution PlanningComponent::plan(const PlanRequestParamet
...
-    last_plan_solution_->error_code = moveit::core::MoveItErrorCode::FAILURE;
-    return *last_plan_solution_;
+    last_plan_solution_.error_code_ = moveit::core::MoveItErrorCode::FAILURE;
+    return last_plan_solution_;
   }
   const planning_pipeline::PlanningPipelinePtr pipeline =
       moveit_cpp_->getPlanningPipelines().at(parameters.planning_pipeline);
   pipeline->generatePlan(planning_scene, req, res);
-  last_plan_solution_->error_code = res.error_code_.val;
+  last_plan_solution_.error_code_ = res.error_code_.val;
```
henningkayser, line 180: Why is there a .val here? I thought these would be the same type and we could just copy the struct completely.

moveit_core/planning_interface/include/moveit/planning_interface/planning_response.h
```diff
@@ -37,23 +37,38 @@
...
+  // Construct a ROS message from struct data
   void getMessage(moveit_msgs::msg::MotionPlanResponse& msg) const;
 
+  // Trajectory generated by the queried planner
   robot_trajectory::RobotTrajectoryPtr trajectory_;
+  // Time to plan the respond to the planning query
   double planning_time_;
-  moveit_msgs::msg::MoveItErrorCodes error_code_;
+  // Result status of the query
+  moveit::core::MoveItErrorCode error_code_;
```
henningkayser, line 63: This change is not consistent with `MotionPlanDetailedResponse`. Whatever you change here should be reflected in the other struct as well.

sjahr: I'll add it there as well :+1:

sjahr: @henningkayser What do you think about using inheritance
```cpp
struct MotionPlanDetailedResponse : public MotionPlanResponse
{
  // Make this a free function for both classes
  // void getMessage(moveit_msgs::msg::MotionPlanDetailedResponse& msg) const; 

  std::vector<std::string> description_;
  std::vector<double> processing_time_;
};
```
or aggregation here?
```cpp
struct MotionPlanDetailedResponse : public MotionPlanResponse
{
   void getMessage(moveit_msgs::msg::MotionPlanDetailedResponse& msg) const; 

  MotionPlanResponse response;
  std::vector<std::string> description_;
  std::vector<double> processing_time_;
};
```
I think both would avoid changing the other if one changes in the future. Or to remove `MotionPlanDetailedResponse` completely and make processing time and description `std::optional`

henningkayser: well, this got stale... I think neither option is applicable, the detailed response includes a vector of trajectories.

sjahr: I see, I'll update this PR

sjahr: The message definition states
```
# Multiple solution paths are reported, each reflecting intermediate steps in the trajectory processing

# The list of reported trajectories
```
But is that used anywhere/ how is it interpretable when the response is just a vector of unlabeled trajectories?

moveit_core/planning_interface/include/moveit/planning_interface/planning_response.h
```diff
@@ -37,23 +37,38 @@
...
+  /// The full starting state used for planning
+  moveit_msgs::msg::RobotState start_state_;
+
+  // Enable checking of query success or failure, for example if(response) ...
+  explicit operator bool() const
+  {
+    return bool(error_code_);
```
AndyZe, line 70: I don't think this is going to work. If `error_code_` is negative, like `int32 PLANNING_FAILED=-1`, it will cast to true.

`int32 SUCCESS=1` also casts to true.

AndyZe:
```suggestion
    return error_code_ == moveit::core::MoveItErrorCode::SUCCESS;
```

moveit_core/planning_interface/include/moveit/planning_interface/planning_response.h
henningkayser: I'm not sure if it's being used anywhere, but my understanding is that the trajectories show you the result for each step in the process. For instance, you should be able to compare the raw planner solution with the smoothed output.

moveit_core/planning_interface/include/moveit/planning_interface/planning_response.h
henningkayser: This is not the error code value. It's using the `MoveItErrorCode` type which is convertible to bool. See https://github.com/ros-planning/moveit2/blob/main/moveit_core/utils/include/moveit/utils/moveit_error_code.h#L62

henningkayser: I suggested to use the same in your PR. https://github.com/ros-planning/moveit2/pull/1605#discussion_r1007826518

sjahr:
I'll close this PR since it is also included in #1420 and additional addressed reviews are included there

PR #1543, opened 2022-08-29
[Servo] Use a WallRate so the clock is monotonically increasing

AndyZe:
### Description

The default `rclcpp::Rate` uses a `[std::chrono::system_clock](http://en.cppreference.com/w/cpp/chrono/system_clock.html)`. Better to use a `rclcpp::WallRate` here since we only care about the change in time.

Hat tip for @ChrisThrasher for bringing this type of issue to my attention.

If we're really lucky, this will fix the Servo tests.


AndyZe:
CI has passed 3X in a row now. I'm going to see if I can re-enable some of the tests that were previously commented.

vatanaksoytezer (approved):

vatanaksoytezer:
@AndyZe do you want to try to re-enable other tests in this PR or is this good to merge?

AndyZe:
> @AndyZe do you want to try to re-enable other tests in this PR or is this good to merge?

I tried re-enabling them but saw some compiler errors because the API has changed, I guess. I'd be happy to push that to later.

PR #1547, opened 2022-09-02
Improve Cartesian interpolation

rhaschke:
Port from Noetic of https://github.com/ros-planning/moveit/pull/3197.
Prepares addressing https://github.com/ros-planning/moveit_task_constructor/issues/359 (see https://github.com/ros-planning/moveit_task_constructor/issues/380 as well.)

### DIFF
--- a/moveit_core/robot_state/include/moveit/robot_state/cartesian_interpolator.h
+++ b/moveit_core/robot_state/include/moveit/robot_state/cartesian_interpolator.h
@@ -200,8 +200,8 @@ public:
 
      In contrast to the previous function, the Cartesian path is specified as a target frame to be reached (\e target)
      for a virtual frame attached to the robot \e link with the given \e link_offset.
-     The target frame is assumed to be specified either w.r.t. to the global reference frame or the virtual link frame (\e global_reference_frame is false).
-     This function returns the percentage (0..1) of the path that was achieved.
+     The target frame is assumed to be specified either w.r.t. to the global reference frame or the virtual link frame
+     (\e global_reference_frame is false). This function returns the percentage (0..1) of the path that was achieved.
      All other comments from the previous function apply. */
   static Percentage
   computeCartesianPath(RobotState* start_state, const JointModelGroup* group,
--- a/moveit_core/robot_state/include/moveit/robot_state/cartesian_interpolator.h
+++ b/moveit_core/robot_state/include/moveit/robot_state/cartesian_interpolator.h
@@ -140,18 +140,16 @@ public:
     double meters;
   };
 
-  /** \brief Compute the sequence of joint values that correspond to a straight Cartesian path for a particular group.
+  /** \brief Compute the sequence of joint values that correspond to a straight Cartesian path for a particular link.
 
-     The Cartesian path to be followed is specified as a direction of motion (\e direction, unit vector) for the origin
-     The Cartesian path to be followed is specified as a direction of motion (\e direction, unit vector) for the origin
-     of a robot link (\e link). The direction is assumed to be either in a global reference frame or in the local
-     reference frame of the link. In the latter case (\e global_reference_frame is false) the \e direction is rotated
-     accordingly. The link needs to move in a straight line, following the specified direction, for the desired \e
-     distance. The resulting joint values are stored in the vector \e traj, one by one. The maximum distance in
+     The Cartesian path to be followed is specified as a \e translation vector to be followed by the robot \e link.
+     This vector is assumed to be specified either in the global reference frame or in the local
+     reference frame of the link.
+     The resulting joint values are stored in the vector \e traj, one by one. The maximum distance in
      Cartesian space between consecutive points on the resulting path is specified in the \e MaxEEFStep struct which
      provides two fields: translation and rotation. If a \e validCallback is specified, this is passed to the internal
      call to setFromIK(). In case of IK failure, the computation of the path stops and the value returned corresponds to
-     the distance that was computed and for which corresponding states were added to the path.  At the end of the
+     the distance that was achieved and for which corresponding states were added to the path.  At the end of the
      function call, the state of the group corresponds to the last attempted Cartesian pose.
 
      During the computation of the trajectory, it is usually preferred if consecutive joint values do not 'jump' by a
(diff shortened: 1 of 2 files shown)


moveit_core/robot_state/include/moveit/robot_state/cartesian_interpolator.h
```diff
@@ -173,43 +171,63 @@ class CartesianInterpolator
...
   static Percentage
   computeCartesianPath(RobotState* start_state, const JointModelGroup* group,
                        std::vector<std::shared_ptr<RobotState>>& traj, const LinkModel* link,
                        const Eigen::Isometry3d& target, bool global_reference_frame, const MaxEEFStep& max_step,
                        const JumpThreshold& jump_threshold,
                        const GroupStateValidityCallbackFn& validCallback = GroupStateValidityCallbackFn(),
                        const kinematics::KinematicsQueryOptions& options = kinematics::KinematicsQueryOptions(),
-                       kinematics::KinematicsBase::IKCostFn cost_function = kinematics::KinematicsBase::IKCostFn());
+                       kinematics::KinematicsBase::IKCostFn cost_function = kinematics::KinematicsBase::IKCostFn(),
+                       const Eigen::Isometry3d& link_offset = Eigen::Isometry3d::Identity());
```
henningkayser, line 214: What's the advantage of `link_offset` compared to offsetting `target` accordingly? These functions already have way too many arguments, adding this will make them even more difficult to understand.

moveit_core/robot_state/src/cartesian_interpolator.cpp
```diff
@@ -56,46 +56,45 @@ static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_robot_state.cart
 
 CartesianInterpolator::Distance CartesianInterpolator::computeCartesianPath(
     RobotState* start_state, const JointModelGroup* group, std::vector<RobotStatePtr>& traj, const LinkModel* link,
-    const Eigen::Vector3d& direction, bool global_reference_frame, double distance, const MaxEEFStep& max_step,
+    const Eigen::Vector3d& translation, bool global_reference_frame, const MaxEEFStep& max_step,
```
henningkayser, line 59: If you want to change it to translation, I'd suggest using the `Eigen::Translation3d` type. I like this simplification, though.

moveit_core/robot_state/include/moveit/robot_state/cartesian_interpolator.h
rhaschke: There is a fundamental difference between adding the offset once to the target and considering the offset during the whole motion - but only when it comes to rotations. Consider rotating about this offset point (see the video [here](https://github.com/ros-planning/moveit_task_constructor/pull/380) for illustration):
In both approaches the _final_ end-effector poses will be identical. However, how you reach there is dramatically different:
- in the former case, you will perform a straigh-line motion along the secant of the circle
- in the latter case, you will perform a circular motion

moveit_core/robot_state/src/cartesian_interpolator.cpp
rhaschke: Tried this, but changing the type to `Translation3d` make the code much more complicated, because you cannot easily pass `scalar * Eigen::Vector3d` anymore. It's debatable, but here we actually want to pass the translation direction, while `Translation3d` is a transform.

moveit_core/robot_state/include/moveit/robot_state/cartesian_interpolator.h
henningkayser: I get that. I just don't get why the Cartesian motion can't be computed outside of the IK+interpolation functions (at least that's what I thought these were), and then passed as `EigenSTL::vector_Isometry3d&` waypoints. To me it just looks like the functions do to much, and a separate function that transforms your waypoints before passing it to `computeCartesianPath()` would be more adequate.

moveit_core/robot_state/src/cartesian_interpolator.cpp
henningkayser: In this case it would really be about semantics, but I see your point. It's not overly important here.

rhaschke: I was talking about semantics too. We don't move towards a given (translation) _transformation_ (given w.r.t. local or global frame) but along some _direction vector_ :wink:

moveit_core/robot_state/include/moveit/robot_state/cartesian_interpolator.h
rhaschke: Such a "transformation" function would do exactly the same as `computeCartesianPath()`, just w/o IK computation. Hence, I deem it more reasonable to avoid the corresponding code duplication.

AndyZe (commented):
It looks good to me, just some documentation suggestions.

moveit_core/robot_state/include/moveit/robot_state/cartesian_interpolator.h
```diff
@@ -173,43 +171,63 @@ class CartesianInterpolator
 
      Kinematics solvers may use cost functions to prioritize certain solutions, which may be specified with \e cost_function. */
   static Distance
+  computeCartesianPath(RobotState* start_state, const JointModelGroup* group,
```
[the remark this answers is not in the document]
rhaschke: A similar statement can be found a few lines above. If you want to add the term "successful", it should be done there.

moveit_core/robot_state/include/moveit/robot_state/cartesian_interpolator.h
```diff
@@ -140,18 +140,16 @@ class CartesianInterpolator
...
+
+     The Cartesian path to be followed is specified as a \e translation vector to be followed by the robot \e link.
+     This vector is assumed to be specified either in the global reference frame or in the local
+     reference frame of the link.
+     The resulting joint values are stored in the vector \e traj, one by one. The maximum distance in
      Cartesian space between consecutive points on the resulting path is specified in the \e MaxEEFStep struct which
      provides two fields: translation and rotation. If a \e validCallback is specified, this is passed to the internal
      call to setFromIK(). In case of IK failure, the computation of the path stops and the value returned corresponds to
-     the distance that was computed and for which corresponding states were added to the path.  At the end of the
+     the distance that was achieved and for which corresponding states were added to the path.  At the end of the
```
rhaschke, line 152:
```suggestion
     the distance that was successfully achieved and for which corresponding states were added to the path.  At the end of the
```

AndyZe (approved):

henningkayser (approved):

swiz23:
Note that when I tried to pull this branch and do a clean build, I ran into a linker error caused by moveit_ros_robot_interaction. The MR at https://github.com/ros-planning/moveit2/pull/1617 fixed the issue.

PR #1550, opened 2022-09-05
Improve CMake usage

sjahr:
### Description

Apply some CMake recommendations from @ChrisThrasher + Update `RUNTIME DESTINATION` for some packages to `bin`
as [recommended by the Ament-CMake docs](https://docs.ros.org/en/rolling/How-To-Guides/Ament-CMake-Documentation.html) 

### Checklist


moveit_core/collision_detection_bullet/CMakeLists.txt
```diff
@@ -66,21 +66,21 @@ if(BUILD_TESTING)
   ament_add_gtest(test_bullet_collision_detection test/test_bullet_collision_detection_pr2.cpp)
   target_link_libraries(test_bullet_collision_detection moveit_test_utils ${MOVEIT_LIB_NAME})
   # TODO: remove if transition to gtest's new API TYPED_TEST_SUITE_P is finished
-  if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  if(CMAKE_CXX_COMPILER_ID MATCHES "(GNU|Clang)")
```
gavanderhoorn, line 69: wouldn't a generator expression be more idiomatic CMake nowadays?

ChrisThrasher, line 69: A single `add_compile_options` call would let you avoid repeating this same snippet in this directory.
```cmake
if(CMAKE_CXX_COMPILER_ID MATCHES "(GNU|Clang)")
    add_compile_options(-Wno-deprecated-declarations)
endif()
```

moveit/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.10.2)
+cmake_minimum_required(VERSION 3.16.3)
```
ChrisThrasher, line 1: What's the oldest Ubuntu LTS release we're expected to support?

moveit_core/collision_detection_bullet/CMakeLists.txt
sjahr: Good idea, I'll take a look at it. Thanks!

moveit/CMakeLists.txt
sjahr: There is a REP about [ROS 2 Releases and Target Platforms ](https://www.ros.org/reps/rep-2000.html). I've chosen the minimal version based on that

moveit_core/collision_detection_bullet/CMakeLists.txt
ChrisThrasher: I don't believe it would be more idiomatic nor do I believe it's even possible. There is no generic regex match generator expression last time I checked.

moveit/CMakeLists.txt
ChrisThrasher: > A single ROS 2 distribution will only have full Tier 1 support for a single Ubuntu LTS. The rationale is that fully supporting two LTS versions - which means 2-year-different versions of upstream dependencies - is a tremendous overhead and sometimes even impossible. 

If we're only supporting Ubuntu 22 with this code then you can raise that CMake requirement to 3.22.

moveit_core/collision_detection_bullet/CMakeLists.txt
sjahr: No, you're right there does not seem a regex match generator impression. Maybe it would have worked with a boolean variable which is then used as condition in the generator expression, but I like your proposal to refactor this into a "global" setting to avoid copy&pastes better.

moveit/CMakeLists.txt
sjahr: This is actually a good point, I'll bump the version to 3.22. @henningkayser @vatanaksoytezer @JafarAbdi any objections?

henningkayser: You can safely bump this to 3.22, we don't support Focal on `main` or `humble`.

henningkayser (approved):

vatanaksoytezer (approved):

tylerjw (approved):

ChrisThrasher (approved):
Really glad to see modern CMake being used in MoveIt!

PR #1553, opened 2022-09-06
Fix moveit_core building

rhaschke:
Trying to build branch `main` on `Galactic`, I ran into the following compiler error:
```
collision_matrix.cpp: In constructor ‘collision_detection::AllowedCollisionMatrix::AllowedCollisionMatrix(const srdf::Model&)’:
collision_matrix.cpp:62:39: error: ‘const class srdf::Model’ has no member named ‘getNoDefaultCollisionLinks’
```
There were actually two issues (having srdfdom part of the current workspace):
- `srdfdom` was missing as an explicit dependency. Its includes were only found accidentally in /opt/ros
- The include order was screwed because moveit_planning_scene's include directories (not comprising srdfdom but /opt/ros) were included before all other includes. The latter are pulled in via `ament_target_dependencies()`, which handles correct ordering for the overlay. Thus srdfdom was found in /opt/ros instead of the local overlay.

Skimming through the cmake files, I noticed inconsistent naming of export targets: Most packages used `export_${PROJECT_NAME}`, while some also used the [standard (by convention)](https://docs.ros.org/en/rolling/How-To-Guides/Ament-CMake-Documentation.html#building-a-library) `${PROJECT_NAME}Targets`. I changed all to the latter.

### DIFF
--- a/moveit_core/CMakeLists.txt
+++ b/moveit_core/CMakeLists.txt
@@ -208,14 +208,14 @@ add_subdirectory(collision_detection_fcl)
 
 install(
   TARGETS ${THIS_PACKAGE_LIBRARIES}
-  EXPORT export_${PROJECT_NAME}
+  EXPORT ${PROJECT_NAME}Targets
   LIBRARY DESTINATION lib
   ARCHIVE DESTINATION lib
   RUNTIME DESTINATION bin
   INCLUDES DESTINATION include
 )
 
-ament_export_targets(export_${PROJECT_NAME} HAS_LIBRARY_TARGET)
+ament_export_targets(${PROJECT_NAME}Targets HAS_LIBRARY_TARGET)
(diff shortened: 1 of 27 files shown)


henningkayser:
Why would you re-enable Galactic support on main? We dropped it a while ago because of API breakages, Focal, etc...

rhaschke:
> Why would you re-enable Galactic support on main? We dropped it a while ago because of API breakages, Focal, etc...

I'm running Focal. However, all fixes suggested here are independent of Galactic. Galactic-specific patches I will keep locally only.

rhaschke:
@AndyZe, looks like `test_servo_singularity` is flaky!

AndyZe:
Sorry. I thought it was fixed recently in #1543 (which passed CI 5x successfully) but I guess we'll have to disable it again, for now.

tylerjw (approved):

PR #1556, opened 2022-09-07
[Servo] CI simplification

### LINKED ISSUE #1499: CI: Servo singularity test is flaky
See https://github.com/ros-planning/moveit2/runs/7751780196?check_suite_focus=true

AndyZe:
- Delete the flaky, timing-based singularity integration test
- Add a new, simple, reliable unit test for singularities
- Delete the "basic integration tests" that were redundant with some other tests
- Use `rclcpp::WallRate` everywhere, since `rclcpp::Rate` can move backward in time

Fixes #1499


moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
```diff
@@ -77,6 +77,28 @@ enum class ServoType
...
+ * @param[in] svd                 A singular value decomposition of the Jacobian
+ * @param[in] pseudo_inverse      The pseudo-inverse of the Jacobian
+ * @param[in] hard_stop_singularity_threshold  Halt if condition(Jacobian) > hard_stop_singularity_threshold
+ * @param[in] lower_singularity_threshold      Decelerate if condition(Jacobian) > lower_singularity_threshold
+ * @param[in] leaving_singularity_threshold_multiplier      Allow faster motion away from singularity
+ * @param[in, out] clock          A ROS clock, for logging
+ * @param[in, out] current_state  The state of the robot. Used in internal calculations.
+ * @param[out] status             Singularity status
+ */
+double velocityScalingFactorForSingularity(const moveit::core::JointModelGroup* joint_model_group,
```
AndyZe, line 93: This was changed to a free function, to make testing easier. All of the inputs and outputs are specified explicitly now

moveit_ros/moveit_servo/test/servo_calcs_unit_tests.cpp
```diff
@@ -123,6 +128,43 @@ TEST_F(EnforceLimitsTests, LowJointVelocityDeltaTest)
...
+
+  rclcpp::Clock clock;
+  std::shared_ptr<moveit::core::RobotState> robot_state = std::make_shared<moveit::core::RobotState>(robot_model_);
+  moveit_servo::StatusCode status;
+
+  double scaling_factor = moveit_servo::velocityScalingFactorForSingularity(
+      joint_model_group_, commanded_twist, svd, pseudo_inverse, hard_stop_singularity_threshold,
+      lower_singularity_threshold, leaving_singularity_threshold_multiplier, clock, robot_state, status);
+
+  EXPECT_EQ(scaling_factor, 0);
```
tylerjw, line 165: suggest using `EXPECT_DOUBLE_EQ` when comparing floating point values

reference: https://google.github.io/googletest/reference/assertions.html#EXPECT_DOUBLE_EQ

AndyZe: I expect exactly zero here, i.e. all bits should be equal to zero.

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
tylerjw: You can move it to a separate file then if you'd like, that would help a small amount with the compile time issues of servo_calcs.cpp.o.

AndyZe: done

henningkayser (commented):
I agree with these changes overall, but deleting flaky tests is really not a good solution, it only helps with hiding and forgetting bugs. Please only comment the tests so that CI succeeds.

tylerjw (approved):
Merging because I hope this resolves the servo flaky test.

PR #1568, opened 2022-09-12
Use generate_parameter_library to load kinematics parameters

Abishalini:
### Description

This PR replaces how kinematics parameters are declared and loaded in the `KinematicsPluginLoader` class.
The default values for the kinematics parameters are now declared in `kinematics_parameters.yaml`

We could specify multiple kinematics solver for each group. That is being deprecated in this PR. 
We have instead decided to declare just one solver per group. The solver parameter will be a string and the search resolution will be a double type. 

Parameter `kinematics_solver_ik_links` is deprecated since we have not found an example that uses it. The code gets the tip link from `JointModelGroup` instead.

Kinematic parameters that are loaded in the plugins themselves still use the old format and will be replaced in subsequent PRs after this is merged.

If anyone has recommendations on additional validations for the parameters, please let me know.

Notes:
Adding `generate_parameter_library` to the `.repos` file since the feature required to read parameters with dynamic name-spacing is not yet released.

### Checklist


moveit_ros/planning/kinematics_plugin_loader/include/moveit/kinematics_plugin_loader/kinematics_plugin_loader.h
```diff
@@ -54,27 +55,8 @@ class KinematicsPluginLoader
...
-  /** \brief Use a default kinematics solver (\e solver_plugin) for
-      all the groups in the robot model. The default timeout for the
-      solver is \e solve_timeout and the default number of IK attempts
-      is \e ik_attempts. Take node as an argument and as optional argument the name of the ROS
-      parameter under which the robot description can be found. This
-      is passed to the kinematics solver initialization as well as
-      used to read the SRDF document when needed. */
-  KinematicsPluginLoader(const rclcpp::Node::SharedPtr& node, const std::string& solver_plugin, double solve_timeout,
-                         const std::string& robot_description = "robot_description",
-                         double default_search_resolution = 0.0)
```
Abishalini, line 72: Deleting this constructor since it's not used anywhere. All default kinematics parameters are listed in `kinematics_parameters.yaml`

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -296,129 +285,58 @@ moveit::core::SolverAllocatorFn KinematicsPluginLoader::getLoaderFunction(const
...
+
+        std::string kinematics_solver_res_param_name = kinematics_param_prefix + ".kinematics_solver_search_resolution";
+        const auto kinematics_solver_search_resolution =
+            kinematics_params_.at(known_group.name_).kinematics_solver_search_resolution;
+        RCLCPP_DEBUG(LOGGER, "Param %s : %f", kinematics_solver_res_param_name.c_str(),
+                     kinematics_solver_search_resolution);
+        search_res[known_group.name_].push_back(kinematics_solver_search_resolution);
+
+        // Allow a kinematic solver's tip links to be specified on the rosparam server as an array
+        std::string ksolver_ik_links_param_name = kinematics_param_prefix + ".kinematics_solver_ik_links";
```
Abishalini, line 320: This parameter is not set anywhere and I'm not sure if this is being used. Should I keep this remove this parameter check?

moveit_planners/test_configs/prbt_moveit_config/config/kinematics.yaml
```diff
@@ -4,7 +4,7 @@ manipulator:
   #kinematics_solver_timeout: 0.05
   #link_prefix: prbt_
   #prefix: prbt_
-  kinematics_solver: prbt_manipulator/IKFastKinematicsPlugin
-  kinematics_solver_search_resolution: 0.005
+  kinematics_solver: ["prbt_manipulator/IKFastKinematicsPlugin"]
+  kinematics_solver_search_resolution: [0.005]
```
tylerjw, line 8: I don't think you want to make these array types. For the first one the quotes are optional as the default type with characters in it is string when writing yaml.

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_parameters.yaml
```diff
@@ -0,0 +1,24 @@
+kinematics:
+  kinematics_solver: {
+    type: string_array,
+    default_value: ["kdl_kinematics_plugin/KDLKinematicsPlugin"],
```
tylerjw, line 4: I don't think you want this to be an array type. Was it an array type before?

Abishalini: It was not an array type before but multiple solvers could be listed using white spaces.
The original code parses the string and splits the string using white space - https://github.com/ros-planning/moveit2/blob/main/moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp#L327-L339

In examples, I have not seen more than one solver be listed

moveit_planners/test_configs/prbt_moveit_config/config/kinematics.yaml
Abishalini: I thought of making it array type since in code, it takes those parameters as string and splits for multiple solvers using white space. 
Even the search resolution is declared as a string if it needs to be different values for different solvers - https://github.com/ros-planning/moveit2/blob/main/moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp#L357-L372

tylerjw: To me this is an unnecessary extension point. @henningkayser what do you think of changing this so it only creates one solver / group.

tylerjw: If you make this an array (which is a much more correct way to specify multiple things in a parameter than a string with a space) it will break existing configs. I think I'd rather break configs that were using the space hack.

moveit_ros/planning/kinematics_plugin_loader/include/moveit/kinematics_plugin_loader/kinematics_plugin_loader.h
```diff
@@ -102,17 +80,14 @@ class KinematicsPluginLoader
 
 private:
   const rclcpp::Node::SharedPtr node_;
+  std::unordered_map<std::string, std::shared_ptr<kinematics::ParamListener>> kinematics_param_listener_;
+  std::unordered_map<std::string, kinematics::Params> kinematics_params_;
```
tylerjw, line 84: These are already inside the kinematics_plugin_loader, so I don't think you need to restate the name kinematics in the name for these. More importantly, I think the name should include what the key of the map is.

```suggestion
  std::unordered_map<std::string, std::shared_ptr<kinematics::ParamListener>> group_param_listener_;
  std::unordered_map<std::string, kinematics::Params> group_params_;
```

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_parameters.yaml
```diff
@@ -0,0 +1,30 @@
...
+    validation: {
+      lower_bounds<>: [ 0.0 ]
+    }
+  }
+  kinematics_solver_timeout: {
+    type: double,
+    default_value: 0.05,
+    description: "kinematics solver timeout",
+    validation: {
+      lower_bounds<>: [ 0.0 ]
```
tylerjw, line 23: Should 0 be a possible value or should the lower bounds for this be something slightly larger than zero? Would it be better to change the names of these validators to be something like:

* gt
* gt_eq
* lt
* lt_eq

Then you could specify that this is greater than 0 (but not 0).

tylerjw: https://github.com/PickNikRobotics/generate_parameter_library/pull/80/files

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_parameters.yaml
```diff
@@ -4,23 +4,30 @@ kinematics:
...
     validation: {
-      not_empty<>: []
+      not_empty<>: [],
+      one_of<>: [[kdl_kinematics_plugin/KDLKinematicsPlugin,
+                  prbt_manipulator/IKFastKinematicsPlugin,
+                  bio_ik/BioIKKinematicsPlugin,
+                  lma_kinematics_plugin/LMAKinematicsPlugin,
+                  srv_kinematics_plugin/SrvKinematicsPlugin,
+                  cached_ik_kinematics_plugin/CachedKDLKinematicsPlugin,
+                  cached_ik_kinematics_plugin/CachedSrvKinematicsPlugin ]]
```
tylerjw, line 14: This is a hard-ish problem. We want this to be an extension point of moveit but we also want to validate that the value set here is a valid kinematics_base plugin. I would probably leave this as just `not_empty` and I will work on writing a validator for you that uses pluginlib to validate that the string your receive here is a valid plugin and not just part of some hard coded set but instead plugins that your system enviroment has.

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_parameters.yaml
```diff
@@ -4,23 +4,30 @@ kinematics:
...
+      gt_eq<>: [ 0.0 ]
     }
   }
   kinematics_solver_timeout: {
     type: double,
     default_value: 0.05,
     description: "kinematics solver timeout",
     validation: {
-      lower_bounds<>: [ 0.0 ]
+      gt_eq<>: [ 0.0 ]
```
tylerjw, line 30: What does a timeout of 0.0 mean? If that has a special meaning please add that to the description. Otherwise I think this should just be `gt`.

Abishalini: Forgot to make this change. Thanks for catching!

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_parameters.yaml
```diff
@@ -0,0 +1,30 @@
...
+    description: "kinematics solver timeout",
+    validation: {
+      lower_bounds<>: [ 0.0 ]
+    }
+  }
+  kinematics_solver_ik_links: {
+    type: string_array,
+    default_value: [],
+    description: "TODO",
+  }
```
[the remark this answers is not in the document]
Abishalini: I'm going to go ahead and remove it since @henningkayser was not sure about its use.

tylerjw (approved):

PR #1574, opened 2022-09-16
[Servo] Remove the option for "stop distance"-based collision checking

AndyZe:
### Description

This was an experimental method for collision checking that was added back in MoveIt1 [here](https://github.com/ros-planning/moveit/pull/2100).

It didn't work very well in practice and I thought it had already been removed. It definitely isn't functional at this point (you can verify this by `grep -r worst_case_stop_time_pub_` -- you'll see that the publisher is never actually initialized.

or `grep -r collision_check_type` and you'll see that the parameter isn't actually read.


AndyZe:
CI will pass if #1556 gets merged

henningkayser (approved):

PR #1577, opened 2022-09-20
Use generate_parameter_library to load pilz cartesian limit parameters

Abishalini:
### Description

Use `generate_parameter_library` to load Cartesian limit parameters inside Pilz planner

TODO:
- [x] How to check if a parameter is set by the user? (i.e) user should override default or we could not set a default
- [x] Remove commented code
- [x] Delete `CartesianLimit` and `CartesianLimitsAggregator`classes
- [x] Does the cartesian limits need unit tests to check if they are set? - The generate parameter library would make sure to throw an error if the parameter is not set so remove the test

### Checklist


moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/pilz_industrial_motion_planner.h
```diff
@@ -138,6 +140,9 @@ class CommandPlanner : public planning_interface::PlannerManager
   pilz_industrial_motion_planner::JointLimitsContainer aggregated_limit_active_joints_;
 
   /// cartesian limit
+  std::shared_ptr<cartesian_limits::ParamListener> cartesian_limits_param_listener_;
+  cartesian_limits::Params cartesian_limits_params_;
```
tylerjw, line 144: These can probably just be named `param_listener_` and `params_` unless you plan on there being other listener or parmas struct in the private members of this class.

moveit_planners/pilz_industrial_motion_planner/test/unit_tests/src/unittest_cartesian_limits_aggregator.cpp
```diff
@@ -32,63 +32,63 @@
  *  POSSIBILITY OF SUCH DAMAGE.
```
henningkayser, line 32: This test seems somewhat obsolete with the new changes, no? Please, either comment with a TODO or remove.

moveit_planners/pilz_industrial_motion_planner/src/trajectory_generator_circ.cpp
```diff
@@ -58,11 +58,9 @@ TrajectoryGeneratorCIRC::TrajectoryGeneratorCIRC(const moveit::core::RobotModelC
                                                  const std::string& /*group_name*/)
   : TrajectoryGenerator::TrajectoryGenerator(robot_model, planner_limits)
 {
-  if (!planner_limits_.hasFullCartesianLimits())
-  {
-    throw TrajectoryGeneratorInvalidLimitsException(
-        "Cartesian limits are not fully set for CIRC trajectory generator.");
-  }
+  // TODO: Need to check if all the parameters are set up the user in cartesian_limits.yaml
```
henningkayser, line 61: Can't we just check if the parameters have an override?

henningkayser (commented):
Changes look good to me overall, only two comments before approval

moveit_planners/pilz_industrial_motion_planner/src/cartesian_limits_parameters.yaml
```diff
@@ -0,0 +1,25 @@
...
+    # default_value: ,
+    description: "Max translation decceleration",
+    # validation:
+  }
+  max_rot_vel: {
+    type: double,
+    # default_value: ,
+    description: "Max rotational velocity",
+    # validation:
+  }
```
tylerjw, line 25: Do these have any defaults or validations that would make sense for them?

moveit_planners/pilz_industrial_motion_planner/src/trajectory_generator_circ.cpp
tylerjw: The current way she's done it with no defaults will fail to load if the user does not specify all the required parameters.

moveit_planners/pilz_industrial_motion_planner/test/unit_tests/src/unittest_cartesian_limits_aggregator.cpp
Abishalini: I have removed the tests

moveit_planners/pilz_industrial_motion_planner/test/unit_tests/src/unittest_trajectory_generator_circ.cpp
```diff
@@ -271,12 +272,12 @@ TEST_F(TrajectoryGeneratorCIRCTest, TestExceptionErrorCodeMapping)
...
-  LimitsContainer planner_limits;
-  EXPECT_THROW(TrajectoryGeneratorCIRC(this->robot_model_, planner_limits, planning_group_),
-               TrajectoryGeneratorInvalidLimitsException);
-}
+// TEST_F(TrajectoryGeneratorCIRCTest, noLimits)
+// {
+//   LimitsContainer planner_limits;
+//   EXPECT_THROW(TrajectoryGeneratorCIRC(this->robot_model_, planner_limits, planning_group_),
+//                TrajectoryGeneratorInvalidLimitsException);
+// }
```
tylerjw, line 280: If it makes sense to remove this test, please remove it instead of commenting it.

moveit_planners/pilz_industrial_motion_planner/src/cartesian_limits_parameters.yaml
Abishalini: I was thinking of leaving the defaults empty so that the user would be forced to set values. Do you think that's a good approach?
Otherwise I can default the defaults to 0 and figure out another way to ensure the use overrides them

tylerjw (approved):
Sweet, thank you for doing this!

moveit_planners/pilz_industrial_motion_planner/src/cartesian_limits_parameters.yaml
tylerjw: My comment was more that we should not leave commented-out lines here. I was asking if you planned on adding defaults. If we need the user to specify the values, not having a default is the right choice.

tylerjw (approved):

tylerjw (approved):

PR #1581, opened 2022-09-26
[RViz] Check valid interactive marker pointer in Motion Planning Display when changing fixed frame

sea-bass:
### Description

While going through the [MoveIt QuickStart in RViz Tutorial](https://moveit.picknik.ai/humble/doc/tutorials/quickstart_in_rviz/quickstart_in_rviz_tutorial.html), I was getting RViz crashing when I changed the fixed frame as shown here:

image

I chased this down to `MotionPlanningDisplay::fixedFrameChanged()` in `moveit2/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp` and found that the issue was resolved if I checked that the pointers `frame_->scene_marker_` were non-null.

### Checklist


moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp
```diff
@@ -1422,8 +1422,9 @@ void MotionPlanningDisplay::fixedFrameChanged()
   if (int_marker_display_)
     int_marker_display_->setFixedFrame(fixed_frame_);
   // When the fixed frame changes we need to tell RViz to update the rendered interactive marker display
-  frame_->scene_marker_->requestPoseUpdate(frame_->scene_marker_->getPosition(),
-                                           frame_->scene_marker_->getOrientation());
+  if (frame_ && frame_->scene_marker_)
+    frame_->scene_marker_->requestPoseUpdate(frame_->scene_marker_->getPosition(),
+                                             frame_->scene_marker_->getOrientation());
```
tylerjw, line 1427: Should we log something when this is false? Ideally, we'd send the error some other way... I generally don't like reporting errors in logs if we can help it but it is definitely better to not dereference an invalid pointer than what is here.

sea-bass: I took a deeper look, and `frame_` is valid but `frame->scene_marker_` is not.

This is totally fine given how `MotionPlanningFrame::createSceneInteractiveMarker()` in `moveit2/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp` doesn't always instantiate a scene marker -- in fact, this code path only makes a marker if we have a collision object selected in the scene, as in here:

image

Also, only a few lines above we have a similar valid pointer check for the `int_marker_display_` variable (which is the interactive marker on the move group's end effector frame) and we don't log anything on that pointer being null.

tylerjw: Thank you for the detailed response.

tylerjw (approved):

vatanaksoytezer (approved):

MarcoMagriDev:
I have encountered the same issue using `humble` branch and I noticed that this PR has not been backported to that branch [see here](https://github.com/ros-planning/moveit2/blob/b679dd870031ad86f8fb5c58df42f7649ab64fc2/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp#L1424).

Could you please backport this PR to `humble`?

sea-bass:
@MarcoMagriDev good call -- opened up https://github.com/ros-planning/moveit2/pull/2366

PR #1583, opened 2022-09-28
Do not add Pilz parameters to MoveIt Configs Utils if Pilz is not used

stephanie-eng:
### Description
After #1571, if MoveIt Configs Utils is used and `self.pilz_cartesian_limits` is not set (which can happen if not using Pilz), I get this error when  launching

`[ERROR] [launch]: Caught exception in launch (see debug for traceback): Caught exception when trying to load file of format [py]: 'robot_description_planning'`

This can be seen in https://github.com/ros-planning/moveit2_tutorials/pull/492.

This PR adds a condition checking if `self.pilz_cartesian_limits` is populated before adding parameters.


sjahr (approved):

stephanie-eng:
Potentially related issue: https://github.com/ros-planning/moveit2_tutorials/issues/518

Issue #1474, opened 2022-08-03
MGI CallbackGroup doesn't trigger when running external Executor + Fast-DDS


AndyZe:
### Description

The OMPL constrained planning tutorial [here](https://github.com/ros-planning/moveit2_tutorials/blob/main/doc/how_to_guides/using_ompl_constrained_planning/src/ompl_constrained_planning_tutorial.cpp) no longer works.

I've traced it down to this:

```
  rclcpp::NodeOptions node_options;
  node_options.automatically_declare_parameters_from_overrides(true);
  auto node = rclcpp::Node::make_shared("ompl_constrained_planning_demo_node", node_options);

  rclcpp::executors::SingleThreadedExecutor executor;
  executor.add_node(node);
  auto spinner = std::thread([&executor]() { executor.spin(); });
  
  moveit::planning_interface::MoveGroupInterface move_group_interface(node, "panda_arm");
  auto moveit_visual_tools =
      moveit_visual_tools::MoveItVisualTools{ node, "panda_link0", rviz_visual_tools::RVIZ_MARKER_TOPIC,
                                              move_group_interface.getRobotModel() };
      ...
   success = (move_group_interface.plan(plan) == moveit::core::MoveItErrorCode::SUCCESS);
   RCLCPP_INFO(LOGGER, "Plan with plane constraint %s", success ? "" : "FAILED");   // Never prints
```

The last line does not print. In other words, the `mgi.plan()` call is blocking. I suspect this has to do with conflicting callback groups.

I tried  a multi-threaded executor as well as two separate nodes.

I'm suspicious that this PR caused the bug:  https://github.com/ros-planning/moveit2/pull/1305

AndyZe:
Indeed, rolling back #1305 prevents the issue.

henningkayser:
Does MoveGroup receive the planning call and does it return a result?

AndyZe:
Yes, planning succeeds. I'm not sure if it tries to return a value or not. (If it does try to return a value, it doesn't make it.) Easy to test with the tutorial

henningkayser:
What do you mean with "parallel with 'other things'"? Running MGI requests from multiple threads?

AndyZe:
`rviz_visual_tools` in this case. You can look at the linked tutorial code

henningkayser:
[...]

PR #1589, opened 2022-09-29
Fix logic with enforcing constrained planning state space

### LINKED ISSUE #1552: Mixed constraints support issue
Based on what stephanie-eng:seng/multiple-constraints(https://github.com/ros-planning/moveit2/pull/1319) has, the mixed constraints should be supported, but the code looks strange to me. In the planning_context_manager.cpp, line 539 to line [544,](https://github.com/stephanie-eng/moveit2/blob/68da67770605628c1043abaf9c60c2d331951036/moveit_planners/ompl/ompl_interface/src/planning_context_manager.cpp#L542), it still said "Mixed constraints are not supported", and it only allows one constraint based on the if statement.
`  if (constrained_planning_iterator != pc->second.config.end() &&
      boost::lexical_cast<bool>(constrained_planning_iterator->second) &&
      ((req.path_constraints.position_constraints.size() == 1) !=
       (req.path_constraints.orientation_constraints.size() == 1)))`

_Originally posted by @jih189 in https://github.com/ros-planning/moveit2/issues/1319#issuecomment-1237524367_

stephanie-eng:
### Description

Fixes #1552, change to logic when enforcing constrained planning state space in the mixed constraint case.

When running the constrained planning tutorial with:
`ros2 launch moveit2_tutorials ompl_constrained_planning.launch.py`

You should see the following log for the last example (required debug log level):
`[moveit.ompl_planning.planning_context_manager]: planning_context_manager: Using OMPL's constrained state space for planning.`

### DIFF
--- a/moveit_planners/ompl/ompl_interface/src/planning_context_manager.cpp
+++ b/moveit_planners/ompl/ompl_interface/src/planning_context_manager.cpp
@@ -536,11 +536,10 @@ ModelBasedPlanningContextPtr PlanningContextManager::getPlanningContext(
   auto constrained_planning_iterator = pc->second.config.find("enforce_constrained_state_space");
   auto joint_space_planning_iterator = pc->second.config.find("enforce_joint_model_state_space");
 
-  // Use ConstrainedPlanningStateSpace if there is exactly one position constraint or one orientation constraint
-  // Mixed constraints are not supported
+  // Use ConstrainedPlanningStateSpace if there is exactly one position constraint and/or one orientation constraint
   if (constrained_planning_iterator != pc->second.config.end() &&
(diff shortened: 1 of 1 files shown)


sjahr:
Does the tutorial work for you? I tried to test your change but the robot never moved and I got stuck with the first example. Same happened with the last one (after I commented the ones before out) and I did not see the print statement you described above

stephanie-eng:
The tutorial doesn't work as per https://github.com/ros-planning/moveit2/issues/1474#issuecomment-1227509627, but Andy has a fix here: https://github.com/ros-planning/moveit2_tutorials/pull/519

The print statement should appear as DEBUG, so you'll need to set your log level accordingly to see it.

henningkayser (approved):
I thought we had this fixed a while ago?

PR #1591, opened 2022-09-29
Consistent orientation constraint definition

### LINKED ISSUE #1586: Orientation constraints are inconsistent for path constraints vs goal constraints
### Description

Re. the OrientationConstraint msg (https://github.com/ros-planning/moveit_msgs/blob/ros2/msg/OrientationConstraint.msg) I just noticed that it's handled differently throughout the codebase.

Here in[ ompl_constraints.h](https://github.com/ros-planning/moveit2/blob/7fec51b564be7d1076de57869f26833c7ba767d3/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/detail/ompl_constraints.h#L368) it's:

`absolute_x_axis_tolerance / 2 < error[0] < absolute_x_axis_tolerance / 2`

But [here](https://github.com/ros-planning/moveit/blob/327b5b09cc48b97ba3f2a9ac2acf1edcd1bb0650/moveit_core/kinematic_constraints/src/kinematic_constraint.cpp#L709) in kinematic_constraint.cpp it's:

```
  else if (parameterization_type_ == moveit_msgs::OrientationConstraint::ROTATION_VECTOR)
  {
    Eigen::AngleAxisd aa(diff.linear());
    xyz_rotation = aa.axis() * aa.angle();
    xyz_rotation(0) = fabs(xyz_rotation(0));
    xyz_rotation(1) = fabs(xyz_rotation(1));
    xyz_rotation(2) = fabs(xyz_rotation(2));
  }
  ...
    bool result = xyz_rotation(2) < absolute_z_axis_tolerance_ + std::numeric_limits<double>::epsilon() &&
                xyz_rotation(1) < absolute_y_axis_tolerance_ + std::numeric_limits<double>::epsilon() &&
                xyz_rotation(0) < absolute_x_axis_tolerance_ + std::numeric_limits<double>::epsilon();
```

Which (I believe) is equivalent to -tol / +tol

AndyZe:
### Description

Fixes #1586. (There's a full description there)

### DIFF
--- a/moveit_core/kinematic_constraints/src/kinematic_constraint.cpp
+++ b/moveit_core/kinematic_constraints/src/kinematic_constraint.cpp
@@ -724,9 +724,9 @@ ConstraintEvaluationResult OrientationConstraint::decide(const moveit::core::Rob
     RCLCPP_ERROR(LOGGER, "The parameterization type for the orientation constraints is invalid.");
   }
 
-  bool result = xyz_rotation(2) < absolute_z_axis_tolerance_ + std::numeric_limits<double>::epsilon() &&
-                xyz_rotation(1) < absolute_y_axis_tolerance_ + std::numeric_limits<double>::epsilon() &&
-                xyz_rotation(0) < absolute_x_axis_tolerance_ + std::numeric_limits<double>::epsilon();
+  bool result = xyz_rotation(2) < absolute_z_axis_tolerance_ / 2 + std::numeric_limits<double>::epsilon() &&
+                xyz_rotation(1) < absolute_y_axis_tolerance_ / 2 + std::numeric_limits<double>::epsilon() &&
+                xyz_rotation(0) < absolute_x_axis_tolerance_ / 2 + std::numeric_limits<double>::epsilon();
 
   if (verbose)
   {


henningkayser:
What do you think about this comment from [here](https://github.com/ros-planning/moveit2/blob/main/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/detail/ompl_constraints.h#L372)?
```
 * **IMPORTANT** It is NOT how orientation error is handled in the default MoveIt constraint samplers, where XYZ
 * intrinsic euler angles are used. Using exponential coordinates is analog to how orientation error is calculated in
 * the TrajOpt  motion planner.
 ```

The exponentiated OMPL constraints are a special case. I don't see why we should change the MoveIt constraints over that. Dividing by 2 doesn't make much sense to me, since the absolute tolerance describes the absolute tolerance from your target, not half the absolute tolerance.

**Comment by AndyZe** [comment]
OK. Well, notice that the message definition says this:

```
# Tolerance on the three vector components of the orientation error (optional)
# This is a total tolerance, e.g. (+ x_axis_tol / 2, - x_axis_tol / 2)
```

Mostly I just wants things to be consistent and documented.

Shall I update the msg definition to say this instead, and make a PR against `ompl_constraints.h`?

```
# Tolerance on the three vector components of the orientation error (optional)
# This is a +/- tolerance, e.g. (+ x_axis_tol, - x_axis_tol)
```

AndyZe:
> the absolute tolerance describes the absolute tolerance from your target, not half the absolute tolerance.

"Absolute" as opposed to "relative". AFAIK it doesn't have anything to do with dividing by 2. That's why a little extra comment in the message definition is important.

AndyZe:
It looks like there would be a lot of `kinematic_constraint` unit tests to change, anwyay. OK, closing this PR in favor of #1592

PR #1601, opened 2022-10-05
Rename MoveItControllerManager. Add deprecation warning

AndyZe:
### Description

We had a talk about what is confusing with MoveIt controller managers in July. The only action item that came out of it was renaming `MoveItControllerManager->Ros2ControlManager` and `MoveItMultiControllerManager->Ros2ControlMultiManager`.

The reason for this is to distinguish these 2 options from the `MoveItSimpleControllerManager`, which is a basic action interface and should be used when `ros2_control` isn't involved.

### Merge with...

moveit2_tutorials PR:  https://github.com/ros-planning/moveit2_tutorials/pull/526


mikeferguson:
Do we need to make any updates to the setup assistant, due to the name changes?

moveit_core/controller_manager/include/moveit/controller_manager/controller_manager.h
```diff
@@ -147,14 +147,14 @@ class MoveItControllerHandle
...
-MOVEIT_CLASS_FORWARD(MoveItControllerManager);  // Defines MoveItControllerManagerPtr, ConstPtr, WeakPtr... etc
+MOVEIT_CLASS_FORWARD(Ros2ControlManager);  // Defines Ros2ControlManagerPtr, ConstPtr, WeakPtr... etc
 
 /** @brief MoveIt does not enforce how controllers are implemented.
     To make your controllers usable by MoveIt, this interface needs to be implemented.
     The main purpose of this interface is to expose the set of known controllers and
     potentially to allow activating and deactivating them, if multiple controllers are available.
  */
-class MoveItControllerManager
+class Ros2ControlManager
```
JafarAbdi, line 157: Why change the base class name? This class has nothing to do with ros2 control

AndyZe:
@mikeferguson I looked at the MSA code. It defaults to a MoveItSimpleControllerManager by default so I don't think it's affected.

https://github.com/ros-planning/moveit2/blob/5c3f98ee4733829154e5de395f9d6eee0b7fd2d9/moveit_setup_assistant/moveit_setup_controllers/src/moveit_controllers_config.cpp#L164

@JafarAbdi thanks, I'll revert that.

gavanderhoorn:
`Ros2ControlControllerManager`?

AndyZe:
@gavanderhoorn that's too redundant for my liking but I'll change it if others like it

gavanderhoorn:
I see what you're saying but it does seem more "correct": it's managing `ros2_control` controllers. Not `ros2_control` itself. Or is it?

AndyZe:
> Or is it?

You're right, it's making service calls to switch controllers.

moveit_plugins/moveit_ros_control_interface/moveit_core_plugins.xml
```diff
@@ -1,12 +1,12 @@
 <class_libraries>
   <library path="moveit_ros_control_interface_plugin">
-    <class name="moveit_ros_control_interface/MoveItControllerManager"
-           type="moveit_ros_control_interface::MoveItControllerManager"
+    <class name="moveit_ros_control_interface/Ros2ControlManager"
```
JafarAbdi, line 3: I think you need to add the following lines; otherwise it will break current setups
 
```xml
<class name="moveit_ros_control_interface/MoveItControllerManager"
       type="moveit_ros_control_interface::Ros2ControlManager"
       base_class_type="moveit_controller_manager::MoveItControllerManager">
    <description>ros_control controller manager interface for MoveIt</description>
</class>
<class name="moveit_ros_control_interface/MoveItMultiControllerManager"
       type="moveit_ros_control_interface::Ros2ControlMultiManager"
       base_class_type="moveit_controller_manager::MoveItControllerManager">
    <description>multiple ros_control controller managers interface for MoveIt</description>
</class>
```

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -133,6 +133,18 @@ void TrajectoryExecutionManager::initialize()
       }
     }
 
+    // Deprecation errors, October 2022
+    if (controller == "moveit_ros_control_interface/MoveItControllerManager")
+    {
+      RCLCPP_FATAL(LOGGER, "moveit_ros_control_interface/MoveItControllerManager is deprecated. Replace with "
```
JafarAbdi, line 139: nit: use warning here?

moveit_plugins/moveit_ros_control_interface/moveit_core_plugins.xml
AndyZe: We're thinking about this differently. My thought is, we change the plugin name and force users to adapt. There probably aren't many users out there using `MoveItControllerManager` on Rolling. Maybe just your client and PickNik.

The name `MoveItControllerManager` is misleading so I think we should stop using it.

AndyZe: Please disregard my previous comment. I understand now.

JafarAbdi (approved):

PR #1605, opened 2022-10-06
Planning request adapters: short-circuit if failure, return code rather than bool

AndyZe:
### Description

Previously, if one of the planning request adapters failed, the pipeline would continue onward anyway. Also, only a `bool` was returned, although `moveit_msgs::MoveItErrorCodes.msg` has some enums to report more detail.

This PR short-circuits in case there is an adapter failure and reports the actual error code.

For example, here's the output when there are 6 adapters and `FixStartStateCollision` fails:

```
[moveit_cpp_tutorial-4] [ERROR] [1665063576.588225372] [moveit_ros.fix_start_state_collision]: Running 'Fix Start State In Collision'
[moveit_cpp_tutorial-4] [WARN] [1665063576.588230842] [moveit_ros.fix_start_state_collision]: Unable to find a valid state nearby the start state (using jiggle fraction of 0.020000 and 100 sampling attempts).
[moveit_cpp_tutorial-4] [ERROR] [1665063576.588242808] [moveit.planning_request_adapter]: Adapter  result: -10
[moveit_cpp_tutorial-4] [ERROR] [1665063576.588261618] [moveit.planning_request_adapter]: Adapter  result: -10
[moveit_cpp_tutorial-4] [ERROR] [1665063576.588265353] [moveit.planning_request_adapter]: Adapter  result: -10
[moveit_cpp_tutorial-4] [ERROR] [1665063576.588267620] [moveit.planning_request_adapter]: Adapter  result: -10
[moveit_cpp_tutorial-4] [ERROR] [1665063576.588291731] [moveit.planning_request_adapter]: Adapter  result: -10
[moveit_cpp_tutorial-4] [ERROR] [1665063576.588300902] [moveit.ros_planning.planning_pipeline]: Res return: -10
[moveit_cpp_tutorial-4] [ERROR] [1665063576.588302635] [moveit.ros_planning_interface.planning_component]: Could not compute plan successfully
```

A code of -10 means ["START_STATE_IN_COLLISION"](https://github.com/ros-planning/moveit_msgs/blob/master/msg/MoveItErrorCodes.msg)  :+1:

Prior to this PR, the output would be:

```
[moveit_cpp_tutorial-4] [ERROR] [1665067792.760674407] [moveit.planning_request_adapter]: Adapter result: 0
[moveit_cpp_tutorial-4] [ERROR] [1665067792.760687109] [moveit.planning_request_adapter]: Adapter result: 0
[moveit_cpp_tutorial-4] [ERROR] [1665067792.760691608] [moveit.planning_request_adapter]: Adapter result: 0
[moveit_cpp_tutorial-4] [ERROR] [1665067792.760693864] [moveit.planning_request_adapter]: Adapter result: 0
[moveit_cpp_tutorial-4] [ERROR] [1665067792.760695824] [moveit.planning_request_adapter]: Adapter result: 0
[moveit_cpp_tutorial-4] [ERROR] [1665067792.760701750] [moveit.ros_planning.planning_pipeline]: Res return: 0
[moveit_cpp_tutorial-4] [ERROR] [1665067792.760704577] [moveit.ros_planning_interface.planning_component]: Could not compute plan successfully
```

Which is not very useful for debugging.

### Testing this PR

I've been testing with the [move_group_interface tutorial](https://moveit.picknik.ai/main/doc/examples/move_group_interface/move_group_interface_tutorial.html). You can publish collision objects conveniently from the GUI:

`ros2 launch moveit2_tutorials move_group.launch.py`

`ros2 launch moveit2_tutorials move_group_interface_tutorial.launch.py`

This image shows a START_STATE_IN_COLLISION result code.

box_collision

### DIFF
--- a/moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
+++ b/moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
@@ -74,25 +74,27 @@ public:
     return "";
   }
 
-  bool adaptAndPlan(const planning_interface::PlannerManagerPtr& planner,
-                    const planning_scene::PlanningSceneConstPtr& planning_scene,
-                    const planning_interface::MotionPlanRequest& req,
-                    planning_interface::MotionPlanResponse& res) const;
+  moveit::core::MoveItErrorCode adaptAndPlan(const planning_interface::PlannerManagerPtr& planner,
+                                             const planning_scene::PlanningSceneConstPtr& planning_scene,
+                                             const planning_interface::MotionPlanRequest& req,
+                                             planning_interface::MotionPlanResponse& res) const;
 
-  bool adaptAndPlan(const planning_interface::PlannerManagerPtr& planner,
-                    const planning_scene::PlanningSceneConstPtr& planning_scene,
-                    const planning_interface::MotionPlanRequest& req, planning_interface::MotionPlanResponse& res,
-                    std::vector<std::size_t>& added_path_index) const;
+  moveit::core::MoveItErrorCode adaptAndPlan(const planning_interface::PlannerManagerPtr& planner,
+                                             const planning_scene::PlanningSceneConstPtr& planning_scene,
+                                             const planning_interface::MotionPlanRequest& req,
+                                             planning_interface::MotionPlanResponse& res,
+                                             std::vector<std::size_t>& added_path_index) const;
 
   /** \brief Adapt the planning request if needed, call the planner
       function \e planner and update the planning response if
       needed. If the response is changed, the index values of the
       states added without planning are added to \e
       added_path_index */
-  virtual bool adaptAndPlan(const PlannerFn& planner, const planning_scene::PlanningSceneConstPtr& planning_scene,
-                            const planning_interface::MotionPlanRequest& req,
-                            planning_interface::MotionPlanResponse& res,
-                            std::vector<std::size_t>& added_path_index) const = 0;
+  virtual moveit::core::MoveItErrorCode adaptAndPlan(const PlannerFn& planner,
+                                                     const planning_scene::PlanningSceneConstPtr& planning_scene,
+                                                     const planning_interface::MotionPlanRequest& req,
+                                                     planning_interface::MotionPlanResponse& res,
+                                                     std::vector<std::size_t>& added_path_index) const = 0;
 
 protected:
   /** \brief Helper param for getting a parameter using a namespace **/
@@ -129,15 +131,16 @@ public:
     adapters_.push_back(adapter);
   }
 
-  bool adaptAndPlan(const planning_interface::PlannerManagerPtr& planner,
-                    const planning_scene::PlanningSceneConstPtr& planning_scene,
-                    const planning_interface::MotionPlanRequest& req,
-                    planning_interface::MotionPlanResponse& res) const;
+  moveit::core::MoveItErrorCode adaptAndPlan(const planning_interface::PlannerManagerPtr& planner,
+                                             const planning_scene::PlanningSceneConstPtr& planning_scene,
+                                             const planning_interface::MotionPlanRequest& req,
+                                             planning_interface::MotionPlanResponse& res) const;
 
-  bool adaptAndPlan(const planning_interface::PlannerManagerPtr& planner,
-                    const planning_scene::PlanningSceneConstPtr& planning_scene,
-                    const planning_interface::MotionPlanRequest& req, planning_interface::MotionPlanResponse& res,
-                    std::vector<std::size_t>& added_path_index) const;
+  moveit::core::MoveItErrorCode adaptAndPlan(const planning_interface::PlannerManagerPtr& planner,
+                                             const planning_scene::PlanningSceneConstPtr& planning_scene,
+                                             const planning_interface::MotionPlanRequest& req,
+                                             planning_interface::MotionPlanResponse& res,
+                                             std::vector<std::size_t>& added_path_index) const;
 
 private:
   std::vector<PlanningRequestAdapterConstPtr> adapters_;
--- a/moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
+++ b/moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
@@ -52,7 +52,7 @@ class PlanningRequestAdapter
 {
 public:
   using PlannerFn =
-      std::function<bool(const planning_scene::PlanningSceneConstPtr&, const planning_interface::MotionPlanRequest&,
+      std::function<void(const planning_scene::PlanningSceneConstPtr&, const planning_interface::MotionPlanRequest&,
                          planning_interface::MotionPlanResponse&)>;
 
   PlanningRequestAdapter()
--- a/moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
+++ b/moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
@@ -39,6 +39,7 @@
 #include <moveit/macros/class_forward.h>
 #include <moveit/planning_interface/planning_interface.h>
 #include <moveit/planning_scene/planning_scene.h>
+#include <moveit/utils/moveit_error_code.h>
 #include <functional>
 #include <rclcpp/logging.hpp>
 #include <rclcpp/node.hpp>
@@ -52,7 +53,7 @@ class PlanningRequestAdapter
 {
 public:
   using PlannerFn =
-      std::function<void(const planning_scene::PlanningSceneConstPtr&, const planning_interface::MotionPlanRequest&,
+      std::function<moveit::core::MoveItErrorCode(const planning_scene::PlanningSceneConstPtr&, const planning_interface::MotionPlanRequest&,
                          planning_interface::MotionPlanResponse&)>;
 
   PlanningRequestAdapter()
--- a/moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
+++ b/moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
@@ -52,9 +52,9 @@ MOVEIT_CLASS_FORWARD(PlanningRequestAdapter);  // Defines PlanningRequestAdapter
 class PlanningRequestAdapter
 {
 public:
-  using PlannerFn =
-      std::function<moveit::core::MoveItErrorCode(const planning_scene::PlanningSceneConstPtr&, const planning_interface::MotionPlanRequest&,
-                         planning_interface::MotionPlanResponse&)>;
+  using PlannerFn = std::function<moveit::core::MoveItErrorCode(const planning_scene::PlanningSceneConstPtr&,
+                                                                const planning_interface::MotionPlanRequest&,
+                                                                planning_interface::MotionPlanResponse&)>;
 
   PlanningRequestAdapter()
   {
--- a/moveit_core/planning_request_adapter/src/planning_request_adapter.cpp
+++ b/moveit_core/planning_request_adapter/src/planning_request_adapter.cpp
@@ -57,25 +57,23 @@ bool callPlannerInterfaceSolve(const planning_interface::PlannerManager& planner
     return false;
 }
 
-bool callAdapter(const PlanningRequestAdapter& adapter,
-                                                   const PlanningRequestAdapter::PlannerFn& planner,
-                                                   const planning_scene::PlanningSceneConstPtr& planning_scene,
-                                                   const planning_interface::MotionPlanRequest& req,
-                                                   std::vector<std::size_t>& added_path_index)
+void callAdapter(const PlanningRequestAdapter& adapter, const PlanningRequestAdapter::PlannerFn& planner,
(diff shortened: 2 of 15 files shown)


moveit_core/planning_request_adapter/src/planning_request_adapter.cpp
```diff
@@ -140,7 +144,12 @@ bool PlanningRequestAdapterChain::adaptAndPlan(const planning_interface::Planner
       fn = [&adapter = *adapters_[i], fn, &added_path_index = added_path_index_each[i]](
                const planning_scene::PlanningSceneConstPtr& scene, const planning_interface::MotionPlanRequest& req,
                planning_interface::MotionPlanResponse& res) {
-        return callAdapter(adapter, fn, scene, req, res, added_path_index);
+        res = callAdapter(adapter, fn, scene, req, added_path_index);
+        // If any adapter fails, return false
+        if (res.error_code_.val != moveit_msgs::msg::MoveItErrorCodes::SUCCESS)
+        {
+          return false;
+        }
```
AndyZe, line 152: This short-circuits if any of the adapters fail.

moveit_core/planning_request_adapter/src/planning_request_adapter.cpp
```diff
@@ -140,7 +144,9 @@ bool PlanningRequestAdapterChain::adaptAndPlan(const planning_interface::Planner
       fn = [&adapter = *adapters_[i], fn, &added_path_index = added_path_index_each[i]](
                const planning_scene::PlanningSceneConstPtr& scene, const planning_interface::MotionPlanRequest& req,
                planning_interface::MotionPlanResponse& res) {
-        return callAdapter(adapter, fn, scene, req, res, added_path_index);
+        res = callAdapter(adapter, fn, scene, req, added_path_index);
+        // If any adapter fails, return false
```
sjahr, line 148: Can you add another sentence here that documents for future code readers why this is necessary? Something like:
```suggestion
        // If any adapter fails, return false. Otherwise, if one of the planning request adapters failed, the pipeline would continue onward anyway and the failure error code would get lost.
```

sjahr (commented):
One small comment, besides that it looks good

moveit_core/planning_request_adapter/src/planning_request_adapter.cpp
AndyZe: done!

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_path_constraints.cpp
```diff
@@ -119,19 +120,22 @@ class FixStartStatePathConstraints : public planning_request_adapter::PlanningRe
           return true;
         }
         else
+        {
```
henningkayser, line 123: This block is redundant with the other error case below. Maybe move them after the if/else conditions so that it's used as the last default case.

moveit_core/planning_request_adapter/src/planning_request_adapter.cpp
```diff
@@ -140,7 +144,10 @@ bool PlanningRequestAdapterChain::adaptAndPlan(const planning_interface::Planner
       fn = [&adapter = *adapters_[i], fn, &added_path_index = added_path_index_each[i]](
                const planning_scene::PlanningSceneConstPtr& scene, const planning_interface::MotionPlanRequest& req,
                planning_interface::MotionPlanResponse& res) {
-        return callAdapter(adapter, fn, scene, req, res, added_path_index);
+        res = callAdapter(adapter, fn, scene, req, added_path_index);
+        // If any adapter fails, return false. Otherwise, the pipeline would continue onward anyway and the failure
```
henningkayser, line 148: No need to document the old behavior now that it's fixed. Simply "Abort pipeline and return error code in case of failure" is sufficient, imo.

moveit_core/planning_request_adapter/src/planning_request_adapter.cpp
```diff
@@ -57,21 +57,25 @@ bool callPlannerInterfaceSolve(const planning_interface::PlannerManager& planner
...
+    adapter.adaptAndPlan(planner, planning_scene, req, res, added_path_index);
+    return res;
   }
   catch (std::exception& ex)
   {
     RCLCPP_ERROR(LOGGER, "Exception caught executing adapter '%s': %s\nSkipping adapter instead.",
                  adapter.getDescription().c_str(), ex.what());
     added_path_index.clear();
-    return planner(planning_scene, req, res);
+    planner(planning_scene, req, res);
```
henningkayser, line 77: Are we omitting any useful bool return value here? `res` should always include success and error code, correct?

henningkayser (commented):
Thank you, this is really a useful improvement!

moveit_core/planning_request_adapter/src/planning_request_adapter.cpp
AndyZe: Well, `res` always contains a `MoveItErrorCodes error_code` that indicates success or failure. https://github.com/ros-planning/moveit_msgs/blob/ros2/msg/MotionPlanResponse.msg

I don't see any useful bool from `planner(planning_scene, req, res);` -- this just updates the arguments to the `planner` function. It doesn't return anything.

AndyZe: I contemplated adding something like this to check the boolean return of `planner(planning_scene, req, res);`:

```
    bool plan_result = planner(planning_scene, req, res);
    if (!plan_result)
    {
      res.error_code_.val = moveit_msgs::msg::MoveItErrorCodes::FAILURE;
    }
```

The problem I see is that this would overwrite a more specific error code with a generic FAILURE.

I guess the proper way to fix this is like:

```
  using PlannerFn =
      std::function<bool(const planning_scene::PlanningSceneConstPtr&, const planning_interface::MotionPlanRequest&,
                         planning_interface::MotionPlanResponse&)>;
```

Should become:

```
  using PlannerFn =
      std::function<void(const planning_scene::PlanningSceneConstPtr&, const planning_interface::MotionPlanRequest&,
                         planning_interface::MotionPlanResponse&)>;
```

(`MotionPlanResponse` ought to be passed by reference so it retains a value from the previous adapter.)

AndyZe: Changing the return type of PlannerFn caused quite a few changes because of all the plugins that use it. We can just revert this commit if you don't think it's worthwhile. I do, though.

moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
```diff
@@ -52,7 +52,7 @@ class PlanningRequestAdapter
 {
 public:
   using PlannerFn =
-      std::function<bool(const planning_scene::PlanningSceneConstPtr&, const planning_interface::MotionPlanRequest&,
+      std::function<void(const planning_scene::PlanningSceneConstPtr&, const planning_interface::MotionPlanRequest&,
```
AndyZe, line 55: Success/failure will be reported in the MotionPlanResponse now so it's not ambiguous

moveit_core/planning_request_adapter/src/planning_request_adapter.cpp
tylerjw: Should the response be the return value of the function and not an out parameter? That would make it clear that the response replaces the bool and is where the user should look for success / failure.

tylerjw: If the response is also an input (I'm sorry I haven't read the code yet, only this conversation) you could still express that by having it be both an input parameter and a return value. I doubt this function is in the hot loop of the problem and copying the `planning_interface::MotionPlanResponse` to make it more obvious to the user what is the input and output seems like an ok compromise to me.

moveit_core/planning_request_adapter/src/planning_request_adapter.cpp
```diff
@@ -157,7 +163,7 @@ bool PlanningRequestAdapterChain::adaptAndPlan(const planning_interface::Planner
         added_path_index.push_back(added_index);
       }
     std::sort(added_path_index.begin(), added_path_index.end());
-    return result;
+    return res.error_code_.val == moveit_msgs::msg::MoveItErrorCodes::SUCCESS;
```
tylerjw, line 166: I do really like the direction of this change. Even better would be to have all these functions return the error type instead of the bool, but I know that is probably an even larger change.

moveit_core/planning_request_adapter/src/planning_request_adapter.cpp
sjahr: Although I'd prefer returning something over passing it as a mutable reference I think passing it as input + returning it is a not very intuitive. Why not merge it this way and open an issue to perform the large refactoring to return a value & remove the input parameter?

henningkayser: Couldn't we simply return a `MoveItErrorCode` which [converts to bool](https://github.com/ros-planning/moveit2/blob/main/moveit_core/utils/include/moveit/utils/moveit_error_code.h#L62) and keep the short evaluations? Right now, the bool result seems to not be used at all, as an alternative I'd prefer `void` and keep an issue open.

AndyZe: The tricky thing here is that `res` is used recursively [here](https://github.com/ros-planning/moveit2/blob/749c5ae0777d093fad1be6514bae2bc188fc6926/moveit_core/planning_request_adapter/src/planning_request_adapter.cpp#L143). So we NEED to pass it by mutable reference.

I'm trying to do what Henning suggests by returning a `MoveItErrorCode` in addition, but no luck so far.

AndyZe: OK, I believe this is resolved now

sjahr (approved):

moveit_core/planning_request_adapter/src/planning_request_adapter.cpp
```diff
@@ -140,24 +142,28 @@ bool PlanningRequestAdapterChain::adaptAndPlan(const planning_interface::Planner
       fn = [&adapter = *adapters_[i], fn, &added_path_index = added_path_index_each[i]](
                const planning_scene::PlanningSceneConstPtr& scene, const planning_interface::MotionPlanRequest& req,
                planning_interface::MotionPlanResponse& res) {
-        return callAdapter(adapter, fn, scene, req, res, added_path_index);
+        // Abort pipeline and return in case of failure
+        callAdapter(adapter, fn, scene, req, res, added_path_index);
+        return (res.error_code_.val == moveit_msgs::msg::MoveItErrorCodes::SUCCESS);
```
henningkayser, line 147: Here you are returning a bool, even though the return type should be MoveItErrorCode. Why can't all functions (`callAdapter()` etc) return a MoveItErrorCode, and once you detect a non-successful error anywhere in the pipeline, you return that as result. No need to pass in a mutable if you fail early and pass back the result in the pipeline function chain.

moveit_ros/planning/planning_request_adapter_plugins/src/add_ruckig_traj_smoothing.cpp
```diff
@@ -64,17 +64,17 @@ class AddRuckigTrajectorySmoothing : public planning_request_adapter::PlanningRe
...
       if (!smoother_.applySmoothing(*res.trajectory_, req.max_velocity_scaling_factor,
                                     req.max_acceleration_scaling_factor))
       {
-        result = false;
+        moveit_code = moveit::core::MoveItErrorCode::FAILURE;
       }
     }
 
-    return result;
+    return bool(moveit_code);
```
henningkayser, line 77: If you pass back the error code here instead of bool, and then pass it through in `callAdapter()`, that should be all you need. The termination condition is already built in, since smoothing is only applied in the happy path.

AndyZe: I will try it. I think it's going to require another large set of changes.

After this request, I'm going to claim **scope creep** because our client needs to get this merged soon.

moveit_core/planning_request_adapter/src/planning_request_adapter.cpp
```diff
@@ -45,22 +45,30 @@ rclcpp::Logger LOGGER = rclcpp::get_logger("moveit").get_child("planning_request
...
+                                                        const planning_scene::PlanningSceneConstPtr& planning_scene,
+                                                        const planning_interface::MotionPlanRequest& req,
+                                                        planning_interface::MotionPlanResponse& res)
 {
   planning_interface::PlanningContextPtr context = planner.getPlanningContext(planning_scene, req, res.error_code_);
   if (context)
-    return context->solve(res);
+  {
+    // TODO(andyz): consider returning a moveit::core::MoveItErrorCode from context->solve()
+    bool result = context->solve(res);
```
AndyZe, line 57: @henningkayser I replaced `bool` with `moveit::core::MoveItErrorCode` up to this point in planning_interface::PlanningContext:

`bool solve(MotionPlanDetailedResponse& res)`

If this PR gets merged, I'll make an issue to recommend replacing `bool` even further up the pipeline.

moveit_ros/planning/planning_request_adapter_plugins/src/add_ruckig_traj_smoothing.cpp
AndyZe: Done

PR #1606, opened 2022-10-06
Convert OMPL status to MoveItErrorCode in the OMPL interface

AndyZe:
### Description

Get better detail from planning errors by converting OMPL codes to MoveItErrorCode and returning them.

OMPL code documentation:  https://github.com/ompl/ompl/blob/main/src/ompl/base/PlannerStatus.h
MoveItErrorCode documentation:  https://github.com/ros-planning/moveit_msgs/blob/master/msg/MoveItErrorCodes.msg

### DIFF
--- a/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
+++ b/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
@@ -980,15 +980,15 @@ void ompl_interface::ModelBasedPlanningContext::omplPlannerStatusToMoveItErrorCo
       break;
     case ompl::base::PlannerStatus::INVALID_START:
       RCLCPP_WARN(LOGGER, "Invalid start state");
-      res.val = moveit_msgs::msg::MoveItErrorCodes::PLANNING_FAILED;
+      res.val = moveit_msgs::msg::MoveItErrorCodes::START_STATE_INVALID;
       break;
     case ompl::base::PlannerStatus::INVALID_GOAL:
       RCLCPP_WARN(LOGGER, "Invalid goal state");
-      res.val = moveit_msgs::msg::MoveItErrorCodes::PLANNING_FAILED;
+      res.val = moveit_msgs::msg::MoveItErrorCodes::GOAL_STATE_INVALID;
       break;
     case ompl::base::PlannerStatus::UNRECOGNIZED_GOAL_TYPE:
       RCLCPP_WARN(LOGGER, "Unrecognized goal type");
-      res.val = moveit_msgs::msg::MoveItErrorCodes::PLANNING_FAILED;
+      res.val = moveit_msgs::msg::MoveItErrorCodes::UNRECOGNIZED_GOAL_TYPE;
       break;
     case ompl::base::PlannerStatus::TIMEOUT:
       RCLCPP_WARN(LOGGER, "Timed out");
@@ -1003,11 +1003,11 @@ void ompl_interface::ModelBasedPlanningContext::omplPlannerStatusToMoveItErrorCo
       break;
     case ompl::base::PlannerStatus::CRASH:
       RCLCPP_WARN(LOGGER, "OMPL crashed!");
-      res.val = moveit_msgs::msg::MoveItErrorCodes::FAILURE;
+      res.val = moveit_msgs::msg::MoveItErrorCodes::CRASH;
       break;
     case ompl::base::PlannerStatus::ABORT:
       RCLCPP_WARN(LOGGER, "OMPL was aborted");
-      res.val = moveit_msgs::msg::MoveItErrorCodes::FAILURE;
+      res.val = moveit_msgs::msg::MoveItErrorCodes::ABORT;
       break;
     default:
       // This should never happen
--- a/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
+++ b/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
@@ -861,7 +861,7 @@ const moveit_msgs::msg::MoveItErrorCodes ompl_interface::ModelBasedPlanningConte
     last_plan_time_ = ompl_simple_setup_->getLastPlanComputationTime();
     unregisterTerminationCondition();
     // fill the result status code
-    omplPlannerStatusToMoveItErrorCode(ompl_simple_setup_, result);
+    result.val = logPlannerStatus(ompl_simple_setup_);
   }
   else
   {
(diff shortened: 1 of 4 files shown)


moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
```diff
@@ -959,6 +968,62 @@ void ompl_interface::ModelBasedPlanningContext::unregisterTerminationCondition()
...
+  ompl::base::PlannerStatus ompl_status = ompl_simple_setup->getLastPlannerStatus();
+  if (ompl_status == ompl::base::PlannerStatus::UNKNOWN)
+  {
+    RCLCPP_WARN(LOGGER, "Motion planning failed for an unknown reason");
+    res.val = moveit_msgs::msg::MoveItErrorCodes::PLANNING_FAILED;
+  }
+  else if (ompl_status == ompl::base::PlannerStatus::INVALID_START)
+  {
+    RCLCPP_WARN(LOGGER, "Invalid start state");
+    res.val = moveit_msgs::msg::MoveItErrorCodes::PLANNING_FAILED;
```
sjahr, line 983: Do you think we could add something like INVALID_START to the MoveItError codes?
Right now we have `START_STATE_IN_COLLISION` and `START_STATE_VIOLATES_PATH_CONSTRAINTS` but none both do not correspond to OMPL's message definition, yet I think it is helpful to have that information in the error code as well.

You could add a TODO here and open an issue in MoveIt2 so we can add this later

moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
```diff
@@ -959,6 +968,62 @@ void ompl_interface::ModelBasedPlanningContext::unregisterTerminationCondition()
...
+  }
+  else if (ompl_status == ompl::base::PlannerStatus::INVALID_GOAL)
+  {
+    RCLCPP_WARN(LOGGER, "Invalid goal state");
+    res.val = moveit_msgs::msg::MoveItErrorCodes::PLANNING_FAILED;
```
sjahr, line 988: Same here. Would GOAL_CONSTRAINTS_VIOLATED work?

moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
```diff
@@ -959,6 +968,62 @@ void ompl_interface::ModelBasedPlanningContext::unregisterTerminationCondition()
   ptc_ = nullptr;
 }
 
+void ompl_interface::ModelBasedPlanningContext::omplPlannerStatusToMoveItErrorCode(
+    og::SimpleSetupPtr ompl_simple_setup, moveit_msgs::msg::MoveItErrorCodes& res)
+{
+  ompl::base::PlannerStatus ompl_status = ompl_simple_setup->getLastPlannerStatus();
+  if (ompl_status == ompl::base::PlannerStatus::UNKNOWN)
```
sjahr, line 975: Nit: Wouldn't a switch case statement work here as well? I think it can be a bit more efficient

sjahr (commented):
This is great, thanks for adding it! After reviewing it, I felt like we should maybe extend the existing MoveItError codes, so we loose even fewer information during the translation OMPLStatus -> MoveItError code. But this can be done in a follow up PR

moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
AndyZe: I feel like that's slightly different. Prob best to add a new code that's an exact match

moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
AndyZe: I'm glad you suggested this because I had the same thought. I started a PR here:

https://github.com/ros-planning/moveit_msgs/pull/146

sjahr (approved):

moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
```diff
@@ -959,6 +968,54 @@ void ompl_interface::ModelBasedPlanningContext::unregisterTerminationCondition()
   ptc_ = nullptr;
 }
 
+void ompl_interface::ModelBasedPlanningContext::omplPlannerStatusToMoveItErrorCode(
```
tylerjw, line 971: Why does this use an output parameter and why does it log?

AndyZe: I'll get rid of the output parameter.

As far as logging goes, this seems like a pretty good place to do it. We are already enumerating all possible cases, so... that code would have to be duplicated if the logging moved elsewhere. Maybe the function name just needs to be changed to indicate it logs? Like,  `logOmplPlannerStatus()`

tylerjw: Maybe do something like this and then the call site can decide what they want:

```cpp
// convert ompl status into moveit error code 
moveit_msgs::msg::MoveItErrorCodes toMoveItError(ompl::base::PlannerStatus status);

// test if ompl status or moveit error code is success 
bool isSuccess(ompl::base::PlannerStatus status);
bool isSuccess(moveit_msgs::msg::MoveItErrorCodes error_code);

// convert ompl status or moveit code into string, used in logging
std::string toString(ompl::base::PlannerStatus status);
std::string toString(moveit_msgs::msg::MoveItErrorCodes error_code);

// log warning if ompl status or moveit code is error
void logWarningIfError(ompl::base::PlannerStatus status);
void logWarningIfError(moveit_msgs::msg::MoveItErrorCodes error_code);
```

At the call site where you need to create a moveit error and log the ompl status you can do this:

```cpp
auto ompl_status = ompl_simple_setup_->getLastPlannerStatus();
logWarningIfError(ompl_status);
result.val = toMoveItError(ompl_status);
```

In model_based_planning_context.cpp it can make the code simpler if you use a `using` statement for success and failure. This is fine to do because it is not in a header and won't propagate into other files.

```cpp
using SUCCESS = moveit_msgs::msg::MoveItErrorCodes::SUCCESS;
using FAILURE = moveit_msgs::msg::MoveItErrorCodes::FAILURE;
```

Then, combined with the isSuccess functions you can make those lines where you need to test if both the moveit error code and the ompl status are good results and return a success/failure moveit code can become a bit less verbose:
```cpp
result.val = (isSuccess(result.val) && isSuccess(ompl_status)) ? SUCCESS : FAILURE;
```

These are just ideas for how you can make some smaller functions you can combine to make the error types easier to use in various places. The important idea here is just separating the function of converting the types from the function that logs.

tylerjw: Thank you for getting rid of the output parameter and using a return type. Above are just some suggestions on how we could maybe make this code a bit more flexible and easy to understand at the call site of the function.

tylerjw: @AndyZe I'm also happy to accept this change as is and try to make the change I described above myself. Would you rather that?

AndyZe: Give me a couple days. I want to try what you've suggested but prob won't get to it today

tylerjw (approved):

Issue #1594, opened 2022-10-02
Moveit build error - Ubuntu 22.04


lonebots:
### Description

The terminal crashed when trying to build Moveit using the source. All the dependencies where correctly installed. and further details about the crash and logs are provided below.

### Your environment
* ROS Distro: Humble (ros2)
* OS Version: Ubuntu 22.04
* Source build, (binary not available)
* Branch : stable and unstable (the issue prevails)

### Steps to reproduce
Follow the source build instruction provided [here.](https://moveit.ros.org/install-moveit2/source/) The error occurred after even after successfully installing all the dependencies given in the installation (exception used a different version of clang-format, because clang-format-10 is not available anymore). It occurs when running the build command. 
> cd $COLCON_WS
colcon build --event-handlers desktop_notification- status- --cmake-args -DCMAKE_BUILD_TYPE=Release

### Expected behaviour
The expected behaviour is to successfully complete the build with minimum warning about deprecation. 

### Actual behaviour
The source build exited with 2 errors.
> --- stderr: moveit_ros_robot_interaction
/usr/bin/ld: libmoveit_robot_interaction.so.2.5.3: undefined reference to `moveit::core::RobotState::setFromIK(moveit::core::JointModelGroup const*, geometry_msgs::msg::Pose_<std::allocator<void> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, double, std::function<bool (moveit::core::RobotState*, moveit::core::JointModelGroup const*, double const*)> const&, kinematics::KinematicsQueryOptions const&, std::function<double (geometry_msgs::msg::Pose_<std::allocator<void> > const&, moveit::core::RobotState const&, moveit::core::JointModelGroup*, std::vector<double, std::allocator<double> > const&)>)'
collect2: error: ld returned 1 exit status
gmake[2]: *** [CMakeFiles/locked_robot_state_test.dir/build.make:329: locked_robot_state_test] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:185: CMakeFiles/locked_robot_state_test.dir/all] Error 2
gmake: *** [Makefile:146: all] Error 2

### Backtrace or Console output

> Starting >>> moveit_ros_perception
> [Processing: moveit_kinematics, moveit_planners_ompl, moveit_ros_perception, moveit_ros_robot_interaction, moveit_ros_warehouse]
[...]

PR #1617, opened 2022-10-17
Add moveit_core dependency to robot_interaction

sea-bass:
### Description

This PR addresses the core issue in https://github.com/ros-planning/moveit2/issues/1594, which is a missing dependency that makes moveit2's `robot_interaction` package build successfully if building with a single worker, but fail at linking time otherwise against some functions available in `moveit_core`.


sea-bass:
If all is good here, this should additionally get backported to Humble.

PR #1624, opened 2022-10-19
Added brace initialization in multiple modules

abhijelly:
### Description

### Checklist

### DIFF
--- a/moveit_core/distance_field/src/distance_field.cpp
+++ b/moveit_core/distance_field/src/distance_field.cpp
@@ -110,11 +110,11 @@ void DistanceField::getIsoSurfaceMarkers(double min_distance, double max_distanc
   // inf_marker.lifetime = ros::Duration(30.0);
 
   inf_marker.points.reserve(100000);
-  for (int x = 0; x < getXNumCells(); ++x)
+  for (int x{ 0 }; x < getXNumCells(); ++x)
   {
-    for (int y = 0; y < getYNumCells(); ++y)
+    for (int y{ 0 }; y < getYNumCells(); ++y)
     {
-      for (int z = 0; z < getZNumCells(); ++z)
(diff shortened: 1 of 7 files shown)


moveit_core/distance_field/src/distance_field.cpp
```diff
@@ -428,7 +428,7 @@ void DistanceField::getPlaneMarkers(PlaneVisualizationType type, double length,
           continue;
         }
         double dist = getDistance(x, y, z);
-        int last{ plane_marker.points.size() };
+        int last = plane_marker.points.size();
```
tylerjw, line 431: This is still a narrowing conversion, it is just implicit when you use operator=. It would be better to change this to not be a narrowing conversion. There are a couple ways you can do that. The first is to use auto:

```c++
auto last{ plane_marker.points.size() };
```
This will cause the type of last to be whatever the type returned by the size() function is.
The second would be to use an unsigned type like `size_t` (this is likely the type returned by size()):

```cpp
size_t last{ plane_marker.points.size() };
```

abhijelly: Thank you for the suggestions! I will fix this.

henningkayser (commented):
Thank you for proposing these changes. I'm not convinced that we should switch to brace initialization, but I'm also not entirely against it. IMO, it's way more important to be consistent, and there are already examples in the GitHub diff where this is not the case. Changes like these are style changes that should be discussed and documented, and then applied to the entire codebase, possibly using incremental PRs for each module/package.

ChrisThrasher:
I agree with Henning that this is not a worthwhile change. Braced initialization is wonderful for implementing templates but adds little value outside of that context. We're better off with `=`-based initializers. If we're worried about implicit conversions, that's what `auto`, `-Wconversion`, and `-Wsign-conversion` are for.

Issue #1112, opened 2022-03-10
Deprecate getPlanningScene() and the cast operator


AndyZe:
As mentioned by v4hn [here](https://github.com/ros-planning/moveit2/pull/1100#pullrequestreview-904301643):

> Additionally I would propose to deprecate the getPlanningScene() method that goes along with these interfaces.
> It's just ludicrous to have a method with its documentation saying [DON'T USE ME](https://github.com/ros-planning/moveit2/blob/main/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h#L189-L190).
> 
> It would also help usability to have a simple getScene() method in the locking classes instead of the [awful cast operator](https://github.com/ros-planning/moveit2/blob/main/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h#L651-L654).

abhijelly:
Hi @AndyZe, I cannot seem to find `getPlanningScene()` definition and `getScene()` in the codebase. Is this issue resolved or do I have to look elsewhere? Thank you!

AndyZe:
`getPlanningScene()` is definitely still an issue and you can easily find all usages like so:

`grep -r "getPlanningScene("`

I wouldn't worry about `getScene()` for now, let's take care of `getPlanningScene()` first.

Basically this issue asks you to add a deprecation warning to the [function declaration](https://github.com/ros-planning/moveit2/blob/44821258a71c54a1b2381d8e664dfcdac79c40fa/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h#L212) like this:

`[[deprecated("getPlanningScene() is deprecated, use LockedPlanningSceneRO or LockedPlanningSceneRW instead.")]] const planning_scene::PlanningScenePtr& getPlanningScene()`

Since getPlanningScene() is used internally in a bunch of places, we need to replace it somehow. Use LockedPlanningSceneRO or LockedPlanningSceneRW as described in this tutorial:  https://ros-planning.github.io/moveit_tutorials/doc/motion_planning_pipeline/motion_planning_pipeline_tutorial.html#start

@v4hn @tylerjw @henningkayser  does that sound like a good path forward to you? It's going to mean a lot more internal locking of the planning scene.

abhijelly:
@AndyZe Thank you for your help! I've raised #1635.

PR #1635, opened 2022-10-23
added deprecation warning for getPlanningScene()

abhijelly:
### Description

### Checklist


abhijelly:
I suspect the checks are failing because the warnings are treated as errors. I'm think adding `-Wno-deprecated-declarations` in the CMakeLists.txt should prevent this from occurring. But I'm not sure where do I make this change? @AndyZe any thoughts?

moveit_core/constraint_samplers/CMakeLists.txt
```diff
@@ -1,5 +1,7 @@
 set(MOVEIT_LIB_NAME moveit_constraint_samplers)
 
+add_compile_options(-Wdeprecated-declarations)
```
henningkayser, line 3: Please remove, deprecation warnings are intended to be treated as errors. If code is affected by deprecations, it needs to be fixed as well.

moveit_core/constraint_samplers/include/moveit/constraint_samplers/constraint_sampler.h
```diff
@@ -115,7 +115,9 @@ class ConstraintSampler
    *
    * @return The planning scene as a const ptr
    */
-  const planning_scene::PlanningSceneConstPtr& getPlanningScene() const
+  [[deprecated("getPlanningScene() is deprecated, use LockedPlanningSceneRO or LockedPlanningSceneRW "
+               "instead.")]] const planning_scene::PlanningSceneConstPtr&
+  getPlanningScene() const
```
henningkayser, line 120: The deprecation doesn't make sense in this context, since there is no PSM provided by this class that could be locked.

moveit_core/planning_interface/include/moveit/planning_interface/planning_interface.h
```diff
@@ -100,7 +100,9 @@ class PlanningContext
   }
 
   /** \brief Get the planning scene associated to this planning context */
-  const planning_scene::PlanningSceneConstPtr& getPlanningScene() const
+  [[deprecated("getPlanningScene() is deprecated, use LockedPlanningSceneRO or LockedPlanningSceneRW "
```
henningkayser, line 103: Same thing

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
```diff
@@ -209,15 +209,19 @@ class MOVEIT_PLANNING_SCENE_MONITOR_EXPORT PlanningSceneMonitor
    * @see LockedPlanningSceneRO
    * @see LockedPlanningSceneRW.
    * @return A pointer to the current planning scene.*/
-  const planning_scene::PlanningScenePtr& getPlanningScene()
+  [[deprecated("getPlanningScene() is deprecated, use LockedPlanningSceneRO or LockedPlanningSceneRW "
+               "instead.")]] const planning_scene::PlanningScenePtr&
+  getPlanningScene()
```
henningkayser, line 214: With this I can go along

henningkayser (commented):
What's your motivation behind these changes? I agree that it's only safe to access the PSM using the locked wrappers, but the other classes should not be affected by this at all.

abhijelly:
> 

https://github.com/ros-planning/moveit2/issues/1112#issuecomment-1287925276
Thank you getting back! As mentioned in the issue discussion above, I assumed every getPlanningScene() declaration will require a deprecation warning. I've pushed changes you suggested, fyi.

AndyZe (dismissed):

AndyZe:
CI needs to be fixed before this can be merged, which means you need to replace the internal usage of `getPlanningScene()` with something else. You can click on `details` of failed CI jobs to see exactly where this needs to be done. I'll try to push one example to your branch.

AndyZe:
Please ignore my previous comment. I see the difficulty now. Working on a follow-up...

AndyZe (approved):

AndyZe:
I used the trick described here to prevent compiler warnings where `getPlanningScene()` was unavoidably used internally.

https://stackoverflow.com/a/3394268/3499467

abhijelly:
> I used the trick described here to prevent compiler warnings where `getPlanningScene()` was unavoidably used internally.
> 
> https://stackoverflow.com/a/3394268/3499467

thank you for sharing the resource!

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
```diff
@@ -657,16 +661,23 @@ class LockedPlanningSceneRO
...
   operator bool() const
   {
-    return planning_scene_monitor_ && planning_scene_monitor_->getPlanningScene();
+    planning_scene_monitor::LockedPlanningSceneRO ls(planning_scene_monitor_);
+    return planning_scene_monitor_ && ls;
   }
 
   operator const planning_scene::PlanningSceneConstPtr&() const
   {
+// Ignore getPlanningScene() deprecation
```
henningkayser, line 670: It makes absolutely no sense to deprecate a function but then instead of adapting the code to only suppress the deprecation warnings. This needs to be fixed. Obviously, for this kind of locking feature we still need access to the internal planning scene, we just don't want to expose it publicly. One solution: deprecate the old function, make the `LockedPlanningScene*` classes friend classes to allow them access to PSM members and private functions, (optional but recommended) provide a new private function `PlanningSceneMonitor::getInternalPlanningScene()` and use that instead.

I'm still not entirely sure about the possible side effects of this deprecation, and will do some more reviewing before approval. We should also discuss if the public `PlanningSceneMonitor::lockScene*()` functions should be deprecated as well with this change because there is no use for them anymore.

AndyZe: `provide a new private function PlanningSceneMonitor::getInternalPlanningScene()`

I don't think we should do this because that might encourage future MoveIt developers to use it in the future. We don't want that.

AndyZe: > We should also discuss if the public PlanningSceneMonitor::lockScene*() functions should be deprecated as well with this change because there is no use for them anymore.

My thinking is, we ought to keep those methods because they have other functionality, like this:

```
void PlanningSceneMonitor::lockSceneRead()
{
  scene_update_mutex_.lock_shared();
  if (octomap_monitor_)
    octomap_monitor_->getOcTreePtr()->lockRead();
}
```

AndyZe: `One solution: ...`

Done!

henningkayser: > `provide a new private function PlanningSceneMonitor::getInternalPlanningScene()`
> 
> I don't think we should do this because that might encourage future MoveIt developers to use it in the future. We don't want that.

They can't, that's what private is for. And if they do, they have do it explicitly with a friend class like here, in which case I trust they have a good reason, like here.

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
```diff
@@ -657,17 +661,18 @@ class LockedPlanningSceneRO
 
   operator bool() const
   {
-    return planning_scene_monitor_ && planning_scene_monitor_->getPlanningScene();
+    planning_scene_monitor::LockedPlanningSceneRO ls(planning_scene_monitor_);
+    return planning_scene_monitor_ && ls;
```
henningkayser, line 665: Why would you change the behavior of this function and did you try running this? Before, it would check if psm and planning scene are valid (!=nullptr). Now it seems to recursively check `LockedPlanningSceneRO::bool()`.

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
```diff
@@ -657,17 +661,18 @@ class LockedPlanningSceneRO
...
   operator const planning_scene::PlanningSceneConstPtr&() const
   {
-    return static_cast<const PlanningSceneMonitor*>(planning_scene_monitor_.get())->getPlanningScene();
+    return static_cast<const PlanningSceneMonitor*>(planning_scene_monitor_.get())->scene_;
   }
 
   const planning_scene::PlanningSceneConstPtr& operator->() const
   {
-    return static_cast<const PlanningSceneMonitor*>(planning_scene_monitor_.get())->getPlanningScene();
+    return static_cast<const PlanningSceneMonitor*>(planning_scene_monitor_.get())->scene_;
```
sjahr, line 675:
```suggestion
    return static_cast<const PlanningSceneMonitor*>(planning_scene_monitor_.get())->scene_const_;
```

tylerjw:
This work seems to have been abandoned. I'll open an issue for finishing it. Feel free to re-open this PR.

PR #1638, opened 2022-10-27
Add joint acceleration validator methods to Pilz limits container

sea-bass:
### Description

While developing, I found a bug in Pilz where not setting joint acceleration limits causes us to directly access a map with a non-existent key and segfault while validating a trajectory against joint limits.

This had already been handled for the position and velocity limits by first checking the existence of this key using `hasLimit()`, so it just required a little bit of copy-paste for the acceleration / deceleration case.

NOTE: As this work is doing in service of MoveIt Studio, it should be backported to Humble.

### Checklist

### DIFF
--- a/moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/joint_limits_container.h
+++ b/moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/joint_limits_container.h
@@ -130,7 +130,7 @@ public:
    * @param joint_position
    * @return true if within limits, false otherwise
    */
-  bool verifyPositionLimit(const std::string& joint_name, const double& joint_position) const;
+  bool verifyPositionLimit(const std::string& joint_name, double joint_position) const;
 
   /**
    * @brief verify velocity limit of single joint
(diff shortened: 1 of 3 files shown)


moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/joint_limits_container.h
```diff
@@ -128,17 +128,33 @@ class JointLimitsContainer
...
-   * @brief verify position limit of single joint
+   * @brief verify acceleration limit of single joint
    * @param joint_name
-   * @param joint_position
-   * @return
+   * @param joint_acceleration
+   * @return true if within limits, false otherwise
    */
-  bool verifyPositionLimit(const std::string& joint_name, const double& joint_position) const;
+  bool verifyAccelerationLimit(const std::string& joint_name, const double& joint_acceleration) const;
```
tylerjw, line 149: Odd question, why pass a const-ref to a double? Doesn't that just add indirection by creating a pointer to a value of the same size as the pointer?

tylerjw (approved):

PR #1679, opened 2022-10-31
Switch back to the moveit_msgs debian

AndyZe:
This cleans up the TODO [here](https://github.com/ros-planning/moveit2/blob/66a64b4a72b6ddef1af2329f20ed8162554d5bcb/.github/workflows/ci.yaml#L39):  

```
      # Uninstall binaries that are duplicated in the .repos file
      # TODO(andyz): remove this once a sync containing 35b93c8 happens
      AFTER_SETUP_UPSTREAM_WORKSPACE_EMBED: sudo apt remove -y ros-${{ matrix.env.ROS_DISTRO }}-moveit-msgs || true
```


sea-bass:
Does the TODO in https://github.com/ros-planning/moveit2/blob/main/moveit2.repos need to get put back as well?

tylerjw (approved):

PR #1694, opened 2022-11-04
Short-circuit planning adapters

rhaschke:
This is a simpler implementation of #1605, which is forward-ported from https://github.com/ros-planning/moveit/pull/3262.
In contrast to #1605 we can keep the existing boolean return types as the error code can be propagated in `res.error_code_`.
There is no need to additionally use the return value to propagate an error code. By the way, eventually the propagated return value wasn't used at all, but simply [cast back to a boolean](https://github.com/ros-planning/moveit2/pull/1605/files#diff-fa45d6f7c1ebdbae2bd647690436c1b3936b94fc24e2dd07c4b3f4e187b4feb3R262):
https://github.com/ros-planning/moveit2/blob/abbbb9d906d21cb7fe2d3ee7c2a2a31799b9d0aa/moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp#L270-L271

As this PR first reverts #1605, the diff is primarily cluttered by this revert. For review, one ideally compares directly to the original main branch: https://github.com/ros-planning/moveit2/compare/749c5ae0777d093fad1be6514bae2bc188fc6926...rhaschke:moveit2:short-circuit-planning-adapters


AndyZe:
I'm fine with the general idea of returning `bool`. We went back and forth in the original PR between bool, void, and MoveItErrorCode. I think `void` or `bool` is the best option so it's not redundant with `res`. Testing this now.

moveit_core/utils/include/moveit/utils/moveit_error_code.h
```diff
@@ -47,11 +47,7 @@ namespace core
 class MoveItErrorCode : public moveit_msgs::msg::MoveItErrorCodes
 {
 public:
-  MoveItErrorCode()
-  {
-    val = 0;
-  }
-  MoveItErrorCode(int code)
+  MoveItErrorCode(int code = 0)
```
AndyZe, line 50: Nice simplification here.

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_bounds.cpp
```diff
@@ -167,18 +165,16 @@ class FixStartStateBounds : public planning_request_adapter::PlanningRequestAdap
       }
     }
 
-    moveit::core::MoveItErrorCode moveit_code;
+    bool solved;
```
AndyZe, line 168:
```suggestion
    bool solved = false;
```

moveit_ros/planning/plan_execution/src/plan_execution.cpp
```diff
@@ -109,35 +110,6 @@ void plan_execution::PlanExecution::stop()
   preempt_.request();
 }
 
-std::string plan_execution::PlanExecution::getErrorCodeString(const moveit_msgs::msg::MoveItErrorCodes& error_code)
```
AndyZe, line 112: Also great to remove this!

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_collision.cpp
```diff
@@ -147,17 +145,16 @@ class FixStartStateCollision : public planning_request_adapter::PlanningRequestA
...
       {
         RCLCPP_WARN(LOGGER,
                     "Unable to find a valid state nearby the start state (using jiggle fraction of %lf and %u sampling "
-                    "attempts).",
+                    "attempts). Passing the original planning request to the planner.",
                     jiggle_fraction_, sampling_attempts_);
         res.error_code_.val = moveit_msgs::msg::MoveItErrorCodes::START_STATE_IN_COLLISION;
-        return moveit::core::MoveItErrorCode(moveit_msgs::msg::MoveItErrorCodes::START_STATE_IN_COLLISION);
-        ;
+        return false;  // skip remaining planners
```
AndyZe, line 157:
```suggestion
        return false;  // skip remaining planning adapters
```

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_path_constraints.cpp
```diff
@@ -120,22 +116,24 @@ class FixStartStatePathConstraints : public planning_request_adapter::PlanningRe
...
+      else
       {
-        RCLCPP_WARN(LOGGER, "Unable to meet path constraints at the start.");
-        moveit::core::MoveItErrorCode moveit_code(
-            moveit_msgs::msg::MoveItErrorCodes::START_STATE_VIOLATES_PATH_CONSTRAINTS);
+        RCLCPP_WARN(LOGGER, "Unable to plan to path constraints. Running usual motion plan.");
         res.error_code_.val = moveit_msgs::msg::MoveItErrorCodes::START_STATE_VIOLATES_PATH_CONSTRAINTS;
-        return moveit_code;
+        res.planning_time_ = res2.planning_time_;
+        return false;  // skip remaining planners
```
AndyZe, line 129:
```suggestion
        return false;  // skip remaining planning adapters
```

AndyZe (commented):
Generally looks good, I verified functionality

henningkayser (commented):
This actually looks simpler. I somewhat like the idea of returning the bool-convertible error code more (which could work just the same way), but either way is fine to me.

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_collision.cpp
rhaschke: I considered that before, but not only adapters are skipped but the main planner as well. As both are named with the variable `planner` in source, I decided for planners.

moveit_ros/planning/planning_request_adapter_plugins/src/fix_workspace_bounds.cpp
```diff
@@ -81,8 +79,8 @@ class FixWorkspaceBounds : public planning_request_adapter::PlanningRequestAdapt
       default_wp.max_corner.x = default_wp.max_corner.y = default_wp.max_corner.z = workspace_extent_;
       return planner(planning_scene, req2, res);
     }
-
-    return planner(planning_scene, req, res);
+    else
+      return planner(planning_scene, req, res);
```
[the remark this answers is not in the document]
rhaschke: Our formatting style doesn't enforce this, so why do you want to enforce it?

rhaschke:
> This actually looks simpler. I somewhat like the idea of returning the bool-convertible error code more (which could work just the same way).

It _is_ simpler and it is removing redundancy (returning the error code by return value _and_ in response).

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_collision.cpp
AndyZe:
```suggestion
        return false;  // skip remaining planners & adapters
```

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_path_constraints.cpp
AndyZe:
```suggestion
        return false;  // skip remaining planners & adapters
```

AndyZe (approved):

moveit_ros/planning/planning_request_adapter_plugins/src/fix_workspace_bounds.cpp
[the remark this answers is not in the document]
henningkayser: For consistency, and it's a cleaner style imo

PR #1697, opened 2022-11-06
Preparing Release 2.6.0 for Rolling

vatanaksoytezer:


henningkayser (approved):
Thank you for preparing this. Only complaint is that there really is a lot of noise in the logs. Many commits have way too much meaningless information. Also, there are several duplicate "Co-authored" entries, it's debatable if we really want them in there, or if "Contributors" is sufficient.

vatanaksoytezer:
> Thank you for preparing this. Only complaint is that there really is a lot of noise in the logs. Many commits have way too much meaningless information. Also, there are several duplicate "Co-authored" entries, it's debatable if we really want them in there, or if "Contributors" is sufficient.

I tried to clean a lot of the information (the sub comments and duplicate co-authors), if there is anything you would like me to remove I'd be happy to.

PR #1698, opened 2022-11-06
Fix deprecation of std::random_shuffle in C++17

light-tech:
### Description

With the release of `libc++` 15 that removes `std::random_shuffle` (c.f. https://releases.llvm.org/15.0.0/projects/libcxx/docs/ReleaseNotes.html), I am making the necessary changes to conform with C++ 17.

The old pull request https://github.com/ros-planning/moveit2/pull/353 made a similar change (together with other stuffs) but there doesn't appear any progress. I also address @henningkayser comment there about making the random number generator a member for reuse.

### Checklist


tylerjw:
@christhrasher would you like to review this? Do you think this would be a good place to use random from RSL, or is this entirely different?

ChrisThrasher:
> @ChrisThrasher would you like to review this? Do you think this would be a good place to use random from RSL, or is this entirely different?

`rsl::rng` could be a good fit, especially if there are other places in MoveIt where a random number generated is needed.

light-tech:
> > @ChrisThrasher would you like to review this? Do you think this would be a good place to use random from RSL, or is this entirely different?
> 
> `rsl::rng` could be a good fit, especially if there are other places in MoveIt where a random number generated is needed.

Thanks for the info. That sounds great.

PR #1699, opened 2022-11-07
Port parallel planning reviews from MoveIt1

sjahr:
### Description

This PR ports addressed reviews from @rhaschke to MoveIt2. Most importantly, `bool generatePlan(...)` becomes const again :+1: 

### Checklist

### DIFF
--- a/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/plan_solutions.hpp
+++ b/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/plan_solutions.hpp
@@ -39,6 +39,7 @@
 
 #include <algorithm>
 #include <moveit/planning_interface/planning_response.h>
+#include <mutex>
 #include <thread>
 
 namespace moveit_cpp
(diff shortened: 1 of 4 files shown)


moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/plan_solutions.hpp
```diff
@@ -39,6 +39,7 @@
 
 #include <algorithm>
 #include <moveit/planning_interface/planning_response.h>
+#include <mutex>
```
rhaschke, line 42: Why is that required now (and not before)?

rhaschke (commented):
Only two nitpicks:

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/plan_solutions.hpp
sjahr: It build before, probably because we're lucky and the required header was included from somewhere else but I think it is nicer to ensure that everything is included.

rhaschke (approved):

PR #1707, opened 2022-11-09
Small optimization in constructGoalConstraints()

AndyZe:
We can use the input orientation here since, as the comment says:

> // orientation of constraint region does not affect anything, since it is a sphere

### DIFF
--- a/moveit_core/kinematic_constraints/src/utils.cpp
+++ b/moveit_core/kinematic_constraints/src/utils.cpp
@@ -182,7 +182,7 @@ moveit_msgs::msg::Constraints constructGoalConstraints(const std::string& link_n
   pcm.header = pose.header;
   pcm.constraint_region.primitive_poses.resize(1);
   // orientation of constraint region does not affect anything, since it is a sphere
-  pcm.constraint_region.primitive_poses[0] = pose.pose;
+  pcm.constraint_region.primitive_poses[0].position = pose.pose.position;
   pcm.weight = 1.0;
 
   goal.orientation_constraints.resize(1);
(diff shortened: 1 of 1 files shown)


moveit_core/kinematic_constraints/src/utils.cpp
```diff
@@ -181,13 +181,8 @@ moveit_msgs::msg::Constraints constructGoalConstraints(const std::string& link_n
...
   pcm.header = pose.header;
   pcm.constraint_region.primitive_poses.resize(1);
-  pcm.constraint_region.primitive_poses[0].position = pose.pose.position;
-
   // orientation of constraint region does not affect anything, since it is a sphere
-  pcm.constraint_region.primitive_poses[0].orientation.x = 0.0;
-  pcm.constraint_region.primitive_poses[0].orientation.y = 0.0;
-  pcm.constraint_region.primitive_poses[0].orientation.z = 0.0;
-  pcm.constraint_region.primitive_poses[0].orientation.w = 1.0;
+  pcm.constraint_region.primitive_poses[0] = pose.pose;
```
henningkayser, line 185:
```suggestion
  pcm.constraint_region.primitive_poses[0].position = pose.pose.position;
```
Just to make sure this doesn't break anything (who trusts a comment these days?).
Quarternions default to identity https://docs.ros2.org/galactic/api/geometry_msgs/msg/Quaternion.html

rhaschke (commented):
Now orientation isn't initialized at all.

AndyZe:
It's initialized to identity by default in ROS2:  https://docs.ros2.org/galactic/api/geometry_msgs/msg/Quaternion.html

henningkayser (approved):

PR #1706, opened 2022-11-09
Fix clang-tidy issues

rhaschke:
Having the clang-tidy check `performance-unnecessary-value-param` disabled for a few years, lead to the [introduction of several shortcomings in the code](https://github.com/ros-planning/moveit2/actions/runs/3427540403), which should be fixed asap.
Applying all automatic fixes from clang-tidy is not a good idea as you can see from [second commit, trying to cleanup a few of them](https://github.com/ros-planning/moveit2/pull/1706/commits/25c798e9e3726ca76b672ec04a828ef358d15a75). Somebody from the MoveIt2 team should take over and clean up the remaining ones. @henningkayser, can you please coordinate this work?

PR #1710, opened 2022-11-10
Cleanup lookup of planning pipelines in MoveItCpp

rhaschke:
This PR attempts to cleanup the mess in configuring multiple planning pipelines in MoveItCpp, which exists since the very beginning and was changed back and forth since then, see #1096, #1114, #1522.

This PR removes `MoveItCpp::getPlanningPipelineNames()`, which was obviously intended initially to provide a planning-group-based filter for all available planning pipelines:
A pipeline was discarded for a group if there were no `planner_configs` defined for that group on the parameter server.

As pointed out in #1522, only OMPL actually explicitly declares `planner_configs` on the parameter server.
To enable all other pipelines as well, #1522 introduced empty dummy `planner_configs` for all other planners (CHOMP + Pilz) - and thus, obviously, circumventing the original filter mechanism!

As the filter mechanism is useless now, we can just remove the function `getPlanningPipelineNames()` and revert #1522 instead.
@sjahr, this is what I was thinking of for https://github.com/ros-planning/moveit/pull/3244 (actually the PR here simplifies even more).


henningkayser:
Are you suggesting to remove the `PlannerConfigurationSettings` as well? I think this would be the direct consequence of omitting the interface, and I would agree that it would make a lot of sense to do so. Also, if we don't require a mapping, the `groups_pipelines_map_` should also be removed with this PR.

rhaschke:
> Are you suggesting to remove the `PlannerConfigurationSettings` as well? I think this would be the direct consequence of omitting the interface.

No, not at all. I think it makes a lot of sense to allow different planner configs for different groups!
The PR here just provides a _fallback_ in case a planner plugin doesn't setup any config settings. Most user's assumption is that the plugin will serve all groups will all its planning capabilities = planners.

> Also, if we don't require a mapping, the `groups_pipelines_map_` should also be removed with this PR.

I agree with that and pushed another commit.

sjahr (approved):
I tested this with OMPL and PILZ and multiple pipelines are usable. CI failure is not cause by this PR but unrelated clang-tidy errors that are probably fixed by https://github.com/ros-planning/moveit2/pull/1706.

rhaschke:
@sjahr, if this gets approved here, can you please adapt https://github.com/ros-planning/moveit/pull/3244 correspondingly?

sjahr:
> @sjahr, if this gets approved here, can you please adapt [ros-planning/moveit#3244](https://github.com/ros-planning/moveit/pull/3244) correspondingly?

Yes, thanks a lot for making these changes :pray:

henningkayser (approved):

PR #1712, opened 2022-11-11
Fix some more clang compiler warnings

rhaschke:
This depends on https://github.com/ament/googletest/pull/21

### DIFF
--- a/moveit_planners/pilz_industrial_motion_planner_testutils/include/pilz_industrial_motion_planner_testutils/goalconstraintsmsgconvertible.h
+++ b/moveit_planners/pilz_industrial_motion_planner_testutils/include/pilz_industrial_motion_planner_testutils/goalconstraintsmsgconvertible.h
@@ -48,6 +48,7 @@ namespace pilz_industrial_motion_planner_testutils
 class GoalConstraintMsgConvertible
 {
 public:
+  GoalConstraintMsgConvertible() = default;
   GoalConstraintMsgConvertible(const GoalConstraintMsgConvertible&) = default;
   GoalConstraintMsgConvertible(GoalConstraintMsgConvertible&&) = default;
   GoalConstraintMsgConvertible& operator=(const GoalConstraintMsgConvertible&) = default;
--- a/moveit_planners/pilz_industrial_motion_planner_testutils/include/pilz_industrial_motion_planner_testutils/goalconstraintsmsgconvertible.h
+++ b/moveit_planners/pilz_industrial_motion_planner_testutils/include/pilz_industrial_motion_planner_testutils/goalconstraintsmsgconvertible.h
@@ -48,6 +48,10 @@ namespace pilz_industrial_motion_planner_testutils
 class GoalConstraintMsgConvertible
 {
 public:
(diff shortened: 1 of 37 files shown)


rhaschke:
After a full day of work, all compiler warnings in the `moveit2` repo are fixed again. I don't know why we have such an elaborated CI framework if error-checking gets disabled (or downgraded to warnings) as soon as they provide hurdles.
We should really trust the compilers and fix those warnings.

tylerjw (approved):
Thank you for this change. I don't remember warnings being disabled in CI; I didn't know they were. I'm sorry I didn't see that before.

I was the one who disabled the python test that you re-enabled yesterday. Thank you for that; I didn't how to deal with that one. It looks like it was as simple as removing that import as it was unused.

tylerjw:
Would you like to merge this now or wait on that pull request to update google test. I think merging now would be nice but I'm unsure what affect that has on people who build locally.

moveit_planners/pilz_industrial_motion_planner_testutils/include/pilz_industrial_motion_planner_testutils/goalconstraintsmsgconvertible.h
```diff
@@ -48,6 +48,7 @@ namespace pilz_industrial_motion_planner_testutils
 class GoalConstraintMsgConvertible
 {
 public:
+  virtual ~GoalConstraintMsgConvertible() = default;
```
ChrisThrasher, line 51: This will implicitly delete move semantics. Is that acceptable for these types?

ChrisThrasher (commented):
I like a lot about this. My only feedback is about the addition of destructors and how that implicitly deletes move semantics. This won't break user code per se but it can have performance impacts by converting some moves into copies. Do we care enough to `= default` the move operators as well?

moveit_planners/pilz_industrial_motion_planner_testutils/include/pilz_industrial_motion_planner_testutils/goalconstraintsmsgconvertible.h
rhaschke: Valid point. I added default move/copy constructors and assignment operators.

rhaschke:
> Would you like to merge this now or wait on that pull request to update google test?

I suggest merging now to reenable stricter CI. The newer google test doesn't pose a problem locally. Your build log is just more cluttered with warnings (and you can't use `-Werror`). But that's the current status quo anyway.

moveit_planners/pilz_industrial_motion_planner_testutils/include/pilz_industrial_motion_planner_testutils/goalconstraintsmsgconvertible.h
rhaschke: … and cleaned them up again where default constructors were not feasible.

henningkayser (approved):
Why haven't we merged this yet? The only hold-back I see is that we might want to add a note to some documentation to locally switch to gtest1.11, but otherwise I agree with @rhaschke that we should not have to wait for the ament fix. Also, we should backport this to humble, right?

PR #1721, opened 2022-11-16
Use a stronger source of randomness

ChrisThrasher:
### Description

`rsl::rng()` seeds a Mersenne twister to the full extent allowed by the `<random>` API so that we can ensure we're getting the highest quality randomness possible. Using this global random number generator also lets us remove a few member variables that existed for this same purpose.

`std::random_shuffle` has been deprecated and removed from the standard so I also replaced the remaining uses of that function with `std::shuffle`.

### Checklist

### DIFF
--- a/moveit2.repos
+++ b/moveit2.repos
@@ -16,4 +16,8 @@ repositories:
   generate_parameter_library:
     type: git
     url: https://github.com/PickNikRobotics/generate_parameter_library.git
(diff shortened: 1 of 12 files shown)


tylerjw (approved):
Looks good, will merge after CI runs.

tylerjw (approved):

PR #1748, opened 2022-11-19
Use upstream dependencies from ROS

tylerjw:
Signed-off-by: Tyler Weaver <maybe@tylerjw.dev>

### Description

All upstream dependencies have synced. This removes them from the repos file.


tylerjw:
I forgot this is needed for both rolling and humble and humble has not synced. I will close this for now.

PR #1751, opened 2022-11-19
Remove `MOVEIT_LIB_NAME`

### LINKED ISSUE #1720: Replace `MOVEIT_LIB_NAME` with target names
### Description

The `MOVEIT_LIB_NAME` CMake variable creates an extra layer of indirection when writing target-based CMake commands. This practice does not aid readability while harming the grep-ability of the build scripts. The variable ends up getting re-used for multiple targets which further harms ones' ability to understand the build scripts.

I propose we simply remove all instances of that variable and instead spell out the target name where it's being used. 

I'm not positive where this pattern came from but I believe it originated in the ROS 1 documentation and has been copied ever since them. I can't think of any compelling reasons to keep doing this. I've never heard any CMake experts recommend doing this nor do I think Kitware themselves recommend it.

ChrisThrasher:
### Description

Closes #1720

### Checklist

### DIFF
--- a/doc/MIGRATION_GUIDE.md
+++ b/doc/MIGRATION_GUIDE.md
@@ -35,7 +35,7 @@ For instance, the file `joint_model_group.cpp` inside the library `moveit_robot_
 
     static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_robot_model.joint_model_group");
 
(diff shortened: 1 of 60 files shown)


tylerjw (approved):
This is a nice cleanup, thank you.

henningkayser (approved):

ChrisThrasher:
Do I need to rebase on `main` before this can be merged? I'm not sure if that's a requirement or not.

Issue #854, opened 2021-12-01
{} Are Required for Blocks.


tylerjw:
https://github.com/cpp-best-practices/cppbestpractices/blob/master/03-Style.md#-are-required-for-blocks

kiwipeace01:
Hi, I'd like to work on this issue if that's alright

tylerjw:
hello @kiwipeace01 this issue is already taken by cory

kiwipeace01:
oh okay sorry about that! I'm new to open-source so I was a little confused 😅

PR #1753, opened 2022-11-21
Allow omitting braces

ChrisThrasher:
### Description

Our use of clang-format allows us to avoid this check by elimiting all potential misleading indentation. We can and should avoid adding braces in many circumstances where the shorter brace-free form leads to more readable code. The best example of this are guard clauses where

```cpp
if(vector.empty())
  return;
```

becomes the much more verbose

```cpp
if(vector.empty())
{
  return;
}
```

### Checklist


tylerjw (approved):

corycrean:
I'm fine either way on this one.  My personal preference is to add braces around everything for consistency, but that's just me.

One practical downside of _not_ having the braces is that if someone writes:
```
if(vector.empty())
  return;
```
and then I want to modify it to be:
```
if(vector.empty())
{
  doThing();
  return;
}
```
I'll have to add the braces myself, and if I forget, it creates a problem that won't be caught by clang-tidy or the compiler (clang-tidy will fix the indentation, of course, but not the main problem).

ChrisThrasher:
If you forget to add the braces, GCC's `-Wmisleading-indentation` will save you. If you run clang-format to fix the formatting then you'll still likely end up with code that is easily caught in code review. 

https://godbolt.org/z/4sqaoWcMf

sjahr (commented):
As @corycrean, I personally like the version with braces more because a single line 'if' body becomes more prominent (and the compiler won't help me (yet?) with reading the code carefully :sweat_smile: ). That is also pointed out as an argument for the current style in the referenced book of #854. 

Additionally, it would be nice to handle this uniform across the codebase and #999 enforced the '{}' style, so if we'll merge this, we should undo the previous change.

henningkayser:
I'm personally against omitting braces and I find single-line blocks better readable and maintainable. It enforces consistency, allows easy modifications in the condition's scope, and the extra verbosity forces the author to reduce branching. Higher information density does not always equal better readability, imo.

tylerjw:
I'll close this for now, as we do not have an agreement. We can revisit this in the future.

PR #1758, opened 2022-11-22
Require CMake 3.22 everywhere

ChrisThrasher:
### Description

Most of MoveIt already requires CMake 3.22 but there were a few projects that seemingly missed getting updated. `main` includes 42 instances of `cmake_minimum_required(VERSION 3.22)` and this PR only adds 5 more.

### Checklist

### DIFF
--- a/moveit_commander/CMakeLists.txt
+++ b/moveit_commander/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1.3)
+cmake_minimum_required(VERSION 3.22)
 project(moveit_commander)
 
 find_package(catkin REQUIRED)
--- a/moveit_experimental/collision_distance_field_ros/CMakeLists.txt
+++ b/moveit_experimental/collision_distance_field_ros/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1.3)
+cmake_minimum_required(VERSION 3.22)
 include($ENV{ROS_ROOT}/core/rosbuild/rosbuild.cmake)
 
 # Set the build type.  Options are:
--- a/moveit_experimental/kinematics_cache/v1/kinematics_cache/CMakeLists.txt
+++ b/moveit_experimental/kinematics_cache/v1/kinematics_cache/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1.3)
+cmake_minimum_required(VERSION 3.22)
 include($ENV{ROS_ROOT}/core/rosbuild/rosbuild.cmake)
 
 # Set the build type.  Options are:
--- a/moveit_experimental/kinematics_cache/v1/kinematics_cache_ros/CMakeLists.txt
+++ b/moveit_experimental/kinematics_cache/v1/kinematics_cache_ros/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1.3)
+cmake_minimum_required(VERSION 3.22)
 include($ENV{ROS_ROOT}/core/rosbuild/rosbuild.cmake)
 
 # Set the build type.  Options are:
--- a/moveit_planners/trajopt/CMakeLists.txt
+++ b/moveit_planners/trajopt/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1.3)
+cmake_minimum_required(VERSION 3.22)
 project(moveit_planners_trajopt)
 
 find_package(catkin REQUIRED


tylerjw (approved):

jdlangs:
@ChrisThrasher I had some interest in building from source in Ubuntu 20.04 but this is an obstacle. I assume this version was chosen not because of a needed CMake feature but because it's the one packaged with Ubuntu 22.04. Would there be willingness to lower the version either based on the needed functionality or just to the one installed on Focal? I understand there's no official support for building/running on Focal but it doesn't seem necessary to force workarounds for anyone who would like to make the effort.

ChrisThrasher:
You're right it was chosen because it's what ships with Ubuntu 22. I'm not willing to lower the version back down to 3.16 to support an unsupported platform, sorry.

jdlangs:
Understood, thanks for the quick reply.

gavanderhoorn:
@jdlangs: you might be able to get away with `pip install --user cmake`. Even on Focal, this should get you `3.25` right now.

ChrisThrasher:
I'll also recommend the [Kitware Apt Repo](https://apt.kitware.com) which I've used before and quite like. It continuously ships the latest CMake release.

jdlangs:
Yeah, now it seems obvious I should have just upgraded CMake instead of asking about modifying the code. Thanks for the quick help, guys!

gavanderhoorn:
> Thanks, I would never have guessed CMake was packaged on PyPI. I was looking into manually installing the latest version but this is even easier.

well no guarantee. That CMake might not be able to find all the same modules and/or config files as your system's package manager installed one. It's worth a try though.

> I'll also recommend the [Kitware Apt Repo](https://apt.kitware.com) which I've used before and quite like. It continuously ships the latest CMake release. 

I was under the impression it was actually deprecated and Kitware now recommends using `pip` whenever possible.

> Yeah, now it seems obvious I should have just upgraded CMake instead of asking about modifying the code. Thanks for the quick help, guys!

it's not that obvious actually. Whether it'll work is rather dependent on what you're trying to do and how your system is configured.

ChrisThrasher:
> I was under the impression it was actually deprecated and Kitware now recommends using pip whenever possible.

I haven't heard anything of the sort. Looks like it's still maintained but I reached out to a CMake maintainer for confirmation.

> Replacing the platform default CMake with an upgraded one might also not work reliably or have negative consequences. A user install mostly avoids those.

I'd doubt that. I've never run into issues with using a different version of CMake than what the system package manager happened to provide.

gavanderhoorn:
> I'd doubt that. I've never run into issues with using a different version of CMake than what the system package manager happened to provide.

I'm happy for you :) But I have, on multiple occasions, so I thought I'd report what I've observed.

ChrisThrasher:
What in particular went wrong when you tried to do that?

gavanderhoorn:
I've worked with/on platforms where a patched CMake was needed in order to work properly (platform detection, compiler settings, etc). The generic ones provided by Kitware doesn't have those patches.

Some platforms also install CMake find scripts and other modules in a version specific directory. Those don't always get picked up by generic CMake binaries installed using `pip`. The `apt` repository provided by Kitware causes the OS default version to get pruned, which will likely not have that problem, but then the problem I described above has a chance to pop up.

v4hn:
> > Yeah, now it seems obvious I should have just upgraded CMake instead of asking about modifying the code. Thanks for the quick help, guys!
> 
> it's not that obvious actually

I do not think it's obvious either. There is no need to break code compatibility with a platform on that level, unless the cmake files actually *required* features from that new version. Valid reasons to upgrade it are basically:

- some new feature is used, in which case the version set should correspond to the oldest version that is compatible
- cmake on the main target platform (I Ubuntu 22.04 right now) warns excessively at configure time that the specified version is not or only partially compatible with the current version

The commit message nonchalantly states

>  Require CMake 3.22 everywhere #1758 

and the reason given is that many other packages use this version. That's neither of the reasons above. The other locations might actually require it in which case it makes sense to be consistent, but that's at least not clear from this PR.

If you successfully compile the main branch on a different platform and the only error you see is the required cmake version throughout the branch, I consider it a clear bugfix to set the number to a lower value again.

ChrisThrasher:
It's not a "bug" to require CMake 3.22. We've made the statement that we're going to try to get the most out of what Ubuntu 22 gives us. Over time we will start using features added after 3.16. It's an easy first step to bump up the version so all those features are unlocked for us.

gavanderhoorn:
Guys, relax. It's just a different way of working. One tries to only bump versions when (absolutely) needed based on features / functionality required. The other just increases versions whenever possible to 'get access' to new functionality, even if it isn't used yet.

Bw compatibility will suffer with the second approach, but ok, that's a choice that's apparently been made.

ChrisThrasher:
It's more than just features we care about. Upgrading the requirement to CMake 3.22 also changes [all these policies](https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html) that have been introduced between CMake 3.16 and 3.22 which improve the default behavior of CMake.

gavanderhoorn:
Fine.

It doesn't change the fact that something which could still build with different versions of CMake on other platforms now can't, something which your users might want to do.

But again, that's a choice. And one apparently made by you guys.

AndyZe:
I think there are some communication issues here, partially based around the word "we." We does not include me, for the record  ;)   I don't actually care much about the result of this discussion.

@ChrisThrasher you realize @v4hn and @gavanderhoorn are both MoveIt maintainers, right?

ChrisThrasher:
I've been operating under the assumption that MoveIt only supports Ubuntu 22 and my PRs got merged as such. If the maintainers want to change what platforms are supported that is their prerogative. Take this up with the maintainers who approved https://github.com/ros-planning/moveit2/pull/1550.

henningkayser:
I don't really get what the big deal is. If someone wants to build the main branch on an unsupported distro, installing a recent cmake version is the smallest hurdle. At the same time it's introducing new policies that improve cleanliness and style. We are also not reducing the clang-tidy version or similar to allow the defaults from older distros to work.

gavanderhoorn:
There is no *big deal* as far as I'm concerned. Just a couple of differing opinions. Sometimes a discussion is just a discussion.

tylerjw:
I think it is generally a waste of time to try to support old Ubuntu versions on the main branch. There is much more to Ubuntu than just CMake. Just because it builds doesn't mean the behavior is not different and we have a hard enough time keeping it working on 22.04 without worrying about 20.04.

As a compromise how about we don't backport this to the humble branch as humble is tier 3 on 20.04. Windows is tier 1, but good luck getting that to work.

v4hn:
> I think it is generally a waste of time to try to support old Ubuntu versions on the main branch.

Fully agreed (It's a nightmare in moveit/master that we still target melodic). 
Just as much as that I think it is a waste of time to write patches that do nothing (improve readability is not "nothing" here) but break backward compatibility.

I basically wanted to say that if someone comes along and tries to build the project  with a different version and these lines are the only change they need, I'll gladly accept the pr if it passes CI.
As Gijs pointed out, it's different work modes.
I also would not have merged the commit @ChrisThrasher linked, but Vatan did at the time and I don't care as long as nobody complains.

PR #1764, opened 2022-11-28
Migrate to Ogre.h

stephanie-eng:
### Description

When building, I get the following message:
```
In file included from /workspaces/ros2-rolling-ws/src/moveit2/moveit_ros/visualization/rviz_plugin_render_tools/src/planning_link_updater.cpp:38:
/opt/ros/rolling/opt/rviz_ogre_vendor/include/OGRE/OgreVector3.h:2:62: note: ‘#pragma message: /opt/ros/rolling/opt/rviz_ogre_vendor/include/OGRE/OgreVector3.h is deprecated, migrate to Ogre.h’
    2 | #pragma message( __FILE__ " is deprecated, migrate to Ogre.h")
```

This PR uses Ogre.h instead of the deprecated OgreVector3.h as per the message suggests.

### DIFF
--- a/moveit_ros/visualization/rviz_plugin_render_tools/src/planning_link_updater.cpp
+++ b/moveit_ros/visualization/rviz_plugin_render_tools/src/planning_link_updater.cpp
@@ -36,7 +36,6 @@
 
 #include <moveit/rviz_plugin_render_tools/planning_link_updater.h>
 #include <Ogre.h>
-#include <OgreQuaternion.h>
 
 bool moveit_rviz_plugin::PlanningLinkUpdater::getLinkTransforms(const std::string& link_name,
                                                                 Ogre::Vector3& visual_position,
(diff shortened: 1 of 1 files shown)


moveit_ros/visualization/rviz_plugin_render_tools/src/planning_link_updater.cpp
```diff
@@ -35,7 +35,7 @@
 /* Author: Ioan Sucan */
 
 #include <moveit/rviz_plugin_render_tools/planning_link_updater.h>
-#include <OgreVector3.h>
+#include <Ogre.h>
 #include <OgreQuaternion.h>
```
AndyZe, line 39: Can you check if Ogre.h includes OgreQuaternion.h?

AndyZe (approved):

moveit_ros/visualization/rviz_plugin_render_tools/src/planning_link_updater.cpp
stephanie-eng: It does not include it directly, but it does include OgreVector.h which includes OgreQuaternion.h

AndyZe: Let's remove OgreQuaternion.h from here, then

PR #1767, opened 2022-11-29
Do not allow a mutable ACM to be returned from the planning scene

AndyZe:
As discussed here:  https://github.com/orgs/ros-planning/discussions/1760#discussioncomment-4265269

This is part of simplifying locking of the planning scene.

Should be merged with the MTC PR:  https://github.com/ros-planning/moveit_task_constructor/pull/398

There are still some other PS methods that return mutable values that should get cleaned up.

### DIFF
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -329,7 +329,9 @@ void PlanningScene::pushDiffs(const PlanningScenePtr& scene)
   }
 
   if (acm_)
-    scene->getAllowedCollisionMatrixNonConst() = *acm_;
+  {
+    scene->update(*acm_);
+  }
 
   collision_detection::CollisionEnvPtr active_cenv = scene->getCollisionEnvNonConst();
   active_cenv->setLinkPadding(collision_detector_->cenv_->getLinkPadding());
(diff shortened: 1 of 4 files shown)


moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -532,11 +534,40 @@ void PlanningScene::setCollisionObjectUpdateCallback(const collision_detection::
   current_world_object_update_callback_ = callback;
 }
 
-collision_detection::AllowedCollisionMatrix& PlanningScene::getAllowedCollisionMatrixNonConst()
+void PlanningScene::update(collision_detection::AllowedCollisionMatrix const& new_acm)
 {
-  if (!acm_)
-    acm_ = std::make_shared<collision_detection::AllowedCollisionMatrix>(parent_->getAllowedCollisionMatrix());
-  return *acm_;
```
rhaschke, line 539: You are missing the creation of a new ACM in all your modification functions!

rhaschke (commented):
I don't yet see how these changes improve performance. 
See https://github.com/orgs/ros-planning/discussions/1760#discussioncomment-4268025.

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -329,7 +329,9 @@ void PlanningScene::pushDiffs(const PlanningScenePtr& scene)
   }
 
   if (acm_)
-    scene->getAllowedCollisionMatrixNonConst() = *acm_;
+  {
+    scene->update(*acm_);
+  }
```
tylerjw, line 334: You removed the check for if acm_ is a valid pointer?

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -532,11 +534,40 @@ void PlanningScene::setCollisionObjectUpdateCallback(const collision_detection::
   current_world_object_update_callback_ = callback;
 }
 
-collision_detection::AllowedCollisionMatrix& PlanningScene::getAllowedCollisionMatrixNonConst()
+void PlanningScene::update(collision_detection::AllowedCollisionMatrix const& new_acm)
 {
-  if (!acm_)
-    acm_ = std::make_shared<collision_detection::AllowedCollisionMatrix>(parent_->getAllowedCollisionMatrix());
-  return *acm_;
+  acm_ = std::make_shared<collision_detection::AllowedCollisionMatrix>(new_acm);
```
tylerjw, line 539: If internally, this is stored as a shared_ptr, maybe we should make it a shared_ptr to const, and the update parameter should take a shared_ptr to const. The way you wrote it here it takes the acm as a parameter on the stack and then allocates a acm on the heap inside this shared ptr and copies it into the shared_ptr. While this is safe, it is creating a bunch of unnecessary copies.

AndyZe:
I'm just going to close this until the path forward is better-defined.

PR #1770, opened 2022-11-30
Enable `-Wold-style-cast`

ChrisThrasher:
### Description

Add `-Wold-style-cast` and fix all new warnings. Expands on #1628.

In many places I'm just using C++ casts and those are fine. However, the big issue is that all the ROS `_THROTTLE` logging macros (through a few more macros nested within) eventually call `RCUTILS_LOG_CONDITION_THROTTLE_BEFORE` which being from a C library understandably uses a C-style cast. See `rcutils/logging_macros.h:176` copied below for convenience.

```c
/**
 * \def RCUTILS_LOG_CONDITION_THROTTLE_BEFORE
 * A macro initializing and checking the `throttle` condition.
 */
#define RCUTILS_LOG_CONDITION_THROTTLE_BEFORE(get_time_point_value, duration) { \
    static rcutils_duration_value_t __rcutils_logging_duration = RCUTILS_MS_TO_NS((rcutils_duration_value_t)duration); \
    static rcutils_time_point_value_t __rcutils_logging_last_logged = 0; \
    rcutils_time_point_value_t __rcutils_logging_now = 0; \
    bool __rcutils_logging_condition = true; \
    if (get_time_point_value(&__rcutils_logging_now) != RCUTILS_RET_OK) { \
      rcutils_log( \
        &__rcutils_logging_location, RCUTILS_LOG_SEVERITY_ERROR, "", \
        "%s() at %s:%d getting current steady time failed\n", \
        __func__, __FILE__, __LINE__); \
    } else { \
      __rcutils_logging_condition = __rcutils_logging_now >= __rcutils_logging_last_logged + __rcutils_logging_duration; \
    } \
 \
    if (RCUTILS_LIKELY(__rcutils_logging_condition)) { \
      __rcutils_logging_last_logged = __rcutils_logging_now;
```

A proper fix would be to upstream a fix that adds pragmas within these macros that silence the warnings, preferably at the source inside `RCUTILS_LOG_CONDITION_THROTTLE_BEFORE` but potentially at a higher level inside rclcpp macros. My use of `#pragma`s is a noisy but practical option that lets us silence these warnings at the callsite while still enabling the warning which as seen by this PR, has caught many remaining C-style casts.

### Checklist

### DIFF
--- a/moveit_common/cmake/moveit_package.cmake
+++ b/moveit_common/cmake/moveit_package.cmake
@@ -44,7 +44,7 @@ macro(moveit_package)
   if(NOT CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
     # Enable warnings
     add_compile_options(-Wall -Wextra
-      -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls -Wcast-qual)
+      -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls -Wcast-qual -Wold-style-cast)
   else()
     # Defaults for Microsoft C++ compiler
     add_compile_options(/W3 /wd4251 /wd4068 /wd4275)
--- a/moveit_core/collision_detection/src/collision_common.cpp
+++ b/moveit_core/collision_detection/src/collision_common.cpp
@@ -47,6 +47,8 @@ void CollisionResult::print() const
   rclcpp::Clock clock;
   if (!contacts.empty())
   {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wold-style-cast"
     RCLCPP_WARN_STREAM_THROTTLE(LOGGER, clock, LOG_THROTTLE_PERIOD,
                                 "Objects in collision (printing 1st of "
                                     << contacts.size() << " pairs): " << contacts.begin()->first.first << ", "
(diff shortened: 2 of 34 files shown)


henningkayser:
Thank you for enabling this. I think we should get upstream fixed before merging this, though. Alternatively, wouldn't it be possible to silence these warnings by including rclcpp as [SYSTEM](https://github.com/ament/ament_cmake/blob/rolling/ament_cmake_target_dependencies/cmake/ament_target_dependencies.cmake#L32) dependency?

ChrisThrasher:
> I think we should get upstream fixed before merging this, though. 

Would the rcutils maintainers accept a change that disables a C++ warning in their C library? I wasn't sure about that. These pragma statements are really easy to `grep` for and easy to remove later once this fix is hopefully upstreamed and shipped in a few months. In the meantime I think it's worth having this warning.

> Alternatively, wouldn't it be possible to silence these warnings by including rclcpp as [SYSTEM](https://github.com/ament/ament_cmake/blob/rolling/ament_cmake_target_dependencies/cmake/ament_target_dependencies.cmake#L32) dependency?

Sadly no because the macros expand into something the compiler is treating as 1st party code.

tylerjw (approved):
I think this is a useful enough warning to live with the pragmas until this is upstreamed. Once it is upstreamed it should be easy enough to grep for and remove them.

tylerjw (approved):

tylerjw:
@henningkayser I think it is easier to accept this PR now and have Chris make a PR to remove these pragma statements when his upstream change is accepted than to have him rebase or remake this PR in 2-6mo after the upstream change is made.

vatanaksoytezer (approved):
I think I am happy with the changes as is as well. Thank you for doing this!

henningkayser (approved):

ChrisThrasher:
Fixed another C-style cast that just got merged. Looks like this PR will have an immediate impact once merged.

PR #1771, opened 2022-12-01
Add a version of TOTG computeTimeStamps() for a fixed num waypoints

### LINKED ISSUE #1769: Add a capability to specify the exact num_waypoints of a trajectory
I think Henning was the first one to propose this method to resample a trajectory, yielding the exact `num_waypoints` a user requests:

- Run TOTG once to figure out the optimal trajectory duration
- Calculate the timestep to get the desired num_waypoints:
-     delta_t = duration / (num_waypoints-1)     // subtract one for the initial waypoint
-     Run TOTG again with the new timestep. This should give the exact num_waypoints you want

The TOTG function is here:[https://github.com/ros-planning/moveit/blob/1f4435190f7498d746b8066fb72c956eebe8ee[…]veit/trajectory_processing/time_optimal_trajectory_generation.h](https://github.com/ros-planning/moveit/blob/1f4435190f7498d746b8066fb72c956eebe8eeef/moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h#L167)

It would be cool to add this to the MoveIt API.

AndyZe:
Fixes #1769

### DIFF
--- a/moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
+++ b/moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
@@ -175,7 +175,7 @@ public:
                          const std::unordered_map<std::string, double>& acceleration_limits) const override;
 
   /**
-   * \brief Compute a trajectory with the given number of waypoints
+   * \brief Compute a trajectory with the given number of waypoints (plus or minus 1, due to rounding)
    */
   bool computeTimeStamps(robot_trajectory::RobotTrajectory& trajectory, const size_t num_waypoints,
                          const double max_velocity_scaling_factor = 1.0,
--- a/moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
+++ b/moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
@@ -174,10 +174,17 @@ public:
                          const std::unordered_map<std::string, double>& velocity_limits,
                          const std::unordered_map<std::string, double>& acceleration_limits) const override;
 
+  // clang-format off
   /**
    * \brief Compute a trajectory with the given number of waypoints (plus or minus 1, due to rounding).
    * Caution: if the waypoint spacing is sparse, execution of the trajectory may vary from the expected path.
+   * \param num_waypoints The desired number of waypoints.
+   * \param[in,out] trajectory A path which needs time-parameterization. It's OK if this path has already been
+   * time-parameterized; this function will re-time-parameterize it.
+   * \param max_velocity_scaling_factor Joint velocity limits are scaled by this factor.
+   * \param max_acceleration_scaling_factor Joint acceleration limits are scaled by this factor.
    */
+  // clang-format on
   bool computeTimeStamps(const size_t num_waypoints, robot_trajectory::RobotTrajectory& trajectory,
                          const double max_velocity_scaling_factor = 1.0,
                          const double max_acceleration_scaling_factor = 1.0);
--- a/moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
+++ b/moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
@@ -177,7 +177,7 @@ public:
   /**
    * \brief Compute a trajectory with the given number of waypoints (plus or minus 1, due to rounding)
    */
-  bool computeTimeStamps(robot_trajectory::RobotTrajectory& trajectory, const size_t num_waypoints,
+  bool computeTimeStamps(const size_t num_waypoints, robot_trajectory::RobotTrajectory& trajectory,
                          const double max_velocity_scaling_factor = 1.0,
                          const double max_acceleration_scaling_factor = 1.0);
 
--- a/moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
+++ b/moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
@@ -177,7 +177,8 @@ public:
   // clang-format off
   /**
    * \brief Compute a trajectory with the given number of waypoints (plus or minus 1, due to rounding).
-   * Caution: if the waypoint spacing is sparse, execution of the trajectory may vary from the expected path.
+   * Resampling the trajectory to get the desired num_waypoints doesn't change the shape of the trajectory,
+   * but controller execution may deviate from the intended path if waypoint spacing is too sparse.
    * \param num_waypoints The desired number of waypoints.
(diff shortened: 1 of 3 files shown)


moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -1057,6 +1057,32 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(
...
+
+  // auto input_trajectory = robot_trajectory::RobotTrajectory(trajectory, true /* deep copy */);
+
+  // computeTimeStamps(input_trajectory, max_velocity_scaling_factor, max_acceleration_scaling_factor);
+  // double optimal_duration = input_trajectory.getDuration();
+
+  computeTimeStamps(trajectory, max_velocity_scaling_factor, max_acceleration_scaling_factor);
+  double optimal_duration = trajectory.getDuration();
+
+  std::cout << "Optimal duration: " << optimal_duration << std::endl;
```
AndyZe, line 1079: TODO: remove cout's before merging

moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
```diff
@@ -174,13 +174,20 @@ class TimeOptimalTrajectoryGeneration : public TimeParameterization
                          const std::unordered_map<std::string, double>& velocity_limits,
                          const std::unordered_map<std::string, double>& acceleration_limits) const override;
 
+  /**
+   * \brief Compute a trajectory with the given number of waypoints (plus or minus 1, due to rounding)
```
sjahr, line 178: Would it make sense here to motivate why a finer-grained / coarser trajectory might be necessary and potential damage that can be caused e.g if you down-sample a long trajectory with 120 waypoints to 2? Is it correct to assume that the only guarantee that can be made here is that the start and goal state will be the same? The path characteristics between those two points can change completely

sjahr (commented):
I think this is a useful feature of people know what they're doing

moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
```diff
@@ -174,13 +174,20 @@ class TimeOptimalTrajectoryGeneration : public TimeParameterization
...
+   */
+  bool computeTimeStamps(const size_t num_waypoints, robot_trajectory::RobotTrajectory& trajectory,
```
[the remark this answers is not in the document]
AndyZe: So, the way to look at this is just another overloaded version of `computeTimeStamps`. It can be used the same way, just with the additional `num_waypoints` argument.

moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
AndyZe: Definitely needs a better comment.

AndyZe: This part of your comment is actually an incorrect assumption:

> Is it correct to assume that the only guarantee that can be made here is that the start and goal state will be the same? The path characteristics between those two points can change completely

Resampling doesn't change the path since it occurs after time-parameterization. You can see it here:

https://github.com/ros-planning/moveit2/blob/50b03a1674537613e2b6f26a220e78d8572f0216/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp#L1130

Possibly @henningkayser can confirm this. I'll clarify with a better comment.

moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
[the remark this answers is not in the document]
sjahr: You're right, that is also a way to look at it. I am fine with that :+1:

moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
sjahr: Maybe I am missing something here, but how can that be? If you have an 'S'-shaped path in 2D with 10 waypoints and you down-sample it to 3 waypoints how can this path of 3 waypoints still describe an 'S'?

moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
```diff
@@ -174,13 +174,29 @@ class TimeOptimalTrajectoryGeneration : public TimeParameterization
                          const std::unordered_map<std::string, double>& velocity_limits,
                          const std::unordered_map<std::string, double>& acceleration_limits) const override;
 
+  // clang-format off
+  /**
+   * \brief Compute a trajectory with the given number of waypoints (plus or minus 1, due to rounding).
+   * Resampling the trajectory to get the desired num_waypoints doesn't change the shape of the trajectory,
+   * but controller execution may deviate from the intended path if waypoint spacing is too sparse.
```
sjahr, line 181:
```suggestion
   * Resampling the trajectory to get the desired num_waypoints doesn't change the start and goal point,
   * and all re-sampled waypoints will be on the path of the original trajectory (within the configured tolerances)
   * but controller execution may deviate from the intended path if waypoint spacing is too sparse.
```

sjahr (approved):
I think the documentation is good now. If you don't like the small clarification of your comment I suggested, feel free to reject it

moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
```diff
@@ -174,13 +174,30 @@ class TimeOptimalTrajectoryGeneration : public TimeParameterization
...
+                         const double max_acceleration_scaling_factor = 1.0);
+
 private:
   bool doTimeParameterizationCalculations(robot_trajectory::RobotTrajectory& trajectory,
                                           const Eigen::VectorXd& max_velocity,
                                           const Eigen::VectorXd& max_acceleration) const;
 
   const double path_tolerance_;
-  const double resample_dt_;
+  double resample_dt_;
```
henningkayser, line 200: @AndyZe it's actually error-prone that `resample_dt_` is being changed implicitly by the new `computeTimeStamps()` function. The other member functions have been `const` for a reason, namely to allow repeatable results. Now the results of the old `computeTimeStamps()` depends on if the new function has been called and whatever `num_waypoints` has been used. Please fix that, either by implementing a free function that creates two instances of TOTG (one for each resample_dt), or by providing a private `computeTimeStampsInternal` function that accepts `resample_dt` as argument so that you can call that one twice without changing the member variable.

AndyZe: OK, I'll fix this soon

PR #1772, opened 2022-12-01
Fix trajectory unwind bug

### LINKED ISSUE #1777: RobotTrajectory unwind() functions only account for one extra "wrap"
Unwinding should account for any angle. For example, a UR3e robot's wrist can rotate infinitely. But, the [RobotTrajectory unwind()](https://github.com/ros-planning/moveit2/blob/f0d3202d637569ccac5fcc357a37e0d12a077c87/moveit_core/robot_trajectory/src/robot_trajectory.cpp#L158) function only accounts for one "unwrapping" of 2*pi.

```
      if (last_value > current_value + M_PI)
      {
        running_offset += 2.0 * M_PI;
      }
      else if (current_value > last_value + M_PI)
      {
        running_offset -= 2.0 * M_PI;
      }
```

mechwiz:
### Description
This PR fixes some unwinding/bounds issues for continuous joints.

1. If a trajectory that is being unwound from a reference-state starts (i.e. has an initial waypoint) where the continuous joint has a value above 2PI or below -2PI as could be the case when a trajectory from MTC is generated, I have noticed in a project that I was working on that the `running_offset` can be off by a multiple of 2PI leading to 2PI jumps in the command for that continous joint. This occurs because in the current implementation, the running_offset is not only being calculated using the offset between the reference state and the unwound reference state but also the offset between the initial waypoint (**which may not be unwound**) and the unwound current state which is not valid and is like comparing apples and oranges. To fix this, the initial waypoint of the trajectory also has its bounds enforced so that when it's being compared to the reference value, it's comparing apples to apples. Furthermore, I added logic to ensure that the running_offset takes into account when the unwound reference state and the unwound initial waypoint are not on the same side of PI by adding 2PI or subtracting 2PI accordingly.
2. When `executeAndMonitor` is called, unwinding is not handled properly if there are multiple trajectories within the plan as would happen when executing a task or a serial container from MTC. The way unwinding should work is that the current plan_component's trajectory (item i) should be unwound based on any group-matching previous plan_component's trajectory (item j). The current implementation for some unknown reason is starting at index zero and working up (which is not valid since there could be a matching group plan-component but it might not be the one that precedes component "i") instead of starting at the previous plan compoment and working down.

Fixes #1777 

@AndyZe 

### Checklist

### DIFF
--- a/moveit_core/robot_trajectory/src/robot_trajectory.cpp
+++ b/moveit_core/robot_trajectory/src/robot_trajectory.cpp
@@ -216,9 +216,13 @@ RobotTrajectory& RobotTrajectory::unwind(const moveit::core::RobotState& state)
     double last_value = waypoints_[0]->getJointPositions(cont_joint)[0];
     cont_joint->enforcePositionBounds(&last_value);
     if (last_value > reference_value + M_PI)
+    {
       running_offset -= 2.0 * M_PI;
+    }
     else if (last_value < reference_value - M_PI)
+    {
       running_offset += 2.0 * M_PI;
+    }
     double current_start_value = last_value + running_offset;
     waypoints_[0]->setJointPositions(cont_joint, &current_start_value);
 
--- a/moveit_core/robot_trajectory/src/robot_trajectory.cpp
+++ b/moveit_core/robot_trajectory/src/robot_trajectory.cpp
@@ -168,10 +168,13 @@ RobotTrajectory& RobotTrajectory::unwind()
     // unwrap continuous joints
     double running_offset = 0.0;
     double last_value = waypoints_[0]->getJointPositions(cont_joint)[0];
+    cont_joint->enforcePositionBounds(&last_value);
+    waypoints_[0]->setJointPositions(cont_joint, &last_value);
 
     for (std::size_t j = 1; j < waypoints_.size(); ++j)
     {
       double current_value = waypoints_[j]->getJointPositions(cont_joint)[0];
+      cont_joint->enforcePositionBounds(&current_value);
       if (last_value > current_value + M_PI)
       {
         running_offset += 2.0 * M_PI;
@@ -182,12 +185,8 @@ RobotTrajectory& RobotTrajectory::unwind()
       }
 
       last_value = current_value;
-      if (running_offset > std::numeric_limits<double>::epsilon() ||
-          running_offset < -std::numeric_limits<double>::epsilon())
-      {
-        current_value += running_offset;
-        waypoints_[j]->setJointPositions(cont_joint, &current_value);
-      }
+      current_value += running_offset;
+      waypoints_[j]->setJointPositions(cont_joint, &current_value);
     }
   }
   for (moveit::core::RobotStatePtr& waypoint : waypoints_)
@@ -231,6 +230,7 @@ RobotTrajectory& RobotTrajectory::unwind(const moveit::core::RobotState& state)
     for (std::size_t j = 1; j < waypoints_.size(); ++j)
     {
       double current_value = waypoints_[j]->getJointPositions(cont_joint)[0];
+      cont_joint->enforcePositionBounds(&current_value);
       if (last_value > current_value + M_PI)
       {
         running_offset += 2.0 * M_PI;
@@ -241,16 +241,14 @@ RobotTrajectory& RobotTrajectory::unwind(const moveit::core::RobotState& state)
       }
 
       last_value = current_value;
-      if (running_offset > std::numeric_limits<double>::epsilon() ||
-          running_offset < -std::numeric_limits<double>::epsilon())
-      {
-        current_value += running_offset;
-        waypoints_[j]->setJointPositions(cont_joint, &current_value);
-      }
+      current_value += running_offset;
+      waypoints_[j]->setJointPositions(cont_joint, &current_value);
     }
   }
   for (moveit::core::RobotStatePtr& waypoint : waypoints_)
+  {
     waypoint->update();
+  }
 
   return *this;
 }
--- a/moveit_core/robot_trajectory/src/robot_trajectory.cpp
+++ b/moveit_core/robot_trajectory/src/robot_trajectory.cpp
@@ -215,8 +215,12 @@ RobotTrajectory& RobotTrajectory::unwind(const moveit::core::RobotState& state)
 
     double last_value = waypoints_[0]->getJointPositions(cont_joint)[0];
     cont_joint->enforcePositionBounds(&last_value);
-    double current_value = last_value + running_offset;
-    waypoints_[0]->setJointPositions(cont_joint, &current_value);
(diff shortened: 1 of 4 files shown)


JafarAbdi:
Do you mind adding a test that reproduces the fixed issue?

AndyZe:
I can write a test, as part of reviewing this anyway.

moveit_core/robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -214,12 +214,17 @@ RobotTrajectory& RobotTrajectory::unwind(const moveit::core::RobotState& state)
     double running_offset = reference_value0 - reference_value;
 
     double last_value = waypoints_[0]->getJointPositions(cont_joint)[0];
```
AndyZe, line 216: I'm confused why this is named `last_value`? It looks like `waypoint_value` would be a better name.

mechwiz:
Thanks Andy!
Sorry haven't gotten to the test writing part yet.

moveit_ros/planning/plan_execution/src/plan_execution.cpp
```diff
@@ -357,7 +357,7 @@ moveit_msgs::msg::MoveItErrorCodes plan_execution::PlanExecution::executeAndMoni
     // in the meantime we do a hack:
 
     bool unwound = false;
-    for (std::size_t j = 0; j < i; ++j)
+    for (int j = i - 1; j >= 0; --j)
```
AndyZe, line 360: Let's use longer names to improve readability

```suggestion
    for (size_t prev_component = current_component - 1; prev_component >= 0; --prev_component)
```

moveit_ros/planning/plan_execution/src/plan_execution.cpp
```diff
@@ -357,7 +357,7 @@ moveit_msgs::msg::MoveItErrorCodes plan_execution::PlanExecution::executeAndMoni
...
     {
       // if we ran unwind on a path for the same group
```
AndyZe, line 362:
```suggestion
      // if `current_component` and `prev_component` use the same group, unwind the trajectory
```

moveit_ros/planning/plan_execution/src/plan_execution.cpp
AndyZe: Overall, I'm not sure out this approach. Won't each component be "unwound" repeatedly? Once at the for-loop of L352, then again at L360...

AndyZe: @mechwiz what would you think about splitting these `executeAndMonitor` changes to a different PR, so they don't hold up getting the other bug fix merged?

mechwiz: I don't think so since there is a "break" at L369

AndyZe: I see. Basically the for-loop at L360 goes backward until it finds an earlier component for the same group, then unwinds it. So it can skip components which are not for the same group. Is that right?

mechwiz: correct - just to clarify to make sure we are on the same page "it finds an earlier component for the same group, then unwinds it" -> "it finds an earlier component for the same group, then unwinds the current component based on the last waypoint of the earlier component". yep it should skip any  components not of the same group.

mechwiz: I'm OK with making a separate PR but feel that the time difference between getting each of these issues resolved won't be significant to make it worth it.

AndyZe: Thanks for the explanation, I'm OK with it now. Shame on MoveIt for not having some basic tests for `unwind()`

mechwiz: on a side note - I could be wrong but I think the compiler might complain about using `size_t` here since the value of `j`/`prev_component` on the first pass is - 1.

AndyZe:
Started writing a test, got distracted by other bugs  :/   Will come back to this. Or maybe we merge it as it is, then fix the other bugs and write tests.

AndyZe (approved):
I'm OK merging this without tests. `unwind()` clearly has some other bugs that need to be fixed, anyway. See https://github.com/ros-planning/moveit2/pull/1778

We need to wait for one more review, though.

mechwiz:
I'm cool with merging as is as well!

mechwiz:
Okedoke, I think I fixed up the other unwind bugs and got the unwind test case to be successful

moveit_core/robot_trajectory/test/test_robot_trajectory.cpp
```diff
@@ -150,6 +151,210 @@ class RobotTrajectoryTestFixture : public testing::Test
   }
 };
 
+class OneRobot : public testing::Test
+{
+protected:
+  void SetUp() override
+  {
+    static const std::string MODEL2 =
```
mechwiz, line 159: Note - I pretty much copied this model from here https://github.com/ros-planning/moveit2/blob/main/moveit_core/robot_state/test/test_cartesian_interpolator.cpp#L57

AndyZe (approved):

moveit_core/robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -168,10 +168,13 @@ RobotTrajectory& RobotTrajectory::unwind()
     // unwrap continuous joints
```
AndyZe, line 168: @mechwiz Here's a question. No need to address in this PR but I'm just interested in what you think.

Does it make sense to only unwrap continuous joints? I think you could still have a joint jump if a revolute joint had a large limit range.

mechwiz: I think I only saw this joint jump issue with continuous joints since the unwrapping was being done incorrectly. With revolute joints with large ranges, I don't think there will be any joint jump issues since they are never unwrapped. For example if the range is +/- 6PI, the current state is 6PI, and you specify a goal state of -6PI, the joint will actually move 6 full revolutions whereas for a continuous joint it will just stay as is since they are the same state. In other words, the values used for the revolute joint are absolute (and so no joint jumps will occur) whereas the continuous joint are relative (and so need to be unwinded correctly).

All this said - I could understand the usecase of a revolute joint with large limits "acting" like a continuous joint within the limits it has so that there wouldn't potentially be extra revolutions when going to a named state (which has a hardcoded value) but I guess a line had to be drawn somewhere between the two kinds of joints.

sjahr:
@mechwiz Can you rebase this branch, so we can merge it please

mechwiz:
Done

AndyZe (approved):

PR #1776, opened 2022-12-02
Set the resample_dt_ member of TOTG back to const

AndyZe:
Based on the comment here: https://github.com/ros-planning/moveit2/pull/1771#discussion_r1037646816

Shuffle a function around so that the `resample_dt_` member of TOTG can be const again.


moveit_core/trajectory_processing/test/test_time_optimal_trajectory_generation.cpp
```diff
@@ -519,7 +519,8 @@ TEST(time_optimal_trajectory_generation, testFixedNumWaypoints)
   trajectory.addSuffixWayPoint(waypoint_state, delta_t);
 
   TimeOptimalTrajectoryGeneration totg;
```
henningkayser, line 521:
```suggestion
```

henningkayser (commented):
Thanks for fixing this so quickly. Only request would be to change the free function name to include TOTG somewhere, since `trajectory_processing::computeTimeStamps()` doesn't reveal that.

henningkayser (approved):

PR #1778, opened 2022-12-02
Bug demo: Add an "unwind()" unit test that currently fails

AndyZe:
Am I crazy or should this be a very easy test case for [RobotTrajectory unwind()](https://github.com/ros-planning/moveit2/blob/f0d3202d637569ccac5fcc357a37e0d12a077c87/moveit_core/robot_trajectory/src/robot_trajectory.cpp#L158)? I'll try to fix this.


mechwiz:
I think I fixed this in #1772.
Other than the fact that their were bugs in the function, another reason why this test case didn't work is because the panda robot does not contain a continuous joint. This test case just changes the value of a revolute joint (which is assumed can't be greater than pi or -pi) to a wrapped value which is not valid since the unwind function only looks at continuous joints when unwinding. I therefore added another model in the test which has a continuous joint and now it works as expected.

AndyZe:
If you don't mind, I'll cherry-pick this into #1772 and check. That would be great.

PR #1780, opened 2022-12-03
Remove Iterative Spline and Iterative Parabola time-param algorithms (v2)

AndyZe:
We've been using TOTG by default for awhile now. The only drawback of TOTG I'm aware of is that it can't handle nonzero initial or final velocities/accelerations. So I was thinking that would be a reason to keep the older algorithms (ISTP and IPTP). But, I added a little test and it looks like neither of those algorithms handles it either. Even worse -- they just fail without throwing any error. So I think we should delete them.

3 commits:

- The first commit adds a simple test for IPTP. The test passes but the output is wrong. (It didn't do anything.)

```
20: [ RUN      ] TestTimeParameterization.NonzeroInitialVelAccel
20: Trajectory has 2 points over 0.000 seconds
20:   waypoint   0 time 0.000 pos  0.000 vel  0.000 acc  0.000
20:   waypoint   1 time 0.000 pos  0.000 vel  0.000 acc  0.000
20: [       OK ] TestTimeParameterization.NonzeroInitialVelAccel (0 ms)
```

- The second commit adds a simple test for ISTP. Again, the test passes but the output is wrong. The state at the final waypoint should be (position, vel, accel) == (0.01, 0.1, 0.2).

```
20: [ RUN      ] TestTimeParameterization.NonzeroInitialVelAccel
20: Trajectory has 4 points over 0.343 seconds
20:   waypoint   0 time 0.000 pos  0.000 vel  0.000 acc  0.742
20:   waypoint   1 time 0.118 pos  0.003 vel  0.021 acc -0.394
20:   waypoint   2 time 0.253 pos  0.005 vel  0.029 acc  0.526
20:   waypoint   3 time 0.343 pos  0.010 vel  0.100 acc  1.036
20: [       OK ] TestTimeParameterization.NonzeroInitialVelAccel (0 ms)
```

- The third commit deletes both of the plugins.

Note, this is a rebased version of #1191. I wasn't able to re-open that PR because "the branch was recreated or force-pushed." I'm re-opening the PR because there is new interest in changing the API of TOTG. Specifically, some want to be able to change tolerances or timesteps without constructing a new TOTG object.


tylerjw (approved):
Deleting code is an easy approve. Thank you for looking into this and determining we shouldn't keep this.

PR #1784, opened 2022-12-05
Minimize use of `this->`

ChrisThrasher:
### Description

Using the `this` pointer is often unnecessary. MoveIt already avoids this in most cases so this PR better cements that existing pattern. In some places it must remain because a method parameter shadows a function we're trying to call so removing `this->` causes the build to break.


tylerjw:
clang-format failed on this PR, do you mind running pre-commit and pushing to fix that?

vatanaksoytezer (approved):

PR #1785, opened 2022-12-05
Install moveit_core headers within additional moveit_core directory

ChrisThrasher:
### Description

I changed where headers get installed to fix issues with building MoveIt from source when another copy is already installed. By installing all headers inside a `moveit_core` directory, we add an extra layer of indirection that ensures we know what headers are getting used at any given time.

While I was at it, I also made sure to stop using `include_directories` inside moveit_core. CMake best practices are to avoid using `include_directories`, particularly in large projects like this because it makes it difficult to understand what targets have access to what include directories. Removing them from moveit_core makes it easier for us to know what headers we need to install plus it revealed an issue where kinematics_base and robot_model depended on each other. This makes me wonder if perhaps moveit_core should be consolidated into fewer targets.

### DIFF
--- a/moveit_core/CMakeLists.txt
+++ b/moveit_core/CMakeLists.txt
@@ -55,34 +55,6 @@ include(ConfigExtras.cmake)
 # Set target file path for version.h
 set(VERSION_FILE_PATH ${CMAKE_BINARY_DIR}/include)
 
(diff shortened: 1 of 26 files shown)


vatanaksoytezer (approved):

vatanaksoytezer:
Looks good to me, thank you for doing this Chris!

PR #1790, opened 2022-12-06
Remove unnecessary CMake variables and lists

tylerjw:
### Description

In moveit_core CMakeLists.txt

- Place all the `find_package` calls in a single group and sort them alphabetically.
- Eliminate the lists `THIS_PACKAGE_LIBRARIES` and `THIS_PACKAGE_INCLUDE_DEPENDS` that are only used in one place by putting the contents in the install.
- Update the exported dependencies to include all dependencies we `find_package` on. I think this fixes some transitive dependencies not being exported.


moveit_core/CMakeLists.txt
```diff
@@ -173,16 +118,70 @@ add_subdirectory(collision_detection_fcl)
...
+    moveit_exceptions
+    moveit_kinematics_base
+    moveit_kinematic_constraints
+    moveit_kinematics_metrics
+    moveit_macros
+    moveit_planning_interface
+    moveit_planning_scene
+    moveit_planning_request_adapter
+    # TODO: Port python bindings
+    # moveit_python_tools
```
ChrisThrasher, line 139: Do this TODO need to stick around?

ChrisThrasher (approved):

moveit_core/CMakeLists.txt
tylerjw: There is already the TODO above on line 91 for this, so I'm going to remove this comment.

moveit_core/CMakeLists.txt
```diff
@@ -55,63 +53,10 @@ include(ConfigExtras.cmake)
 # Set target file path for version.h
 set(VERSION_FILE_PATH ${CMAKE_BINARY_DIR}/include)
```
ChrisThrasher, line 54: This might deserve inlining too.

tylerjw: good catch

vatanaksoytezer (approved):
Looks good to me!

PR #1791, opened 2022-12-07
Update the Servo dependency on realtime_tools

AndyZe:
Update the location of the `thread_priority` library which was previously lumped in with the `controller_manager` package

### DIFF
--- a/moveit2.repos
+++ b/moveit2.repos
@@ -1,4 +1,5 @@
 repositories:
+  # Keep moveit_* repos here because they are released with moveit
   moveit_msgs:
     type: git
     url: https://github.com/ros-planning/moveit_msgs.git
--- a/moveit2.repos
+++ b/moveit2.repos
@@ -1,17 +1,9 @@
 repositories:
+  moveit_msgs:
+    type: git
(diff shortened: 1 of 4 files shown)


tylerjw:
To get this to work I think you'll have to remove ros2_control from the repos file.

tylerjw:
Per the CI failure on this one we are waiting on a Rolling sync for this to pass.

moveit2.repos
```diff
@@ -1,21 +1,2 @@
 repositories:
-  moveit_msgs:
-    type: git
-    url: https://github.com/ros-planning/moveit_msgs.git
-    version: ros2
-  moveit_resources:
-    type: git
-    url: https://github.com/ros-planning/moveit_resources.git
-    version: ros2
```
rhaschke, line 9: As stated in https://github.com/ros-planning/moveit2/pull/1827, these MoveIt-related packages should be kept here as we typically release them together with MoveIt.

rhaschke (commented):
Looks good, but keep `moveit_*` packages.

rhaschke (approved):

tylerjw (approved):

PR #1792, opened 2022-12-07
Unit test LinearPathSegment

tylerjw:
### Description

This pull request adds tests and documentation to the LinearPathSegment class. The tests check that the constructor works correctly, that the getConfig(), getTangent(), getCurvature(), and getSwitchingPoints() methods return the expected results, and that the clone() method creates a deep copy. The documentation provides a description of the class and detailed descriptions of all the public methods.

### DIFF
--- a/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
+++ b/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
@@ -36,13 +36,15 @@
  *   POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <moveit/trajectory_processing/time_optimal_trajectory_generation.h>
+#include <moveit/trajectory_processing/linear_path_segment.hpp>
+
 #include <rclcpp/logger.hpp>
 #include <rclcpp/logging.hpp>
 #include <limits>
 #include <Eigen/Geometry>
 #include <algorithm>
 #include <cmath>
-#include <moveit/trajectory_processing/time_optimal_trajectory_generation.h>
 #include <vector>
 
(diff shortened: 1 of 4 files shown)


moveit_core/trajectory_processing/test/test_linear_path_segment.cpp
```diff
@@ -0,0 +1,102 @@
+#include <gtest/gtest.h>
+#include <moveit/trajectory_processing/linear_path_segment.hpp>
+
+using trajectory_processing::LinearPathSegment;
+
+// Tests LinearPathSegment constructor.
+TEST(LinearPathSegment, Constructor)
+{
+  Eigen::VectorXd start(2);
```
ChrisThrasher, line 9: Can we test with fixed size vectors or are those not convertible to the `Xd` variable-sized vectors?

ChrisThrasher (changes requested):

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -55,46 +57,6 @@ constexpr double DEFAULT_TIMESTEP = 1e-3;
 constexpr double EPS = 1e-6;
 }  // namespace
 
-class LinearPathSegment : public PathSegment
```
ChrisThrasher, line 58: So we're promoting `LinearPathSegment` from an implementation detail to a public interface? Are we sure we want to do that and start making guarantees on its stability?

tylerjw:
I'm closing this as I abandoned this work and did not finish it.

PR #1794, opened 2022-12-08
Require velocity and acceleration limits in TOTG

### LINKED ISSUE #1793: If no acceleration limit is given in joint_limits.yaml or URDF, it defaults to 1.0 rad/s^2
Defaulting to some random-ish value can cause unexpected behavior. [In the case of UR robots](https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/issues/46), it was causing slower-than-expected motion.

The correct way to handle this would probably be to print a warning then exit if no acceleration limit is given.

AndyZe:
This PR prints an error and returns false if vel/accel isn't defined for a robot joint, instead of assigning a default value and moving on.

Fixes #1793 and this UR issue:  https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/issues/46


henningkayser:
Things like this make me wonder if we should always enforce having velocity and acceleration limits set at runtime. There is really no actual use case of MoveIt that makes sense without them, imo.

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -938,12 +937,12 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
     }
     else
     {
-      RCLCPP_WARN_STREAM_ONCE(
-          LOGGER, "Joint velocity limits are not defined. Using the default "
-                      << max_velocity[j] << " rad/s. You can define velocity limits in the URDF or joint_limits.yaml.");
+      RCLCPP_ERROR_STREAM(LOGGER, "No velocity limit was defined for joint "
+                                      << vars[j].c_str()
+                                      << "! You can define velocity limits in the URDF or joint_limits.yaml");
```
sjahr, line 942:
```suggestion
                                      << "! You have to define velocity limits in the URDF or joint_limits.yaml");
```

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -957,10 +956,11 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
     }
     else
     {
-      RCLCPP_WARN_STREAM_ONCE(LOGGER,
-                              "Joint acceleration limits are not defined. Using the default "
-                                  << max_acceleration[j]
-                                  << " rad/s^2. You can define acceleration limits in the URDF or joint_limits.yaml.");
+      RCLCPP_ERROR_STREAM(LOGGER, "No acceleration limit was defined for joint "
+                                      << vars[j].c_str()
+                                      << "! You can define acceleration limits in the URDF or "
```
sjahr, line 961:
```suggestion
                                      << "! You have to define acceleration limits in the URDF or "
```

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -1016,10 +1016,10 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(
 
     if (!set_velocity_limit)
     {
-      max_velocity[j] = 1.0;
-      RCLCPP_WARN_STREAM_ONCE(
-          LOGGER, "Joint velocity limits are not defined. Using the default "
-                      << max_velocity[j] << " rad/s. You can define velocity limits in the URDF or joint_limits.yaml.");
+      RCLCPP_ERROR_STREAM(LOGGER, "No velocity limit was defined for joint "
+                                      << vars[j].c_str()
+                                      << "! You can define velocity limits in the URDF or "
```
sjahr, line 1021:
```suggestion
                                      << "! You have to define velocity limits in the URDF or "
```

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -1016,10 +1016,10 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(
...
+                                         "joint_limits.yaml");
```
sjahr, line 1022:
```suggestion
                                         "joint_limits.yaml");
      return false;
```

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -1046,11 +1046,10 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(
...
     {
-      max_acceleration[j] = 1.0;
-      RCLCPP_WARN_STREAM_ONCE(LOGGER,
-                              "Joint acceleration limits are not defined. Using the default "
-                                  << max_acceleration[j]
-                                  << " rad/s^2. You can define acceleration limits in the URDF or joint_limits.yaml.");
+      RCLCPP_ERROR_STREAM(LOGGER, "No acceleration limit was defined for joint "
+                                      << vars[j].c_str()
+                                      << "! You can define acceleration limits in the URDF or "
+                                         "joint_limits.yaml");
```
sjahr, line 1052:
```suggestion
                                      << "! You have to define acceleration limits in the URDF or "
                                         "joint_limits.yaml");
      return false;
```

moveit_core/trajectory_processing/test/test_time_optimal_trajectory_generation.cpp
```diff
@@ -274,96 +306,14 @@ TEST(time_optimal_trajectory_generation, testLastWaypoint)
...
-  add_waypoint({ 0.032136103, 0.032136103 });
-  add_waypoint({ 0.032532845, 0.032532845 });
-  add_waypoint({ 0.032929587, 0.032929587 });
-  add_waypoint({ 0.033326329, 0.033326329 });
-  add_waypoint({ 0.033723071, 0.033723071 });
-  add_waypoint({ 0.034119813, 0.034119813 });
-  add_waypoint({ 0.034516555, 0.034516555 });
-
-  const std::vector<double> expected_last_waypoint = { 0.034913297, 0.034913297 };
+  add_waypoint({ 0.000000000, 0.000000000, 0, 0, 0, 0, 0 });
```
sjahr, line 309: I guess you removed a number of waypoints because the test trajectory does not have to be that long, right?

sjahr (commented):
I think that change does make sense :+1: Here are a couple of comments

AndyZe:
I think I'll just close this one because [the discussion](https://github.com/ros-planning/moveit2/discussions/1804) is tending toward requiring acceleration limits when the RobotModel is loaded.

AndyZe:
Reviving this PR since the discussion about requiring accel limits seems to be a dead end.

moveit_core/trajectory_processing/test/test_time_optimal_trajectory_generation.cpp
AndyZe: yeah, exactly.

sjahr (approved):

v4hn:
I'm surprised to see this change merged. Functionally the only thing it does is that you prohibit a use-case that did something intuitive but possibly sub-optimal before while printing an explicit warning. I don't see what the patch improves...

The least you should do is document migration notes telling users to explicitly add the default 1.0 rad/s^2 limits for all joints if they want it to work the way it did so far.
E.g., when controlling a robotiq 3-finger adaptive gripper, I am actually annoyed if I have to specify acceleration limits for each joint just to close the gripper. As pointed out in the linked UR issue, manufacturers do not provide that information.
Additionally, especially for beginners acceleration is completely irrelevant when they just want to move from A to B. I tend to explain the difference between perceived speed, velocity and acceleration quite some time after students set up their first configs and moved robots around via MoveIt.

sjahr:
I don't think that this patch prohibits a use-case but rather helps users to avoid an unexpected behavior of the robot moving very slow. Yes, a warning was printed but it was easy to overlook in the huge terminal output that MoveIt produces. As an engineer, I'd assume the acceleration limits to be treated as unlimited (which would prob. be an unsafe default value :sweat_smile: ), if I don't explicitly set them and not a magic number.

But it is a good point, that we should recommend  1.0 rad/s^2 as default for beginners :+1:

v4hn:
> I don't think that this patch prohibits a use-case

Well, it clearly does not add any functionality and it exits an algorithm when the algorithm ran without error before. ;)

> but rather helps users to avoid an unexpected behavior of the robot moving very slow.

They will still get that because of the independent default scaling parameters which make the robot slow unless overridden. :D

It's a valid question to ask which variables should have defaults and which should not.
MoveIt traditionally has defaults for everything to ease initial setup cost and the values are rarely great.
In my opinion, acceleration is far enough away from beginner use though to warrant a default (of 1.0/s^2).

Thinking along these lines, would it be a viable way around this whole discussion to add explicit defaults for velocity and acceleration to the joint_limits.yaml that apply whenever a joint does not define limits explicitly?
That way `has_velocity_limits`/`has_acceleration_limits` could be removed altogether. All RobotModel joints would alway come with valid (default) limits and the question of how to time trajectories for joints without acceleration limits does not arise.
But still, users would not have to think about each joint independently without having the chance to get official information from the manufacturer.

> Yes, a warning was printed but it was easy to overlook in the huge terminal output that MoveIt produces.

[rqt_console](https://docs.ros.org/en/humble/Tutorials/Beginner-CLI-Tools/Using-Rqt-Console/Using-Rqt-Console.html)? :-)
And yes, MoveIt prints too much information/spam by default.

AndyZe:
> Thinking along these lines, would it be a viable way around this whole discussion to add explicit defaults for velocity and acceleration to the joint_limits.yaml that apply whenever a joint does not define limits explicitly?

That seems like a good idea. The URDF format doesn't even allow acceleration limits so we'd have to require a `joint_limits.yaml`

sea-bass:
I just ran into issues with this and have my two cents to contribute:

Agreed that a default of 1.0 rad/s doesn't make any sense and I'm in favor of removing that. However, I don't think this should require ALL joint limits to be defined.

Take, for example, a robot with 5 joints where 3 of them have velocity/acceleration limits and 2 don't. Rather than erroring out, should we just bypass those 2 unbounded joints and have the 3 bounded joints set the limits?

I'd be OK with throwing an error if NONE of the active joints in the group had bounds, but so long as there is one valid bounded joint, we should be able to progress.

I'd be happy to work on that update, if we come to some agreement.

PR #1796, opened 2022-12-08
Fix BSD license in package.xml

ct2034:
### TL;DR
This is not a change of license, just the correction in the `package.xml`s

### Description
All the moveit2 packages are licensed under BSD-3-Clause as of [LICENSE.txt](https://github.com/ros-planning/moveit2/blob/main/LICENSE.txt), like most ROS1 packages. But the `package.xml`s mentioned BSD as license. This is wrong, because BSD refers to the original BSD (4-clause) license https://en.wikipedia.org/wiki/BSD_licenses#4-clause_license_(original_%22BSD_License%22). So it is literally another license.

I found this with https://github.com/boschresearch/ros_license_linter/blob/main/field-trials/moveit2.log which I am currently working on.

### DIFF
--- a/moveit/package.xml
+++ b/moveit/package.xml
@@ -9,7 +9,7 @@
   <maintainer email="me@v4hn.de">Michael Görner</maintainer>
   <maintainer email="rhaschke@techfak.uni-bielefeld.de">Robert Haschke</maintainer>
 
-  <license>BSD</license>
+  <license>BSD-3-Clause</license>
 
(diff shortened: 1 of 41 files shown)


AndyZe (approved):
I did a quick check that you didn't miss any `package.xml's`. (Other than a couple scattered package.xml's that are Apache). Looks good to me.

vatanaksoytezer (approved):

PR #1783, opened 2022-12-05
fix: resolve bugs in MoveGroupSequenceAction class

MarcoMagriDev:
### Description

Bugs fixed in MoveGroupSequenceAction class.

### Modifications

1) [Invalid goal_handle transition](https://github.com/MarcoMagriDev/moveit2/commit/312e56e68ca13465463274d19969b410f6fd06aa#diff-786763d71c09b7d98cceb9e9b210f23ae92fea36cee24fa613b2e56b97ae93d2L96)
  `[move_group-1]   what():  goal_handle attempted invalid transition from state EXECUTING with event EXECUTE, at ./src/rcl_action/goal_handle.c:95`

2) [move_feedback_ not initialized](https://github.com/MarcoMagriDev/moveit2/commit/312e56e68ca13465463274d19969b410f6fd06aa#diff-786763d71c09b7d98cceb9e9b210f23ae92fea36cee24fa613b2e56b97ae93d2R61)
  `[move_group-1] Segmentation fault (Address not mapped to object [(nil)])`

3) [plan_callback_ not assigned](https://github.com/MarcoMagriDev/moveit2/commit/312e56e68ca13465463274d19969b410f6fd06aa#diff-786763d71c09b7d98cceb9e9b210f23ae92fea36cee24fa613b2e56b97ae93d2R166)
  `[move_group-1]   what():  bad_function_call`

PR #1797, opened 2022-12-09
fix: resolve bugs in MoveGroupSequenceAction class (main branch)

MarcoMagriDev:
### Description

Bugs fixed in MoveGroupSequenceAction class.

### Modifications

1) [Invalid goal_handle transition](https://github.com/MarcoMagriDev/moveit2/commit/312e56e68ca13465463274d19969b410f6fd06aa#diff-786763d71c09b7d98cceb9e9b210f23ae92fea36cee24fa613b2e56b97ae93d2L96)
  `[move_group-1]   what():  goal_handle attempted invalid transition from state EXECUTING with event EXECUTE, at ./src/rcl_action/goal_handle.c:95`

2) [move_feedback_ not initialized](https://github.com/MarcoMagriDev/moveit2/commit/312e56e68ca13465463274d19969b410f6fd06aa#diff-786763d71c09b7d98cceb9e9b210f23ae92fea36cee24fa613b2e56b97ae93d2R61)
  `[move_group-1] Segmentation fault (Address not mapped to object [(nil)])`

3) [plan_callback_ not assigned](https://github.com/MarcoMagriDev/moveit2/commit/312e56e68ca13465463274d19969b410f6fd06aa#diff-786763d71c09b7d98cceb9e9b210f23ae92fea36cee24fa613b2e56b97ae93d2R166)
  `[move_group-1]   what():  bad_function_call`

Original pull request on humble branch #1783


henningkayser (approved):
Thank you for these fixes! I have not tested them yet, but at least 2 and 3 seem obviously correct to me (@v4hn might be interested in a cherry-pick of 3). I'm a bit puzzled about proper GoalHandle usage though, since I don't really find consistent information about who is calling `execute()`. `ACCEPT_AND_EXECUTE` seems to imply this call but I couldn't find it in rclcpp besides in the unit tests. The examples also don't use it, so I'll go with this change.

MarcoMagriDev:
> Thank you for these fixes! I have not tested them yet, but at least 2 and 3 seem obviously correct to me (@v4hn might be interested in a cherry-pick of 3). I'm a bit puzzled about proper GoalHandle usage though, since I don't really find consistent information about who is calling `execute()`. `ACCEPT_AND_EXECUTE` seems to imply this call but I couldn't find it in rclcpp besides in the unit tests. The examples also don't use it, so I'll go with this change.
> 
> Side-note: you can close #1783, we're backporting this to humble right after merging this.

You are welcome!

Regarding point 1) I think that the answer that you are looking for is [here](https://github.com/ros2/rclcpp/blob/rolling/rclcpp_action/src/server.cpp#L388).
The goal state is updated to EXECUTING if the ACCEPT_AND_EXECUTE is used.
Using ACCEPT_AND_DEFER instead of ACCEPT_AND_EXECUTE [here ](https://github.com/ros-planning/moveit2/blob/main/moveit_planners/pilz_industrial_motion_planner/src/move_group_sequence_action.cpp#L74) the execute has to be manually called as it was [here](https://github.com/ros-planning/moveit2/blob/main/moveit_planners/pilz_industrial_motion_planner/src/move_group_sequence_action.cpp#L96).

PR #1798, opened 2022-12-09
feat: adds ompl compatibility to the move_group_sequence

MarcoMagriDev:
### Description

The current implementation of MoveGroupSequence would allow you to specify the `pipeline_id` with which to plan the motion. Choosing the OMPL pipeline as planner causes the plan to fail. 

This is due to the fact that internally to the MoveGroupSequence performs checks on the single planned path  before joining two path together. The only checks that fails (in case of path planned with the OMPL pipeline) is the one on the stationarity of the last point of each planned trajectory. In particular, the condition regarding the zero acceleration value of the last point of the trajectory is never fulfilled. 

### Proposed solution
The proposed solution to this problem is quite brutal but empirically gives good results and would allow to use MoveGroupSequence with the OMPL planner.

### DIFF
--- a/moveit_planners/pilz_industrial_motion_planner/src/command_list_manager.cpp
+++ b/moveit_planners/pilz_industrial_motion_planner/src/command_list_manager.cpp
@@ -92,6 +92,14 @@ RobotTrajCont CommandListManager::solve(const planning_scene::PlanningSceneConst
 
   MotionResponseCont resp_cont{ solveSequenceItems(planning_scene, planning_pipeline, req_list) };
 
+  if (planning_pipeline->getPlannerPluginName() == "ompl_interface/OMPLPlanner")
+  {
+    for (MotionResponseCont::size_type i = 0; i < resp_cont.size(); ++i)
+    {
+      resp_cont.at(i).trajectory_->getLastWayPointPtr()->zeroAccelerations();
+    }
+  }
+  
   assert(model_);
   RadiiCont radii{ extractBlendRadii(*model_, req_list) };
   checkForOverlappingRadii(resp_cont, radii);


ct2034:
Why is this issue unique to ompl?

sjahr:
I was wondering the same. @MarcoMagriDev do you have a minimal demo to reproduce the the error? Most planner plugins in MoveIt with PILZ beeing an exception create paths instead of trajectories and require running trajectory generation algorithm such as TOTG or Ruckig after a plan is generated. This is usually done in MoveIt with `planning_request_adapter`s.

MarcoMagriDev:
@sjahr @ct2034 I don't know if the issue is related only to OMPL or it occurs for each planner different from PILZ.

@sjahr to reproduce the issue perform the following steps with reference to the [panda_moveit_config](https://github.com/ros-planning/moveit_resources/tree/ros2/panda_moveit_config)

1) add the `pilz_industrial_motion_planner/MoveGroupSequenceAction` capability to the move_group node
2) run the demo `ros2 launch moveit_resources_panda_moveit_config demo.launch.py`
3) run [this script](https://github.com/ros-planning/moveit2/files/10209201/send_move_group_sequence.zip) with different arguments:
a) `python3 send_move_group_sequence.py -t single` : the robot should reach the extended position without any problem. When the motion is over, command a motion to the ready pose using the RViz interface before running the next tests.
b) `python3 send_move_group_sequence.py -t multi` : you will get `[ERROR] [1670856446.681808505] [panda_arm_controller]: Time between points 1 and 2 is not strictly increasing, it is 0.079133 and 0.079133 respectively` and the robot will not move.
c) `python3 send_move_group_sequence.py -t multi` : you will get
`[move_group-4] [ERROR] [1670856733.300848155] [moveit.pilz_industrial_motion_planner.trajectory_blender_transition_window]: During blending the last point of the preceding and the first point of the succeeding trajectory
[move_group-4] [ERROR] [1670856733.300851601] [moveit.pilz_industrial_motion_planner.trajectory_blender_transition_window]: Trajectory blend request is not valid.
[move_group-4] [ERROR] [1670856733.300890539] [moveit.pilz_industrial_motion_planner.move_group_sequence_action]: Planning pipeline threw an exception (error code: 99999): Blending failed`

In both cases b) and c) the error is caused by the fact that the accelerations of the last point of each sub-trajectory coming from OMPL are not zero. By uncommenting lines 22 and 23 of the script you can also test what happen with the same targets with the PILZ pipeline.

henningkayser (commented):
Are the acceleration values of the last waypoint approximately 0 though? If anything, the fault should be in [TOTG](https://github.com/ros-planning/moveit_resources/blob/ros2/panda_moveit_config/config/ompl_planning.yaml#L3), that's where the time profile is coming from. A fix could be to explicityl zero out velocity and acceleration in the last waypoint there. I don't see why not since non-zero velocity is not supported anyway. For the general case, we can also add an `HaltTrajectory` planner adapter that ensures a smooth transition from the last waypoint to full stop.

MarcoMagriDev:
> Are the acceleration values of the last waypoint approximately 0 though? If anything, the fault should be in [TOTG](https://github.com/ros-planning/moveit_resources/blob/ros2/panda_moveit_config/config/ompl_planning.yaml#L3), that's where the time profile is coming from. A fix could be to explicityl zero out velocity and acceleration in the last waypoint there. I don't see why not since non-zero velocity is not supported anyway. For the general case, we can also add an `HaltTrajectory` planner adapter that ensures a smooth transition from the last waypoint to full stop.

By adding the following line waypoints can be visualized.
`RCLCPP_ERROR_STREAM(LOGGER, "" << *(resp_cont.at(i).trajectory_));` 
 
Looking at acceleration values (for a simple motion that involves only one joint) it is possible to notice that the produced path has a trapezoidal velocity profile, thus acceleration are piece wise linear in time. 
In particular, given a trajectory with `n` waypoints:

1. the acceleration of the first point is 0
2. from point 1 to point `m` has constant values
3. from point `m` to point `n-m` has value 0
4. from point `n-m` to point `n` (included, here is the problem) has the same constant value as in point 2 but with opposite sign.

So actually the acceleration values of the last waypoint are **not** approximately 0 (they have the same value as all waypoints from `n-m` to `n`)

ct2034:
If the velocity profile is trapezoidal, then I think it would be a better approach to append a point with zero acceleration. This would make it symmetric with the first point, as you described.

ct2034 (changes requested):
I am not happy with the check for ompl. Either this applies to multiple planners and should be checked somehow depending on the trajectory, not the planner to apply to all or it depends purely on ompl, then it should be fixed there.

MarcoMagriDev:
> If the velocity profile is trapezoidal, then I think it would be a better approach to append a point with zero acceleration. This would make it symmetric with the first point, as you described.
> 
> What do you think?

Instead of appending a point I think that it is better to just modify the last one.

> I am not happy with the check for ompl. Either this applies to multiple planners and should be checked somehow depending on the trajectory, not the planner to apply to all or it depends purely on ompl, then it should be fixed there.

Totally agree with you. I am not happy with this check too. 
I really like your idea to perform the check on the trajectory instead of the planner used. 
Two possibile check substitutions  comes to my mind:
- check if the last point has zero velocity and a non zero acceleration (by doing so in practice the stationarity condition of the last point is only checked on velocities)
- performs checks on the whole trajectory accelerations to assess if it is trapezoidal

What do you think?

tylerjw:
I'm not sure what should be done with this PR. It seems there has been no work on it in the last 6mo, so I'm going to close it. Feel free to re-open and ping us for reviews.

PR #1799, opened 2022-12-09
Print a warning from TOTG if the robot model mixes revolute/prismatic joints

AndyZe:
### Description

I believe TOTG can't handle mixed units well because the `path_tolerance_` parameter is just a scalar value. So there's no good way to measure angular error mixed with linear error. Best just to fail and print a nice warning for the user.

### DIFF
--- a/moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
+++ b/moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
@@ -207,6 +207,9 @@ private:
                                           const Eigen::VectorXd& max_velocity,
                                           const Eigen::VectorXd& max_acceleration) const;
 
+  /// @brief Check if a combination of revolute and prismatic joints is used. path_tolerance_ is not valid, if so.
+  bool checkMixedUnits(const moveit::core::JointModelGroup* group) const;
+
   const double path_tolerance_;
   const double resample_dt_;
   const double min_angle_change_;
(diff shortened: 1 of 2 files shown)


nbbrooks:
Can we add a unit test for this?

AndyZe:
I was looking into that but the JointType is protected, so it will take more time than I have right now.

https://github.com/ros-planning/moveit2/blob/ddd50ce124a4f2c6ca67ad9c5738d4f16e718168/moveit_core/robot_model/include/moveit/robot_model/joint_model.h#L473

henningkayser:
I agree that the path tolerance doesn't really make sense with mixed units. But do we have a good alternative for mixed joints?

AndyZe:
That's a good point, I can't really think of an alternative. I could change this to just print a warning...

moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
```diff
@@ -207,6 +207,9 @@ class TimeOptimalTrajectoryGeneration : public TimeParameterization
                                           const Eigen::VectorXd& max_velocity,
                                           const Eigen::VectorXd& max_acceleration) const;
 
+  /// @brief Check if a combination of revolute and prismatic joints is used. path_tolerance_ is not valid, if so.
+  bool checkMixedUnits(const moveit::core::JointModelGroup* group) const;
```
henningkayser, line 211: The function name and comment does not really specify what to expect from the bool (it's the opposite of what I was expecting). Could you rename this to `hasMixedJointTypes()` and directly return `(have_prismatic && have_revolute)` from the function to streamline the logic? Obviously, this flips the bool from the current implementation, but it would improve redability.

henningkayser (approved):
thanks!

tylerjw (approved):

PR #1801, opened 2022-12-12
Refactor moveit_core package to export a single library

ChrisThrasher:
### Description

moveit_core is currently comprised of a few dozen, small libraries. Some of these libraries are small as a single .cpp/.h pair. Some libraries have circular dependencies with others which mean they're difficult if not impossible to use in isolation. 

This PR consolidates those libraries into a single `moveit::core` target which better expresses the fact that moveit_core is intended to be used as one library, not a collection of them. Along the way I was able to make some satisfying simplifications like how many dozens of `target_link_libraries`, `target_include_directories`, and `install` commands were collapsed into just a small few. There's a certain amount of CMake overhead related to each sub-library which is now gone. This will make it easier for others to understand how moveit_core is built and understand how to modify it as needed as requirements change in the future.

This change also required no modification to downstream packages which continue to use moveit_core exactly as they have before.

I'm submitting this PR with the full, verbose commit history so reviewers can see the incremental steps I took, but I plan on squashing this into a single commit before merging since these individual commits are not particularly valuable.


moveit_core/src/collision_detection/allvalid/collision_env_allvalid.cpp
```diff
@@ -40,7 +40,7 @@
 #include <rclcpp/logging.hpp>
 
 // Logger
-static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_collision_detection.collision_world_allvalid");
+static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_core.collision_world_allvalid");
```
henningkayser, line 43: I agree that the logger names should be revised, but setting them all to `moveit_core` doesn't really make them better. Let's omit these changes for now and first discuss how we want to log MoveIt in general.
My thought is that some context is still needed (here maybe `collision_detection`) and that this context should be reflected by file structure and namespace. Ideally we should define a pattern that we could use everywhere, something like `moveit.<package>.<context>.<optional detail>` -> `moveit.core.collision_detection`. (copied to https://github.com/ros-planning/moveit2/discussions/1614#discussioncomment-4390022)

Curiously, this file is `*_env_allvalid` and the logger says `_world_allvalid`...

moveit_core/src/collision_detection_fcl/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+    set(append_library_dirs "${CMAKE_CURRENT_BINARY_DIR};${CMAKE_CURRENT_BINARY_DIR}/../collision_detection;${CMAKE_CURRENT_BINARY_DIR}/../robot_state;${CMAKE_CURRENT_BINARY_DIR}/../robot_model;${CMAKE_CURRENT_BINARY_DIR}/../utils")
+  endif()
+
+  # TODO: remove if transition to gtest's new API TYPED_TEST_SUITE_P is finished
+  if(CMAKE_CXX_COMPILER_ID MATCHES "(GNU|Clang)")
+    add_compile_options(-Wno-deprecated-declarations)
+  endif()
+
+  ament_add_gtest(test_fcl_collision_env test/test_fcl_env.cpp)
+  target_link_libraries(test_fcl_collision_env moveit::core)
```
henningkayser, line 20: Doesn't this change the behavior of the test? The old cmake didn't link against the plugin code so it implicitly tested for valid plugin registration and loading. This version might omit this step, no?

henningkayser (commented):
Great simplification, I really appreciate this effort! Some notes from my end:

* Should we maybe keep separate plugin libraries? I think it would make sense to eventually separate out plugin implementations along with their dependencies (e.g. Bullet/FCL).  That would cleanly delineate what core really does (feature integration+abstraction), and the feature implementation itself (collision checking, smoothing, etc). This is already the case for IK and planners.
* Let's split out the logger name changes (see comment)
* Could you add an entry to the [migration notes](https://github.com/ros-planning/moveit2/blob/main/MIGRATION.md) for those depending on modules?
* There is still a `moveit_core/dummy.cpp`

@lilustga would you have some time to review and test the export header for Windows?

moveit_core/CMakeLists.txt
```diff
@@ -57,46 +56,49 @@ find_package(visualization_msgs REQUIRED)
...
-add_subdirectory(trajectory_processing)
-add_subdirectory(transforms)
-add_subdirectory(utils)
-add_subdirectory(version)
+include(ConfigExtras.cmake) # Finds Boost Components
+include(GenerateExportHeader)
+
+add_library(moveit_core SHARED)
+add_library(moveit::core ALIAS moveit_core)
+generate_export_header(moveit_core EXPORT_FILE_NAME include/moveit/export.h)
```
henningkayser, line 64: should we check for `GenerateExportHeader` or make it `REQUIRED`?

tylerjw: It is included on line 60, and I believe it is part of CMake itself so you don't have to find_package it. @ChrisThrasher is that how it works?

moveit_core/src/collision_detection/allvalid/collision_env_allvalid.cpp
tylerjw: I would suggest we change the logger names to just:

```
<package_name>.<source_file_name>
```

There are two uses of these identifiers I know of:

1. Enabling DEBUG level on a small amount of code to get the debug output.
2. Searching through the code for an ERROR log to figure out where it came from.

I believe the format above would be effective for both of these use cases.

henningkayser: third use case is setting up logger filters for debugging of packages/contexts/detail, and with detail I mean plugin or class instance. Actually, adding file would be redundant because that's already reported by the logger.

moveit_core/CMakeLists.txt
ChrisThrasher: It's not a package, it's a 1st party include module meaning it's always available and never have to `find_package` it.

moveit_core/src/collision_detection_fcl/CMakeLists.txt
ChrisThrasher: That's a good point. I think it's reasonable to keep the plugins as separate targets. I'll have to think about that more.

moveit_core/src/collision_detection/allvalid/collision_env_allvalid.cpp
ChrisThrasher: No problem, I can revert these logger name changes. I was even 100% they were worthwhile changes to begin with but I was just doing pattern matching.

moveit_core/src/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+add_subdirectory(collision_detection_bullet)
```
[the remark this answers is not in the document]
ChrisThrasher: No it's not strictly necessary. My thought was that it was nice to bury some of this code to avoid having to relocate and change the path to many dozens of source files. I could be convinced to move all those `target_sources` call up the main CMakeLists.txt thus eliding them.

My bigger holdup was whether it was worth reorganizing all the tests. Currently I'm keeping them in place but I'm on the fence about whether they should get moved out into their own `tests` directory alongside `include` and `src`. If we do this, then we can delete all these extra CMakeLists.txt in src/. That would be more consistent with how else I've reorganized things but I'd like someone else's input.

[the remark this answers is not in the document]
henningkayser: I was more thinking about using `add_subdirectory(src/...)`. The separate target sources make sense to me, and would even make more sense if we could group all of these targets into fewer but more meaningful namespaces/contexts. Until then, I think it's ok to leave tests where they are for now.

moveit_core/src/collision_detection_fcl/CMakeLists.txt
v4hn: It is best practice to keep plugin lib instantiations in a separate standalone library and I would strongly propose you keep it that way. We had a similar case [in MTC last year](https://github.com/ros-planning/moveit_task_constructor/pull/271#issuecomment-850671203) (that I brought up because of the infamous memory leaks) and ended up splitting the plugins off again.

Personally my main argument is that unless strictly necessary, libraries should not execute code on so-load, unless strictly needed for *every use of the library*. But that's exactly what the PLUGINLIB/CLASSLOADER macros do and not every use of `moveit::core` requires plugins. You have no idea how often boost libraries broke my neck as a linux distribution maintainer because of some random unnecessary initialization it does on load.

sjahr:
@ChrisThrasher Can you rebase this PR so we can merge it?

ChrisThrasher:
> @ChrisThrasher Can you rebase this PR so we can consider merging it?

I wish I could. At a minimum, we need to find a home for plugins in `moveit_core` that need to be removed before we can start consolidating targets and get that merged. After that I plan on rebasing as much of this PR as I can then reimplementing the rest as needed.

If anyone still has lingering issues with this PR, please share them now before I put any more work into reimplementing this transition for the 2nd (and hopefully last) time.

moveit_core/src/CMakeLists.txt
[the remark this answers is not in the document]
v4hn: > I was more thinking about using add_subdirectory(src/...).

That sounds slightly better to me as well.

> whether they should get moved out into their own tests directory alongside include and src

I do think it makes much more sense to keep the tests in the individual source folders.
moveit_core might become one library now, conceptually, but it's still pretty big and the separate folders help to show what component has how many tests.

moveit_core/CMakeLists.txt
```diff
@@ -57,46 +56,49 @@ find_package(visualization_msgs REQUIRED)
...
+target_include_directories(moveit_core PUBLIC
+  $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
```
v4hn, line 66: I'm not convinced we gain anything from splitting sources and headers for the different components. I'll not block the decision though if you think it more convenient.

The decision to split the headers of is fully independent of whether or not there should be one `libmoveit_core.so` and it's definitely not considered separately in this discussion so far.
Yes, it consolidates paths a bit. But it also adds three subdirectory changes to go from a component's sources to the respective headers...

moveit_core/src/collision_detection_fcl/CMakeLists.txt
v4hn: Viable options I see to move the plugins out:
- new moveit_plugins subfolders (actually new packages) for all
- a separate `moveit::core_plugins` library for the three of them
- add `moveit::collision_plugins` and split the butterworth filter out into moveit_plugins. fcl/bullet are definitely close to moveit::core, but the signal smoother is only used in moveit_servo. On the other hand moveit_core does not use the plugins but directly relies on the classes in most cases.

I slightly prefer the last option.

ChrisThrasher:
I spent yesterday trying to split out the three plugins in moveit_core and it was awful.

* collision_detector_bullet
* collision_detector_fcl
* filter_plugin_butterworth

I didn't even get to the last one. I started with the two collision_detector plugins which I tried to move to a new package within the moveit_plugins meta package. The major issue is that there are parts of moveit_core than depend heavily on collision_detector_fcl so extract that whole directory is very messy. I tried to move all the relevant code out of that directory back into the collision_detection package but I think I ended up created a cyclic dependency in the `#include` graph forcing me to forward declare many types. I can try to keep pushing on that until it finally builds but the solution will be very gross and perhaps not good enough to merge.

Does anyone have other ideas for extract collision_detector_fcl when parts of moveit_core are already depending on it?

PR #1805, opened 2022-12-13
Add xacro subsititution class and use it for loading urdf & srdf

JafarAbdi:
### Description

This make it possible to pass `LaunchConfiguration` (or any other complex launch substitutable) to the xacro mappings, which remove the need to use an `OpaqueFunction` to access the context (see https://github.com/AndyZe/tiago_moveit_config/commit/054900ae668e0bcf11e831da5affd4fb061c82e4 for an example)

Another thing worth doing is making all the `file_path`s accept `SomeSubstitutionsType`


JafarAbdi:
The last commit makes it possible to use with the new python binding and complex cases where a launch substitution is required

tylerjw (approved):

PR #1806, opened 2022-12-14
MoveItConfigBuilder robot_description from topic

MarcoMagriDev:
It seems that the current implementation of the MoveItConfigBuilder forces the user to place the robot urdf under the `robot_name_moveit_config/config` folder (if the file is not found the launch fails).
In my opinion, it should not be mandatory since the move_group node is able to load the robot description from a topic by means of [this](https://github.com/ros-planning/moveit2/blob/main/moveit_ros/planning/rdf_loader/include/moveit/rdf_loader/synchronized_string_parameter.h) class (correct me if I am wrong).
With the proposed modification a warn is print to inform the user that the robot description will be loaded from the corresponding topic but the launch can proceed without errors.


henningkayser (approved):
You are correct that the robot description should not be mandatory and I agree with this change. I still see that there is an argument that a launch file should fail and not fall back to other behavior if an explicitly defined config file is not found. I would really like @JafarAbdi's opinion on this.

JafarAbdi (approved):
I always wanted to make it optional, thanks for the fix!

I'll clean it up the logic in a separate.

PR #1807, opened 2022-12-14
MoveItConfigBuilder support for multiple instaces

MarcoMagriDev:
Creating two instances of `MoveItConfigBuilder` seems to be unsupported due to conflicting `MoveItConfigs`.
This is an example launcher to reproduce the issue [test_builder.zip](https://github.com/ros-planning/moveit2/files/10230041/test_builder.zip). 
As you can notice by running the script (with the current version of the `MoveItConfigBuilder`) the attributes of two different `MoveItConfigs` (produced by two different `MoveItConfigBuilder`) share the same id and values (the one of the fanuc in the provided example).

The proposed solution simply consists in moving the instatiation of the `MoveItConfigs` inside the `__init__() `method of `MoveItConfigBuilder`. By doing so the problem seems to be solved (running again the script after applying the modification result in the expected behavior).


JafarAbdi (approved):
Thanks!

henningkayser (approved):

PR #1812, opened 2022-12-15
Add a "lookahead time" parameter

AndyZe:
### Description

This is the second time I've made a PR like this (prev. #886). It looks like Ignition testing requires it again.

The idea is, increase the timestep when calculating the target pose, to move the target pose farther away. Here's a hypothetical example where it helps:

> If there is a net latency of 30 ms (from when the Servo message is published to when the low-level controller executes), it can reduce the path-tracking error quite a bit. Thus the robot's low-level PID output is small --> the robot moves slowly. In the perfect world where the latency is 0, the path-tracking accuracy will be much larger and the robot would move with the expected speed.


AndyZe:
This didn't help with the Ignition simulation. Must be a different issue.

PR #1813, opened 2022-12-16
Bug fix: RobotTrajectory append()

AndyZe:
### Description

The first commit seems to show something unexpected - an extra timestep of 0.1s is appended at index 5. The second commit fixes it.

```
/home/andy/ws_ros2/src/moveit2/moveit_core/robot_trajectory/test/test_robot_trajectory.cpp:217: Failure
Expected equality of these values:
  initial_trajectory->getWayPointCount()
    Which is: 9
  10
/home/andy/ws_ros2/src/moveit2/moveit_core/robot_trajectory/test/test_robot_trajectory.cpp:219: Failure
Expected equality of these values:
  initial_trajectory->getWayPointDurationFromPrevious(5)
    Which is: 0.2
  EXPECTED_DURATION
    Which is: 0.1
```

### DIFF
--- a/moveit_core/robot_trajectory/test/test_robot_trajectory.cpp
+++ b/moveit_core/robot_trajectory/test/test_robot_trajectory.cpp
@@ -204,20 +204,21 @@ TEST_F(RobotTrajectoryTestFixture, Append)
 {
   robot_trajectory::RobotTrajectoryPtr initial_trajectory;
   initTestTrajectory(initial_trajectory);
+  EXPECT_EQ(initial_trajectory->getWayPointCount(), size_t(5));
 
   // Append to the first
   robot_trajectory::RobotTrajectoryPtr traj2;
   initTestTrajectory(traj2);
+  EXPECT_EQ(traj2->getWayPointCount(), size_t(5));
 
   // After append() we should have 10 waypoints, all with 0.1s duration
-  const double EXPECTED_DURATION = 0.1;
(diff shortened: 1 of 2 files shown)


moveit_core/robot_trajectory/test/test_robot_trajectory.cpp
```diff
@@ -200,6 +200,27 @@ TEST_F(RobotTrajectoryTestFixture, ChainEdits)
...
+  EXPECT_EQ(initial_trajectory->getWayPointCount(), size_t(5));
+
+  // Append to the first
+  robot_trajectory::RobotTrajectoryPtr traj2;
+  initTestTrajectory(traj2);
+  EXPECT_EQ(traj2->getWayPointCount(), size_t(5));
+
+  // After append() we should have 10 waypoints, all with 0.1s duration
+  const double EXPECTED_DURATION = 0.1;
+  initial_trajectory->append(*traj2, 0.1, 0, 5);
```
AndyZe, line 216: The **5** here doesn't make sense to me. Since the trajectory has 5 waypoints, I think the last index should be 4. But the test passes. I'm confused.

henningkayser (approved):
Good find! I wonder if anyone would disagree with this, but the fixed behavior looks correct to me. I was still somewhat under the impression that `duration_from_previous_` would start indexing at the second waypoint since the first waypoint doesn't have a "previous" :thinking:.

Issue #1800, opened 2022-12-09
Kinematics solver of type 'kdl_kinematics_plugin/KDLKinematicsPlugin' could not be initialized for group X


sjahr:
### Description

When using starting a node that is using moveit_cpp I see the following error:
```
[moveit_combine_predefined_poses_benchmark-1] [ERROR] [1670626148.094761003] [moveit_kinematics_base.kinematics_base]: Group 'hand' is not a chain
[moveit_combine_predefined_poses_benchmark-1] [ERROR] [1670626148.094765752] [kinematics_plugin_loader]: Kinematics solver of type 'kdl_kinematics_plugin/KDLKinematicsPlugin' could not be initialized for group 'hand'
[moveit_combine_predefined_poses_benchmark-1] [ERROR] [1670626148.094780698] [moveit_kinematics_base.kinematics_base]: Group 'hand' is not a chain
[moveit_combine_predefined_poses_benchmark-1] [ERROR] [1670626148.094785028] [kinematics_plugin_loader]: Kinematics solver of type 'kdl_kinematics_plugin/KDLKinematicsPlugin' could not be initialized for group 'hand'
[moveit_combine_predefined_poses_benchmark-1] [ERROR] [1670626148.094798018] [moveit_kinematics_base.kinematics_base]: Group 'hand' is not a chain
[moveit_combine_predefined_poses_benchmark-1] [ERROR] [1670626148.094803257] [kinematics_plugin_loader]: Kinematics solver of type 'kdl_kinematics_plugin/KDLKinematicsPlugin' could not be initialized for group 'hand'
[moveit_combine_predefined_poses_benchmark-1] [ERROR] [1670626148.094808425] [moveit_ros.robot_model_loader]: Kinematics solver could not be instantiated for joint group hand.
[moveit_combine_predefined_poses_benchmark-1] [ERROR] [1670626148.094827561] [moveit_kinematics_base.kinematics_base]: Group 'panda_arm_hand' is not a chain
[moveit_combine_predefined_poses_benchmark-1] [ERROR] [1670626148.094832311] [kinematics_plugin_loader]: Kinematics solver of type 'kdl_kinematics_plugin/KDLKinematicsPlugin' could not be initialized for group 'panda_arm_hand'
[moveit_combine_predefined_poses_benchmark-1] [ERROR] [1670626148.094846069] [moveit_kinematics_base.kinematics_base]: Group 'panda_arm_hand' is not a chain
[...]

Issue #1633, opened 2022-10-21
Group X is not a chain


swiz23:
### Description
There seems to be a few unexpected warning messages that appear in the terminal if there is a planning group defined in an SRDF file which does not have an associated kinematics plugin. Reasons to do this might be to have a group of links that you want to disable collisions for in a given application. This did not used to happen, and it's unclear why it's happening now.

### Your environment
* ROS Distro: Humble
* OS Version: e.g. Ubuntu 22.04
* Source or Binary build? Source
* If source, which branch? main

### Steps to reproduce
Add the following text to the SRDF in the MoveIt Fanuc Config package [here](https://github.com/ros-planning/moveit_resources/blob/ros2/fanuc_moveit_config/config/fanuc.srdf). This will create an 'eoat' group with 'tool0' as the single link.
```
    <group name="eoat">
        <link name="tool0"/>
    </group>
```

Build and Launch the package by typing `ros2 launch moveit_resources_fanuc_moveit_config demo.launch.py `

### Expected behaviour
The warnings shown below should not appear.

### Actual behaviour
These warning messages appear in the terminal. It doesn't affect usage, but it's unclear why this is happening now when it never happened before.
```
[move_group-4] [ERROR] [1666380869.436552686] [moveit_kinematics_base.kinematics_base]: Group 'eoat' is not a chain
[move_group-4] [ERROR] [1666380869.436566050] [kinematics_plugin_loader]: Kinematics solver of type 'kdl_kinematics_plugin/KDLKinematicsPlugin' could not be initialized for group 'eoat'
[move_group-4] [ERROR] [1666380869.436577384] [moveit_kinematics_base.kinematics_base]: Group 'eoat' is not a chain
[move_group-4] [ERROR] [1666380869.436580639] [kinematics_plugin_loader]: Kinematics solver of type 'kdl_kinematics_plugin/KDLKinematicsPlugin' could not be initialized for group 'eoat'
[move_group-4] [ERROR] [1666380869.436584469] [moveit_ros.robot_model_loader]: Kinematics solver could not be instantiated for joint group eoat.
```

@JafarAbdi - any suggestions?

swiz23:
Thanks @sjahr, I'll try the latest changes and let you know how it goes.

swiz23:
[...]

PR #1816, opened 2022-12-19
No default IK solver

tylerjw:
### Description

Before converting to generate_parameter_library there was no default IK solver. This meant that if you did not specify an IK solver for a given joint group it wouldn't create one. This restores that behavior.

### DIFF
--- a/moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
+++ b/moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
@@ -259,6 +259,13 @@ moveit::core::SolverAllocatorFn KinematicsPluginLoader::getLoaderFunction(const 
 
         std::string kinematics_solver_param_name = kinematics_param_prefix + ".kinematics_solver";
         const auto kinematics_solver = group_params_.at(known_group.name_).kinematics_solver;
+
+        if (kinematics_solver.empty())
+        {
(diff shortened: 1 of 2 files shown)


Abishalini (approved):
Looks good to me

henningkayser (approved):

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -259,6 +259,13 @@ moveit::core::SolverAllocatorFn KinematicsPluginLoader::getLoaderFunction(const
 
         std::string kinematics_solver_param_name = kinematics_param_prefix + ".kinematics_solver";
         const auto kinematics_solver = group_params_.at(known_group.name_).kinematics_solver;
+
+        if (kinematics_solver.empty())
+        {
+          RCLCPP_DEBUG(LOGGER, "No kinematics solver specified for group '%s'.", known_group.name_.c_str());
```
sjahr, line 265: What do you think about raising this log-level to 'WARN'?

sjahr (approved):
This probably fixes #1800 and #1633

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
henningkayser: I think it's fine this way, otherwise it would show up for grippers etc. A warning will be shown if you attempt to run IK on groups without an IK configuration.

Issue #1826, opened 2022-12-22
Servo build fails in build farm (Rolling)


henningkayser:
Servo fails to compile the Rolling binaries (build farm) due to a missing `ros2_control` header https://build.ros2.org/job/Rbin_uJ64__moveit_servo__ubuntu_jammy_amd64__binary/89/console

```
08:22:20 /tmp/binarydeb/ros-rolling-moveit-servo-2.6.0/src/servo_calcs.cpp:45:10: fatal error: controller_manager/realtime.hpp: No such file or directory
08:22:21    45 | #include <controller_manager/realtime.hpp>
08:22:21       |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
08:22:21 compilation terminated.
08:22:21 make[4]: *** [CMakeFiles/moveit_servo_lib.dir/build.make:121: CMakeFiles/moveit_servo_lib.dir/src/servo_calcs.cpp.o] Error 1
08:22:21 make[4]: Leaving directory '/tmp/binarydeb/ros-rolling-moveit-servo-2.6.0/.obj-x86_64-linux-gnu'
08:22:21 make[3]: *** [CMakeFiles/Makefile2:206: CMakeFiles/moveit_servo_lib.dir/all] Error 2
08:22:21 make[3]: Leaving directory '/tmp/binarydeb/ros-rolling-moveit-servo-2.6.0/.obj-x86_64-linux-gnu'
08:22:21 make[2]: *** [Makefile:149: all] Error 2
08:22:21 make[2]: Leaving directory '/tmp/binarydeb/ros-rolling-moveit-servo-2.6.0/.obj-x86_64-linux-gnu'
08:22:21 dh_auto_build: error: cd .obj-x86_64-linux-gnu && make -j1 VERBOSE=1 returned exit code 2
08:22:21 make[1]: *** [debian/rules:44: override_dh_auto_build] Error 25
08:22:21 make[1]: Leaving directory '/tmp/binarydeb/ros-rolling-moveit-servo-2.6.0'
08:22:21 make: *** [debian/rules:27: build] Error 2
08:22:21 dpkg-buildpackage: error: debian/rules build subprocess returned exit status 2
08:22:21 E: Building failed
```

The header has been removed with https://github.com/ros-controls/ros2_control/pull/794, and then the breaking change has been released into Rolling.

[...]

PR #1827, opened 2022-12-22
Remove third-party workspace dependencies

henningkayser:
This removes third-party packages that should be build against debians. Building with debian dependencies only should succeed by default, so that we can actually detect upstream breakages. An exception can be made for things that we would release at the same time (say, moveit_msgs, moveit_resources), everything else should be handled either in a backwards-compatible way, or by preparing and fixing breaking changes when they actually occur (this is the case if the testing build fails in CI). If we are paranoid that we won't have enough time, I'm open for adding a staging CI build that includes the latest head of third-party repos.
 
See #1826 for additional reasoning.

(this PR will probably fail, even though it shouldn't)

### DIFF
--- a/moveit2.repos
+++ b/moveit2.repos
@@ -3,15 +3,3 @@ repositories:
     type: git
     url: https://github.com/ros-planning/moveit_resources.git
     version: ros2
-  ros2_control:
-    type: git
-    url: https://github.com/ros-controls/ros2_control.git
-    version: 2.13.0
-  generate_parameter_library:
-    type: git
-    url: https://github.com/PickNikRobotics/generate_parameter_library.git
-    version: 0.3.0
-  rsl:
-    type: git
-    url: https://github.com/PickNikRobotics/RSL.git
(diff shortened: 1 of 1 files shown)


moveit2.repos
```diff
@@ -3,15 +3,3 @@ repositories:
...
     url: https://github.com/ros-planning/moveit_resources.git
     version: ros2
-  ros2_control:
-    type: git
-    url: https://github.com/ros-controls/ros2_control.git
-    version: 2.13.0
-  generate_parameter_library:
-    type: git
-    url: https://github.com/PickNikRobotics/generate_parameter_library.git
-    version: 0.3.0
```
henningkayser, line 13: This is the currently released version, but CI is not happy with the debian (https://github.com/ros-planning/moveit2/actions/runs/3758355166/jobs/6386597488#step:12:574). @tylerjw any ideas?

rhaschke:
Resolved by #1791

PR #1828, opened 2022-12-22
keep printf color change on the same line

zflat:
### Description

I found that the `printf` statements were not actually resetting colors in the terminal. So this is a change to keep `printf` color change on the same line. This ensures the color reset is applied because printing the color reset after new lines seems to prevent the color from actually being reset.

Example before making this change:
Screenshot from 2022-12-22 10-26-18-before

Example after this change:
Screenshot from 2022-12-22 10-23-58-after

### Checklist

### DIFF
--- a/moveit_ros/move_group/src/move_group.cpp
+++ b/moveit_ros/move_group/src/move_group.cpp
@@ -99,12 +99,12 @@ public:
       {
         if (capabilities_.empty())
         {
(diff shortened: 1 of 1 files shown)


vatanaksoytezer (approved):
Looks good to me, thank you for fixing this @zflat!

tylerjw (approved):

PR #1834, opened 2022-12-28
moveit_cpp: handle the case where blocking==false

AndyZe:
### Description

Currently, the `blocking` arg of `MoveItCpp::execute()` must be true. The case where it's false wasn't handled well.


sjahr (commented):
Wouldn't it be better to remove `blocking` completely if it never worked well? Or are there any plans/ PRs to fix or make this work in the future?

sjahr:
@AndyZe Can you remove `blocking`?

AndyZe:
> @AndyZe Can you remove `blocking`?

@cambel [responded](https://github.com/ros-planning/moveit/pull/3243#discussion_r1088478827) that he does use `blocking`. So I guess this PR should stay as it is.

sjahr (approved):

PR #1835, opened 2022-12-28
Do not allow traj execution from PlanningComponent

AndyZe:
Merge with https://github.com/ros-planning/moveit2_tutorials/pull/563

MoveItCpp and PlanningComponent have very similar `execute()` functions. ~~One does not call the other.~~ Here, I advocate for deleting the PlanningComponent version because I don't think something focused on planning should handle execution. If you like this, there are some small updates to make in moveit_tutorials but that is the only effect I foresee.

```
bool PlanningComponent::execute(bool blocking)

MoveItCpp::execute(const std::string& group_name, const robot_trajectory::RobotTrajectoryPtr& robot_trajectory,
                   bool blocking)
```

If you do NOT like this PR, then can we flip it to delete the MoveItCpp version of `execute()`? AFAIK it is not used anywhere

### DIFF
--- a/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
+++ b/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
@@ -163,9 +163,6 @@ public:
   trajectory_execution_manager::TrajectoryExecutionManagerPtr getTrajectoryExecutionManagerNonConst();
 
   /** \brief Execute a trajectory on the planning group specified by group_name using the trajectory execution manager.
-   * If blocking is set to false, the execution is run in background and the function returns immediately. */
-
-  /** \brief Execute a trajectory on the planning group specified by group_name using the trajectory execution manager.
    *  \param [in] group_name MoveIt group to execute for.
    *  \param [in] robot_trajectory Contains trajectory info as well as metadata such as a RobotModel.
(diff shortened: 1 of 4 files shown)


henningkayser:
But, one does call the other? The intention was to make it easier to forward the last plan solution for execution, similar to how MGI is doing it. I agree though, that we should keep our API clean and that removing it would improve separation of concerns.

sjahr (approved):
I like this change too :+1: 
The API diff looks something like this, I guess:

```diff
-planning_component.plan();
+auto motion_plan_response = planning_component.plan();
-planning_component.execute();
+planning_component(group_name, motion_plan_reponse.trajectory)
```
I think the new version is more readable.

Wouldn't it be possible to drop the group name argument for MoveItCpp::execute(...) as well, since the robot_trajectory::RobotTrajectoryPtr in MotionPlanResponse contains a group name?

henningkayser:
Can you first deprecate this function before removing it without notice?

AndyZe:
> Wouldn't it be possible to drop the group name argument for MoveItCpp::execute(...) as well, since the robot_trajectory::RobotTrajectoryPtr in MotionPlanResponse contains a group name?

Ooh, I like that idea  :+1:

AndyZe:
I'll make a `moveit2_tutorials` PR that should be merged simultaneously with this one.

sjahr:
@henningkayser @AndyZe Are we ready to merge this?

AndyZe:
Yeah, it's ready as far as I know.

PR #1837, opened 2022-12-29
Delete unused variable in TEM

AndyZe:


AndyZe:
Actually, this interferes with stopping. I'm not sure why. Closing it.

PR #1838, opened 2022-12-30
Add optional list of controllers to MoveItCpp::execute()

AndyZe:
### Description

One of our projects has had some difficulty in switching between a joint trajectory controller and an admittance controller. This adds a parameter so the user can specify which controller they want to execute the trajectory.

### DIFF
--- a/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
+++ b/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
@@ -164,9 +164,18 @@ public:
 
   /** \brief Execute a trajectory on the planning group specified by group_name using the trajectory execution manager.
    * If blocking is set to false, the execution is run in background and the function returns immediately. */
-  moveit_controller_manager::ExecutionStatus execute(const std::string& group_name,
-                                                     const robot_trajectory::RobotTrajectoryPtr& robot_trajectory,
-                                                     bool blocking = true);
+
+  /** \brief Execute a trajectory on the planning group specified by group_name using the trajectory execution manager.
+   *  \param [in] group_name MoveIt group to execute for.
+   *  \param [in] robot_trajectory Contains trajectory info as well as metadata such as a RobotModel.
+   *  \param [in] blocking If blocking, wait for the trajectory to execute before continuing. Defaults to `true`.
+   *  \param [in] controllers An optional list of ros2_controllers to execute with. If none, MoveIt will attempt to find
+   * a controller. The exact behavior of finding a controller depends on which MoveItControllerManager plugin is active.
+   * \return moveit_controller_manager::ExecutionStatus::SUCCEEDED if successful
+   */
+  moveit_controller_manager::ExecutionStatus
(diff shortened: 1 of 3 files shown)


moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
```diff
@@ -164,9 +164,18 @@ class MoveItCpp
...
+   *  \param [in] group_name MoveIt group to execute for.
+   *  \param [in] robot_trajectory Contains trajectory info as well as metadata such as a RobotModel.
+   *  \param [in] blocking If blocking, wait for the trajectory to execute before continuing. Defaults to `true`.
+   *  \param [in] controllers An optional list of ros2_controllers to execute with. If none, MoveIt will attempt to find
+   * a controller. The exact behavior of finding a controller depends on which MoveItControllerManager plugin is active.
+   * \return moveit_controller_manager::ExecutionStatus::SUCCEEDED if successful
+   */
+  moveit_controller_manager::ExecutionStatus
+  execute(const std::string& group_name, const robot_trajectory::RobotTrajectoryPtr& robot_trajectory,
+          bool blocking = true, const std::vector<std::string>& controllers = std::vector<std::string>(1, ""));
```
rhaschke, line 178: I think that's the wrong default. Usually, you should pass an empty vector. At least, that's the default for an empty controller string:
https://github.com/ros-planning/moveit2/blob/b60ee8eb21af6821cee805e923c23ba89b081730/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp#L321-L324

If you insist on a vector containing the empty string, use brace initializer:
```suggestion
          bool blocking = true, const std::vector<std::string>& controllers = {""});
```

rhaschke (commented):
Was this tested? Passing `{""}` as controller list should get rejected:
https://github.com/ros-planning/moveit2/blob/b60ee8eb21af6821cee805e923c23ba89b081730/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp#L1011-L1015

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
AndyZe: @rhaschke I did test it but I think your suggestion is better. I'll test it too and probably make the change.

@tylerjw the list of controllers simply gets passed to the tem::execute()  https://github.com/ros-planning/moveit2/blob/9e3f6b3eb578b55722ea84d87a543a2042c6017b/moveit_ros/planning/trajectory_execution_manager/include/moveit/trajectory_execution_manager/trajectory_execution_manager.h#L154

`bool push(const moveit_msgs::msg::RobotTrajectory& trajectory, const std::vector<std::string>& controllers);`

So I think what you suggest would be API-breaking. Probably should create a new PR for that if you feel strongly.

henningkayser: why can't we simply use an empty list here?

PR #1839, opened 2023-01-02
Remove clipped cells from pointcloud_octomap_updater plugin

ummy11:
### Description

This PR is removes the clipped cells from the octomap.

Previously, undesired voxels would stay and not update due to the clipped cells not being removed. This is especially the case in when the camera observes a dynamic environment. A wall would form at max range from the sensor origin. This can be seen in the following images that show the scene before and after dynamic activity.

scene_after_dynamic_activity
static_scene

The clipped cells where therefore erased from occupied cells and the clipped_cells nodes were updated to unoccupied.
When tested, this solved the issue and the octomap updates properly.

### Checklist

### DIFF
--- a/moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
+++ b/moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
@@ -334,7 +334,7 @@ void PointCloudOctomapUpdater::cloudMsgCallback(const sensor_msgs::msg::PointClo
   for (const octomap::OcTreeKey& occupied_cell : occupied_cells)
     free_cells.erase(occupied_cell);
 
-  /* clipped cells are not free*/
+  /* clipped cells are not occupied*/
   for (const octomap::OcTreeKey& clip_cell : clip_cells){ 
     occupied_cells.erase(clip_cell);}
 


henningkayser (commented):
Thank you for your fix! I'm not really following, why a wall would form at all. Is it from objects moving out of the max_range volume? Have you reviewed https://github.com/ros-planning/moveit/issues/2709? The clipped cells are not treated as occupied by themselves. @v4hn might have some thoughts on this.

ummy11:
> Thank you for your fix! I'm not really following, why a wall would form at all. Is it from objects moving out of the max_range volume? Have you reviewed [ros-planning/moveit#2709](https://github.com/ros-planning/moveit/issues/2709)? The clipped cells are not treated as occupied by themselves. @v4hn might have some thoughts on this.

Hi there @henningkayser . Perhaps I could explain it a bit better.

From my understanding, the clipped cells are those points in the cloud that are beyond the max_range parameter and are clipped by modifying the magnitude of the vector between the sensor and the said points beyond max_range.

I see the value in this since it can be used to ray cast and compute the free cells that ends at the found clipped cell. 

The clipped cells have been inserted into the free_cells keyset as they should. However, they have not been erased from the occupied_cells keyset  in case of overlap. The clip_cells nodes have also not been updated to mark those nodes as unoccupied. 

This PR implements those in order to ensure that the clipped cells are marked as free_cells in the octree.

This “wall” that I mention forming at the max_range occurs when I physically move the camera around to a different scene to kind of simulate an extreme of a dynamic scene change.  (Note: the camera frame doesn’t change and is assumed to be in the same location in Rviz).

This “wall” clears out when the depth camera is returned to the original scene using the changes in this PR.

I hope that makes it clearer.

AndyZe:
Thanks for the explanation  :+1:

@MarqRazz or @Abishalini have you seen walls forming at max range? (I think you guys use RealSense a lot, right?)

sjahr:
@ummy11 Can you fix the CI failures? You can format the code by running [pre-commit run -a](https://moveit.ros.org/documentation/contributing/code/#pre-commit-formatting-checks)

MarqRazz:
> @MarqRazz or @Abishalini have you seen walls forming at max range? (I think you guys use RealSense a lot, right?)

Sorry for the slow reply @AndyZe! We do use the RealSense but have our own node for maintaining the octomap and we perform our own clipping of the point cloud before its passed in so I have not see this issue.

tylerjw:
Closing as this work seems to be abandoned. Please re-open if you take back up this work.

PR #1844, opened 2023-01-06
Fix wrong use of C msg headers

rhaschke:
You inconsistently use C++ and C msg headers! This PR fixes all occurrences of wrong C header usage.

I'm really wondering why you didn't run into issues due to this before. When including `moveit/utils/moveit_error_code.h`, I'm getting these errors:
```
/opt/ros/rolling/include/moveit/utils/moveit_error_code.h:48:1: error: expected class-name before ‘{’ token
   48 | {
      | ^
/opt/ros/rolling/include/moveit/utils/moveit_error_code.h:54:43: error: ‘MoveItErrorCodes’ in namespace ‘moveit_msgs::msg’ does not name a type
   54 |   MoveItErrorCode(const moveit_msgs::msg::MoveItErrorCodes& code)
```
That's not surprising as the C++ and C types are fundamentally different, e.g.
`moveit_msgs::msg::MoveItErrorCodes` vs. `moveit_msgs__msg__MoveItErrorCodes`.

I guess you include the corresponding `.hpp` header somewhere else then...

### DIFF
--- a/moveit_core/distance_field/src/propagation_distance_field.cpp
+++ b/moveit_core/distance_field/src/propagation_distance_field.cpp
@@ -35,7 +35,7 @@
 /* Author: Mrinal Kalakrishnan, Ken Anderson */
 
 #include <moveit/distance_field/propagation_distance_field.h>
-#include <visualization_msgs/msg/marker.h>
+#include <visualization_msgs/msg/marker.hpp>
 #include <boost/iostreams/filtering_stream.hpp>
 #include <boost/iostreams/copy.hpp>
 #include <boost/iostreams/filter/zlib.hpp>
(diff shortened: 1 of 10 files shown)


AndyZe (commented):
I found quite a few more, for example:

`moveit_core/planning_scene/src/planning_scene.cpp:#include <octomap_msgs/conversions.h>`

Found with:

`grep -r ".h>" | grep "_msg"`

rhaschke:
You need to use this regex: `/msgs/.*\.h(?!p)` which I did. 
The file you mention, @AndyZe, is _not_ a generated message header.

AndyZe:
You mean that messages from debians should be included differently than messages built along with MoveIt packages? Hmm, that's annoying. I'll defer to somebody else to review the PR, then.

rhaschke:
I mean that only header files auto-generated via `rosidl` should be handled as suggested. The one you pointed out is a manually provided file: [octomap_msgs/conversions.h](https://github.com/OctoMap/octomap_msgs/blob/melodic-devel/include/octomap_msgs/conversions.h)

AndyZe (approved):

henningkayser (approved):
Thank you!

PR #1849, opened 2023-01-07
Cleanup msg includes: Use C++ instead of C header (backport #1844)

rhaschke:
Replacing #1847


rhaschke:
I really don't understand why GHA fails with including a moveit_msgs header file here.
Running `industrial_ci` locally works as expected.

Issue #1857, opened 2023-01-11
Servo: Address drift in Gazebo arm joint position control


sea-bass:
### Description
MoveIt Servo can cause drift in joint angles for some configurations if the target arm does not have sufficient filtering/smoothing in place. 

The example in this issue is with an Ignition Gazebo simulation used in MoveIt Studio, where the controllers are set up using the standard [`gz_ros2_control` plugin](https://github.com/ros-controls/gz_ros2_control). This arm is now being controlled at 200 Hz using sim time; however, we also see the same issues with other hardware we have tried such as the Kinova Gen3 Lite arm.

https://user-images.githubusercontent.com/4603398/211877454-6a1e2b0c-81e3-4117-9c0a-d44890961836.mp4

### Your environment
* ROS Distro: ROS 2 Humble
* OS Version: e.g. Ubuntu 22.04
* Source build of MoveIt 2.5.4
* RMW: CycloneDDS

### What we have tried
* Switching the arm to use velocity control, but it was quite unstable. Maybe we configured it wrong?
* Adding a lookahead parameter (as seen in https://github.com/ros-planning/moveit2/pull/1812)

### Other info
Here is the servo config in YAML format. Does anything stand out?

```yaml
use_gazebo: true # Whether the robot is started in a Gazebo simulation environment

## Properties of incoming commands
command_in_type: "unitless" # "unitless"> in the range [-1:1], as if from joystick. "speed_units"> cmds are in m/s and rad/s
scale:
  # Scale parameters are only used if command_in_type=="unitless"
  linear:  2.0  # Max linear velocity. Meters per publish_period. Unit is [m/s]. Only used for Cartesian commands.
  rotational:  3.0 # Max angular velocity. Rads per publish_period. Unit is [rad/s]. Only used for Cartesian commands.
  # Max joint angular/linear velocity. Rads or Meters per publish period. Only used for joint commands on joint_command_in_topic.
  joint: 2.0

# Disable Servo velocity scaling calculation and drive the arm at 50% max speed.
override_velocity_scaling_factor: 0.5

## Properties of outgoing commands
publish_period: 0.005  # 1/Nominal publish rate [seconds]
low_latency_mode: true  # Set this to true to publish as soon as an incoming Twist command is received (publish_period is ignored)

# What type of topic does your robot driver expect?
# Currently supported are std_msgs/Float64MultiArray or trajectory_msgs/JointTrajectory
command_out_type: std_msgs/Float64MultiArray

[...]

Issue #1428, opened 2022-07-13
moveit_servo build on Rolling build farm may be exceeding resource limits.


nuclearsandwich:
This issue is largely informational regarding build farm builds and it can be closed without action.
If there is a more preferred communication channel to use in the future, please let me know! 🙇🏼 

### Your environment
* ROS Distro: Rolling so far, but I think Humble is also susceptible.
* OS Version: Ubuntu 22.04 amd64
* Binary deb package: 2.5.1-1

---

Two prior builds ([12](https://build.ros2.org/view/Rbin_uJ64/job/Rbin_uJ64__moveit_servo__ubuntu_jammy_amd64__binary/12/) and [16](https://build.ros2.org/view/Rbin_uJ64/job/Rbin_uJ64__moveit_servo__ubuntu_jammy_amd64__binary/16)) of `moveit_servo` have failed during compilation on the build farm  while compiling `servo_calcs.cpp.o`. 

Build agents on the build farm are quad-core instances with 8GiB of RAM (specifically AWS c5.xlarge / c5a.xlarge instances), but are configured to run four single core builds in parallel which yields a rough RAM allocation of 8GiB/4 = 2GiB per package build.
This limit is not strictly enforced by the build isolation mechanism but if a build agent is saturated and memory is exhausted the host operating system may take action and kill build processes consuming high amounts of memory. The killed process is usually, but not always, the highest memory consumer among the build processes so there is the possibility of a "false positive" where `moveit_servo` is getting killed but is not actually consuming more than the expected resource limits. However, since the build was stopped in the same place both times. It is most likely that this compile operation in `moveit_servo` is the high memory consumer.

It is still possible for builds of `moveit_servo` to succeed if they are triggered at a time when the build farm is relatively quiet and there are fewer builds running on the agent alongside it. This is why build [17](https://build.ros2.org/view/Rbin_uJ64/job/Rbin_uJ64__moveit_servo__ubuntu_jammy_amd64__binary/17/), which I triggered manually this morning, succeeded. However rebuilds of Rolling and new releases of `moveit` may see regular build failures for `moveit_servo` while the build farm is under load.

### Build output

```
02:39:42 [ 19%] Building CXX object CMakeFiles/moveit_servo_lib.dir/src/servo_calcs.cpp.o
[...]

PR #1861, opened 2023-01-13
[Servo] Add an optional "Anti-drift" mode

AndyZe:
### Description

Addresses #1857

Also moves `composeJointTraj()` to a free utility function. That's better for:

- testing
- should decrease the size of the `servo_calcs` translation unit, which is starting to become a problem in CI:  https://github.com/ros-planning/moveit2/issues/1428

### DIFF
--- a/moveit_ros/moveit_servo/config/panda_simulated_config.yaml
+++ b/moveit_ros/moveit_servo/config/panda_simulated_config.yaml
@@ -69,3 +69,10 @@ check_collisions: true # Check collisions?
 collision_check_rate: 10.0 # [Hz] Collision-checking can easily bog down a CPU if done too often.
 self_collision_proximity_threshold: 0.01 # Start decelerating when a self-collision is this far [m]
 scene_collision_proximity_threshold: 0.02 # Start decelerating when a scene collision is this far [m]
+
+## Joint drift thresholding
+# Joints of certain robots such as Kinova and Ignition simulations tend to drift over time.
+# It often manifests as a slow sag due to gravity.
+# This enables a special mode to prevent joint drift by only updating positions of the actuated joints.
+# A value of false is fine for most arms with good position control, such as UR robots.
+apply_anti_drift: false
--- a/moveit_ros/moveit_servo/config/panda_simulated_config_pose_tracking.yaml
+++ b/moveit_ros/moveit_servo/config/panda_simulated_config_pose_tracking.yaml
@@ -62,3 +62,10 @@ check_collisions: true # Check collisions?
 collision_check_rate: 10.0 # [Hz] Collision-checking can easily bog down a CPU if done too often.
 self_collision_proximity_threshold: 0.01 # Start decelerating when a self-collision is this far [m]
 scene_collision_proximity_threshold: 0.02 # Start decelerating when a scene collision is this far [m]
+
+## Joint drift thresholding
+# Joints of certain robots such as Kinova and Ignition simulations tend to drift over time.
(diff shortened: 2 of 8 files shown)


moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -802,16 +802,32 @@ bool ServoCalcs::applyJointUpdate(const Eigen::ArrayXd& delta_theta, sensor_msgs
...
+    // Else, increment if Servo is actuating the joint, i.e. delta_theta[i] != 0
+    else if (delta_theta[i] != 0)
+    {
+      joint_state.position.at(i) += delta_theta[i];
+    }
+    // Else, keep the joint's original position from original_joint_state_
+    // This is what mitigates drift.
+    else
+    {
+      joint_state.position.at(i) != original_joint_state_.position.at(i);
```
sea-bass, line 821: This should just be an `=` operator, yeah?

moveit_ros/moveit_servo/config/panda_simulated_config.yaml
```diff
@@ -69,3 +69,10 @@ check_collisions: true # Check collisions?
 collision_check_rate: 10.0 # [Hz] Collision-checking can easily bog down a CPU if done too often.
 self_collision_proximity_threshold: 0.01 # Start decelerating when a self-collision is this far [m]
 scene_collision_proximity_threshold: 0.02 # Start decelerating when a scene collision is this far [m]
+
+## Joint drift thresholding
+# Joints of certain robots such as Kinova and Ignition simulations tend to drift over time.
+# It often manifests as a slow sag due to gravity.
+# This enables a special mode to prevent joint drift.
+# A value of false is fine for most arms such as UR robots.
```
sea-bass, line 77: Would something like this be appropriate?

```suggestion
# A value of false is fine for most arms with good position control, such as UR robots.
```

moveit_ros/moveit_servo/config/panda_simulated_config_pose_tracking.yaml
```diff
@@ -62,3 +62,10 @@ check_collisions: true # Check collisions?
 collision_check_rate: 10.0 # [Hz] Collision-checking can easily bog down a CPU if done too often.
 self_collision_proximity_threshold: 0.01 # Start decelerating when a self-collision is this far [m]
 scene_collision_proximity_threshold: 0.02 # Start decelerating when a scene collision is this far [m]
+
+## Joint drift thresholding
+# Joints of certain robots such as Kinova and Ignition simulations tend to drift over time.
+# It often manifests as a slow sag due to gravity.
+# This enables a special mode to prevent joint drift.
```
sea-bass, line 69: Can you add a little more color to this?

```suggestion
# This enables a special mode to prevent joint drift by <doing something>
```

sea-bass (commented):
Looks good and we can test with it... but I think I found a small bug

moveit_ros/moveit_servo/src/servo_calcs.cpp
AndyZe: yep, good catch. Obviously I didn't test this with anti-drift enabled.

AndyZe: i've tested now with the Panda in sim and it seems fine

henningkayser (commented):
Changes look good, I just have some cleanup suggestions.

AndyZe:
Unfortunately, @sea-bass tells me this does not help with the issue. I think I'll just close this.

PR #1865, opened 2023-01-16
gigantic speed up to move group interface

azalutsky:
### Description

This speedup removes the 100ms delay during a planning time, which increases planning times by a substantial amount. 
For example, my 2 robot arms without this takes 600ms+ to plan. By reducing this sleep to 1ms the plans are then in the 150ms range. 

### Checklist

### DIFF
--- a/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
+++ b/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
@@ -817,7 +817,7 @@ public:
     // wait until send_goal_opts.result_callback is called
     while (!done)
     {
-      std::this_thread::sleep_for(std::chrono::milliseconds(100));
+      std::this_thread::sleep_for(std::chrono::milliseconds(1));
     }
 
     if (code != rclcpp_action::ResultCode::SUCCEEDED)
@@ -895,7 +895,7 @@ public:
     // wait until send_goal_opts.result_callback is called
     while (!done)
     {
-      std::this_thread::sleep_for(std::chrono::milliseconds(100));
+      std::this_thread::sleep_for(std::chrono::milliseconds(1));
     }
 
     if (code != rclcpp_action::ResultCode::SUCCEEDED)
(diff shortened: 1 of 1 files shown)


azalutsky:
Thanks guys, if anyone has any other tricks to speed up planning times- I'd love to know!

azalutsky:
I am noticing however with multiple robots that there is a bug somewhere with these actions. 

Independent of the timeout, but the timeout alleviates it a little bit. 

For dual robot, the done flag claims it plans both and completes it however only the first robot in the group executes the trajectory while the second robot stays put. 

Seems to be a bug. I started tracking it but if someone has an idea I'd love to hear your thoughts as I investigate. 

I posted in discord's bug's channel about it. 

Would love a second opinion.

rhaschke:
@ros-planning/moveit-maintainers, why does this code use busy waiting??? There are [condition variables](https://en.cppreference.com/w/cpp/thread/condition_variable) for this purpose...

samialperen:
@azalutsky I did not see a performance increase even after this PR. I was expecting it would increase the performance but did not happen. Did you manage to find any other tricks to decrease planning time for humble?

azalutsky:
If you use move group interface, when you call plan the action has this 100ms latency. 
What planning times are you seeing?

azalutsky:
Is it possible you're not using moveit2 via source? Did you perhaps install moveit2 through apt-get or didn't colcon build with making sure that the package was built by source? I believe foxy requires some additional output.

samialperen:
I built it from source and ros2 pkg prefix shows my workspace. For the reference, I am using moveit humble, (2.5.4)

samialperen:
I find another issue stating that foxy has better performance than humble when it comes to planning times, but I could not test it on my system since I have too many dependencies for humble, I can not change it to foxy easily

samialperen:
@azalutsky you were right. Somehow, there is a package collision even if ros2 pkg prefix shows the right package. After applying the fix manually, planning time reduced from 600 ms to 170ms.

PR #1872, opened 2023-01-19
Remove redundant Bullet install+export from subdirectory

### LINKED ISSUE #1858: Wrong use of ament_export_targets from package sub folders
As pointed out [here](https://github.com/ament/ament_cmake/pull/416#issuecomment-1379685273), MoveIt's cmake files use `ament_export_targets` from package subdirectories, which is forbidden and will definitely fail in future releases of `ament_cmake`.

henningkayser:
Bullet is already installed and exported at the top level https://github.com/ros-planning/moveit2/blob/main/moveit_core/CMakeLists.txt#L122.
 
This should fix #1858 and hopefully #1869 as well.

### DIFF
--- a/moveit_core/collision_detection_bullet/CMakeLists.txt
+++ b/moveit_core/collision_detection_bullet/CMakeLists.txt
@@ -51,12 +51,6 @@ target_link_libraries(collision_detector_bullet_plugin
 
 install(DIRECTORY include/ DESTINATION include/moveit_core)
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/moveit_collision_detection_bullet_export.h DESTINATION include/moveit_core)
-install(TARGETS moveit_collision_detection_bullet collision_detector_bullet_plugin EXPORT export_moveit_collision_detection_bullet
-  LIBRARY DESTINATION lib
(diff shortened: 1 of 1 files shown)


sjahr (commented):
What about `collision_detector_bullet_plugin`? Doesn't this need to be installed?

henningkayser:
> What about `collision_detector_bullet_plugin`? Doesn't this need to be installed?

Great questions! It should be installed by the tutorials, and also by common sense. Export maybe not, since it's not provided for import in another package. However, has anyone noticed issues with FCL? We are not installing or exporting it either. This could be an issue that only occurs with the debians, though. @tylerjw @ChrisThrasher any thoughts?

tylerjw:
@henningkayser I made this PR because I believe this change is not complete as there is the plugin library that would not be installed with just this change: https://github.com/ros-planning/moveit2/pull/1889

henningkayser:
> @henningkayser I made this PR because I believe this change is not complete as there is the plugin library that would not be installed with just this change: https://github.com/ros-planning/moveit2/pull/1889

I haven't seen yours. Could you update your branch? I'll close this PR.

PR #1875, opened 2023-01-19
Clean up kinematic_constraints/utils

AndyZe:
### Description
The first commit is minor cleanup.

The second commit adds a new function, `updateConstraints()`. The idea is, it's annoying and inefficient to construct a new Constraint object every time. This allows the user to update only the important information. If you like the idea, I'll add the other versions, like:

`bool updateConstraints(moveit_msgs::msg::Constraints& constraints,  const geometry_msgs::msg::PoseStamped& pose);`

and so on.

### DIFF
--- a/moveit_core/kinematic_constraints/include/moveit/kinematic_constraints/utils.h
+++ b/moveit_core/kinematic_constraints/include/moveit/kinematic_constraints/utils.h
@@ -101,6 +101,18 @@ moveit_msgs::msg::Constraints constructGoalConstraints(const moveit::core::Robot
                                                        double tolerance = std::numeric_limits<double>::epsilon());
 
 /**
+ * \brief Update joint constraints with a new JointModelGroup state
+ *
+ * @param [in, out] constraints Previously-constructed constraints to update
+ * @param [in] state The new target state
+ * @param [in] jmg Specify which JointModelGroup to update
+ *
+ * @return true if successful, false if any joint_constraint within constraints was not able to be updated
+ */
+bool updateJointConstraints(moveit_msgs::msg::Constraints& constraints, const moveit::core::RobotState& state,
+                            const moveit::core::JointModelGroup* jmg);
+
+/**
  * \brief Generates a constraint message intended to be used as a goal
  * constraint for a given link.  The full constraint will contain a
  * \ref PositionConstraint and a \ref OrientationConstraint,
--- a/moveit_core/kinematic_constraints/include/moveit/kinematic_constraints/utils.h
+++ b/moveit_core/kinematic_constraints/include/moveit/kinematic_constraints/utils.h
@@ -107,7 +107,7 @@ moveit_msgs::msg::Constraints constructGoalConstraints(const moveit::core::Robot
  * @param [in] state The new target state
  * @param [in] jmg Specify which JointModelGroup to update
  *
- * @return true if successful, false if any joint_constraint within constraints was not able to be updated
+ * @return true if all joint constraints were updated
  */
 bool updateJointConstraints(moveit_msgs::msg::Constraints& constraints, const moveit::core::RobotState& state,
                             const moveit::core::JointModelGroup* jmg);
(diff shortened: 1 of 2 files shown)


moveit_core/kinematic_constraints/include/moveit/kinematic_constraints/utils.h
```diff
@@ -66,9 +66,6 @@ namespace kinematic_constraints
 moveit_msgs::msg::Constraints mergeConstraints(const moveit_msgs::msg::Constraints& first,
                                                const moveit_msgs::msg::Constraints& second);
 
-/** \brief Check if any constraints were specified */
-[[deprecated("Use moveit/utils/message_checks.h instead")]] bool isEmpty(const moveit_msgs::msg::Constraints& constr);
```
AndyZe, line 70: This has been deprecated for 4 years now, prob safe to remove

moveit_core/kinematic_constraints/src/utils.cpp
```diff
@@ -165,6 +160,23 @@ moveit_msgs::msg::Constraints constructGoalConstraints(const moveit::core::Robot
   return goal;
 }
 
+bool updateConstraints(moveit_msgs::msg::Constraints& constraints, const moveit::core::RobotState& state,
```
henningkayser, line 163:
```suggestion
bool updateJointConstraints(moveit_msgs::msg::Constraints& constraints, const moveit::core::RobotState& state,
```

moveit_core/kinematic_constraints/src/utils.cpp
```diff
@@ -165,6 +160,23 @@ moveit_msgs::msg::Constraints constructGoalConstraints(const moveit::core::Robot
...
+                       const moveit::core::JointModelGroup* jmg)
+{
+  std::vector<double> vals;
+  state.copyJointGroupPositions(jmg, vals);
+  if (constraints.joint_constraints.size() != vals.size())
```
henningkayser, line 168: If you have to resize the joint constraints, you are using a completely different JMG which would require also initializing `joint_name`, `weight` and tolerances. If you really just want to update the position values, I would rule out this case by only iterating the constraints and updating the values of those joints that are part of the group.

I would even say that it should explicitly only update the constraints values that are already present without adding or removing new ones. Otherwise, a new constraint should be constructed with all the required member values.

AndyZe: That makes a lot of sense. Done.

moveit_core/kinematic_constraints/src/utils.cpp
```diff
@@ -227,6 +246,36 @@ moveit_msgs::msg::Constraints constructGoalConstraints(const std::string& link_n
   return goal;
 }
 
+bool updatePoseConstraint(moveit_msgs::msg::Constraints& constraints, const std::string& link_name,
```
henningkayser, line 249: This function should just use something like `return updatePositionConstraint() && updateOrientationConstraint()`

AndyZe: it's not quite that tidy b/c:

-  updatePositionConstraint() takes a geometry_msgs/Point instead of a geometry_msgs/Pose
- updateOrientationConstraint() takes a geometry_msgs::msg::QuaternionStamped instead of a geometry_msgs/Quaternion

So I'd have to convert the message types before calling those functions. I guess it's slightly cleaner that way

henningkayser (approved):

moveit_core/kinematic_constraints/src/utils.cpp
```diff
@@ -227,6 +246,22 @@ moveit_msgs::msg::Constraints constructGoalConstraints(const std::string& link_n
...
+                          const geometry_msgs::msg::PoseStamped& pose)
+{
+  // Convert message types so the existing functions can be used
+  geometry_msgs::msg::PointStamped point;
```
henningkayser, line 253: @AndyZe I don't know why I missed this, but I think we have to pass the stamp information with this as well, since that one is defining the reference frame

AndyZe: https://github.com/ros-planning/moveit2/pull/1890

AndyZe: https://github.com/ros-planning/moveit2/pull/1890

PR #1876, opened 2023-01-19
Add a default stopping criterion for parallel planning

AndyZe:
Add a parallel planning stop condition that will be commonly used. It stops at the first solution from any planner.

### DIFF
--- a/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/plan_solutions.hpp
+++ b/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/plan_solutions.hpp
@@ -80,7 +80,7 @@ private:
  *  \param [in] solutions Vector of solutions to chose the shortest one from
  *  \return Shortest solution, trajectory_ of the returned MotionPlanResponse is a nullptr if no solution is found!
  */
-static planning_interface::MotionPlanResponse
+static inline planning_interface::MotionPlanResponse
 getShortestSolution(const std::vector<planning_interface::MotionPlanResponse>& solutions)
 {
   // Find trajectory with minimal path
--- a/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/plan_solutions.hpp
+++ b/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/plan_solutions.hpp
@@ -80,7 +80,7 @@ private:
  *  \param [in] solutions Vector of solutions to chose the shortest one from
  *  \return Shortest solution, trajectory_ of the returned MotionPlanResponse is a nullptr if no solution is found!
  */
-planning_interface::MotionPlanResponse
+static planning_interface::MotionPlanResponse
 getShortestSolution(const std::vector<planning_interface::MotionPlanResponse>& solutions)
 {
   // Find trajectory with minimal path
--- a/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
+++ b/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
@@ -209,11 +209,6 @@ public:
   const planning_interface::MotionPlanResponse& getLastMotionPlanResponse();
 
 private:
-  /** \brief A callback function that is used as the default parallel planning stop criterion.
-   *          It stops parallel planning as soon as any planner finds a solution. */
-  static bool stopAtFirstSolution(PlanSolutions const& plan_solutions,
-                                  PlanningComponent::MultiPipelinePlanRequestParameters const& plan_request_parameters);
-
   // Core properties and instances
   rclcpp::Node::SharedPtr node_;
   MoveItCppPtr moveit_cpp_;
--- a/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
+++ b/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
@@ -49,6 +49,7 @@
 namespace moveit_cpp
 {
 MOVEIT_CLASS_FORWARD(PlanningComponent);  // Defines PlanningComponentPtr, ConstPtr, WeakPtr... etc
+
 class PlanningComponent
 {
 public:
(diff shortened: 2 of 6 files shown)


moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
```diff
@@ -208,6 +209,11 @@ class PlanningComponent
   const planning_interface::MotionPlanResponse& getLastMotionPlanResponse();
 
 private:
+  /** \brief A callback function that is used as the default parallel planning stop criterion.
+   *          It stops parallel planning as soon as any planner finds a solution. */
+  static bool stopAtFirstSolution(PlanSolutions const& plan_solutions,
+                                  PlanningComponent::MultiPipelinePlanRequestParameters const& plan_request_parameters);
```
AndyZe, line 215: I tried to make this a free function, but:

It needs knowledge of `PlanningComponent::MultiPipelinePlanRequestParameters`. [Apparently you can't forward-declare a struct nested in a class in C++](https://stackoverflow.com/questions/951234/forward-declaration-of-nested-types-classes-in-c), so... I don't think a free function is feasible.

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
```diff
@@ -198,7 +199,7 @@ class PlanningComponent
   planning_interface::MotionPlanResponse
   plan(const MultiPipelinePlanRequestParameters& parameters,
        const SolutionCallbackFunction& solution_selection_callback = &getShortestSolution,
-       StoppingCriterionFunction stopping_criterion_callback = nullptr);
+       StoppingCriterionFunction stopping_criterion_callback = &moveit_cpp::PlanningComponent::stopAtFirstSolution);
```
sjahr, line 202:
```suggestion
       StoppingCriterionFunction stopping_criterion_callback = nullptr);
```

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
sjahr: If we don't make it a default, you could make it a free function. Another option would be to move `MultiPipelinePlanRequestParameters` out of the `PlanningComponentClass`. The main reason to keep it there was to treat it similarly as `PlanRequestParameters` and the lack of a need to keep it outside of the class but that would justify moving it out

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
```diff
@@ -419,4 +419,25 @@ const planning_interface::MotionPlanResponse& PlanningComponent::getLastMotionPl
...
+
+bool PlanningComponent::stopAtFirstSolution(
+    PlanSolutions const& plan_solutions,
+    PlanningComponent::MultiPipelinePlanRequestParameters const& /*plan_request_parameters*/)
+{
+  // Read solutions that are found up to this point from a thread safe storage
+  auto const& solutions = plan_solutions.getSolutions();
+
+  // Stop at the first successful plan
+  for (auto const& solution : solutions)
```
sjahr, line 431: Nit: Does this work?

```suggestion
  // Stop at the first successful plan
  for (auto const& solution : plan_solutions.getSolutions())
```

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
```diff
@@ -419,4 +419,25 @@ const planning_interface::MotionPlanResponse& PlanningComponent::getLastMotionPl
...
+  {
+    if (bool(solution))
```
sjahr, line 433: Nit: You could clarify with a small comment here that `bool(solution)` evaluate the error code of the solution and is true if it is `SUCCESS`

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
```diff
@@ -419,4 +419,25 @@ const planning_interface::MotionPlanResponse& PlanningComponent::getLastMotionPl
...
+    {
+      // Return true to abort the other pipelines
+      return true;
+    }
+  }
+  // Return false when parallel planning should continue
```
sjahr, line 439:
```suggestion
  // Return false when parallel planning should continue because it hasn't found a successful solution yet
```

sjahr (commented):
I think adding this as an option for a Stopping criterion is a very good idea. Nevertheless, I wouldn't use it as default because it breaks one of the main advantages of parallel planning: Optimizing for path length by default.
I assume most users would like the see better paths (shorter is likely to be less "wacky") and if they run for example CHOMP and RRTC in parallel, this criteria would never let CHOMP find a better solution (and there is a chance that CHOMP does that) because RRTC is always faster, and CHOMP will be aborted. I think the default settings shouldn't disable any feature and this stopping criteria would change the default behavior from "Find the shortest solution" to "Find any solution". 

Besides that, I'd like to merge this because there are definitely use cases where this stopping criterion is the right one to use!

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
AndyZe: indeed!

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
AndyZe: This is tricky. `MultiPipelinePlanRequestParameters` needs knowledge of `PlanRequestParameters` at L112:

`std::vector<PlanRequestParameters> multi_plan_request_parameters;`

AndyZe: I moved the new callback to a new translation unit. Seems ok to me.

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/plan_solutions.hpp
```diff
@@ -80,7 +80,7 @@ class PlanSolutions
  *  \param [in] solutions Vector of solutions to chose the shortest one from
  *  \return Shortest solution, trajectory_ of the returned MotionPlanResponse is a nullptr if no solution is found!
  */
-planning_interface::MotionPlanResponse
+static planning_interface::MotionPlanResponse
```
AndyZe, line 83: Since the new `parallel_planning_callbacks.h` includes this header, the `static` is necessary to avoid multiple definitions.

henningkayser: I think it should also make this function `inline` since the definition is provided with the header file.

AndyZe: done

sjahr (approved):

PR #1877, opened 2023-01-19
Switch to clang-format-14

### LINKED ISSUE #1736: Migrate to clang-format-14
As we are now firmly on Ubuntu 22.04 it makes sense to migrate to clang-format-14.

henningkayser:
Fixes #1736

### DIFF
--- a/.docker/ci/Dockerfile
+++ b/.docker/ci/Dockerfile
@@ -25,7 +25,7 @@ RUN \
         # Some basic requirements
         wget git sudo curl \
         # Preferred build tools
-        clang clang-format-12 clang-tidy clang-tools \
+        clang clang-format-14 clang-tidy clang-tools \
         ccache && \
     #
     # Globally disable git security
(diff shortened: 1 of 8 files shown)


AndyZe (approved):
Note that our installation instructions recommend clang-format-10:

https://moveit.ros.org/install-moveit2/source/

I can make a PR for that, although it will be a little annoying b/c we have different versions for every ROS distro.

henningkayser:
> Note that our installation instructions recommend clang-format-10:
> 
> https://moveit.ros.org/install-moveit2/source/
> 
> I can make a PR for that, although it will be a little annoying b/c we have different versions for every ROS distro.

Right, we should really fix that. Maybe we can even drop the "-10" since we are sticking with the distro default anyway.

AndyZe:
Actually I don't think clang-format is necessary in building MoveIt. Will remove it completely from the installation instructions.

henningkayser:
There is a bunch of stuff that we can probably drop. The typical ROS 2 setup instructions + rosdep should be all we need.

Issue #1862, opened 2023-01-15
Remove last_plan_solution_ in PlanningComponent


sjahr:
Let us remove last_plan_solution_ from the PlanningComponent class once #1835 is merged. With the `execute()` method removed, the only actual usage of last_plan_solution_ within the class is removed. It would simplify this class and I think it is better to let the user of PlanningComponent handle the lifetime of results from `plan()` rather than having a member in the planning component who's state might be unknown.
What do you think @henningkayser @AndyZe ?

AndyZe:
I'd agree with that because it seems this is not used anywhere:

```
const planning_interface::MotionPlanResponse& PlanningComponent::getLastMotionPlanResponse()
{
  return last_plan_solution_;
}
```

henningkayser:
I like this idea a lot. Without a meaningful state the only real use case of the PlanningComponent is building of motion plan requests, which should in turn be addressed by #1455. I think that way the API would become much easier to use and to extend.

sjahr:
I'll create a PR. Small annoyance: This will definitely break the API since
```
[[deprecated]]
bool PlanningComponent::execute(bool blocking);
```
wont work anymore

sjahr:
Closed with #1879

PR #1879, opened 2023-01-20
Remove last_plan_solution_ in PlanningComponent

sjahr:
### Description

Addresses discussion in #1862 

BREAKING CHANGE

### Checklist

### DIFF
--- a/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
+++ b/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
@@ -191,7 +191,7 @@ public:
   planning_interface::MotionPlanResponse plan();
   /** \brief Run a plan from start or current state to fulfill the last goal constraints provided by setGoal() using the
    * provided PlanRequestParameters. */
-  planning_interface::MotionPlanResponse plan(const PlanRequestParameters& parameters, const bool store_solution = true);
+  planning_interface::MotionPlanResponse plan(const PlanRequestParameters& parameters);
 
   /** \brief Run a plan from start or current state to fulfill the last goal constraints provided by setGoal() using the
    * provided PlanRequestParameters. */
@@ -203,9 +203,6 @@ public:
   /** \brief Execute the latest computed solution trajectory computed by plan(). By default this function terminates
    * after the execution is complete. The execution can be run in background by setting blocking to false. */
   [[deprecated("Use MoveItCpp::execute()")]] bool execute(bool blocking = true);
-
-  /** \brief Return the last plan solution*/
-  const planning_interface::MotionPlanResponse& getLastMotionPlanResponse();
 
 private:
   // Core properties and instances
@@ -224,7 +221,6 @@ private:
   PlanRequestParameters plan_request_parameters_;
   moveit_msgs::msg::WorkspaceParameters workspace_parameters_;
   bool workspace_parameters_set_ = false;
-  planning_interface::MotionPlanResponse last_plan_solution_;
 
   // common properties for goals
   // TODO(henningkayser): support goal tolerances
@@ -236,12 +232,3 @@ private:
   // std::unique_ptr<moveit_msgs::msg::TrajectoryConstraints> trajectory_constraints_;
 };
 }  // namespace moveit_cpp
-
-namespace moveit
-{
-namespace planning_interface
-{
-using PlanningComponent [[deprecated("use moveit_cpp")]] = moveit_cpp::PlanningComponent;
-[[deprecated("use moveit_cpp")]] MOVEIT_DECLARE_PTR(PlanningComponent, moveit_cpp::PlanningComponent);
-}  // namespace planning_interface
-}  // namespace moveit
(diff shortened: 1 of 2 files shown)


moveit_ros/planning/moveit_cpp/src/planning_component.cpp
```diff
@@ -398,25 +376,7 @@ bool PlanningComponent::setGoal(const std::string& goal_state_name)
...
-  //  RCLCPP_ERROR("Failed to parameterize trajectory");
-  //  return false;
-  //}
-  return moveit_cpp_->execute(last_plan_solution_.trajectory_, blocking);
-}
-
-const planning_interface::MotionPlanResponse& PlanningComponent::getLastMotionPlanResponse()
-{
-  return last_plan_solution_;
+  RCLCPP_ERROR(LOGGER, "There is no successful plan to execute");
```
sjahr, line 379: I am wondering whether or not this means we can remove the whole function :thinking:

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
```diff
@@ -236,12 +232,3 @@ class PlanningComponent
   // std::unique_ptr<moveit_msgs::msg::TrajectoryConstraints> trajectory_constraints_;
 };
 }  // namespace moveit_cpp
-
-namespace moveit
```
sjahr, line 240: This is deprecated for 22 months. I'd remove it with this PR but I can also revert this change if you don't feel so

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
AndyZe: I would say so

AndyZe (commented):
Looks good to me besides removing that last function

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
AndyZe: Prob should add a deprecation warning

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
```diff
@@ -396,27 +374,9 @@ bool PlanningComponent::setGoal(const std::string& goal_state_name)
...
-  // max_acceleration_scaling_factor_))
-  //{
-  //  RCLCPP_ERROR("Failed to parameterize trajectory");
-  //  return false;
-  //}
-  return moveit_cpp_->execute(last_plan_solution_.trajectory_, blocking);
-}
-
-const planning_interface::MotionPlanResponse& PlanningComponent::getLastMotionPlanResponse()
+bool PlanningComponent::execute(bool /*blocking*/)
```
henningkayser, line 377: we should really drop the blocking flag altogether.

henningkayser (approved):

moveit_ros/planning/moveit_cpp/src/planning_component.cpp
AndyZe: The "simultaneous execution" PR to ROS2 eventually (I hope so!). It looks like it does use a blocking flag in MoveItCpp but not the blocking flag in PlanningComponent. So I agree, the `blocking` flag here could probably be dropped.

https://github.com/ros-planning/moveit/pull/3243/files

sjahr:
@AndyZe I removed the execute function from the source file but kept it barebone in the header so the deprecation warning will still be there. Does this work for you or is there an even smarter way?

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
```diff
@@ -202,10 +202,10 @@ class PlanningComponent
 
   /** \brief Execute the latest computed solution trajectory computed by plan(). By default this function terminates
    * after the execution is complete. The execution can be run in background by setting blocking to false. */
-  [[deprecated("Use MoveItCpp::execute()")]] bool execute(bool blocking = true);
-
-  /** \brief Return the last plan solution*/
-  const planning_interface::MotionPlanResponse& getLastMotionPlanResponse();
```
AndyZe, line 208: @sjahr I think you might have screwed up the previous commit. Didn't you mean to add a deprecation warning here?

sjahr: I'll fix that

sjahr: :thinking: After giving it a second thought, I am unsure how I can deprecate this function in a good way. It returns a reference to a not existing member variable. If I'd keep the member variable the whole point of the PR is gone. @AndyZe Should we keep it this way? It is either way a breaking change

AndyZe: I see. Well, let's just delete it then. The nice thing about Rolling is we don't have to worry too much about API breakage  :+1:

AndyZe (approved):

PR #1887, opened 2023-01-25
Add ament export

galou:
Also sort `find_package` entries alphabetically.

### Description

I had an issue when trying to compile a package that depended on `moveit_kinematics`. IIRC the issue was with `tf2_kdl` but I added also all other packages except `orocos_kdl` which is added separately as `orocos_kdl_vendor` for some reason.

I didn't change and `find_package` line, just sorted them alphabetically.

### Checklist

### DIFF
--- a/moveit_kinematics/CMakeLists.txt
+++ b/moveit_kinematics/CMakeLists.txt
@@ -43,13 +43,14 @@ set(THIS_PACKAGE_LIBRARIES
 )
 
 set(THIS_PACKAGE_INCLUDE_DEPENDS
-  Boost
+  Eigen3
   class_loader
   generate_parameter_library
   kdl_parser
   moveit_core
   moveit_msgs
   moveit_ros_planning
+  orocos_kdl_vendor
   pluginlib
   random_numbers
   rclcpp
@@ -80,7 +81,7 @@ install(
 )
 
 ament_export_targets(${PROJECT_NAME}Targets HAS_LIBRARY_TARGET)
-ament_export_dependencies(${THIS_PACKAGE_INCLUDE_DEPENDS} orocos_kdl_vendor)
+ament_export_dependencies(${THIS_PACKAGE_INCLUDE_DEPENDS})
 
 if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
--- a/moveit_kinematics/CMakeLists.txt
+++ b/moveit_kinematics/CMakeLists.txt
@@ -6,18 +6,19 @@ find_package(moveit_common REQUIRED)
 moveit_package()
 
 find_package(ament_cmake REQUIRED)
+
 find_package(Eigen3 REQUIRED)
-find_package(orocos_kdl REQUIRED)
-find_package(tf2_kdl REQUIRED)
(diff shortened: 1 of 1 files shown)


moveit_kinematics/CMakeLists.txt
```diff
@@ -6,18 +6,19 @@ find_package(moveit_common REQUIRED)
 moveit_package()
 
 find_package(ament_cmake REQUIRED)
+
 find_package(Eigen3 REQUIRED)
```
AndyZe, line 10: Thanks for the PR!

I have some questions that @ChrisThrasher would probably know the answer to:

- Boost is only under `set(THIS_PACKAGE_INCLUDE_DEPENDS
  Boost`
- Should Eigen3 go under 
`set(THIS_PACKAGE_INCLUDE_DEPENDS` as well?

ChrisThrasher: I believe so if Eigen types appear in the interface of this library.

AndyZe: Well, Boost isn't part of the interface of this library. So, remove Boost from `set(THIS_PACKAGE_INCLUDE_DEPENDS`. Make sense, @ChrisThrasher?

Eigen is part of the interface so I think it should be added to `set(THIS_PACKAGE_INCLUDE_DEPENDS`

henningkayser: Yes, Eigen3 should be set as a dependency. If we are exporting headers with Eigen includes, we should probably even use `eigen3_cmake_module` for better ament support https://github.com/ros2/eigen3_cmake_module.

AndyZe (commented):
@galou as you mentioned, it's weird to have `orocos_kdl_vendor` by itself at L83. Can you please add it to `set(THIS_PACKAGE_INCLUDE_DEPENDS` as well

moveit_kinematics/CMakeLists.txt
ChrisThrasher: That seems right to not export Boost. No need to export private dependencies since downstream projects ought not know about them.

AndyZe (approved):
I went ahead and pushed the small changes mentioned here

moveit_kinematics/CMakeLists.txt
```diff
@@ -43,13 +43,14 @@ set(THIS_PACKAGE_LIBRARIES
 )
 
 set(THIS_PACKAGE_INCLUDE_DEPENDS
-  Boost
+  Eigen3
```
galou, line 46: Boost is included L.24. @AndyZe are you sure it's good to remove it?

AndyZe: @ChrisThrasher thought it was OK. He's the expert. If not, we can make another quick PR to put it back

AndyZe: I believe the thinking is, as long as these plugins build, other packages do not need to be aware of Boost. It's not part of the interface.

sjahr (approved):

vatanaksoytezer (approved):

PR #1899, opened 2023-01-26
Fix timeout in waitForCurrentState

c-rizz:
### Description

CurrentStateMonitor::waitForCurrentState() was getting stuck in an endless loop when use_sim_time==True and joint_states is not being received, also, it could not be terminated by SIGINT.
Now it should always properly check for the timeout and also checks that the rcl context is still up.

(Originally made for humble branch: see [https://github.com/ros-planning/moveit2/pull/1894](https://github.com/ros-planning/moveit2/pull/1894))

### Checklist

### DIFF
--- a/moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
+++ b/moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
@@ -238,6 +238,7 @@ bool CurrentStateMonitor::waitForCurrentState(const rclcpp::Time& t, double wait
   rclcpp::Duration elapsed(0, 0);
   rclcpp::Duration timeout = rclcpp::Duration::from_seconds(wait_time_s);
 
+  rclcpp::Clock steady_clock(RCL_STEADY_TIME);
   std::unique_lock<std::mutex> lock(state_update_lock_);
   while (current_state_time_ < t)
   {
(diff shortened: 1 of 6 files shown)


moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
```diff
@@ -247,30 +248,32 @@ bool CurrentStateMonitor::waitForCurrentState(const rclcpp::Time& t, double wait
...
-                  "Didn't received robot state (joint angles) with recent timestamp within "
-                  "%f seconds.\n"
+                  "Didn't receive robot state (joint angles) with recent timestamp within "
+                  "%f seconds. Requested time %f, but latest received state has time %f.\n"
                   "Check clock synchronization if your are running ROS across multiple machines!",
-                  wait_time_s);
+                  wait_time_s, t.seconds(), current_state_time_.seconds());
+      return false;
+    }
+    if (!rclcpp::ok())
```
henningkayser, line 274: We have a mocked test for the CurrentStateMonitor that works without rclcpp runtime. You might have to add a function `CurrentStateMonitor::MiddlewareHandle::ok()` that injects `rclcpp::ok()` and adapt the unit test mock here https://github.com/ros-planning/moveit2/blob/main/moveit_ros/planning/planning_scene_monitor/test/current_state_monitor_tests.cpp#L48

henningkayser (commented):
Thank you! The fix looks correct to me. The only thing left would be to remove the rclcpp runtime dependency as described in the comment. Maybe @tylerjw or @Abishalini could help out with what's needed for this.

c-rizz:
I updated the code by adding an ok() function in the middleware handle. I couldn't fully check it for now as my main setup is on humble, but I tried on a rolling docker and it compiles and I think the tests are fine.

henningkayser (approved):

sjahr (approved):

PR #1901, opened 2023-01-26
Install moveit_kinematics headers within additional directory

ChrisThrasher:
### Description

Expanding on the work we started in #1786.

### DIFF
--- a/moveit_kinematics/CMakeLists.txt
+++ b/moveit_kinematics/CMakeLists.txt
@@ -69,7 +69,7 @@ install(
   LIBRARY DESTINATION lib
   ARCHIVE DESTINATION lib
   RUNTIME DESTINATION bin
-  INCLUDES DESTINATION include
+  INCLUDES DESTINATION include/moveit_kinematics
 )
 
 ament_export_targets(${PROJECT_NAME}Targets HAS_LIBRARY_TARGET)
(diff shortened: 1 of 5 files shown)


moveit_kinematics/CMakeLists.txt
```diff
@@ -69,7 +69,7 @@ install(
   LIBRARY DESTINATION lib
   ARCHIVE DESTINATION lib
   RUNTIME DESTINATION bin
-  INCLUDES DESTINATION include
+  INCLUDES DESTINATION include/moveit_kinematics
```
tylerjw, line 72: Are you sure we need this line at all?

ChrisThrasher: I believe it can be removed. So long as another package in this repo depends on this and uses its headers, we can properly test this change. Because I'm not sure if that's true, I wasn't going to make that change.

henningkayser: do we even need to install headers at all for all these plugins?

ChrisThrasher: I think it's not unreasonable to install plugin headers

ChrisThrasher:
I keep amending the latest commit to force CI to re-run. Yesterday it was failing on some Docker step that seemed unrelated to any of the changes I made.

tylerjw (approved):

sjahr (approved):

PR #1902, opened 2023-01-26
Fix kinematics plugin loader bug

mechwiz:
### Description

There is a bug where the wrong IK solver plugin is assigned to the planning groups defined in the SRDF. I noticed this issue because I had multiple kinematic plugins defined in my yaml file with different IK solvers and after doing some debug logging, it was definitely apparant that groups that shouldn't be having an IK solver had one and that groups that were supposed to have an IK solver sometimes had the wrong one.

Turns out that in this for loop here within the [`allocKinematicsSolver`](https://github.com/ros-planning/moveit2/blob/main/moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp#L143) function, we essentially try to assign the first solver within `possible_kinematics_solvers_` to the joint group passed into the function **without even checking to see if that solver even corresponds to that joint group**. Therefore when this function is called for each joint group in the SRDF in the [`robot_model_loader`](https://github.com/ros-planning/moveit2/blob/main/moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp#L279), it is possible to get IK solvers set to groups that were never meant to have them or for the wrong one to be set. 

This PR fixes this issue. It also fixes this issue #1633 .

@AndyZe @nbbrooks @swiz23

### Checklist

### DIFF
--- a/moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
+++ b/moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
@@ -142,6 +142,11 @@ public:
     std::scoped_lock slock(lock_);
     for (auto const& [group, solver] : possible_kinematics_solvers_)
     {
+      // Don't bother trying to load a solver for the wrong group
+      if (group != jmg->getName())
+      {
+        continue;
+      }
       try
       {
         result = kinematics_loader_->createUniqueInstance(solver);


mechwiz:
Yeah it looks like both PRs are needed to fix the issue completely

sjahr:
@mechwiz We had some issues with CI that are fixed by the latest commit on main. If you rebase this PR on the latest version of main, CI should pass!

mechwiz:
done :)

sjahr (approved):

Issue #1885, opened 2023-01-24
Servo: Returns to last sent position when unpaused


HX2003:
If joint positions move by some other method when servo is paused or commands are stale, upon receiving new commands that are zero it returns to last sent position. Hence, there is a sudden movement; I don't think that is supposed to be the intended behaviour?
https://github.com/ros-planning/moveit2/blob/e2e0003c0e170a05747c699da05258382a338edf/moveit_ros/moveit_servo/src/servo_calcs.cpp#L452-L453

AndyZe:
Thanks for tracking down one of the lines involved. I'll file a PR to fix this soon. Can you test it for me?

henningkayser:
Is this issue related to https://github.com/ros-planning/moveit2/issues/1036?

AndyZe:
I'm starting to look at this. Servo has these flags, `latest_twist_cmd_is_nonzero_` and `latest_joint_cmd_is_nonzero_`. The original thinking was, Servo should skip publishing commands if the user is requesting zero velocity. This was done to reduce spammy messages to ros_control.

I'm now thinking, those flags should just be dropped. No harm done if Servo publishes zero-velocity commands always.

AndyZe:
@HX2003 can you please test #1904?

HX2003:
@AndyZe I can confirm that the proposed fix works as intended. Additionally, can you expose the paused/unpaused status as a topic, I have multiple nodes that need to know whether servo is active.

AndyZe:
Isn't there a topic for this already, something like `/servo_server/status`?

Thanks for testing it so quickly!

HX2003:
There is a topic similar to it, but it does not reflect anything about pause status.
https://github.com/ros-planning/moveit2/blob/2fe4b30266fff91580234cbe0f05c6d1e1cefe8e/moveit_ros/moveit_servo/include/moveit_servo/status_codes.h

AndyZe:
OK. Well, I probably won't have time to make that PR in the next couple weeks. If you can make, I'll review, though.

Feel free to open a new issue requesting the feature, too.

HX2003:
Would it just involve publishing the paused status here as a separate topic? In that case I should be able to make a PR. And I should close this issue soon.
https://github.com/ros-planning/moveit2/blob/2fe4b30266fff91580234cbe0f05c6d1e1cefe8e/moveit_ros/moveit_servo/src/servo_calcs.cpp#L1232-L1235

[...]

PR #1904, opened 2023-01-27
Servo: always keep publishing zero-velocity cmds. Eliminate possibility of joint jumps

### LINKED ISSUE #1036: Joint jump when Servo is activated
Hello everyone,

I am facing issues, that seem to be related to a call to getCurrentJointValues(). I am planning a path to a certain point near a collision object, and then doing servo for the last part towards the collision object, after which is has to be attached to the robot. Between executing the path and starting the servo, I try to rotate the end-effector to 0, like so:
```
std::vector<double> joint_values = move_group_interface.getCurrentJointValues(); 
joint_values.at(5) = 0.0;
move_group_interface.setJointValueTarget(joint_values);
```
After this, I also add the collision object and the end-effector of the robot to the ACM, and then start servoing. When I do this sequence without calling getCurrentJointValues(), everything works fine. However, as soon as I call it, I get weird "jumps" at the start of the servoing. After that, the servo does proceed "normally" (as far as possible, it does execute the commands that are given). See the video for some visualisation. However, it seems like this is also related to a "Didn't received robot state (joint angles) with recent timestamp within 1 seconds." error, that I also only get when rotating the end-effector, but I am not _completely_ sure whether that is indeed the case. 

I have no clue how to deal with this, so any help is welcome. Thanks a lot in advance!
Sabine

### Your environment
* ROS Distro: ROS2 Foxy
* OS Version: Ubuntu  20.04
* Source or Binary build? Binary

### Steps to reproduce
Not sure if this gives the same behaviour for everyone, but:
- create a PoseStamped with a tf2 frame as frame_id and orientation as `quaternion.setRPY(0.0, 0.0, 0.0)`
- Plan a path using setPoseTarget;
- Execute
- Call getCurrentJointValues()
- Start servoing

### Expected behaviour
The servo motion starts at the same location as where the execution of the planned path ended.

### Actual behaviour
The robot jumps to a (random) location elsewhere

https://user-images.githubusercontent.com/54582035/151384215-b4009324-cc19-48c9-a202-aa39f5d54f0f.mp4

AndyZe:
### Description

Likely fixes #1036, #1885 

There are a lot of simplifications here:

- Always keep publishing zero-velocity commands. Previously, we kept track of the number of zero-velocity commands published to try to  prevent spammy messages. I think the extra complexity wasn't worth it.
- If there is not a valid command, don't jump back to the previously-commanded state. This was causing issue #1885. Instead, come to a gradual halt based on the robot's current position.

This was tested by the bug reporter here:  https://github.com/ros-planning/moveit2/issues/1885#issuecomment-1406923402

### DIFF
--- a/moveit_ros/moveit_servo/config/panda_simulated_config.yaml
+++ b/moveit_ros/moveit_servo/config/panda_simulated_config.yaml
@@ -47,9 +47,6 @@ robot_link_command_frame:  panda_link0  # commands must be given in the frame of
 
 ## Stopping behaviour
 incoming_command_timeout:  0.1  # Stop servoing if X seconds elapse without a new command
-# If 0, republish commands forever even if the robot is stationary. Otherwise, specify num. to publish.
-# Important because ROS may drop some messages and we need the robot to halt reliably.
-num_outgoing_halt_msgs_to_publish: 4
 
 ## Configure handling of singularities and joint limits
 lower_singularity_threshold:  17.0  # Start decelerating when the condition number hits this (close to singularity)
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
@@ -305,14 +305,12 @@ protected:
   // Main tracking / result publisher loop
   std::thread thread_;
   bool stop_requested_;
-  std::atomic<bool> done_stopping_;
 
   // Status
   StatusCode status_ = StatusCode::NO_WARNING;
   std::atomic<bool> paused_;
   bool twist_command_is_stale_ = false;
   bool joint_command_is_stale_ = false;
-  bool ok_to_publish_ = false;
   double collision_velocity_scale_ = 1.0;
 
   // Use ArrayXd type to enable more coefficient-wise operations
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
@@ -264,10 +264,6 @@ protected:
   // Pointer to the collision environment
   planning_scene_monitor::PlanningSceneMonitorPtr planning_scene_monitor_;
 
-  // Track the number of cycles during which motion has not occurred.
-  // Will avoid re-publishing zero velocities endlessly.
-  int zero_velocity_count_ = 0;
-
   // Flag for staying inactive while there are no incoming commands
   bool wait_for_servo_commands_ = true;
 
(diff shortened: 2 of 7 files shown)


moveit_ros/moveit_servo/config/panda_simulated_config.yaml
```diff
@@ -47,9 +47,6 @@ robot_link_command_frame:  panda_link0  # commands must be given in the frame of
 
 ## Stopping behaviour
 incoming_command_timeout:  0.1  # Stop servoing if X seconds elapse without a new command
-# If 0, republish commands forever even if the robot is stationary. Otherwise, specify num. to publish.
-# Important because ROS may drop some messages and we need the robot to halt reliably.
-num_outgoing_halt_msgs_to_publish: 4
```
AndyZe, line 52: Drop this functionality and always just keep republishing, even if the robot is stationary.

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
```diff
@@ -264,21 +264,12 @@ class ServoCalcs
...
   // Flag for staying inactive while there are no incoming commands
   bool wait_for_servo_commands_ = true;
 
   // Flag saying if the filters were updated during the timer callback
   bool updated_filters_ = false;
 
-  // Nonzero status flags
-  bool have_nonzero_twist_stamped_ = false;
-  bool have_nonzero_joint_command_ = false;
-  bool have_nonzero_command_ = false;
```
AndyZe, line 280: ^No longer track this. Always accept all incoming commands.

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
```diff
@@ -314,14 +305,12 @@ class ServoCalcs
   // Main tracking / result publisher loop
   std::thread thread_;
   bool stop_requested_;
-  std::atomic<bool> done_stopping_;
```
AndyZe, line 317: This variable was only used when incoming velocity commands were all zero. We don't bother tracking that anymore.

henningkayser (approved):
This is a great simplification, and I see why this would help with the joint jump. Only question is, are we now completely relying on ros2_control to correct for drift when we are halted?

AndyZe:
> Only question is, are we now completely relying on ros2_control to correct for drift when we are halted?

Let's merge it and find out in the long term. Somebody could always pause the Servo server if that's an issue for them.

sea-bass:
@AndyZe turns out this PR actually broke things for Gazebo with MoveIt Studio because of https://github.com/ros-planning/moveit2/issues/1857.

Now that we're publishing zero-velocity commands constantly, our simulated arm now sags indefinitely.

Is there a possibility of reverting this, but maybe making it an option via parameter?

AndyZe:
I don't think we should revert it, this was a nice cleanup. It would be a lot cleaner on my end if you could use the `pause` and `unpause` services. Is that feasible?

https://github.com/ros-planning/moveit2/blob/a1fce2d077f16673d8dd540f5b1072b5bc0e56f5/moveit_ros/moveit_servo/src/servo_node.cpp#L66

PR #1546, opened 2022-09-01
Python Bindings - moveit_py

peterdavidfagan:
### Description

This pull request adds a new Python library `moveit` which can be used to interface with MoveIt 2. Below is a basic summary of the initial feature set of the library: 

* **moveit:** 
    + **moveit.core:** a python module containing implementations of moveit_core objects.
        + **moveit.core.collision_detection:**
            + AllowedCollisionMatrix
            + CollisionRequest
            + CollisionResult
        + **moveit.core.kinematic_constraints:**
            + KinematicConstraints
        + **moveit.core.planning_interface:**
           + MotionPlanResponse
        + **moveit.core.planning_scene:**
            + PlanningScene
        + **moveit.core.robot_model:**
            + JointModelGroup
            + RobotModel
        + **moveit.core.robot_state:**
            + RobotState
        + **moveit.core.robot_trajectory**
            + RobotTrajectory
        + **moveit.core.transforms**
            + Transforms

    + **moveit.planning:** a python module that exposes the planning functionalities of moveit_cpp
        + MoveItCpp  
        + MultiPipelinePlanRequestParameters 
        + PlanningComponent   
        + PlanningSceneMonitor
        + SinglePlanRequestParameters
 
    + **moveit.servo_client:** a python module that defines a client for interacting with moveit_servo.
        + TeleopDevice
            + PS4DualShock
            + OcculusQuest2 (to be added)

Note: this is an initial release, in future releases further functionalities will be added (e.g. moveit task constructor; task planning tools; interfaces for machine learning experiments). 

### Related Pull Requests
As this is a new library we also have associated API documentation and tutorials that should be merged around the same time as the initial library release. Below is a list of the currently open pull requests: 

API Documentation:  https://github.com/ros-planning/moveit2_tutorials/pull/504
Robot State and Model Tutorial: TBC
Motion Planning Tutorial: https://github.com/ros-planning/moveit2_tutorials/pull/565
Jupyter Notebook Tutorial: https://github.com/ros-planning/moveit2_tutorials/pull/564
Robot Learning Tutorial (Simplified application of [QT-OPT](https://arxiv.org/abs/1806.10293) or similar ): TBC (likely to be moved until after the initial release)

Issue #1262, opened 2022-05-19
[MSA] Port `moveit_setup_simulation`


DLu:
The `moveit_setup_simulation` package is currently ignored, and does not compile.

In ROS 1, this step would attempt to modify your original URDF file to add the needed attributes for simulation and write it to `config/gazebo_[ROBOT_NAME].urdf` if the original location could not be modified. It would generate
 * `config/gazebo_controllers.yaml` - Configuration of Gazebo controllers
 * `launch/gazebo.launch` - Gazebo launch file which also launches ros_controllers and sends robot urdf to param server, then using gazebo_ros pkg the robot is spawned to Gazebo
 * `launch/demo_gazebo.launch` - Run a demo of MoveIt with Gazebo and Rviz
 * `config/gazebo_[ROBOT_NAME].urdf` (sometimes)

The edits made to the urdf include:
 * Adding `<inertial>` tags to `<collision>` elements of `<link>` elements.
 * Adding `<transmission>` tags for each `<joint>` element.
 * Adding the `<gazebo>` tag (globally) and specifying `<plugin>` elements inside.

Which of these steps need to be done for ROS 2 ~~Ignition~~ Gazebo? Is it within the scope of this repo? Could it just be made into its own Setup Assistant using this framework?

DLu:
@abake48 I made some boilerplate for generating the simulation xacro here: https://github.com/DLu/moveit2/tree/feature/sim_xacro/moveit_setup_assistant/moveit_setup_simulation

abake48:
Thank you! I'll be spending all day tomorrow working on this so this can get released!

DLu:
@ahcorde @chapulina  Can you help us define what is needed in a URDF for simulation? Assuming it already has `ros2_control` tags, it looks like we need to replace the `gazebo_ros_control` plugin with [`libign_ros2_control-system.so`](https://github.com/ros-controls/gz_ros2_control#add-the-ign_ros2_control-plugin). I'm guessing we don't need transmissions anymore?

PR #1916, opened 2023-02-02
Delete packages which are not being built

ChrisThrasher:
### Description

I deleted all packages containing `COLCON_IGNORE` since by the fact that they're not being built, they're not being used. We can resurrect this code later if we change our mind.


tylerjw (approved):

sjahr (approved):
Good idea :+1: The code in these packages is probably 3 years outdated by now because I am not sure if the unported moveit1 packages got synchronized. If :man_shrugging:  any of these packages will be ported to moveit2 in the future, the first step should be to copy the latest version of the package from moveit1 anyways.

henningkayser (changes requested):
Did you investigate what you are removing? Originally, these packages have been ignored due to migration in order to preserve the commit history, which is really useful for understanding the context for a lot of the code. I'm ok with removing `moveit_commander` and `moveit_experimental` since the first one is being made obsolete by the new python bindings (#1546 removes these parts as well), and I can't recall `moveit_experimental` ever being of much use. Trajopt should definitely not be removed, it's way too recent and it's still being worked on (https://github.com/ros-planning/moveit/pull/3235), and we should rather find a way to complete migration. I'm unsure about the MSA parts, @DLu (author) might have some thoughts if it's worth migrating or if it should be reimplemented from scratch. In general, I think it's very important in OSS to preserve authorship and commit history. Ignored packages are very little dead weight and trivial to sync with changes from ROS 1.

Edit: I just noticed that there is even an open issue to complete the MSA simulation work https://github.com/ros-planning/moveit2/issues/1262, so I would suggest not removing the code but rather try to find some time to make it work.

ChrisThrasher:
> I think it's very important in OSS to preserve authorship and commit history.

What is not being preserved? Anyone can still look through the commit history to see who added this now-unused code.

ChrisThrasher:
I put trajopts and moveit_setup_assistant back so now this PR only removes moveit_experimental and moveit_commander.

henningkayser:
> What is not being preserved? Anyone can still look through the commit history to see who added this now-unused code. That's the whole point of version control.

The last version will not list the commits with git blame, and the person that copied back the code or reverted the removal commit will be the new author. https://moveit.ros.org/documentation/contributing/maintainer_pr/#the-latest-branch-preserves-all-original-commits

henningkayser (approved):

PeterMitrano:
Very happy to see `moveit_commander` being deleted... makes more spaces for #1546 to be the one and only option!!! :tada:

PR #1918, opened 2023-02-03
TEM: Fix segfault when no manager is specified

JafarAbdi:
### Description

This fixes a bug when no controller_manager is specified and there are no declared class found.

We only create a node when `!controller.empty()` but we access it always later in the code which is causing a segfault.

Same for the executor, I added a check before accessing it otherwise it segfaults for the same reason

### Checklist

### DIFF
--- a/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
+++ b/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
@@ -74,10 +74,10 @@ TrajectoryExecutionManager::TrajectoryExecutionManager(const rclcpp::Node::Share
 TrajectoryExecutionManager::~TrajectoryExecutionManager()
 {
   stopExecution(true);
-  private_executor_->cancel();
+  if (private_executor_)
+    private_executor_->cancel();
   if (private_executor_thread_.joinable())
     private_executor_thread_.join();
-  private_executor_.reset();
 }
 
 void TrajectoryExecutionManager::initialize()
@@ -130,6 +130,7 @@ void TrajectoryExecutionManager::initialize()
         RCLCPP_FATAL(LOGGER, "Parameter '~moveit_controller_manager' not specified. This is needed to "
                              "identify the plugin to use for interacting with controllers. No paths can "
                              "be executed.");
+        return;
       }
     }
 


AndyZe (approved):

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -74,10 +74,10 @@ TrajectoryExecutionManager::TrajectoryExecutionManager(const rclcpp::Node::Share
 TrajectoryExecutionManager::~TrajectoryExecutionManager()
 {
   stopExecution(true);
-  private_executor_->cancel();
+  if (private_executor_)
+    private_executor_->cancel();
   if (private_executor_thread_.joinable())
     private_executor_thread_.join();
-  private_executor_.reset();
```
sjahr, line 80: @JafarAbdi Why are you removing this?

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -74,10 +74,10 @@ TrajectoryExecutionManager::TrajectoryExecutionManager(const rclcpp::Node::Share
 TrajectoryExecutionManager::~TrajectoryExecutionManager()
 {
   stopExecution(true);
-  private_executor_->cancel();
+  if (private_executor_)
+    private_executor_->cancel();
   if (private_executor_thread_.joinable())
     private_executor_thread_.join();
```
sjahr, line 80:
```suggestion
  if (private_executor_)
  {
    private_executor_->cancel();
  }
  if (private_executor_thread_.joinable())
  {
    private_executor_thread_.join();
  }
```

sjahr (commented):
Besides a small question, this looks good to me. Feel free to ignore the nitpick suggestion if you don't like it

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
JafarAbdi: It will get called by the destructor

sjahr (approved):

PR #1931, opened 2023-02-06
Install headers within additional directory

ChrisThrasher:
### Description

Continuing from #1915. There are about 7 more packages left to convert which I will do in the last PR of this series.

### DIFF
--- a/moveit_ros/benchmarks/CMakeLists.txt
+++ b/moveit_ros/benchmarks/CMakeLists.txt
@@ -59,7 +59,7 @@ install(
   LIBRARY DESTINATION lib
   ARCHIVE DESTINATION lib
   RUNTIME DESTINATION bin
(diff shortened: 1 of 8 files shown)


tylerjw (approved):

AndyZe:
In the future, I think it would be good to combine smaller PR's that have a common theme into one large PR

ChrisThrasher:
> I think it would be good to combine smaller PR'

I did. I could have made one PR per package. One massive PR would have delayed merging this sooner thus making conflicts more likely plus it would have put more burden on reviewers by forcing them to review a massive diff all at once instead of the 4 smaller PRs I ended up submitting. Determining PR size is hard and we all have own preferences.

AndyZe:
Well, my preference is toward larger PR's so the git history is less noisy.

PR #1946, opened 2023-02-13
[WIP] Cartesian Tool Path Planning

sea-bass:
### Description

This reimplements #377 with the suggestions from review and continues the work to make it a complete feature.

### Checklist

### DIFF
--- a/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/detail/ompl_constraints.h
+++ b/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/detail/ompl_constraints.h
@@ -481,16 +481,17 @@ struct EigenIsometry3dWrapper
 
   bool operator==(const EigenIsometry3dWrapper& other) const
   {
-    return tform_.isApprox(other.tform_, 0.0001);
+    return tform_.isApprox(other.tform_, EQUALITY_CONSTRAINT_THRESHOLD);
   }
 
   bool operator!=(const EigenIsometry3dWrapper& other) const
   {
-    return !(tform_.isApprox(other.tform_, 0.0001));
+    return !(tform_.isApprox(other.tform_, EQUALITY_CONSTRAINT_THRESHOLD));
   }
 
   size_t path_idx_ = 0;
   Eigen::Isometry3d tform_;
+  static constexpr double EQUALITY_CONSTRAINT_THRESHOLD{ 0.001 };
 };
 
 /**
--- a/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/detail/ompl_constraints.h
+++ b/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/detail/ompl_constraints.h
@@ -45,8 +45,8 @@
 #include <moveit/ompl_interface/detail/threadsafe_state_storage.h>
 #include <moveit/robot_model/robot_model.h>
 #include <moveit/macros/class_forward.h>
-#include <moveit_msgs/msg/cartesian_trajectory.hpp>
 #include <moveit_msgs/msg/constraints.hpp>
+#include <moveit_msgs/msg/path_constraint.hpp>
 
 namespace ompl_interface
 {
@@ -504,23 +504,30 @@ double arcLength(const Eigen::Isometry3d& pose1, const Eigen::Isometry3d& pose2)
  */
 Eigen::Isometry3d interpolatePose(const Eigen::Isometry3d& pose1, const Eigen::Isometry3d& pose2, double step);
 
-/** \brief A path constraint for the EEF useful for sanding, painting, drawing, welding, etc
+/** \brief A Cartesian path constraint a specified link, useful for appliations such as sanding, painting, drawing,
+ * welding, etc.
  * */
 class ToolPathConstraint : public BaseConstraint
 {
 public:
   ToolPathConstraint(const moveit::core::RobotModelConstPtr& robot_model, const std::string& group,
-                     unsigned int num_dofs, const unsigned int num_cons = 6, double step = 0.01);
+                     unsigned int num_dofs, const unsigned int num_cons = 6);
 
   void parseConstraintMsg(const moveit_msgs::msg::Constraints& constraints) override;
 
-  void setPath(const moveit_msgs::msg::CartesianTrajectory& msg);
+  void setPath(const moveit_msgs::msg::PathConstraint& msg);
 
   void function(const Eigen::Ref<const Eigen::VectorXd>& joint_values, Eigen::Ref<Eigen::VectorXd> out) const override;
 
 private:
+  // Nearest-neighbors structure
   mutable ompl::NearestNeighborsGNAT<EigenIsometry3dWrapper> pose_nn_;  // store poses
-  const double step_;                                                   // step size between each waypoint
+
+  // Path representation
+  std::vector<EigenIsometry3dWrapper> path_;
+
+  // Interpolation distance for path
+  double interp_distance_;
 };
 
 }  // namespace ompl_interface
--- a/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/detail/ompl_constraints.h
+++ b/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/detail/ompl_constraints.h
@@ -488,6 +488,14 @@ struct EigenIsometry3dWrapper
 };
 
 /**
+ * @brief Find arc length between two quaternions
+ * @param pose1 - first pose
+ * @param pose2 - second pose
+ * @return Arc length
+ */
+double arcLength(const Eigen::Isometry3d& pose1, const Eigen::Isometry3d& pose2);
+
+/**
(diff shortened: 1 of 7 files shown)


moveit_planners/ompl/ompl_interface/src/detail/ompl_constraints.cpp
```diff
@@ -408,4 +408,106 @@ ompl::base::ConstraintPtr createOMPLConstraints(const moveit::core::RobotModelCo
...
+/******************************************
+ * Toolpath Constraint
+ * ****************************************/
+ToolPathConstraint::ToolPathConstraint(const moveit::core::RobotModelConstPtr& robot_model, const std::string& group,
+                                       unsigned int num_dofs, const unsigned int num_cons, double step)
+  : BaseConstraint(robot_model, group, num_dofs, num_cons), step_(step)
+{
+  pose_nn_.setDistanceFunction([this](const EigenIsometry3dWrapper& pose1, const EigenIsometry3dWrapper& pose2) {
+    double translation_distance = (pose1.tform_.translation() - pose2.tform_.translation()).norm();
+    double rotation_distance = 0.0;  // TODO
```
mamoll, line 421: See https://github.com/ompl/ompl/blob/main/src/ompl/base/spaces/src/SO3StateSpace.cpp#L250-L277
There may be equivalent code in Eigen, tf2, ...

sea-bass: It's just the arc cosine of the quaternion dot product, so easy enough to do with Eigen. Thanks!

sea-bass:
The code is now at the point where I can pass in a path constraint, it gets instantiated, and we actually get a plan. 

However, the plan right now either fails because of tight bounds or succeeds but doesn't honor the tool path at all.

There are lots of OMPL interface nuances that are still mysterious to me, and I may need to some help with this to head in the right direction vs. just doing things with trial and error.

mamoll:
@zkingston do you have time to take a look?

moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/detail/ompl_constraints.h
```diff
@@ -459,4 +466,72 @@ inline Eigen::Matrix3d angularVelocityToAngleAxis(const double& angle, const Eig
...
+struct EigenIsometry3dWrapper
+{
+  EigenIsometry3dWrapper(){};
+  EigenIsometry3dWrapper(const Eigen::Isometry3d& tform) : tform_{ tform }
+  {
+  }
+
+  bool operator==(const EigenIsometry3dWrapper& other) const
+  {
+    return tform_.isApprox(other.tform_);
```
zkingston, line 483: Should this use the default precision value?

moveit_planners/ompl/ompl_interface/src/detail/ompl_constraints.cpp
```diff
@@ -323,6 +323,134 @@ Eigen::MatrixXd OrientationConstraint::calcErrorJacobian(const Eigen::Ref<const
...
+
+  // Extract the interpolation distance
+  if (path_constraint.interp_distance > 0.0)
+  {
+    interp_distance_ = path_constraint.interp_distance;
+  }
+  else
+  {
+    RCLCPP_WARN_STREAM(LOGGER, "Cannot use zero interpolation distance. Defaulting to 0.01.");
+    interp_distance_ = 0.01;
```
zkingston, line 380: Given that you are discretizing the polyline at some resolution `interp_distance_`, the allowable constraint error has to be greater than this interpolation distance so that it would appear as a continuous function rather than a disjoint collection of points. Where is the constraint tolerance set in this example?

moveit_planners/ompl/ompl_interface/src/detail/ompl_constraints.cpp
```diff
@@ -323,6 +323,134 @@ Eigen::MatrixXd OrientationConstraint::calcErrorJacobian(const Eigen::Ref<const
...
+  out[4] = eef_aa_vec(1) - eef_nearest_aa_vec(1);
+  out[5] = eef_aa_vec(2) - eef_nearest_aa_vec(2);
+  return out;
+}
+
+double arcLength(const Eigen::Isometry3d& pose1, const Eigen::Isometry3d& pose2)
+{
+  const auto q1 = Eigen::Quaterniond(pose1.rotation());
+  const auto q2 = Eigen::Quaterniond(pose2.rotation());
+  return std::acos(std::abs(std::clamp(q1.dot(q2), -1.0, 1.0)));
```
zkingston, line 439: Can't you use `angularDistance` from Eigen here?

zkingston (commented):
Could you elaborate more on the failure case? e.g., when you say the tolerances are too tight and planning fails. I took a quick look through the code.

moveit_planners/ompl/ompl_interface/src/detail/ompl_constraints.cpp
sea-bass: I'm not sure what this is. Do you know off the top of your head what the OMPL name / function call would be so I can search for it?

sea-bass:
> Could you elaborate more on the failure case? e.g., when you say the tolerances are too tight and planning fails. I took a quick look through the code.

Thanks for looking at this!

I don't know if I have anything super specific rather than generally trying to understand how this all works. Some pointed questions:

* What does the `bounds_` member correspond to for a constraint like this. Should it be the bounds of the workspace? The bounds for each of the elements in the error function? Something else?
* Does implementing `function()` vs `calcError()` make a difference? What is the right way of creating something that constrains a search space to be within a polyline like this?
* Should the error be a scalar or a vector?

moveit_planners/ompl/ompl_interface/src/detail/ompl_constraints.cpp
zkingston: In [`ompl::base::Constraint`](https://ompl.kavrakilab.org/classompl_1_1base_1_1Constraint.html) this is the `setTolerance()` function - I'm not sure which solver MoveIt defaults to, but it's whatever is considered the cutoff for saying the constraint is satisfied.

sea-bass: Aha, thanks! Yeah, it's just using the default of `1e-4` at the moment, so I can try bring that up.

zkingston:
1. I think MoveIt's box constraint bounds here define the region in which EE positions are considered valid (I am not 100% on this). Since your constraint is a sequence of smaller position constraints (maybe you could think of this as a collection of MoveIt's box / position / equality constraints tiled together so that each error region overlaps with each other...), the bounds here you inherit from `BaseConstraint` probably don't do what you want, you might need to overload the Jacobian.
2. `calcError` is a MoveIt thing, `function` is the OMPL constraint version. I don't know enough about the MoveIt implementation to say.
3. Vector should work, but if you can get the gradient for scalar that should be fine as well.

sea-bass:
Interesting... so is there a way in OMPL to do a "`ConstraintUnion`" rather than `ConstraintIntersection` to tile these together?

zkingston:
I think an idea similar to what you have currently, but using the NN to store the centers of each constraint region (which tile the path similar to what you have currently), and then evaluating the error / jacobian to the closest constraint may work.

There isn't a way in OMPL to do a "constraint union" like this - I'm not sure if it even is a well-formed concept.

sea-bass:
Thank you for all this input -- these are helpful things to try out!

sea-bass:
I tried going down this path and have something that runs... but again fails to plan with this error at the end. Will dig more when I find more time / energy.

```
[ERROR] [1676674916.342760349] [ompl]: ./src/ompl/base/src/SpaceInformation.cpp:468 - State samplers generate constraint unsatisfying states.
```

moveit_planners/ompl/ompl_interface/src/detail/ompl_constraints.cpp
```diff
@@ -323,6 +323,166 @@ Eigen::MatrixXd OrientationConstraint::calcErrorJacobian(const Eigen::Ref<const
...
+void ToolPathConstraint::jacobian(const Eigen::Ref<const Eigen::VectorXd>& joint_values,
+                                  Eigen::Ref<Eigen::MatrixXd> out) const
+{
+  const Eigen::Isometry3d eef_pose = forwardKinematics(joint_values);
+  const auto eef_nearest = pose_nn_.nearest(EigenIsometry3dWrapper(eef_pose));
+  // std::cout << "Jacobian, nearest: " << eef_nearest.path_idx_ << std::endl;
+
+  const auto bc = box_constraints_.at(eef_nearest.path_idx_);
+  bc->jacobian(joint_values, out);
+}
```
sea-bass, line 465: These are the "magic" functions that map the EEF pose to the nearest box constraint for that path segment and compute the standard function/Jacobian for that box constraints.

I think this is close, except that I think the state sampler needs to be configured.

tylerjw:
Closing as this seems to be abandoned. Feel free to re-open if you take it back up.

PR #1955, opened 2023-02-17
Doxygen tag

henningkayser:
I'm working on fixing the dead links in the tutorials, and making it easier to reference the API using [doxylink](https://github.com/sphinx-contrib/doxylink). We need to install a tag file for that to work. Also, the QT property is breaking the references because of the identical naming, with the attached fixed everything works fine.

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -2081,7 +2081,7 @@ TAGFILES               =
 # tag file that is based on the input files it reads. See section "Linking to
 # external documentation" for more information about the usage of tag files.
 
-GENERATE_TAGFILE       = MoveIt.tag
+GENERATE_TAGFILE       = $(DOXYGEN_OUTPUT_DIRECTORY)/MoveIt.tag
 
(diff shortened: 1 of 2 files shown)


tylerjw (approved):
If this passes, lets merge this and backport to humble?

tylerjw:
```
  --- stderr: moveit_msgs
  CMake Error at /opt/ros/rolling/share/sensor_msgs/cmake/ament_cmake_export_dependencies-extras.cmake:21 (find_package):
    By not providing "Findservice_msgs.cmake" in CMAKE_MODULE_PATH this project
    has asked CMake to find a package configuration file provided by
    "service_msgs", but CMake did not find one.
  
    Could not find a package configuration file provided by "service_msgs" with
    any of the following names:
  
      service_msgsConfig.cmake
      service_msgs-config.cmake
```

is there something broken about moveit_msgs now?

tylerjw:
I'm going to merge this as the failure has nothing to do with this PR and should be addressed separately.

PR #1963, opened 2023-02-21
Fix Ruckig termination condition

### LINKED ISSUE #1960: Ruckig trajectory smoothing plugin: duration is extended unnecessarily
### Description

#1719 seems to have introduced a bug, as commented here:  https://github.com/ros-planning/moveit2/issues/1439#issuecomment-1374005027

AndyZe:
### Description

Previously I didn't understand the difference between `ruckig::Finished` and `ruckig::Working` so I focused only on `ruckig::Finished`. There's not much documentation but here it is:  https://github.com/pantor/ruckig/blob/master/include/ruckig/result.hpp

This corrects that error and adds a comment. Fixes #1960 

This should significantly speed up Ruckig-smoothed trajectories. I'm seeing speedups in the range of 50% although it depends on the specific robot configuration.

Fixes #1960. It is a port of this MoveIt1 PR:  https://github.com/ros-planning/moveit/pull/3348

@swiz23 @mechwiz @nbbrooks

Plots. The performance is close to optimal but not quite optimal. This makes sense because the iterative process for multiple waypoints involves extending durations in 10% increments -- so it could be up to 10% non-optimal.

J1, vel limit=2.18 rad/s, accel limit=3.75 rad/s^2, jerk limit=100
j1

J3, vel limit=2.18, accel limit=2.5, jerk limit=100
j3

### DIFF
--- a/moveit_core/trajectory_processing/test/test_ruckig_traj_smoothing.cpp
+++ b/moveit_core/trajectory_processing/test/test_ruckig_traj_smoothing.cpp
@@ -63,6 +63,8 @@ TEST_F(RuckigTests, basic_trajectory)
 {
   moveit::core::RobotState robot_state(robot_model_);
   robot_state.setToDefaultValues();
+  robot_state.zeroVelocities();
+  robot_state.zeroAccelerations();
   // First waypoint is default joint positions
   trajectory_->addSuffixWayPoint(robot_state, DEFAULT_TIMESTEP);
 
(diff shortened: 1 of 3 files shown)


swiz23:
Can you link to the issue that this is fixing?

AndyZe:
It might fix #1960 but I'm not sure. That could also be a different bug.

moveit_core/trajectory_processing/test/test_ruckig_traj_smoothing.cpp
```diff
@@ -106,7 +106,7 @@ TEST_F(RuckigTests, basic_trajectory_with_custom_limits)
 TEST_F(RuckigTests, trajectory_duration)
 {
   // Compare against the OJET online trajectory generator: https://www.trajectorygenerator.com/ojet-online/
-  const double ideal_duration = 0.21025;
+  const double ideal_duration = 0.310;
```
AndyZe, line 109: I was looking at the wrong joint limits -- it is reading from the robot model, not joint_limits.yaml for this test.

AndyZe:
@swiz23 I got rid of `runRuckigInBatches()` for now so this should definitely fix #1960. I'll put some plots in the description now, too.

henningkayser (approved):

Issue #1815, opened 2022-12-19
Remove "blocking" semantics from MoveItCpp::execute()


henningkayser:
Non-blocking trajectory execution is currently not supported anymore, it has been removed with https://github.com/ros-planning/moveit2/commit/24c18f68a80eeac43627af4097a72703a32d42c5. Calling `execute()` with `blocking=false` will not do anything anymore, so the argument should be removed.

sjahr:
Done by #1984

PR #1984, opened 2023-03-02
Deprecating MoveItCpp::Execute() use of bool blocking

CursedRock17:
### Description

In this pull request I targetted issue #1815 which was meant to remove the `bool blocking` since we had shifted the non-blocking trajectory execution into a different function. Since I saw that bool blocking had no further use and it was possible to make deprecation warnings I adjusted all cases of the MoveItCpp::Execute() function to work without the bool blocking.

### Checklist


moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
```diff
@@ -188,7 +188,7 @@ void MotionPlanningFrame::computePlanButtonClicked()
 
   if (success)
   {
-    ui_->execute_button->setEnabled(true);
+    ui_->ebxecute_button->setEnabled(true);
```
henningkayser, line 191: this looks like an accident, please fix

henningkayser (commented):
Thanks! Please rebase onto main, CI should be fixed now. The changes look correct to me, I'll give this another pass once CI succeeds.

CursedRock17:
It should be rebased, is it just the CI acting up that causes some of these tests to fail, because it didn't seem like anything was missing that would've directly caused an error?

CursedRock17:
It should be rebased and the CI passes the workflow tests on my forked project, so these should pass.

CursedRock17:
Is there a valid reason that the CI on the ccov debug version cannot find service_msgsConfig, but the other CI's in release version pass with no problem. None of this code should affect external CMake packages, but I know that the general CI was having some hiccups recently. For reference this is the error in the industrial_ci: 
- ` 
      Could not find a package configuration file provided by "service_msgs" with
      any of the following names:
      service_msgsConfig.cmake
      service_msgs-config.cmake
`

sjahr:
> Is there a valid reason that the CI on the ccov debug version cannot find service_msgsConfig, but the other CI's in release version pass with no problem. None of this code should affect external CMake packages, but I know that the general CI was having some hiccups recently. For reference this is the error in the industrial_ci:
> 
> * ` Could not find a package configuration file provided by "service_msgs" with any of the following names: service_msgsConfig.cmake service_msgs-config.cmake`

This is not caused by this PR and should have been fixed at the end of last week. Let's update this PR a second time

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
```diff
@@ -165,18 +165,17 @@ class MoveItCpp
...
-   *  \param [in] blocking If blocking, wait for the trajectory to execute before continuing. Defaults to `true`.
    *  \param [in] controllers An optional list of ros2_controllers to execute with. If none, MoveIt will attempt to find
    * a controller. The exact behavior of finding a controller depends on which MoveItControllerManager plugin is active.
    * \return moveit_controller_manager::ExecutionStatus::SUCCEEDED if successful
    */
   [[deprecated(
       "MoveItCpp::execute() no longer requires a group_name parameter")]] moveit_controller_manager::ExecutionStatus
   execute(const std::string& group_name, const robot_trajectory::RobotTrajectoryPtr& robot_trajectory,
-          bool blocking = true, const std::vector<std::string>& controllers = std::vector<std::string>());
+          const std::vector<std::string>& controllers = std::vector<std::string>());
```
henningkayser, line 175: This function is deprecated, just remove it

henningkayser (commented):
I agree that with these changes overall, but it would be sensible to actually deprecate the functions first before removing them (what this PR does).

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
CursedRock17: To be clear replace the old deprecation of the group_name parameter, to now just three parameters with`bool blocking` as one of them. Then move forward with the true two parameter version of execute.

henningkayser (approved):

PR #1986, opened 2023-03-03
Fix collision checking in VisibilityConstraint

schornakj:
### Description

Previously, planning with the VisibilityConstraint could cause segfaults because modifying collision objects in the collision environment is not thread-safe. The planning threads would interfere with each other by trying to add and remove the visibility cone collision object out of sync.

This PR fixes that by creating a local collision environment within `VisibilityConstraint::decide`.

I also modified how VisibilityConstraints are evaluated so that if a negative or zero `target_radius` value is used but valid `max_range_angle` and `max_view_angle` are provided, then the visibility cone check is skipped and the range and view angle checks are performed as usual. Previously setting an invalid `target_radius` value would cause the entire constraint evaluation to be skipped.

While I was at it I renamed some of the variables used for transforms to be more verbose and easier to follow.

Note: I'm keeping this as a draft for now since I'm still testing, but I'd be happy to get feedback on what I have so far.

### Checklist

### DIFF
--- a/moveit_core/kinematic_constraints/src/kinematic_constraint.cpp
+++ b/moveit_core/kinematic_constraints/src/kinematic_constraint.cpp
@@ -1073,11 +1073,6 @@ void VisibilityConstraint::getMarkers(const moveit::core::RobotState& state,
 
 ConstraintEvaluationResult VisibilityConstraint::decide(const moveit::core::RobotState& state, bool verbose) const
 {
-  if (target_radius_ <= std::numeric_limits<double>::epsilon())
-  {
-    return ConstraintEvaluationResult(true, 0.0);
-  }
-
   if (max_view_angle_ > 0.0 || max_range_angle_ > 0.0)
   {
     // getFrameTransform() returns a valid isometry by contract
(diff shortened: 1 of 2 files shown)


moveit_core/kinematic_constraints/src/kinematic_constraint.cpp
```diff
@@ -1142,38 +1140,49 @@ ConstraintEvaluationResult VisibilityConstraint::decide(const moveit::core::Robo
...
+    shapes::Mesh* m = getVisibilityCone(state);
+    if (!m)
+    {
+      RCLCPP_ERROR(LOGGER, "Visibility constraint is violated because we could not create the visibility cone mesh.");
+      return ConstraintEvaluationResult(false, 0.0);
+    }
 
-  collision_env_->getWorld()->removeObject("cone");
+    // add the visibility cone as an object
+    collision_env_local->getWorld()->addToObject("cone", shapes::ShapeConstPtr(m), Eigen::Isometry3d::Identity());
```
henningkayser, line 1155: I see why this would cause problems with multi-threading, this should work. Another way to solve this multi-threading issue is to attach the cone to the (local) robot state and check for self-collisions. That way, there shouldn't be any race conditions since there are no variable states being shared between threads. `checkCollision()` alone is thread-safe afaik. What do you think?

schornakj: I think recreating the collision environment every time we check this constraint is probably quite inefficient, so if there's a way to reduce that I think that's preferable.

> Another way to solve this multi-threading issue is to attach the cone to the (local) robot state and check for self-collisions.

I'd appreciate some pointers on how to do this -- is this using a function of the robot state or the collision environment?

An important thing to consider is that the visibility cone needs to work both when it moves with the robot and when it's fixed in the world. The way I have it written here works for both cases, but it loses some of the efficiency of the previous (attempted but buggy) approach of precalculating the transforms for fixed frames.

henningkayser (approved):
I think we should get this fixed merged for now to get rid of the segfault. @schornakj since you are using this, do you think you could evaluate the performance impact and decide if we need to address this further?

sea-bass (approved):
LGTM

Issue #586, opened 2021-08-01
TOTG does not handle mimic joints properly


AndyZe:
The TOTG trajectory processing plugin uses this:

`const std::vector<int>& idx = group->getVariableIndexList();`

[Here](https://github.com/ros-planning/moveit2/pull/571#discussion_r680264144), Henning said that could cause issues with mimic joints.

Perhaps best to replace it with `group->getActiveJointModelNames()`?

PR #1989, opened 2023-03-04
Fix mimic joints with TOTG

sea-bass:
### Description

Made some modifications to TOTG and joint model group to reason in *active* joints, which solves issues with TOTG with mimic joints. I can confirm this change fixed things for a robot configuration with a Robotiq 2f-85 gripper that has mimic joints.

Fixes https://github.com/ros-planning/moveit2/issues/586.

### Checklist

### DIFF
--- a/moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
+++ b/moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
@@ -166,19 +166,6 @@ public:
     return active_joint_model_name_vector_;
   }
 
-  /** \brief Get the names of the active variables that make up the joints included in this group. The number of
-      returned elements is always equal to getVariableCount(). This does not include mimic joints. */
-  const std::vector<std::string>& getActiveVariableNames() const
-  {
-    return active_variable_names_;
-  }
-
-  /** \brief Get the active index locations in the complete robot state for all the variables in this group */
-  const std::vector<int>& getActiveVariableIndexList() const
-  {
-    return active_variable_index_list_;
-  }
-
   /** \brief Get the fixed joints that are part of this group */
   const std::vector<const JointModel*>& getFixedJointModels() const
   {
@@ -595,9 +582,11 @@ public:
   bool isValidVelocityMove(const double* from_joint_pose, const double* to_joint_pose, std::size_t array_size,
                            double dt) const;
 
-protected:
-  bool computeIKIndexBijection(const std::vector<std::string>& ik_jnames, std::vector<size_t>& joint_bijection) const;
+  /** \brief Computes the indices of joint variables given a vector of joint names to look up */
+  bool computeJointVariableIndices(const std::vector<std::string>& joint_names,
+                                   std::vector<size_t>& joint_bijection) const;
 
+protected:
   /** \brief Update the variable values for the state of a group with respect to the mimic joints. This only updates
       mimic joints that have the parent in this group. If there is a joint mimicking one that is outside the group,
       there are no values to be read (\e values is only the group state) */
@@ -637,14 +626,6 @@ protected:
   /** \brief The names of the DOF that make up this group (this is just a sequence of joint variable names; not
       necessarily joint names!) */
   std::set<std::string> variable_names_set_;
-
-  /** \brief The names of the active DOF that make up this group (this is just a sequence of joint variable names;
-      not necessarily joint names!) */
-  std::vector<std::string> active_variable_names_;
-
-  /** \brief The list of active index values this group includes, with respect to a full robot state;
-      this does not include mimic joints. */
-  std::vector<int> active_variable_index_list_;
 
   /** \brief A map from joint names to their instances. This includes all joints in the group. */
   JointModelMapConst joint_model_map_;
--- a/moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
+++ b/moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
@@ -166,6 +166,19 @@ public:
     return active_joint_model_name_vector_;
   }
 
+  /** \brief Get the names of the active variables that make up the joints included in this group. The number of
+      returned elements is always equal to getVariableCount(). This does not include mimic joints. */
+  const std::vector<std::string>& getActiveVariableNames() const
+  {
+    return active_variable_names_;
+  }
+
+  /** \brief Get the active index locations in the complete robot state for all the variables in this group */
+  const std::vector<int>& getActiveVariableIndexList() const
+  {
+    return active_variable_index_list_;
+  }
+
   /** \brief Get the fixed joints that are part of this group */
   const std::vector<const JointModel*>& getFixedJointModels() const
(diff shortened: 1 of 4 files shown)


moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -734,6 +744,10 @@ void JointModelGroup::printGroupInfo(std::ostream& out) const
   out << "    ";
   for (int variable_index : variable_index_list_)
     out << variable_index << ' ';
+  out << "  * Active Variables Index List:\n";
+  out << "    ";
+  for (int active_variable_index : active_variable_index_list_)
+    out << active_variable_index << ' ';
```
AndyZe, line 750: I think you need a newline `"\n"` at the end of this

AndyZe (commented):
Looks good to me except one small question

moveit_core/robot_model/src/joint_model_group.cpp
sea-bass: Good call. I moved it after the `(contiguous)` vs `(non-contiguous)` printout that was in the same line too.

AndyZe (approved):

sea-bass:
Is rolling CI broken atm or is this something introduced by this PR?

AndyZe:
Yeah, it's broken

moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
```diff
@@ -166,6 +166,19 @@ class JointModelGroup
     return active_joint_model_name_vector_;
   }
 
+  /** \brief Get the names of the active variables that make up the joints included in this group. The number of
+      returned elements is always equal to getVariableCount(). This does not include mimic joints. */
+  const std::vector<std::string>& getActiveVariableNames() const
```
henningkayser, line 171: What's wrong with getActiveJointModelNames() https://moveit.picknik.ai/humble/api/html/classmoveit_1_1core_1_1JointModelGroup.html#ad3a2b2661f3303f1d6fb8a8a08008174

henningkayser: ... ok I see. The fact that you have to do something like this really shows that the API and mapping implementation of RobotState and joints is very awkward. We know that TOTG only supports active joints with single variables, and in that case the variable names always equal the joint model names (https://moveit.picknik.ai/humble/api/html/classmoveit_1_1core_1_1JointModel.html#a09b0ad5f0d57c67a1fd7cde8d494519d). But that's convention and not clear from the API. I'm not against adding these functions, but they somewhat make the API even more complicated. Would it help to just expose and use this function instead https://moveit.picknik.ai/humble/api/html/classmoveit_1_1core_1_1JointModelGroup.html#aa11b464a43bb475addb128a9f1b8bbb1?

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -884,11 +884,11 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
   double velocity_scaling_factor = verifyScalingFactor(max_velocity_scaling_factor, VELOCITY);
   double acceleration_scaling_factor = verifyScalingFactor(max_acceleration_scaling_factor, ACCELERATION);
 
-  const std::vector<std::string>& vars = group->getVariableNames();
+  // Get the velocity and acceleration limits for all active joints
```
henningkayser, line 887: I had some concerns about possible side-effects at first, but I think they are completely fine now.

moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
sea-bass: I also agree that adding more "getters" is awkward, so if the suggested function makes this a little better, I'm open to try it.

I also have to find time to figure out how/why this breaks rolling but not humble, as the changes are purely additive from what I can tell.

henningkayser: maybe related to this? https://github.com/ros-planning/moveit2/pull/1821

sea-bass: Looking through, I think the test is just fundamentally broken, actually.

The test initializes a trajectory of 5 waypoints in [this function](https://github.com/ros-planning/moveit2/blob/main/moveit_core/robot_trajectory/test/test_robot_trajectory.cpp#L68-L89), but all the waypoints have the same exact robot state... so it's actually correct that the path length is 0.0. But the test wants it to be (greater than 0](https://github.com/ros-planning/moveit2/blob/70af62818f0e0be8c572c2f51b3357854418b5fd/moveit_core/robot_trajectory/test/test_robot_trajectory.cpp#L326).

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -982,9 +982,9 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(
   double velocity_scaling_factor = verifyScalingFactor(max_velocity_scaling_factor, VELOCITY);
   double acceleration_scaling_factor = verifyScalingFactor(max_acceleration_scaling_factor, ACCELERATION);
 
-  const unsigned num_joints = group->getVariableCount();
-  const std::vector<std::string>& vars = group->getVariableNames();
   const moveit::core::RobotModel& rmodel = group->getParentModel();
+  const std::vector<std::string>& vars = group->getActiveVariableNames();
+  const unsigned num_joints = group->getActiveVariableCount();
```
sea-bass, line 987: @henningkayser this is the change that breaks the unit tests. I think because this just scales all the joints (mimic or not), we probably want to keep it as it was

moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
```diff
@@ -582,9 +582,10 @@ class JointModelGroup
   bool isValidVelocityMove(const double* from_joint_pose, const double* to_joint_pose, std::size_t array_size,
                            double dt) const;
 
-protected:
+  /** \brief Computes the indices of joints given a vector of joint names to look up */
   bool computeIKIndexBijection(const std::vector<std::string>& ik_jnames, std::vector<size_t>& joint_bijection) const;
```
sea-bass, line 586: @henningkayser I switched this implementation to your suggestion and seems to work fine.

Question, though: Should we rename this function to not have a name targeted to IK, at this point?

sea-bass: `computeJointVariableIndices()` SGTM.

I'm building these changes against Studio + running CI here to verify that they work, and if they do I'll do this as well.

sea-bass:
So now the Humble job fails because we renamed the function ... how do we reconcile things like these?

henningkayser (approved):

Issue #1970, opened 2023-02-23
CI is broken on Rolling


mikeferguson:
The fix is almost certainly needed to be upstream - just ticketing here for visibility (so we don't have multiple people working on it)

 * @AndyZe tried to fix this with https://github.com/ros-planning/moveit_msgs/pull/153 - BUT:
 * service_msgs was added only to Rolling

It looks like the failure is actually in sensor_msgs:

<img width="798" alt="image" src="https://user-images.githubusercontent.com/899410/220978159-29730e52-0506-425a-aaf0-b8a25875e83c.png">

mikeferguson:
Locally, I can build moveit_msgs against ros2-testing debs (and I uninstalled all of ros-rolling-* before tasting, so it's not a missing dependency).

sjahr:
This commit could be the culprit: https://github.com/ros2/rosidl_defaults/commit/a3f753d1527cbc8be3b6c41c172d68732414584d. But it still does not explain why service_msgs is not installed :thinking:

jspricke:
The job uses an old Docker image of the ros2-testing repos where ros-rolling-service-msgs is not installed and it does not `apt dist-upgrade` the system: https://github.com/ros-planning/moveit2/blob/main/.docker/source/Dockerfile#L28. So service_msgs is never picked up and thus missing. The logical fix is to use dist-upgrade in the job, as proposed in a144a3c but Docker lint complains about that: https://github.com/jspricke/moveit2/actions/runs/4346334681. So you can either fix Docker lint or build a new Docker base image and run into this problem again, next time. Your call.

mikeferguson:
@tylerjw - I think this may be more up your alley - since it looks like it is our CI, not some upstream break

jspricke:
Discussed this in the standup, the problem is that the ci-testing Docker images are based on the ci images which the added ros2-testing apt sources: https://github.com/ros-planning/moveit2/blob/main/.docker/ci-testing/Dockerfile and the `apt-get upgrade` in there is not enough to install a new dependency package. @henningkayser will rework the Dockerfile to work from a base image.

vatanaksoytezer:
I think we only need to add a rosdep step to ci-testing and that should be sufficient. The dependency in question (service_msgs) is not being installed in apt update / apt upgrade, it should be installed and triggered through `rosdep install`

jspricke:
@vatanaksoytezer no, service_msgs is a dependency of rcl already.

PR #1994, opened 2023-03-09
Fix ci testing

henningkayser:
Fixes https://github.com/ros-planning/moveit2/issues/1970.

The underlying issue was that ci-testing was based on ros-base which didn't provide the same package dependencies. Apt can fix removed or added packages with `dist-upgrade`, but that command is discouraged by hadolint, and it also wouldn't make much sense in the context of reusing image layers in docker.

The fix is to rely on osrf's testing image and to manually install the base dependencies directly from testing. This adds more work to our testing build, but by splitting it out into a separate RUN command we can rely on the build cache most of the time (testing is updated very rarely).

New docker images have already been created pushed from this branch.

### DIFF
--- a/.docker/ci-testing/Dockerfile
+++ b/.docker/ci-testing/Dockerfile
@@ -2,15 +2,59 @@
 # CI image using the ROS testing repository
 
 ARG ROS_DISTRO=rolling
-FROM moveit/moveit2:${ROS_DISTRO}-ci
+FROM osrf/ros2:testing
 LABEL maintainer Robert Haschke rhaschke@techfak.uni-bielefeld.de
 
-# Switch to ros-testing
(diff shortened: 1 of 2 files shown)


jspricke (approved):

henningkayser:
The fix works, CI fails for other reasons.

The old issue was that moveit_msgs failed to compile in the upstream workspace because of missing service_msgs. This succeeds now!

In the meantime, we must have missed an upstream API breakage in controller_manager_msgs. moveit_ros_control_interface fails here https://github.com/ros-planning/moveit2/actions/runs/4374898377/jobs/7654908084#step:12:352.

henningkayser:
#1996 fixes the API breakage and hopefully the whole rolling-ci-testing job

PR #1949, opened 2023-02-15
Fix member naming

rhaschke:
@tylerjw, this PR attempts to address https://github.com/ros-planning/moveit2/pull/1948#issuecomment-1431964275.
As I don't have time to deeply look into this, I just cherry-picked corresponding commits from https://github.com/ros-planning/moveit/pull/3317, with some manual cleanup. I didn't even compile locally. So I expect some more work to be done on this. Any help is welcome as I won't have time to finish this within the next days.

PR #1998, opened 2023-03-09
Fix unreachable child logger instance

henningkayser:
https://github.com/ros2/rcl/pull/921 fixed child loggers so that CI now detected an issue with one of our loggers. The problem was that the parent logger instance was no longer available at call time. The fix here creates a new logger instance for the child's namespace that would not go out of scope.


tylerjw (approved):

rhaschke:
Instead of manually fixing naming-related clang-tidy issues when they pop up, somebody should take the time and finish https://github.com/ros-planning/moveit2/pull/1949.

PR #2001, opened 2023-03-10
Fix MoveItCpp issues (port from MoveIt1)

rhaschke:
Forward port of https://github.com/ros-planning/moveit/pull/3369. Untested!

- Fix `const` member accessors. They should return a `ConstPtr` instead of a `const Ptr&`!
- Fix SEVERE ClassLoader warning on destruction of `MoveItCpp`:
  - PSM was released before copy of its RobotModel -> removed extra RobotModel copy
  - `clearContents()` was broken:
    - resets in wrong order: psm_ should be last
    - trajectory_execution_manager_ was missing
    
    I suggest to omit `clearContents()` and just rely on the (correct) ordering of member variables.
    While this is not explicit, we ensure that we don't miss any newly added member variable.


henningkayser (commented):
Thank you, these changes look good to me. We'll probably have to fixup moveit_py, though:
* https://github.com/ros-planning/moveit2/blob/main/moveit_py/src/moveit/moveit_ros/moveit_cpp/moveit_cpp.cpp#L153 should return non-const
* ~https://github.com/ros-planning/moveit2/blob/main/moveit_py/src/moveit/moveit_ros/moveit_cpp/moveit_cpp.cpp#L118 should use non-const~

sjahr (approved):

JafarAbdi (approved):

Issue #2015, opened 2023-03-14
Flaky test: moveit_planners_ompl testPathConstraints


stephanie-eng:
### Description
#1300 caused [this test](https://github.com/ros-planning/moveit2/blob/39070093610699990e145e752f18cd5158cc262c/moveit_planners/ompl/ompl_interface/test/test_planning_context_manager.cpp#L114) to become flaky. 

This test tries to create a simple orientation-constrained plan. The failure is inconsistent, and only happens occasionally. The failure occurs at [this line](https://github.com/ros-planning/moveit2/blob/39070093610699990e145e752f18cd5158cc262c/moveit_planners/ompl/ompl_interface/test/test_planning_context_manager.cpp#L157); it is caused by a planning failure. Test results indicate that planning times out before the planning problem is solved. This same test is repeated for both the Panda and the FANUC, but the test only seems to be flaky for the Panda, and does not seem to fail for the FANUC.

The constrained planning problem already has an extended timeout (10 seconds) to account for the more-difficult planning problem, and further extending the timeout does not seem to resolve this issue. Similarly, relaxing constraints also does not seem to resolve this issue.

PR #2016, opened 2023-03-14
Temporarily disable TestPathConstraints with the Panda robot

stephanie-eng:
This test has become flaky since it was modified to use the OMPL constrained state space. #2015 tracks the flaky test issue.


tylerjw (approved):
Thank you for digging into this and creating an issue for it.

henningkayser:
IKFast tests are failing because of an outdated gpg key https://github.com/ros-planning/moveit2/issues/2017

henningkayser (approved):

Issue #1490, opened 2022-08-09
Stale bot for PRs


vatanaksoytezer:
Add a GH bot that would warn the authors and/or reviewers when a PR goes stale (2 weeks of response?). This should be annoying enough so that we prevent valuable PRs and work go stale. We may add an additional feature that would close the PR if it warned enough times and didn't get responses / updates.

v4hn:
Stale-PR/Issue notifications are a lot of junk in your notification list/inbox. More unread/marked-as-read content does not help.
I'm sorry, but I think it's the wrong response to the problem (which is a normal one in Open Source projects).

Instead, if PickNik can afford it, add dedicated well-defined working hours for some engineer(s) explicitly on PR handling (instead of specific grander projects) and decide a ranking for all open PRs/issues on a bi-weekly basis. Then have them spend time on the top of the list and enforce that people do not wait for feedback from PR authors but are allowed to push their own cleanup patches or create follow-up PRs.

henningkayser:
I'm proposing a version of this here https://github.com/ros-planning/moveit2/pull/2022.

@v4hn I agree that stale labels and notifications won't help with identifying inactivity (it's fairly easy to do this using the normal GitHub tools). The purpose of this action is really to require the authors of issues to be responsible for their proposal, and to not bloat the issue list with ill-considered wishes or thoughts just to get it out there and hope that someone else will magically spend a few days to produce a solution. Issues are a communication tool, and they should be seen as an invitation for an active discussion by the author. If no-one even bothers to discuss an issue for 3 months, it's obviously not worthy of discussion and it can be closed (same thing I'm doing manually all the time). Note that feature discussions and roadmap items can be exempt from this explicitly. With PRs it's also different, that is why I disabled automatic closing for them as well. Here the GHA acts as an automatic reminder for the author and for us to take care of this work item and to find a solution for moving forward. It obviously doesn't replace putting in the hours for doing the reviews.

PR #2022, opened 2023-03-17
Add stale GHA

henningkayser:
Enables a [stale action](https://github.com/marketplace/actions/close-stale-issues) that labels issues and PRs as stale after 45 days, and closes issues after another 45 days. The action also takes care about warning the author appropriately, and it ignores a set of persistent labels. The intention of this is to really only keep issues open that are persistent explicitly (there is a new label for that), or that belong to a roadmap, milestone, etc.. All other issues should require at least some activity within the 3 months of time it would take for this GHA to close them.

Fixes https://github.com/ros-planning/moveit2/pull/2022

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -1,0 +1,26 @@
+name: 'Stale issues and PRs'
+description: 'Labels inactive issues and PRs as stale, and only closes stale issues after ongoing inactivity'
+on:
+  schedule:
+    # UTC noon every workday
+    - cron: '0 12 * * MON-FRI'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v7
+        with:
+          stale-issue-label: 'stale'
+          stale-pr-label: 'stale'
+          stale-issue-message: 'This issue is being labeled as stale because it has been open 45 days with no activity. It will be automatically closed after another 45 days without follow-ups.'
+          close-issue-message: 'This issue was closed because it has been stalled for 45 days with no activity.'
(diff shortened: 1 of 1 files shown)


tylerjw (approved):

v4hn:
I still think this is superficial make-up and the only thing it does is add bloat where maintainer actions are required.
An issue list (even when the issues age) is a valuable resource of user input, not a junkyard that needs automatic discarding. In an ideal world all items in it warrant a manually added "can't reproduce/not enough information/won't fix" label before they are closed without a merged patch. 

I guess you weakened the conditions enough to avoid further complains.

AndyZe:
I thought discussions were supposed to go in the Discussions tab that was the cool new thing a couple months ago  ;)

This type of shuffling of deck chairs amuses me. I hope no valuable old issues get lost because of it.

henningkayser:
Same response as here https://github.com/ros-planning/moveit2/issues/1490#issuecomment-1473867153. 


> I still think this is superficial make-up
yes
> and the only thing it does is add bloat where maintainer actions are required
it automates superficial maintainer actions
> An issue list (even when the issues age) is a valuable resource of user input, not a junkyard
it's both
> that needs automatic discarding
it doesn't, but it helps
> In an ideal world
... 
> all items in it warrant a manually added "can't reproduce/not enough information/won't fix" label before they are closed without a merged patch.
We do have that and use it frequently
> I guess you weakened the conditions enough to avoid further complains.
I hope so. My concern is really not about rejecting valuable issues. It's about requiring people to think about the content and value of their issue first before posting it in the first place. There is a huge amount of content that is along the lines of "we could...", "it would be nice..." that no one ever responds to because it's apparently not important at all. If only all maintainers would post issues of whatever wishes or concerns they have about the code base, the list would be endless and it would still be just as useless. Let's focus on actual problems and action items that are important enough to discuss them.

> I thought discussions were supposed to go in the Discussions tab that was the cool new thing a couple months ago ;)
GitHub discussions are obviously something different than issues but it's hard to argue that issues are not meant for discussions.

PR #2023, opened 2023-03-18
For convenience, add a vector version of RobotState::getVariableVelocities() #2018

### LINKED ISSUE #2018: For convenience, add a vector version of `RobotState::getVariableVelocities()`
It would be convenient to add a std::vector-ized version of this:

```
  const double* getVariableVelocities() const
  {
    return velocity_;
  }
```

Tejal-19:
### Description

added a vector version of RobotState::getVariableVelocities()

Fixes #2018

### Checklist


AndyZe:
@Tejal-19 notice that I changed the description, adding "Fixes #2018". That's so it will close the issue automatically if this gets merged.

AndyZe:
@henningkayser What would you think about just deleting the functions `getVariableVelocities()` and `getVariableAccelerations()`? Here's my thinking:

- getVariableVelocities() and getVariableAccelerations() are only used in one place in the codebase. That's here in current_state_monitor.cpp, `setToCurrentState()`.  https://github.com/ros-planning/moveit2/blob/9f45e6c2fce97ed0a38cc61445f0b2fe4506aca8/moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp#L124
- setToCurrentState() is also only used once. That's in planning_scene_monitor.cpp `updateSceneWithCurrentState()`. I don't believe the PSM would ever need to know robot vels or accels, so this is unnecessary
- getVariableVelocities() and getVariableAccelerations() are only necessary if dynamics are copied. I think that's pretty rare.
- There's already another function that's very similar to `getVariableVelocities()`. It's called `copyJointGroupVelocities()` as you pointed out.  https://moveit.picknik.ai/humble/api/html/classmoveit_1_1core_1_1RobotState.html#a3c2938fb8c9467dff253db4b5c1992ca
- This `double*` API is messy

The more I think about this, the more I think it's a good simplification.

AndyZe:
Pinging @henningkayser to respond to the above comment

henningkayser (changes requested):
This is just plain wrong. What are you trying to do?

ibrahiminfinite:
> * setToCurrentState() is also only used once. That's in planning_scene_monitor.cpp `updateSceneWithCurrentState()`. I don't believe the PSM would ever need to know robot vels or accels, so this is unnecessary

Would this not be relevant for this issue ? #1975

Tejal-19:
> This is just plain wrong. What are you trying to do? If you want to get a vector of the internal array, you'll have to copy the elements using a separate getter function.

Can someone please provide a reference to this. Thankyou

henningkayser:
> > This is just plain wrong. What are you trying to do? If you want to get a vector of the internal array, you'll have to copy the elements using a separate getter function.
> 
> Can someone please provide a reference to this. Thankyou

The reference is the source code. `velocity_` is a member variable with a double pointer for the value array. There is no direct conversion from a double array to a vector, so the only option is to create a new vector and to copy over the values. We certainly don't want to break API over a convenience function, so a new copy function is the best solution for now. The provided approach is plain wrong since it ignores the member variable and only returns the vector passed in as function argument. This doesn't work.

sjahr:
Closing this due to inactivity. Feel free to re-open if you get it working

PR #2028, opened 2023-03-20
Delete the Ruckig "batches" option, deprecated by #1990

AndyZe:
### Description

You could say that #1990 took the "batches" idea to the extreme - it extends durations by one segment at a time. So this complicated function isn't helpful anymore (and it was also unused).

Thanks @ibrahiminfinite


ibrahiminfinite:
oh wow!
This is an interesting consequence.

sjahr (approved):

PR #2030, opened 2023-03-20
Fix getting variable bounds in mimic joints for TOTG

sea-bass:
### Description

Following the changes in #1989, I found another bug with this implementation.

In TOTG, we were getting joint limits for only non-mimic joints, so planning in a group that had mimic joints would use a max velocities / acceleration vector with zero values. This caused an infinite loop in TOTG's `Trajectory::integrateForward()` where the solver kept taking zero steps and never breaking out of its loop.

This PR fixes the problem by building a vector of non-mimic joint bounds correctly, instead of creating uninitialized (zero) elements for the indices pertaining to mimic joints.

I tested this in MoveIt Studio and it solved my issues.

### Checklist


moveit_core/robot_model/include/moveit/robot_model/robot_model.h
```diff
@@ -429,10 +429,23 @@ class RobotModel
...
+    const auto mimic = joint->getMimic();
+    if (!mimic)
+    {
+      return joint->getVariableBounds(variable);
+    }
+    else
+    {
+      const auto orig_idx = joint->getLocalVariableIndex(variable);
+      const auto var_names = mimic->getVariableNames();
+      return mimic->getVariableBounds(var_names.at(orig_idx));
```
sea-bass, line 447: Is this valid to do?

I didn't see an easy way in `JointModel` to know which variable in a mimic joint corresponds to which variable in its parent joint.

henningkayser:
This is a bit confusing to me. Mimic joints should not really be part of TOTG's trajectory profile and not be considered during the process. Mimic joint values are instantiated automatically from the parent joints.

sea-bass:
> This is a bit confusing to me. Mimic joints should not really be part of TOTG's trajectory profile and not be considered during the process. Mimic joint values are instantiated automatically from the parent joints.

Yeah, in hindsight I think I now see a different way to fix it. Let me test this change today and get back.

sea-bass:
OK @henningkayser this latest commit still works for me in MoveIt Studio and doesn't actually consider mimic joints in the loop.

The changes to `RobotModel::getVariableBounds()` are technically not needed anymore, but I don't see the harm in keeping that for other applications where you may want to get the variable bounds for a mimic joint.

Let me know what you think!

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -891,12 +891,12 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
     RCLCPP_ERROR(LOGGER, "Failed to get active variable indices.");
   }
 
-  const size_t num_joints = indices.size();
-  Eigen::VectorXd max_velocity(num_joints);
-  Eigen::VectorXd max_acceleration(num_joints);
-  for (const auto idx : indices)
+  const size_t num_active_joints = indices.size();
```
PeterMitrano, line 894: So if I understand this correctly, you want to use `indices` when getting bounds, but the raw `idx` everywhere else?

moveit_core/robot_model/include/moveit/robot_model/robot_model.h
PeterMitrano: This seems fine to me. It wouldn't make sense for the two mimic'd joints to have a different number of variables, and the ordering of variables should be deterministic. And even if that's not true, I've never seen a 2dof mimic joint.

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
sea-bass: Yeah, I had originally messed this up, so here is the idea.

Suppose you have a robot with 5 real joints and 1 mimic joint: `indices` might give you something like `[0, 1, 3, 4]`, let's say, so index `2` is the mimic.

The limits then have to look through 4 joints (this is `num_active_joints`), and each element should correspond to the limits for joints `0`, `1`, `3`, and `4`, respectively.

tl;dr
* `indices = [0, 1, 3, 4]`
* `idx` in the loop takes values `[0, 1, 2, 3]`

PeterMitrano: Thanks, that makes sense!

PeterMitrano (approved):

moveit_core/robot_model/include/moveit/robot_model/robot_model.h
sea-bass: I actually ended up changing it so that the `setMimic()` function itself updates the bounds when the mimic relationship is first established.

moveit_core/robot_model/src/joint_model.cpp
```diff
@@ -242,6 +242,28 @@ void JointModel::setMimic(const JointModel* mimic, double factor, double offset)
   mimic_ = mimic;
   mimic_factor_ = factor;
   mimic_offset_ = offset;
+
+  // Scale and offset the variable bounds based on the mimic joint.
+  if (getVariableCount() != mimic->getVariableCount())
+    throw Exception("Variable counts of this joint and mimic joint do not match.");
```
PeterMitrano, line 248: curly braces for one-line if statements?

PeterMitrano (approved):

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -891,12 +891,12 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
...
+  Eigen::VectorXd max_velocity(num_active_joints);
+  Eigen::VectorXd max_acceleration(num_active_joints);
+  for (size_t idx = 0; idx < num_active_joints; ++idx)
```
AndyZe, line 897: Just like Peter, I was confused initially. Let's improve the naming:

```suggestion
  for (size_t active_joint_idx = 0; active_joint_idx < num_active_joints; ++active_joint_idx)
```

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -891,12 +891,12 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
...
   {
-    const moveit::core::VariableBounds& bounds = rmodel.getVariableBounds(vars[idx]);
+    const moveit::core::VariableBounds& bounds = rmodel.getVariableBounds(vars[indices[idx]]);
```
AndyZe, line 899:
```suggestion
    // For active joints only (skip mimic joints and other types)
    const moveit::core::VariableBounds& bounds = rmodel.getVariableBounds(vars[indices[active_joint_idx]]);
```

AndyZe (commented):
LGTM after this name change and a comment.

AndyZe (approved):

sea-bass:
> LGTM after this name change and a comment.

I think there was a slight issue in the variable naming. How does this look instead?

Pomorondza:
Hello, I am getting the following error when trying to control the robot with JointTrajectoryController:

_[moveit_trajectory_processing.time_optimal_trajectory_generation]: No acceleration limit was defined for joint shoulder_pan_joint! You have to define acceleration limits in the URDF or joint_limits.yaml_

Here is how the shoulder_pan_joint limits look in the joint_limits.yaml file:
```
joint_limits:
  shoulder_pan_joint:
    # acceleration limits are not publicly available
    has_position_limits: true
    has_velocity_limits: true
    has_acceleration_limits: true
    has_effort_limits: true
    max_position: !degrees  200.0
    min_position: !degrees -30.0
    max_velocity: !degrees  180.0
    max_acceleration: !degrees 60.0
    max_effort: 56.0
```
Do you know what might be the issue?

sea-bass:
> Hello, I am getting the following error when trying to control the robot with JointTrajectoryController on the UR3e robot:
> 
> _[moveit_trajectory_processing.time_optimal_trajectory_generation]: No acceleration limit was defined for joint shoulder_pan_joint! You have to define acceleration limits in the URDF or joint_limits.yaml_
> 
> Here is how the shoulder_pan_joint limits look in the joint_limits.yaml file:
> 
> ```
> joint_limits:
>   shoulder_pan_joint:
>     has_position_limits: true
>     has_velocity_limits: true
>     has_acceleration_limits: true
>     has_effort_limits: true
>     max_position: !degrees  200.0
>     min_position: !degrees -30.0
>     max_velocity: !degrees  180.0
>     max_acceleration: !degrees 60.0
>     max_effort: 56.0
> ```
> 
> Do you know what might be the issue?

Could you please make a new git issue for this? And share what example you're running?

My guess is the joint limits params are either not being passed in in the launch file, or the !degrees bits are not being parsed correctly depending on what is being used.

Pomorondza:
You were right in both, the fix in my case was to add robot_description_planning (before it was only added through robot_description) to the moveit_ros_move_group node, and also change the values to radians.
```
    robot_description_planning = {
        "robot_description_planning": {
            **joint_limits_yaml,
        }
    }
```
Thank you!

Issue #1944, opened 2023-02-12
MoveIt Setup Assistant crashes when URDF is invalid path or not on ROS2 path


vidalperezbohoyo:
### Description
I'm trying to setup a new robot, but I'm barely able to get past the first step. The program stops responding when clicking on create a new package and trying to load an urdf. Even writing something in the textbox already stops working. It doesn't get any messages from the terminal.

### Your environment
* ROS Distro: Humble
* OS Version: 22.04
* Source and binary (same with this error) 
* Latest
* Latest
* Cyclone DDS

### Steps to reproduce
If write on textbox in create new package import urdf or browse and select any urdf.

patricia241:
I have the same problem.

LSBigum:
I haven't got this issue. Did the following:
```sudo apt install ros-humble-moveit``` 
Sourced the underlay and my overlay (```source /opt/ros/humble/setup.bash``` and ```source ros2_ws/install/setup.bash```)
Ran ```ros2 launch moveit_setup_assistant setup_assistant.launch.py verbose:=true```
And then I was able to load my .xacro file and generate a moveit config.

- ROS Distro: Humble
- OS Version: 22.04
- Binary install
- Fast DDS

**Comment by vidalperezbohoyo** [comment]
Doesn't work. I have this installed. With verbose nothing is displayed when crashes.

**Comment by vidalperezbohoyo** [comment]
I have tried to unistall it , then autoremove to clean all. Then changed Ubuntu to US lenguage and then reboot. Then installed again with your command. But is NOT working...

**Comment by vidalperezbohoyo** [comment]
I have retried in a virtualbox from zero. Same error. 

Now there is more info:

```
[INFO] [launch]: All log files can be found below /home/vidi/.ros/log/2023-02-19-14-07-51-890511-vidi-VirtualBox-53720
[INFO] [launch]: Default logging verbosity is set to INFO
[INFO] [moveit_setup_assistant-1]: process started with pid [53721]
[moveit_setup_assistant-1] Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
[ERROR] [moveit_setup_assistant-1]: process has died [pid 53721, exit code -9, cmd '/opt/ros/humble/lib/moveit_setup_assistant/moveit_setup_assistant --ros-args'].

```

Is this version of moveit. Assistant is broken!

mikeferguson:
[...]

PR #2032, opened 2023-03-20
Add URDF Loader Exceptions and Fix Infinite While-Loop when URDF file isn't in a ROS Package

chancecardona:
### Description
This PR fixes [Issue 1944](https://github.com/ros-planning/moveit2/issues/1944). When selecting a URDF from a file path, MSA would hang forever before needing to be killed if the URDF was not in a ROS package; with no error messages printed. This was due to an infinite while-loop in the moveit_setup_framework/src/utilities.cpp file when extracting the package name from the file path. This fixes that while-condition to check against the root file path when truncating the path. Also added a catch so if ament throws an error finding the package it will just return an empty filepath instead. 

This PR also adds some error handling to the start screen widget to propagate URDF loading errors to the user via the GUI, rather than just crash. Previous behavior was that when selecting an invalid URDF, the MSA tool crashed and the ROS logs displayed the exception.

Partially based off of initial work done by @mikeferguson 

### Checklist

### DIFF
--- a/moveit_setup_assistant/moveit_setup_core_plugins/src/start_screen_widget.cpp
+++ b/moveit_setup_assistant/moveit_setup_core_plugins/src/start_screen_widget.cpp
@@ -312,8 +312,8 @@ void StartScreenWidget::onUrdfPathChanged(const QString& path)
   }
   catch (const std::runtime_error& e)
   {
-    RCLCPP_ERROR(setup_step_.getLogger(), "%s", e.what());
-    QMessageBox::warning(this, "Error Loading URDF path", QString(e.what()));
+    RCLCPP_ERROR(setup_step_.getLogger(), "Error Loading URDF Path. %s", e.what());
+    QMessageBox::warning(this, "Error Loading URDF Path", QString(e.what()));
   }
 }
 
@@ -351,7 +351,7 @@ bool StartScreenWidget::loadExistingFiles()
   }
   catch (const std::runtime_error& e)
   {
-    RCLCPP_ERROR(setup_step_.getLogger(), "%s", e.what());
+    RCLCPP_ERROR(setup_step_.getLogger(), "Error Loading SRDF. %s", e.what());
     QMessageBox::warning(this, "Error Loading SRDF", QString(e.what()));
     return false;
   }
--- a/moveit_setup_assistant/moveit_setup_core_plugins/src/start_screen_widget.cpp
+++ b/moveit_setup_assistant/moveit_setup_core_plugins/src/start_screen_widget.cpp
@@ -305,8 +305,16 @@ void StartScreenWidget::onPackagePathChanged(const QString& /*path*/)
 
 void StartScreenWidget::onUrdfPathChanged(const QString& path)
 {
-  setup_step_.loadURDFFile(path.toStdString(), urdf_file_->getArgs().toStdString());
-  urdf_file_->setArgsEnabled(setup_step_.isXacroFile());
+  try
+  {
+    setup_step_.loadURDFFile(path.toStdString(), urdf_file_->getArgs().toStdString());
+    urdf_file_->setArgsEnabled(setup_step_.isXacroFile());
+  }
+  catch (const std::runtime_error& e)
+  {
+    RCLCPP_ERROR(setup_step_.getLogger(), "%s", e.what());
+    QMessageBox::warning(this, "Error Loading URDF path", QString(e.what()));
+  }
 }
 
 bool StartScreenWidget::loadPackageSettings(bool show_warnings)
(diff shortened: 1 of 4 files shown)


moveit_setup_assistant/moveit_setup_core_plugins/src/start_screen_widget.cpp
```diff
@@ -305,8 +305,16 @@ void StartScreenWidget::onPackagePathChanged(const QString& /*path*/)
...
-  setup_step_.loadURDFFile(path.toStdString(), urdf_file_->getArgs().toStdString());
-  urdf_file_->setArgsEnabled(setup_step_.isXacroFile());
+  try
+  {
+    setup_step_.loadURDFFile(path.toStdString(), urdf_file_->getArgs().toStdString());
+    urdf_file_->setArgsEnabled(setup_step_.isXacroFile());
+  }
+  catch (const std::runtime_error& e)
+  {
+    RCLCPP_ERROR(setup_step_.getLogger(), "%s", e.what());
```
sea-bass, line 315: Should we also add the "Error Loading URDF path" prefix to the `RCLCPP_ERROR` macro for consistency?

moveit_setup_assistant/moveit_setup_core_plugins/src/start_screen_widget.cpp
```diff
@@ -343,8 +351,8 @@ bool StartScreenWidget::loadExistingFiles()
   }
   catch (const std::runtime_error& e)
   {
-    QMessageBox::warning(this, "Error Loading SRDF", QString(e.what()));
     RCLCPP_ERROR(setup_step_.getLogger(), "%s", e.what());
```
sea-bass, line 354: ... and here too, including the "Error Loading SRDF" prefix?

moveit_setup_assistant/moveit_setup_framework/src/urdf_config.cpp
```diff
@@ -97,20 +97,18 @@ void URDFConfig::setPackageName()
...
-    {
-      RCLCPP_WARN(*logger_,
-                  "Package Not Found In ROS Workspace. ROS was unable to find the package name '%s'"
-                  " within the ROS workspace. This may cause issues later.",
-                  urdf_pkg_name_.c_str());
-    }
-  }
+    RCLCPP_WARN(*logger_,
+                "Package Not Found In ROS Workspace. ROS was unable to find the package name '%s'"
+                " within the ROS workspace. This may cause issues later.",
```
sea-bass, line 109: (I know this is from before, but...) This message seems redundant. Maybe just take off that first sentence?

sea-bass (commented):
This all looks good to me. Thanks for adding in all these safety rails!

I have cosmetic suggestions about the error messages, but that's about it.

sea-bass:
Oh also, you need to go through the formatting checks (`pre-commit run -a`) and also the clang-tidy CI failures shown in this job.

moveit_setup_assistant/moveit_setup_framework/src/urdf_config.cpp
chancecardona: I would argue to keep it consistent (with your other comments) that this should stay as the "subject line" as it were for the error that would go into a dialogue box if desired ever. I'm also fine just stripping away the redundant parts for all these error messages instead if so desired.

sea-bass (approved):
Thanks!

moveit_setup_assistant/moveit_setup_framework/src/urdf_config.cpp
```diff
@@ -98,18 +98,15 @@ void URDFConfig::setPackageName()
     urdf_pkg_name_ = "";
     urdf_pkg_relative_path_ = urdf_path_;  // just the absolute path
   }
-  else
+  // Check that ROS can find the package
+  const std::filesystem::path robot_desc_pkg_path = getSharePath(urdf_pkg_name_);
```
henningkayser, line 102: nit, getSharePath() could return a std::optional

moveit_setup_assistant/moveit_setup_framework/src/utilities.cpp
```diff
@@ -53,7 +53,8 @@ bool extractPackageNameFromPath(const std::filesystem::path& path, std::string&
   }
 
   // truncate path step by step and check if it contains a package.xml
-  while (!sub_path.empty())
+  // This runs until the path is either empty "" or at the root "/" or "C:\\"
```
henningkayser, line 56: ouch! great find

henningkayser (approved):
Awesome, thanks for the fix! My request about using std::optional is a nice to have, I'd go with either version.

mikeferguson (approved):

moveit_setup_assistant/moveit_setup_framework/src/urdf_config.cpp
chancecardona: I attempted this change but it ended up affecting multiple other / unrelated files that needed to do `value_or`. May attempt in a future PR.

PR #2039, opened 2023-03-23
Upgrade apt dependencies --with-new-pkgs

### LINKED ISSUE #2038: Rolling CI broken
We have again a broken CI job with a similar situation as in this issue https://github.com/ros-planning/moveit2/issues/1970, only that it affects the ci job not the ci-testing job. This time it is because the ros-rolling-base image is outdated and doesn't provide `service_msgs` as a dependency from `sensor_msgs`. Our `apt update && apt upgrade` doesn't fix the missing package, but using `apt dist-upgrade` is discouraged by our docker lint (we could suppress this rule). Additionally, there seems to be an issue with Ubuntu having had an update right in time to break the latest Rolling build, so it's unclear how soon `ros-rolling-base` will be fixed.

Using `apt upgrade --with-new-pkgs` also [seems to do the trick](https://discourse.ros.org/t/new-packages-for-ros-2-rolling-ridley-2023-03-17/30365/13).

henningkayser:
This should fix #2038, I already have a docker job running to test this

### DIFF
--- a/.docker/ci/Dockerfile
+++ b/.docker/ci/Dockerfile
@@ -18,7 +18,7 @@ COPY . src/moveit2
 RUN \
     # Update apt package list as previous containers clear the cache
     apt-get -q update && \
-    apt-get -q -y upgrade && \
+    apt-get -q -y upgrade --with-new-pkgs && \
     #
     # Install some base dependencies
     apt-get -q install --no-install-recommends -y \


.docker/ci/Dockerfile
```diff
@@ -18,7 +18,7 @@ COPY . src/moveit2
 RUN \
     # Update apt package list as previous containers clear the cache
     apt-get -q update && \
-    apt-get -q -y upgrade && \
+    apt-get -q -y upgrade --with-new-pkgs && \
```
AndyZe, line 21: Do you want to add a TODO or create an issue to remove later? Seems fine either way, to me

AndyZe (approved):

.docker/ci/Dockerfile
henningkayser: We might even want to leave it this way. I don't think there is anything wrong with always allowing to install new dependencies if they show up.

PR #2049, opened 2023-03-24
Use $DISPLAY rather than assuming :0

### LINKED ISSUE #2026: generate_move_group_launch assumes DISPLAY=:0
In moveit_configs_utils launches.py generate_move_group_launch the code assumes the users DISPLAY is :0 mine happens to be :1 which causes the following error and then crash of move_group

`[move_group-1] freeglut (<): failed to open display ':0'`

Here is a snippet of the code that overwrites the users DISPLAY environment variable.

```python
add_debuggable_node(
        ld,
        package="moveit_ros_move_group",
        executable="move_group",
        commands_file=str(moveit_config.package_path / "launch" / "gdb_settings.gdb"),
        output="screen",
        parameters=move_group_params,
        extra_debug_args=["--debug"],
        # Set the display variable, in case OpenGL code is used internally
        additional_env={"DISPLAY": ":0"},
    )
```

Removing the `additional_env` line and setting my DISPLAY environment variable correctly resolved the issue.

stephanie-eng:
### Description

Fixes #2026 
Use the $DISPLAY environmental variable rather than assuming `:0`

### DIFF
--- a/moveit_py/src/moveit/moveit_core/robot_trajectory/robot_trajectory.cpp
+++ b/moveit_py/src/moveit/moveit_core/robot_trajectory/robot_trajectory.cpp
@@ -50,6 +50,13 @@ get_robot_trajectory_msg(const robot_trajectory::RobotTrajectoryConstPtr& robot_
   return msg;
 }
 
+robot_trajectory::RobotTrajectory
+set_robot_trajectory_msg(const std::shared_ptr<robot_trajectory::RobotTrajectory>& robot_trajectory,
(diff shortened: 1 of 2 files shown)


JafarAbdi (approved):

amalnanavati:
Is it possible to get this change backported to humble? We have had to manually make this change in our site-packages folder on every computer that uses `moveit_config_utils` in humble, and then re-make that change when the package updates. It would be awesome if this could be included in the next humble package update.

PR #2060, opened 2023-03-31
Update srv_kinematics_parameters.yaml

namansharma7125:
I am Naman Sharma

### Description 
In this updated version of the code, I've added a new validation rule that uses the regex validator to check if the value of kinematics_solver_service_name is a valid ROS namespace. The regular expression ^(\/?[a-zA-Z_][\w\/]*)*$ matches strings that start with an optional forward slash (/) followed by one or more word characters or forward slashes, and which do not end with a forward slash.

The effect of this change is that the configuration file now includes a more robust validation rule for the kinematics solver service name, which helps ensure that the ROS system is configured correctly and reduces the risk of errors due to invalid service names.

### Checklist


JafarAbdi:
This's not even a supported validator in https://github.com/PickNikRobotics/generate_parameter_library#built-in-validators

PR #2064, opened 2023-03-31
Update moveit2.repos

namansharma7125:
I am Naman Sharma  

### Description
These two entries specify two Git repositories (moveit_msgs and moveit_resources) that are related to MoveIt, a popular motion planning framework for ROS. The type field specifies the type of repository (in this case Git), the url field specifies the URL of the repository, and the version field specifies the Git branch or tag to use.

Adding these dependencies means that the moveit_ros package will depend on the moveit_msgs and moveit_resources packages, which will be downloaded and built when the moveit_ros package is built.

### Checklist

### DIFF
--- a/moveit2.repos
+++ b/moveit2.repos
@@ -1,10 +1,10 @@
-repositories:
-  # Keep moveit_* repos here because they are released with moveit
-  moveit_msgs:
-    type: git
-    url: https://github.com/ros-planning/moveit_msgs.git
-    version: ros2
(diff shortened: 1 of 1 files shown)


moveit2.repos
```diff
@@ -1,10 +1,10 @@
-repositories:
```
AndyZe, line 1: I believe that removing the `repositories:` tag will break things. I'll run CI to verify it.

PR #2069, opened 2023-03-31
Update trajopt_planner_manager.cpp

namansharma7125:
I am Naman Sharma 

### Description
Added #include <ros/ros.h>: This header file was missing in the original code, and is required to use ROS_INFO and ros::NodeHandle.

Added #include <moveit_msgs/MoveItErrorCodes.h>: This header file is required to use moveit_msgs::MoveItErrorCodes.

Changed ROS_ERROR to ROS_WARN in the getPlanningContext() function: The original code used ROS_ERROR to print a warning message, which would be printed in red in the terminal. I changed it to ROS_WARN, which prints the message in yellow and is more appropriate for a warning.

Added a check to getPlanningContext() to ensure that the requested group name is valid: The original code assumed that the requested group name is always valid, but this may not always be the case. I added a check to ensure that the requested group name exists in the planning_contexts_ map before attempting to retrieve the corresponding TrajOptPlanningContext.

Added a check to getPlanningContext() to ensure that the TrajOptPlanningContext object for the requested group name has been initialized: The original code assumed that the TrajOptPlanningContext object for the requested group name has been initialized during initialize(), but this may not always be the case. I added a check to ensure that the object exists before attempting to use it.

The effects of these changes are as follows:

Adding #include <ros/ros.h> allows the code to use ROS_INFO and ros::NodeHandle.

Adding #include <moveit_msgs/MoveItErrorCodes.h> allows the code to use moveit_msgs::MoveItErrorCodes.

Changing ROS_ERROR to ROS_WARN ensures that the warning message is printed in yellow instead of red.

Adding a check to getPlanningContext() to ensure that the requested group name is valid prevents the code from attempting to use an uninitialized TrajOptPlanningContext.

Adding a check to getPlanningContext() to ensure that the TrajOptPlanningContext object for the requested group name has been initialized prevents the code from attempting to use a null PlanningContextPtr.

### Checklist

### DIFF
--- a/moveit_planners/trajopt/src/trajopt_planner_manager.cpp
+++ b/moveit_planners/trajopt/src/trajopt_planner_manager.cpp
@@ -37,104 +37,101 @@
 */
 
 #include <moveit/planning_interface/planning_interface.h>
-
 #include <moveit/collision_detection_fcl/collision_detector_allocator_fcl.h>
-
 #include <class_loader/class_loader.hpp>
-
 #include <trajopt_interface/trajopt_planning_context.h>
 
 namespace trajopt_interface
 {
-class TrajOptPlannerManager : public planning_interface::PlannerManager
-{
-public:
-  TrajOptPlannerManager() : planning_interface::PlannerManager()
+  class TrajOptPlannerManager : public planning_interface::PlannerManager
   {
-  }
-
-  bool initialize(const moveit::core::RobotModelConstPtr& model, const std::string& ns) override
-  {
-    ROS_INFO(" ======================================= initialize gets called");
-
-    if (!ns.empty())
-      nh_ = ros::NodeHandle(ns);
-    std::string trajopt_ns = ns.empty() ? "trajopt" : ns + "/trajopt";
-
-    for (const std::string& gpName : model->getJointModelGroupNames())
+  public:
+    TrajOptPlannerManager() : planning_interface::PlannerManager()
     {
-      ROS_INFO(" ======================================= group name: %s, robot model: %s", gpName.c_str(),
-               model->getName().c_str());
-      planning_contexts_[gpName] = std::make_shared<TrajOptPlanningContext>("trajopt_planning_context", gpName, model);
     }
 
-    return true;
-  }
-
-  bool canServiceRequest(const moveit_msgs::MotionPlanRequest& req) const override
-  {
-    return req.trajectory_constraints.constraints.empty();
-  }
-
-  std::string getDescription() const override
-  {
-    return "TrajOpt";
-  }
-
-  void getPlanningAlgorithms(std::vector<std::string>& algs) const override
-  {
-    algs.clear();
-    algs.push_back("trajopt");
-  }
-
-  planning_interface::PlanningContextPtr getPlanningContext(const planning_scene::PlanningSceneConstPtr& planning_scene,
-                                                            const planning_interface::MotionPlanRequest& req,
-                                                            moveit_msgs::MoveItErrorCodes& error_code) const override
-  {
-    ROS_INFO(" ======================================= getPlanningContext() is called ");
-    error_code.val = moveit_msgs::MoveItErrorCodes::SUCCESS;
-
-    if (req.group_name.empty())
+    bool initialize(const moveit::core::RobotModelConstPtr& model, const std::string& ns) override
     {
-      ROS_ERROR("No group specified to plan for");
-      error_code.val = moveit_msgs::MoveItErrorCodes::INVALID_GROUP_NAME;
-      return planning_interface::PlanningContextPtr();
+      ROS_INFO(" ======================================= initialize gets called");
+
+      if (!ns.empty())
+        nh_ = ros::NodeHandle(ns);
+
+      std::string trajopt_ns = ns.empty() ? "trajopt" : ns + "/trajopt";
+      for (const std::string& gpName : model->getJointModelGroupNames())
+      {
+        ROS_INFO(" ======================================= group name: %s, robot model: %s", gpName.c_str(),
+                 model->getName().c_str());
+        planning_contexts_[gpName] = std::make_shared<TrajOptPlanningContext>("trajopt_planning_context", gpName, model);
+      }
+
+      return true;
     }
 
-    if (!planning_scene)
+    bool canServiceRequest(const planning_interface::MotionPlanRequest& req) const override
     {
-      ROS_ERROR("No planning scene supplied as input");
-      error_code.val = moveit_msgs::MoveItErrorCodes::FAILURE;
-      return planning_interface::PlanningContextPtr();
+      return req.trajectory_constraints.constraints.empty();
     }
 
-    // create PlanningScene using hybrid collision detector
-    planning_scene::PlanningScenePtr ps = planning_scene->diff();
+    std::string getDescription() const override
+    {
+      return "TrajOpt";
+    }
 
-    // set FCL for the collision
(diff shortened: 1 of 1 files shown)


moveit_planners/trajopt/src/trajopt_planner_manager.cpp
```diff
@@ -37,104 +37,101 @@
 */
 
 #include <moveit/planning_interface/planning_interface.h>
-
 #include <moveit/collision_detection_fcl/collision_detector_allocator_fcl.h>
-
 #include <class_loader/class_loader.hpp>
-
 #include <trajopt_interface/trajopt_planning_context.h>
```
AndyZe, line 42: It would be nice if you alphabetized these includes.  :+1:

moveit_planners/trajopt/src/trajopt_planner_manager.cpp
```diff
@@ -37,104 +37,101 @@
...
 {
-class TrajOptPlannerManager : public planning_interface::PlannerManager
-{
-public:
-  TrajOptPlannerManager() : planning_interface::PlannerManager()
-  {
-  }
-
-  bool initialize(const moveit::core::RobotModelConstPtr& model, const std::string& ns) override
+  class TrajOptPlannerManager : public planning_interface::PlannerManager
```
AndyZe, line 46: When CI runs, I think the `clang format` check will fail. It automatically formats the indentation so it's consistent across all MoveIt repositories. So this change will probably need to be reverted.

moveit_planners/trajopt/src/trajopt_planner_manager.cpp
```diff
@@ -37,104 +37,101 @@
...
-      ROS_INFO(" ======================================= group name: %s, robot model: %s", gpName.c_str(),
-               model->getName().c_str());
-      planning_contexts_[gpName] = std::make_shared<TrajOptPlanningContext>("trajopt_planning_context", gpName, model);
     }
 
-    return true;
-  }
+    bool initialize(const moveit::core::RobotModelConstPtr& model, const std::string& ns) override
+    {
+      ROS_INFO(" ======================================= initialize gets called");
```
AndyZe, line 55: Suggest deleting this line. It looks like a debug print which doesn't add any value.
```suggestion
```

moveit_planners/trajopt/src/trajopt_planner_manager.cpp
```diff
@@ -37,104 +37,101 @@
...
 
-  std::string getDescription() const override
-  {
-    return "TrajOpt";
-  }
+      std::string trajopt_ns = ns.empty() ? "trajopt" : ns + "/trajopt";
+      for (const std::string& gpName : model->getJointModelGroupNames())
+      {
+        ROS_INFO(" ======================================= group name: %s, robot model: %s", gpName.c_str(),
+                 model->getName().c_str());
```
AndyZe, line 64: Same comment, this doesn't add any value.

```suggestion
```

moveit_planners/trajopt/src/trajopt_planner_manager.cpp
```diff
@@ -37,104 +37,101 @@
...
-  }
+      if (!ns.empty())
+        nh_ = ros::NodeHandle(ns);
 
-  std::string getDescription() const override
-  {
-    return "TrajOpt";
-  }
+      std::string trajopt_ns = ns.empty() ? "trajopt" : ns + "/trajopt";
+      for (const std::string& gpName : model->getJointModelGroupNames())
```
AndyZe, line 61: The naming convention for local variables is `snake_case`

```suggestion
      for (const std::string& group_name : model->getJointModelGroupNames())
```

moveit_planners/trajopt/src/trajopt_planner_manager.cpp
```diff
@@ -37,104 +37,101 @@
...
+      return true;
+    }
 
-  planning_interface::PlanningContextPtr getPlanningContext(const planning_scene::PlanningSceneConstPtr& planning_scene,
-                                                            const planning_interface::MotionPlanRequest& req,
-                                                            moveit_msgs::MoveItErrorCodes& error_code) const override
-  {
-    ROS_INFO(" ======================================= getPlanningContext() is called ");
-    error_code.val = moveit_msgs::MoveItErrorCodes::SUCCESS;
+    bool canServiceRequest(const planning_interface::MotionPlanRequest& req) const override
```
AndyZe, line 71: Why did you reorder where these functions are in the file? You should try to minimize changes unless there's a good reason for them.

moveit_planners/trajopt/src/trajopt_planner_manager.cpp
```diff
@@ -37,104 +37,101 @@
...
+      algs.push_back("trajopt");
     }
 
-    // create PlanningScene using hybrid collision detector
-    planning_scene::PlanningScenePtr ps = planning_scene->diff();
+    planning_interface::PlanningContextPtr getPlanningContext(const planning_scene::PlanningSceneConstPtr& planning_scene,
+                                                              const planning_interface::MotionPlanRequest& req,
+                                                              moveit_msgs::MoveItErrorCodes& error_code) const override
+    {
+      ROS_INFO(" ======================================= getPlanningContext() is called ");
```
AndyZe, line 91:
```suggestion
```

moveit_planners/trajopt/src/trajopt_planner_manager.cpp
```diff
@@ -37,104 +37,101 @@
...
+      }
 
-    // retrieve and configure existing context
-    const TrajOptPlanningContextPtr& context = planning_contexts_.at(req.group_name);
+      // create PlanningScene using hybrid collision detector
+      planning_scene::PlanningScenePtr ps = planning_scene->diff();
 
-    ROS_INFO(" ======================================= context is made ");
+      // set FCL for the collision
+      ps->getCollisionDetectionPlugin()->initialize("FCL");
```
AndyZe, line 112: Why this change? I don't know the difference between:

`ps->allocateCollisionDetector(collision_detection::CollisionDetectorAllocatorFCL::create());`

and

`ps->getCollisionDetectionPlugin()->initialize("FCL");`

moveit_planners/trajopt/src/trajopt_planner_manager.cpp
```diff
@@ -37,104 +37,101 @@
...
 
-    context->setPlanningScene(ps);
-    context->setMotionPlanRequest(req);
+      // retrieve and configure existing context
+      const TrajOptPlanningContextPtr& context = planning_contexts_.at(req.group_name);
 
-    error_code.val = moveit_msgs::MoveItErrorCodes::SUCCESS;
+      ROS_INFO(" ======================================= context is made ");
```
AndyZe, line 117:
```suggestion
```

AndyZe (commented):
Thanks for tidying up a bit  :+1:

Since there are some good changes mixed in with a lot of rather useless changes, I'd suggest to run `git reset --hard HEAD~` (start from a clean slate) then apply just the good, desired changes. Then push again (of course).

moveit_planners/trajopt/src/trajopt_planner_manager.cpp
henningkayser: it's also for ROS 1 logging

henningkayser (changes requested):
this mostly looks like random formatting changes. What is the purpose of these updates?

henningkayser:
I also don't see the changes you are talking about in your description.

Issue #413, opened 2021-04-11
Write a unit test for joint limit parsing


AndyZe:
A unit test for PR #408 would be great

PR #2074, opened 2023-04-01
Add extended World test for joint limits (#413)

adeeb10abbas:
Hi! First time contributor. All feedback is welcome! Thanks :) 
### Description

This PR adds an extended test case for the World class, focusing on joint limit parameters, joint velocity limits, and joint acceleration limits. The test case ensures that the RobotModelBuilder correctly sets joint limits and that the RobotState enforces these bounds as expected. Solves #413 

#### Test Case Overview
- [x] Set up a simple robot model using the RobotModelBuilder with two revolute joints and a single chain.
- [x] Define joint position, velocity, and acceleration limits for the robot model.
- [x] Load the robot model into a World object.
- [x] Create two RobotState objects with joint values, velocities, and accelerations.
- [x] Set joint values, velocities, and accelerations within the defined limits for the first RobotState.
- [x] Set joint values, velocities, and accelerations beyond the defined limits for the second RobotState.
- [x] Enforce joint position, velocity, and acceleration limits for the second RobotState.
- [x] Check that the joint values, velocities, and accelerations are within the defined limits after enforcement.
- [x] Verify that the joint limits have been applied correctly to the joint models.

### DIFF
--- a/moveit_core/collision_detection/test/test_world.cpp
+++ b/moveit_core/collision_detection/test/test_world.cpp
@@ -458,6 +458,120 @@ TEST(World, ObjectPoseAndSubframes)
   EXPECT_EQ(1.0, pose(2, 3));  // z
 }
 
+TEST(World, JointLimitParameters)
+{
+  World world;
+
+  TestAction ta;
+  World::ObserverHandle observer_ta;
+  observer_ta = world.addObserver([&ta](const World::ObjectConstPtr& object, World::Action action) {
+    return TrackChangesNotify(ta, object, action);
+  });
+
+  // Create a simple robot model
+  moveit::core::RobotModelBuilder builder("test_robot", "base_link");
+  builder.addChain("base_link->joint1->link1->joint2->link2");
+
+  // Set joint limits
+  builder.setJointLimits("joint1", { -M_PI, M_PI, 1.0, 2.0 });  // position, velocity, acceleration limits
+  builder.setJointLimits("joint2", { -M_PI / 2, M_PI / 2, 0.5, 1.0 });
+
+  moveit::core::RobotModelConstPtr robot_model = builder.build();
+
+  // Load the robot model into the world
+  world.setRobotModel(robot_model);
+
+  // Get the joint model group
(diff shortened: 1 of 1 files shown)


henningkayser:
Thank you for these test cases. Why did you choose to add them to world? The test cases don't seem related at all.

adeeb10abbas:
@henningkayser Where should they go? I wasn't entirely clear on that

moveit_core/collision_detection/test/test_world.cpp
```diff
@@ -458,6 +458,120 @@ TEST(World, ObjectPoseAndSubframes)
...
+  observer_ta = world.addObserver([&ta](const World::ObjectConstPtr& object, World::Action action) {
+    return TrackChangesNotify(ta, object, action);
+  });
+
+  // Create a simple robot model
+  moveit::core::RobotModelBuilder builder("test_robot", "base_link");
+  builder.addChain("base_link->joint1->link1->joint2->link2");
+
+  // Set joint limits
+  builder.setJointLimits("joint1", { -M_PI, M_PI, 1.0, 2.0 });  // position, velocity, acceleration limits
```
henningkayser, line 476: where did you find this member function? Also, 4 values?

moveit_core/collision_detection/test/test_world.cpp
```diff
@@ -458,6 +458,120 @@ TEST(World, ObjectPoseAndSubframes)
...
 }
 
+TEST(World, JointLimitParameters)
+{
+  World world;
+
+  TestAction ta;
+  World::ObserverHandle observer_ta;
+  observer_ta = world.addObserver([&ta](const World::ObjectConstPtr& object, World::Action action) {
+    return TrackChangesNotify(ta, object, action);
```
henningkayser, line 468: what is happening here?

henningkayser (changes requested):
Tests like these would belong next to RobotModel in moveit_core

henningkayser:
doesn't compile, code doesn't make sense and is using non-existent functions.

PR #2080, opened 2023-04-04
Compute velocity using central difference

### LINKED ISSUE #2077: Use a "central difference" when calculating velocities in Servo
For lines like this:

https://github.com/ros-planning/moveit2/blob/06e5a0b29107432c8a42e1986d356193383347b8/moveit_ros/moveit_servo/src/servo_calcs.cpp#L783

h/t to @kylc

Reference:  https://dmpeli.math.mcmaster.ca/Matlab/Math4Q3/NumMethods/Lecture3-1.html

ibrahiminfinite:
### Description

This PR changes the calculations in `ServoCalcs` to use the central difference to compute the joint velocities more accurately.

Fixes #2077 

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
@@ -278,6 +278,7 @@ protected:
   // internal_joint_state_ is used in servo calculations. It shouldn't be relied on to be accurate.
   // original_joint_state_ is the same as incoming_joint_state_ except it only contains the joints the servo node acts
   // on.
+  // last_joint_state_ caches the previous `original_joint_state_`
   sensor_msgs::msg::JointState internal_joint_state_, original_joint_state_, last_joint_state_;
   std::map<std::string, std::size_t> joint_state_name_map_;
 
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
@@ -278,7 +278,7 @@ protected:
   // internal_joint_state_ is used in servo calculations. It shouldn't be relied on to be accurate.
   // original_joint_state_ is the same as incoming_joint_state_ except it only contains the joints the servo node acts
   // on.
-  sensor_msgs::msg::JointState internal_joint_state_, original_joint_state_;
+  sensor_msgs::msg::JointState internal_joint_state_, original_joint_state_, last_joint_state_;
   std::map<std::string, std::size_t> joint_state_name_map_;
 
   // Smoothing algorithm (loads a plugin)
--- a/moveit_ros/moveit_servo/src/servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/src/servo_calcs.cpp
@@ -366,10 +366,10 @@ void ServoCalcs::calculateSingleIteration()
   // After we publish, status, reset it back to no warnings
   status_ = StatusCode::NO_WARNING;
 
-  // original_joint_state_ contains state x(t - 1)
-  // internal_joint_state_ will be updated with the state x(t + dt) in this iteration.
-  // last_joint_state_ will preserve the state x(t - dt) for this iteration to be used in central difference.
-  // original_joint_state_ will get updated with current state x(t) in updateJoints()
+  // original_joint_state_ contains state q(t - dt)
+  // internal_joint_state_ will be updated with the state q(t + dt) in this iteration.
+  // last_joint_state_ will preserve the state q(t - dt) for this iteration to be used in central difference.
+  // original_joint_state_ will get updated with current state q(t) in updateJoints()
   last_joint_state_ = original_joint_state_;
 
   // Always update the joints and end-effector transform for 2 reasons:
@@ -784,14 +784,14 @@ bool ServoCalcs::applyJointUpdate(const Eigen::ArrayXd& delta_theta, sensor_msgs
   smoother_->doSmoothing(joint_state.position);
 
   // Lambda that calculates velocity using central difference.
-  // (x(t + dt) - x(t - dt)) / ( 2 * dt )
+  // (q(t + dt) - q(t - dt)) / ( 2 * dt )
   auto compute_velocity = [&](const double& next_pos, const double& previous_pos) {
     return (next_pos - previous_pos) / (2 * parameters_->publish_period);
   };
 
   // Transform that applies the lambda to all joints.
-  // joint_state contains the future position x(t + dt)
-  // last_joint_state contains past position x(t - dt)
+  // joint_state contains the future position q(t + dt)
+  // last_joint_state contains past position q(t - dt)
   std::transform(joint_state.position.begin(), joint_state.position.end(), last_joint_state_.position.begin(),
                  joint_state.velocity.begin(), compute_velocity);
 
--- a/moveit_ros/moveit_servo/src/servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/src/servo_calcs.cpp
@@ -366,6 +366,12 @@ void ServoCalcs::calculateSingleIteration()
   // After we publish, status, reset it back to no warnings
   status_ = StatusCode::NO_WARNING;
 
+  // original_joint_state_ contains state x(t - 1)
+  // internal_joint_state_ will be updated with the state x(t + dt) in this iteration.
+  // last_joint_state_ will preserve the state x(t - dt) for next iteration to be used in for central difference.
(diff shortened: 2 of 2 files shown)


AndyZe:
My first impression from testing it out in simulation is that the behavior doesn't change noticeably. That's a good sign  :+1:

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
```diff
@@ -278,7 +278,7 @@ class ServoCalcs
   // internal_joint_state_ is used in servo calculations. It shouldn't be relied on to be accurate.
   // original_joint_state_ is the same as incoming_joint_state_ except it only contains the joints the servo node acts
   // on.
-  sensor_msgs::msg::JointState internal_joint_state_, original_joint_state_;
+  sensor_msgs::msg::JointState internal_joint_state_, original_joint_state_, last_joint_state_;
```
AndyZe, line 281: Let's add a comment above this. Here's my suggestion, feel free to modify it.
```suggestion
  // last_joint_state_ caches the previous `original_joint_state_`
  sensor_msgs::msg::JointState internal_joint_state_, original_joint_state_, last_joint_state_;
```

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -366,6 +366,12 @@ void ServoCalcs::calculateSingleIteration()
   // After we publish, status, reset it back to no warnings
   status_ = StatusCode::NO_WARNING;
 
+  // original_joint_state_ contains state x(t - 1)
+  // internal_joint_state_ will be updated with the state x(t + dt) in this iteration.
+  // last_joint_state_ will preserve the state x(t - dt) for this iteration to be used in central difference.
+  // original_joint_state_ will get updated with current state x(t) in updateJoints()
```
AndyZe, line 372: I appreciate the explanation here!

I think you should do the caching at the end of each iteration, though. That seems to flow better. Can you move this code to L513 instead?

```
  // Cache the state for a finite difference calculation in the next iteration
  // last_joint_state_ = original_joint_state_;
```

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -777,12 +783,17 @@ bool ServoCalcs::applyJointUpdate(const Eigen::ArrayXd& delta_theta, sensor_msgs
...
   smoother_->doSmoothing(joint_state.position);
 
-  for (std::size_t i = 0; i < joint_state.position.size(); ++i)
-  {
-    // Calculate joint velocity
-    joint_state.velocity[i] =
-        (joint_state.position.at(i) - original_joint_state_.position.at(i)) / parameters_->publish_period;
-  }
+  // Lambda that calculates velocity using central difference.
+  // (x(t + dt) - x(t - dt)) / ( 2 * dt )
```
AndyZe, line 787: Let's use `q(t)` and `q(t + dt)` everywhere, instead of `x(t)` and `x(t + dt)`? Because we're in the joint domain.

```suggestion
  // (q(t + dt) - q(t - dt)) / ( 2 * dt )
```

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -777,12 +783,17 @@ bool ServoCalcs::applyJointUpdate(const Eigen::ArrayXd& delta_theta, sensor_msgs
...
+  auto compute_velocity = [&](const double& next_pos, const double& previous_pos) {
+    return (next_pos - previous_pos) / (2 * parameters_->publish_period);
+  };
+
+  // Transform that applies the lambda to all joints.
+  // joint_state contains the future position x(t + dt)
+  // last_joint_state contains past position x(t - dt)
+  std::transform(joint_state.position.begin(), joint_state.position.end(), last_joint_state_.position.begin(),
+                 joint_state.velocity.begin(), compute_velocity);
```
AndyZe, line 796: This looks correct to me  :+1:

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -777,12 +783,17 @@ bool ServoCalcs::applyJointUpdate(const Eigen::ArrayXd& delta_theta, sensor_msgs
...
-  }
+  // Lambda that calculates velocity using central difference.
+  // (x(t + dt) - x(t - dt)) / ( 2 * dt )
+  auto compute_velocity = [&](const double& next_pos, const double& previous_pos) {
+    return (next_pos - previous_pos) / (2 * parameters_->publish_period);
+  };
+
+  // Transform that applies the lambda to all joints.
+  // joint_state contains the future position x(t + dt)
+  // last_joint_state contains past position x(t - dt)
```
AndyZe, line 794:
```suggestion
  // joint_state contains the future position q(t + dt)
  // last_joint_state contains past position q(t - dt)
```

AndyZe (commented):
Thanks a lot! I'll see if I can get somebody to test it on a real robot.

moveit_ros/moveit_servo/src/servo_calcs.cpp
AndyZe, line 372:
```suggestion
  // original_joint_state_ contains state q(t - dt)
  // internal_joint_state_ will be updated with the state q(t + dt) in this iteration.
  // last_joint_state_ will preserve the state q(t - dt) for this iteration to be used in central difference.
  // original_joint_state_ will get updated with current state x(t) in updateJoints()
```

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
ibrahiminfinite: Thanks . I forgot to add this.

moveit_ros/moveit_servo/src/servo_calcs.cpp
ibrahiminfinite: Makes sense , lets go with `q`

moveit_ros/moveit_servo/src/servo_calcs.cpp
ibrahiminfinite: Sophistication !

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -777,12 +783,17 @@ bool ServoCalcs::applyJointUpdate(const Eigen::ArrayXd& delta_theta, sensor_msgs
...
 
-  for (std::size_t i = 0; i < joint_state.position.size(); ++i)
-  {
-    // Calculate joint velocity
-    joint_state.velocity[i] =
-        (joint_state.position.at(i) - original_joint_state_.position.at(i)) / parameters_->publish_period;
-  }
+  // Lambda that calculates velocity using central difference.
+  // (q(t + dt) - q(t - dt)) / ( 2 * dt )
+  auto compute_velocity = [&](const double& next_pos, const double& previous_pos) {
```
sea-bass, line 788: nit, but there's no need to pass basic types like double by reference here

AndyZe: yep, it actually hurts performance (a bit...)

ibrahiminfinite: Would it not be faster than pass by value which makes a copy ?

AndyZe: it's a complex question but our rule of thumb is what Sebastian commented

https://blog.bytellect.com/software-development/cplusplus/which-is-faster-in-cplusplus-call-by-value-call-by-reference-or-call-by-pointer/

ibrahiminfinite: Oh nvm, it is slower for basic types, due to unnecessary extra memory access.

sea-bass: tl;dr the address for basic types basically uses the same memory as the payload itself, so pass-by-reference only pays off when the data types are nontrivial

ibrahiminfinite: Might also be good to restrict the variables captured by the lambda to just the 3 arrays it access?

AndyZe: I don't think that will matter from a performance standpoint

AndyZe (approved):

ibrahiminfinite:
Is this still in testing?

AndyZe:
Waiting for a second "maintainer" to review it. It worked well when I tried it.

sjahr (approved):

PR #2084, opened 2023-04-05
Fix: TOTG can return vels/accels greater than the limits

AndyZe:
### Description

This is a port of the MoveIt1 PR:  https://github.com/ros-planning/moveit/pull/3394

The trajectory in the new test looks like this:

Position:

position

Velocity:

velocity

Acceleration:

acceleration

### DIFF
--- a/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
+++ b/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
@@ -831,7 +831,6 @@ Eigen::VectorXd Trajectory::getVelocity(double time) const
   const double acceleration =
       2.0 * (it->path_pos_ - previous->path_pos_ - time_step * previous->path_vel_) / (time_step * time_step);
 
-  time_step = time - previous->time_;
   const double path_pos =
       previous->path_pos_ + time_step * previous->path_vel_ + 0.5 * time_step * time_step * acceleration;
   const double path_vel = previous->path_vel_ + time_step * acceleration;
@@ -849,7 +848,6 @@ Eigen::VectorXd Trajectory::getAcceleration(double time) const
   const double acceleration =
       2.0 * (it->path_pos_ - previous->path_pos_ - time_step * previous->path_vel_) / (time_step * time_step);
 
-  time_step = time - previous->time_;
   const double path_pos =
       previous->path_pos_ + time_step * previous->path_vel_ + 0.5 * time_step * time_step * acceleration;
   const double path_vel = previous->path_vel_ + time_step * acceleration;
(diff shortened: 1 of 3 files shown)


moveit_core/robot_trajectory/test/test_robot_trajectory.cpp
```diff
@@ -538,7 +538,7 @@ TEST_F(RobotTrajectoryTestFixture, RobotTrajectorySmoothness)
 
   const auto smoothness = robot_trajectory::smoothness(*trajectory);
   ASSERT_TRUE(smoothness.has_value());
-  EXPECT_GT(smoothness.value(), 0.0);
+  EXPECT_EQ(smoothness.value(), 0.0);
```
AndyZe, line 541: @sjahr is this change OK? It seems correct, to me, because `trajectory` contains 5 repeated waypoints, right?

sjahr (commented):
Can you elaborate a bit more on what this fix does? Which effect does removing` time_step = time - previous->time_;` have? The test makes sense, although I am wondering what the vel/accel profile for a single-point trajectory looks like if we apply the constraint that both need to be 0 at the end. Does TOTG insert additional points to "interpolate" between the two points?

AndyZe:
Sure. There was a comment on the original MoveIt1 PR but I'll expand it here.

> although I am wondering what the vel/accel profile for a single-point trajectory looks like if we apply the constraint that both need to be 0 at the end. Does TOTG insert additional points to "interpolate" between the two points?

Yes it does! I added a plot of the test trajectory to the description

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -849,7 +848,6 @@ Eigen::VectorXd Trajectory::getAcceleration(double time) const
   const double acceleration =
       2.0 * (it->path_pos_ - previous->path_pos_ - time_step * previous->path_vel_) / (time_step * time_step);
 
-  time_step = time - previous->time_;
```
AndyZe, line 852: Notice that a different `time_step` is used previously at L848. The `time_step` at L848 is exactly at one of the waypoints that TOTG used in calculation so vel/accel limits aren't exceeded.

The `time_step` at L852 is not exactly on one of the waypoints that TOTG used in calculation and that's why it exceeds the vel/accel limits.

There will be a very small loss in accuracy by removing this but it should be fine if TOTG is run with a small `delta_t`. Since the default `delta_t` is 0.001sec, that's certainly true.

sjahr: Thanks for the explanation!

sjahr (approved):

PR #2086, opened 2023-04-05
Seperating a calculation from an action in collision_env

TaMaN2031A:
Separating calculations from actions produces reusable, and more testable code. 
Function setScale contains a calculation. This calculation can be separated into an independent calculation. This calculation iterates over scales of links. It compares them with the to-be-set scale. In case of inequality, it pushes this link into a vector. This vector, called u, will be used in updatePaddingOrScaling. 
Same behavior of setPadding.
### Description

### Checklist

### DIFF
--- a/.codespell_words
+++ b/.codespell_words
@@ -13,3 +13,5 @@ planed
 tork
 uint
 whis
+sinic
+padd
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@
 repos:
   # Standard hooks
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v3.4.0
+    rev: v4.4.0
     hooks:
       - id: check-added-large-files
       - id: check-case-conflict
@@ -33,7 +33,7 @@ repos:
       - id: trailing-whitespace
 
   - repo: https://github.com/psf/black
-    rev: 22.3.0
+    rev: 23.3.0
     hooks:
       - id: black
 
@@ -47,8 +47,8 @@ repos:
         files: \.(c|cc|cxx|cpp|frag|glsl|h|hpp|hxx|ih|ispc|ipp|java|js|m|proto|vert)$
         args: ['-fallback-style=none', '-i']
   - repo: https://github.com/codespell-project/codespell
-    rev: v2.0.0
+    rev: v2.2.4
     hooks:
       - id: codespell
-        args: ['--write-changes', '--ignore-words=.codespell_words']
+        args: ['--write-changes', '--ignore-words=.codespell_words', '--skip="*.eps"']
         exclude: CHANGELOG.rst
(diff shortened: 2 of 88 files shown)


sjahr (changes requested):
I don't see how this refactors the action to a calculation. `compareWithScale` still uses the member `robot_model_`

TaMaN2031A:
> I don't see how this refactors the action to a calculation. `compareWithScale` still uses the member `robot_model_`

Does sending robot_model_ as a parameter to this function with the scale-or padding as this function could be used with setPadding also, solve this problem? Or the problem is bigger than this, if yes I'd be very grateful if you explain it to me. 
And thank you for your time.

TaMaN2031A:
> > I don't see how this refactors the action to a calculation. `compareWithScale` still uses the member `robot_model_`
> 
> Does sending links as a parameter to this function with the scale-or padding as this function could be used with setPadding also, solve this problem? Or the problem is bigger than this, if yes I'd be very grateful if you explain it to me. And thank you for your time.

Something like this: 
`std::vector<std::string> CollisionEnv::compareWithScale(const double scale,
                                                        const std::vector<const moveit::core::LinkModel*>& links)
{
  std::vector<std::string> u;
  for (const auto link : links)
  {
    if (getLinkScale(link->getName()) != scale)
      u.push_back(link->getName());
  }
  return u;
}

void CollisionEnv::setScale(const double scale)
{
  if (!validateScale(scale))
    return;
  std::vector<std::string> u;
  const std::vector<const moveit::core::LinkModel*>& links = robot_model_->getLinkModelsWithCollisionGeometry();
  for (const auto link : links)
  {
    link_scale_[link->getName()] = scale;
  }
  u = compareWithScale(scale, links);
  if (!u.empty())
    updatedPaddingOrScaling(u);
}`

TaMaN2031A:
> > I don't see how this refactors the action to a calculation. `compareWithScale` still uses the member `robot_model_`
> 
> Does sending links as a parameter to this function with the scale-or padding as this function could be used with setPadding also, solve this problem? Or the problem is bigger than this, if yes I'd be very grateful if you explain it to me. And thank you for your time.

Something like this: https://drive.google.com/file/d/1fHmlrq7lXFU9qfZoDeqAZzGDZ5v0vU1X/view?usp=sharing

TaMaN2031A:
@sjahr Can you review the pull request?

TaMaN2031A:
@sjahr Could you review the code?

sjahr:
I am sorry, but while this technically extracts a calculation from the function, there is no use in doing so. This simple comparison doesn't need testing nor is it re-used, so the only effect of this change is that the number of lines of code for this functionality blows up. Whether or not these changes make sense highly depends on the context and keeping a balance between applying them and keeping the code in a good structure is important. If you're further interested in contributing to MoveIt I'd recommend fixing one of the many open issues. The ones labeled as "good first issue" are in particular good for beginners. Good luck!

TaMaN2031A:
> I am sorry, but while this technically extracts a calculation from the function, there is no use in doing so. This simple comparison doesn't need testing nor is it re-used, so the only effect of this change is that the number of lines of code for this functionality blows up. Whether or not these changes make sense highly depends on the context and keeping a balance between applying them and keeping the code in a good structure is important. If you're further interested in contributing to MoveIt I'd recommend fixing one of the many open issues. The ones labeled as "good first issue" are in particular good for beginners. Good luck!

 I'll go for these issues as soon as possible inshallah. Thank you very much, Mr. Sebastian!

PR #2091, opened 2023-04-07
Change filter coefficient to ROS param

### LINKED ISSUE #2089: Convert the Servo filter coefficient to a ROS parameter
Instead of hard-coding this filter coefficient, it should be a ROS parameter.

https://github.com/ros-planning/moveit2/blob/4d3420cca312b51073d78df909293380fe73a2ff/moveit_core/online_signal_smoothing/src/butterworth_filter.cpp#L108

ibrahiminfinite:
### Description

This PR makes the Butterworth filter coefficient into a ROS param.

Fixes #2089 

### Checklist


moveit_core/online_signal_smoothing/src/butterworth_filter.cpp
```diff
@@ -38,6 +38,7 @@
 
 #include <moveit/online_signal_smoothing/butterworth_filter.h>
 #include <rclcpp/clock.hpp>
+#include <rclcpp/rclcpp.hpp>
```
Abishalini, line 41: I think this can be removed. `rclcpp.hpp` is an expensive header in terms of compilation time.

Abishalini (commented):
I have one comment. Otherwise this PR looks good to me. Thanks!

moveit_core/online_signal_smoothing/src/butterworth_filter.cpp
ibrahiminfinite: Thanks ! forgot to remove that one after including it for some debugging.

ibrahiminfinite:
This should be done now.

moveit_core/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ endforeach()
 set(LIBFCL_LIBRARIES "${LIBFCL_LIBRARIES_FULL}")
 
 find_package(ament_cmake REQUIRED)
+find_package(generate_parameter_library REQUIRED)
```
AndyZe, line 19: nitpick: please alphabetize

moveit_core/CMakeLists.txt
```diff
@@ -139,6 +141,7 @@ install(
 
 ament_export_targets(moveit_coreTargets HAS_LIBRARY_TARGET)
 ament_export_dependencies(
+  generate_parameter_library
```
AndyZe, line 144: nit: please alphabetize

moveit_core/online_signal_smoothing/CMakeLists.txt
```diff
@@ -28,16 +28,26 @@ target_include_directories(moveit_butterworth_filter PUBLIC
...
 )
 ament_target_dependencies(moveit_butterworth_filter
   srdfdom  # include dependency from moveit_robot_model
 )
 
 # Installation
-
+install(
+  TARGETS butterworth_parameters
+  EXPORT ${PROJECT_NAME}Targets
```
AndyZe, line 46: I hope @tylerjw or @ChrisThrasher will stop by to say if this was done correctly. It's a little different from what was done here, for example:

https://github.com/ros-controls/ros2_controllers/blob/fa7499f0eb28ccca11408e703b8d5fce4f4deb30/joint_trajectory_controller/CMakeLists.txt#L97

Abishalini (approved):

moveit_core/online_signal_smoothing/src/butterworth_filter.cpp
```diff
@@ -100,12 +100,12 @@ bool ButterworthFilterPlugin::initialize(rclcpp::Node::SharedPtr node, moveit::c
 {
   node_ = node;
   num_joints_ = num_joints;
+  param_listener_ = std::make_shared<online_signal_smoothing::ParamListener>(node_);
```
AndyZe, line 103: Since `param_listener_` is only used in this one place, it doesn't need to be a member variable. Just make it a local variable here.

moveit_core/online_signal_smoothing/src/butterworth_parameters.yaml
```diff
@@ -0,0 +1,9 @@
+online_signal_smoothing:
+  butterworth_filter_coeff: {
+        type: double,
+        default_value: 1.5,
+        description: "Filter coefficient for the Butterworth filter",
+        validation: {
+          gt<>: 1.0
```
AndyZe, line 7: Thanks for adding this check.

moveit_core/online_signal_smoothing/CMakeLists.txt
ibrahiminfinite: Would be great to get some feedback on this point.
The reference I followed : https://github.com/ros-planning/moveit2/blob/9d6ceb20d3f9046b287a1e4540af8aa73199e2e0/moveit_kinematics/CMakeLists.txt#L74

moveit_core/online_signal_smoothing/CMakeLists.txt
```diff
@@ -28,16 +28,26 @@ target_include_directories(moveit_butterworth_filter PUBLIC
 set_target_properties(moveit_butterworth_filter PROPERTIES VERSION
   "${${PROJECT_NAME}_VERSION}"
 )
+
+generate_parameter_library(butterworth_parameters src/butterworth_parameters.yaml)
```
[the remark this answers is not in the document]
ChrisThrasher: Yes let's please use the `moveit_` prefix on all of our build targets.

moveit_core/online_signal_smoothing/CMakeLists.txt
ChrisThrasher: This is probably not doing what you want. To make CMake targets visible to downstream packages there are a few things you need to do of which you're doing one of them. The first step is creating an export set which is what `install(TARGETS ... EXPORT` is doing. `${PROJECT_NAME}Targets` is your export set. This is a collection of CMake targets that you want to install. Alongside creating an export set you specify where various binaries from that export set get installed which you're already doing.

You also have to install that export set via `install(EXPORT` then create a config module which includes the file that `install(EXPORT` generates. I'm guessing that this machinery already exists elsewhere in moveit_core and that you may jus need to list this target in some other export set to automatically get that behavior.

I'm not 100% sure how moveit_core architectures its install nor am I 100% how the ament CMake wrappers ~~obscure~~ simplify this so you may not need to explicitly write those other two things I mentioned if ament is automagically doing that for you.

---

As an aside, please don't use `${PROJECT_NAME}` here though. It puts extra burden on those reading the code and makes it much harder to `grep` for things wherever it is used. It's analogous to using a C++ macro to specify a variable name instead simply spelling out that variable name each time you need it.

ChrisThrasher (changes requested):

moveit_core/online_signal_smoothing/src/butterworth_filter.cpp
```diff
@@ -100,12 +100,12 @@ bool ButterworthFilterPlugin::initialize(rclcpp::Node::SharedPtr node, moveit::c
 {
   node_ = node;
   num_joints_ = num_joints;
+  auto param_listener_ = std::make_unique<online_signal_smoothing::ParamListener>(node_);
```
ibrahiminfinite, line 103: The parameter is only read and set once when `initialize()` is called.
Is this the expected behaviour? 
or should this also check for updates in the parameter and reset the coefficient ?

moveit_core/online_signal_smoothing/src/butterworth_filter.cpp
```diff
@@ -100,12 +100,12 @@ bool ButterworthFilterPlugin::initialize(rclcpp::Node::SharedPtr node, moveit::c
...
+  auto filter_coeff = param_listener_->get_params().butterworth_filter_coeff;
```
AndyZe, line 104:
```suggestion
  auto param_listener = std::make_unique<online_signal_smoothing::ParamListener>(node_);
  auto filter_coeff = param_listener->get_params().butterworth_filter_coeff;
```

moveit_core/online_signal_smoothing/CMakeLists.txt
```diff
@@ -28,16 +28,19 @@ target_include_directories(moveit_butterworth_filter PUBLIC
 set_target_properties(moveit_butterworth_filter PROPERTIES VERSION
   "${${PROJECT_NAME}_VERSION}"
 )
+
+generate_parameter_library(moveit_butterworth_parameters src/butterworth_parameters.yaml)
+
 target_link_libraries(moveit_butterworth_filter
   moveit_smoothing_base
   moveit_robot_model
+  moveit_butterworth_parameters
```
AndyZe, line 37:
```suggestion
  moveit_butterworth_parameters
  moveit_robot_model
  moveit_smoothing_base
```

AndyZe (approved):
It works! Thanks for your patience with the review process.

ibrahiminfinite:
Happy to contribute!

PR #2092, opened 2023-04-07
Add JointModel::satisfiesAccelerationBounds()

AndyZe:
### Description

This is a port of this MoveIt1 PR:  https://github.com/ros-planning/moveit/pull/3396

We already have versions for satisfiesPositionBounds() and satisfiesVelocityBounds(), why not Acceleration and Jerk too?

### DIFF
--- a/moveit_core/robot_model/src/joint_model.cpp
+++ b/moveit_core/robot_model/src/joint_model.cpp
@@ -130,6 +130,16 @@ bool JointModel::satisfiesVelocityBounds(const double* values, const Bounds& oth
   return true;
 }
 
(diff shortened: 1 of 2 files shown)


moveit_core/robot_model/src/joint_model.cpp
```diff
@@ -118,6 +118,10 @@ bool JointModel::satisfiesVelocityBounds(const double* values, const Bounds& oth
 {
   for (std::size_t i = 0; i < other_bounds.size(); ++i)
   {
+    if (!other_bounds[i].velocity_bounded_)
+    {
+      continue;
+    }
```
AndyZe, line 124: Not having this previously was a bug

PR #2096, opened 2023-04-08
Update Moveit Servo to use generate_parameter_library

### LINKED ISSUE #2095: Use generate_parameter_library with MoveIt Servo
Currently the paraments in MoveIt Servo are loaded using a custom parameter library, it will be better to make use of the generate_parameter_library  to load the servo parameters in a more cleaner way. And also adhere to a consistent way of loading parameters in MoveIt.

ibrahiminfinite:
### Description

This PR modifies Moveit Servo to use the `generate_parameter_library` to load the servo parameters.
Fixes #2095 

There are still things to be done, but I wanted to get some initial feedback on this.

TODO : 
The old parameter library is still used in parallel at the moment due to a dependency on  `registerSetParameterCallback()`
- [ ] Alternative to handle `registerSetParameterCallback()`  method 
- [ ] Remove the usages of `moveit_servo::ServoParameters`
- [ ] Fix pose tracking demo
- [ ] Fix `moveit2_tutorials` cpp_interface_demo
- [ ] validations for some parameters that require comparison against other parameters. 
- [ ] Fix the tests to use `generate_parameter_library`
- [ ] Cleanup debug stuff

Questions : 
- How should the overriding of default  values be done ?

### Checklist


Abishalini:
Thanks for the initiative! I will review the PR as you progress.

Just an initial feedback to get the formatting CI check passing - install pre-commit
Here are instructions on how to run pre-commit -  https://moveit.ros.org/documentation/contributing/code/#pre-commit-formatting-checks

ibrahiminfinite:
There is a ` make_shared_from_pool.h` file in the package which does not seem to be used anywhere.
Is that there for some future purpose ?
If not , I can just remove it in this PR itself.

moveit_ros/moveit_servo/src/servo_parameters.yaml
```diff
@@ -0,0 +1,261 @@
...
+    type: string,
+    default_value: "/joint_states",
+    description: "The topic on which joint states can be monitored"
+  }
+
+  smoothing_filter_plugin_name: {
+    type: string,
+    default_value: "online_signal_smoothing::ButterworthFilterPlugin",
+    description: "The name of the smoothing plugin to be used"
+  }
```
sea-bass, line 119: I recently learned that the smoothing filter coefficient was hard-coded in the code to be 1.5. Could you make this a parameter in this PR too, if that's possible?

ibrahiminfinite:
> Please do remove it, if possible 👍

Turns out it is used in `pose_tracking`
I did not notice this earlier since I had disabled the pose_tracking demos.
```
  // use the shared pool to create a message more efficiently
  auto msg = moveit::util::make_shared_from_pool<geometry_msgs::msg::TwistStamped>();
```
@AndyZe  , could this be replaced with a `geometry_msgs::msg::TwistStamped` member variable which is reused ?

AndyZe:
Can you please create an issue about it and we'll decide later. It doesn't matter for this PR, in fact it would be nice to keep it separate from this PR.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -317,38 +301,68 @@ void ServoCalcs::stop()
...
     {
       input_cv_.wait(main_loop_lock, [this] { return (new_input_cmd_ || stop_requested_); });
     }
 
     // reset new_input_cmd_ flag
     new_input_cmd_ = false;
 
+    // Check if any parameters changed
+    // TODO : Remove debug info here
+    if (servo_param_listener_->is_old(servo_params_))
```
sea-bass, line 322: To answer your top-level question about what to do about parameters: Many other nodes that use this library basically have a boolean parameter that configures whether the node should be continuously checking for parameter updates on every tick vs. just initialize once and not check again. For performance reasons.

an example of this would be [the ros2_control admittance controller](https://github.com/ros-controls/ros2_controllers/blob/master/admittance_controller/src/admittance_controller_parameters.yaml#L155-L159), where the use of this parameter would be [here](https://github.com/ros-controls/ros2_controllers/blob/master/admittance_controller/include/admittance_controller/admittance_rule_impl.hpp#L169-L172).

moveit_ros/moveit_servo/src/servo_parameters.yaml
ibrahiminfinite: There is a PR going to make that a parameter .
https://github.com/ros-planning/moveit2/pull/2091
Is this what you are referring to ?

moveit_ros/moveit_servo/src/servo_calcs.cpp
ibrahiminfinite: Thanks !
Will add that here as well .

moveit_ros/moveit_servo/src/servo_parameters.yaml
ibrahiminfinite: @sea-bass  #2114  and #2091  have taken care of this.

ibrahiminfinite:
At present , the parameters do not have a "moveit_servo" prefix like before.
`rotational_scale` instead of `moveit_servo.rotational_scale`.
Should this prefix be kept , or is this something that can be dropped since the parameters are already in a node namspace.

ibrahiminfinite:
Will rebase this once #2121 is merged.

moveit_ros/moveit_servo/package.xml
```diff
@@ -23,6 +23,7 @@
   <buildtool_depend>ament_cmake</buildtool_depend>
   <depend>moveit_common</depend>
 
+  <depend>generate_parameter_library</depend>
```
sea-bass, line 26: Add this in alphabetical order, so before `geometry_msgs`

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -72,7 +75,8 @@ void Servo::start()
   servo_calcs_.start();
 
   // Check collisions in this timer
-  if (parameters_->check_collisions)
+  RCLCPP_INFO(LOGGER, "SERVO Check Collision : %d", servo_params_.check_collisions);
```
sea-bass, line 78: Seems to be a lingering debug statement that should probably be removed

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -73,52 +73,33 @@ int const THREAD_PRIORITY = 40;
...
 // Constructor for the class that handles servoing calculations
 ServoCalcs::ServoCalcs(const rclcpp::Node::SharedPtr& node,
-                       const std::shared_ptr<const moveit_servo::ServoParameters>& parameters,
-                       const planning_scene_monitor::PlanningSceneMonitorPtr& planning_scene_monitor)
+                       const planning_scene_monitor::PlanningSceneMonitorPtr& planning_scene_monitor,
+                       std::shared_ptr<const servo::ParamListener>& servo_param_listener)
   : node_(node)
-  , parameters_(parameters)
+  , servo_param_listener_(servo_param_listener)
+  , servo_params_(servo_param_listener_->get_params())
```
sea-bass, line 80: If this is in the constructor, does it mean that none of the parameters can be dynamically updated at runtime?

EDIT: Ah, never mind, I see `updateParams()` now.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -319,40 +300,77 @@ void ServoCalcs::stop()
   }
 }
 
+void ServoCalcs::updateParams()
```
sea-bass, line 303: not necessarily for this PR, but is this where we would choose to add more parameters that we would like to be dynamically settable in future?

If so, worth saying that in the header file docstring for this?

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -319,40 +300,77 @@ void ServoCalcs::stop()
...
     }
 
     // reset new_input_cmd_ flag
     new_input_cmd_ = false;
 
+    // Check if any parameters changed
+    if (servo_params_.enable_parameter_update)
+    {
+      updateParams();
+    }
```
sea-bass, line 356: Shall we move this to the very top of the function? Because let's say we eventually have `low_latency_mode` also be responsive to this function, it wouldn't take effect til the next tick.

moveit_ros/moveit_servo/src/servo_parameters.yaml
```diff
@@ -0,0 +1,350 @@
...
+    default_value: "/joint_states",
+    description: "The topic on which joint states can be monitored"
+  }
+
+################################ GENERAL CONFIG #############################
+  enable_parameter_update: {
+    type: bool,
+    default_value: false,
+    description: "If true, continuous check for parameter update is enabled"
+  }
```
sea-bass, line 98: What needs to be clearly communicated here is that right now, only 2 parameters are actually updated!

sea-bass (commented):
Took another pass through and only saw a few small things.

moveit_ros/moveit_servo/src/servo_calcs.cpp
ibrahiminfinite: > not necessarily for this PR, but is this where we would choose to add more parameters that we would like to be dynamically settable in future?

yes , that is right.
Will add that information in the docstring. Thanks!

sea-bass:
I will also say that you can run `clang-tidy` locally so you don't have to wait for CI to get feedback:

First, make sure you build your workspace with the `compile-commands` mixin:

```
colcon build --mixin compile-commands
```

Then, from that same workspace root:

```
run-clang-tidy -p build/ src/
```

moveit_ros/moveit_servo/package.xml
ibrahiminfinite: Does the `moveit_common` also need to be moved ?
Or is that put there on purpose ?

sea-bass: Not sure. It seems to be at the top for most moveit packages, so maybe just leave that one there?

ibrahiminfinite: Did not see any issues after I moved it.
Is there anyone who can clarify on this?
Feels weird to leave it like that now.

AndyZe: I'd be surprised if order mattered.

Personally, I'd prefer alphabetized but no strong opinion

sea-bass: I don't think there's any issues with rearranging things in `package.xml`. Typically they are alphabetized, but I'm sure there was a decision to have `moveit_common` at the top for some visibility reason. Don't worry about it.

ibrahiminfinite:
There are a couple of checks from `validate()`  method of old parameter library that are left out now.
Not sure if those can be done in the YAML file itself. 
If not I can add a `validate()` in this one as well which is called once during construction and then from within `updateParams()` at the end, to ensure parameters values are still valid after an update.

moveit_ros/moveit_servo/CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ set(THIS_PACKAGE_INCLUDE_DEPENDS
 )
 
 find_package(ament_cmake REQUIRED)
+find_package(generate_parameter_library REQUIRED)
```
AndyZe, line 27: alphabetize plz

moveit_ros/moveit_servo/include/moveit_servo/collision_check.h
```diff
@@ -47,7 +47,7 @@
 #include <sensor_msgs/msg/joint_state.hpp>
 #include <std_msgs/msg/float64.hpp>
 
-#include <moveit_servo/servo_parameters.h>
+#include <moveit_servo_lib_parameters.hpp>
```
AndyZe, line 50:
```suggestion
// Auto-generated
#include <moveit_servo_lib_parameters.hpp>
```

moveit_ros/moveit_servo/include/moveit_servo/collision_check.h
```diff
@@ -86,8 +86,8 @@ class CollisionCheck
   // Pointer to the ROS node
   const std::shared_ptr<rclcpp::Node> node_;
 
-  // Parameters from yaml
-  const ServoParameters::SharedConstPtr parameters_;
+  // The servo parameters
+  servo::Params servo_params_;
```
AndyZe, line 90: Nitpick: it's pretty obvious what these are from the name I think, the comment isn't needed.

```suggestion
  servo::Params servo_params_;
```

moveit_ros/moveit_servo/include/moveit_servo/pose_tracking.h
```diff
@@ -87,7 +87,7 @@ class PoseTracking
 {
 public:
   /** \brief Constructor. Loads ROS parameters under the given namespace. */
-  PoseTracking(const rclcpp::Node::SharedPtr& node, const ServoParameters::SharedConstPtr& servo_parameters,
+  PoseTracking(const rclcpp::Node::SharedPtr& node, std::shared_ptr<const servo::ParamListener>& servo_param_listener,
```
AndyZe, line 90: Any reason to have `shared_ptr` or can it be a `unique_ptr`?

moveit_ros/moveit_servo/include/moveit_servo/servo.h
```diff
@@ -54,8 +54,9 @@ namespace moveit_servo
 class Servo
 {
 public:
-  Servo(const rclcpp::Node::SharedPtr& node, const ServoParameters::SharedConstPtr& parameters,
-        const planning_scene_monitor::PlanningSceneMonitorPtr& planning_scene_monitor);
+  Servo(const rclcpp::Node::SharedPtr& node,
+        const planning_scene_monitor::PlanningSceneMonitorPtr& planning_scene_monitor,
+        std::shared_ptr<const servo::ParamListener>& servo_param_listener);
```
AndyZe, line 59: I wonder if we should be concerned with thread safety here? B/c at least 2 threads will have access to this ParamListener at once (PoseTracker and ServoCalcs).

@tylerjw does ParamListener happen to be threadsafe?

moveit_ros/moveit_servo/src/collision_check.cpp
```diff
@@ -49,21 +49,21 @@ constexpr size_t ROS_LOG_THROTTLE_PERIOD = 30 * 1000;  // Milliseconds to thrott
 namespace moveit_servo
 {
 // Constructor for the class that handles collision checking
-CollisionCheck::CollisionCheck(const rclcpp::Node::SharedPtr& node, const ServoParameters::SharedConstPtr& parameters,
+CollisionCheck::CollisionCheck(const rclcpp::Node::SharedPtr& node, const servo::Params& servo_params,
                                const planning_scene_monitor::PlanningSceneMonitorPtr& planning_scene_monitor)
   : node_(node)
-  , parameters_(parameters)
+  , servo_params_(servo_params)
```
AndyZe, line 55: It's good that you're creating a copy of these parameters so we don't have to worry about thread safety.

moveit_ros/moveit_servo/src/servo_parameters.yaml
```diff
@@ -0,0 +1,353 @@
...
+  command_out_type: {
+    type: string,
+    default_value: "trajectory_msgs/JointTrajectory",
+    description: "The type of command that servo will publish",
+    validation: {
+      one_of<>: [["trajectory_msgs/JointTrajectory", "std_msgs/Float64MultiArray"]]
+    }
+  }
+
+  num_outgoing_halt_msgs_to_publish: {
```
AndyZe, line 57: I believe this parameter actually isn't being used anymore. Please do `grep -r num_outgoing_halt_msgs_to_publish` then delete any references to it.

moveit_ros/moveit_servo/test/servo_launch_test_common.hpp
```diff
@@ -58,8 +58,8 @@
 #include <gtest/gtest.h>
 
 // Servo
-#include <moveit_servo/servo_parameters.h>
 #include <moveit_servo/status_codes.h>
+#include <moveit_servo_lib_parameters.hpp>
```
AndyZe, line 62:
```suggestion
// Auto-generated
#include <moveit_servo_lib_parameters.hpp>
```

moveit_ros/moveit_servo/include/moveit_servo/pose_tracking.h
ibrahiminfinite: We are passing the `ParamListener` when creating the `Servo` object.
`unique_ptr` reference might be fine as long as the original copy stays in scope.
Which is the case for pose tracking demo.
Not sure if there will be a case were the `ParamListener` is created in a function and passed and the function goes out of scope and destroys the `ParamListener` `unique_ptr`

moveit_ros/moveit_servo/include/moveit_servo/servo.h
AndyZe: I agree with you about thread safety  :+1:  It seems fine since `servo_calcs.cpp` is the only thread that will be updating at runtime.

ibrahiminfinite: > So, the shared_ptr is always passed by reference. That means, the reference count isn't incremented

Even though it is passed by reference to `Servo` , this is passsed to `ServoCalcs` and  a copy of the `ParamListener`  is created there as `ServoCalcs` needs to have its own copy to check for updates.
If I am correct, the copy of `shared_ptr` [here](https://github.com/ibrahiminfinite/moveit2/blob/4b1083c8b841624462438818ce2a40733ae4cd08/moveit_ros/moveit_servo/src/servo_calcs.cpp#L79) increments the reference count.

ibrahiminfinite: I looked into it a bit more and turns out we never really needed the `shared_ptr`.
Changed the order of some of the initializations and that removed the need for a `shared_ptr`.
Now there is only a `unqiue_ptr` passed (`std::move`) through all the classes till it reaches `ServoCalcs` 
All classes except `ServoCalcs` only need a copy of `servo::Params` and not the `ParamListener` itself.

AndyZe: that's really great!

AndyZe (approved):

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
```diff
@@ -92,6 +92,17 @@ class ServoCalcs
    */
   void start();
 
+  /***
+   * Validate the parameter values
+   */
+  void validateParams();
```
sea-bass, line 98: I missed why we needed this function when `generate_parameter_library` already has parameter validators.

Might be worth saying in the comment that this validates not the individual parameter, but any additional parameters that are interdependent (like the self vs scene collision threshold, etc.)

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -319,17 +303,86 @@ void ServoCalcs::stop()
   }
 }
 
+void ServoCalcs::validateParams()
+{
+  if (servo_params_.hard_stop_singularity_threshold <= servo_params_.lower_singularity_threshold)
+  {
+    RCLCPP_ERROR(LOGGER, "Parameter 'hard_stop_singularity_threshold' "
+                         "should be greater than 'lower_singularity_threshold.' "
+                         "Check yaml file.");
```
sea-bass, line 312:
```suggestion
                         "Check the parameters YAML file used to launch this node.");
```

Same with the others.

sea-bass (approved):
LGTM as well. I don't have anything meaningful, and maybe one comment that's more of a question.

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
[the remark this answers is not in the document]
ibrahiminfinite: Good catch! Thanks
In the [old implementation](https://github.com/ros-planning/moveit2/blob/acbf3770763e58e4f86ce0dea251a50cd4987fe6/moveit_ros/moveit_servo/src/servo_node.cpp#L81) the behaviour was to throw an error.
This was what I indented as well but looks like I forgot to `throw` the error and only printed it.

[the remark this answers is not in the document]
ibrahiminfinite: Maybe it is better to move to this validation to `ServoNode` instead ?
There is no reason to delay the validation till the construction of `ServoCalcs`

moveit_ros/moveit_servo/src/servo_node.cpp
```diff
@@ -106,6 +108,50 @@ ServoNode::ServoNode(const rclcpp::NodeOptions& options)
...
+  {
+    RCLCPP_WARN(LOGGER, "Parameter 'self_collision_proximity_threshold' should probably be less "
+                        "than or equal to 'scene_collision_proximity_threshold'."
+                        "Check the parameters YAML file used to launch this node.");
+    has_error = true;
+  }
+
+  if (has_error)
+  {
+    throw std::runtime_error("Failed to initialize servo");
```
sea-bass, line 151: If this throws, then the logger messages above should either be `RCLCPP_ERROR` or `RCLCPP_FATAL`, not warnings, right?

ibrahiminfinite: Not sure about that, it was like this in the old implementation.
So I followed that pattern.

moveit_ros/moveit_servo/include/moveit_servo/servo_node.h
```diff
@@ -63,6 +63,12 @@ class ServoNode
   std::shared_ptr<tf2_ros::Buffer> tf_buffer_;
   std::shared_ptr<planning_scene_monitor::PlanningSceneMonitor> planning_scene_monitor_;
 
+  /***
+   * \brief Most of parameters are validated using the validation methods provided by the generate_paramter_library
+   * This method validates only the parameters that need to be checked against another parameter
```
sea-bass, line 68: Recommend taking some extra time in documenting things and checking for typos and punctuation -- there is no hurry to turn around feedback, and it takes longer to go back and forth multiple times in the long run.

This is just an idea:

```suggestion
   * \brief Most servo parameters are individually validated using the validation methods in `generate_parameter_library`, with limits specified in the parameters YAML file.
   * This method performs additional validation for parameters whose values depend on each other. 
```

ibrahiminfinite: Got it, will take care.
Thanks.

moveit_ros/moveit_servo/src/servo_node.cpp
ibrahiminfinite: Should the failed checks be considered as errors ?
or can we have a single error message at the end before throwing?
Stating servo failed to initialize due to invalid parameter values

sea-bass: You can definitely do that. Start with an empty `std::string` and append any of the logger messages to it, then if you throw an error you tack that string on there.

sea-bass: Heads-up, clang-tidy may yell at you if you use the `+` operator for this. So do:

```
std::string error_text = "";

if (some_bad_thing_happened)
{
    my_error = "Everything is broken.";
    RCLCPP_ERROR(LOGGER, my_error.c_str());
    my_error.append(error_text).append("\n");
    has_error = true;
}

if (has_error)
{
    throw std::runtime_error("Failed with reason: ".append(error_text));
}
```

ibrahiminfinite: Feels like this would make it less readable.
Since multiple checks could fail and the error message will get quite big (huge multi-line error message)
I think we can convert all the warnings to errors instead and just throw with a `Servo failed to initialize : Invalid parameters` message.

sea-bass: That's fine too so long as the errors are being logged

ibrahiminfinite:
@sea-bass @AndyZe 
I just ran the tested locally with one parameter set to invalid value.
The test just keeps running indefinitely. 
We might need a better mechanism to handle the validation failure.

ibrahiminfinite:
> Another question is, if the default values in the `servo_parameters.yaml` should be set to something generic rather than the panda specific values like `panda_link8` Since the values are overridden anyway

Bringing this up again since I just found out the tests were passing because of the default values and not loading the config properly.

AndyZe:
Ideally all these parameters that are specific to the robot would NOT have defaults so they throw an error if not configured. Thanks for taking the time to address these issues instead of sweeping them under the rug.

> move_group_name, planning_frame, ee_frame_name, robot_link_command_frame

ibrahiminfinite:
> Ideally all these parameters that are specific to the robot would NOT have defaults so they throw an error if not configured.

The tests fail properly now if the parameters are not specified during launch.

ibrahiminfinite:
Maybe the parameters that use default values should be removed from the configs in the `config` folder to keep things cleaner ?
So that the file only contains params that are overridden or those that do not have a default value.

moveit_ros/moveit_servo/include/moveit_servo/servo_node.h
```diff
@@ -62,6 +63,13 @@ class ServoNode
   std::shared_ptr<tf2_ros::Buffer> tf_buffer_;
   std::shared_ptr<planning_scene_monitor::PlanningSceneMonitor> planning_scene_monitor_;
 
+  /***
+   * \brief Most servo parameters are individually validated using the validation methods in
+   * `generate_parameter_library`, with limits specified in the parameters YAML file. This method performs additional
+   * validation for parameters whose values depend on each other.
+   */
+  void validateParams(const servo::Params);
```
AndyZe, line 71: CI suggests this

```suggestion
  void validateParams(const servo::Params&);
```

moveit_ros/moveit_servo/src/servo_node.cpp
```diff
@@ -108,7 +105,51 @@ ServoNode::ServoNode(const rclcpp::NodeOptions& options)
   }
 
   // Create Servo
-  servo_ = std::make_unique<moveit_servo::Servo>(node_, servo_parameters, planning_scene_monitor_);
+  servo_ = std::make_unique<moveit_servo::Servo>(node_, planning_scene_monitor_, std::move(param_listener));
+}
+
+void ServoNode::validateParams(const servo::Params servo_params)
```
AndyZe, line 111:
```suggestion
void ServoNode::validateParams(const servo::Params& servo_params)
```

AndyZe (approved):

AndyZe:
Just waiting on a review from a second maintainer now...

sea-bass (approved):
Looks good to me as well -- let's do this. Thanks, @ibrahiminfinite!

ibrahiminfinite:
Thanks for the guidance!

moveit_ros/moveit_servo/include/moveit_servo/servo_node.h
```diff
@@ -62,6 +63,13 @@ class ServoNode
   std::shared_ptr<tf2_ros::Buffer> tf_buffer_;
   std::shared_ptr<planning_scene_monitor::PlanningSceneMonitor> planning_scene_monitor_;
 
+  /***
+   * \brief Most servo parameters are individually validated using the validation methods in
+   * `generate_parameter_library`, with limits specified in the parameters YAML file. This method performs additional
+   * validation for parameters whose values depend on each other.
+   */
+  void validateParams(const servo::Params&);
```
sea-bass, line 71:
```suggestion
  void validateParams(const servo::Params& servo_params);
```

ibrahiminfinite:
> @ibrahiminfinite this broke the tutorials from building, would you mind updating the [tutorials site](https://github.com/ros-planning/moveit2_tutorials)?
> 
> See [ros-planning/moveit2_tutorials#684](https://github.com/ros-planning/moveit2_tutorials/issues/684)

Yea sure .

develiberta:
I'm waiting the  moveit2_tutorials update

Issue #2090, opened 2023-04-06
The `margin` argument to satisfiesVelocityBounds() is reversed


AndyZe:
### Description

See https://github.com/ros-planning/moveit2/blob/4d3420cca312b51073d78df909293380fe73a2ff/moveit_core/robot_model/src/joint_model.cpp#L121

```
bool JointModel::satisfiesVelocityBounds(const double* values, const Bounds& other_bounds, double margin) const
{
  for (std::size_t i = 0; i < other_bounds.size(); ++i)
  {
    if (other_bounds[i].max_velocity_ + margin < values[i])
    {
      return false;
```

I think this should be:

```
bool JointModel::satisfiesVelocityBounds(const double* values, const Bounds& other_bounds, double margin) const
{
  for (std::size_t i = 0; i < other_bounds.size(); ++i)
  {
    if (other_bounds[i].max_velocity_ margin < values[i] + margin)
    {
      return false;
```

AndyZe:
OK, thanks. For the reference of others, the PR is #2097.

Issue #121, opened 2019-11-07
Port all package dependencies


henningkayser:
I've given you access to our forked package dependencies which are partially migrated but still need some fixup (see warnings and upstream PRs).
https://github.com/PickNikRobotics/geometric_shapes
https://github.com/PickNikRobotics/object_recognition_msgs
https://github.com/PickNikRobotics/octomap
https://github.com/PickNikRobotics/octomap_msgs
https://github.com/PickNikRobotics/srdfdom

@RoboticsYY Could you start fixing these up and filing new PRs? That would be really helpful.

RoboticsYY:
@henningkayser Thanks for giving me access! I have fixed up some warnings of moveit_core, I will continue to fix the warnings in these dependency packages as well.

RoboticsYY:
Submitted PRs to the upstream projects:

- [ ] object_recognition_msgs: [PR](https://github.com/wg-perception/object_recognition_msgs/pull/8)
- [ ] octomap_msgs: [PR](https://github.com/OctoMap/octomap_msgs/pull/13)

PR #2097, opened 2023-04-09
fix: issue #2090

SuyashSapkal:
Updated moveit_core/robot_model/src/joint_model.cpp at line #121

### Description

### Checklist


AndyZe:
Unfortunately the `margin` argument is used in lots of other places, too. So this PR will need to be bigger in scope. Before diving into the coding, is there a consensus that it is indeed backward?  @henningkayser  @tylerjw   @sjahr

tylerjw:
I'm confused about what the usefulness of this change is.

AndyZe:
Don't you think the margin should be subtracted from `max_velocity` instead of added to it? And also, is this margin actually used by anyone?

SuyashSapkal:
from the code below:

`bool JointModel::satisfiesVelocityBounds(const double* values, const Bounds& other_bounds, double margin) const
{
  for (std::size_t i = 0; i < other_bounds.size(); ++i)
  {
    if (other_bounds[i].max_velocity_ + margin < values[i])
    {
      return false;
    }
    else if (other_bounds[i].min_velocity_ - margin > values[i])
    {
      return false;
    }
  }
  return true;
}`

I am assuming the function should return `true` 
if `other_bounds[i].min_velocity_ - margin < values[i] < other_bounds[i].min_velocity_ + margin`.

Currently,
it's returning `false`.

Is that the issue?

moveit_core/robot_model/src/joint_model.cpp
```diff
@@ -118,7 +118,7 @@ bool JointModel::satisfiesVelocityBounds(const double* values, const Bounds& oth
 {
   for (std::size_t i = 0; i < other_bounds.size(); ++i)
   {
-    if (other_bounds[i].max_velocity_ + margin < values[i])
+    if (other_bounds[i].max_velocity_ < values[i] + margin)
```
henningkayser, line 121: I would prefer this version
```suggestion
    if (other_bounds[i].max_velocity_ - margin < values[i])
```

henningkayser (commented):
The reasoning makes sense to me, a safety margin should define a range "before" the limit and not an additional tolerance. Please fix the min_velocity case as well. Also, appropriate documentation of the result seems needed as well.

AndyZe:
@henningkayser what about removing this `margin` argument completely? It's used in lots of places throughout the codebase, not just here. I'm convinced it was a hack which allows small overshoots from time parameterization. Which we should not be having any longer.

moveit_core/robot_model/src/joint_model.cpp
SuyashSapkal: Hi @AndyZe and @henningkayser ,
let me know if I should go ahead with this change?
or 
wait and try to work on removing the `margin` argument completely?

AndyZe: It would be great if you could work on removing `margin` arguments completely. I would just do it in this same PR and we can rename the PR.

I expect there will be some CI failures when the change is made. (In fact, I tried it myself and saw the CI failures). But it will be good to see these issues get exposed.

AndyZe: @SuyashSapkal here is a list of all usage of `margin` I could find in the codebase. I found this with `grep -r margin`. It's a pretty big list, let me know if you're up for it.

[use_of_margin.txt](https://github.com/ros-planning/moveit2/files/11215979/use_of_margin.txt)

SuyashSapkal: ok will work on this and get back

henningkayser:
> @henningkayser what about removing this `margin` argument completely? It's used in lots of places throughout the codebase, not just here. I'm convinced it was a hack which allows small overshoots from time parameterization. Which we should not be having any longer.

we might certainly break some things by fixing the margin behavior. If you think the currently implemented use cases are hacks, I would agree with removing support for it.

SuyashSapkal:
Hi @AndyZe and @henningkayser ,

I did a quick search for `satisfiesVelocityBounds(` using find command in the whole repo and found it being present in the below files with their lines in which it is being used.

1. moveit_core/robot_model/include/moveit/robot_model/joint_model.h
	311:  bool satisfiesVelocityBounds(const double* values, double margin = 0.0) const
	313:    return satisfiesVelocityBounds(values, variable_bounds_, margin);
	317:  virtual bool satisfiesVelocityBounds(const double* values, const Bounds& other_bounds, double margin) const;
	
2. moveit_core/robot_model/src/joint_model.cpp
	117:bool JointModel::satisfiesVelocityBounds(const double* values, const Bounds& other_bounds, double margin) const
	
3. moveit_core/robot_state/include/moveit/robot_state/robot_state.h
	1552:    return satisfiesPositionBounds(joint, margin) && (!has_velocity_ || satisfiesVelocityBounds(joint, margin));
	1558:  bool satisfiesVelocityBounds(const JointModel* joint, double margin = 0.0) const
	1560:    return joint->satisfiesVelocityBounds(getJointVelocities(joint), margin);  

4. moveit_planners/pilz_industrial_motion_planner/src/joint_limits_aggregator.cpp
	187:  if (!joint_model->satisfiesVelocityBounds(&joint_limit.max_velocity))

I am assuming I need to make changes in all these lines for removing `margin` parameter.
Let me know what changes should I start making.

AndyZe:
Please remove it everywhere, not just `satisfiesVelocityBounds()`. Here's a list.

[use_of_margin.txt](https://github.com/ros-planning/moveit2/files/11226179/use_of_margin.txt)

SuyashSapkal:
I have made the changes as per the list given by you. Except for these 3 lines

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 14">
<link id=Main-File rel=Main-File
href="file:///C:\Users\suyas\AppData\Local\Temp\msohtmlclip1\01\clip.htm">
<link rel=File-List
href="file:///C:\Users\suyas\AppData\Local\Temp\msohtmlclip1\01\clip_filelist.xml">
<style>

</style>
</head>

<body link=blue vlink=purple>



moveit_core/kinematic_constraints/test/test_constraints.cpp | // if different, test margin behavior
-- | --
moveit_core/kinematic_constraints/test/test_constraints.cpp | // but they are within   this margin
moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/bullet_integration/bullet_utils.h | void setMargin(btScalar /*margin*/) override



</body>

</html>

AndyZe:
@SuyashSapkal can I have write access to your fork of `moveit2` please? I'll see what I can do to fix up these CI failures.

SuyashSapkal:
I can see that it's failing the CI tests.
How can I work on the changes so that it passes the CI tests?

AndyZe:
If you click `Details` for the failed CI jobs, then expand `Run industrial ci` -- you'll see this:

 /home/runner/work/moveit2/moveit2/.work/target_ws/src/moveit2/moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_bounds.cpp:136:40: error: no matching function for call to ‘moveit::core::RobotState::satisfiesBounds(const moveit::core::JointModel*&, const double&)’
    136 |         if (start_state.satisfiesBounds(jmodel, bounds_dist_))
    
That particular error is self-explanatory, just remove `bounds_dist_` since that argument is gone.

AndyZe:
We're waiting on this PR:

https://github.com/ros-planning/moveit2/pull/2111

AndyZe:
It would also be wise to create a backup branch of your changes, with a name other than `main`. Possibly call it `remove_margin_arg` or something similar.

`git checkout -b remove_margin_arg`
`git push origin remove_margin_arg`

SuyashSapkal:
do you want me to create a backup branch? and push the changes back to my forked `main` branch?

SuyashSapkal:
> remove_margin_arg

ok sure. I have created a branch with name `remove_margin_arg` will open PR with that

PR #2106, opened 2023-04-13
[Servo] Change variable names in `ServoCalcs` for improved readability

ibrahiminfinite:
### Description

This PR renames variables within `ServoCalcs` class to improve readability.

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
@@ -199,7 +199,7 @@ protected:
    * @param previous_vel Eigen vector of previous velocities being updated
    * @return Returns false if there is a problem, true otherwise
    */
-  bool applyJointUpdate(const Eigen::ArrayXd& delta_theta, sensor_msgs::msg::JointState& joint_state);
+  bool applyJointUpdate(const Eigen::ArrayXd& delta_theta, sensor_msgs::msg::JointState& next_joint_state);
 
   /** \brief Gazebo simulations have very strict message timestamp requirements.
    * Satisfy Gazebo by stuffing multiple messages into one.
@@ -274,12 +274,12 @@ protected:
 
   moveit::core::RobotStatePtr current_state_;
 
-  // (mutex protected below)
-  // internal_joint_state_ is used in servo calculations. It shouldn't be relied on to be accurate.
-  // original_joint_state_ is the same as incoming_joint_state_ except it only contains the joints the servo node acts
-  // on.
-  // last_joint_state_ caches the previous `original_joint_state_`
-  sensor_msgs::msg::JointState internal_joint_state_, original_joint_state_, last_joint_state_;
+  // There variables are mutex protected
+  // previous_joint_state holds the state q(t - 1)
+  // current_joint_state holds the  state q(t)
+  // next_joint_state holds the computed state q(t + 1)
+
+  sensor_msgs::msg::JointState previous_joint_state_, current_joint_state_, next_joint_state_;
   std::map<std::string, std::size_t> joint_state_name_map_;
 
   // Smoothing algorithm (loads a plugin)
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
@@ -132,9 +132,6 @@ protected:
   /** \brief Do servoing calculations for direct commands to a joint. */
   bool jointServoCalcs(const control_msgs::msg::JointJog& cmd, trajectory_msgs::msg::JointTrajectory& joint_trajectory);
 
-  /** \brief Parse the incoming joint msg for the joints of our MoveGroup */
-  void updateJoints();
-
   /**
    * Checks a JointJog msg for valid (non-NaN) velocities
    * @param cmd the desired joint servo command
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
@@ -272,9 +272,9 @@ protected:
   moveit::core::RobotStatePtr current_state_;
 
   // These variables are mutex protected
-  // previous_joint_state holds the state q(t - 1)
-  // current_joint_state holds the  state q(t)
-  // next_joint_state holds the computed state q(t + 1)
+  // previous_joint_state holds the state q(t - dt)
+  // current_joint_state holds the  state q(t) as retrieved from the planning scene monitor.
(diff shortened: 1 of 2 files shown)


ibrahiminfinite:
@AndyZe I find the names more fitting in context now. What do you think ?

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -790,10 +791,10 @@ bool ServoCalcs::applyJointUpdate(const Eigen::ArrayXd& delta_theta, sensor_msgs
   };
 
   // Transform that applies the lambda to all joints.
-  // joint_state contains the future position q(t + dt)
-  // last_joint_state contains past position q(t - dt)
-  std::transform(joint_state.position.begin(), joint_state.position.end(), last_joint_state_.position.begin(),
-                 joint_state.velocity.begin(), compute_velocity);
+  // next_joint_state contains the future position q(t + dt)
+  // previous_joint_state_ contains past position q(t - dt)
```
AndyZe, line 795: This comment isn't quite right. We're only iterating forward by one `dt`. I think it should be:

```
  // next_joint_state contains the future position q(t)
  // previous_joint_state_ contains past position q(t - dt)
```

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
```diff
@@ -274,12 +271,12 @@ class ServoCalcs
...
-  // (mutex protected below)
-  // internal_joint_state_ is used in servo calculations. It shouldn't be relied on to be accurate.
-  // original_joint_state_ is the same as incoming_joint_state_ except it only contains the joints the servo node acts
-  // on.
-  // last_joint_state_ caches the previous `original_joint_state_`
-  sensor_msgs::msg::JointState internal_joint_state_, original_joint_state_, last_joint_state_;
+  // These variables are mutex protected
+  // previous_joint_state holds the state q(t - 1)
+  // current_joint_state holds the  state q(t)
+  // next_joint_state holds the computed state q(t + 1)
```
AndyZe, line 277: We're only iterating forward by one `dt` in each cycle so this comment isn't quite right.

```suggestion
  // current_joint_state holds the  current state as retrieved from the planning scene monitor
  // next_joint_state holds the computed state q(t)
```

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -366,19 +370,18 @@ void ServoCalcs::calculateSingleIteration()
...
-  updateJoints();
-
-  // Update from latest state
+  // Get the latest joint group positions
   current_state_ = planning_scene_monitor_->getStateMonitor()->getCurrentState();
+  current_state_->copyJointGroupPositions(joint_model_group_, current_joint_state_.position);
+  current_state_->copyJointGroupVelocities(joint_model_group_, current_joint_state_.velocity);
+
+  // copy current state to temp state to use for calculating next state
+  // This is done so that current_state_ is preserved and can be used as backup.
```
AndyZe, line 382:
```suggestion
  // This is done so that current_joint_state_ is preserved and can be used as backup.
```

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -366,19 +370,18 @@ void ServoCalcs::calculateSingleIteration()
...
+  // All computations related to computing state q(t + 1) acts only on next_joint_state_ variable.
```
AndyZe, line 383:
```suggestion
  // All computations related to computing state q(t) acts only on next_joint_state_ variable.
```

AndyZe (commented):
Here are some comments. I want to test it still. Thanks for the PR!

AndyZe:
I didn't notice any regressions while testing  :+1:

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
ibrahiminfinite: Oh yea, the `current_joint_state` makes more sense this way.
But `next_joint_state` is still  going to contain the expected future state  `q(t+dt)` right?

AndyZe: That's fair. OK, just change `current_joint_state` and I'll be happy.

AndyZe:
@ibrahiminfinite could you also change the function signature of `applyJointUpdate()` to be like this:

`bool applyJointUpdate(const Eigen::ArrayXd& delta_theta, const sensor_msgs::msg::JointState& previous_joint_state, sensor_msgs::msg::JointState& next_joint_state);`

Reason being, it's weird to pass in `next_joint_state_` but not `previous_joint_state_`.

At that point, it doesn't have to be a member function anymore so you could move that function to `utilities.cpp`

ibrahiminfinite:
>At that point, it doesn't have to be a member function anymore so you could move that function to utilities.cpp

I am not sure about this.
The computations in `applyJointUpdate()`  are important steps in the servo loop. So I am not in favour of moving it out.
That would make it more harder for people to follow the code.

A better idea might be to just  replace the function call to `applyJointUpdate()` with the content. 
This will in addition to making the code more readable ,  avoid an [extra check](https://github.com/ibrahiminfinite/moveit2/blob/cf542c5045afbb0822b8962362c9a8fb83d085fe/moveit_ros/moveit_servo/src/servo_calcs.cpp#L728) and reduce the number of return statements, since the function only used once internally anyway.

Either way, I think this should be done in a separate PR since it is not related to the changes from renaming.

AndyZe:
> A better idea might be to just replace the function call to `applyJointUpdate()` with the content.

This would be one giant, monolithic block of code if we didn't have some helper functions to break it up and make it more readable! That function is fairly long, too. It's nice to break it out.

> Either way, I think this should be done in a separate PR since it is not related to the changes from renaming.

Fine, I'll make an issue:  https://github.com/ros-planning/moveit2/issues/2120

AndyZe:
We'll have to wait on this PR to get CI passing:  https://github.com/ros-planning/moveit2/pull/2111

AndyZe:
@ibrahiminfinite if you can rebase (at your leisure, no rush) CI should pass now

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
```diff
@@ -199,7 +196,7 @@ class ServoCalcs
    * @param previous_vel Eigen vector of previous velocities being updated
```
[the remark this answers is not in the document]
AndyZe: cool, thanks!

AndyZe (approved):

sea-bass (approved):

PR #2114, opened 2023-04-14
[Servo] Document the new low-pass filter param

AndyZe:
### Description

Add the new parameter to a launch file as a means of documentation

@ibrahiminfinite can you please review?

### DIFF
--- a/moveit_ros/moveit_servo/launch/servo_example.launch.py
+++ b/moveit_ros/moveit_servo/launch/servo_example.launch.py
@@ -143,10 +143,10 @@ def generate_launch_description():
         executable="servo_node_main",
         parameters=[
             servo_params,
+            low_pass_filter_coeff,
             moveit_config.robot_description,
             moveit_config.robot_description_semantic,
             moveit_config.robot_description_kinematics,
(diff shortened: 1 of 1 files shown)


moveit_ros/moveit_servo/launch/servo_example.launch.py
```diff
@@ -144,6 +146,7 @@ def generate_launch_description():
             moveit_config.robot_description,
             moveit_config.robot_description_semantic,
             moveit_config.robot_description_kinematics,
+            low_pass_filter_coeff,
```
ibrahiminfinite, line 149: How about we put this just after `servo_params` , feels easy to miss when looking through otherwise.

ibrahiminfinite (approved):

moveit_ros/moveit_servo/launch/servo_example.launch.py
AndyZe: good point, will do

ibrahiminfinite (approved):

sea-bass (approved):

Abishalini (approved):

Issue #859, opened 2021-12-01
Do not pass and return simple types by const ref


tylerjw:
https://github.com/cpp-best-practices/cppbestpractices/blob/master/04-Considering_Safety.md#do-not-pass-and-return-simple-types-by-const-ref

Nils-ChristianIseke:
I made the necessary changes:
https://github.com/ros-planning/moveit2/pull/2047
Regarding this PR i have two more questions:
1. How to handle variables that are const ref?
2. Doesn't it make sense to replace not only const references of primitive data types, but also references without const? If that is desired, I can also do that quickly.

Nils-ChristianIseke:
Made the necessary changes: https://github.com/ros-planning/moveit2/pull/2057.

PR #2119, opened 2023-04-17
 Do not pass and return simple types by const ref #859 

Nils-ChristianIseke:
### Description

Removed const references of simple types with regex.

### Checklist
- [x ] **Required by CI**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)


henningkayser (commented):
CI fails for unrelated formatting reasons https://github.com/ros-planning/moveit2/actions/runs/4862228141/jobs/8670864807?pr=2119#step:12:9677. I'm fine with these changes in general

tylerjw (commented):
I'm also fine with these changes. I do think having `const` in parameters where it is meaningless should also be removed.

Here is what I mean. Take this function you want to improve:
```c++
void do_magic(const int& input_number);
```

If you change it to this (just remove the &) you leave `const` which is meaningless in modifying the `int`. Previously it modified the way the `&` worked but that is no longer there:
```c++
void do_magic(const int input_number);
```

A much better change in my opinion is to also remove the const keyword in cases where we took a `const&` and made it a value parameter like this:
```c++
void do_magic(int input_number);
```

Nils-ChristianIseke:
Thanks for the suggestions @tylerjw. What are the advantages of not using const? Is it really bad to use a const, if it is not necessary?

gavanderhoorn:
Without the reference, the function is getting a copy. The rationale is that if you're passing a copy already, `const` doesn't really make any difference any more, as the "original" can't be modified anyway.

The function/method could still change its argument, but that would not affect the variable/value that was passed to it.

Hence no need for `const`.

(this is not the only opinion about this btw. Some developers consider the `const` to also say something about the semantics of the function, as in: this argument is used in a read-only manner only. I don't have any opinion, but did want to point out that it's not necessarily the case there's full consensus about these kinds of things)

Nils-ChristianIseke:
Thanks @gavanderhoorn and @tylerjw for the detailled explanation, that was exactly what i was looking for. If it is consensus to remove the const keyword, i can make the change as soon as i find the time (hopefully within 1 week).

tylerjw:
> Thanks @gavanderhoorn and @tylerjw for the detailled explanation, that was exactly what i was looking for. If it is consensus to remove the const keyword, i can make the change as soon as i find the time (hopefully within 1 week).

Yes, please remove const where it is meaningless (value parameters).

Shobuj-Paul:
The clang-tidy warnings in CI will be fixed by #2343
This PR can be rebased and merged right after.

Shobuj-Paul:
@sjahr @tylerjw 
If no further work is being done on this PR, should I take over and finish it according to the above recommendations?

tylerjw:
@Shobuj-Paul you are welcome to take over this work and finish it. Ping me for reviews.

sjahr:
Merged with #2453

Nils-ChristianIseke:
Thanks for catching up on this issue @Shobuj-Paul.

PR #2121, opened 2023-04-18
[Servo] Make `applyJointUpdate()` a free function

### LINKED ISSUE #2120: [Servo] Make `applyJointUpdate()` into a free function
The function signature should look like this and it should go in `utilities.h`

`bool applyJointUpdate(const Eigen::ArrayXd& delta_theta, const sensor_msgs::msg::JointState& previous_joint_state, sensor_msgs::msg::JointState& next_joint_state);`

ibrahiminfinite:
### Description

Fixes #2120 
Should be merged after #2106 

This PR make the `applyJointUpdate()` method of `ServoCalcs` a free function and moves it to `utilities`.
This is done in order to keep the core servo code cleaner and enable better testing.

### Checklist


sjahr (approved):

moveit_ros/moveit_servo/src/utilities.cpp
```diff
@@ -144,4 +144,44 @@ double velocityScalingFactorForSingularity(const moveit::core::JointModelGroup*
...
+{
+  // All the sizes must match
+  if (next_joint_state.position.size() != static_cast<std::size_t>(delta_theta.size()) ||
+      next_joint_state.velocity.size() != next_joint_state.position.size())
+  {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wold-style-cast"
+    RCLCPP_ERROR_STREAM_THROTTLE(LOGGER, clock, ROS_LOG_THROTTLE_PERIOD,
+                                 "Lengths of output and increments do not match.");
+#pragma GCC diagnostic pop
```
AndyZe, line 160: This GCC stuff is new to me. What is this?

moveit_ros/moveit_servo/include/moveit_servo/utilities.h
```diff
@@ -72,4 +73,18 @@ double velocityScalingFactorForSingularity(const moveit::core::JointModelGroup*
...
+ * @param delta_theta Eigen vector of joint delta's
+ * @param previous_joint_state The previous joint state
+ * @param next_joint_state The joint state object which will hold the next joint state.
+ * @param smoother The trajectory smoother to be used.
+ * @return Returns false if there is a problem, true otherwise
+ */
+bool applyJointUpdate(rclcpp::Clock& clock, const double publish_period, const Eigen::ArrayXd& delta_theta,
+                      const sensor_msgs::msg::JointState& previous_joint_state,
+                      sensor_msgs::msg::JointState& next_joint_state,
+                      std::shared_ptr<online_signal_smoothing::SmoothingBaseClass>& smoother);
```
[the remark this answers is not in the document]
ibrahiminfinite: That sounds reasonable.
In that case we would also need to replace [this](https://github.com/ros-planning/moveit2/blob/d343ed5b793c4bdb207b9df02200d60c9c4ccc68/moveit_ros/moveit_servo/src/servo_calcs.cpp#L176) , with `createUniqueInstance()` instead?
Since this is only used within `ServoCalcs` and only one copy is needed.

moveit_ros/moveit_servo/src/utilities.cpp
ibrahiminfinite: This supresses the warning
```
 use of old-style cast to ‘rcutils_duration_value_t’ {aka ‘long int’} [-Wold-style-cast]
 RCLCPP_ERROR_STREAM_THROTTLE(LOGGER, clock, ROS_LOG_THROTTLE_PERIOD
```
However a proper fix would be to do `static_cast`
```
constexpr auto ROS_LOG_THROTTLE_PERIOD = static_cast<rcutils_duration_value_t>(std::chrono::milliseconds(3000).count());
```

AndyZe: best go for the proper fix. There are other compilers besides GCC  :+1:

ibrahiminfinite: The fix has been applied , removed the older warning suppressions as well.

ibrahiminfinite: This suppression is present in other files as well.
Might need to raise an issue for this.

ibrahiminfinite: Unfortunately casting it does not seem to fix the issue.
[This](https://gitlab.com/autowarefoundation/autoware.auto/AutowareAuto/-/issues/955) issue seems to be deeper.

AndyZe: ok, feel free to ignore my comment

moveit_ros/moveit_servo/include/moveit_servo/utilities.h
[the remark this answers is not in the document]
ibrahiminfinite: Used a `pluginlib::UniquePtr` here since that is what `createUniqueInstance()` returns.

AndyZe (approved):

moveit_ros/moveit_servo/src/utilities.cpp
ibrahiminfinite: Have changed the warnings to global within `utilities`  rather than per call to keep things a bit cleaner.

Issue #2103, opened 2023-04-12
Constraint Planning Segfault


MarqRazz:
### Description

I have been working on a problem where I need orientation constraints to be held when transporting an object and have been observing segmentation faults in MoveIt. In my setup I'm also trying to build databases to speed up planning but when I enable [multi_query_planning_enabled](https://moveit.picknik.ai/main/doc/examples/ompl_interface/ompl_interface_tutorial.html?highlight=multi_query_planning_enabled#persistent-roadmaps) motion planning crashes.

### Your environment
* ROS Distro: [Humble & Rolling]
* OS Version: e.g. Ubuntu 22.04
* Source build
* Cyclone DDS

### Steps to reproduce
I have made a working [docker container with instructions](https://github.com/MarqRazz/moveit2_tutorials/tree/pr-constraint_segfault_example#debug-constraint-planning-segmentation-fault) that includes some details on where I have been seeing the faults. In this container I have modified [ompl_constrained_planning_tutorial](https://github.com/MarqRazz/moveit2_tutorials/blob/pr-constraint_segfault_example/doc/how_to_guides/using_ompl_constrained_planning/src/ompl_constrained_planning_tutorial.cpp) to include some collision objects around the robot and to attempt to find motion plans forever because the fault happens on the second or third attempt. I'm asking for multiple plans because in my actual application I'm using MTC and this happens when my constraint stage has a parent with multiple solutions.

### Expected behaviour
I should be able to run the modified tutorial example without MoveIt crashing.

### Actual behaviour
Program crashes with a segfault.

### Backtrace or Console output

[segfault backtrace](https://gist.github.com/MarqRazz/4fa972f736746913c105f6be899e1cad)

henningkayser:
@MarqRazz I think I've found what is broken in OMPL. However, I don't have a great suggestion for a fix but maybe a better workaround.

[...]

PR #2130, opened 2023-04-21
Fix Constraint Planning Segfault

MarqRazz:
### Description
This PR is an attempt to understand and hopefully fix a segfault I have been running into when asking MoveIt to solve constrained motion as reported in #2103.

While debugging we have found that the `ompl::base::PlannerPtr` generated on the first planning request is going out of scope causing the next planning request to fault the program. As you can see in my initial commit I have removed the ability for the planning allocator to reuse the planner between requests which fixes the segfault. This is not ideal if the user is attempting to build a [roadmap database](https://moveit.picknik.ai/humble/doc/examples/ompl_interface/ompl_interface_tutorial.html#persistent-roadmaps) because it would only contain the map of a single plan if written to disk, or thrown away on the next request. The planner is created [on this line](https://github.com/MarqRazz/moveit2/blob/b19e25adc724b27ffef350313bd8f666946916c8/moveit_planners/ompl/ompl_interface/src/planning_context_manager.cpp#L190) by passing it the `SpaceInformationPtr` that is created [here](https://github.com/MarqRazz/moveit2/blob/b19e25adc724b27ffef350313bd8f666946916c8/moveit_planners/ompl/ompl_interface/src/planning_context_manager.cpp#L375-L378). If I'm understanding OMPL correctly this `SpaceInformationPtr` is being moved to the Planner object [here](https://github.com/ompl/ompl/blob/main/src/ompl/base/src/Planner.cpp#L44-L49) and the `ConfiguredPlannerAllocator` containing it should be kept alive/in context by the [known_planners_](https://github.com/ros-planning/moveit2/blob/acbf3770763e58e4f86ce0dea251a50cd4987fe6/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/planning_context_manager.h#L232) object of the Planning Context Manager.

I would really appreciate any help in identifying why the pointer, or objects inside of it, is going out of scope of if my understanding of the program is incorrect! As I have been learning this section of the software I have also wondered if the current implementation is worth fixing / there is a better way to preserve the planning scene between motion requests. Because the planner is constructed with the initial planning request goal's [path_constraint](https://github.com/MarqRazz/moveit2/blob/b19e25adc724b27ffef350313bd8f666946916c8/moveit_planners/ompl/ompl_interface/src/planning_context_manager.cpp#L366-L367) does this mean that if it is reused it could have invalid constraints specified if they change between requests?


henningkayser:
Please see my comment here https://github.com/ros-planning/moveit2/issues/2103#issuecomment-1523720752

mamoll:
If this fixes the crash, then this seems like a useful workaround that should be merged. Recreating a planner instance from planner data adds a bit of overhead, but shouldn't be too bad. I'd add a comment along these lines so that we don't forget why this is done:
```cpp
// FIXME: make reusing PlannerPtr not crash, so that we don't have to reconstruct a PlannerPtr instance
```

moveit_planners/ompl/ompl_interface/src/planning_context_manager.cpp
```diff
@@ -117,11 +118,17 @@ ompl::base::PlannerPtr MultiQueryPlannerAllocator::allocatePlanner(const ob::Spa
...
   {
-    // If we already have an instance, use that one
+    // If we already have an instance, reuse it's planning data
+    // FIXME: make reusing PlannerPtr not crash, so that we don't have to reconstruct a PlannerPtr instance
     auto planner_map_it = planners_.find(new_name);
     if (planner_map_it != planners_.end())
     {
-      return planner_map_it->second;
+      ob::PlannerData data(si);
+      planner_map_it->second->getPlannerData(data);
```
henningkayser, line 127: I'm confused now. How can we access the planner data if the planner instance is invalid? Or are there actually any members inside the planner instance that go out of scope? This version looks like it should produce the same segfaults

henningkayser (approved):
I'm fine with merging this as a temporary solution. The overhead of creating a new planner instance is really not too much of an issue, but we should keep on looking into the underlying cause anyway.

MarqRazz:
Just so everyone knows, I have done some profiling and re-constructing the planner is taking a little under 0.2 seconds which is not ideal for application that are going for fast cycle times!
```
ompl::geometric::LazyPRM::LazyPRM(const base::PlannerData &data, bool starStrategy)
*** Profiling statistics. Total counted time : 0.181746 seconds
Blocks of time:
lazyPRMConstructor: 0.181721s (99.9865%), [0.181721s --> 0.181721 s], 1 parts, 0.181721 s on average
Unaccounted time : 2.4583e-05 (0.013526 %)
```

I have also found that re-using the planning graph does not speed up LazyPRM. If you give it the exact same problem to solve over and over again the graph will increasing in size with each attempt but the planning time always uses the max allowed set by the user. This could be an error with my configuration so I will update this PR with more details if I can get it working better.

moveit_planners/ompl/ompl_interface/src/planning_context_manager.cpp
```diff
@@ -117,11 +118,17 @@ ompl::base::PlannerPtr MultiQueryPlannerAllocator::allocatePlanner(const ob::Spa
   }
   if (multi_query_planning_enabled)
   {
-    // If we already have an instance, use that one
+    // If we already have an instance, reuse it's planning data
+    // FIXME: make reusing PlannerPtr not crash, so that we don't have to reconstruct a PlannerPtr instance
```
sjahr, line 122: Can you open a follow-up issue for this? With a summery of you findings and (if possible) a path forward so somebody else can take this up in the future?

sjahr (approved):
Let's merge this fix and open a follow-up issue :+1:

sjahr (approved):
Let's merge this fix and open a follow-up issue :+1:

MarqRazz:
Can we back port this to Humble? I don't have permissions to ask mergify.

PR #2132, opened 2023-04-21
Refactor the `colors` portion of the PlanningScene class

### LINKED ISSUE #2033: Remove Colors from the PlanningScene class
This is just an idea for discussion, let's not start coding it yet.

As I trim the giant PlanningScene class, it seems like removing Color would be easy to do. I wouldn't expect any of the color-related functions to be called rapidly, thus the locking mechanism doesn't need to be fast.

I suggest making a new `Color` class in the `planning_scene` package. It could simply have a mutex for thread safety.

Here are the functions and objects related to Color in the PS:

```
/** \brief A map from object names (e.g., attached bodies, collision objects) to their colors */
using ObjectColorMap = std::map<std::string, std_msgs::msg::ColorRGBA>;

  /** \brief Construct a vector of messages (\e object_colors) with the colors of the objects from the planning_scene */
  void getObjectColorMsgs(std::vector<moveit_msgs::msg::ObjectColor>& object_colors) const;
  
  bool hasObjectColor(const std::string& id) const;

  const std_msgs::msg::ColorRGBA& getObjectColor(const std::string& id) const;
  void setObjectColor(const std::string& id, const std_msgs::msg::ColorRGBA& color);
  void removeObjectColor(const std::string& id);
  void getKnownObjectColors(ObjectColorMap& kc) const;
  
  std::unique_ptr<ObjectColorMap> object_colors_;
```

We could do the same thing for `using ObjectTypeMap = std::map<std::string, object_recognition_msgs::msg::ObjectType>;`

I'm pretty excited about this idea. For example, RViz would no longer need to lock the Planning Scene to make a color change.

AndyZe:
Fixes #2033

I put object color management in a new class. This is nice b/c, potentially, you wouldn't have to lock the whole planning scene just to change an object color.If everybody's cool with this, it looks like the same technique could be used on some other class members to break it up even more.

- There's no more checking of nullptrs to see if a planning scene has been initialized. Instead, I did this:  `if (object_colors_.getColorMap().empty())`

- This would have been easy if we didn't have to worry about maintaining a "parent" planning scene. IMO, i've never used the parent planning scene functionality so I'd be inclined to delete it.


moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
```diff
@@ -77,11 +77,44 @@ typedef std::function<bool(const moveit::core::RobotState&, bool)> StateFeasibil
     whether the check should be verbose or not. */
 using MotionFeasibilityFn = std::function<bool(const moveit::core::RobotState&, const moveit::core::RobotState&, bool)>;
 
+/** \brief A map from object names (e.g., attached bodies, collision objects) to their types */
+using ObjectTypeMap = std::map<std::string, object_recognition_msgs::msg::ObjectType>;
+
 /** \brief A map from object names (e.g., attached bodies, collision objects) to their colors */
 using ObjectColorMap = std::map<std::string, std_msgs::msg::ColorRGBA>;
```
AndyZe, line 84: I just changed the order of these declarations so `ObjectColorMap` is closer to the color-related stuff. No functional change here.

AndyZe:
Maybe @rhaschke or @v4hn has an opinion

moveit_py/src/moveit/moveit_core/planning_scene/planning_scene.cpp
```diff
@@ -181,15 +181,15 @@ void init_planning_scene(py::module& m)
...
+      //       Set the color of a collision object.
 
-	   Args:
-               object_id (str): The id of the collision object to set the color of.
-               color (std_msgs.msg.ObjectColor): The color of the collision object.
-           )")
+      //    Args:
+      //           object_id (str): The id of the collision object to set the color of.
+      //           color (std_msgs.msg.ObjectColor): The color of the collision object.
+      //       )")
```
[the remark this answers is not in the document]
peterdavidfagan: Hi @AndyZe, 

Thanks for notifying me of this. For now it would make sense to comment out this code and add a todo item. In order to account for these changes I would need to add bindings for the new class you introduced or reference the `object_colors_` from the planning scene instance. I think this is relatively low priority for the Python library right now so I would be happy to just go with a todo item in the code that can be addressed if/when it becomes a priority.

1. Bind the new class PlanningSceneColors
2. Refer to object colors using `object_colors_` from the planning scene

rhaschke (commented):
I don't yet see a code simplification in factoring out the color handling into a separate class.
Your primary goal seems to be relaxing locking requirements, but that wasn't touched yet.

> I've never used the parent planning scene functionality so I'd be inclined to delete it.

The parent functionality (and scene diffs) is at the heart of MoveIt's PlanningScene. It is used all over the place internally in MoveIt and MTC and allows for minimal changes w/o copies of _all_ the data over and over again.

However, for colors, I agree that we don't really need the hierarchical diffs.
I even suggest to move colors completely out of the PS class.

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -632,15 +634,14 @@ void PlanningScene::getPlanningSceneDiffMsg(moveit_msgs::msg::PlanningScene& sce
   collision_detector_->cenv_->getScale(scene_msg.link_scale);
 
   scene_msg.object_colors.clear();
-  if (object_colors_)
+
+  unsigned int i = 0;
+  scene_msg.object_colors.resize(object_colors_.getConstObjectColorMap().size());
+  for (ObjectColorMap::const_iterator it = object_colors_.getConstObjectColorMap().begin();
+       it != object_colors_.getConstObjectColorMap().end(); ++it, ++i)
```
v4hn, line 641: range-based loop?

moveit_py/src/moveit/moveit_core/planning_scene/planning_scene.cpp
```diff
@@ -53,7 +53,7 @@ void apply_collision_object(std::shared_ptr<planning_scene::PlanningScene>& plan
   if (color_msg.has_value())
   {
     // set object color
-    planning_scene->setObjectColor(color_msg.value().id, color_msg.value().color);
+    planning_scene->getPlanningSceneColors()->setObjectColor(color_msg.value().id, color_msg.value().color);
```
v4hn, line 56:
```
planning_scene->colors()->setObjectColor(...)
```

or something similar looks much better I guess?

v4hn (commented):
> There's no more checking of nullptrs to see if a planning scene has been initialized. Instead, I did this: if (object_colors_.getColorMap().empty())

As I said inline, you should allow overriding a parent color map with an empty map (all colors unassigned).

> This would have been easy if we didn't have to worry about maintaining a "parent" planning scene. IMO, i've never used the parent planning scene functionality so I'd be inclined to delete it.

You use it every time you call `PS::diff`. MTC makes extensive use of it, so does a lot of other code that plans in parallel/in many alternative worlds.

some superficial inline comments, but no complete review.
I believe the color support could be superseded by proper visual support on the basis of https://github.com/ros-planning/moveit/pull/2854

tylerjw:
Closing as this seems to be abandoned work.

PR #2139, opened 2023-04-24
[Servo] Fix stop callback, delete pause/unpause mode

sea-bass:
### Description

This PR fixes the stop callback (`stopCB()`) in the MoveIt Servo node implementation by actually ensuring that the `servo_calcs` and `collision_checking` sub-components are stopped (not paused) when the stop service is triggered.


AndyZe (approved):

sea-bass:
Actually, I misunderstood something here. Not a bug.

AndyZe:
Can you leave a comment so this doesn't happen again?

AndyZe:
I find one issue when testing. Call `/servo_node/pause_servo` --> this stops all publication of commands to the controller.

Per [this PR](https://github.com/ros-planning/moveit2/pull/2128) a couple days ago, `paused` mode should continue publishing. (I guess that behavior is debatable, though)

sea-bass:
> I find one issue when testing. Call `/servo_node/pause_servo` --> this stops all publication of commands to the controller.
> 
> Per [this PR](https://github.com/ros-planning/moveit2/pull/2128) a couple days ago, `paused` mode should continue publishing. (I guess that behavior is debatable, though)

That seems like a bug in the other PR because the JTC / vector publishers are inside an `if (paused_)` block:
https://github.com/ros-planning/moveit2/blob/main/moveit_ros/moveit_servo/src/servo_calcs.cpp#L471

AndyZe:
Hmm. Can you think of any reason not to just delete the `pause/unpause` functionality then? It seems identical in almost every way to `stop/start`

sea-bass:
> Hmm. Can you think of any reason not to just delete the `pause/unpause` functionality then? It seems identical in almost every way to `stop/start`

I can't, but I don't have as much experience using Servo compared to other users. If there is no value in keeping the robot publishing (stay at current joint state) commands, then deleting pause/unpause seems good to me.

AndyZe:
There are a couple reasons to use `paused`/`unpaused` instead of `stop`/`start`:

- stop/start involves a bit more overhead. It involves filling some messages and joining/launching a thread

- paused/unpaused keeps the getEndEffectorTransform()/getCommandFrameTransform() functions active

I'd recommend to delete `start`/`stop`, keep `pause`/unpause`. I'm happy to make the code changes if you are busy with other stuff.

sea-bass:
> There are a couple reasons to use `paused`/`unpaused` instead of `stop`/`start`:
> 
> * stop/start involves a bit more overhead. It involves filling some messages and joining/launching a thread
> * paused/unpaused keeps the getEndEffectorTransform()/getCommandFrameTransform() functions active
> 
> I'd recommend to delete `start`/`stop`, keep `pause`/unpause`. I'm happy to make the code changes if you are busy with other stuff.

That's fine by me -- can you make those changes? Up to you if you want to use this PR or a new one, just let me know.

AndyZe:
On further testing, I still see some "joint jump" behavior when Servo resumes from a paused state but not from a stopped state. I'll use this same PR to delete `pause`/`unpause` tomorrow   :+1:   (Please check that I have write access to your fork)

sea-bass:
No, please don't delete low-latency mode! We need it for our Gazebo config, since for other limitations we don't have the Servo node running with sim time.

AndyZe:
OK. Bummer. It makes things so complicated. Will revert it.

henningkayser (approved):
This looks much better to me. @AndyZe are you good with this approach?

sea-bass:
> This looks much better to me. @AndyZe are you good with this approach?

I plan to test this w Studio today, now that I'm back from vacation. If this works well we can merge

AndyZe (approved):

sea-bass:
@AndyZe @henningkayser this also worked for me on Studio, so we can merge it.

Although,I had to fight through some issues in the switch to `generate_parameter_library` since it seems the default namespace for the params changed from `moveit_servo` to no namespace. 

Also, updating the `override_velocity_scaling_factor` parameter at runtime seems to have no effect again. I will dig into this one.

(@ibrahiminfinite FYI)

JafarAbdi:
@AndyZe @sea-bass This PR introduced a regression, if you servo and then call the stop service and move the robot with the joint trajectory controller then start servo, the arm to move fast to the last state before the servo was stopped, reverting this PR fixes it for me

AndyZe:
Thanks @JafarAbdi! Pinging @ibrahiminfinite too (GSOC coder)

ibrahiminfinite:
> Possibly the current state and smoother, which is initialized in the _constructor_ of `ServoCalcs` but maybe we should consider doing it in the `start()` method instead?

The current and previous state variables need to be updated, but they are  already updated in `Servo::start()` [here](https://github.com/ros-planning/moveit2/blob/cc1ef618d9c02728bfc58723f15b29ed1efc6aa2/moveit_ros/moveit_servo/src/servo_calcs.cpp#L218) 
The smoother should be fine if we call the `reset()` method in `Servo::start()`
The issue is possibly due to the smoother still having its state as the robot state when `Servo::stop()` was called while the actual state of the robot has changed now.

PR #2146, opened 2023-04-27
[Servo] Avoid unnecessary checks for initializing `ik_base_to_tip_frame`

ibrahiminfinite:
### Description

The variable `ik_base_to_tip_frame_` was being initialized in `ServoCalcs::start()` and  `ServoCalcs::calculateSingleIteration()`  and both places checked if `use_inv_jacobian` was `False`.
Since this variable is only used in [this block](https://github.com/ros-planning/moveit2/blob/dc7643502adbeaa41dfd370605460cfd45f26b7f/moveit_ros/moveit_servo/src/servo_calcs.cpp#L616) which is already inside a ` if (!use_inv_jacobian_)` check. 
The previous 2 initializations can be removed and replaced with a single initialization at the place it is used.

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
@@ -43,6 +43,7 @@
 #include <mutex>
 #include <thread>
 #include <atomic>
+#include <optional>
 
 // ROS
 #include <control_msgs/msg/joint_jog.hpp>
@@ -337,6 +338,6 @@ protected:
   // Load a smoothing plugin
   pluginlib::ClassLoader<online_signal_smoothing::SmoothingBaseClass> smoothing_loader_;
 
-  kinematics::KinematicsBaseConstPtr ik_solver_ = NULL;
+  kinematics::KinematicsBaseConstPtr ik_solver_ = nullptr;
 };
 }  // namespace moveit_servo
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
@@ -338,7 +338,6 @@ protected:
   pluginlib::ClassLoader<online_signal_smoothing::SmoothingBaseClass> smoothing_loader_;
 
   kinematics::KinematicsBaseConstPtr ik_solver_;
-  Eigen::Isometry3d ik_base_to_tip_frame_;
   bool use_inv_jacobian_ = false;
 };
 }  // namespace moveit_servo
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
@@ -337,7 +337,6 @@ protected:
   // Load a smoothing plugin
   pluginlib::ClassLoader<online_signal_smoothing::SmoothingBaseClass> smoothing_loader_;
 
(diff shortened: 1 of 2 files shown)


ibrahiminfinite:
@AndyZe 
This takes care of the if/else involving `ik_solver_`.
The check happens only once now during construction.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -612,7 +601,8 @@ bool ServoCalcs::cartesianServoCalcs(geometry_msgs::msg::TwistStamped& cmd,
 
     // Poses passed to IK solvers are assumed to be in some tip link (usually EE) reference frame
     // First, find the new tip link position without newly applied rotation
-
+    ik_base_to_tip_frame_ = current_state_->getGlobalLinkTransform(ik_solver_->getBaseFrame()).inverse() *
```
AndyZe, line 604: I would suggest that `ik_base_to_tip_frame_` doesn't need to be a member variable anymore, since it's only used here. There isn't any extra computatoinal expense to allocate an Eigen::Isometry3d here locally.

AndyZe (commented):
Good cleanup PR! Thanks.

AndyZe:
I have an idea for further cleanup. How about:

`std::optional<kinematics::KinematicsBase> ik_solver_` instead of `kinematics::KinematicsBaseConstPtr ik_solver_`

Then we can replace `use_inv_jacobian_` like this:

`if (!ik_solver_)` instead of `if (use_inv_jacobian_)`

Delete `use_inv_jacobian_`

ibrahiminfinite:
> I have an idea for further cleanup. How about:
> 
> `std::optional<kinematics::KinematicsBase> ik_solver_` instead of `kinematics::KinematicsBaseConstPtr ik_solver_`
> 
> Then we can replace `use_inv_jacobian_` like this:
> 
> `if (!ik_solver_)` instead of `if (use_inv_jacobian_)`
> 
> Delete `use_inv_jacobian_`

Used a simple `nullptr` solution.
Is there a reason using `std::optional` will be better ?
Asking since I have not used `std::optional` much.

ibrahiminfinite:
The local variable naming was fixed in #2149.
So I did not make the change here.

AndyZe:
> Used a simple `nullptr` solution. Is there a reason using `std::optional` will be better ? Asking since I have not used `std::optional` much.

Thanks for the effort! I'd make the effort to learn `std::optional`. It's nice to use when you get the hang of it and it makes the "optionality" explicit. Here's a good answer about it:

https://stackoverflow.com/a/74243517/3499467

With that said, it's really a minor detail

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
```diff
@@ -43,6 +43,7 @@
 #include <mutex>
 #include <thread>
 #include <atomic>
+#include <optional>
```
AndyZe, line 46: Sorry to nitpick: please alphabetize

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
```diff
@@ -39,10 +39,11 @@
 #pragma once
 
 // C++
+#include <atomic>
 #include <condition_variable>
 #include <mutex>
+#include <optional>
```
AndyZe, line 45: Looks like this header is not actually used yet.

sjahr (approved):
Nice cleanup!

AndyZe (approved):

PR #2153, opened 2023-05-02
State cost function interface

sjahr:
### Description

This PR enhances the API to use a user-defined state cost function in the planners.

Testing can be done with https://github.com/ros-planning/moveit2_tutorials/pull/679

### Checklist


moveit_core/planning_interface/include/moveit/planning_interface/planning_interface.h
```diff
@@ -117,6 +120,9 @@ class PlanningContext
   /** \brief Set the planning request for this context */
   void setMotionPlanRequest(const MotionPlanRequest& request);
 
+  /** \brief Set the planning request for this context */
+  void setStateCostFunction(const StateCostFn& state_cost_function);
```
henningkayser, line 124: I think that state cost functions should be provided with the planning request, just as constraints

moveit_core/planning_interface/include/moveit/planning_interface/planning_interface.h
```diff
@@ -75,6 +75,9 @@ struct PlannerConfigurationSettings
 /** \brief Map from PlannerConfigurationSettings.name to PlannerConfigurationSettings */
 typedef std::map<std::string, PlannerConfigurationSettings> PlannerConfigurationMap;
 
+using StateCostFn = std::function<double(const moveit::core::RobotState& robot_state, const MotionPlanRequest& request,
```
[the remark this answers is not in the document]
henningkayser: I agree, ideally the cost function signature should only use primitive types so that it's easy to adapt for the different planners (see STOMP cost functions). Also, costs should rather be part of the MotionPlanRequest and not depend on them. STOMP currently has a generic cost function for all trajectory waypoints (https://github.com/ros-planning/moveit2/blob/main/moveit_planners/stomp/include/stomp_moveit/stomp_moveit_task.hpp#LL65C83-L65C88). For a simple state cost, I would therefore propose something like: `CostFn = std::function<double(const Eigen::VectorXd& values, bool& validity)` where validity is even up for debate. The actual implementation and MoveIt-specific state of the cost functions can likewise be managed with lambda bindings. Is there an issue with supporting this type of function with OMPL?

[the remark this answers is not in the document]
henningkayser: Also, consider the [StateValidatorFn](https://github.com/ros-planning/moveit2/blob/main/moveit_planners/stomp/include/stomp_moveit/cost_functions.hpp#L53) as a use case that could be applicable to all planners.

moveit_core/planning_interface/include/moveit/planning_interface/planning_interface.h
```diff
@@ -144,6 +150,9 @@ class PlanningContext
 
   /// The planning request for this context
   MotionPlanRequest request_;
+
+  // Cost function for individual states
+  StateCostFn state_cost_function_ = StateCostFn();
```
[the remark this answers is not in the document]
sjahr: Good point! I'd like to have a way to check for an "unset" cost function, so I think this should be the default, but specific implementations should default to cost = 1

moveit_core/planning_interface/include/moveit/planning_interface/planning_interface.h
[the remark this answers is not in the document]
sjahr: I've updated the signature to `std::function<double(const std::vector<double>& state_vector)` & you're right it simplifies using the cost function a lot. I also like the idea of adding the cost function to the MotionPlanRequest since that will simplify the implementation even more :+1:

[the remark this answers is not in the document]
sjahr: I have to investigate if the concept of a state validator function is applicable to other planners. How does it differ from the processing of constraints within the planner?

henningkayser (commented):
Finally trying to come to a conclusion on this, testing right now...

moveit_core/planning_interface/include/moveit/planning_interface/planning_request.h
```diff
@@ -36,12 +36,22 @@
...
 
-typedef moveit_msgs::msg::MotionPlanRequest MotionPlanRequest;
+/** \brief Definition for a cost function to measure the cost of a single state during motion planning */
+using StateCostFn = std::function<double(const Eigen::VectorXd& state_vector)>;
+
+struct MotionPlanRequest : moveit_msgs::msg::MotionPlanRequest
+{
+  MotionPlanRequest(moveit_msgs::msg::MotionPlanRequest request_msg = moveit_msgs::msg::MotionPlanRequest(),
+                    planning_interface::StateCostFn state_cost_function = nullptr);
+  planning_interface::StateCostFn state_cost_function = nullptr;
```
marioprats, line 52: I guess we need this MotionPlanRequest specialization because the message doesn't have a way to encode cost functions? Does that mean that we can use cost functions only in the C++ interface, but not through messages?
I'm sure I'm missing a lot of context here, but is there a plan to add cost functions to messages, maybe as an enum with pre-defined options?

moveit_core/planning_interface/include/moveit/planning_interface/planning_request.h
```diff
@@ -36,12 +36,22 @@
...
+
+  [[nodiscard]] moveit_msgs::msg::MotionPlanRequest toMessage() const;
```
marioprats, line 54: I'm confused about why this is needed. Since this is inheriting from `moveit_msgs::msg::MotionPlanRequest` and we are not doing anything special with the `state_cost_function` member, this shouldn't be needed, right?

moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/parameterization/joint_space/constrained_planning_state_space_factory.h
```diff
@@ -58,7 +58,7 @@ class ConstrainedPlanningStateSpaceFactory : public ModelBasedStateSpaceFactory
    * For more details on this state space selection process, see:
    * https://github.com/JeroenDM/moveit/pull/2
    * **/
-  int canRepresentProblem(const std::string& group, const moveit_msgs::msg::MotionPlanRequest& req,
+  int canRepresentProblem(const std::string& group, const planning_interface::MotionPlanRequest& req,
```
marioprats, line 61: I like the idea of all these functions now taking a MoveIt! struct, as opposed to a ROS message. Even though the struct is still inheriting from the message, this can be a step towards additional ROS independency.
Not to do now, but what would you think of a `planning_interface::MotionPlanRequest` struct totally independent from the ROS message, with separate to/from functions to convert between one or the other?

moveit_planners/stomp/include/stomp_moveit/cost_functions.hpp
```diff
@@ -219,7 +220,28 @@ CostFn get_constraints_cost_function(const std::shared_ptr<const planning_scene:
...
+ *
+ * @return                    STOMP Cost function based on MoveIt state cost function
+ */
+CostFn get_cost_function_from_moveit_state_cost_fn(const ::planning_interface::StateCostFn& state_cost_function)
+{
+  CostFn cost_fn = [&](const Eigen::MatrixXd& values, Eigen::VectorXd& costs, bool& validity) {
+    validity = true;  // TODO(sjahr): Discuss if that is always the case
+    for (int timestep = 0; timestep < values.cols() - 1; ++timestep)
+    {
+      Eigen::VectorXd current_state_vector = values.col(timestep);
```
marioprats, line 235: const& should avoid a copy here I think, since Eigen block operations return references to the underlying data.

moveit_planners/stomp/include/stomp_moveit/cost_functions.hpp
```diff
@@ -219,7 +220,28 @@ CostFn get_constraints_cost_function(const std::shared_ptr<const planning_scene:
...
+ *
+ * @param state_cost_function The planning scene instance to use for computing transforms
+ *
+ * @return                    STOMP Cost function based on MoveIt state cost function
+ */
+CostFn get_cost_function_from_moveit_state_cost_fn(const ::planning_interface::StateCostFn& state_cost_function)
+{
+  CostFn cost_fn = [&](const Eigen::MatrixXd& values, Eigen::VectorXd& costs, bool& validity) {
+    validity = true;  // TODO(sjahr): Discuss if that is always the case
+    for (int timestep = 0; timestep < values.cols() - 1; ++timestep)
```
marioprats, line 233: Is the cost not required for the last waypoint?

moveit_planners/stomp/src/stomp_moveit_planning_context.cpp
```diff
@@ -151,16 +151,21 @@ stomp::TaskPtr createStompTask(const stomp::StompConfiguration& config, StompPla
   // Cost, noise and filter functions are provided for planning.
   // TODO(henningkayser): parameterize cost penalties
   using namespace stomp_moveit;
-  CostFn cost_fn;
+  std::vector<CostFn> cost_functions;
+  cost_functions.reserve(3);  // Expect a collision, a constraint and a state cost function
```
marioprats, line 155: Since this isn't always the case, maybe remove the reserve?

moveit_planners/stomp/src/stomp_moveit_planning_context.cpp
```diff
@@ -151,16 +151,21 @@ stomp::TaskPtr createStompTask(const stomp::StompConfiguration& config, StompPla
...
-                                                                1.0 /* constraint penalty */) });
+    cost_functions.push_back(costs::get_constraints_cost_function(
+        planning_scene, group, constraints.getAllConstraints(), 1.0 /* constraint penalty */));
   }
-  else
+
+  if (context.getMotionPlanRequest().state_cost_function != nullptr)
   {
-    cost_fn = costs::get_collision_cost_function(planning_scene, group, 1.0 /* collision penalty */);
+    std::cout << "Using user defined state cost function!" << std::endl;
```
marioprats, line 166: remove

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
```diff
@@ -230,6 +238,7 @@ class PlanningComponent
   const std::string group_name_;
   // The robot_model_ member variable of MoveItCpp class will manually free the joint_model_group_ resources
   const moveit::core::JointModelGroup* joint_model_group_;
+  planning_interface::StateCostFn state_cost_function_;
```
marioprats, line 241: = nullptr;

moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
```diff
@@ -203,7 +203,7 @@ bool planning_pipeline::PlanningPipeline::generatePlan(const planning_scene::Pla
   // broadcast the request we are about to work on, if needed
   if (publish_received_requests)
   {
-    received_request_publisher_->publish(req);
+    received_request_publisher_->publish(req.toMessage());
```
marioprats, line 206: Related to my other comment about why `toMessage` is needed, since the request struct inherits from the message.
Maybe you need to reinterpret_cast to the base class?

marioprats (commented):
Looks good!
I have a few minor comments and optional changes, but I'm probably missing context, so feel free to ignore!
I also have a question about how to add support for cost functions on the ROS message, if that's planned.
Other than that, with some additional documentation and testing I think this should be good to merge.

moveit_core/cost_functions/src/cost_functions.cpp
```diff
@@ -0,0 +1,74 @@
...
+namespace moveit
+{
+namespace cost_functions
+{
+[[nodiscard]] ::planning_interface::StateCostFn
+getClearanceCostFn(moveit::core::RobotState& robot_state, const std::string& group_name,
+                   const planning_scene::PlanningSceneConstPtr& planning_scene)
+{
+  // Create cost function
+  return [robot_state, group_name, planning_scene](const Eigen::VectorXd& state_vector) mutable {
```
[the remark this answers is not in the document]
sjahr: There are multiple sources that discourage using std::bind (like [abseil](https://abseil.io/tips/108) and [turner](https://www.youtube.com/watch?v=ZlHi8txU4aQ)) because of the compile and runtime overhead it causes. Maybe C++20's  std::bind_front fixes that but since we're not using it at the moment I'd prefer the lambdas for now. I guess if I'd like to avoid the dependency on the planning_interface I could also just abstain from using the typedef and use `std::function<double(const Eigen::VectorXd& state_vector)>` instead of `StateCostFn`. So the dependency is not mandatory.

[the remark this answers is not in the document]
sjahr: I think it is Ok because I capture references (e.g. `robot_state`) by value:
```
getClearanceCostFn(moveit::core::RobotState& robot_state, const std::string& group_name,
                   const planning_scene::PlanningSceneConstPtr& planning_scene)
{
  return [robot_state, group_name, planning_scene](const Eigen::VectorXd& state_vector)
}
```
Is that assumption correct?

moveit_core/planning_interface/include/moveit/planning_interface/planning_request.h
sjahr: > because the message doesn't have a way to encode cost functions?

Yes :disappointed:, but we had the idea with the pre-defined options too but refrained from adding it for now because we first need to evaluate a set of useful default functions. Until then this is only usable with the cpp interface

moveit_core/planning_interface/include/moveit/planning_interface/planning_request.h
sjahr: I don't remember if it was at runtime or during compilation but in either one of these situations the state_cost_function member function pointer could not be handled and I got either a compiler error or a segfault

moveit_planners/stomp/include/stomp_moveit/cost_functions.hpp
sjahr: Good catch!

moveit_planners/stomp/include/stomp_moveit/cost_functions.hpp
sjahr: Oh, you're right!

moveit_core/cost_functions/src/cost_functions.cpp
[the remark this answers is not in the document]
marioprats: I think it is still making a copy, even if the original parameter is a reference (the pointer is ok though, because even if it makes a copy, it's a copy of the pointer). To verify I ran this minimal example:

```
#include <iostream>
#include <functional>

std::function<void()> create_function(int &p) {
   auto lambda=[p]() mutable {
        p += 50;
   };
   return lambda;
}

int main() {
        int a = 0;
        auto function = create_function(a);
        function();
        std::cout << a << std::endl;
}
```

This returns 0, i.e., the lambda works with a copy of `a`. But worse than that, the copy will be made when `create_function` is called, not when the actual lambda function is called, so it could be working on an old outdated state.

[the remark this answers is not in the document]
marioprats: What I mean with avoiding the dependency is that if you just have the function that computes the cost here, you can turn that into a StateCostFn or any other function signature at the caller side (e.g. stomp or other libraries may have their own cost function signature, but they could still use these cost computation functions and transform them into what they need via bind or lambdas. Of course this doesn't matter if we think that `double(Eigen::VectorXd)` is a general enough cost-definition function, which it probably is. I'm okay with leaving this as is for now and iterate later if needed, but see my comment below for the lambda captures.

On a separate note, I wish we had the absl dependency to be able to use absl::bind_front, StatusOr and others, and to follow their TotW and style guide.

moveit_core/cost_functions/include/moveit/cost_functions/cost_functions.hpp
```diff
@@ -0,0 +1,56 @@
...
+{
+namespace cost_functions
+{
+[[nodiscard]] ::planning_interface::StateCostFn
+createMinJointDisplacementCostFn(moveit::core::RobotState& robot_state, const std::string& group_name,
+                                 const planning_scene::PlanningSceneConstPtr& planning_scene);
+
+[[nodiscard]] ::planning_interface::StateCostFn
+createClearanceCostFn(moveit::core::RobotState& robot_state, const std::string& group_name,
+                      const planning_scene::PlanningSceneConstPtr& planning_scene);
```
marioprats, line 54: We will need comments on what these do, and tests for them.

moveit_core/planning_interface/include/moveit/planning_interface/planning_request.h
marioprats: hmm... maybe worth investigating more? See my other comment about the `reinterpret_cast` to the base class. Or I believe a `static_cast` would work too, since you'd be casting to the base class.

moveit_core/cost_functions/src/cost_functions.cpp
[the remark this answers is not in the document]
sjahr: Good to know! Thanks for the minimal example. I'll update the code

marioprats (commented):
I think this looks good to start experimenting and iterating on what cost functions work and make sense.
Longer term, if we want to add cost functions to the PlanningRequest msg as enums, I would maybe update the PlanningRequest C++ type to reflect that (i.e. an enum instead of a StateCostFn, or go back to use the original msg), and handle the enum inside stomp. That would also avoid the double type-erasure with the lambda functions which I find hard to read (i.e. cost-function -> StateCostFn -> CostFn), since stomp could wrap directly MoveIt cost functions into CostFn. But I would leave all this for the future if needed.

moveit_planners/stomp/include/stomp_moveit/cost_functions.hpp
```diff
@@ -222,7 +223,28 @@ CostFn get_constraints_cost_function(const std::shared_ptr<const planning_scene:
...
+ * Creates a STOMP cost function from
+ *
+ * @param state_cost_function The planning scene instance to use for computing transforms
+ *
+ * @return                    STOMP Cost function based on MoveIt state cost function
+ */
+CostFn get_cost_function_from_moveit_state_cost_fn(const ::planning_interface::StateCostFn& state_cost_function)
+{
+  CostFn cost_fn = [&](const Eigen::MatrixXd& values, Eigen::VectorXd& costs, bool& validity) {
+    validity = true;  // TODO(sjahr): Discuss if that is always the case
```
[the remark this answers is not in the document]
henningkayser: The threshold needs to be provided by the cost function implementation. Unless we normalize all costs (which is not easy to do) there is no way to define a general threshold. The linked PR only changed the cost metric to use actual distance values, it doesn't really deal with validity.

moveit_core/cost_functions/src/cost_functions.cpp
```diff
@@ -0,0 +1,74 @@
...
+    robot_state.setJointGroupActivePositions(group_name, state_vector);
+    auto const shortest_distance_to_collision = planning_scene->distanceToCollision(robot_state);
+
+    // Return cost based on shortest_distance if the robot is not in contact or penetrated a collision object
+    if (shortest_distance_to_collision > 0.0)
+    {
+      // The closer the collision object the higher the cost
+      return 1.0 / shortest_distance_to_collision;
+    }
+    return std::numeric_limits<double>::infinity();  // Return a max cost cost by default
```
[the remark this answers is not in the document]
henningkayser: I still think that infinity will basically break optimization since there is no way to approximate a gradient based in this value. IMO, a better cost function would take a specific clearance target as 0 (or max_distance if set), and then returning a linear cost for distances smaller than that, including negative distances. It's very useful to also consider negative distances since they also produce a gradient and STOMP can optimize out collisions. The cost function could looks like this ```max(0, target_clearance - distance)```

henningkayser:
@sjahr I think this looks good for merging. Can you please add the comments that @marioprats requested, and track my suggestions about the cost functions in a follow-up issue?

PR #2154, opened 2023-05-03
fix: fix Jacobian Quarternion representation bug

shaohedu:
### Description

Fix: robot_state.cpp getJacobian() function quaternion representation
[reference: ](https://hal.science/hal-01122406v4/file/kinematics.pdf) 1.7.2 Global perturbations
It should be global perturbation rather than local perturbation.
I verify the resulte with the following method of estimation:
$Y = F(q)         \qquad  \Delta Y = J(q_0)  \Delta q$
$Y_0 = F(q_0) $
$Y_1 = F(q_1)  \qquad  \Delta Y_1 = J(q_0)  \Delta q_1$
$Y_2 = F(q_2)  \qquad  \Delta Y_2 = J(q_0)  \Delta q_2$
$...$
$Y_n = F(q_n)  \qquad  \Delta Y_n = J(q_0)  \Delta q_n$

$[ \Delta Y_1,  \Delta Y_2,  ... , \Delta Y_n] = J(q_0)[ \Delta q_1,  \Delta q_2, ... , \Delta q_n]$
$[ \Delta Y_1,  \Delta Y_2,  ... , \Delta Y_n] = A$
$[ \Delta q_1,  \Delta q_2, ... , \Delta q_n] = B$
$J(q_0) = AB^T(BB^T)^-1$

### Checklist

### DIFF
--- a/moveit_core/robot_state/src/robot_state.cpp
+++ b/moveit_core/robot_state/src/robot_state.cpp
@@ -1436,15 +1436,17 @@ bool RobotState::getJacobian(const JointModelGroup* group, const LinkModel* link
   }
   if (use_quaternion_representation)
   {  // Quaternion representation
-    // From "Advanced Dynamics and Motion Simulation" by Paul Mitiguy
+    // From "Quaternion kinematics for the error-state KF" 1.7.2 Global perturbations
+    // https://hal.science/hal-01122406v4/file/kinematics.pdf
+    // dq/qt = 0.5 * wg ⊗ q
     // d/dt ( [w] ) = 1/2 * [ -x -y -z ]  * [ omega_1 ]
-    //        [x]           [  w -z  y ]    [ omega_2 ]
-    //        [y]           [  z  w -x ]    [ omega_3 ]
-    //        [z]           [ -y  x  w ]
+    //        [x]           [  w  z -y ]    [ omega_2 ]
+    //        [y]           [ -z  w  x ]    [ omega_3 ]
+    //        [z]           [  y -x  w ]
     Eigen::Quaterniond q(link_transform.linear());
     double w = q.w(), x = q.x(), y = q.y(), z = q.z();
     Eigen::MatrixXd quaternion_update_matrix(4, 3);
-    quaternion_update_matrix << -x, -y, -z, w, -z, y, z, w, -x, -y, x, w;
+    quaternion_update_matrix << -x, -y, -z, w, z, -y, -z, w, x, y, -x, w;
     jacobian.block(3, 0, 4, columns) = 0.5 * quaternion_update_matrix * jacobian.block(3, 0, 3, columns);
   }
   return true;


henningkayser:
@shaohedu thank you for your fix. Can you elaborate on why you think that this is a bug and maybe demonstrate how it affects the results? Why do you think that the quaternion representation should used instead of local? For cases like this it would be great to validate the fix with an adequate unit test.

shaohedu:
@henningkayser 

## The experimental results：
I estimate the Jacobian matrix using the least squares algorithm, and the result is consistent with the global perturbation. Here is the result:

_tmp_dingtalkgov_qt_pic_1683900321210
## Theoretical analysis：
The time-derivatives of quaternion is obtained from Euler angle transformation rates. The independent variable is the Euler angles. Therefore, I think the quaternion representation should use global pertubation instead of local.
## Least squares estimate function:
$Y = F(q)         \qquad  \Delta Y = J(q_0)  \Delta q$
$Y_0 = F(q_0) $
$Y_1 = F(q_1)  \qquad  \Delta Y_1 = J(q_0)  \Delta q_1$
$Y_2 = F(q_2)  \qquad  \Delta Y_2 = J(q_0)  \Delta q_2$
$...$
$Y_n = F(q_n)  \qquad  \Delta Y_n = J(q_0)  \Delta q_n$

$[ \Delta Y_1,  \Delta Y_2,  ... , \Delta Y_n] = J(q_0)[ \Delta q_1,  \Delta q_2, ... , \Delta q_n]$
$[ \Delta Y_1,  \Delta Y_2,  ... , \Delta Y_n] = A$
$[ \Delta q_1,  \Delta q_2, ... , \Delta q_n] = B$
$J(q_0) = AB^T(BB^T)^-1$
q is joints state，Y is SE3 pose(eg. end-effector pose). add a pertubation $\Delta q$, Y changes $\Delta Y$, then using Least-squares can estimate the Jacobian matrix
## Estimation function code:
```C++
Eigen::MatrixXd jacobian_estimate(const State& q)
{
    Eigen::VectorXd y0 = endeffectorPose(q);
    int raw = y0.size();
    int colums = q.size();
    Eigen::MatrixXd dY = Eigen::MatrixXd::Zero(y0.size(), colums); // dY
    Eigen::MatrixXd dQ = Eigen::MatrixXd::Zero(q.size(), colums); // dQ
    const double dq = 0.001;
    std::vector<State> X;
    std::vector<Eigen::VectorXd> Y;

    for (int i = 0; i < columns; i++) {
        State qn = q;
        qn[i] += dq;
        Eigen::VectorXd yn = endeffectorPose(qn);
        dY.block(0, i, raw, 1) = yn - y0;
        Eigen::VectorXd dqn = Eigen::VectorXd::Zero(q.size());
        dqn(i) = dq;
        dQ.block(0, i, q.size(), 1) = dqn;
    }
    return dY * dQ.transpose() * (dQ * dQ.transpose()).inverse();
}
```

shaohedu:
@henningkayser Do I need to add more explanation or information?

henningkayser:
> @henningkayser Do I need to add more explanation or information?

Your explanation is great, I just didn't have the time to look into this properly. It would also be great if @AndyZe could take a brief look at this.

shaohedu:
> > @henningkayser Do I need to add more explanation or information?
> 
> Your explanation is great, I just didn't have the time to look into this properly. It would also be great if @AndyZe could take a brief look at this. My biggest concern is that there is code somewhere that depends on the old convention, and that we could break without notice.

Any update?

AndyZe:
I'm swamped and won't have time to look soon. So don't wait for me. Sorry guys

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -1436,15 +1436,17 @@ bool RobotState::getJacobian(const JointModelGroup* group, const LinkModel* link
...
-    //        [z]           [ -y  x  w ]
+    //        [x]           [  w  z -y ]    [ omega_2 ]
+    //        [y]           [ -z  w  x ]    [ omega_3 ]
+    //        [z]           [  y -x  w ]
     Eigen::Quaterniond q(link_transform.linear());
     double w = q.w(), x = q.x(), y = q.y(), z = q.z();
     Eigen::MatrixXd quaternion_update_matrix(4, 3);
-    quaternion_update_matrix << -x, -y, -z, w, -z, y, z, w, -x, -y, x, w;
+    quaternion_update_matrix << -x, -y, -z, w, z, -y, -z, w, x, y, -x, w;
     jacobian.block(3, 0, 4, columns) = 0.5 * quaternion_update_matrix * jacobian.block(3, 0, 3, columns);
```
henningkayser, line 1450: in case of global pertubation, the ordering in this function and the comment needs to be fixed per your reference: `0.5 * omega * quaternion_update_matrix`

henningkayser (commented):
(sorry for so long response time lately)
Ok, I see where you are coming from. However, this looks rather like a matter of convention and use case instead of a real bug. Both local and global pertubations can be useful. And from the original implementation it doesn't look like it's clearly depending on global Euler angles, does it? Instead, the documentation even states that the Jacobian is defined in [reference of the provided link](https://github.com/ros-planning/moveit2/blob/main/moveit_core/robot_state/include/moveit/robot_state/robot_state.h#L1142), not a global frame. I'm inclined to suggest adding an alternative API for your use case, instead.

sea-bass:
@ibrahiminfinite You've been looking at this code closely -- what do you think?

ibrahiminfinite:
The usage I was interested in was singularity analysis at a specific point, and in that context local pertubation was more relevant.
I also took a look at the other places in MoveIt where the `getJacobian()` is used,  and local pertubation looked more relevant there for things like manipulability analysis.

I also don't think there is any real bug here.
If the user needs to use the global pertubation specifically for some reason, it will be best to go with a seperate API like @henningkayser said.

shaohedu:
do not need fix

PR #2157, opened 2023-05-08
MoveIt Python Library - Add Gripper Interface

peterdavidfagan:
### Description

This pull request looks to begin addressing a request for a Gripper Interface made by @AndrejOrsula in the following [thread](https://github.com/ros-planning/moveit2/issues/1279#issuecomment-1143660241). Ideally the API should look as follows with the gripper client added to the library: 

```python
# Instantiate gripper client
from moveit.gripper_action_client import GripperClient
robotiq_85_gripper = GripperClient("/robotiq_gripper/gripper_cmd")

def open_gripper():
    robotiq_85_gripper.send_gripper_command(0.0, 0.0)
    
def close_gripper():
    robotiq_85_gripper.send_gripper_command(1.0, 1.0)

# close gripper applying max_effort of 1.0 N.
close_gripper()
```

### Features
- [x] Support `GripperCommand`

### Checklist

### DIFF
--- a/moveit_py/moveit/gripper_action_client/__init__.py
+++ b/moveit_py/moveit/gripper_action_client/__init__.py
@@ -1,0 +1,1 @@
+from .client import GripperClient
--- a/moveit_py/moveit/gripper_action_client/client.py
+++ b/moveit_py/moveit/gripper_action_client/client.py
@@ -1,0 +1,22 @@
+import subprocess
+
+
+class GripperClient:
+    def __init__(self, action_name):
+        self.action_name = action_name
+
+    def send_gripper_command(self, position, max_effort):
+        subprocess.run(
+            [
+                "ros2",
+                "action",
+                "send_goal",
+                self.action_name,
+                "control_msgs/action/GripperCommand",
+                "{command: { position: "
+                + str(position)
+                + ", max_effort: "
+                + str(max_effort)
+                + "}}",
+            ]
+        )


peterdavidfagan:
@AndrejOrsula your feedback and thoughts on features you would like as part of the Gripper interface is much appreciated. Right now it is a basic action client, this works well for most basic applications IMO.

henningkayser (commented):
I see that this can be useful for very simple test setups. However, the same action should be supported by MoveIt's gripper controller handle which is supposed to also handle the runtime state of the action call. Would it be too much to ask to bind to the controller interface? And if yes, how can it be improved to streamline usage through the Python bindings?
I bet @MarqRazz also has a few words to say about supporting custom grippers.

AndrejOrsula:
Hello @peterdavidfagan,

Thank you very much for looking into it!

Yes, I think having a simple-to-use interface for grippers would be quite beneficial. Besides sending commands, it would be great if the same interface also enabled users to query the current state of the gripper. Here are some primary categories that could be discussed (some of these might be beyond the scope of a gripper interface):
- Support grippers with [`GripperCommand`](https://github.com/ros-controls/control_msgs/blob/fe4ce167bf6db6e6e0209e997759ead8764e529d/control_msgs/action/GripperCommand.action) and [`FollowJointTrajectory`](https://github.com/ros-controls/control_msgs/blob/fe4ce167bf6db6e6e0209e997759ead8764e529d/control_msgs/action/FollowJointTrajectory.action) (`JointTrajectoryController`) interfaces. From what I have seen, many simulated grippers and multi-finger grippers (hands) implement only the `FollowJointTrajectory` interface or something similar (but not `GripperCommand`).
- When sending goals via action requests, support getting feedback and result in a synchronous or asynchronous way.
- Simplistic `open()` / `close()` commands with optional `position` and `max_effort` arguments.
	- If a gripper with `FollowJointTrajectory` interface supports controlling each finger separately, it might be nice to have an interface for that too (e.g. `set_target_configuration()`).
- Simplistic `is_open` / `is_closed` getters (properties), with the option of also directly getting the `JointState` of all joints.
	- If a gripper has some additional sensors (e.g. force/tactile), it might be nice to have a way of passing the output of these through the interface.
- Option to automatically parse joint names, link names (and other necessities) directly from SRDF/URDF based on the `MoveGroup` of the gripper.

Looking at the current action client in this PR, I am not so sure about the use of a new subprocess for each gripper action. It is probably good enough for most use cases, but I feel like the overhead might be too much for some applications (it could become a bottleneck in some robot learning scenarios). Handling of feedback/results could also become cumbersome. Just my personal opinion, as I have not tested it. What do you think?

I tried two separate interfaces in my previous attempt, but both have some limitations: [MoveIt2Gripper](https://github.com/AndrejOrsula/pymoveit2/blob/f55428ab7a31887436aa5921979abbefe9e648b8/pymoveit2/moveit2_gripper.py) (`FollowJointTrajectory` interface | [example](https://github.com/AndrejOrsula/pymoveit2/blob/f55428ab7a31887436aa5921979abbefe9e648b8/examples/ex_gripper.py)) and [GripperCommand](https://github.com/AndrejOrsula/pymoveit2/blob/f55428ab7a31887436aa5921979abbefe9e648b8/pymoveit2/gripper_command.py) (`GripperCommand` interface | [example](https://github.com/AndrejOrsula/pymoveit2/blob/f55428ab7a31887436aa5921979abbefe9e648b8/examples/ex_gripper_command.py)). I used the first one for a simulated gripper inside Gazebo, and the second one to control a real gripper. Both of these use `rclpy` as the interface for all communication. Do you think there are better alternatives to using `rclpy` based on your experience with `moveit_py` (e.g. binding directly to the MoveIt 2 codebase, which could eventually provide some of the functionalities for free)?

peterdavidfagan:
Thanks @AndrejOrsula for the detailed response. 

> Besides sending commands, it would be great if the same interface also enabled users to query the current state of the gripper. 

Makes sense.

> Support grippers with [GripperCommand](https://github.com/ros-controls/control_msgs/blob/fe4ce167bf6db6e6e0209e997759ead8764e529d/control_msgs/action/GripperCommand.action) and [FollowJointTrajectory](https://github.com/ros-controls/control_msgs/blob/fe4ce167bf6db6e6e0209e997759ead8764e529d/control_msgs/action/FollowJointTrajectory.action) (JointTrajectoryController) interfaces. From what I have seen, many simulated grippers and multi-finger grippers (hands) implement only the FollowJointTrajectory interface or something similar (but not GripperCommand).

For this point, my current thought process was that the `FollowJointTrajectory` interface is exposed through planning groups with the necessary ros2 controls configuration. For instance with the existing library once could define a planning group for the entire manipulator or the hand alone. Do you see this as too cumbersome and am I missing something? 

> When sending goals via action requests, support getting feedback and result in a synchronous or asynchronous way.

Thanks for this I agree, this should be supported.

> Simplistic open() / close() commands with optional position and max_effort arguments.
If a gripper with FollowJointTrajectory interface supports controlling each finger separately, it might be nice to have an interface for that too (e.g. set_target_configuration()).
Simplistic is_open / is_closed getters (properties), with the option of also directly getting the JointState of all joints.
If a gripper has some additional sensors (e.g. force/tactile), it might be nice to have a way of passing the output of these through the interface.

I planned to make the open/close gripper methods, members of the GripperClient class for convenience. It would be nice to have the other methods you listed as well. 

> I tried two separate interfaces in my previous attempt

Thanks for sharing these. 

> Do you think there are better alternatives to using rclpy based on your experience with moveit_py (e.g. binding directly to the MoveIt 2 codebase, which could eventually provide some of the functionalities for free)?

In this case of the new Python library, I believe there are performance benefits to leveraging the C++ library directly versus relying on message passing via ROS client libraries. In the case of interacting with a Gripper, I think the ROS client libraries will be necessary unless bypassing the use of ROS 2 control.

AndrejOrsula:
> > Support grippers with [`GripperCommand`](https://github.com/ros-controls/control_msgs/blob/fe4ce167bf6db6e6e0209e997759ead8764e529d/control_msgs/action/GripperCommand.action) and [`FollowJointTrajectory`](https://github.com/ros-controls/control_msgs/blob/fe4ce167bf6db6e6e0209e997759ead8764e529d/control_msgs/action/FollowJointTrajectory.action) (`JointTrajectoryController`) interfaces. From what I have seen, many simulated grippers and multi-finger grippers (hands) implement only the `FollowJointTrajectory` interface or something similar (but not `GripperCommand`).
>
> For this point, my current thought process was that the `FollowJointTrajectory` interface is exposed through planning groups with the necessary ros2 controls configuration. For instance with the existing library one could define a planning group for the entire manipulator or the hand alone. Do you see this as too cumbersome and am I missing something?

That makes sense @peterdavidfagan, and it is certainly usable that way. Both `open()` and `close()` commands could be just motions to named targets. But I wonder if it might be beneficial to have such a gripper-specific wrapper around `FollowJointTrajectory` (while being compatible with the `GripperCommand` interface). :thinking:

sea-bass:
I'm skeptical about the decision to use `subprocess` to call out to the ROS 2 CLI when you could just write a Python node that creates an action client and abstracts it away, e.g. [this](https://github.com/DLu/simple_actions). Is there a specific reason you decided to go this route?

peterdavidfagan:
>  Is there a specific reason you decided to go this route?

Hi @sea-bass, 

This was an initial quick but unclean solution for interfacing with available action for the `GripperCommand`, it was working fine for my use case so I didn't go much further. I started by implementing an action client in Python but was encountering issues when running it from within a notebook.

It's far from perfect, I wanted to open the discussion around this and then consider adding it to my task list to develop the gripper interface for the Python library.

peterdavidfagan:
> If you want to make a GripperActionClient class that helps speed up integration/standardize how users interact with the GripperActionController then I think ros2_controllers/gripper_controllers would be a better home for this code.

Thanks @MarqRazz, and this makes sense. 

If I do write such a class in ros2_controllers/gripper_controllers, for the Python bindings do you think it makes sense to start writing a separate Python binding library for ROS 2 control? I think it may be useful for replicating some of the functionalities of libraries like https://facebookresearch.github.io/fairo/polymetis/overview.html as it is closer to the control stack.

peterdavidfagan:
Thanks @MarqRazz, I'll look to follow up with this approach once I have finished the work I am doing on grasp pose estimation for my lab.

henningkayser:
We already have a gripper action with this [controller plugin](https://github.com/ros-planning/moveit2/blob/main/moveit_plugins/moveit_ros_control_interface/src/gripper_controller_plugin.cpp) coming [from here](https://github.com/ros-planning/moveit2/blob/main/moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/gripper_controller_handle.h#L50). If it's really about the same action interface, we should not duplicate functionality, but rather think about making it easier to access it. (I know that MoveIt's controller interfaces are not ideal, but this would be a good opportunity to think about how to improve them)

PR #2171, opened 2023-05-11
[Servo] Restore namespace to parameters

### LINKED ISSUE #2162: Restore namespace in MoveIt Servo parameters
It seems that https://github.com/ros-planning/moveit2/pull/2096 removed the namespace in the servo parameters.

These should be added back in, as the removal of the `moveit_servo` namespace changes the level at which user parameters are specified (for example, we had to do that in MoveIt Studio).

This can be changed by adding the namespace in the constructor of the parameter listener.

ibrahiminfinite:
### Description

Fixes #2162 
This PR adds the `moveit_servo` namespace to MoveIt Servo parameters.

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/launch/servo_example.launch.py
+++ b/moveit_ros/moveit_servo/launch/servo_example.launch.py
@@ -19,6 +19,7 @@ def generate_launch_description():
         .yaml("config/panda_simulated_config.yaml")
         .to_dict()
     )
+    servo_params = {"moveit_servo": servo_params}
     # This filter parameter should be >1. Increase it for greater smoothing but slower motion.
     low_pass_filter_coeff = {"butterworth_filter_coeff": 1.5}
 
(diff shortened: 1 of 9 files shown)


sjahr (approved):

moveit_ros/moveit_servo/launch/servo_example.launch.py
```diff
@@ -19,6 +19,7 @@ def generate_launch_description():
         .yaml("config/panda_simulated_config.yaml")
         .to_dict()
     )
+    servo_params = {"moveit_servo": servo_params}
```
sea-bass, line 22: Can you do this in one line without overwriting the `servo_params` variable like you did in the pose tracking launch file?

sea-bass (approved):
I did mean to approve the PR, since my comment is minor

moveit_ros/moveit_servo/launch/servo_example.launch.py
ibrahiminfinite: ah yes, will do.

sea-bass (approved):

PR #2180, opened 2023-05-16
[Servo] Move `enforcePositionLimits` and `enforceVelocityLimits` to utilities

ibrahiminfinite:
### Description

This PR moves the functions `enforcePositionLimits` and `enforceVelocityLimits`  into utilities from `servo_calcs.cpp` and `enforce_limits.cpp` respectively.
The dependency of some of the utility functions on `rclcpp:Clock` has also been removed.

### Checklist


henningkayser (commented):
Is there a reason to not use [RobotState::enforceBounds()](https://moveit.picknik.ai/main/api/html/classmoveit_1_1core_1_1RobotState.html#aca89bd2b257eef35f03e424f1d5c2ff0) or the alternatives for position/velocities?

ibrahiminfinite:
> Is there a reason to not use [RobotState::enforceBounds()](https://moveit.picknik.ai/main/api/html/classmoveit_1_1core_1_1RobotState.html#aca89bd2b257eef35f03e424f1d5c2ff0) or the alternatives for position/velocities?

Probably because of this (from PR #451 )
> Previously, the incoming robot state velocity was used to check if the robot would move beyond the joint limit. That was a problem if the robot state did not contain velocity data, i.e. if "copy dynamics" hadn't been turned on.
This PR uses the prospective robot velocity from Servo itself to check if the joint would move beyond the position limit and halt, if so. This velocity should always be available regardless of whether "copy dynamics" is ON.


Maybe @AndyZe  can clarify more on this ?

Also @sea-bass , did you have any cases where you needed to use the `override_velocity_scaling_factor` parameter ?
If that is not useful, we can take that out as well.

sea-bass:
We are currently using `override_velocity_scaling,_factor` actively right now, so please keep it!

moveit_ros/moveit_servo/include/moveit_servo/utilities.h
```diff
@@ -105,4 +105,28 @@ void transformTwistToPlanningFrame(geometry_msgs::msg::TwistStamped& cmd, const
...
 geometry_msgs::msg::Pose poseFromCartesianDelta(const Eigen::VectorXd& delta_x,
                                                 const Eigen::Isometry3d& base_to_tip_frame_transform);
+
+/**
+ * @brief Decrease robot position change and velocity, if needed, to satisfy joint velocity limits
+ * @param joint_model_group Active joint group. Used to retrieve limits
+ * @param publish_period Period of the servo loop
+ * @param joint_state The command that will go to the robot
+ * @param override_velocity_scaling_factor Option if the user wants a constant override of the velocity scaling
+ *        a value greater than 0 will override the internal calculations done by getVelocityScalingFactor
```
AndyZe, line 115:
```suggestion
 *        A value greater than 0 will override the internal calculations done by getVelocityScalingFactor
```

moveit_ros/moveit_servo/include/moveit_servo/utilities.h
```diff
@@ -105,4 +105,28 @@ void transformTwistToPlanningFrame(geometry_msgs::msg::TwistStamped& cmd, const
...
+
+/** \brief Avoid overshooting joint limits
+ * @param joint_state The joint state to be checked
+ * @param joint_limit_margin The allowed margin for joint limit
+ * @param joint_model_group The MoveIt group
+ * @param clock The ROS clock of the calling node
+   \return Vector of the joints that would move farther past position margin limits
+*/
+std::vector<const moveit::core::JointModel*>
+enforcePositionLimits(sensor_msgs::msg::JointState& joint_state, double joint_limit_margin,
```
AndyZe, line 129:
```suggestion
enforcePositionLimits(sensor_msgs::msg::JointState& joint_state, const double joint_limit_margin,
```

moveit_ros/moveit_servo/include/moveit_servo/utilities.h
```diff
@@ -105,4 +105,28 @@ void transformTwistToPlanningFrame(geometry_msgs::msg::TwistStamped& cmd, const
...
+ * @param override_velocity_scaling_factor Option if the user wants a constant override of the velocity scaling
+ *        a value greater than 0 will override the internal calculations done by getVelocityScalingFactor
+ */
+void enforceVelocityLimits(const moveit::core::JointModelGroup* joint_model_group, const double publish_period,
+                           sensor_msgs::msg::JointState& joint_state,
+                           const double override_velocity_scaling_factor = 0.0);
+
+/** \brief Avoid overshooting joint limits
+ * @param joint_state The joint state to be checked
+ * @param joint_limit_margin The allowed margin for joint limit
```
AndyZe, line 123: Suggesting this b/c the user may be confused whether the margin is over the limit or prior to the limit.

```suggestion
 * @param joint_limit_margin The allowed margin for joint limit. This is a buffer, prior to the actual limit.
```

ibrahiminfinite:
@AndyZe 
Can you clarify on this https://github.com/ros-planning/moveit2/pull/2180#pullrequestreview-1429273829

AndyZe:
I agree with your comment there

AndyZe:
Overall I'm not sure this PR accomplishes much. It moves 2 free functions from `enforce_limits.cpp` to `utilities.cpp`. No functional change, right? I'd suggest to close the PR if that is correct. Sorry @ibrahiminfinite

AndyZe:
Well, I see that `getVelocityScalingFactor()` was not a free function previously but it is with this PR. That's a positive change.

moveit_ros/moveit_servo/include/moveit_servo/utilities.h
```diff
@@ -105,4 +105,28 @@ void transformTwistToPlanningFrame(geometry_msgs::msg::TwistStamped& cmd, const
...
+ * @param joint_model_group The MoveIt group
+ * @param clock The ROS clock of the calling node
+   \return Vector of the joints that would move farther past position margin limits
+*/
+std::vector<const moveit::core::JointModel*>
+enforcePositionLimits(sensor_msgs::msg::JointState& joint_state, double joint_limit_margin,
+                      const moveit::core::JointModelGroup* joint_model_group, rclcpp::Clock& clock);
```
AndyZe, line 130:
```suggestion
                      const moveit::core::JointModelGroup* joint_model_group, const rclcpp::Clock& clock);
```

moveit_ros/moveit_servo/src/utilities.cpp
```diff
@@ -249,4 +249,101 @@ geometry_msgs::msg::Pose poseFromCartesianDelta(const Eigen::VectorXd& delta_x,
...
+    velocity *= velocity_scaling_factor;
+    // Back to sensor_msgs type
+    joint_state.velocity = std::vector<double>(velocity.data(), velocity.data() + velocity.size());
+    joint_state.position = std::vector<double>(positions.data(), positions.data() + positions.size());
+  }
+}
+
+std::vector<const moveit::core::JointModel*>
+enforcePositionLimits(sensor_msgs::msg::JointState& joint_state, double joint_limit_margin,
+                      const moveit::core::JointModelGroup* joint_model_group, rclcpp::Clock& clock)
```
AndyZe, line 300:
```suggestion
enforcePositionLimits(sensor_msgs::msg::JointState& joint_state, const double joint_limit_margin,
                      const moveit::core::JointModelGroup* joint_model_group, const rclcpp::Clock& clock)
```

ibrahiminfinite:
This PR is a cleanup PR, mostly to keep things cleaner in the current `ServoCalcs` and also pull out functions that can be reused into utilities so that it could be used in the upcoming refactor.

moveit_ros/moveit_servo/src/utilities.cpp
```diff
@@ -249,4 +247,101 @@ geometry_msgs::msg::Pose poseFromCartesianDelta(const Eigen::VectorXd& delta_x,
...
+  }
+  if (!joints_to_halt.empty())
+  {
+    std::ostringstream joints_names;
+    std::transform(joints_to_halt.cbegin(), std::prev(joints_to_halt.cend()),
+                   std::ostream_iterator<std::string>(joints_names, ", "),
+                   [](const auto& joint) { return joint->getName(); });
+    joints_names << joints_to_halt.back()->getName();
+    RCLCPP_WARN_STREAM_THROTTLE(LOGGER, clock, ROS_LOG_THROTTLE_PERIOD,
+                                joints_names.str() << " close to a position limit. Halting.");
```
sea-bass, line 342: Given that the `clock` variable is used only to print a log message, I think this should be removed from the input arguments.

If you want to return `joints_to_halt` in some way, or at least a Boolean that indicates whether joints were halting, you could consider that as another argument of the function instead.

sea-bass (changes requested):
Just one comment here

moveit_ros/moveit_servo/src/utilities.cpp
sea-bass: Actually, `joints_to_halt` is the thing being returned! So this is easy. Just remove the clock argument and move the log message to `servo_calcs.cpp` itself.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -618,7 +616,8 @@ bool ServoCalcs::internalServoUpdate(Eigen::ArrayXd& delta_theta,
                         servo_params_.override_velocity_scaling_factor);
 
   // Enforce SRDF position limits, might halt if needed, set prev_vel to 0
-  const auto joints_to_halt = enforcePositionLimits(next_joint_state_);
+  const auto joints_to_halt = enforcePositionLimits(next_joint_state_, servo_params_.joint_limit_margin,
+                                                    joint_model_group_, *node_->get_clock());
   if (!joints_to_halt.empty())
   {
     status_ = StatusCode::JOINT_BOUND;
```
sea-bass, line 623: basically, your log message would get pasted in here

moveit_ros/moveit_servo/src/utilities.cpp
ibrahiminfinite: Yup, passing the clock did not look very sensible, but I lazed out on that 
Might as well get that fixed for other functions as also in this PR itself 👍

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -618,8 +618,17 @@ bool ServoCalcs::internalServoUpdate(Eigen::ArrayXd& delta_theta,
   // Enforce SRDF position limits, might halt if needed, set prev_vel to 0
   const auto joints_to_halt = enforcePositionLimits(next_joint_state_, servo_params_.joint_limit_margin,
                                                     joint_model_group_, *node_->get_clock());
+
   if (!joints_to_halt.empty())
   {
+    std::ostringstream joints_names;
```
sea-bass, line 639: Call it `joint_names`

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -618,8 +618,17 @@ bool ServoCalcs::internalServoUpdate(Eigen::ArrayXd& delta_theta,
...
+    std::transform(joints_to_halt.cbegin(), std::prev(joints_to_halt.cend()),
+                   std::ostream_iterator<std::string>(joints_names, ", "),
+                   [](const auto& joint) { return joint->getName(); });
+    joints_names << joints_to_halt.back()->getName();
+    RCLCPP_WARN_STREAM_THROTTLE(LOGGER, *node_->get_clock(), ROS_LOG_THROTTLE_PERIOD,
+                                joints_names.str() << " close to a position limit. Halting.");
```
sea-bass, line 645: right now this will print something like

```
joint1 joint2 joint3 close to a position limit. Halting.
```

Consider formatting it a bit more clearly , for example

```
Joints 'joint1' 'joint2' 'joint3' close to a position limit. Halting.
```

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -503,9 +503,15 @@ bool ServoCalcs::cartesianServoCalcs(geometry_msgs::msg::TwistStamped& cmd,
     return false;
 
   // Transform the command to the MoveGroup planning frame
+  if (cmd.header.frame_id.empty())
+  {
+    RCLCPP_WARN_STREAM_THROTTLE(LOGGER, *node_->get_clock(), ROS_LOG_THROTTLE_PERIOD,
+                                "No frame specified for command ,will use planning_frame");
```
sea-bass, line 509:
```suggestion
                                "No frame specified for command, will use planning_frame");
```

also... can you print in there what the planning frame is? It's just in the next line below.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -559,11 +559,17 @@ bool ServoCalcs::cartesianServoCalcs(geometry_msgs::msg::TwistStamped& cmd,
...
   }
 
+  auto last_status = status_;
   delta_theta_ *= velocityScalingFactorForSingularity(joint_model_group_, delta_x, svd, pseudo_inverse,
                                                       servo_params_.hard_stop_singularity_threshold,
                                                       servo_params_.lower_singularity_threshold,
                                                       servo_params_.leaving_singularity_threshold_multiplier,
-                                                      *node_->get_clock(), current_state_, status_);
+                                                      current_state_, status_);
+  // Status will have changed if approcaching singularity
```
sea-bass, line 568:
```suggestion
  // Status will have changed if approaching singularity
```

Also it is worth actually printing something in front of it like "Servo status changed to BLAH"

sea-bass (commented):
Thanks for this! Found a few typos and suggestions.

moveit_ros/moveit_servo/src/servo_calcs.cpp
ibrahiminfinite: Not sure about that since it already prints a verbose message for the statuses when near singularity, would that not be enough ?

```c++
{ StatusCode::DECELERATE_FOR_LEAVING_SINGULARITY, "Moving away from a singularity, decelerating" },
{ StatusCode::DECELERATE_FOR_APPROACHING_SINGULARITY, "Moving closer to a singularity, decelerating" },
{ StatusCode::HALT_FOR_SINGULARITY, "Very close to a singularity, emergency stop" },
```

sea-bass: Ah yeah I think that's fine then. Thanks!

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -555,11 +560,17 @@ bool ServoCalcs::cartesianServoCalcs(geometry_msgs::msg::TwistStamped& cmd,
     delta_theta_ = pseudo_inverse * delta_x;
   }
 
+  auto last_status = status_;
```
sea-bass, line 563: const

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -555,11 +560,17 @@ bool ServoCalcs::cartesianServoCalcs(geometry_msgs::msg::TwistStamped& cmd,
...
   delta_theta_ *= velocityScalingFactorForSingularity(joint_model_group_, delta_x, svd, pseudo_inverse,
                                                       servo_params_.hard_stop_singularity_threshold,
                                                       servo_params_.lower_singularity_threshold,
                                                       servo_params_.leaving_singularity_threshold_multiplier,
-                                                      *node_->get_clock(), current_state_, status_);
+                                                      current_state_, status_);
+  // Status will have changed if approaching  singularity
```
sea-bass, line 569: Remove double space

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -618,9 +632,20 @@ bool ServoCalcs::internalServoUpdate(Eigen::ArrayXd& delta_theta,
...
   if (!joints_to_halt.empty())
   {
+    std::ostringstream joint_names;
+    std::transform(joints_to_halt.cbegin(), std::prev(joints_to_halt.cend()),
+                   std::ostream_iterator<std::string>(joint_names, " "),
+                   [](const auto& joint) { return " '" + joint->getName() + "' "; });
+
+    joint_names << joints_to_halt.back()->getName();
+    RCLCPP_WARN_STREAM_THROTTLE(LOGGER, *node_->get_clock(), ROS_LOG_THROTTLE_PERIOD,
+                                joint_names.str() << " close to a position limit. Halting.");
```
sea-bass, line 647: Shouldnt this also start with the string "Joints" before showing joint names?

Also this will put 2 spaces after the last joint since the std::ostream stuff will already end with a space

moveit_ros/moveit_servo/src/servo_calcs.cpp
ibrahiminfinite: Removed the `auto` as well and replaced with the type

moveit_ros/moveit_servo/src/servo_calcs.cpp
ibrahiminfinite: Ah you are right.
Fixed it 
Should print this now
```
Joints 'joint1' 'joint2' 'joint3' close to a position limit. Halting.
```

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -618,9 +632,20 @@ bool ServoCalcs::internalServoUpdate(Eigen::ArrayXd& delta_theta,
...
+      enforcePositionLimits(next_joint_state_, servo_params_.joint_limit_margin, joint_model_group_);
+
   if (!joints_to_halt.empty())
   {
+    std::ostringstream joint_names;
+    std::transform(joints_to_halt.cbegin(), std::prev(joints_to_halt.cend()),
+                   std::ostream_iterator<std::string>(joint_names, ""),
+                   [](const auto& joint) { return " '" + joint->getName() + "'"; });
+
+    joint_names << joints_to_halt.back()->getName();
```
sea-bass, line 645: I missed this, but why is this line needed? Wouldn't the line above address the full set of joints if you remove the `std::prev()` ?

Here is a godbolt link showing your implementation and then the fix: https://godbolt.org/z/5hd68YGTa

Would strongly recommend testing your code before making changes. In the long run, it's easier than going back and forth with reviews.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -505,9 +503,16 @@ bool ServoCalcs::cartesianServoCalcs(geometry_msgs::msg::TwistStamped& cmd,
     return false;
 
   // Transform the command to the MoveGroup planning frame
+  if (cmd.header.frame_id.empty())
+  {
+    RCLCPP_WARN_STREAM_THROTTLE(LOGGER, *node_->get_clock(), ROS_LOG_THROTTLE_PERIOD,
+                                "No frame specified for command, will use planning_frame : "
```
sea-bass, line 509:
```suggestion
                                "No frame specified for command, will use planning_frame: "
```

sea-bass (changes requested):
Almost there! I found a bug in the log message printing.

moveit_ros/moveit_servo/src/servo_calcs.cpp
ibrahiminfinite: oh ! Really sorry about that.
Did not give it much thought since the code was from the original implementation and assumed it would work.
I wonder why it was done like this (could be bug that went unnoticed till now)

sea-bass: I think the previous code worked when there were no single quotes around, and spaces after the joint name.

But because there was a `std::transform()` and another line for the last element without the trailing space, the quotes/spaces had to be updated consistently.

Anyways, after fixing that we should be good.

ibrahiminfinite: Some observations while testing

If I first jog the 7th joint to reach the limit, the message correctly prints 
```
Joints 'panda_joint7' close to a position limit. Halting.
```
After this if I jog the 6th joint to the limit, the message only tells about the 6th joint while the 7th is also actually at the limit

```
Joints 'panda_joint6' close to a position limit. Halting.
```

But if I move the robot with the 6th and 7th joints at limit, using the Twist command, it gives the message as expected
```
Joints 'panda_joint6' 'panda_joint7' close to a position limit. Halting.
```

Is this the expected behaviour ? shouldn't the JointJog also print the message similar to how it works for Twsit ?
The reason I think is this check, where it skips the other joints when  the joint velocity is set as zero during jointJog for all joints except the one being moved

sea-bass: That seems fine. Makes sense we don't care to check joints we're not moving.

sea-bass (approved):
Awesome!

ibrahiminfinite:
@sjahr  @sea-bass 
I think this can go in now, all tests passing

Issue #2178, opened 2023-05-15
Allow dynamically toggling collision checking in MoveIt Servo without stopping and restarting.


sea-bass:
I have bandwidth to work on this feature next week, but I am simply capturing this in an issue to open up discussions.

Now that we use `generate_parameter_library` and can easily change parameter values at runtime, this lets us be able to react to the `check_collisions` parameter more easily.

However, right now, the collision checking timer thread is only accessible in the start/stop service callbacks of MoveIt Servo. Meaning you can't dynamically enable/disable collision checking capabilities. A more performant runtime check could also be doable, so I'll be working on that.

ibrahiminfinite:
I was looking into this same thing.
It might also be good to not rely on the pub-sub mechanism and use a shared variable to update the `collision_velocity_scale` value instead.

mechwiz:
linking this issue as well https://github.com/ros-planning/moveit2/issues/409

sea-bass:
> I was looking into this same thing. It might also be good to not rely on the pub-sub mechanism and use a shared variable to update the `collision_velocity_scale` value instead.

I think this is a good idea performance-wise -- how about you make a separate issue for this if you'd like to tackle it as part of your refactoring efforts? Not super high priority, though.

AndyZe:
The original idea was that people would launch the servo node as a node component so pub/sub would be more efficient. There's a comment about that here:

https://github.com/ros-planning/moveit2/blob/243b0b2ccac156a75bcd158eefa86c8acf409110/moveit_ros/moveit_servo/launch/servo_example.launch.py#L80

I can't remember why L80 is commented, though. We should look into making a node component the default way. (Yet another issue...)

Agree, low priority.

PR #2183, opened 2023-05-17
Change servo collision checking parameters to dynamically update

sea-bass:
In this PR, we change the MoveIt servo parameter listener to be a shared pointer so we can listen to servo parameters in the main servo class, as well as the servo calculations and collision checking classes.

This, in turn, allows for collision checking parameters to dynamically be updated!

To test, you can directly change parameters in the servo node as follows, for example:

```
ros2 param set /servo_server moveit_servo.self_collision_proximity_threshold 0.1
ros2 param set /servo_server moveit_servo.check_collisions False
```

Closes https://github.com/ros-planning/moveit2/issues/2178


moveit_ros/moveit_servo/src/collision_check.cpp
```diff
@@ -49,14 +49,16 @@ constexpr size_t ROS_LOG_THROTTLE_PERIOD = 30 * 1000;  // Milliseconds to thrott
...
+                               std::shared_ptr<const servo::ParamListener> servo_param_listener)
   : node_(node)
-  , servo_params_(servo_params)
+  , servo_param_listener_(std::move(servo_param_listener))
+  , servo_params_(servo_param_listener_->get_params())
   , planning_scene_monitor_(planning_scene_monitor)
   , self_velocity_scale_coefficient_(-log(0.001) / servo_params_.self_collision_proximity_threshold)
   , scene_velocity_scale_coefficient_(-log(0.001) / servo_params_.scene_collision_proximity_threshold)
-  , period_(1. / servo_params_.collision_check_rate)
+  , period_(1.0 / servo_params_.collision_check_rate)
```
tylerjw, line 61: This change seems extra, does it do something I don't understand?

tylerjw (commented):
I went and looked at parameter updating and found this variable in the config that controls if it ever checks for updated parameters in the main loop: `enable_parameter_update`. Why was this done? Shouldn't we allow users to set any parameter? If we want to have this behavior, should the collision checking code also pay attention to this too?

moveit_ros/moveit_servo/src/collision_check.cpp
sea-bass: No, I just disliked that there was no number after the decimal point

sea-bass:
> I went and looked at parameter updating and found this variable in the config that controls if it ever checks for updated parameters in the main loop: `enable_parameter_update`. Why was this done? Shouldn't we allow users to set any parameter? If we want to have this behavior, should the collision checking code also pay attention to this too?

The collision checking code should pay attention to it -- tagged you in a separate comment.

The parameter itself was done to copy how the admittance controller was set up: if this param is false, the servo loop will never update its internal parameters as a way of "locking" the system configuration. But if true, it will allow runtime changes.

In the big PR that switched servo to generate_parameter_library, the dynamic updates only happened in ServoCalcs, but not in CollisionChecking

moveit_ros/moveit_servo/src/collision_check.cpp
tylerjw: I don't care either way, just seemed odd to me. (I regularly just follow a number with a . to say it is a floating point but I could see how that syntax would look odd).

AndyZe:
I'd be happy to get rid of `enable_parameter_update`. It's not like somebody accidentally updates a ROS parameter. They usually know what they're doing  ;)

moveit_ros/moveit_servo/src/collision_check.cpp
```diff
@@ -100,58 +102,68 @@ void CollisionCheck::stop()
...
-  collision_result_.print();
-
-  velocity_scale_ = 1;
-  // If we're definitely in collision, stop immediately
-  if (collision_detected_)
+  // Update the latest parameters
+  if (servo_params_.enable_parameter_update)
   {
-    velocity_scale_ = 0;
+    servo_params_ = servo_param_listener_->get_params();
```
AndyZe, line 108: How long does this take? If it's more than a few milliseconds, maybe not a good idea.

Can you put a timer around it to benchmark?

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
```diff
@@ -79,7 +79,7 @@ class ServoCalcs
 public:
   ServoCalcs(const rclcpp::Node::SharedPtr& node,
              const planning_scene_monitor::PlanningSceneMonitorPtr& planning_scene_monitor,
-             std::unique_ptr<const servo::ParamListener> servo_param_listener);
+             std::shared_ptr<const servo::ParamListener> servo_param_listener);
```
AndyZe, line 82:
```suggestion
             const std::shared_ptr<const servo::ParamListener> servo_param_listener);
```

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
```diff
@@ -200,7 +200,7 @@ class ServoCalcs
   std::shared_ptr<rclcpp::Node> node_;
 
   // Servo parameters
-  std::unique_ptr<const servo::ParamListener> servo_param_listener_;
+  std::shared_ptr<const servo::ParamListener> servo_param_listener_;
```
AndyZe, line 203:
```suggestion
  const std::shared_ptr<const servo::ParamListener> servo_param_listener_;
```

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -50,12 +50,12 @@ constexpr double ROBOT_STATE_WAIT_TIME = 10.0;  // seconds
 
 Servo::Servo(const rclcpp::Node::SharedPtr& node,
              const planning_scene_monitor::PlanningSceneMonitorPtr& planning_scene_monitor,
-             std::unique_ptr<const servo::ParamListener> servo_param_listener)
-  : servo_params_{ servo_param_listener->get_params() }
+             std::shared_ptr<const servo::ParamListener> servo_param_listener)
```
AndyZe, line 53:
```suggestion
             const std::shared_ptr<const servo::ParamListener> servo_param_listener)
```

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -74,7 +74,7 @@ int const THREAD_PRIORITY = 40;
 // Constructor for the class that handles servoing calculations
 ServoCalcs::ServoCalcs(const rclcpp::Node::SharedPtr& node,
                        const planning_scene_monitor::PlanningSceneMonitorPtr& planning_scene_monitor,
-                       std::unique_ptr<const servo::ParamListener> servo_param_listener)
+                       std::shared_ptr<const servo::ParamListener> servo_param_listener)
```
AndyZe, line 77:
```suggestion
                       const std::shared_ptr<const servo::ParamListener> servo_param_listener)
```

moveit_ros/moveit_servo/src/collision_check.cpp
sea-bass: It takes between 2 and 10 microseconds on my machine. So I think we're good.

AndyZe (approved):

PR #2186, opened 2023-05-18
Reset filters when starting MoveIt Servo calculations

sea-bass:
This PR fixes the regression @JafarAbdi pointed out in https://github.com/ros-planning/moveit2/pull/2139.

Turns out we just have to reset the filters on `ServoCalcs::start()`.

This fixed it for me on MoveIt Studio, but would love for someone else to test.


JafarAbdi:
@sea-bass Thanks! I'll test it on hardware as soon as I get a chance

AndyZe (approved):

bgill92 (approved):
Tested it in Moveit Studio with gazebo, saw the joints sag while jogging, but stopped sagging when the jogging ended (which I believe is the intended behavior)

PR #2189, opened 2023-05-19
Run SonarCloud only on push

tylerjw:
### Description

This enables sending the coverage report to sonar cloud.

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -5,4 +5,5 @@ sonar.python.version=3
 sonar.sources=.work/target_ws/src
 sonar.cfamily.llvm-cov.reportPath=.work/target_ws/coverage.info
 sonar.scm.exclusions.disabled=true
+sonar.exclusions=.work/target_ws/src/moveit2/moveit_planners/test_configs/prbt_ikfast_manipulator_plugin/src/prbt_manipulator_ikfast_solver.cpp
 sonar.cfamily.compile-commands=.work/target_ws/build/compile_commands.json
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -4,5 +4,5 @@ sonar.organization=moveit
 sonar.python.version=3
 sonar.sources=.work/target_ws/src
 sonar.cfamily.llvm-cov.reportPath=.work/target_ws/coverage.info
-
+sonar.scm.exclusions.disabled=true
 sonar.cfamily.compile-commands=.work/target_ws/build/compile_commands.json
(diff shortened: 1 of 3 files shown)


sonar-project.properties
```diff
@@ -3,5 +3,6 @@ sonar.organization=moveit
 
 sonar.python.version=3
 sonar.sources=.work/target_ws/src
+sonar.cfamily.llvm-cov.reportPath=.work/target_ws/coverage.info
```
EzraBrooks, line 6: It doesn't seem to have picked up any coverage info. Maybe this has to be a directory?

```suggestion
sonar.cfamily.llvm-cov.reportPath=.work/target_ws
```

sonar-project.properties
```diff
@@ -3,5 +3,6 @@ sonar.organization=moveit
...
```
EzraBrooks, line 7: just for now can we see if disabling SCM exclusion importing fixes the report? then we can debug why it thinks everything is gitignored after we have a successful scan.

```suggestion
sonar.scm.exclusions.disabled=true
```

sonar-project.properties
EzraBrooks: actually nope (hiding my above incorrect comment) I just looked at the [example for this](https://github.com/sonarsource-cfamily-examples/linux-cmake-llvm-cov-gh-actions-sc/blob/main/.github/workflows/build.yml#L48) and it is a file path, not a directory path.

EzraBrooks:
We should add the ikfast directory to the sonar exclusions list

sonar-project.properties
```diff
@@ -3,5 +3,6 @@ sonar.organization=moveit
 
 sonar.python.version=3
 sonar.sources=.work/target_ws/src
-
+sonar.cfamily.llvm-cov.reportPath=.work/target_ws/coverage.info
+sonar.scm.exclusions.disabled=true
```
EzraBrooks, line 7: This generated file seems to be particularly noisy.

```suggestion
sonar.scm.exclusions.disabled=true
sonar.exclusions=.work/target_ws/src/moveit2/moveit_planners/test_configs/prbt_ikfast_manipulator_plugin/src/prbt_manipulator_ikfast_solver.cpp
```

sonar-project.properties
```diff
@@ -3,5 +3,7 @@ sonar.organization=moveit
 
 sonar.python.version=3
 sonar.sources=.work/target_ws/src
-
+sonar.cfamily.llvm-cov.reportPath=.work/target_ws/coverage.info
```
EzraBrooks, line 6:
```suggestion
sonar.cfamily.gcov.reportPath=.work/target_ws/coverage.info
```

EzraBrooks (approved):
approving to unblock PRs from forks

PR #2194, opened 2023-05-21
Fix the launching of Servo as a node component

AndyZe:
### Description

Now that this seems to work reliably, make it the default way. It should decrease latency.

I tested with the tutorial, like normal:

https://moveit.picknik.ai/main/doc/examples/realtime_servo/realtime_servo_tutorial.html


AndyZe:
@ibrahiminfinite

moveit_ros/moveit_servo/launch/servo_example.launch.py
```diff
@@ -79,16 +79,16 @@ def generate_launch_description():
         composable_node_descriptions=[
             # Example of launching Servo as a node component
             # Assuming ROS2 intraprocess communications works well, this is a more efficient way.
```
sea-bass, line 81: I would say if this now works, let's rephrase this comment.

```suggestion
            # Launching as a node component makes ROS 2 intraprocess communication more efficient.
```

sea-bass (approved):

sea-bass (approved):

henningkayser (approved):
Looks good. When are we ready to use component nodes by default? ;)

sea-bass:
> Looks good. When are we ready to use component nodes by default? ;)

If it works, I'd be totally fine with the default param being true. @AndyZe up to you if you wanna tweak that before merging.

AndyZe:
I believe it is the default with this PR:

```
    launch_as_standalone_node = LaunchConfiguration(
        "launch_as_standalone_node", default="false"
    )
```

sea-bass:
Woops, I got my logicals mixed up. Yes, you're right :)

sea-bass:
Unfortunately this change actually has a bug, my apologies for not actually testing it. Turns out a component node doesn't have the ability to accept a launch condition parameter... fun

From a user on Discord,

I was following this tutorial (https://moveit.picknik.ai/main/doc/examples/realtime_servo/realtime_servo_tutorial.html#launching-a-servo-node) and after running the command:
`ros2 launch moveit_servo servo_example.launch.py`

I got an error:
```
[INFO] [launch]: Default logging verbosity is set to INFO
[ERROR] [launch]: Caught exception in launch (see debug for traceback): Caught exception when trying to load file of format [py]: ComposableNode.__init__() got an unexpected keyword argument 'condition'
```
Commenting out the condition argument in launch file at `~/ws_moveit/src/moveit2/moveit_ros/moveit_servo/launch/servo_example.launch.py` solved my problem :
```
    container = launch_ros.actions.ComposableNodeContainer(
        name="moveit_servo_demo_container",
        namespace="/",
        package="rclcpp_components",
        executable="component_container_mt",
        composable_node_descriptions=[
            # Example of launching Servo as a node component
            # Launching as a node component makes ROS 2 intraprocess communication more efficient.
            launch_ros.descriptions.ComposableNode(
                package="moveit_servo",
                plugin="moveit_servo::ServoNode",
                name="servo_node",
                parameters=[
                    servo_params,
                ],
                # condition=UnlessCondition(launch_as_standalone_node), <------- Commented out(Line 98)
            ),
```

sea-bass:
Update: This actually works on Rolling, but not Humble, so I think we shouldn't do much.

See https://github.com/ros2/launch_ros/issues/294 for more details.

AndyZe:
Ok. I thought I had tested it before pushing

PR #2200, opened 2023-05-23
Ignore more common tooling-generated folders

EzraBrooks:
### Description

Ignore colcon `install/` and `log/` and clangd `.cache`.
- it's pretty reasonable to use the `moveit2` repo root as your workspace root for `colcon` if you're making quick changes.
  - accordingly, also ignore the two folders that would be generated from a vcs import.
- Myself and many others use the `clangd` tooling in VS Code or other IDEs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,12 @@
 *~
 ~$
 .swp$
+
+# colcon
 build/
+install/
+log/
+
 bin/
 lib/
 msg_gen/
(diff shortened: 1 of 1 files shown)


tylerjw (approved):
Checking out other git repos inside a repo that doesn't use submodules is weird IMO but I'll allow it, I don't see any harm.

EzraBrooks:
> Checking out other git repos inside a repo that doesn't use submodules is weird IMO but I'll allow it, I don't see any harm.

Well ideally they'd be in a subdirectory instead of in ROS's weird inverted-workspace layout, but /shrug

tylerjw:
> Well ideally they'd be in a subdirectory instead of in ROS's weird inverted-workspace layout, but /shrug

```
mkdir log
vcs import log < moveit2.repos
```

:rocket: :rocket: :rocket:

PR #2201, opened 2023-05-23
Use `isnan` in `DepthImageOctomapUpdater`

EzraBrooks:
### Description

Uses `std::isnan`for a particularly nasty conditional we saw in the Sonar scan that haunted me, instead of relying on the fact that `NaN != NaN`.


moveit_ros/perception/depth_image_octomap_updater/src/depth_image_octomap_updater.cpp
```diff
@@ -400,7 +401,7 @@ void DepthImageOctomapUpdater::depthImageCallback(const sensor_msgs::msg::Image:
     inv_fy_ = 1.0 / K4_;
 
     // if there are any NaNs, discard data
-    if (!(px == px && py == py && inv_fx_ == inv_fx_ && inv_fy_ == inv_fy_))
+    if (isnan(px) || isnan(py) || isnan(inv_fx_) || isnan(inv_fy_))
```
ChrisThrasher, line 404: `isnan()` from the C standard library is actually [an implementation-defined macro](https://en.cppreference.com/w/c/numeric/math/isnan) where as `std::isnan()` from the C++ standard library is [a proper function](https://en.cppreference.com/w/cpp/numeric/math/isnan) and overloaded on all floating point types. I really doubt the difference matters but I wanted to point that.

ChrisThrasher (approved):
Common SonarCloud W

moveit_ros/perception/depth_image_octomap_updater/src/depth_image_octomap_updater.cpp
EzraBrooks: huh, I was confused by exactly that cppreference page

PR #2203, opened 2023-05-25
Revert central differencing calculation in servo

sea-bass:
This is effectively a revert of https://github.com/ros-planning/moveit2/pull/2080, just that flat out reverting the commit had a lot of merge conflicts due to newer PRs that landed.

We found this was affecting us in MoveIt Studio in that the velocity scaling was not kicking in correctly. For example, going from `override_velocity_scaling_factor` value of `0.1` to `1.0` did not actually increase the target velocity 10x; it was being damped out by the central difference.

After a bunch of debugging and thinking, I've convinced myself that central differencing is not what we need here -- it's useful for state estimation, but not when we're trying to calculate a command given a known current state and a desired next state. We know exactly what our target velocity should be.

To test:
* Start some kind of Servo node with the ability to command a robot
* Try change speed scaling: `ros2 param set /servo_node moveit_servo.override_velocity_scale_factor 0.1`
* Check that the velocity actually scales linearly when not scaled down by other factors (collision, singularity, etc.)
* Compare w/ `main`

### DIFF
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
@@ -216,9 +216,9 @@ protected:
   moveit::core::RobotStatePtr current_state_;
 
   // These variables are mutex protected
-  // previous_joint_state holds the state q(t - dt)
-  // current_joint_state holds the state q(t) as retrieved from the planning scene monitor.
-  sensor_msgs::msg::JointState previous_joint_state_, current_joint_state_;
+  // previous_joint_state_ holds the state q(t - dt)
+  // internal_joint_state_ holds the state q(t) as retrieved from the planning scene monitor.
+  sensor_msgs::msg::JointState previous_joint_state_, internal_joint_state_;
   std::map<std::string, std::size_t> joint_state_name_map_;
 
   // Smoothing algorithm (loads a plugin)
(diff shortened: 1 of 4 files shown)


sea-bass:
@ibrahiminfinite  FYI

ibrahiminfinite:
Do we still need to remove the central difference ?

sea-bass:
> Do we still need to remove the central difference ?

Yes, I think so. I filled in more details in the issue description.

bgill92 (approved):
Tested with Moveit Studio and Gazebo, reverting the central difference does allow for proper velocity scaling.

AndyZe:
I don't have a convenient way to test this at the moment, but I'm concerned if there will be a joint jump if Servo is stopped, the robot moves, then Servo starts again. So could you try:

- launch Servo
- ros2 service call /servo_node/stop_servo std_srvs/srv/Trigger {}\
- move the robot somehow (maybe `rqt_joint_trajectory_controller`)
- ros2 service call /servo_node/start_servo std_srvs/srv/Trigger {}\

And look carefully for a joint jump

sea-bass:
> I don't have a convenient way to test this at the moment, but I'm concerned about a joint jump if Servo is stopped, the robot moves, then Servo starts again. 

We fixed this with https://github.com/ros-planning/moveit2/pull/2186. It wasn't related to the velocity differencing calculation, but rather to the low-pass filters retaining old state from before the servo calcs were stopped.

AndyZe:
Please test regardless.

AndyZe (approved):

sjahr (approved):

PR #2207, opened 2023-05-29
Export moveit_py_utils' cmake target(s)

rhaschke:
To use moveit_py_utils in downstream projects (e.g. MTC), its targets need to be exported.

### DIFF
--- a/moveit_py/CMakeLists.txt
+++ b/moveit_py/CMakeLists.txt
@@ -90,4 +90,6 @@ if(BUILD_TESTING)
   endforeach()
 endif()
 
+ament_export_targets(moveit_py_utilsTargets HAS_LIBRARY_TARGET)
+ament_export_dependencies(moveit_ros_planning_interface)
 ament_package()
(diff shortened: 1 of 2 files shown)


JafarAbdi:
Should we move it to a separate package outside this repo? I needed the same utilities for https://github.com/PickNikRobotics/rviz_visual_tools/pull/239 and my solution was to copy past them

rhaschke:
> Should we move it to a separate package outside this repo?

Of course, we could provide a MoveIt-independent utility package for python bindings. I thought about this too already.
However, for simplicity, I always decided to stick with a moveit package. In any case, that's a different decision.
For now, we should make moveit_py reusable by downstream packages.

> I needed the same utilities for [PickNikRobotics/rviz_visual_tools#239](https://github.com/PickNikRobotics/rviz_visual_tools/pull/239) and my solution was to copy past them.

I'm always trying to avoid duplicating code!

JafarAbdi (approved):

PR #2208, opened 2023-05-29
Cleanup to ros_control_interface and trajectory execution

stephanie-eng:
### Description

Fix some comments, variable naming, documentation, etc. in moveit_ros_control_interface and trajectory execution


moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -243,7 +243,6 @@ class Ros2ControlManager : public moveit_controller_manager::MoveItControllerMan
   Ros2ControlManager()
     : loader_("moveit_ros_control_interface", "moveit_ros_control_interface::ControllerHandleAllocator")
   {
-    RCLCPP_INFO_STREAM(LOGGER, "Started moveit_ros_control_interface::Ros2ControlManager for namespace " << ns_);
```
stephanie-eng, line 246: I moved this to the other constructor that takes ns as an argument, because it looks like if using this constructor, ns_ would always be an empty string. That makes this print statement not very useful.

JafarAbdi (approved):

vatanaksoytezer (approved):

sjahr (approved):

PR #2211, opened 2023-05-30
Fix servo speed scaling YAML parameters

sea-bass:
After https://github.com/ros-planning/moveit2/pull/2096, the servo speed scaling parameters were renamed from, e.g., `scale.joint` to `joint_scale` but were not updated in the YAML files. So we were using the defaults.

This PR fixes the parameter files to the values actually take effect.


sea-bass:
@ibrahiminfinite fyi

ibrahiminfinite:
What do you think of  preserving the old format using nested params ? , generate_parameter_library supports that.
So that users who are already using the old style config don't have to change their yaml

sea-bass:
> What do you think of preserving the old format using nested params ? , generate_parameter_library supports that. So that users who are already using the old style config don't have to change their yaml
> 
> I may have originally planned to change it to nested but then forgot later on.

Wow, yeah, that's definitely better. Let me do that real quick.

PR #2213, opened 2023-05-31
Warn if optimization objective does not match expected values

stephanie-eng:
Currently, if you specify an optimization objective that does not match any of the expected values, the path length optimization objective will be used without warning the user. This can lead to unexpected behaviour if you have mistyped the optimization objective name. This PR adds a warning for this case.


sjahr (commented):
Thanks for making that change! What do you think about printing a warning in both cases?

sjahr (approved):

henningkayser (approved):

Issue #2219, opened 2023-06-01
MoveItPy only supports a single set of parameters


henningkayser:
### Description

MoveItPy parses the `--param-file` arg at constructions [here](https://github.com/ros-planning/moveit2/blob/main/moveit_py/moveit/utils.py#L56), ignoring that there can be multiple.
This is the case when multiple dictionaries are listed as launch parameters, as [for example here with RViz](https://github.com/ros-planning/moveit_resources/blob/ros2/panda_moveit_config/launch/demo.launch.py#L69).

I ran into this when trying to pass `use_sim_time:=true` to moveit_py which always failed, even though MoveGroup/MoveItCpp didn't have any issues with that. The additional parameters where simply ignored. Side note, the temporary parameter files are being added to `/tmp/`, the actual file names are listed in the ROS log.

### Your environment
* ROS Distro: Humble
* OS Version: Ubuntu 22.04
* [jammy-humble](https://github.com/moveit/moveit2_packages/tree/jammy-humble) binaries (main branch)

### Steps to reproduce
Add {use_sim_time:=True} to the [moveit_py parameters](https://github.com/ros-planning/moveit2_tutorials/blob/main/doc/examples/motion_planning_python_api/launch/motion_planning_python_api_tutorial.launch.py#L38), then query `ros2 param get moveit_py use_sim_time` (it's False).

### Expected behaviour
All passed parameters should be forwarded to MoveItCpp.

### Actual behaviour
All parameters that are not in the first dict are ignored by moveit_py.

PR #2223, opened 2023-06-02
Fix moveit_py rclcpp::init()

### LINKED ISSUE #2220: MoveItPy does not support use_sim_time:=True
### Description

When working around #2219 by merging `use_sim_time:=True` into the MoveIt config dictionary, I'm seeing mixed results:

* `use_sim_time` is actually set to `True` :tada: 
* The moveit_py node is crashing due to an `rclcpp::exceptions::InvalidParameterValueException`

In particular it prints:
```
[motion_planning_python_api_tutorial.py-2] terminate called after throwing an instance of 'rclcpp::exceptions::InvalidParameterValueException'
[motion_planning_python_api_tutorial.py-2]   what():  parameter 'qos_overrides./clock.subscription.durability' could not be set:
```

A quick search brought me to [this issue comment](https://github.com/ros2/rclcpp/issues/1566#issuecomment-788779572), pointing out that `use_sim_time:=True` will enable qos_override callbacks on the `/clock` topic. Question is, why does everything work with the C++ versions MoveGroup/MoveItCpp, and why do we run into this exception with moveit_py?
I have briefly skimmed the node initialization and didn't find any obvious issues.

For context, I've been working on a MoveIt+Gazebo setup and wanted to add a MoveItPy example.
[Here](https://github.com/henningkayser/moveit2_tutorials/blob/panda_gz/doc/examples/motion_planning_python_api/motion_planning_python_api_tutorial.rst#running-the-tutorial-in-a-gazebo-simulation) are instructions for running different launch files in a docker compose container. You'll find that all service targets besides **python_gazebo** work, **python_gazebo** fails for the described reason.

### Your environment
* ROS Distro: Humble
* OS Version: Ubuntu 22.04
* [jammy-humble](https://github.com/moveit/moveit2_packages/tree/jammy-humble) binaries (main branch)

### Steps to reproduce
Try passing `use_sim_time:=True` [like this](https://github.com/henningkayser/moveit2_tutorials/blob/panda_gz/doc/examples/motion_planning_python_api/launch/motion_planning_python_api_tutorial.launch.py#L67).

### Expected behaviour
The launch file should work and allow using sim time with moveit_py.

### Actual behaviour
The Exception is thrown, node dies.

henningkayser:
Fixes #2220. Rclcpp has been initialized without args which was problematic for some use cases like clock simulation. Parameters like use_sim_time:=true need to be passed to rclcpp, also NodeOptions access the global rcl state on construction.

@peterdavidfagan maybe we should just pass all the launch arguments from moveit_py over to rclcpp through the MoveItPy constructor. That would also take care of #2219


henningkayser:
of course clang-tidy is not happy...

so, this does indeed fix simulation, but it somehow changes something with the topic internals.
The Python tutorial works completely, but moveit_py has started to publish planning scene updates to `/monitored_planning_scene` even though the topic parameter is set to `/moveit_cpp/monitored_planning_scene` (I could even verify this with `ros2 param get`). No idea why topic and parameter diverge here. All other communication channels (ros2_control, path publisher to RViz, etc) just work. Changing the monitor topic to `/monitored_planning_scene` in RViz also works without issues. I'm simply confused why this PR would break a single topic that way. What am I missing?

moveit_py/src/moveit/moveit_ros/moveit_cpp/moveit_cpp.cpp
```diff
@@ -58,40 +58,44 @@ void init_moveit_py(py::module& m)
...
              // This section is used to load the appropriate node parameters before spinning a moveit_cpp instance
              // Priority is given to parameters supplied directly via a config_dict, followed by launch parameters
              // and finally no supplied parameters.
+             std::vector<std::string> launch_arguments;
              if (!config_dict.is(py::none()))
              {
                auto utils = py::module::import("moveit.utils");
                // TODO (peterdavidfagan): replace python method with C++ method
                std::string params_filepath =
                    utils.attr("create_params_file_from_dict")(config_dict, "moveit_py").cast<std::string>();
```
JafarAbdi, line 70:
```suggestion
                   utils.attr("create_params_file_from_dict")(config_dict, node_name).cast<std::string>();
```

moveit_py/src/moveit/moveit_ros/moveit_cpp/moveit_cpp.cpp
```diff
@@ -58,40 +58,44 @@ void init_moveit_py(py::module& m)
...
              }
-             else
+
+             // Initialize ROS, pass launch arguments with rclcpp::init()
+             if (!rclcpp::ok())
              {
-               // TODO (peterdavidfagan): consider failing initialization if no params file is provided
-               node_options.allow_undeclared_parameters(true).automatically_declare_parameters_from_overrides(true);
+               RCLCPP_INFO(LOGGER, "Initialize rclcpp");
+               std::vector<char*> chars;
```
JafarAbdi, line 82:
```suggestion
               std::vector<const char*> chars;
```

moveit_py/src/moveit/moveit_ros/moveit_cpp/moveit_cpp.cpp
```diff
@@ -58,40 +58,44 @@ void init_moveit_py(py::module& m)
...
+               chars.reserve(launch_arguments.size());
+               for (const auto& arg : launch_arguments)
+               {
+                 chars.push_back(const_cast<char*>(arg.c_str()));
```
JafarAbdi, line 86:
```suggestion
                 chars.push_back(arg.c_str());
```

JafarAbdi (approved):

PR #2228, opened 2023-06-06
Don't default to random algorithm if no plugin is defined

sjahr:
### Description

Currently, if `planner_plugin_name_.empty()` & multiple planner plugin classes are defined, we default to the first one (In many cases that is CHOMP because the elements seem to be alphabetically ordered :shrug: ). This PR changes that to throw in case no planner plugin name is defined and multiple options are available because it is more useful to throw here rather than loading a random pipeline.

Also, I added throw statements at the end of the plugin lib exception catch blocks to propagate the exception upwards. Please correct me if that is not correct.

### Checklist

### DIFF
--- a/moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
+++ b/moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
@@ -53,7 +53,13 @@ planning_pipeline::PlanningPipeline::PlanningPipeline(const moveit::core::RobotM
                                                       const std::string& parameter_namespace,
                                                       const std::string& planner_plugin_param_name,
                                                       const std::string& adapter_plugins_param_name)
-  : active_{ false }, node_(node), parameter_namespace_(parameter_namespace), robot_model_(model)
+  : active_{ false }
+  , node_(node)
+  , parameter_namespace_(parameter_namespace)
+  , display_computed_motion_plans_{ false }
+  , publish_received_requests_{ false }
+  , robot_model_(model)
+  , check_solution_paths_{ false }
 {
   std::string planner_plugin_fullname = parameter_namespace_ + "." + planner_plugin_param_name;
   if (parameter_namespace_.empty())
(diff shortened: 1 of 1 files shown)


moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
```diff
@@ -110,42 +110,48 @@ void planning_pipeline::PlanningPipeline::configure()
...
-  {
-    planner_plugin_name_ = classes[0];
-    RCLCPP_INFO(
-        LOGGER,
-        "No '~planning_plugin' parameter specified, but only '%s' planning plugin is available. Using that one.",
-        planner_plugin_name_.c_str());
-  }
-  if (planner_plugin_name_.empty() && classes.size() > 1)
+
+  if (planner_plugin_name_.empty())
```
sjahr, line 120: Alternatively we could consider defaulting to OMPL with `planner_plugin_name_` :thinking:

henningkayser (approved):

moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
```diff
@@ -194,6 +196,10 @@ void planning_pipeline::PlanningPipeline::configure()
       }
     }
   }
+  else
+  {
+    RCLCPP_INFO(LOGGER, "No planning request adapter names specified.");
```
MarqRazz, line 201: This seems like it should be a warning. Is it valid to load MoveIt without any adapters specified?

sjahr: I thought about this too and concluded that it is not necessarily something invalid. If you have a super fancy planning plugin that generates a full motion plan response and that is robust without any pre-processing: Go for it. Nonetheless, I think this INFO print statement is nice to have because none of our existing planning plugins can do this.

MarqRazz: I think WARN sounds like it fits better because as you stated we don't have any planning plugins that can perform an end to end solution.

MarqRazz (approved):
Not going to block on the message severity. Everything else looks good.

moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
sjahr, line 201:
```suggestion
    RCLCPP_WARN(LOGGER, "No planning request adapter names specified.");
```

PR #2230, opened 2023-06-06
Fix CodeCov link on README

henningkayser:


henningkayser:
nevermind, I fixed the codecov settings

PR #2235, opened 2023-06-10
MoveItCpp/MoveItPy: Add blocking parameter back to execute

JafarAbdi:
### Description

Remove the deprecation, too!

### Checklist


sjahr:
Can you elaborate a bit more on why you'd like to re-add this? Is the TrajectoryExecutionManager non-blocking trajectory execution re-enabled? The reason for removing and deprecating this functionality is explained here https://github.com/ros-planning/moveit2/issues/1815

JafarAbdi:
> Can you elaborate a bit more on why you'd like to re-add this? Is the TrajectoryExecutionManager non-blocking trajectory execution re-enabled?

Given the diff in this PR, I don't think it was worth deprecating. Having a non-block execution is useful (at least in my use case :D)

> The reason for removing and deprecating this functionality is explained here #1815

I don't think that one is related to this PR, execute is async by default and it doesn't block. Fixing that TODO is just a few lines (unless I'm miss understood something about the code)

@cambel Do you mind giving this PR a review?

moveit_ros/planning/moveit_cpp/src/moveit_cpp.cpp
```diff
@@ -225,6 +218,10 @@ MoveItCpp::execute(const robot_trajectory::RobotTrajectoryPtr& robot_trajectory,
   robot_trajectory->getRobotTrajectoryMsg(robot_trajectory_msg);
   trajectory_execution_manager_->push(robot_trajectory_msg, controllers);
   trajectory_execution_manager_->execute();
+  if (!blocking)
```
cambel, line 221: The reason why non-blocking does not make sense to me without the "non-blocking execution" is that you don't get the result of the execution and you cannot execute anything unless the execution is completed or canceled. 
So I am not sure about your use case. Could you explain a bit about it?

sjahr:
@JafarAbdi What's the status of this PR?

JafarAbdi:
Closing because I no longer need this feature!

PR #2237, opened 2023-06-14
Add path publisher to STOMP smoothing adapter

sjahr:
### Description

Enable path visualization in STOMP planning request adapter + some minor cleanups

### Checklist


henningkayser (commented):
This would create a new path publisher with every planning request which is really not ideal. Better make it a member. I just checked, and it's also just a temporary instance for the planner mode. Unless there is a need for creating new publishers all the time - say, we are changing the topic all the time - these should be class members.

Additional nit: the former implementation has hidden the fact that the iteration/done callbacks are used for visualization. The explicit API for setting a path publisher adds another direct dependency to the planning context. I would find it more elegant to wrap and hide the functionality in callbacks and only pass those to the planning context, improving readability, separation of concerns, and DRY.

sjahr:
Closing this because it was not the right approach.

PR #2251, opened 2023-06-26
Prefer to use the active controller if multiple controllers apply

### LINKED ISSUE #1631: TrajectoryExecutionManager's default controller selection logic incorrectly prefers to use inactive controllers
### Your environment
* Humble
* Ubuntu 22.04
* 2.5.3

### Steps to reproduce
Use a robot with two controllers that both actuate the same set of joints, and code where MoveIt has to independently select which controller to use for a given trajectory (i.e., no user-specified controller).

### Expected behaviour
The TrajectoryExecutionManager uses the already-active controller to execute the trajectory.

### Actual behaviour
The TrajectoryExecutionManager deactivates the already-active controller, activates the inactive controller, and attempts to use it to execute the trajectory (without success, since this controller does not necessarily accept FollowJointTrajectory action goals).

### Suspicious Code

I think [these lines in `trajectory_execution_manager.cpp`](https://github.com/ros-planning/moveit2/blob/b050890a2632a723dd95c08542ef6ed57d77fb06/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp#L500-L504) are at fault:

```
    if (nractive[a] < nractive[b])
      return true;
    if (nractive[a] > nractive[b])
      return false;
```

As written, the comparison gives a higher rank to controllers where `nractive` is smaller, which is the case where the controller is not active.

stephanie-eng:
### Description

Resolves #1631 

Currently, if controller `a` is active , `nractive[a]` will be `1`. If it is inactive, `nractive[a]` will be `0`. Assuming controller `a` is active (`nractive[a] = 1`) and controller `b` is inactive (`nractive[b] = 0`), we would follow the second condition (`nractive[a] > nractive[b]`)and return `false`. This means that preference is given to inactive controllers.

In practice, if two controllers A and B are able to control the same number of joints and only controller A is active, `TrajectoryExecutionManager` will select the inactive controller B first and activate it. For a subsequent move, we select the now-inactive controller A, and use that, switching between controller A and B for each move.

This change prioritizes the active controller first, such that if controller A is active, it will be used for all subsequent moves until it is deactivated.

### DIFF
--- a/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
+++ b/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
@@ -539,9 +539,9 @@ struct OrderPotentialControllerCombination
       return false;
 
     // and then to active ones
-    if (nractive[a] < nractive[b])
-      return true;
     if (nractive[a] > nractive[b])
+      return true;
+    if (nractive[a] < nractive[b])
       return false;
 
     return false;


moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -539,9 +539,9 @@ struct OrderPotentialControllerCombination
       return false;
 
     // and then to active ones
-    if (nractive[a] < nractive[b])
-      return true;
     if (nractive[a] > nractive[b])
+      return true;
+    if (nractive[a] < nractive[b])
```
sjahr, line 544: You did not create them but can we replace these horrible variable names with something verbose? What does `nr` mean :see_no_evil: ?

sjahr (approved):

sjahr:
After talking to @stephanie-eng I'd say we merge this fix and then do a bigger refactoring of this code

sea-bass (approved):
The plan sounds good...

I was trying to read through it and have no idea what's going on, haha.

PR #2253, opened 2023-06-27
Add utility function to check trajectory continuity

sjahr:
### Description

This function checks the continuity of a trajectory by comparing all joint distances to a threshold value.

### Checklist


moveit_core/robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -630,6 +630,32 @@ void RobotTrajectory::print(std::ostream& out, std::vector<int> variable_indexes
   out.flush();
 }
 
+bool RobotTrajectory::isContinuous(const double threshold)
+{
+  bool is_continuous = true;
+  // Iterate over joint states & check if distance is higher than max_vel
```
sjahr, line 636:
```suggestion
  // Iterate over joint states & check if distance is higher than the given threshold
```

moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
```diff
@@ -382,6 +382,11 @@ class RobotTrajectory
    */
   void print(std::ostream& out, std::vector<int> variable_indexes = std::vector<int>()) const;
 
+  /** @brief Check if trajectory contains any joint jumps
+   *  @return True if trajectory is continuous
+   */
+  [[nodiscard]] bool isContinuous(const double threshold);
```
simonschmeisser, line 388:
```suggestion
  [[nodiscard]] bool isContinuous(const double threshold) const;
```

moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
```diff
@@ -319,7 +319,11 @@ bool planning_pipeline::PlanningPipeline::generatePlan(const planning_scene::Pla
       arr.markers.push_back(m);
 
       std::vector<std::size_t> index;
-      if (!planning_scene->isPathValid(*res.trajectory, req.path_constraints, req.group_name, false, &index))
+      if(!res.trajectory->isContinuous(0.1)){
```
simonschmeisser, line 322: shouldn't this be configurable somehow? Or depend on some existing parameters?

sjahr: @simonschmeisser I decided to remove the default check because the issue we originally seem to be caused by the code hierarchically above the planning pipeline. So this PR just adds some nice to have functionality.

moveit_core/robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -630,6 +630,32 @@ void RobotTrajectory::print(std::ostream& out, std::vector<int> variable_indexes
...
+    {
+      auto const joint_model_ptr = group_->getJointModel(joint_name);
+      auto const distance = joint_model_ptr->getDistanceFactor() *
+                            joint_model_ptr->distance(waypoints_.at(index)->getJointPositions(joint_name),
+                                                      waypoints_.at(index - 1)->getJointPositions(joint_name));
+      if (distance > threshold)
+      {
+        RCLCPP_WARN(rclcpp::get_logger("RobotTrajectory"),
+                    "Joint discontinuity detected for joint %s between waypoint %li and %li. The distance is %f and "
+                    "accedes the threshold %f",
```
sea-bass, line 649:
```suggestion
                    "exceeds the threshold %f",
```

moveit_core/robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -630,6 +630,32 @@ void RobotTrajectory::print(std::ostream& out, std::vector<int> variable_indexes
   out.flush();
 }
 
+bool RobotTrajectory::isContinuous(const double threshold)
+{
+  bool is_continuous = true;
+  // Iterate over joint states & check if distance is higher than the given threshold
+  for (std::size_t index = 1; index < waypoints_.size(); ++index)
```
sea-bass, line 637: We're not protecting this function if the trajectory has 0 or 1 points. I would just return true immediately at the beginning if that's the case.

moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
```diff
@@ -382,6 +382,11 @@ class RobotTrajectory
    */
   void print(std::ostream& out, std::vector<int> variable_indexes = std::vector<int>()) const;
 
+  /** @brief Check if trajectory contains any joint jumps
+   *  @return True if trajectory is continuous
+   */
+  [[nodiscard]] bool isContinuous(const double threshold) const;
```
sea-bass, line 388: The docstring should also describe what `threshold` is.

moveit_core/robot_trajectory/src/robot_trajectory.cpp
sjahr: :sweat_smile:

moveit_core/robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -630,6 +630,32 @@ void RobotTrajectory::print(std::ostream& out, std::vector<int> variable_indexes
   out.flush();
 }
 
+bool RobotTrajectory::isContinuous(const double threshold) const
+{
+  bool is_continuous = true;
```
Abishalini, line 635: Nit: Can we return true/false instead of assigning value to this variable and returning at the end?

Abishalini: It makes sense to print all discontinuities instead of early termination after detecting the first discontinuity.

moveit_core/robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -630,6 +630,32 @@ void RobotTrajectory::print(std::ostream& out, std::vector<int> variable_indexes
...
+  // Iterate over joint states & check if distance is higher than the given threshold
+  for (std::size_t index = 1; index < waypoints_.size(); ++index)
+  {
+    for (auto const& joint_name : group_->getJointModelNames())
```
henningkayser, line 639: group_ can be undefined

moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
```diff
@@ -382,6 +382,13 @@ class RobotTrajectory
    */
   void print(std::ostream& out, std::vector<int> variable_indexes = std::vector<int>()) const;
 
+  /** @brief Check if trajectory contains any joint jumps
+   *  @param threshold If the distance between two consecutive joint values in the trajectory exceeds this threshold,
+   * the trajectory is classified as discontinuous.
+   *  @return True if trajectory is continuous
+   */
+  [[nodiscard]] bool isContinuous(const double threshold) const;
```
henningkayser, line 390: instead of changing the API, let's make this a free function

henningkayser: and consider reusing the same code from the CartesianInterpolator

henningkayser (commented):
I think this implementation is not very useful, unfortunately. It only checks for position distances, ignoring that a RobotTrajectory has timestamps. A trajectory that only contains start and end state should be considered continuous as long as the distance can be moved in time. If you want to still keep this function, please rename to `hasJointJumps()` or add a function `getWayPointDistances()` (after all, there already is `getWayPointDurations()`) and call something like `std::max_element()` on it.

sjahr:
I'll close this PR and I've opened an issue to keep track of the task to free the Cartesian interpolator function.

PR #2257, opened 2023-06-29
Specify controller name in MGI execution

### LINKED ISSUE #1632: Allow specifying controllers when executing a plan via move group
## Background
This is motivated by MoveIt Studio, but has a more general use case. Suppose there are several controllers that can be toggled for a specific robot.

When executing a trajectory, either through `MoveItCpp::execute()` or `MoveGroupInterface::execute()`, there is currently no option to explicitly specify controller names. 

`TrajectoryExecutionManager` has its own logic to enable controllers based on the joints that need to be controller and/or the currently active controller, but sometimes it's preferred to be explicit about this. Additionally, in a separate Gitissue (https://github.com/ros-planning/moveit2/issues/1631), we additionally found a logic bug in which this logic is not always selecting the active controller that we manually switched to by directly calling the controller switching service.

## Desired Solution
What seems plausible is to expose a way to explicitly pass in controller names to override the logic that auto-assigns controllers based on joint names. This could happen in two ways I can think of, both of which unfortunately require some finessing of the interfaces.

### 1. Modify the `execute()` function interfaces
The following are just examples of adding an optional vector of strings for the controller names to specify, at the interface level.

Before:
```cpp
// MoveItCpp
MoveItCpp::execute(const std::string& group_name, const robot_trajectory::RobotTrajectoryPtr& robot_trajectory,
                   bool blocking)
// MoveGroupInterface
moveit::core::MoveItErrorCode execute(const moveit_msgs::msg::RobotTrajectory& trajectory, bool wait)
```

After:
```cpp
// MoveItCpp
MoveItCpp::execute(const std::string& group_name, const robot_trajectory::RobotTrajectoryPtr& robot_trajectory,
                   const std::optional<std::vector<std::string>>& controller_names, bool blocking)
// MoveGroupInterface
moveit::core::MoveItErrorCode execute(const moveit_msgs::msg::RobotTrajectory& trajectory, 
                  const std::optional<std::vector<std::string>>& controller_names, bool wait)
```

### 2. Modify the robot trajectory representations
Alternatively, we could keep the same arguments and modify the trajectory information so that the trajectory itself, or maybe each trajectory portion, has its set of controllers. We've done this with MTC in [this branch](https://github.com/JafarAbdi/moveit_task_constructor/blob/pr-ms_humble/msgs/msg/SubTrajectory.msg), where we add a new [`TrajectoryExecutionInfo` message](https://github.com/JafarAbdi/moveit_task_constructor/blob/pr-ms_humble/msgs/msg/TrajectoryExecutionInfo.msg) that contains controller names.

A similar thing could be done here for the `moveit_msgs::msg::RobotTrajectory` message definition in `moveit_msgs` (in the case of `MoveGroupInterface`), and in the `RobotTrajectory` class definition of `moveit_core`.

stephanie-eng:
Resolve #1632 

This PR adds the ability to execute via MGI while specifying a controller name. This can be useful for more fine control over controller-switching applications for which multiple controllers apply, and it is undesirable to let TrajectoryExecutionManager selet the controller for you.

Related and required change in moveit_msgs: https://github.com/ros-planning/moveit_msgs/pull/160

### DIFF
--- a/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
+++ b/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
@@ -716,7 +716,7 @@ public:
   moveit::core::MoveItErrorCode plan(Plan& plan);
 
   /** \brief Given a \e plan, execute it without waiting for completion.
-   *  \param [in] plan The motion plan for which
+   *  \param [in] plan The motion plan for which to execute
    *  \param [in] controllers An optional list of ros2_controllers to execute with. If none, MoveIt will attempt to find
    * a controller. The exact behavior of finding a controller depends on which MoveItControllerManager plugin is active.
    *  \return moveit::core::MoveItErrorCode::SUCCESS if successful
--- a/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
+++ b/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
@@ -715,17 +715,41 @@ public:
       target. No execution is performed. The resulting plan is stored in \e plan*/
   moveit::core::MoveItErrorCode plan(Plan& plan);
 
-  /** \brief Given a \e plan, execute it without waiting for completion. */
-  moveit::core::MoveItErrorCode asyncExecute(const Plan& plan);
+  /** \brief Given a \e plan, execute it without waiting for completion.
+   *  \param [in] plan The motion plan for which
+   *  \param [in] controllers An optional list of ros2_controllers to execute with. If none, MoveIt will attempt to find
+   * a controller. The exact behavior of finding a controller depends on which MoveItControllerManager plugin is active.
+   *  \return moveit::core::MoveItErrorCode::SUCCESS if successful
+   */
+  moveit::core::MoveItErrorCode asyncExecute(const Plan& plan,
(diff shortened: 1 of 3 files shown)


moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
```diff
@@ -715,17 +715,41 @@ class MOVEIT_MOVE_GROUP_INTERFACE_EXPORT MoveGroupInterface
...
-  /** \brief Given a \e robot trajectory, execute it without waiting for completion. */
-  moveit::core::MoveItErrorCode asyncExecute(const moveit_msgs::msg::RobotTrajectory& trajectory);
-
-  /** \brief Given a \e plan, execute it while waiting for completion. */
-  moveit::core::MoveItErrorCode execute(const Plan& plan);
-
-  /** \brief Given a \e robot trajectory, execute it while waiting for completion. */
-  moveit::core::MoveItErrorCode execute(const moveit_msgs::msg::RobotTrajectory& trajectory);
+  /** \brief Given a \e plan, execute it without waiting for completion.
+   *  \param [in] plan The motion plan for which
```
sea-bass, line 719: For which what?

sea-bass (approved):
Looks great, save one minor comment -- so LGTM!

sjahr (approved):

sea-bass (approved):

PR #2266, opened 2023-07-13
Cleanup planning request adapter interface

sjahr:
### Description

Cleans up the planning_request_adapter code and updates param handling to use generate_paramters_library.

### Checklist

### DIFF
--- a/moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
+++ b/moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
@@ -30,7 +30,7 @@ default_plan_request_adapter_parameters:
     description: "FixStartStateCollision: Maximum number of attempts to re-sample a valid start state.",
     default_value: 100,
     validation: {
-        lower_bounds<>: [ 0 ],
+        gt_eq<>: [ 0 ],
     }
   }
   start_state_max_bounds_error: {


moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
```diff
@@ -0,0 +1,44 @@
+default_plan_request_adapter_parameters:
```
sea-bass, line 1: It's not usually standard to call these files `default_` ... the file defines the parameters, but also happens to provide default values.

Maybe just remove the `default_` from the name?

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
```diff
@@ -0,0 +1,44 @@
...
+  path_tolerance: {
+    type: double,
+    description: "AddTimeOptimalParameterization: Tolerance per joint in which the time parameterization is allowed to deviate from the original path.",
+    default_value: 0.1,
+  }
+  resample_dt: {
+    type: double,
+    description: "AddTimeOptimalParameterization:Time steps between two waypoints of the parameterized trajectory. The trajectory is re-sampled from the original path.",
```
sea-bass, line 9: missing space

```suggestion
    description: "AddTimeOptimalParameterization: Time steps between two waypoints of the parameterized trajectory. The trajectory is re-sampled from the original path.",
```

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
```diff
@@ -0,0 +1,44 @@
...
+    default_value: 0.1,
+  }
+  min_angle_change: {
+    type: double,
+    description: "AddTimeOptimalParameterization: ?",
```
sea-bass, line 14: fill it in

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
```diff
@@ -0,0 +1,44 @@
...
+    default_value: 0.001,
+  }
+  start_state_max_dt: {
+    type: double,
+    description: "FixStartStateCollision/ FixStartStateBounds: Maximum temporal distance of the fixed start state from the original state.",
```
sea-bass, line 19:
```suggestion
    description: "FixStartStateCollision/FixStartStateBounds: Maximum temporal distance of the fixed start state from the original state.",
```

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
```diff
@@ -0,0 +1,44 @@
...
+    default_value: 0.5,
+  }
+  jiggle_fraction: {
+    type: double,
+    description: "FixStartStateCollision: ?",
```
sea-bass, line 24: fill it in

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
```diff
@@ -0,0 +1,44 @@
...
+    type: int,
+    description: "FixStartStateCollision: Maximum number of attempts to re-sample a valid start state.",
+    default_value: 100,
+    validation: {
+        bounds<>: [0, 255]
+    }
+  }
+  start_state_max_bounds_error: {
+    type: double,
+    description: "FixStartStateBounds: ?",
```
sea-bass, line 37: fill it in

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
```diff
@@ -0,0 +1,44 @@
...
+    default_value: 0.05,
+  }
+  default_workspace_bounds: {
+    type: double,
+    description: "FixWorkspaceBounds: Default workspace bounds representing a 10m x 10m x 10m box around the robot's origin.",
```
sea-bass, line 42: I wouldn't specify the actual 10m dimension in this comment. Just say that it's a box of this size in each dimension.

Also, weird that this parameter wouldn't be a vector...

moveit_ros/planning/planning_request_adapter_plugins/CMakeLists.txt
```diff
@@ -18,15 +20,3 @@ ament_target_dependencies(moveit_default_planning_request_adapter_plugins
...
-ament_target_dependencies(moveit_list_request_adapter_plugins
-  Boost
-  moveit_core
-  rclcpp
-  pluginlib
-)
-
-install(TARGETS moveit_list_request_adapter_plugins
-  DESTINATION lib/${PROJECT_NAME}
-)
```
sea-bass, line 32: I like the code cleanup here, and also agree that it's weird just having this as a standalone executable... but is there an alternative way to list planning adapters that we can still use?

I'm assuming yes, and it's a `ros2 param get` type of situation, right?

sjahr: No really, because this executable checks the available c++ plugins

sjahr: I moved the executable into its own directory so it is still usable + detectable. In a follow up PR this executable could be refactored to list planner_plugins or IK plugins as well :thinking:

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
sjahr: The default is referring to the group of plugins for which this file provides parameters. Those are called `default_plan_request_adapters` but I get why this might be confusing. Do you think we should rename it nonetheless?

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
sjahr: Yes :sweat_smile: We could consider changing that in a follow up PR

sjahr, line 42:
```suggestion
    description: "FixWorkspaceBounds: Default workspace bounds representing a cube around the robot's origin whose edge length this parameter defines.",
```

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
```diff
@@ -0,0 +1,44 @@
...
+    default_value: 0.1,
+  }
+  resample_dt: {
+    type: double,
+    description: "AddTimeOptimalParameterization: Time steps between two waypoints of the parameterized trajectory. The trajectory is re-sampled from the original path.",
+    default_value: 0.1,
+  }
+  min_angle_change: {
+    type: double,
+    description: "AddTimeOptimalParameterization: ?",
```
sjahr, line 14:
```suggestion
    description: "min_angle_change: ?",
```

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
sjahr: Help wanted with the description

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
sjahr: Help wanted with the description

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
sea-bass: Ah! I see, then it's fine to leave it.

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
sea-bass:
```suggestion
    description: "AddTimeOptimalParameterization: Minimum angle change to consider two waypoints unique.",
```

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
sea-bass:
```suggestion
    description: "FixStartStateCollision: Joint value perturbation as a percentage of the total range of motion for the joint.",
```

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
sea-bass:
```suggestion
    description: "FixStartStateBounds: Maximum allowable joint value displacement from starting configuration.",
```

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
```diff
@@ -0,0 +1,44 @@
...
+    default_value: 0.1,
+  }
+  resample_dt: {
+    type: double,
+    description: "AddTimeOptimalParameterization: Time steps between two waypoints of the parameterized trajectory. The trajectory is re-sampled from the original path.",
+    default_value: 0.1,
+  }
+  min_angle_change: {
+    type: double,
+    description: "min_angle_change: ?",
```
sjahr, line 14:
```suggestion
    description: "AddTimeOptimalParameterization: Minimum joint value change to consider two waypoints unique.",
```

sjahr:
@sea-bass Thank you!

sea-bass (approved):

moveit_ros/planning/planning_request_adapter_plugins/src/fix_workspace_bounds.cpp
```diff
@@ -91,11 +93,8 @@ class FixWorkspaceBounds : public planning_request_adapter::PlanningRequestAdapt
   }
 
 private:
-  rclcpp::Node::SharedPtr node_;
-  double workspace_extent_;
+  std::shared_ptr<default_plan_request_adapter_parameters::ParamListener> param_listener_;
```
kylc, line 96: Could this be a `std::unique_ptr`?

moveit_ros/planning/planning_request_adapter_plugins/src/fix_workspace_bounds.cpp
```diff
@@ -53,13 +54,10 @@ static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_ros.fix_workspac
...
 public:
-  static const std::string WBOUNDS_PARAM_NAME;
-
   void initialize(const rclcpp::Node::SharedPtr& node, const std::string& parameter_namespace) override
   {
-    node_ = node;
-    workspace_extent_ = getParam(node_, LOGGER, parameter_namespace, WBOUNDS_PARAM_NAME, 10.0);
-    workspace_extent_ /= 2.0;
+    param_listener_ =
+        std::make_shared<default_plan_request_adapter_parameters::ParamListener>(node, parameter_namespace);
```
kylc, line 60: nit: `initialize` should take the node by const-value and then `std::move` it here.

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_collision.cpp
```diff
@@ -184,15 +178,8 @@ class FixStartStateCollision : public planning_request_adapter::PlanningRequestA
   }
 
 private:
-  rclcpp::Node::SharedPtr node_;
-  double max_dt_offset_;
-  double jiggle_fraction_;
-  int sampling_attempts_;
+  std::shared_ptr<default_plan_request_adapter_parameters::ParamListener> param_listener_;
```
kylc, line 181: Can this be `std::unique_ptr`?

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_collision.cpp
```diff
@@ -58,21 +60,10 @@ static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_ros.fix_start_st
...
-    max_dt_offset_ = getParam(node_, LOGGER, parameter_namespace, DT_PARAM_NAME, 0.5);
-    jiggle_fraction_ = getParam(node_, LOGGER, parameter_namespace, JIGGLE_PARAM_NAME, 0.02);
-    sampling_attempts_ = getParam(node_, LOGGER, parameter_namespace, ATTEMPTS_PARAM_NAME, 100);
-    if (sampling_attempts_ < 1)
-    {
-      sampling_attempts_ = 1;
-      RCLCPP_WARN(LOGGER, "Param '%s' needs to be at least 1.", ATTEMPTS_PARAM_NAME.c_str());
-    }
+    param_listener_ =
+        std::make_shared<default_plan_request_adapter_parameters::ParamListener>(node, parameter_namespace);
```
kylc, line 66: nit: `initialize` should take the node by const-value and then `std::move` it here.

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
```diff
@@ -0,0 +1,44 @@
...
+  start_state_max_bounds_error: {
+    type: double,
+    description: "FixStartStateBounds: Maximum allowable error outside joint bounds for the starting configuration.",
+    default_value: 0.05,
+  }
+  default_workspace_bounds: {
+    type: double,
+    description: "FixWorkspaceBounds: Default workspace bounds representing a cube around the robot's origin whose edge length this parameter defines.",
+    default_value: 10.0,
+  }
```
kylc, line 44: As someone not familiar with the code, it's not clear to me why the class names are important in the parameter descriptions ("ClassName: description" format). Is that relevant information to the user?

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
```diff
@@ -0,0 +1,44 @@
...
+    type: double,
+    description: "FixStartStateCollision: Joint value perturbation as a percentage of the total range of motion for the joint.",
+    default_value: 0.02,
+  }
+  max_sampling_attempts: {
+    type: int,
+    description: "FixStartStateCollision: Maximum number of attempts to re-sample a valid start state.",
+    default_value: 100,
+    validation: {
+        bounds<>: [0, 255]
```
kylc, line 32: Why the upper bound here? Maybe I want to sample all day... :smile:

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
```diff
@@ -0,0 +1,44 @@
...
+    default_value: 0.1,
+  }
+  min_angle_change: {
+    type: double,
+    description: "AddTimeOptimalParameterization: Minimum joint value change to consider two waypoints unique.",
+    default_value: 0.001,
+  }
+  start_state_max_dt: {
+    type: double,
+    description: "FixStartStateCollision/FixStartStateBounds: Maximum temporal distance of the fixed start state from the original state.",
```
kylc, line 19: How is this temporal distance computed? As a user, it's now clear to me how to set this parameter because there is no context about how the trajectory to the fixed start state from the original state is derived.

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
```diff
@@ -0,0 +1,44 @@
...
+    type: double,
+    description: "AddTimeOptimalParameterization: Tolerance per joint in which the time parameterization is allowed to deviate from the original path.",
+    default_value: 0.1,
+  }
+  resample_dt: {
+    type: double,
+    description: "AddTimeOptimalParameterization: Time steps between two waypoints of the parameterized trajectory. The trajectory is re-sampled from the original path.",
+    default_value: 0.1,
+  }
+  min_angle_change: {
```
kylc, line 12: Canonical term for this might be `angle_epsilon` or something?

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
```diff
@@ -0,0 +1,44 @@
+default_plan_request_adapter_parameters:
+  path_tolerance: {
+    type: double,
+    description: "AddTimeOptimalParameterization: Tolerance per joint in which the time parameterization is allowed to deviate from the original path.",
+    default_value: 0.1,
+  }
+  resample_dt: {
+    type: double,
+    description: "AddTimeOptimalParameterization: Time steps between two waypoints of the parameterized trajectory. The trajectory is re-sampled from the original path.",
```
kylc, line 9:
```suggestion
    description: "AddTimeOptimalParameterization: Time steps between two adjacent waypoints of the parameterized trajectory. The trajectory is re-sampled from the original path.",
```

moveit_ros/planning/planning_request_adapter_plugins/CMakeLists.txt
```diff
@@ -9,7 +11,7 @@ set(SOURCE_FILES
   src/resolve_constraint_frames.cpp
 )
 
-add_library(moveit_default_planning_request_adapter_plugins SHARED ${SOURCE_FILES})
+target_link_libraries(moveit_default_planning_request_adapter_plugins default_plan_request_adapter_parameters)
```
kylc, line 14:
```suggestion
target_link_libraries(moveit_default_planning_request_adapter_plugins PRIVATE default_plan_request_adapter_parameters)
```

Should be PRIVATE I think, since it is not exposed in the interface. If it will let you, I know ament sometimes complains about this...

moveit_ros/planning/introspection/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+add_executable(moveit_list_planning_adapter_plugins src/list_planning_adapter_plugins.cpp)
+ament_target_dependencies(moveit_list_planning_adapter_plugins
+  Boost
```
kylc, line 3: I don't see any boost references in `list_planning_adapter_plugins.cpp`.

moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
```diff
@@ -103,10 +89,11 @@ class PlanningRequestAdapter
...
    *  @return True if response got generated correctly */
-  bool adaptAndPlan(const planning_interface::PlannerManagerPtr& planner,
-                    const planning_scene::PlanningSceneConstPtr& planning_scene,
-                    const planning_interface::MotionPlanRequest& req, planning_interface::MotionPlanResponse& res,
-                    std::vector<std::size_t>& added_path_index) const;
+  [[nodiscard]] bool adaptAndPlan(const planning_interface::PlannerManagerPtr& planner,
+                                  const planning_scene::PlanningSceneConstPtr& planning_scene,
+                                  const planning_interface::MotionPlanRequest& req,
+                                  planning_interface::MotionPlanResponse& res,
+                                  std::vector<std::size_t>& added_path_index = EMPTY_PATH_INDEX_VECTOR) const;
```
kylc, line 96: This pattern of passing the default `std::vector<std::size_t>& added_path_index = EMPTY_PATH_INDEX_VECTOR` isn't making sense to me. It is an out-parameter? If the user doesn't pass their own, does it not mutate the global `EMPTY_PATH_INDEX_VECTOR`, which surely is not the intent?

kylc (changes requested):

moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
kylc: Also note that clang-tidy CI is complaining about the naming of this and failing.

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
sjahr: Yes, because this file contains the parameters for all DefaultPlanRequestAdapters which are multiple different classes

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
sjahr: :+1: 
upper_bound_gone

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
sjahr: It is not computed but heuristically chosen:
```cpp
// heuristically decide a duration offset for the trajectory (induced by the additional point added as a
// prefix to the computed trajectory)
res.trajectory->setWayPointDurationFromPrevious(0, std::min(params.start_state_max_dt,
                                                                      res.trajectory->getAverageSegmentDuration()));
```
Looking at this it might be worth creating a follow-up PR to change this to something that respects the joint limits & take spatial distance into account :thinking:

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
sjahr: I think `min_angle_change` is more descriptive, so I'd be in favor of keeping it

moveit_ros/planning/introspection/CMakeLists.txt
sjahr: Good catch

sjahr, line 3:
```suggestion
```

moveit_ros/planning/planning_request_adapter_plugins/src/fix_workspace_bounds.cpp
sjahr: Not a huge expert on this but I think the main reason why we pass it by reference is

> The advantage of passing the shared_ptr by const& is that the reference count doesn't have to be increased and then decreased. Because these operations have to be thread-safe, they can be expensive.

(Copied from more detailed answer here: https://stackoverflow.com/questions/37610494/passing-const-shared-ptrt-versus-just-shared-ptrt-as-parameter)

moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
sjahr: My intend with this was to remove these functions:

```cpp
bool PlanningRequestAdapterChain::adaptAndPlan(const planning_interface::PlannerManagerPtr& planner,
                                               const planning_scene::PlanningSceneConstPtr& planning_scene,
                                               const planning_interface::MotionPlanRequest& req,
                                               planning_interface::MotionPlanResponse& res) const
{
  std::vector<std::size_t> empty_added_path_index;
  return adaptAndPlan(planner, planning_scene, req, res, empty_added_path_index);
}
```
As far as I understand there are no optional references, so like before we just write the path indices to a memory location we never read and dump them that way. Maybe another possible implementation would be to change this interface to use a pointer but that would be a bigger change. Do you have a smarter idea of how to do this?

moveit_ros/planning/planning_request_adapter_plugins/CMakeLists.txt
sjahr: Yes :disappointed: I tried it and got 
```
  Starting >>> moveit_ros_planning
  --- stderr: moveit_ros_planning
  CMake Error at /opt/ros/rolling/share/ament_cmake_target_dependencies/cmake/ament_target_dependencies.cmake:145 (target_link_libraries):
    The keyword signature for target_link_libraries has already been used with
    the target "moveit_default_planning_request_adapter_plugins".  All uses of
    target_link_libraries with a target must be either all-keyword or
    all-plain.
  
    The uses of the keyword signature are here:
  
     * planning_request_adapter_plugins/CMakeLists.txt:14 (target_link_libraries)
  
    planning_request_adapter_plugins/CMakeLists.txt:17 (ament_target_dependencies)
  
  
  ---
  Failed   <<< moveit_ros_planning [8.90s, exited with code 1]
  Aborted  <<< moveit_planners_stomp [27.2s]
  
  Summary: 6 packages finished [1min 14s]
    1 package failed: moveit_ros_planning
    1 package aborted: moveit_planners_stomp
    3 packages had stderr output: moveit_common moveit_configs_utils moveit_ros_planning
    32 packages not processed
'build_target_workspace' returned with code '1' after 1 min 14 sec
```
I guess the `ament_target_dependencies` macro is the culprit since apparently, [you cannot call target_link_libs with PUBLIC and PRIVATE](https://stackoverflow.com/questions/47737558/uses-of-target-link-libraries-must-be-either-all-keyword-or-all-plain). Let's keep it public?

moveit_ros/planning/planning_request_adapter_plugins/CMakeLists.txt
```diff
@@ -9,7 +11,7 @@ set(SOURCE_FILES
   src/resolve_constraint_frames.cpp
 )
 
-add_library(moveit_default_planning_request_adapter_plugins SHARED ${SOURCE_FILES})
+target_link_libraries(moveit_default_planning_request_adapter_plugins PRIVATE default_plan_request_adapter_parameters)
```
sjahr, line 14:
```suggestion
target_link_libraries(moveit_default_planning_request_adapter_plugins default_plan_request_adapter_parameters)
```

moveit_ros/planning/planning_request_adapter_plugins/src/fix_workspace_bounds.cpp
henningkayser: From the base interface it's not clear if the node is ignored, only used in the function, or transferred. Copying by default doesn't seem like the better option to me.

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
henningkayser: in that case, does the adapter take care of timeouts? If not, unlimited sampling wouldn't be too useful

moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
henningkayser: can't you just use `std::vector<std::size_t>& added_path_index = {}`?

moveit_ros/planning/planning_request_adapter_plugins/src/fix_workspace_bounds.cpp
kylc: Sounds good

moveit_ros/planning/planning_request_adapter_plugins/res/default_plan_request_adapter_params.yaml
kylc: Ok, maybe there is more than the parameter description lets on but it probably belongs in a follow-up.

moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
kylc: @henningkayser C++ won't let you assign a default reference like that.

@sjahr I guess this pattern will technically work as long as `adaptAndPlan` makes sure to clear the "empty" vector before it starts every time. Otherwise it's a memory leak!

kylc (approved):
Approved pending 1 remaining comment above.

moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
henningkayser: oh of course not... content-wise, this parameter would fit into the MotionPlanResponse as well

henningkayser (approved):

PR #2288, opened 2023-08-05
Use generate parameters library in PlanningPipelineClass + general cleanups

sjahr:
### Description

To avoid breaking MTC https://github.com/ros-planning/moveit_task_constructor/pull/450 needs to be merged.

- [X] Use generate parameters library for parameter handling
- [X] Add `nodiscard` to return values
- [X] Some small cleanup suggestions
- [X] Reduce state of PlanningPipeline class

### Checklist


moveit_ros/planning/planning_pipeline/include/moveit/planning_pipeline/planning_pipeline.h
```diff
@@ -63,102 +63,71 @@ class MOVEIT_PLANNING_PIPELINE_EXPORT PlanningPipeline
...
-     specified
-      \param adapter_plugins_param_name The name of the ROS parameter under which the names of the request adapter
-     plugins are specified (plugin names separated by space; order matters)
+      \param parameter_namespace parameter namespace where the planner configurations are stored
   */
   PlanningPipeline(const moveit::core::RobotModelConstPtr& model, const std::shared_ptr<rclcpp::Node>& node,
-                   const std::string& parameter_namespace,
-                   const std::string& planning_plugin_param_name = "planning_plugin",
-                   const std::string& adapter_plugins_param_name = "request_adapters");
+                   const std::string& parameter_namespace);
```
sea-bass, line 80: I think the one comment I have is on this interface.

By removing the extra 2 string arguments, we're breaking compatibility. Is there path to maybe not removing this interface, but marking it with a deprecation warning, at least?

moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
```diff
@@ -387,30 +334,33 @@ bool planning_pipeline::PlanningPipeline::generatePlan(const planning_scene::Pla
...
-  if (display_computed_motion_plans_ && solved)
-  {
-    moveit_msgs::msg::DisplayTrajectory disp;
-    disp.model_id = robot_model_->getName();
-    disp.trajectory.resize(1);
-    res.trajectory->getRobotTrajectoryMsg(disp.trajectory[0]);
-    moveit::core::robotStateToRobotStateMsg(res.trajectory->getFirstWayPoint(), disp.trajectory_start);
-    display_path_publisher_->publish(disp);
+    // Optionally publish DisplayTrajectory msg
+    if (display_computed_motion_plans)
```
sea-bass, line 339: Why remove the `&& solved`?

moveit_ros/planning/planning_pipeline_interfaces/include/moveit/planning_pipeline_interfaces/planning_pipeline_interfaces.hpp
```diff
@@ -111,8 +111,6 @@ const std::vector<::planning_interface::MotionPlanResponse> planWithParallelPipe
 std::unordered_map<std::string, planning_pipeline::PlanningPipelinePtr>
 createPlanningPipelineMap(const std::vector<std::string>& pipeline_names,
                           const moveit::core::RobotModelConstPtr& robot_model, const rclcpp::Node::SharedPtr& node,
-                          const std::string& parameter_namespace = std::string(),
-                          const std::string& planning_plugin_param_name = "planning_plugin",
-                          const std::string& adapter_plugins_param_name = "request_adapters");
+                          const std::string& parameter_namespace = std::string());
```
sea-bass, line 114: Same here re: changing the interface without a deprecation warning?

sea-bass (commented):
Code looks good, and I like the cleanup and extra error checks you've done along the way.

Mostly just have one question about changing a few function signatures and whether we should try do a more gradual change? But maybe the default arguments are most often left default and this doesn't have a huge impact?

moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
sjahr: Hard to see from the diff but I moved the whole thing into the scope of `if (solved && res.trajectory)` to get rid of the additional solved check

sea-bass: That would do it! Lol

moveit_ros/planning/planning_pipeline/include/moveit/planning_pipeline/planning_pipeline.h
sjahr: I don't think that this has a huge impact and I think that very few people (if any) use the pure PlanningPipeline class and I'd be surprised if they manually changed the parameter name of the planning_plugin and request_adapters. The main reason for removing this interface was that I felt it was relatively pointless, so I did not bother about keeping a deprecated version. If you'd prefer a proper deprecation I can add that :+1:

sea-bass: Nah, I believe you if you say it's not gonna affect a ton of people. Let's do this!

moveit_planners/pilz_industrial_motion_planner/src/command_list_manager.cpp
```diff
@@ -239,7 +239,11 @@ CommandListManager::solveSequenceItems(const planning_scene::PlanningSceneConstP
     setStartState(motion_plan_responses, req.group_name, req.start_state);
 
     planning_interface::MotionPlanResponse res;
-    planning_pipeline->generatePlan(planning_scene, req, res);
+    if (!planning_pipeline->generatePlan(planning_scene, req, res))
+    {
+      RCLCPP_ERROR(LOGGER, "Generating a plan with planning pipeline failed.");
+      res.error_code.val = moveit_msgs::msg::MoveItErrorCodes::FAILURE;
```
Abishalini, line 245: Does `generatePlan` already not assign an error code? Just making sure we are not overriding it with the generic FAILURE error code here

Abishalini (approved):
The changes look good to me. I just have one question

moveit_planners/pilz_industrial_motion_planner/src/command_list_manager.cpp
henningkayser: yes, technically some adapter or the planner should put its own error message into the result.

sjahr: I think it is possible for the chain to fail before anything is written to the error code. E.g. if the chain throws an [exception early](https://github.com/ros-planning/moveit2/blob/095b8ae9f6e3f6218a2f32dd1fbe33a843b085a2/moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp#L233). That is also the only case when generatePlan returns false

PR #2292, opened 2023-08-10
Fix linking error with cached_ik_kinematics_plugin

sloretz:
### Description

This fixes a bug caused by #1677 which accidentally inverted the dependency between the plugin and the base library. This fixes it by making the plugin again depend on the base library, and adding missing dependencies to the base library.

I'm not sure why the base library is necessary. It seems like we could add `ik_cache.cpp` to the plugin library, but I fixed it this way because I would like to request this fix be backported to ROS Iron.

Steps to reproduce (sorry I didn't spend the time to make an MRE)

1. Launch a `moveit_ros_move_group move_group` node with parameters that includes

```yaml
robot_description_kinematics:
  some_group_name:
    kinematics_solver: cached_ik_kinematics_plugin/CachedKDLKinematicsPlugin
```

2. Observe that it errors with:

```
symbol lookup error: /path/to/ws/install/moveit_kinematics/lib/libmoveit_cached_ik_kinematics_plugin.so: undefined symbol: _ZN27cached_ik_kinematics_plugin7IKCacheC1Ev
```

### Checklist

### DIFF
--- a/moveit_kinematics/cached_ik_kinematics_plugin/CMakeLists.txt
+++ b/moveit_kinematics/cached_ik_kinematics_plugin/CMakeLists.txt
@@ -9,7 +9,6 @@ ament_target_dependencies(moveit_cached_ik_kinematics_base
   moveit_core
   moveit_msgs
 )
-target_link_libraries(moveit_cached_ik_kinematics_base moveit_cached_ik_kinematics_plugin)
 
 if(trac_ik_kinematics_plugin_FOUND)
     include_directories(${trac_ik_kinematics_plugin_INCLUDE_DIRS})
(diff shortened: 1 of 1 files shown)


sjahr (approved):
Thanks for this fix! I agree that the base library seems to be superfluous. It would be nice to have it removed and the `ik_cache.cpp` added to the two libraries. We can backport it to ROS Iron.

sjahr:
Minor blocker for this: https://github.com/ros-planning/moveit2/pull/2293

moveit_kinematics/cached_ik_kinematics_plugin/CMakeLists.txt
```diff
@@ -20,6 +19,10 @@ generate_parameter_library(
   include/moveit/cached_ik_kinematics_plugin/cached_ik_kinematics_parameters.yaml # path to input yaml file
 )
 
+target_link_libraries(moveit_cached_ik_kinematics_base
```
ChrisThrasher, line 22: While you're touching this code, please specify `PRIVATE` or `PUBLIC` for these link libraries. MoveIt needs to transition to using those keywords on all `target_link_libraries` calls.

sloretz: Using keywords in 68a6ee278

ChrisThrasher (approved):

PR #2295, opened 2023-08-11
[Servo] Add additional info about twist frame conversion 

ibrahiminfinite:
### Description

This PR adds some info for the users about the limitations of twist frame conversion currently used in Servo.

### Checklist

Ping : @AndyZe  @sea-bass

### DIFF
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
@@ -134,8 +134,11 @@ public:
 private:
   /**
    * \brief Convert a give twist command to planning frame
-   * (This implementation assumes that source and target frames are stationary)
+   * This implementation does not differentiate between body twist and spatial twist,
+   * i.e if the command is as seen from a stationary or non-stationary frame.
+   * The result of this transformation will only be accruate if the command frame is a stationary frame.
    * See issue: https://github.com/ros-planning/moveit2/issues/2150
+   * TODO: Update with proper transformations.
    * @param command The twist command to be converted
    * @return The transformed twist command.
    */
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
@@ -136,9 +136,8 @@ private:
    * \brief Convert a give twist command to planning frame
    * This implementation does not differentiate between body twist and spatial twist,
    * i.e if the command is as seen from a stationary or non-stationary frame.
-   * The result of this transformation will only be accruate if the command frame is a stationary frame.
+   * The result of this transformation will only be accurate if the command frame is a stationary frame.
    * See issue: https://github.com/ros-planning/moveit2/issues/2150
-   * TODO: Update with proper transformations.
    * @param command The twist command to be converted
    * @return The transformed twist command.
    */
--- a/moveit_ros/moveit_servo/src/servo.cpp
+++ b/moveit_ros/moveit_servo/src/servo.cpp
@@ -47,6 +47,7 @@ namespace
 const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_servo.servo");
 constexpr double ROBOT_STATE_WAIT_TIME = 5.0;  // seconds
 constexpr double STOPPED_VELOCITY_EPS = 1e-4;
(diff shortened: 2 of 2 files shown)


moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
```diff
@@ -134,8 +134,11 @@ class Servo
 private:
   /**
    * \brief Convert a give twist command to planning frame
-   * (This implementation assumes that source and target frames are stationary)
+   * This implementation does not differentiate between body twist and spatial twist,
+   * i.e if the command is as seen from a stationary or non-stationary frame.
+   * The result of this transformation will only be accruate if the command frame is a stationary frame.
    * See issue: https://github.com/ros-planning/moveit2/issues/2150
+   * TODO: Update with proper transformations.
```
AndyZe, line 141: I'd suggest deleting the TODO b/c we know you're going to be fixing it in the next few days, and the fix isn't going to be in this file anyway.

```suggestion
```

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -462,6 +463,11 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
 
 const TwistCommand Servo::toPlanningFrame(const TwistCommand& command)
 {
+  if (command.frame_id != servo_params_.planning_frame)
+  {
+    RCLCPP_WARN_STREAM_THROTTLE(LOGGER, *node_->get_clock(), ROS_LOG_THROTTLE_PERIOD,
+                                "Twist command is not in planning frame, conversion may not be accurate.");
```
AndyZe, line 469:
```suggestion
                                "Twist command is not in planning frame, transformation may not be accurate.");
```

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
```diff
@@ -134,8 +134,11 @@ class Servo
 private:
   /**
    * \brief Convert a give twist command to planning frame
-   * (This implementation assumes that source and target frames are stationary)
+   * This implementation does not differentiate between body twist and spatial twist,
+   * i.e if the command is as seen from a stationary or non-stationary frame.
+   * The result of this transformation will only be accruate if the command frame is a stationary frame.
```
sea-bass:
```suggestion
   * The result of this transformation will only be accurate if the command frame is a stationary frame.

Comment (ibrahiminfinite) [inline moveit_ros/moveit_servo/include/moveit_servo/servo.hpp:139]:
```suggestion
   * The result of this transformation will only be accurate if the command frame is a stationary frame.
```

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -47,6 +47,7 @@ namespace
 const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_servo.servo");
 constexpr double ROBOT_STATE_WAIT_TIME = 5.0;  // seconds
 constexpr double STOPPED_VELOCITY_EPS = 1e-4;
+constexpr auto ROS_LOG_THROTTLE_PERIOD = std::chrono::milliseconds(3000).count();
```
sea-bass, line 50: I think the warning throttle just takes the time in seconds, so it seems roundabout to make a `chrono` duration to then just extract the count.

```suggestion
constexpr size_t ROS_LOG_THROTTLE_PERIOD = 3;  // seconds
```

ibrahiminfinite:
```
 * \param duration The duration of the throttle interval as an integral value in milliseconds.
```
I followed the pattern from the old implementation, it felt more expressive, I guess it can be a `size_t` with a comment also that it is "milliseconds"

AndyZe (approved):

moveit_ros/moveit_servo/src/servo.cpp
sea-bass: Oh wait, it accepts milliseconds, not seconds? So that means that the thing you copied from @AndyZe 's old implementation is actually the correct one, and the other stuff in the moveit2 repo is wrongly using seconds as milliseconds. Cool.

As you can see, my confusion is shared in https://github.com/ros2/rclcpp/issues/1929, which indeed would be nice.

OK you can leave it as is.

sea-bass (approved):

PR #2224, opened 2023-06-03
 [WIP] [Servo] Refactoring servo

ibrahiminfinite:
### Description

This is an effort supported by GSoC 2023 to refactor the MoveIt Servo package to improve its usability and performance.

The joint jog and twist based demos for the C++ API are currently working.
Please feel free to test it out and provide your feedback.

TODO :
- [ ] C++ joint jog API + demo
- [ ] C++ twist API + demo
- [ ] C++ pose API + demo
- [ ] Tests for C++ APIs
- [ ] ROS start, stop services
- [ ] ROS input type switching service
- [ ] ROS joint jog API + demo
- [ ] ROS twist API + demo
- [ ] ROS pose API + demo
- [ ] Tests for ROS APIs
- [ ] Keyboard teleop demo
- [ ] Joystick teleop demo
- [ ] Update moveit_tutorials

PR #2301, opened 2023-08-15
[Servo] Fix twist transformation

### LINKED ISSUE #2150: Wrong equations used in Servo twist transformations
It was pointed out to me that Servo uses the wrong equations to transform twist. Servo does this:

`w' = R * w` and `v' = R * v`

Probably equation 3.83 or 3.84 from here should be used instead:

http://hades.mech.northwestern.edu/images/7/7f/MR.pdf

Whether it's Eq. 3.83 or 3.84 depends on whether the frame is fixed or attached to a robot link.

This is only an issue for large angular velocities.

ibrahiminfinite:
### Description

Fixes #2150 

This was partly addressed in #2224, but the input ordering of the twist command to the `tf2::doTransform` was wrong.
Since Servo used (linear, angular) and `tf2::doTransform` uses (angular, linear).

----

Assumptions about the frames used in the context of twist commands:
- The frame specified by the parameter `planning_frame` is always stationary.
- The command maybe given in any stationary frame or the end effector frame.
- Any other moving frames other than ee frame will also still transform, but not sure if the resulting velocities will be accurate.

### DIFF
--- a/moveit_core/planning_interface/include/moveit/planning_interface/planning_response.h
+++ b/moveit_core/planning_interface/include/moveit/planning_interface/planning_response.h
@@ -65,6 +65,11 @@ struct MotionPlanResponse
   moveit_msgs::msg::RobotState start_state;
   std::string planner_id;
 
+  /// Sometimes planning request adapters may add states on the solution path (e.g., add the current state of the robot
+  /// as prefix, when the robot started to plan only from near that state, as the current state itself appears to touch
+  /// obstacles). This is helpful because the added states should not be considered invalid in all situations.
+  std::vector<std::size_t> added_path_index;  // This won't be included into the MotionPlanResponse ROS 2 message!
+
   // \brief Enable checking of query success or failure, for example if(response) ...
   explicit operator bool() const
   {
--- a/moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
+++ b/moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
@@ -46,10 +46,6 @@
 /** \brief Generic interface to adapting motion planning requests */
 namespace planning_request_adapter
 {
-namespace
-{
-std::vector<std::size_t> empty_path_index_vector = {};
-}
 MOVEIT_CLASS_FORWARD(PlanningRequestAdapter);  // Defines PlanningRequestAdapterPtr, ConstPtr, WeakPtr... etc
 
 /** @brief Concept in MoveIt which can be used to modify the planning problem and resulting trajectory (pre-processing
@@ -84,16 +80,11 @@ public:
    *  @param planning_scene Representation of the environment for the planning
    *  @param req Motion planning request with a set of constraints
    *  @param res Reference to which the generated motion plan response is written to
-   *  @param added_path_index Sometimes planning request adapters may add states on the solution path (e.g.,
-     add the current state of the robot as prefix, when the robot started to plan only from near that state, as the
-     current state itself appears to touch obstacles). This is helpful because the added states should not be considered
-     invalid in all situations.
    *  @return True if response got generated correctly */
   [[nodiscard]] bool adaptAndPlan(const planning_interface::PlannerManagerPtr& planner,
                                   const planning_scene::PlanningSceneConstPtr& planning_scene,
                                   const planning_interface::MotionPlanRequest& req,
-                                  planning_interface::MotionPlanResponse& res,
-                                  std::vector<std::size_t>& added_path_index = empty_path_index_vector) const;
+                                  planning_interface::MotionPlanResponse& res) const;
 
   /** \brief Adapt the planning request if needed, call the planner
       function \e planner and update the planning response if
@@ -104,15 +95,11 @@ public:
    *  @param planning_scene Representation of the environment for the planning
    *  @param req Motion planning request with a set of constraints
    *  @param res Reference to which the generated motion plan response is written to
-   *  @param added_path_index Sometimes planning request adapters may add states on the solution path (e.g.,
-        add the current state of the robot as prefix, when the robot started to plan only from near that state, as the
-        current state itself appears to touch obstacles). This is helpful because the added states should not be
-   considered invalid in all situations.
    *  @return True if response got generated correctly */
-  [[nodiscard]] virtual bool
-  adaptAndPlan(const PlannerFn& planner, const planning_scene::PlanningSceneConstPtr& planning_scene,
-               const planning_interface::MotionPlanRequest& req, planning_interface::MotionPlanResponse& res,
-               std::vector<std::size_t>& added_path_index = empty_path_index_vector) const = 0;
+  [[nodiscard]] virtual bool adaptAndPlan(const PlannerFn& planner,
+                                          const planning_scene::PlanningSceneConstPtr& planning_scene,
+                                          const planning_interface::MotionPlanRequest& req,
+                                          planning_interface::MotionPlanResponse& res) const = 0;
 };
 
 /** \brief Apply a sequence of adapters to a motion plan */
@@ -128,16 +115,11 @@ public:
    *  @param planning_scene Representation of the environment for the planning
    *  @param req Motion planning request with a set of constraints
    *  @param res Reference to which the generated motion plan response is written to
-   *  @param added_path_index Sometimes planning request adapters may add states on the solution path (e.g.,
-          add the current state of the robot as prefix, when the robot started to plan only from near that state, as the
-          current state itself appears to touch obstacles). This is helpful because the added states should not be
-          considered invalid in all situations.
    *  @return True if response got generated correctly */
   [[nodiscard]] bool adaptAndPlan(const planning_interface::PlannerManagerPtr& planner,
                                   const planning_scene::PlanningSceneConstPtr& planning_scene,
                                   const planning_interface::MotionPlanRequest& req,
-                                  planning_interface::MotionPlanResponse& res,
-                                  std::vector<std::size_t>& added_path_index = empty_path_index_vector) const;
+                                  planning_interface::MotionPlanResponse& res) const;
 
 private:
   std::vector<PlanningRequestAdapterConstPtr> adapters_;
--- a/moveit_core/planning_request_adapter/src/planning_request_adapter.cpp
+++ b/moveit_core/planning_request_adapter/src/planning_request_adapter.cpp
@@ -64,12 +64,11 @@ bool callPlannerInterfaceSolve(const planning_interface::PlannerManager& planner
 
 bool callAdapter(const PlanningRequestAdapter& adapter, const PlanningRequestAdapter::PlannerFn& planner,
                  const planning_scene::PlanningSceneConstPtr& planning_scene,
-                 const planning_interface::MotionPlanRequest& req, planning_interface::MotionPlanResponse& res,
-                 std::vector<std::size_t>& added_path_index)
+                 const planning_interface::MotionPlanRequest& req, planning_interface::MotionPlanResponse& res)
 {
   try
   {
-    bool result = adapter.adaptAndPlan(planner, planning_scene, req, res, added_path_index);
+    bool result = adapter.adaptAndPlan(planner, planning_scene, req, res);
     RCLCPP_DEBUG_STREAM(LOGGER, adapter.getDescription() << ": " << moveit::core::error_code_to_string(res.error_code));
     return result;
   }
@@ -77,7 +76,6 @@ bool callAdapter(const PlanningRequestAdapter& adapter, const PlanningRequestAda
   {
     RCLCPP_ERROR(LOGGER, "Exception caught executing adapter '%s': %s\nSkipping adapter instead.",
                  adapter.getDescription().c_str(), ex.what());
-    added_path_index.clear();
     return planner(planning_scene, req, res);
   }
 }
@@ -87,15 +85,14 @@ bool callAdapter(const PlanningRequestAdapter& adapter, const PlanningRequestAda
 bool PlanningRequestAdapter::adaptAndPlan(const planning_interface::PlannerManagerPtr& planner,
                                           const planning_scene::PlanningSceneConstPtr& planning_scene,
                                           const planning_interface::MotionPlanRequest& req,
-                                          planning_interface::MotionPlanResponse& res,
-                                          std::vector<std::size_t>& added_path_index) const
+                                          planning_interface::MotionPlanResponse& res) const
 {
   return adaptAndPlan(
       [&planner](const planning_scene::PlanningSceneConstPtr& scene, const planning_interface::MotionPlanRequest& req,
                  planning_interface::MotionPlanResponse& res) {
         return callPlannerInterfaceSolve(*planner, scene, req, res);
       },
-      planning_scene, req, res, added_path_index);
+      planning_scene, req, res);
 }
 
 void PlanningRequestAdapterChain::addAdapter(const PlanningRequestAdapterConstPtr& adapter)
@@ -106,17 +103,13 @@ void PlanningRequestAdapterChain::addAdapter(const PlanningRequestAdapterConstPt
 bool PlanningRequestAdapterChain::adaptAndPlan(const planning_interface::PlannerManagerPtr& planner,
                                                const planning_scene::PlanningSceneConstPtr& planning_scene,
                                                const planning_interface::MotionPlanRequest& req,
-                                               planning_interface::MotionPlanResponse& res,
-                                               std::vector<std::size_t>& added_path_index) const
+                                               planning_interface::MotionPlanResponse& res) const
 {
   // if there are no adapters, run the planner directly
   if (adapters_.empty())
   {
-    added_path_index.clear();
     return callPlannerInterfaceSolve(*planner, planning_scene, req, res);
   }
-  // the index values added by each adapter
-  std::vector<std::vector<std::size_t>> added_path_index_each(adapters_.size());
 
   // if there are adapters, construct a function for each, in order,
   // so that in the end we have a nested sequence of functions that calls all adapters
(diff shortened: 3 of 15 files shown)


moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -466,15 +465,31 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
 
 const TwistCommand Servo::toPlanningFrame(const TwistCommand& command)
 {
+  // Reverse the twist vector since we need it to be (angular, linear) for transforming using tf2.
+  Eigen::VectorXd reordered_twist(6);
+  reordered_twist.tail<3>() = command.velocities.head<3>();
+  reordered_twist.head<3>() = command.velocities.tail<3>();
```
sea-bass, line 471: I wonder if it would work better if we did things like this: https://godbolt.org/z/oqs16ocfn

So basically:


```cpp
// Defined once
Eigen::VectorXf flip_indices{6};
flip_indices << 3, 4, 5, 0, 1, 2;

// In your loop
auto transformed_twist = command.velocities(flip_indices);
tf2::doTransform(transformed_twist, transformed_twist, body_to_space_tf);
return TwistCommand{ servo_params_.planning_frame, transformed_twist(flip_indices));
```

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
```diff
@@ -134,10 +134,8 @@ class Servo
 private:
   /**
    * \brief Convert a give twist command to planning frame
-   * This implementation does not differentiate between body twist and spatial twist,
-   * i.e if the command is as seen from a stationary or non-stationary frame.
-   * The result of this transformation will only be accurate if the command frame is a stationary frame.
-   * See issue: https://github.com/ros-planning/moveit2/issues/2150
+   * The reference frame specified by `frame_id` should be a stationary frame.
+   * The command maybe expressed in the space frame basis or the body frame basis.
```
sea-bass, line 138:
```suggestion
   * The command may be expressed in the space frame basis or the body frame basis.
```

I also think more explanation here would be required. What does space frame vs. body frame mean?

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -466,15 +465,31 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
...
+
+  if (!servo_params_.command_frame_is_stationary)
+  {
+    RCLCPP_DEBUG_STREAM(LOGGER, "Transforming from body frame to space frame");
```
sea-bass, line 475: Would it help to print the actual frame names here as well?

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -466,15 +465,31 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
...
+  }
+
   if (command.frame_id != servo_params_.planning_frame)
   {
-    RCLCPP_WARN_STREAM_THROTTLE(LOGGER, *node_->get_clock(), ROS_LOG_THROTTLE_PERIOD,
-                                "Twist command is not in planning frame, transformation may not be accurate.");
+    RCLCPP_DEBUG_STREAM(LOGGER, "Transforming from " << command.frame_id << " to " << servo_params_.planning_frame);
+    const auto planning_to_command_tf =
+        transform_buffer_.lookupTransform(servo_params_.planning_frame, command.frame_id, rclcpp::Time(0));
+    tf2::doTransform(reordered_twist, reordered_twist, planning_to_command_tf);
```
sea-bass, line 486: If both of these IF statements are hit, how would this be different from directly doing a single transform lookup from `servo_params_.ee_frame` to `servo_params_.planning_frame`, and should we explicitly capture this condition so we don't have to do redundant transformations + lookups on our end?

Is it true then that `if (!servo_params_.command_frame_is_stationary)`, then you're basically just always converting the command from `ee_frame` to `planning_frame` regardless or what its `command.frame_id` is? And is that correct given the spec?

sea-bass (commented):
I'll admit, I don't fully understand this logic and new parameter, so I've taken a look and commented more on the code side of things rather than the math.

Regardless, this will need some detailed documentation to make sense of it for users, because if I can't figure it out and I've been looking at servo for a few months, I'm not confident that the average user will be able to figure out easily either.

moveit_ros/moveit_servo/src/servo.cpp
ibrahiminfinite: I did not know this existed ! Super convenient. Thanks

moveit_ros/moveit_servo/src/servo.cpp
ibrahiminfinite: The first IF statement converts from `body` (ee frame in case of Servo) frame to `space` frame A (any stationary frame).
The second converts from the stationary frame A to planning frame.

There are 4 situations that can happen:

1. Command is with reference to planning frame and expressed in planning frame coordinates -----> None of the IF conditions are triggered
2. Command is with reference to planning frame and expressed in ee frame (body) coordinates -----> Only the first IF condition is triggered
3. Command is with reference to some stationary frame (A ) other than planning frame and expressed in the A coordinates ----->  Only the  second IF condition is triggered
4. Command is with reference to some stationary frame (A) other than planning frame and expressed in the ee frame (body) coordinates -----> both IF conditions are triggered

moveit_ros/moveit_servo/src/servo.cpp
ibrahiminfinite: Gave this some more thought.
For both the current method and the method you suggested we end up creating 2 vector variables.
In the absence of an advantage, the current method seems more readable to me, since it is clear that we are moving the "angular" and "linear" parts.

moveit_ros/moveit_servo/src/servo.cpp
ibrahiminfinite: Updated it.

2 & 4 are basically same,  twist expressed in body frame is independent of the space frame. so irrespective of the reference frame we can convert it to `planning_frame`.

henningkayser (commented):
I don't understand why we would need this option at all. The PlanningSceneMonitor listens to and keeps track of all transforms (has its own transform buffer), and should probably be used instead of lookupTransform(). It also shouldn't make a difference if a frame is stationary or not just in order to transform it to the planning frame.

ibrahiminfinite:
> I don't understand why we would need this option at all. The PlanningSceneMonitor listens to and keeps track of all transforms (has its own transform buffer), and should probably be used instead of lookupTransform(). It also shouldn't make a difference if a frame is stationary or not just in order to transform it to the planning frame.

There is no real difference in the computation for when the command is in moving frame.
However the `frame_id` we specify in the command is the reference frame, if I am not mistaken.

So when the twist command given is expressed in the moving frame basis , we cannot use the moving frame as the `frame_id` since velocity of the moving frame in the moving frame would be zero.

Does that make sense?

ibrahiminfinite:
Not sure why the "HaltForSingularity"  unit test fails on Humble.
The changes in this PR are not related that function.

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -466,15 +465,35 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
...
+                                    << command.frame_id << "to planning frame: " << servo_params_.planning_frame);
+    const auto command_to_planning_tf =
+        transform_buffer_.lookupTransform(servo_params_.planning_frame, command.frame_id, rclcpp::Time(0));
+    tf2::doTransform(reordered_twist, reordered_twist, command_to_planning_tf);
   }
-  const auto command_to_planning_frame =
-      transform_buffer_.lookupTransform(servo_params_.planning_frame, command.frame_id, rclcpp::Time(0));
-  Eigen::VectorXd transformed_twist = command.velocities;
-  tf2::doTransform(transformed_twist, transformed_twist, command_to_planning_frame);
+  else if (!command_in_space_frame)
```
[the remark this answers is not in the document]
ibrahiminfinite: That means command is in `planning_frame` and expressed in `planning_frame` basis right?
Which means we don't have to do the transformation.

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
```diff
@@ -134,10 +134,8 @@ class Servo
 private:
   /**
    * \brief Convert a give twist command to planning frame
-   * This implementation does not differentiate between body twist and spatial twist,
-   * i.e if the command is as seen from a stationary or non-stationary frame.
-   * The result of this transformation will only be accurate if the command frame is a stationary frame.
-   * See issue: https://github.com/ros-planning/moveit2/issues/2150
+   * The reference frame specified by `frame_id` should be a stationary frame.
+   * The command may be expressed in the space frame basis or the body frame basis.
```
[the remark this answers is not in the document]
ibrahiminfinite: >    * This implementation does not differentiate between body twist and spatial twist,
>    * i.e if the command is as seen from a stationary or non-stationary frame.

This part is no longer true since command can now be given as in body frame or space frame.

>    * The result of this transformation will only be accurate if the command frame is a stationary frame.

can keep this part.

moveit_ros/moveit_servo/config/panda_simulated_config.yaml
```diff
@@ -40,7 +40,9 @@ move_group_name:  panda_arm  # Often 'manipulator' or 'arm'
 planning_frame: panda_link0  # The MoveIt planning frame. Often 'base_link' or 'world'
 
 ## Other frames
-ee_frame: panda_link8  # The name of the end effector link, used to return the EE pose
+ee_frame: panda_hand  # The name of the end effector link, used to return the EE pose
+
+command_frame_is_stationary: true  # Twist commands are as seen from a stationary frame.
```
ibrahiminfinite, line 45:
```suggestion
# See https://github.com/ros-planning/moveit2/issues/2150
# Is the twist command expressed in a stationary frame ?
command_frame_is_stationary: true
```

How about like this ?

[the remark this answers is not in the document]
sea-bass: I'd omit the question mark. Maybe:

```suggestion
# If true, twist commands are given in a stationary frame; otherwise, they are in a moving frame.
# See https://github.com/ros-planning/moveit2/issues/2150 for more details
command_frame_is_stationary: true
````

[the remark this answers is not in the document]
Comment (AndyZe) [reply moveit_ros/moveit_servo/config/panda_simulated_config.yaml:45]: ^sounds good

Comment (AndyZe) [reply moveit_ros/moveit_servo/config/panda_simulated_config.yaml:45]: ok

Comment (AndyZe) [reply moveit_ros/moveit_servo/config/panda_simulated_config.yaml:45]: let's go with sea-bass's suggestion

File: moveit_ros/moveit_servo/include/moveit_servo/servo.hpp (hunk 6 above)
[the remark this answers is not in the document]
Comment (AndyZe) [reply moveit_ros/moveit_servo/include/moveit_servo/servo.hpp:138]: Well, don't these 2 sentences contradict each other?

>   * The reference frame specified by `frame_id` should be a stationary frame.
>   * The command may be expressed in the space frame basis or the body frame basis.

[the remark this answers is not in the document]
Comment (ibrahiminfinite) [reply moveit_ros/moveit_servo/include/moveit_servo/servo.hpp:138]: I agree this part might be a bit confusing.
The question is, in the context of twist command, does `frame_id` represent the reference frame or the basis in which the command is expressed ?
I was thinking of it as the reference frame

Eg: B moving w.r.t to A, --> command reference frame A, but expressed in B. 
Should `frame_id` be A or B?

File: moveit_ros/moveit_servo/config/test_config_panda.yaml [hunk 8]
```diff
@@ -41,7 +41,8 @@ move_group_name:  panda_arm  # Often 'manipulator' or 'arm'
 planning_frame: panda_link0  # The MoveIt planning frame. Often 'base_link' or 'world'
 
 ## Other frames
-ee_frame: panda_link8  # The name of the end effector link, used to return the EE pose
+ee_frame: panda_hand  # The name of the end effector link, used to return the EE pose
+command_frame_is_stationary: true # Twist commands are as seem from the stationary frame.
```
ibrahiminfinite, line 45:
```suggestion
# If true, twist commands are given in a stationary frame; otherwise, they are in a moving frame.
# See https://github.com/ros-planning/moveit2/issues/2150 for more details
command_frame_is_stationary: true
```

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -466,15 +465,35 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
...
+  {
+    RCLCPP_DEBUG_STREAM(LOGGER, "Transforming twist from moving (body) frame: "
+                                    << servo_params_.ee_frame << "to planning frame: " << servo_params_.planning_frame);
+    const auto body_to_space_tf =
+        transform_buffer_.lookupTransform(servo_params_.planning_frame, servo_params_.ee_frame, rclcpp::Time(0));
+    tf2::doTransform(reordered_twist, reordered_twist, body_to_space_tf);
+  }
+
```
ibrahiminfinite, line 492:
```suggestion

// else: the command is in the planning frame so no transformation is needed

```

AndyZe:
Let's have a video chat. I feel like there is an XY communication issue going on here.

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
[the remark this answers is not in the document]
AndyZe: I think if the user is commanding something in the end-effector frame, then `command.header.frame_id` should be `end_effector` (whatever the EE frame is named).

If they're commanding something in the planning frame then `command.header.frame_id` should be `base_link`.

I think it's safe to assume for Servo that the user always wants to move the end effector. They won't move any other links. If they wanted to move a different link, they could set up a new planning group to do that.

For example, if `command.header.frame_id` is "elbow" .... that means the basis for the command is the elbow frame but it is still the end effector that will move in that direction.

ibrahiminfinite:
Removed the complexity in favour of  documentation.

sea-bass (approved):

AndyZe:
Code looks good, i'll approve after testing it.

sea-bass (changes requested):
Following @AndyZe 's need to test, I decided to test this against MoveIt Studio and found a bug...

I set the twist commands to be in the end effector frame, and the translation doesn't quite line up with the frame's axes. More importantly, the twist commands along a single axis are leading to combined translation and rotation.

https://github.com/ros-planning/moveit2/assets/4603398/5ad6204d-63af-4bfa-8150-7f098741348f

Looking at the `tf2::doTransform` code here (https://docs.ros2.org/foxy/api/tf2_geometry_msgs/namespacetf2.html), I see that there isn't a specialization for the Eigen Vector6 class, so actually this is being all treated as a 3-element vector and it's likely why your tests required flipping around the first and last 3 elements -- because it only uses the first 3 elements!

sea-bass:
Behold, I found a PR by @AndyZe which says this, in fact, is translation first and rotation later!

https://github.com/ros2/geometry2/pull/406/files

AndyZe:
That helps explain my confusion. lol

sea-bass:
I guess the unit tests still pass because we're passing all the commands in the planning frame itself, so the transform is a no-op.

Would suggest making some tests that use alternative frames, e.g., the end effector, in the upcoming work we discussed offline

sea-bass:
BTW reverted to `main` and this issue goes away. The frame conversions work correctly. So I actually think we should just close this PR and move on to the next thing.

ibrahiminfinite:
LOL, This is getting quite confusing.

> Behold, I found a PR by @AndyZe which says this, in fact, is translation first and rotation later!
> 
> https://github.com/ros2/geometry2/pull/406/files

I believe the documentation there is wrong.

Screenshot from 2023-08-17 09-08-35



This is the equation being used in that `tf2::doTransform()` method, based on this ,the angular velocities must come first.
[Forces8.pdf](https://github.com/ros-planning/moveit2/files/12365324/Forces8.pdf) This is another source Andy has mentioned in the comment, it is the same here.


> I set the twist commands to be in the end effector frame, along a single axis, but they are leading to combined translation and rotation.

I think it might be because the `ee_frame` parameter was still pointing to `panda_link8` instead of `panda_hand`. The `panda_link8` frame has a rotation relative to `panda_hand`. This change was originally in the PR but got reverted when I reset.

Also turns out, the pose resulting from the application of the twist is computed incorrectly here: https://github.com/ros-planning/moveit2/blob/a2f0183297284ed43ef5d435b4e11b414ec21876/moveit_ros/moveit_servo/src/utils/command.cpp#L236

The `ik_solver->getTipFrame()` points to `panda_link8` while the the pose we are interested in is of `panda_hand`.
I think this needs to be replaced with tf2 transforms as well rather than what is available in `RobotState`.

sea-bass:
Hmm, now I too am confused.

According to the [code](https://github.com/ros2/geometry2/blob/rolling/tf2_eigen/include/tf2_eigen/tf2_eigen.hpp#L143-L174) (not the comments) @ibrahiminfinite is right in that the rotation block seems to come first.

Also, remember that in my example I am using a UR5e robot, not a Panda, so the actual frames used in the Panda servo config do not apply. But you're right -- this was a case where a moving frame was being used as the `frame_id`.

But then, in MoveIt Studio I get the correct results before this PR, so something else might be up. It might boil down to expectations:
* #2150 explicitly mentions that if there is a lever, applying a rotation twist will lead to translation as well. This, in fact, is what I'm seeing with this PR, so in that sense it might be correct
* However, in my case I actually want the twist to be applied in the end effector frame without translating the end effector. In order words, I want the rotation-only part, but not the translation, if I'm commanding a rotation-only twist.

How do we reconcile that some users (like myself) want the rotation-only part of the twist, whereas others (like those who reported #2150) want the actual translation+rotation portion of twisting the EE frame w.r.t. the base frame?

ibrahiminfinite:
> * However, in my case I actually want the twist to be applied in the end effector frame without translating the end effector. In order words, I want the rotation-only part, but not the translation, if I'm commanding a rotation-only twist.

https://github.com/ros-planning/moveit/issues/2891 This was already reported earlier.

AndyZe:
I think we might be getting back to the original idea of not making any changes in code. Provide the transformation options in a header file. Point to the header file from the tutorial and an advanced user can do as they please.

Issue #2066, opened 2023-03-31
Infinite loop in TOTG if zero max acceleration is specified


sea-bass:
### Description

If you pass in a zero max acceleration limit to TOTG, you will actually get stuck in an [infinite loop](https://github.com/ros-planning/moveit2/blob/main/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp#L520) because of the [`integrateForward()`](https://github.com/ros-planning/moveit2/blob/main/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp#L512) function.

While zero acceleration limit is physically meaningless since it means the joints won't be able to move, we should at least have an error/warning. Whether we actually throw an exception or just return a no-op trajectory is up to the implementer.

### Your environment
* ROS Distro: Rolling
* OS Version: Ubuntu 22.04

AndyZe:
Quick comment: you might as well check for 0 vel and 0 jerk limits, too

PR #2307, opened 2023-08-17
Exit loop when position can't change.

uavster:
### Description

Fixes https://github.com/ros-planning/moveit2/issues/2066. Infinite looping happens at [Trajectory::integrateForward](https://github.com/ros-planning/moveit2/blob/a2f0183297284ed43ef5d435b4e11b414ec21876/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp#L512) when three things concur:
* The velocity at a trajectory point is zero
* Any maximum acceleration component is zero
* The corresponding component in the tangent at the point of the trajectory [is not zero](https://github.com/ros-planning/moveit2/blob/a2f0183297284ed43ef5d435b4e11b414ec21876/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp#L693).

The changes in this PR are:
* Break the loop when velocity and acceleration are zero
* Unit test that times out without the fix and passes with it
* Unit test that ensures the loop is not broken if the tangent is zero even if the maximum acceleration is

### Checklist


moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -566,6 +566,16 @@ bool Trajectory::integrateForward(std::list<TrajectoryStep>& trajectory, double
     trajectory.push_back(TrajectoryStep(path_pos, path_vel));
     acceleration = getMinMaxPathAcceleration(path_pos, path_vel, true);
 
+    if (path_vel == 0 && acceleration == 0)
```
sea-bass, line 569: I'm generally wary of doing direct floating-point equality, but I think in this case it's probably okay since it will come directly from a parameter and isn't calculated on the fly.

moveit_core/trajectory_processing/test/test_time_optimal_trajectory_generation.cpp
```diff
@@ -538,6 +538,29 @@ TEST(time_optimal_trajectory_generation, testSingleDofDiscontinuity)
...
+
+TEST(time_optimal_trajectory_generation, testIrrelevantZeroMaxAccelerationsDontInvalidateTrajectory)
+{
+  EXPECT_TRUE(Trajectory(Path(std::list<Eigen::VectorXd>{ Eigen::Vector2d(0, 0), Eigen::Vector2d(0, 1) }),
+                         Eigen::Vector2d(1, 1), Eigen::Vector2d(0, 1))
+                  .isValid());
+  EXPECT_TRUE(Trajectory(Path(std::list<Eigen::VectorXd>{ Eigen::Vector2d(0, 0), Eigen::Vector2d(1, 0) }),
+                         Eigen::Vector2d(1, 1), Eigen::Vector2d(1, 0))
+                  .isValid());
+}
```
sea-bass, line 562: Great that you added unit tests -- for my reference, where in this test are we specifying that max acceleration / max vel is zero?

sea-bass (commented):
Also, what happens in the case where `time_step_` is zero? Should that be handled here as well?

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
uavster: I considered comparing with epsilon, but I'm not sure if I'll be preventing trajectories that would still end but simply take longer. You're correct that, in the case that a component of max_accel==0, acceleration comes from a min() operation, so acceleration is guaranteed to be 0.

moveit_core/trajectory_processing/test/test_time_optimal_trajectory_generation.cpp
uavster: Third parameter of Trajectory() is the maximum acceleration (see that tests set different components of the acceleration to 0, or all of them). Zero max velocities did not make the loop to spin indefinitely.

uavster: I'm going to add parameter annotations to the calls in the tests to make it clearer.

uavster:
> 

Ah right! My original fix does not catch that because time_step==0 makes velocity, but not acceleration. I've added an extra case at the constructor to check for that.

moveit_core/trajectory_processing/test/test_time_optimal_trajectory_generation.cpp
uavster: Added parameter annotations.

moveit_core/trajectory_processing/test/test_time_optimal_trajectory_generation.cpp
```diff
@@ -538,6 +538,37 @@ TEST(time_optimal_trajectory_generation, testSingleDofDiscontinuity)
...
+                         /*max_velocity=*/Eigen::Vector2d(1, 1), /*max_acceleration=*/Eigen::Vector2d(1, 0))
+                  .isValid());
+}
+
+TEST(time_optimal_trajectory_generation, testTimeStepZeroMakesTrajectoryInvalid)
+{
+  EXPECT_FALSE(Trajectory(Path(std::list<Eigen::VectorXd>{ Eigen::Vector2d(0, 0), Eigen::Vector2d(1, 1) }),
+                          /*max_velocity=*/Eigen::Vector2d(1, 1), /*max_acceleration=*/Eigen::Vector2d(1, 1),
+                          /*time_step=*/0)
+                   .isValid());
```
sea-bass, line 569: Oh wow, I had originally read this all as a 4-point trajectory. This explains a lot.

I might have a slight preference to creating variables up front, so e.g.

```
const auto path = Path(std::list<Eigen::VectorXd>{ Eigen::Vector2d(0, 0), Eigen::Vector2d(1, 1) });
const auto max_vel = Eigen::Vector2d(1, 1);
const auto max_acc = Eigen::Vector2d(1, 1);
const double time_step = 0;

EXPECT_FALSE(Trajectory(path, max_vel, max_acc, time_step).is_valid());
```

... but up to you.
```

sea-bass (approved):

moveit_core/trajectory_processing/test/test_time_optimal_trajectory_generation.cpp
uavster: It's muscle memory. One-time-use variables were strongly discouraged in C++ at Google. You're right that, at least, max velocities could benefit from a constant. Will change.

sea-bass: Even for unit tests?

I tend to be way more liberal with local variables in tests for readability, but understand that for source.

uavster: No difference was made (funny though that they were ok in python). Constants were fine, e.g. kMaxVel. I'm fine with both styles as long as readability improves. Turned the paths into a constant as well. Hopefully, it reads better now.

sea-bass (changes requested):
gah. The clang-tidy checks are failing because somehow MoveIt doesn't use the Google style of constant casing. Sorry :/

https://github.com/ros-planning/moveit2/actions/runs/5893944176/job/15986542597?pr=2307#step:12:9198

uavster:
> gah. The clang-tidy checks are failing because somehow MoveIt doesn't use the Google style of constant casing. Sorry :/
> 
> https://github.com/ros-planning/moveit2/actions/runs/5893944176/job/15986542597?pr=2307#step:12:9198

Oh, hahah. Linter-settled discussion.

sea-bass:
> Oh, hahah. Linter-settled discussion.

It's specifically that in the bodies of functions, where things aren't constants, you have to use `snake_case`, whereas `camelCase` is okay for e.g. constexprs in anonymous namespaces.

So if you just use names like `max_velocity` instead of `kMaxVelocity` in the tests, should be good to go

uavster:
> > Oh, hahah. Linter-settled discussion.
> 
> It's specifically that in the bodies of functions, where things aren't constants, you have to use `snake_case`, whereas `camelCase` is okay for e.g. constexprs in anonymous namespaces.
> 
> So if you just use names like `max_velocity` instead of `kMaxVelocity` in the tests, should be good to go

done

uavster:
> > > Oh, hahah. Linter-settled discussion.
> > 
> > 
> > It's specifically that in the bodies of functions, where things aren't constants, you have to use `snake_case`, whereas `camelCase` is okay for e.g. constexprs in anonymous namespaces.
> > So if you just use names like `max_velocity` instead of `kMaxVelocity` in the tests, should be good to go
> 
> done

BTW, are you ok with inlining the instantiations of non-constants (Path and Trajectory)?

sea-bass:
> BTW, are you ok with inlining the instantiations of Path and Trajectory?

That's fine!

sea-bass (approved):

PR #2311, opened 2023-08-21
[Servo] Fix Twist transformation 

### LINKED ISSUE #2150: Wrong equations used in Servo twist transformations

ibrahiminfinite:
### Description

Fixes #2150 
Previous discussions on https://github.com/ros-planning/moveit2/pull/2301

The problem had boiled down to:
> How do we reconcile that some users (like myself) want the rotation-only part of the twist, whereas others (like those who reported https://github.com/ros-planning/moveit2/issues/2150) want the actual translation+rotation portion of twisting the EE frame w.r.t. the base frame?

This PR addresses  the  above by adding the option for proper twist conversion while preserving the existing method of twist conversion (which is wrong).
The selection is made using the parameter  `apply_twist_commands_about_ee_frame`.

@sea-bass 
@AndyZe

### DIFF
--- a/moveit_ros/moveit_servo/config/servo_parameters.yaml
+++ b/moveit_ros/moveit_servo/config/servo_parameters.yaml
@@ -16,7 +16,8 @@ servo:
 
   ee_frame: {
     type: string,
-    description: "The name of end effector frame of your robot \
+    description: "The name of end effector frame of your robot, \
+                  this should also be the IK tip frame of your IK solver. \
                   This parameter does not have a default value and \
                   must be passed to the node during launch time."
   }
@@ -124,6 +125,15 @@ servo:
       default_value: 0.5,
       description: "Max joint angular/linear velocity. Only used for joint commands on joint_command_in_topic."
     }
+
+
+  angular_velocity_about_ee_frame: {
+    type: bool,
+    default_value: true,
+    description: "If true, the angular velocity specified in the twist command is applied about the ee frame axes \
+                  if false, the twist computed will include the linear component from rotation of ee frame about the planning frame, \
+                  due to the existence of a lever between the two frames."
+  }
 
   override_velocity_scaling_factor: {
     type: double,
--- a/moveit_ros/moveit_servo/config/servo_parameters.yaml
+++ b/moveit_ros/moveit_servo/config/servo_parameters.yaml
@@ -127,7 +127,7 @@ servo:
     }
 
 
-  angular_velocity_about_ee_frame: {
+  apply_twist_commands_about_ee_frame: {
     type: bool,
     default_value: true,
     description: "If true, the angular velocity specified in the twist command is applied about the ee frame axes \
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
@@ -134,10 +134,7 @@ public:
 private:
   /**
    * \brief Convert a give twist command to planning frame
-   * This implementation does not differentiate between body twist and spatial twist,
-   * i.e if the command is as seen from a stationary or non-stationary frame.
-   * The result of this transformation will only be accurate if the command frame is a stationary frame.
-   * See issue: https://github.com/ros-planning/moveit2/issues/2150
+   * The result of this transformation will be accurate only if the command frame is a stationary frame or end-effector frame.
    * @param command The twist command to be converted
    * @return The transformed twist command.
    */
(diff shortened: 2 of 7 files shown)


moveit_ros/moveit_servo/config/servo_parameters.yaml
```diff
@@ -125,6 +126,15 @@ servo:
       description: "Max joint angular/linear velocity. Only used for joint commands on joint_command_in_topic."
     }
 
+
+  angular_velocity_about_ee_frame: {
```
sea-bass, line 130: This param name doesn't read like it's a boolean, but rather as the angular velocity itself.

Maybe `apply_twist_commands_about_ee_frame` or something?

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
```diff
@@ -134,10 +134,7 @@ class Servo
 private:
   /**
    * \brief Convert a give twist command to planning frame
-   * This implementation does not differentiate between body twist and spatial twist,
-   * i.e if the command is as seen from a stationary or non-stationary frame.
-   * The result of this transformation will only be accurate if the command frame is a stationary frame.
-   * See issue: https://github.com/ros-planning/moveit2/issues/2150
+   * The result of this transformation will be accurate only if the command frame is a stationary frame or end-effector frame.
```
sea-bass, line 137: Should we be linking here to some of the resources in @AndyZe's Eigen conversion function in case people want to read more about that?

Also, I'm not sure that "not accurate" is quite the right thing here anymore. It really is whether you want the twist to be applied about the planning frame vs. about the end effector frame, which is what caused all our confusion last week in the first place. So probably worth rewording that here as well?

sea-bass (commented):
Looks reasonable.

Got an unused variable failing CI [here](https://github.com/ros-planning/moveit2/actions/runs/5929193908/job/16076360943?pr=2311#step:12:332), but other small comments on naming and ... comments.

moveit_ros/moveit_servo/config/servo_parameters.yaml
ibrahiminfinite: 👍

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
ibrahiminfinite: Yeah that is a good idea.

moveit_ros/moveit_servo/tests/test_integration.cpp
```diff
@@ -92,6 +92,30 @@ TEST_F(ServoCppFixture, TwistTest)
...
+
+  moveit_servo::KinematicState next_state = servo_test_instance_->getNextJointState(twist_non_zero);
+  status_next = servo_test_instance_->getStatus();
+  ASSERT_EQ(status_next, moveit_servo::StatusCode::NO_WARNING);
+
+  // Check against manually verified value
+  constexpr double expected_delta = 0.001693;
+  double delta = next_state.positions[6] - curr_state.positions[6];
+  constexpr double tol = 0.00001;
+  ASSERT_NEAR(delta, expected_delta, tol);
```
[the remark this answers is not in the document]
ibrahiminfinite: Not sure that will be a problem.
I had added an option disable the use of smoothing plugins during the refactor for making testing easier.
So the tests run without using the smoothing plugins, and the command values are well below any type of scaling limit.

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
ibrahiminfinite: > It really is whether you want the twist to be applied about the planning frame vs. about the end effector frame,

The "not accurate" part is about the command being in some moving frame other than ee frame. (eg: the case of elbow frame)

ibrahiminfinite:
> I have to say that the translation is in the opposite direction of what I expected!?

The transformation pulled from TF2 was the wrong one ,`T_command_planning` instead of  `T_planning_command`.

Also restored  `servo_keyboard_input` demo with the option to switch between the `planning` and `end effector` frames.

PR #2312, opened 2023-08-22
Remove humble tags from main branch GHA jobs

henningkayser:
This removes redundant humble tags for the docker images and the pre-release test. Both should be run from the humble branch only.

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ROS_DISTRO: [humble, rolling]
+        ROS_DISTRO: [rolling]
     runs-on: ubuntu-latest
     permissions:
       packages: write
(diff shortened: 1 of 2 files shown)


JafarAbdi:
We do need these containers for MTC's ros2 branch. Maybe we name them humble-main or something like that?

henningkayser:
> We do need these containers for MTC's ros2 branch. Maybe we name them humble-main or something like that?

For what do we need them? If ros2 should be humble-compatible, we should run tests using the humble images from the humble branch. We already have the MTC humble branch as well.

JafarAbdi:
> For what do we need them? If ros2 should be humble-compatible, we should run tests using the humble images from the humble branch. We already have the MTC humble branch as well.

ros2 is compatible with rolling/humble with moveit2's main branch. I don't use MTC personally, but I think it's crucial to make sure that it always compile for both distros with moveit2's main branch

> Also, it looks like the moveit2/humble branch is pushing conflicting docker images already.

Yes, I noticed the same thing a few days ago, maybe we have a different tag name for the main branch?

tylerjw (approved):

Issue #2272, opened 2023-07-27
Add utility function to check trajectory continuity


sjahr:
To verify the result of a motion planner/generator it might be necessary to check a trajectory for joint jumps. It would be nice to have this utility available as a free function. There is already [an implementation available in the Cartesian interpolator](https://github.com/ros-planning/moveit2/blob/main/moveit_core/robot_state/src/cartesian_interpolator.cpp#L222) that needs to be refactored into a free function. For more context take a look at [this (unmerged PR)](https://github.com/ros-planning/moveit2/pull/2253) with a less mature implementation.

abhijelly:
hi @sjahr, can I work on this?

EzraBrooks:
not stale.

@abhijelly thanks for the offer (and sorry that nobody followed up with you here)! PickNik has a commitment to deliver this particular feature to a client soon so we'll be handling the implementation to ensure it meets those requirements.

PR #2329, opened 2023-08-29
Refactor cartesian interpolator into free function

abhijelly:
### Description

Refactored existing implementation of Cartesian Interpolator as a free function. After not any answer back in the pr #2272 about where to move the function, I just moved it outside the class scope but still within the `cartesian_interpolator.h/cartesian_interpolator.cpp` files.

Please let me know if I should move it somewhere else where it would make more sense!

### Checklist

### DIFF
--- a/moveit_core/robot_state/src/cartesian_interpolator.cpp
+++ b/moveit_core/robot_state/src/cartesian_interpolator.cpp
@@ -54,6 +54,62 @@ static const std::size_t MIN_STEPS_FOR_JUMP_THRESH = 10;
 
 static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_robot_state.cartesian_interpolator");
 
+CartesianInterpolator::Percentage checkJointSpaceJump(const JointModelGroup* group,
+                                                                             std::vector<RobotStatePtr>& traj,
+                                                                             const JumpThreshold& jump_threshold)
+{
+  double percentage_solved = 1.0;
+  if (traj.size() <= 1)
+    return percentage_solved;
+
+  if (jump_threshold.factor > 0.0)
+    percentage_solved *= checkRelativeJointSpaceJump(group, traj, jump_threshold.factor);
+
+  if (jump_threshold.revolute > 0.0 || jump_threshold.prismatic > 0.0)
+    percentage_solved *= checkAbsoluteJointSpaceJump(group, traj, jump_threshold.revolute, jump_threshold.prismatic);
+
+  return CartesianInterpolator::Percentage(percentage_solved);
+}
+
+CartesianInterpolator::Percentage CartesianInterpolator::checkRelativeJointSpaceJump(const JointModelGroup* group,
+                                                                                     std::vector<RobotStatePtr>& traj,
+                                                                                     double jump_threshold_factor)
+{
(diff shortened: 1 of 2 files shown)


ibrahiminfinite (commented):
@abhijelly 

Just moving the function outside the class in the file does not make it a free function. 
You have to think of how to pull out the logic from the required `member function` and any other supporting logic into a `free function`. 

For example, the member variables that might be accessed by the function when it is a `member function` will need to be explicitly passed in when it is a `free function`.

This `free function` can then go into an anonymous namespace in the same file allowing it to be used by other functions. 

And this must be done without breaking the functionality of the existing code.
As you can see, the tests are failing.

abhijelly:
> @abhijelly
> 
> Just moving the function outside the class in the file does not make it a free function. You have to think of how to pull out the logic from the required `member function` and any other supporting logic into a `free function`.
> 
> For example, the member variables that might be accessed by the function when it is a `member function` will need to be explicitly passed in when it is a `free function`.
> 
> This `free function` can then go into an anonymous namespace in the same file allowing it to be used by other functions.
> 
> And this must be done without breaking the functionality of the existing code. As you can see, the tests are failing.

thank you for the pointers! Gives something to think about

abhijelly:
hey @ibrahiminfinite , I was looking into using anonymous namespace for `checkJointSpaceJump` function which makes the implementation of the function usable within the source file. However, anonymous namespaces should not be defined in the header file, so I'm kind of blocked on how to make this function accessible so that it can be used for testing or in test files and such that the functionality does not break. Do you have any thoughts on this?

ibrahiminfinite:
I guess you will need to put it in the `moveit::core` namespace itself within `robot_state.h` and `robot_state.cpp`  for it to be available for testing or add a new header and source file with the same namespace for the free functions. 
@sjahr What do you think ?

ibrahiminfinite:
@abhijelly 
You can go ahead with the the first option I mentioned above for now and get the functionality working.
Moving it to another file would be a trivial change and can be done later.

abhijelly:
> @abhijelly You can go ahead with the the first option I mentioned above for now and get the functionality working. Moving it to another file would be a trivial change and can be done later.

Understood, thank you for the insights! I try this by EOD

moveit_core/robot_state/src/cartesian_interpolator.cpp
```diff
@@ -54,6 +54,62 @@ static const std::size_t MIN_STEPS_FOR_JUMP_THRESH = 10;
...
+CartesianInterpolator::Percentage checkJointSpaceJump(const JointModelGroup* group,
+                                                                             std::vector<RobotStatePtr>& traj,
+                                                                             const JumpThreshold& jump_threshold)
+{
+  double percentage_solved = 1.0;
+  if (traj.size() <= 1)
+    return percentage_solved;
+
+  if (jump_threshold.factor > 0.0)
+    percentage_solved *= checkRelativeJointSpaceJump(group, traj, jump_threshold.factor);
```
Shobuj-Paul, line 66: Same here for checkRelativeJointSpaceJump function.

sjahr:
Closing this due to inactivity

Issue #2332, opened 2023-08-29
[Servo] Smoothing plugin does not update with planning group.


ibrahiminfinite:
#### Current behaviour 
The new Servo implementation supports changing the planning group, however the smoothing plugin implementation is not yet updated to support this.
The plugins initialize with a `num_joints` ([ref](https://github.com/ros-planning/moveit2/blob/47d92ef973680aea1badd6cc6080598f040a89f4/moveit_core/online_signal_smoothing/include/moveit/online_signal_smoothing/smoothing_base_class.h#L69)) parameter, but this this can now change after the initialization of the plugin, depending on the planning group.
 

#### Expected behaviour

Smoothing plugins should have the ability to re-initialize the filter size when the planning group changes.

### Your environment
* ROS Distro: [Rolling]
* OS Version: Ubuntu 22.04
* Source build
* Branch: main

**NOTE : This issue is currently being worked on.**

ibrahiminfinite:
**Approach 1:**  Add a check for planning group change [here](https://github.com/ros-planning/moveit2/blob/47d92ef973680aea1badd6cc6080598f040a89f4/moveit_ros/moveit_servo/src/servo.cpp#L213) and re-initialize the plugin.

Currently the planning group is set using a parameter, and this parameter is used to fetch the planning group on every iteration of Servo. This ensures that the planning group can be changed easily using the intended functionality of the ROS parameter.
But since we want to re-initialize the smoothing plugin when planning group changes,  the loop time would take a significant hit for the single iteration when the re-initialization happens.  This is the easiest way to do it if the one time hit to loop time is not a big problem.

**Approach 2:** Add a separate service to update the planning group

This requires adding some easy but considerable amount of changes to Servo.
The plugin would be re-initialized from the service and rest of the code can remain the same.
However functionality of the ROS parameter becomes redundant here, this parameter will also need to be updated from the service.

ibrahiminfinite:
@AndyZe ,@sea-bass, @henningkayser  Which of the approaches would be preferable ?

AndyZe:
I would make the planning group param static. It doesn't seem like a good idea to update which joints are being controlled in a realtime control scenario, anyway.

sea-bass:
We already have spoken to users who want to switch the planning group (arm + linear rail vs. arm only), so I do not think we should have a static planning group.

[...]

PR #2333, opened 2023-08-29
[Servo] Improve support for switching planning group

ibrahiminfinite:
### Description

Fixes https://github.com/ros-planning/moveit2/issues/2332

This PR adds improved support for changing the planning group used by Servo by following  [approach 3](https://github.com/ros-planning/moveit2/issues/2332#issuecomment-1702845344) mentioned in the issue.

### DIFF
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
@@ -97,6 +97,12 @@ public:
   StatusCode getStatus() const;
 
   /**
+   * \brief Set the current status of servo to the provided status.
+   * @param status The new status of servo.
+   */
+  void setStatus(const StatusCode status);
+
+  /**
    * \brief Get the message associated with the current servo status.
    * @return The status message.
    */
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
@@ -128,14 +128,14 @@ public:
    * \brief Get the current state of the robot as given by planning scene monitor.
    * @return The current state of the robot.
    */
-  KinematicState getCurrentRobotState() const;
+  KinematicState getCurrentRobotState();
 
   /**
    * \brief Smoothly halt at a commanded state when command goes stale.
    * @param The last commanded joint states.
    * @return The next state stepping towards the required halting state.
    */
-  std::pair<bool, KinematicState> smoothHalt(const KinematicState& halt_state) const;
+  std::pair<bool, KinematicState> smoothHalt(const KinematicState& halt_state);
 
   /**
    * \brief Updates the servo parameters and performs validations.
@@ -150,7 +150,7 @@ private:
    * @param command_frame The command frame name.
    * @return The transformation between planning frame and command frame.
    */
-  Eigen::Isometry3d getPlanningToCommandFrameTransform(const std::string& command_frame) const;
+  Eigen::Isometry3d getPlanningToCommandFrameTransform(const std::string& command_frame);
 
   /**
    * \brief Convert a given twist command to planning frame,
@@ -161,21 +161,21 @@ private:
    * @param command The twist command to be converted.
    * @return The transformed twist command.
    */
-  TwistCommand toPlanningFrame(const TwistCommand& command) const;
+  TwistCommand toPlanningFrame(const TwistCommand& command);
 
   /**
    * \brief Convert a given pose command to planning frame
    * @param command The pose command to be converted
    * @return The transformed pose command
    */
-  PoseCommand toPlanningFrame(const PoseCommand& command) const;
+  PoseCommand toPlanningFrame(const PoseCommand& command);
 
   /**
    * \brief Compute the change in joint position required to follow the received command.
    * @param command The incoming servo command.
    * @return The joint position change required (delta).
    */
-  Eigen::VectorXd jointDeltaFromCommand(const ServoInput& command, const moveit::core::RobotStatePtr& robot_state);
+  Eigen::VectorXd jointDeltaFromCommand(const ServoInput& command);
 
   /**
    * \brief Updates data depending on joint model group
@@ -220,6 +220,8 @@ private:
   std::unique_ptr<CollisionMonitor> collision_monitor_;
 
   pluginlib::UniquePtr<online_signal_smoothing::SmoothingBaseClass> smoother_ = nullptr;
+
+  moveit::core::RobotStatePtr robot_state_;
 };
 
 }  // namespace moveit_servo
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
@@ -223,6 +223,7 @@ private:
 
   moveit::core::RobotStatePtr robot_state_;
   const moveit::core::JointModelGroup* joint_model_group_;
+  std::vector<std::string> joint_names_;
 };
 
 }  // namespace moveit_servo
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
@@ -222,6 +222,7 @@ private:
   pluginlib::UniquePtr<online_signal_smoothing::SmoothingBaseClass> smoother_ = nullptr;
 
   moveit::core::RobotStatePtr robot_state_;
+  const moveit::core::JointModelGroup* joint_model_group_;
 };
 
 }  // namespace moveit_servo
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
@@ -137,6 +137,11 @@ public:
    */
   std::pair<bool, KinematicState> smoothHalt(const KinematicState& halt_state) const;
 
+  /**
+   * \brief Updates the servo parameters and performs validations.
+   */
+  bool updateParams();
+
 private:
   /**
    * \brief Finds the transform from the planning frame to a specified command frame.
@@ -183,11 +188,6 @@ private:
    * @return True if parameters are valid, else False
    */
   bool validateParams(const servo::Params& servo_params) const;
-
-  /**
-   * \brief Updates the servo parameters and performs validations.
-   */
-  bool updateParams();
 
   /**
    * \brief Create and initialize the smoothing plugin to be used by servo.
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
@@ -224,6 +224,8 @@ private:
   moveit::core::RobotStatePtr robot_state_;
   const moveit::core::JointModelGroup* joint_model_group_;
   std::vector<std::string> joint_names_;
+
+  bool invalid_parameter_update_;
 };
 
 }  // namespace moveit_servo
--- a/moveit_ros/moveit_servo/src/servo.cpp
+++ b/moveit_ros/moveit_servo/src/servo.cpp
@@ -91,12 +91,12 @@ Servo::Servo(const rclcpp::Node::SharedPtr& node, std::shared_ptr<const servo::P
   // Check if the transforms to planning frame and end effector frame exist.
   if (!robot_state->knowsFrameTransform(servo_params_.planning_frame))
   {
-    servo_status_ = StatusCode::INVALID;
+    setStatus(StatusCode::INVALID);
     RCLCPP_ERROR_STREAM(LOGGER, "No transform available for planning frame " << servo_params_.planning_frame);
   }
   else if (!robot_state->knowsFrameTransform(servo_params_.ee_frame))
   {
-    servo_status_ = StatusCode::INVALID;
+    setStatus(StatusCode::INVALID);
     RCLCPP_ERROR_STREAM(LOGGER, "No transform available for end effector frame " << servo_params_.ee_frame);
   }
   else
@@ -116,7 +116,7 @@ Servo::Servo(const rclcpp::Node::SharedPtr& node, std::shared_ptr<const servo::P
         std::make_unique<CollisionMonitor>(planning_scene_monitor_, servo_params_, std::ref(collision_velocity_scale_));
     collision_monitor_->start();
 
-    servo_status_ = StatusCode::NO_WARNING;
+    setStatus(StatusCode::NO_WARNING);
     RCLCPP_INFO_STREAM(LOGGER, "Servo initialized successfully");
   }
 }
@@ -251,6 +251,11 @@ StatusCode Servo::getStatus() const
   return servo_status_;
 }
 
+void Servo::setStatus(const StatusCode status)
+{
+  servo_status_ = status;
+}
+
 std::string Servo::getStatusMessage() const
 {
   return SERVO_STATUS_CODE_MAP.at(servo_status_);
@@ -323,7 +328,7 @@ Eigen::VectorXd Servo::jointDeltaFromCommand(const ServoInput& command, const mo
     if (expected_type == CommandType::JOINT_JOG)
     {
       delta_result = jointDeltaFromJointJog(std::get<JointJogCommand>(command), robot_state, servo_params_);
-      servo_status_ = delta_result.first;
+      setStatus(delta_result.first);
     }
     else if (expected_type == CommandType::TWIST)
     {
@@ -331,11 +336,11 @@ Eigen::VectorXd Servo::jointDeltaFromCommand(const ServoInput& command, const mo
       {
         const TwistCommand command_in_planning_frame = toPlanningFrame(std::get<TwistCommand>(command));
         delta_result = jointDeltaFromTwist(command_in_planning_frame, robot_state, servo_params_);
-        servo_status_ = delta_result.first;
+        setStatus(delta_result.first);
       }
       catch (tf2::TransformException& ex)
       {
-        servo_status_ = StatusCode::INVALID;
+        setStatus(StatusCode::INVALID);
         RCLCPP_ERROR_STREAM(LOGGER, "Could not transform twist to planning frame.");
       }
     }
@@ -345,11 +350,11 @@ Eigen::VectorXd Servo::jointDeltaFromCommand(const ServoInput& command, const mo
       {
         const PoseCommand command_in_planning_frame = toPlanningFrame(std::get<PoseCommand>(command));
         delta_result = jointDeltaFromPose(command_in_planning_frame, robot_state, servo_params_);
-        servo_status_ = delta_result.first;
+        setStatus(delta_result.first);
       }
       catch (tf2::TransformException& ex)
       {
-        servo_status_ = StatusCode::INVALID;
+        setStatus(StatusCode::INVALID);
         RCLCPP_ERROR_STREAM(LOGGER, "Could not transform pose to planning frame.");
       }
     }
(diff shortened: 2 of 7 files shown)


AndyZe (commented):
I'm not sure about this. Lot of complexity added and I don't see anything to prevent a switch when the robot is in motion.

Perhaps it would make more sense to launch 2 Servo instances, one for each `move_group`? That will be needed, anyway, for something like a dual-arm system.

ibrahiminfinite:
> Lot of complexity added

I don't think this adds any complexity.
The only new infrastructure added is the code for switching the move group, and a new service called `ServoMoveGroup` to `moveit_msgs`.

Other than that the changes are:

- The `num_joints` and `joint_model_group`  have been made into member variables and is updated only through the `updateMoveGroup` method. This required the `joint_model_group` to be passed in to a couple of methods.
- The parameter `move_group_name` is made read-only, it is only used once during initialization of `Servo` and is copied into the member variable `active_move_group_`, which is then used everywhere else where the move group name is needed, this is also the variable updated by the service.

> I don't see anything to prevent a switch when the robot is in motion.

The user would have to manage that when using the C++ API like with most of the functionality , but for the ROS API which I assume most users will be using, the service call automatically pauses the servo loop before updating the move group and then resumes it after update.

I think this functionality would be useful in the long run since it lets us use different groups without having to start a new node.
But you might be right @AndyZe,  if the move group changes, the parameters like `ee_frame` also potentially need to be changed. In which case it makes more sense to just launch a new servo node with the required parameters (but at the cost of resource overhead).

Would be good to have some feedback from the MoveIt Studio team.

AndyZe (changes requested):
Blocking until the comments above get resolved

ibrahiminfinite:
This update allows the user to change the planning group and smoothing plugin without restarting the servo node.
A safety check is also in place that will not allow for update of the above parameters unless servo is paused.  

Another important change is that the choice of updating the parameters when using C++ API has been delegated to the user.
This has two benefits:
- Allows for a clean usage of the `Servo::updateParams()` method from `ServoNode`
- Users of C++ API can gain better performance by skipping the call to `Servo::updateParams()` if they know that paramters will not be updated during runtime.

ibrahiminfinite:
@sea-bass

sea-bass (commented):
I kind of agree with @AndyZe about the extra complexity, but at the same time don't think we should require multiple servo nodes to switch planning groups.

To reduce complexity, I'd like to clarify: Why do we need to explicitly call `updateParameters()` with these changes? If this continues to be called automatically when you get the next servo command, this should be fine and you'll still get the warnings if you change planning groups/smoothing plugin while servo is running. Then, the changes in this PR also become much less disruptive.

As a much smaller comment: Seems with these changes we got rid of the `RCLCPP_INFO_STREAM()` printouts that mentioned what parameters were changed to. Can we get those back as well?

sea-bass (commented):
Looking at this again, I'm OK with the non-node examples explicitly calling `updateParams()`.

There is a test failure [here](https://github.com/ros-planning/moveit2/actions/runs/6071638770/job/16470067422?pr=2333#step:12:9123) based on inability to load the smoothing plugin, so take a look at that and this should be good.

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -89,69 +90,46 @@ Servo::Servo(const rclcpp::Node::SharedPtr& node, std::shared_ptr<const servo::P
...
 
 Servo::~Servo()
 {
   setCollisionChecking(false);
 }
 
-void Servo::setSmoothingPlugin()
+void Servo::setSmoothingPlugin(const servo::Params& servo_params)
 {
   // Load the smoothing plugin
```
AndyZe, line 112: redundant

```suggestion
```

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -212,35 +204,40 @@ bool Servo::validateParams(const servo::Params& servo_params)
 
 bool Servo::updateParams()
 {
```
AndyZe, line 206: You ought to check `getStatus() != StatusCode::PAUSED` right away and return immediately if it is. Reason being, `validateParams()` is expensive and you wouldn't want somebody calling it from within a control loop.

Then you don't need to check the status code at L219,L225.

```suggestion
{
  if (getStatus() != StatusCode::PAUSED)
  {
    RCLCPP_WARN_STREAM_THROTTLE(LOGGER, *node_->get_clock(), ROS_LOG_THROTTLE_PERIOD,
                                  "Updating parameters without pausing Servo is not allowed");
  }
```

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -374,10 +376,7 @@ Eigen::VectorXd Servo::jointDeltaFromCommand(const ServoInput& command, moveit::
 KinematicState Servo::getNextJointState(const ServoInput& command)
 {
   // Set status to clear
-  servo_status_ = StatusCode::NO_WARNING;
-
-  // Update the parameters
-  updateParams();
+  setStatus(StatusCode::NO_WARNING);
```
AndyZe, line 379: I think this should be done outside of this function, for readability. It's kind of hidden away here in an unrelated function.

moveit_ros/moveit_servo/src/servo_node.cpp
```diff
@@ -296,42 +298,45 @@ void ServoNode::servoLoop()
...
+      }
+      else if (expected_type == CommandType::POSE && new_pose_msg_)
+      {
+        next_joint_state = processPoseCommand();
+      }
+      else if (new_joint_jog_msg_ || new_twist_msg_ || new_pose_msg_)
       {
-        trajectory_publisher_->publish(composeTrajectoryMessage(servo_->getParams(), next_joint_state.value()));
+        new_joint_jog_msg_ = new_twist_msg_ = new_pose_msg_ = false;
+        RCLCPP_WARN_STREAM(LOGGER, "Command type has not been set, cannot accept input");
```
AndyZe, line 325: Shouldn't the status be set to something invalid here?

moveit_ros/moveit_servo/src/servo.cpp
ibrahiminfinite: Yeah :+1: 
Will include this in the upcoming PR since that PR has some other changes that are need for planning group anyway.

moveit_ros/moveit_servo/src/servo_node.cpp
ibrahiminfinite: Yeah, makes sense.

moveit_ros/moveit_servo/src/servo.cpp
ibrahiminfinite: This is the only place we can set it automatically.
Otherwise user would have to manually call it.

sea-bass: If you have the function either return the status or accept it as a mutable input arg, this could be done!

moveit_ros/moveit_servo/src/servo_node.cpp
```diff
@@ -296,9 +298,12 @@ void ServoNode::servoLoop()
 
   while (rclcpp::ok() && !stop_servo_)
   {
-    // Skip processing if servoing is disabled.
-    if (servo_paused_)
+    // Skip processing commands if servoing is disabled, but allow parameter updates.
```
sea-bass, line 301: I understand not changing the planning group unless servo is paused... but for all the params? Like, if I want to change a scaling factor or enable/disable collision it should now be paused?

ibrahiminfinite: Yeah, this makes things more safer (updating parameters while the control is running doesn't sound safe) and also efficient since update checks are skipped when servo is not paused.

@AndyZe  suggested it during the discussions here.

sea-bass: Does it count as paused when the robot is publishing nothing with no recent input messages, i.e., smoothHalt() is completed?

That seems like a good option vs. Asking uses to explicitly call 2 ROS services before and after changing any param.

sea-bass: Also... what does this mean if you update parameters while servo is unpaused, and then pause/resume? Will you not hit the params out of date flag and have your params be out of sync forever going forward?

I agree it's a good idea for the robot to be stopped before changing values, but we just need to think it through a little to not introduce weird bugs/behaviors.

ibrahiminfinite: > That seems like a good option vs. Asking users to explicitly call 2 ROS services before and after changing any param.

But there is no way to enforce that updates don't happen during the control cycle, I guess we can just add a note about this.

> Also... what does this mean if you update parameters while servo is unpaused, and then pause/resume? Will you not hit the params out of date flag and have your params be out of sync forever going forward?

Yup, I hit this issue while testing.

sea-bass: Yeah, thinking about this, the logic complexity may get a bit complicated.

I would say for the core Servo API, users can do whatever they want and we should add notes about the risks of switching planning groups while moving.

For the servo node, we could still call `updateParameters()` as normal, but keep the "next planning group" stashed in a member variable until Servo has been paused or is otherwise not publishing any commands.

sea-bass: I haven't fully thought about this, but I would say the possibility of desyncing the parameters if not paused is a non-starter for me.

ibrahiminfinite: > For the servo node, we could still call `updateParameters()` as normal, but keep the "next planning group" stashed in a member variable until Servo has been paused or is otherwise not publishing any commands.

Not sure about stashing it, since it would take effect only the next time servo is paused ,which doesn't feel intuitive.

Makes me want to go back to having planning group switch as a service since it would take care pausing , updating and unpausing and the working is very clear.

AndyZe: I think it's a good idea to explicitly require the user to pause Servo before changing any param. No risk of desyncing then, right?

Pretty much agree with what Ibrahim said here:

> this makes things more safer (updating parameters while the control is running doesn't sound safe) and also efficient since update checks are skipped when servo is not paused.

---

> For the servo node, we could still call updateParameters() as normal, but keep the "next planning group" stashed in a member variable until Servo has been paused or is otherwise not publishing any commands.

Sounds complicated

AndyZe: I guess we could add one exception for the collision checking flag, that it's allowed to be updated while unpaused

ibrahiminfinite: > I think it's a good idea to explicitly require the user to pause Servo before changing any param. No risk of desyncing then, right?

We will still have a desync issue since the parameters would be updated in the parameter server. We can only enforce `Servo` not updating it's stored parameters.

sea-bass: Yeah... The service call option could be promising.

I think something that drives the robot to a halt, pauses, switches planning groups, restarts the smoothing plugin, and resumes would be safe.

---

> I guess we could add one exception for the collision checking flag, that it's allowed to be updated while unpaused

My understanding is that most servo parameters can/should be updated while unpaused, and things like the planning group are the exceptions?

AndyZe: But why? Is anybody asking for this? (Other than collision checking toggle)

Definitely anything timing or filtering-related shouldn't be touched while unpaused

sea-bass: On MoveIt Studio we currently modify the collision check toggle and the velocity scaling parameters at runtime -- admittedly, not the others.

I guess my point of view here is servo has a ton of gains/thresholds that would make sense to have runtime tunable because why not? The ROS parameter tooling basically supports this for free.

So I wouldn't want to immediately sacrifice that and take a step back to see how we could handle the planning group switching in a way that doesn't revert a lot of the benefits brought on by switching to `generate_parameter_library`.

AndyZe: i think we're starting to go in circles. Switching the planning group while unpaused isn't a good idea b/c it will screw up filtering, for one issue

sea-bass: 100% agree. I just don't want to introduce bugs or kill all our other runtime parameter tunability to singularly satisfy that requirement.

ibrahiminfinite: How about we allow update of simple parameters like collision toggle etc normally, and check against servo paused status for just planning group change. If not paused we can set status of servo to invalid and print a warning, setting the status to invalid will also skip rest of computations in `getNextJointState` and robot would  be send a command .

ibrahiminfinite: The above method would also avoid the desync issue since parameters of servo would be updated but since status is invalid, no computations will happen until servo was paused and the update is applied properly. It will keep printing warnings till then.

sea-bass: hmm, good point. A warning would go a long way -- then it's user error :)

AndyZe: @ibrahiminfinite that sounds OK.

Can we default nearly all of the parameters to static initially, please? We can unlock them one-by-one as people request it and they get carefully considered. For example, here's a list of parameters that should definitely be static. There may be more.

> monitored_planning_scene_topic
> status_topic
> command_out_topic
> command_out_type
> cartesian_command_in_topic
> joint_command_in_topic
> joint_topic
> thread_priority
> incoming_command_timeout
> publish_period
> use_smoothing
> smoothing_filter_plugin_name

check_collisions and scale can be "unlocked" right now for dynamic updates.

sea-bass: This seems like a good way forward.

We have a MoveIt Studio configuration with UR arm on a linear rail with multiple planning groups (arm+rail vs. rail only), so I'll be happy to test on that once we're at that point.

Wouldn't be a bad idea to use/introduce a similar configuration with multiple planning groups in open-source MoveIt to actually test/showcase these capabilities.

ibrahiminfinite:
> How about we allow update of simple parameters like collision toggle etc normally, and check against servo paused status for just planning group change. If not paused we can set status of servo to invalid and print a warning, setting the status to invalid will also skip rest of computations in getNextJointState and robot would be send a command .

Have implemented this, and kept the call to `updateParams` explicit as before.
Used a `invalid_parameter_update_` flag to handle the setting of status to INVALID

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -226,8 +211,31 @@ bool Servo::updateParams()
       }
       if (params.move_group_name != servo_params_.move_group_name)
       {
-        RCLCPP_INFO_STREAM(LOGGER, "Move group changed from " << servo_params_.move_group_name << " to "
-                                                              << params.move_group_name);
+        if (servo_status_ == StatusCode::PAUSED)
+        {
+          invalid_parameter_update_ = false;
```
sea-bass, line 216: `invalid_parameter_update_` is synonymous with `servo_status != StatusCode::PAUSED` now, so no need for this member variable, right?

moveit_ros/moveit_servo/src/servo.cpp
sea-bass: Reading through the code again, I realize that additionally to everything already said, using the `setStatus()` function the way we use it now is why you had to make all those other member functions non-`const`. So another data point for figuring this out in a more functional way.

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
```diff
@@ -122,14 +128,19 @@ class Servo
    * \brief Get the current state of the robot as given by planning scene monitor.
    * @return The current state of the robot.
    */
-  KinematicState getCurrentRobotState() const;
+  KinematicState getCurrentRobotState();
```
sea-bass, line 131: Why is this not `const` if it just gets the robot state variable?

This is impacting the constness of a bunch of other functions as well

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -527,36 +542,32 @@ TwistCommand Servo::toPlanningFrame(const TwistCommand& command) const
...
-  robot_state->copyJointGroupVelocities(joint_model_group, current_state.velocities);
-  robot_state->copyJointGroupAccelerations(joint_model_group, current_state.accelerations);
+  KinematicState current_state(joint_names_.size());
+  current_state.joint_names = joint_names_;
+  robot_state_->copyJointGroupPositions(joint_model_group_, current_state.positions);
+  robot_state_->copyJointGroupVelocities(joint_model_group_, current_state.velocities);
+  robot_state_->copyJointGroupAccelerations(joint_model_group_, current_state.accelerations);
 
   return current_state;
 }
```
sea-bass, line 562: Yeah, so this function seems to have 2 uses right now, which is why it's not `const`. I think what we need is:

1. An `updateCurrentRobotState()` that does everything done here but doesn't return `current_state`, it simply updates `robot_state_`, to be used in functions that calculate commands. 
2. A `getCurrentRobotState()` function that is `const` and simply returns `robot_state_`, to be used in other tools.

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -527,36 +542,32 @@ TwistCommand Servo::toPlanningFrame(const TwistCommand& command) const
...
-std::pair<bool, KinematicState> Servo::smoothHalt(const KinematicState& halt_state) const
+std::pair<bool, KinematicState> Servo::smoothHalt(const KinematicState& halt_state)
 {
   bool stopped = false;
   auto target_state = halt_state;
   const auto current_state = getCurrentRobotState();
 
-  const size_t num_joints = current_state.joint_names.size();
-  for (size_t i = 0; i < num_joints; i++)
+  for (size_t i = 0; i < joint_names_.size(); i++)
```
sea-bass, line 570:
```suggestion
  for (size_t i = 0; i < joint_names_.size(); ++i)

File: moveit_ros/moveit_servo/include/moveit_servo/servo.hpp [hunk 10]
```diff
@@ -214,6 +220,12 @@ class Servo
   std::unique_ptr<CollisionMonitor> collision_monitor_;
 
   pluginlib::UniquePtr<online_signal_smoothing::SmoothingBaseClass> smoother_ = nullptr;
+
+  moveit::core::RobotStatePtr robot_state_;
+  const moveit::core::JointModelGroup* joint_model_group_;
+  std::vector<std::string> joint_names_;
+
+  bool invalid_parameter_update_;
```
sea-bass, line 228: I don't think we need this if it's used as synonymous with servo being paused.

I also can't find that line in the PR review, but I think your changes make you able to remove the `servo_paused_` member variable as well?

sea-bass (commented):
Read through and I think this looks good.

My biggest thing remaining is ensuring we don't have to un-`const` a lot of our functions. If anything, we need more of them to be `const`. To that point:

* I'll +1 @AndyZe's comment on using `setStatus()` outside the bodies of existing functions so it's more readable and gives us more `const` member functions
* The `getCurrentRobotState()` is now being used both as a getter and to mutate internal state, so I think those 2 things should use 2 separate functions.

sea-bass:
I had one more random thought... if we switch planning groups, things like the planning frame, controller command topic, etc. possibly will need to change too.

ibrahiminfinite:
> I had one more random thought... if we switch planning groups, things like the planning frame, controller command topic, etc. possibly will need to change too.
> 
Yeah, I had mentioned the need for additional parameter changes in one of the earlier comments on this PR.
If things like command topic needs to change, I think the best thing to do is to start a new node.

> If this is getting a bit convoluted, we can take it on at PickNik. Offer still stands!

Yeah that might be a better choice here since you guys will have more grasp on the requirements internally. And I can work on the optimisation PR instead.

AndyZe:
> Yeah, I had mentioned the need for additional parameter changes in one of the earlier comments on this PR. If things like command topic needs to change, I think the best thing to do is to start a new node.

:+1: to this, and coming back full circle now

sea-bass:
Thanks! I'll go ahead and close this PR.

Based on this, I think we can stick with the multiple node solution until/unless we can put some engineering resources on something else.

PR #2335, opened 2023-08-29
Do Servo velocity smoothing in the velocity domain

### LINKED ISSUE #2191: MoveIt Servo moving at vastly different cartesian speed than instructed
### Description

I'm using MoveIt servo to "hover over" a moving object which is moving in a straight line. The position of this moving object is known at all times. 

However, what I am seeing is that when I send a TwistStamped command so the arm will move in a straight line, the actual speed of the arm is vastly different than instructed. I've calculated it to be a 2.7x difference if the speed is 0.01m/s. If the speed changes, the multiplier unfortunately also changes. 

In another [post](https://answers.ros.org/question/411225/moveit-servo-does-not-follow-velocity-commmand-with-ur10e/) it was mentioned this might be due to the JointTrajectoryController not offering a _velocity_ state and command interface, but even when I add this:
```
# This config file is used by ros2_control
controller_manager:
  ros__parameters:
    update_rate: 100  # Control rate, match to robot.

    fanuc_controller:
      type: joint_trajectory_controller/JointTrajectoryController

    joint_state_broadcaster:
      type: joint_state_broadcaster/JointStateBroadcaster


fanuc_controller:
  ros__parameters:
    command_interfaces:
      - position
      - velocity
    state_interfaces:
      - position
      - velocity
    joints:
      - joint_1
      - joint_2
      - joint_3
      - joint_4
      - joint_5
      - joint_6
```
the problem still persists. I've already verified that the correct velocity commands are being sent but they are not being followed. Does anyone know what might be causing this?

This is my configuration file:
```
###############################################
# Modify all parameters related to servoing here
###############################################
use_gazebo: false # Whether the robot is started in a Gazebo simulation environment

## Properties of incoming commands
command_in_type: "speed_units" # "unitless"> in the range [-1:1], as if from joystick. "speed_units"> cmds are in m/s and rad/s


# Optionally override Servo's internal velocity scaling when near singularity or collision (0.0 = use internal velocity scaling)
# override_velocity_scaling_factor = 0.0 # valid range [0.0:1.0]

## Properties of outgoing commands
publish_period: 0.034  # 1/Nominal publish rate [seconds]
low_latency_mode: false  # Set this to true to publish as soon as an incoming Twist command is received (publish_period is ignored)

# What type of topic does your robot driver expect?
# Currently supported are std_msgs/Float64MultiArray or trajectory_msgs/JointTrajectory
command_out_type: trajectory_msgs/JointTrajectory

# What to publish? Can save some bandwidth as most robots only require positions or velocities
publish_joint_positions: true
publish_joint_velocities: true

## Plugins for smoothing outgoing commands
smoothing_filter_plugin_name: "online_signal_smoothing::ButterworthFilterPlugin"

# If is_primary_planning_scene_monitor is set to true, the Servo server's PlanningScene advertises the /get_planning_scene service,
# which other nodes can use as a source for information about the planning environment.
# NOTE: If a different node in your system is responsible for the "primary" planning scene instance (e.g. the MoveGroup node),
# then is_primary_planning_scene_monitor needs to be set to false.
is_primary_planning_scene_monitor: false

## MoveIt properties
move_group_name:  manipulator  # Often 'manipulator' or 'arm'
planning_frame: base_link  # The MoveIt planning frame. Often 'base_link' or 'world'
## Other frames
ee_frame_name: tool0  # The name of the end effector link, used to return the EE pose
robot_link_command_frame: tool0    # commands must be given in the frame of a robot link. Usually either the base or end effector

## Stopping behaviour
incoming_command_timeout:  0.10000 # Stop servoing if X seconds elapse without a new command
# If 0, republish commands forever even if the robot is stationary. Otherwise, specify num. to publish.
# Important because ROS may drop some messages and we need the robot to halt reliably.
num_outgoing_halt_msgs_to_publish: 4

## Configure handling of singularities and joint limits
lower_singularity_threshold:  17.0  # Start decelerating when the condition number hits this (close to singularity)
hard_stop_singularity_threshold: 30.0 # Stop when the condition number hits this
joint_limit_margin: 0.1 # added as a buffer to joint limits [radians]. If moving quickly, make this larger.
leaving_singularity_threshold_multiplier: 2.0 # Multiply the hard stop limit by this when leaving singularity (see https://github.com/ros-planning/moveit2/pull/620)

## Topic names
cartesian_command_in_topic: ~/delta_twist_cmds  # Topic for incoming Cartesian twist commands
joint_command_in_topic: ~/delta_joint_cmds # Topic for incoming joint angle commands
joint_topic: /joint_states
status_topic: ~/status # Publish status to this topic
command_out_topic: /fanuc_controller/joint_trajectory # Publish outgoing commands here

## Collision checking for the entire robot body
check_collisions: true # Check collisions?
collision_check_rate: 10.0 # [Hz] Collision-checking can easily bog down a CPU if done too often.
self_collision_proximity_threshold: 0.01 # Start decelerating when a self-collision is this far [m]
scene_collision_proximity_threshold: 0.02 # Start decelerating when a scene collision is this far [m]
```

### Your environment
* ROS Distro: Humble
* OS Version: Ubuntu 22.04
* RMW: Cyclone

### Steps to reproduce
A geometry_msgs::msg::TwistStamped message is being sent to MoveIt servo. More specifically:
```
        auto msg = std::make_unique<geometry_msgs::msg::TwistStamped>();
        msg->header.stamp = MoveItInterfaceManager::getMoveItInterfaceManager().getROSNode()->now();
        msg->header.frame_id = "base_link";
        msg->twist.linear.x = 0; 
        msg->twist.linear.y = 0.01; 
        twist_cmd_pub_->publish(std::move(msg));
```

### Expected behaviour
The robot moves exactly at the instructed speed. 

### Actual behaviour
The robot moves significantly slower.

AndyZe:
### Description

May fix #2191 

### Test Instructions

No noticeable change when running the demos:  https://moveit.picknik.ai/main/doc/examples/realtime_servo/realtime_servo_tutorial.html

### DIFF
--- a/moveit_ros/moveit_servo/src/servo.cpp
+++ b/moveit_ros/moveit_servo/src/servo.cpp
@@ -425,7 +425,6 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
     // Compute the next joint positions based on the joint position deltas
     target_joint_positions = current_joint_positions + joint_position_delta;
 
-    // Compute velocities based on smoothed joint positions
     target_joint_velocities = (target_joint_positions - current_joint_positions) / servo_params_.publish_period;
 
     // Scale down the velocity based on joint velocity limit or user defined scaling if applicable.
--- a/moveit_ros/moveit_servo/src/servo.cpp
+++ b/moveit_ros/moveit_servo/src/servo.cpp
@@ -425,15 +425,6 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
     // Compute the next joint positions based on the joint position deltas
     target_joint_positions = current_joint_positions + joint_position_delta;
 
-    // TODO : apply filtering to the velocity instead of position
-    // Apply smoothing to the positions if a smoother was provided.
-    // Update filter state and apply filtering in position domain
-    if (smoother_)
-    {
-      smoother_->reset(current_state.positions);
-      smoother_->doSmoothing(target_state.positions);
-    }
-
     // Compute velocities based on smoothed joint positions
     target_joint_velocities = (target_joint_positions - current_joint_positions) / servo_params_.publish_period;
 
@@ -444,6 +435,13 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
       RCLCPP_WARN_STREAM(LOGGER, "Joint velocity limit scaling applied by a factor of " << joint_limit_scale);
 
     target_joint_velocities *= joint_limit_scale;
+
+    // Apply smoothing to the velocity if a smoother was provided.
+    if (smoother_)
+    {
+      smoother_->reset(current_state.velocities);
+      smoother_->doSmoothing(target_state.velocities);
+    }
 
     // Adjust joint position based on scaled down velocity
     target_joint_positions = current_joint_positions + (target_joint_velocities * servo_params_.publish_period);
@@ -553,8 +551,8 @@ std::pair<bool, KinematicState> Servo::smoothHalt(KinematicState halt_state)
 
   if (smoother_)
   {
-    smoother_->reset(current_state.positions);
-    smoother_->doSmoothing(halt_state.positions);
+    smoother_->reset(current_state.velocities);
+    smoother_->doSmoothing(halt_state.velocities);
   }
 
   return std::make_pair(stopped, halt_state);


AndyZe:
@ibrahiminfinite

ibrahiminfinite (commented):
Looks correct, thanks for adding this :+1: 
I had pushed this back in my list since the smoothing plugin API might be updated.
Putting this in now would also be fine.

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -425,15 +425,6 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
...
-    // TODO : apply filtering to the velocity instead of position
-    // Apply smoothing to the positions if a smoother was provided.
-    // Update filter state and apply filtering in position domain
-    if (smoother_)
-    {
-      smoother_->reset(current_state.positions);
-      smoother_->doSmoothing(target_state.positions);
-    }
-
     // Compute velocities based on smoothed joint positions
```
sea-bass, line 428: This comment is no longer accurate

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -445,6 +436,13 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
...
 
+    // Apply smoothing to the velocity if a smoother was provided.
+    if (smoother_)
+    {
+      smoother_->reset(current_state.velocities);
+      smoother_->doSmoothing(target_state.velocities);
+    }
+
     // Adjust joint position based on scaled down velocity
     target_joint_positions = current_joint_positions + (target_joint_velocities * servo_params_.publish_period);
```
sea-bass, line 447: This calculation doesn't seem to use `target_state.velocities` from the previous line at all... is this right?

I guess what is the relationship between `target_state.velocities` and `target_joint_velocities`? Are they linked by reference (if so, that's dangerous and not readable), or have we been using the wrong variables for smoothing?

(this was the case with the positions anyway, so maybe a carry over bug from before unless I'm missing something)

ibrahiminfinite: `target_joint_velocities` is simply an `Eigen:Map` of  `target_state.velocities`.
Used for efficient computation, defined [here](https://github.com/ros-planning/moveit2/blob/47d92ef973680aea1badd6cc6080598f040a89f4/moveit_ros/moveit_servo/src/servo.cpp#L401)

moveit_ros/moveit_servo/src/servo.cpp
AndyZe, line 428: Let's just delete it, eh

```suggestion
```

moveit_ros/moveit_servo/src/servo.cpp
ibrahiminfinite: Btw, the reason `target_state.velocities` and `current_state.velocities` is used with the smoothing API instead of the Eigen map is because the smoothing plugin only accepts `std::vector`, otherwise all of the computations could be performed with the Eigen map.

sea-bass (approved):

sea-bass:
I noticed that running these changes with Gazebo, which drifts a lot, actually makes the drift problem significantly worse.

I wonder if maybe we should add in an option to do the filtering in position vs. velocity?

ibrahiminfinite:
>  I wonder if maybe we should add in an option to do the filtering in position vs. velocity?

I think that would be giving too many choices, we already have a lot of parameters.
Maybe we should just switch to completely using Ruckig which smoothes both positions and velocities. 

BTW, @sea-bass 
Can you try running Servo with `use_smoothing` set to False and see if there is any change in the drifting behaviour.

sea-bass:
Unfortunately, enabling/disabling smoothing doesn't do much... here is what I am seeing. Joint jog seems to work somewhat better, but twist is way off.


https://github.com/ros-planning/moveit2/assets/4603398/80eadc3c-b792-4aa1-8442-623636950341

sea-bass:
Update: Actually, the problem doesn't have to do with the new servo or the filtering -- it's the fact that the `low_latency_mode` (which we removed) was the only thing keeping our Gazebo configuration stable.

I know we chose not to add it... but can I maybe re-add it? lol

AndyZe:
So there's no issue with the PR, right? I think it's a pretty good improvement if you think about the fact that filtering in the position domain encourages the robot to maintain its current position.

AndyZe:
> Maybe we should just switch to completely using Ruckig which smoothes both positions and velocities.

I think the Butterworth option should definitely be kept. Ruckig has some unintuitive behavior. The Butterworth option is about as simple, fast, and idiot-proof as you could get.

ibrahiminfinite (approved):

sea-bass (changes requested):
I don't think we can blanket introduce this change. Here is the UR5e Gazebo config

with position filtering
https://github.com/ros-planning/moveit2/assets/4603398/a9f96527-01d9-4fdd-9c66-f30469d42324

with velocity filtering
https://github.com/ros-planning/moveit2/assets/4603398/26a081fd-cd85-476a-959b-6a55d3e48e90

... so, it should at least be an option.

PR #2344, opened 2023-09-06
Fix collisions_updater CLI if no package is used

### LINKED ISSUE #2338: collisions_updater fails if no package is specified
### Description

Running the collisions_updater results in a runtime error if no package path is used and if the URDF/SRDF pair is provided as absolute paths.

This bug was probably introduced with https://github.com/ros-planning/moveit2/pull/2032, which added the runtime error in case the package path is empty: https://github.com/ros-planning/moveit2/blob/main/moveit_setup_assistant/moveit_setup_framework/src/urdf_config.cpp#L136

henningkayser:
The exceptions introduced with https://github.com/ros-planning/moveit2/pull/2032 broke support for running the collisions updater CLI without a ROS package. This fix makes ROS packages optional again, allowing to use the CLI with absolute paths only.

Fixes #2338


moveit_setup_assistant/moveit_setup_framework/src/urdf_config.cpp
```diff
@@ -131,11 +144,6 @@ void URDFConfig::load()
     throw std::runtime_error("URDF/COLLADA file not found: " + urdf_path_.string());
   }
 
-  if (urdf_pkg_name_.empty())
```
henningkayser, line 134: This exception should not be triggered if we actually don't provide a file path outside of a ROS package. That's why I moved it to `loadFromPackage()` above.

moveit_setup_assistant/moveit_setup_framework/src/urdf_config.cpp
```diff
@@ -92,32 +92,45 @@ void URDFConfig::loadFromPath(const std::filesystem::path& urdf_file_path, const
...
 {
-  bool package_found = extractPackageNameFromPath(urdf_path_, urdf_pkg_name_, urdf_pkg_relative_path_);
-  if (!package_found)
-  {
-    urdf_pkg_name_ = "";
-    urdf_pkg_relative_path_ = urdf_path_;  // just the absolute path
-  }
-  // Check that ROS can find the package
-  const std::filesystem::path robot_desc_pkg_path = getSharePath(urdf_pkg_name_);
+  // Reset to defaults: no package name, relative path is set to absolute path
```
henningkayser, line 95: These changes silence the warning if there is no package name associated with the file.

moveit_setup_assistant/moveit_setup_framework/src/urdf_config.cpp
```diff
@@ -92,32 +92,45 @@ void URDFConfig::loadFromPath(const std::filesystem::path& urdf_file_path, const
...
+    {
+      urdf_pkg_name_ = pkg_name;
+      urdf_pkg_relative_path_ = relative_path;
+    }
+    else
+    {
+      RCLCPP_WARN(*logger_,
+                  "Package Not Found In ROS Workspace. ROS was unable to find the package name '%s'"
+                  " within the ROS workspace. This may cause issues later.",
+                  pkg_name.c_str());
```
sea-bass, line 115: this warning seems very redundant. Can we maybe make it shorter while we're here?

```suggestion
      RCLCPP_WARN(*logger_,
                  "ROS was unable to find the package %s. This may cause issues later.",
                  pkg_name.c_str());
```

sea-bass (approved):
Changes look good -- thanks for this!

sjahr (approved):

PR #2348, opened 2023-09-07
Re-add low_latency_mode parameter to Servo

sea-bass:
This PR re-adds the `low_latency_mode` workaround to servo, but also fixes a bug I found in doing this where the `new_X_msg_` flags were not being unset correctly.

### DIFF
--- a/moveit_ros/moveit_servo/src/servo_node.cpp
+++ b/moveit_ros/moveit_servo/src/servo_node.cpp
@@ -202,6 +202,8 @@ void ServoNode::poseCallback(const geometry_msgs::msg::PoseStamped::ConstSharedP
 
 std::optional<KinematicState> ServoNode::processJointJogCommand()
 {
+  std::optional<KinematicState> next_joint_state = std::nullopt;
+
   const bool command_stale = (node_->now() - latest_joint_jog_.header.stamp) >=
                              rclcpp::Duration::from_seconds(servo_params_.incoming_command_timeout);
   if (!command_stale)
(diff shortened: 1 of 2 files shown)


sea-bass:
@ibrahiminfinite

moveit_ros/moveit_servo/src/servo_node.cpp
```diff
@@ -202,10 +202,6 @@ void ServoNode::poseCallback(const geometry_msgs::msg::PoseStamped::ConstSharedP
 
 std::optional<KinematicState> ServoNode::processJointJogCommand()
 {
-  std::optional<KinematicState> next_joint_state = std::nullopt;
```
sea-bass, line 205: woops, re-add this

sea-bass:
```suggestion
  std::optional<KinematicState> next_joint_state = std::nullopt;

ibrahiminfinite:
How about we give https://github.com/ros-planning/moveit2/issues/1857#issuecomment-1506581979 a try ?
The `publish_period` no longer makes sense if we add `low_latency_mode`.

sea-bass:
> How about we give [#1857 (comment)](https://github.com/ros-planning/moveit2/issues/1857#issuecomment-1506581979) a try ? The `publish_period` no longer makes sense if we add `low_latency_mode`.

I actually just tried this and it performs almost as well as re-adding this parameter! I'll put up a PR soon

PR #2349, opened 2023-09-08
[Servo] Correct for drift by using previously commanded state instead of current state

sea-bass:
This PR implements the suggestion in https://github.com/ros-planning/moveit2/issues/1857#issuecomment-1506581979 to try correct drift for robots without great joint tracking control (like the MoveIt Studio Gazebo simulation).

This adds an `apply_anti_drift` parameter to MoveIt Servo, which when enabled will use the previously commanded joint state instead of the current state to compute joint deltas.

### DIFF
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
@@ -74,9 +74,12 @@ public:
   /**
    * \brief Computes the joint state required to follow the given command.
    * @param command The command to follow, std::variant type, can handle JointJog, Twist and Pose.
+   * @param previous_state_maybe A std::optional containing the previous commanded state. If this value is set, it
+   * is used to calculate an error between the current state and the previous commanded state for anti-drift correction.
    * @return The required joint state.
    */
-  KinematicState getNextJointState(const ServoInput& command);
+  KinematicState getNextJointState(const ServoInput& command,
+                                   const std::optional<KinematicState>& previous_state_maybe = std::nullopt);
 
   /**
    * \brief Set the type of incoming servo command.
@@ -126,10 +129,10 @@ public:
 
   /**
    * \brief Smoothly halt at a commanded state when command goes stale.
-   * @param The last commanded joint states.
+   * @param halt_state_maybe A std::optional containing the last commanded joint states, if available.
    * @return The next state stepping towards the required halting state.
    */
-  std::pair<bool, KinematicState> smoothHalt(KinematicState halt_state);
+  std::pair<bool, KinematicState> smoothHalt(std::optional<KinematicState>& halt_state_maybe);
 
 private:
   /**
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
@@ -74,12 +74,10 @@ public:
   /**
    * \brief Computes the joint state required to follow the given command.
    * @param command The command to follow, std::variant type, can handle JointJog, Twist and Pose.
-   * @param previous_state_maybe A std::optional containing the previous commanded state. If this value is set, it
    * is used to calculate an error between the current state and the previous commanded state for anti-drift correction.
    * @return The required joint state.
    */
-  KinematicState getNextJointState(const ServoInput& command,
-                                   const std::optional<KinematicState>& previous_state_maybe = std::nullopt);
+  KinematicState getNextJointState(const ServoInput& command);
 
   /**
    * \brief Set the type of incoming servo command.
@@ -132,7 +130,14 @@ public:
    * @param halt_state_maybe A std::optional containing the last commanded joint states, if available.
    * @return The next state stepping towards the required halting state.
    */
-  std::pair<bool, KinematicState> smoothHalt(std::optional<KinematicState>& halt_state_maybe);
+  std::pair<bool, KinematicState> smoothHalt();
+
+  /**
+   * TODO
+   */
+  std::optional<KinematicState> getLastCommandedState();
+
+  void resetLastCommandedState();
 
 private:
   /**
@@ -202,6 +207,8 @@ private:
   const rclcpp::Node::SharedPtr node_;
   std::shared_ptr<const servo::ParamListener> servo_param_listener_;
   planning_scene_monitor::PlanningSceneMonitorPtr planning_scene_monitor_;
+
+  std::optional<KinematicState> last_commanded_state_;
 
   // This value will be updated by CollisionMonitor in a separate thread.
   std::atomic<double> collision_velocity_scale_ = 1.0;
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
@@ -133,10 +133,14 @@ public:
   std::pair<bool, KinematicState> smoothHalt();
 
   /**
-   * TODO
+   * \brief Returns the last commanded state.
+   * @return The last commanded state.
    */
   std::optional<KinematicState> getLastCommandedState();
 
+  /**
+   * \brief Resets the last commanded state to std::nullopt.
+   */
   void resetLastCommandedState();
 
 private:
--- a/moveit_ros/moveit_servo/src/servo.cpp
+++ b/moveit_ros/moveit_servo/src/servo.cpp
@@ -394,12 +394,10 @@ KinematicState Servo::getNextJointState(const ServoInput& command,
   KinematicState current_state(num_joints), target_state(num_joints);
   target_state.joint_names = joint_names;
 
-  // If we provide a previous state for drift correction, set that as the robot state.
+  // If we provide a previous state for drift correction, set the robot position to that.
   if (previous_state_maybe)
   {
-    const auto& previous_state = previous_state_maybe.value();
-    robot_state->setJointGroupPositions(joint_model_group, previous_state.positions);
-    robot_state->setJointGroupVelocities(joint_model_group, previous_state.velocities);
+    robot_state->setJointGroupPositions(joint_model_group, previous_state_maybe.value().positions);
   }
 
   // Copy current kinematic data from RobotState.
--- a/moveit_ros/moveit_servo/src/servo.cpp
+++ b/moveit_ros/moveit_servo/src/servo.cpp
@@ -371,8 +371,7 @@ Eigen::VectorXd Servo::jointDeltaFromCommand(const ServoInput& command, moveit::
   return joint_position_deltas;
 }
 
-KinematicState Servo::getNextJointState(const ServoInput& command,
-                                        const std::optional<KinematicState>& previous_state_maybe)
+KinematicState Servo::getNextJointState(const ServoInput& command)
 {
   // Set status to clear
   servo_status_ = StatusCode::NO_WARNING;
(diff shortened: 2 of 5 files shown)


sea-bass:
@ibrahiminfinite

ibrahiminfinite:
The `std::optional` passing approach seems a bit complicated.

How about something like this :

1 . Move  `last_commanded_state_` from `ServoNode` to `Servo`
2 . Compute error [here](https://github.com/ros-planning/moveit2/blob/e9e7ec2e941c9275af717540ed37c362ca04907e/moveit_ros/moveit_servo/src/servo.cpp#L435) and apply it to `target_joint_positions`.



```C++
 
// Compute error from last commanded state
joint_position_error = last_commanded_positions_  - current_joint_positions
// Compute the next joint positions based on the joint position deltas
target_joint_positions = current_joint_positions + joint_position_delta + joint_position_error;
``` 

or 

```C++
 
// Compute error from last commanded state
joint_position_error = last_commanded_positions_  - current_joint_positions
// Compute the next joint positions based on the joint position deltas
target_joint_positions = current_joint_positions + joint_position_delta;

if(servo_params_.apply_anti_drift)
    target_joint_positions += joint_position_error;
``` 

`last_commanded_positions_`  will be an `Eigen::Map` of `last_commanded_state_.positions`
This way none of the public APIs have to change.

The `smoothHalt()` method can then just use the `last_command_state_` without needing to pass it in, since it will already be available as a member variable in `Servo`.

moveit_ros/moveit_servo/src/servo_node.cpp
```diff
@@ -321,7 +323,8 @@ void ServoNode::servoLoop()
       RCLCPP_WARN_STREAM(LOGGER, "Command type has not been set, cannot accept input");
     }
 
-    if (next_joint_state && (servo_->getStatus() != StatusCode::INVALID))
+    if (next_joint_state && (servo_->getStatus() != StatusCode::INVALID) &&
+        (servo_->getStatus() != StatusCode::HALT_FOR_COLLISION))
```
sea-bass, line 327: fun bug: Without this change, the robot would be commanded to go to zero positions when it halted for collision :)

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -460,50 +469,43 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
...
-      adjoint.topRightCorner(3, 3).setZero();
-      adjoint.bottomLeftCorner(3, 3) = skew_translation * rotation;
-      adjoint.bottomRightCorner(3, 3) = rotation;
+      adjoint.topLeftCorner(3, 3) = skew_translation * rotation;
+      adjoint.topRightCorner(3, 3) = rotation;
+      adjoint.bottomLeftCorner(3, 3) = rotation;
+      adjoint.bottomRightCorner(3, 3).setZero();
 
-      reordered_twist = adjoint * reordered_twist;
+      transformed_twist = adjoint * transformed_twist;
```
sea-bass, line 507: simplified here so we don't need to reorder the twist vector -- we just change the matrix

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
```diff
@@ -126,10 +127,21 @@ class Servo
 
   /**
    * \brief Smoothly halt at a commanded state when command goes stale.
-   * @param The last commanded joint states.
+   * @param halt_state_maybe A std::optional containing the last commanded joint states, if available.
```
ibrahiminfinite, line 130:
```suggestion
```
Outdated comment

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
```diff
@@ -200,6 +212,8 @@ class Servo
   std::shared_ptr<const servo::ParamListener> servo_param_listener_;
   planning_scene_monitor::PlanningSceneMonitorPtr planning_scene_monitor_;
 
+  std::optional<KinematicState> last_commanded_state_;
```
ibrahiminfinite, line 215:
```suggestion
  KinematicState last_commanded_state_;
```
I think this should not be optional, since `last_commanded_state_ ` should technically always have a value.

ibrahiminfinite:
I am not sure about computing the required joint delta by setting `robot_state` positions to `last_commanded_state_`.
This would mean we are computing the delta assuming that the robot actually reached the `last_commanded_state_`.

What I think should happen is:
In case of a drift, the robot is currently at positions `q_current` (under/overshoot), but should have been at `q_target`.
So when we compute the new joint delta, we want the error `q_target - q_current` to be accounted for.
If we use the last commanded positions `q_target`, the delta would be smaller/larger than what it should be, depending on if the drift was an undershoot or overshoot.

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
```diff
@@ -74,6 +74,7 @@ class Servo
   /**
    * \brief Computes the joint state required to follow the given command.
    * @param command The command to follow, std::variant type, can handle JointJog, Twist and Pose.
+   * is used to calculate an error between the current state and the previous commanded state for anti-drift correction.
```
AndyZe, line 77: Continuing the formatting nitpick theme  ;)

```suggestion
   * Is used to calculate an error between the current state and the previous commanded state for anti-drift correction.
```

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
AndyZe: This seems like the type of thing that should be a free function, eh? Transparent inputs and outputs

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
```diff
@@ -126,10 +127,21 @@ class Servo
 
   /**
    * \brief Smoothly halt at a commanded state when command goes stale.
-   * @param The last commanded joint states.
+   * @param halt_state_maybe A std::optional containing the last commanded joint states, if available.
    * @return The next state stepping towards the required halting state.
```
AndyZe, line 131: What does the `bool` component of the returned value mean?

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
AndyZe: i'm ok with the optional. Maybe it wouldn't have a value if the program were just started or somebody just called `resetLastCommandedState()`

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -397,7 +397,16 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
   robot_state->copyJointGroupPositions(joint_model_group, current_state.positions);
   robot_state->copyJointGroupVelocities(joint_model_group, current_state.velocities);
 
+  // If we provide a previous state for drift correction, set the robot position to that.
+  auto initial_state = current_state;
+  if (servo_params_.apply_anti_drift && last_commanded_state_.has_value())
+  {
+    robot_state->setJointGroupPositions(joint_model_group, last_commanded_state_.value().positions);
```
AndyZe, line 404: Since we know last_commanded_state_ has a value

```suggestion
    robot_state->setJointGroupPositions(joint_model_group, *last_commanded_state_.positions);
```

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -423,7 +432,7 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
     joint_position_delta *= collision_velocity_scale_;
 
     // Compute the next joint positions based on the joint position deltas
-    target_joint_positions = current_joint_positions + joint_position_delta;
+    target_joint_positions = initial_joint_positions + joint_position_delta;
```
AndyZe, line 435: Maybe i missed it but i don't see that `current_joint_positions` was modified at all between L410 and here. So why not use `current_joint_positions` and don't bother creating `initial_joint_positions`?

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -534,30 +544,47 @@ KinematicState Servo::getCurrentRobotState()
...
+  if (last_commanded_state_)
   {
-    const double vel = (halt_state.positions[i] - current_state.positions[i]) / servo_params_.publish_period;
-    halt_state.velocities[i] = (vel > STOPPED_VELOCITY_EPS) ? vel : 0.0;
-    halt_state.accelerations[i] =
-        (halt_state.velocities[i] - current_state.velocities[i]) / servo_params_.publish_period;
-  }
+    auto& halt_state = last_commanded_state_.value();
+    const size_t num_joints = current_state.joint_names.size();
+    for (size_t i = 0; i < num_joints; i++)
```
AndyZe, line 556: nit

```suggestion
    for (size_t i = 0; i < num_joints; ++i)
```

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -534,30 +544,47 @@ KinematicState Servo::getCurrentRobotState()
...
+      smoother_->reset(current_state.positions);
+      smoother_->doSmoothing(halt_state.positions);
+    }
+    return std::make_pair(stopped, halt_state);
+  }
+  else
   {
-    smoother_->reset(current_state.positions);
-    smoother_->doSmoothing(halt_state.positions);
+    return std::make_pair(stopped, current_state);
```
AndyZe, line 576: `stopped` appears to be false here, I think it should be true

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -534,30 +544,47 @@ KinematicState Servo::getCurrentRobotState()
...
+      const double vel = (halt_state.positions[i] - current_state.positions[i]) / servo_params_.publish_period;
+      halt_state.velocities[i] = (vel > STOPPED_VELOCITY_EPS) ? vel : 0.0;
+      halt_state.accelerations[i] =
+          (halt_state.velocities[i] - current_state.velocities[i]) / servo_params_.publish_period;
+    }
 
-  // If all velocities are zero, robot has decelerated to a stop.
-  stopped = (std::accumulate(halt_state.velocities.begin(), halt_state.velocities.end(), 0.0) == 0.0);
+    // If all velocities are zero, robot has decelerated to a stop.
+    stopped = (std::accumulate(halt_state.velocities.begin(), halt_state.velocities.end(), 0.0) == 0.0);
```
AndyZe, line 565: In practice, i think there will always be some noise so this will prob fail

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
ibrahiminfinite: It is a public API of `Servo` which is used by `ServoNode`, not sure if making it a free function would make sense contextually  even though it can easily be a free function.

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
ibrahiminfinite: > Maybe it wouldn't have a value if the program were just started or somebody just called `resetLastCommandedState()`

Would it  make sense for `resetLastCommandedState()` to reset it to the current state rather than a `std::nullopt`.
We could then call this same function in the constructor to initialize the `last_commanded_state_`.

moveit_ros/moveit_servo/src/servo.cpp
ibrahiminfinite: Ah yes, the specific noise level would also depend on hardware I guess.
Would checking `<=` some threshold like `0.00001` be enough to account for the noise ?

sea-bass: Agreed. There already is a variable for this, so I'll update it in my bug fix PR.

sea-bass:
Guys, I am not sure this ended up helping at all, so I may close this PR. Still need to do some more testing.

PR #2350, opened 2023-09-08
[Servo] Fix bugs when halting for collision + transforming commands to planning frame

### LINKED ISSUE #2352: [Servo] Using TF lookup instead of robot state for transforms causing drift
I was seeing some weird drift in the new MoveIt Servo, and finally found the source of the issue.

In new servo we're doing conversions to the planning frame using a TF buffer:

```cpp
const auto planning_to_command_tf =
    transform_buffer_.lookupTransform(command.frame_id, servo_params_.planning_frame, rclcpp::Time(0));
```

In the old servo, we're doing conversions to the planning frame using the robot state:

```cpp
moveit::core::RobotStatePtr robot_state = planning_scene_monitor_->getStateMonitor()->getCurrentState();
const Eigen::Isometry3d planning_to_command_tf_eigen =
   robot_state->getGlobalLinkTransform(servo_params_.planning_frame).inverse() *
   robot_state->getGlobalLinkTransform(command.frame_id);
```

While functionally this is the same, it's causing significant drift in these systems.

New servo (even after the correction in https://github.com/ros-planning/moveit2/pull/2350):
https://github.com/ros-planning/moveit2/assets/4603398/43caf2e6-bb25-42ab-a115-884b57a9c1f6

Old servo (actually, this is new servo but with the transform approach reverted as above):
https://github.com/ros-planning/moveit2/assets/4603398/c4b778fe-59f3-4f12-a94a-9763fed33d49

---

So, I think we should have some kind of check on whether the frame name can be retrieved using the robot model or not, and that should decide whether to use TF.

@AndyZe @ibrahiminfinite

sea-bass:
This PR fixes a few bugs:
1. Robot was being commanded to zero state when halting for collision, now we simply don't publish outputs in this case.
2. When applying twists about the EE frame, the translation component of the transform was also being used, plus the TF `lookupTransform()` call had arguments reversed, causing drift. Fixes the issue described in https://github.com/ros-planning/moveit2/issues/2352, and applies the same fixes to pose tracking conversions as well.
4. Changed the stopped velocity threshold for the `smoothHalt()` function to use a nonzero threshold instead of exact floating-point comparison to zero.
5. Make some functions accept input arguments as const to speed things up
6. Removed the redundant `tf2::TransformBuffer` and `tf2::TransformListener` that were already implicit in the Planning Scene Monitor.

Additionally, it simplifies some of the new twist conversion code from https://github.com/ros-planning/moveit2/pull/2311 so there is no need to reorder a vector.

Closes #2352


sea-bass:
Moved some of the unrelated fixes from https://github.com/ros-planning/moveit2/pull/2349 (which I'm not sure will actually help) to here. These changes should definitely go in.

cc @ibrahiminfinite

ibrahiminfinite:
Great catch!
I had  set it up so that `getNextJointState()` would return the current state in case of halting for collision, but it seems I forgot to copy the current positions to target state.
Since this should work for both C++ and ROS APIs, I think it might be better to add 

```C++
robot_state->copyJointGroupPositions(joint_model_group, target_state.positions);
```
[here](https://github.com/ros-planning/moveit2/blob/2644cf54f304f20840e81cf401f47b8439b27f66/moveit_ros/moveit_servo/src/servo.cpp#L398), no need to check for the collision flag in `servoLoop()`

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -465,45 +465,37 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
...
-      const Eigen::Vector3d& translation = planning_to_command_tf.translation();
+
+      const auto planning_to_command_tf_eigen = tf2::transformToEigen(planning_to_command_tf);
+      const Eigen::Matrix3d& rotation = planning_to_command_tf_eigen.rotation();
+      const Eigen::Vector3d& translation = planning_to_command_tf_eigen.translation();
 
       Eigen::Matrix3d skew_translation;
       skew_translation.row(0) << 0, -translation(2), translation(1);
       skew_translation.row(1) << translation(2), 0, -translation(0);
       skew_translation.row(2) << -translation(1), translation(0), 0;
```
ibrahiminfinite, line 491: Would be good to have some comments here about the matrix flip to avoid confusion, since the references have the actual equation only.

sea-bass: Good idea -- done!

sea-bass:
@ibrahiminfinite Implemented your comments, and also found a follow-on bug with converting twists to planning frame, so I changed this PR name/description.

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -459,51 +459,55 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
       target_state = haltJoints(joints_to_halt, current_state, target_state);
     }
   }
+  else
+  {
+    robot_state->copyJointGroupPositions(joint_model_group, target_state.positions);
```
AndyZe, line 464: I'm not sure I see the bug here. L397 already does pretty much the same thing, right? Is `robot_state` modified after L397? I don't think so, jointDeltaFromCommand() doesn't change it.

Also, what about velocities like at L398?

Also, can the `robot_state` parameter of jointDeltaFromCommand() be `const`? That would eliminate the confusion.

sea-bass: Good questions

* L397 copies the `robot_state` into the `current_state` variable (for computing the joint deltas), wheras this line copies the `robot_state` into the `target_state` variable (for output). Without this line, the case of halting for collision would leave `target_state` at all zeros.
* I thought about that, but don't we want the target state velocities to be all zeros in the case of a hard stop?

moveit_ros/moveit_servo/src/servo.cpp
sea-bass: I actually did need this check anyway...

moveit_ros/moveit_servo/src/servo.cpp
sea-bass: Actually, for your first question, you're right. We do not need to have this else statement and instead we just don't publish commands in this status.

ibrahiminfinite:
Tested this, all the demos worked fine and applying twist commands to panda also worked as expected.

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -463,28 +462,53 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
...
 }
 
-const TwistCommand Servo::toPlanningFrame(const TwistCommand& command)
+Eigen::Isometry3d Servo::getPlanningToCommandFrameTransform(const std::string& command_frame) const
 {
-  Eigen::Vector<double, 6> transformed_twist = command.velocities;
+  const moveit::core::RobotStatePtr robot_state = planning_scene_monitor_->getStateMonitor()->getCurrentState();
+  if (robot_state->knowsFrameTransform(command_frame))
+  {
+    return robot_state->getGlobalLinkTransform(servo_params_.planning_frame).inverse() *
```
ibrahiminfinite, line 470: Do we need this?
```
robot_state->getGlobalLinkTransform(servo_params_.planning_frame).inverse() *
```
Doesn't `getGlobalLinkTransform(command.frame_id)` already give transform w.r.t to `planning_frame` ?

```
Get the link transform w.r.t. the root link (model frame) of the RobotModel.
This is typically the root link of the URDF unless a virtual joint is present.
```

sea-bass: Is the URDF root link necessarily the base link of the planning group? I would think not.

Also, there's the note about the virtual joint you pasted in as a (distant) second reason to want to keep this?

ibrahiminfinite: I see, in that case , better to keep it :+1:

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -463,28 +462,53 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
...
+           robot_state->getGlobalLinkTransform(command_frame);
+  }
+  else
+  {
+    return tf2::transformToEigen(
```
henningkayser, line 475: why are we running a transform buffer again? PSM already provides one for us, and uses it internally.

henningkayser (approved):

moveit_ros/moveit_servo/src/servo.cpp
sea-bass: Whoa, I didn't know!

I would love to make this change in a follow-on PR. Should I add a public method to PSM to use its `tf_listener_` to grab a transform? Or make it public? Open to suggestions.

sea-bass: Never mind, there already is a `getTFClient()` function...

PR #2351, opened 2023-09-09
Using std types instead of boost for Gaussian sampling

### LINKED ISSUE #2166: Modernize random sampling in STOMP
The STOMP planner is using boost types for sampling from a Gaussian distribution.
We should replace them with std types, possibly featuring a new implementation in [RSL](https://github.com/PickNikRobotics/RSL/blob/main/include/rsl/random.hpp).

Shobuj-Paul:
### Description

Getting rid of boost for sampling from a gaussian distribution and instead using standard library functions, wherever available.

Fixes #2166 

### Checklist

### DIFF
--- a/moveit_planners/stomp/include/stomp_moveit/math/multivariate_gaussian.hpp
+++ b/moveit_planners/stomp/include/stomp_moveit/math/multivariate_gaussian.hpp
@@ -42,7 +42,7 @@
 #include <Eigen/Core>
 #include <Eigen/Cholesky>
 
-#include <random>
+#include <rsl/random.hpp>
 #include <cstdlib>
 
 namespace stomp_moveit
@@ -75,8 +75,6 @@ private:
   Eigen::MatrixXd covariance_cholesky_; /**< Cholesky decomposition (LL^T) of the covariance */
 
   int size_;
-  std::mt19937 rng_;
-  std::random_device rd_{};
   std::normal_distribution<double> normal_dist_;
 };
 
@@ -87,7 +85,6 @@ MultivariateGaussian::MultivariateGaussian(const Eigen::MatrixBase<Derived1>& me
                                            const Eigen::MatrixBase<Derived2>& covariance)
   : mean_(mean), covariance_(covariance), covariance_cholesky_(covariance_.llt().matrixL()), normal_dist_(0.0, 1.0)
 {
-  rng_.seed(rd_());
   size_ = mean.rows();
 }
 
@@ -95,7 +92,7 @@ template <typename Derived>
 void MultivariateGaussian::sample(Eigen::MatrixBase<Derived>& output, bool use_covariance)
 {
   for (int i = 0; i < size_; ++i)
-    output(i) = normal_dist_(rng_);
+    output(i) = normal_dist_(rsl::rng());
 
   if (use_covariance)
   {
--- a/moveit_planners/stomp/include/stomp_moveit/math/multivariate_gaussian.hpp
+++ b/moveit_planners/stomp/include/stomp_moveit/math/multivariate_gaussian.hpp
@@ -76,6 +76,7 @@ private:
 
   int size_;
   std::mt19937 rng_;
+  std::random_device rd_{};
   std::normal_distribution<double> normal_dist_;
 };
 
@@ -86,7 +87,7 @@ MultivariateGaussian::MultivariateGaussian(const Eigen::MatrixBase<Derived1>& me
                                            const Eigen::MatrixBase<Derived2>& covariance)
   : mean_(mean), covariance_(covariance), covariance_cholesky_(covariance_.llt().matrixL()), normal_dist_(0.0, 1.0)
 {
-  rng_.seed(rand());
+  rng_.seed(rd_());
   size_ = mean.rows();
 }
 
--- a/moveit_planners/stomp/include/stomp_moveit/math/multivariate_gaussian.hpp
+++ b/moveit_planners/stomp/include/stomp_moveit/math/multivariate_gaussian.hpp
@@ -80,9 +80,9 @@ private:
   int size_;
   std::mt19937 rng_;
   std::normal_distribution<double> normal_dist_;
-  std::shared_ptr<boost::variate_generator<
-      std::mt19937, STD_MSGS__MSG__DETAIL__COLOR_RGBA__TYPE_SUPPORT_HPP_::std::normal_distribution<>>>
-      gaussian_;
+  // std::shared_ptr<boost::variate_generator<
+  //     std::mt19937, STD_MSGS__MSG__DETAIL__COLOR_RGBA__TYPE_SUPPORT_HPP_::std::normal_distribution<>>>
+  //     gaussian_;
 };
 
 //////////////////////// template function definitions follow //////////////////////////////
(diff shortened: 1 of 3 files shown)


Shobuj-Paul:
@henningkayser 
There isn't a standard function for boost::variate_generator  (refer [this](https://stackoverflow.com/questions/13691813/stdvariate-generator-in-vs2012) link). 
Can we think of writing our own implementation of variate_generator, or try to patch the `gaussian_` variable together with std::bind?

Shobuj-Paul:
I am directly calling the dist_(eng_) for updating output matrix, becasue that is what `()` operator is doing in variate_generator also as per my understanding. Please review and let me know if that is correct. Also, how do I test if this is working correctly?

Shobuj-Paul:
@henningkayser 
Could you review this when you're free? Thanks!

henningkayser (approved):

moveit_planners/stomp/include/stomp_moveit/math/multivariate_gaussian.hpp
```diff
@@ -79,9 +75,8 @@ class MultivariateGaussian
   Eigen::MatrixXd covariance_cholesky_; /**< Cholesky decomposition (LL^T) of the covariance */
 
   int size_;
-  boost::mt19937 rng_;
-  boost::normal_distribution<> normal_dist_;
-  std::shared_ptr<boost::variate_generator<boost::mt19937, boost::normal_distribution<> > > gaussian_;
+  std::mt19937 rng_;
```
sjahr, line 78: I think the rng's seed needs to be randomized. E.g. with random_device:
```
    std::random_device rd{};
    std::mt19937 rng{rd()};
```
If you don't explicitly choose, the random sequence will be always the same since it is initialized with 5489u ([reference](https://en.cppreference.com/w/cpp/numeric/random/mersenne_twister_engine)). Try running this example in [godbolt](https://godbolt.org/) and then replace `std::mt19937 rng` with the two lines above

```cpp
// Type your code here, or load an example.
#include <random>
#include <iostream>

int main(int argc, char** argv)  {
    std::mt19937 rng;
    std::normal_distribution<double> normal_dist;
    
    for(auto i = 0; i < 10; i++){
        std::cout << normal_dist(rng) << std::endl;
    }
    return 0;
}
```
You can re-run the program with the small :arrow_right: icon in the bottom left

sjahr (changes requested):
@henningkayser Not sure if the rng initialized with the default value has the desired effect.

moveit_planners/stomp/include/stomp_moveit/math/multivariate_gaussian.hpp
Shobuj-Paul: You're right @sjahr , I'll make the changes in the evening.

moveit_planners/stomp/include/stomp_moveit/math/multivariate_gaussian.hpp
```diff
@@ -42,11 +42,7 @@
 #include <Eigen/Core>
 #include <Eigen/Cholesky>
 
-// TODO(#2166): Replace with std types
-#include <boost/random/variate_generator.hpp>
-#include <boost/random/normal_distribution.hpp>
-#include <boost/random/mersenne_twister.hpp>
-#include <boost/shared_ptr.hpp>
+#include <random>
```
sjahr, line 45:
```suggestion
#include <rsl/random.hpp>
```

moveit_planners/stomp/include/stomp_moveit/math/multivariate_gaussian.hpp
```diff
@@ -79,9 +75,9 @@ class MultivariateGaussian
   Eigen::MatrixXd covariance_cholesky_; /**< Cholesky decomposition (LL^T) of the covariance */
 
   int size_;
-  boost::mt19937 rng_;
-  boost::normal_distribution<> normal_dist_;
-  std::shared_ptr<boost::variate_generator<boost::mt19937, boost::normal_distribution<> > > gaussian_;
+  std::mt19937 rng_;
+  std::random_device rd_{};
```
sjahr, line 79:
```suggestion
```

moveit_planners/stomp/include/stomp_moveit/math/multivariate_gaussian.hpp
```diff
@@ -91,16 +87,15 @@ MultivariateGaussian::MultivariateGaussian(const Eigen::MatrixBase<Derived1>& me
                                            const Eigen::MatrixBase<Derived2>& covariance)
   : mean_(mean), covariance_(covariance), covariance_cholesky_(covariance_.llt().matrixL()), normal_dist_(0.0, 1.0)
 {
-  rng_.seed(rand());
+  rng_.seed(rd_());
```
sjahr, line 90: I missed in your original PR that the rng is seeded and not just the default constructor used :see_no_evil: 
```suggestion
```

moveit_planners/stomp/include/stomp_moveit/math/multivariate_gaussian.hpp
```diff
@@ -91,16 +87,15 @@ MultivariateGaussian::MultivariateGaussian(const Eigen::MatrixBase<Derived1>& me
...
   size_ = mean.rows();
-  gaussian_.reset(new boost::variate_generator<boost::mt19937, boost::normal_distribution<> >(rng_, normal_dist_));
 }
 
 template <typename Derived>
 void MultivariateGaussian::sample(Eigen::MatrixBase<Derived>& output, bool use_covariance)
 {
   for (int i = 0; i < size_; ++i)
-    output(i) = (*gaussian_)();
+    output(i) = normal_dist_(rng_);
```
sjahr, line 98:
```suggestion
    output(i) = normal_dist_(rsl::rng());
```

sjahr (commented):
Sorry @Shobuj-Paul I just discover an even simpler way to do this. We had a PR ~1 year ago which [unified the use of random number generators across the repository](https://github.com/ros-planning/moveit2/pull/1721/files#top)(see the changes for the CHOMP planner as references for STOMP) by using a [unified implementation to create a random number generator ](https://github.com/PickNikRobotics/RSL/blob/main/src/random.cpp). It is more performant since creating a random device is computationally expensive. Can you test my suggestions? All that's left to do would be to add `<depend>rsl</depend>` to the CMakeLists.txt. If that works this PR is finally good to go :sweat_smile:

moveit_planners/stomp/include/stomp_moveit/math/multivariate_gaussian.hpp
Shobuj-Paul: It was seeded with rand(), but that wasn't generating random sequences either, as I checked at your suggestion, so I changed it to random_device. But either way, the RSL implementation looks a lot cleaner.

Is there any issue/PR reference explaining how it is more performant? Cause from what I see in the above links, rsl::rng() internally uses random_device.

sjahr: Performance issues are mainly caused by creating multiple random devices (e.g. if you construct a lot of objects that have a random device member). Using the RSL implementation improves that by using a thread_local generator and only creating a new one (and a random device instance) if it does not exist within the a thread. Otherwise the already existing instance is shared.

sjahr (approved):

PR #2364, opened 2023-09-14
Update "Store OMPL Planning Data" PR

### LINKED ISSUE #2317: Add service to store OMPL motion planner data
This looks like an easy one. The PR was approved but the author never came back to fix clang-tidy warnings. Here is the PR: https://github.com/ros-planning/moveit2/pull/1183

Shobuj-Paul:
### Description

Resolves the merge conflicts and cling-tidy errors for PR #1183 
Fixes #2317 

### Checklist


Shobuj-Paul:
I have replaced std::bind function with lambdas. If this fix looks good, we can remove the defined function 'storePlannerData'.

moveit_planners/ompl/ompl_interface/src/planning_context_manager.cpp
```diff
@@ -89,17 +89,23 @@ struct PlanningContextManager::CachedContexts
...
 
-MultiQueryPlannerAllocator::~MultiQueryPlannerAllocator()
+bool MultiQueryPlannerAllocator::storePlannerData()
 {
   // Store all planner data
   for (const auto& entry : planner_data_storage_paths_)
   {
     ob::PlannerData data(planners_[entry.first]->getSpaceInformation());
     planners_[entry.first]->getPlannerData(data);
+    if (data.numVertices() == 0)
```
henningkayser, line 99: Please add an debug info for this here. The error in `OMPLInterface::storePlannerData()` shouldn't make assumptions about the reason for failures in this function.

moveit_planners/ompl/ompl_interface/src/ompl_interface.cpp
```diff
@@ -71,10 +88,41 @@ OMPLInterface::OMPLInterface(const moveit::core::RobotModelConstPtr& robot_model
   RCLCPP_DEBUG(LOGGER, "Initializing OMPL interface using specified configuration");
   setPlannerConfigurations(pconfig);
   loadConstraintSamplers();
+
+  store_planner_data_service_ = node_->create_service<std_srvs::srv::Trigger>(
+      "store_planner_data",
+      [this](const std::shared_ptr<std_srvs::srv::Trigger::Request>& /* unused */,
+             const std::shared_ptr<std_srvs::srv::Trigger::Response>& response) -> void {
+        bool success = context_manager_.storePlannerData();
```
henningkayser, line 96: this should call `OMPLInterface::storePlannerData()` instead of copying the function body

moveit_planners/ompl/ompl_interface/src/ompl_interface.cpp
```diff
@@ -56,6 +57,22 @@ OMPLInterface::OMPLInterface(const moveit::core::RobotModelConstPtr& robot_model
   RCLCPP_DEBUG(LOGGER, "Initializing OMPL interface using ROS parameters");
   loadPlannerConfigurations();
   loadConstraintSamplers();
+
+  store_planner_data_service_ = node_->create_service<std_srvs::srv::Trigger>(
+      "store_planner_data",
+      [this](const std::shared_ptr<std_srvs::srv::Trigger::Request>& /* unused */,
+             const std::shared_ptr<std_srvs::srv::Trigger::Response>& response) -> void {
+        bool success = context_manager_.storePlannerData();
```
henningkayser, line 65: same

create_client
henningkayser, line 1: remove?

henningkayser (commented):
Thank you!
While I'm not a huge fan of adding a direct rclcpp service dependency to the ompl plugin, I see why it might be userful. IMO, it would be much cleaner to use some event API that exposes detailed features like this through the planning plugin, and only bind a service to it via moveit_ros. I'm fine with merging this feature anyway, provided the comments are resolved.

moveit_planners/ompl/ompl_interface/src/ompl_interface.cpp
Shobuj-Paul: Previous commit just does that by calling std::bind, but it fails clang-tidy. Any way around that?

create_client
Shobuj-Paul: This PR is a continuation from previous work, where std::bind was used (check previous commits).
I couldn't find any other way to remove clang-tidy warnings other than using explicity using a lambda.

Shobuj-Paul:
> Thank you! While I'm not a huge fan of adding a direct rclcpp service dependency to the ompl plugin, I see why it might be userful. IMO, it would be much cleaner to use some event API that exposes detailed features like this through the planning plugin, and only bind a service to it via moveit_ros. I'm fine with merging this feature anyway, provided the comments are resolved.

I am not sure on how to proceed with that, but I could work on it instead of this PR if given some direction.

create_client
Shobuj-Paul: My bad, not sure how this showed up.

dyackzan:
I have reviewed and am ok approving these changes once Henning's comments are resolved. Tagging @henningkayser for his reply here

dyackzan (commented):
I have reviewed and am ok approving these changes once Henning's comments are resolved. Tagging @henningkayser for his reply here

moveit_planners/ompl/ompl_interface/src/planning_context_manager.cpp
Shobuj-Paul: Added debug info, please check if this is what you meant.

moveit_planners/ompl/ompl_interface/src/planning_context_manager.cpp
```diff
@@ -89,17 +89,24 @@ struct PlanningContextManager::CachedContexts
...
+    {
+      RCLCPP_DEBUG_STREAM(LOGGER, "Could not fetch space information or space information is empty.");
```
sjahr, line 101: I think the log level for this could be ERROR

peterfryd:
Hi, what is the timeline for this PR? 
I am currently doing planning for a 6 DOF in  a rather complex static environment in Ros2 Humble and i have the same issue as this guy: https://github.com/moveit/moveit/issues/2160.
As it is now, the successrate for the persistentLazyPrmStar planner i use from the Ompl library isnt close to being good enough without needing a huge amount of nodes which causes the saving process to slow down. Hence, the node doesnt successfully save the data before being killed by SIGTERM which then causes the data to become corrupt.
I see that there is a service available called /save_map but the saved roadmap cant be read properly by the planner for later uses.
Would love to get some help, and i see that this PR is related to it :-))

PR #2377, opened 2023-09-20
Compare MoveIt! Jacobian against KDL

marioprats:
### Description

Adds the KDL Jacobian to the MoveIt Jacobian benchmark to have a baseline. As it is today, MoveIt! Jacobian is already ~2x faster than KDL.

```
------------------------------------------------------------
Benchmark                  Time             CPU   Iterations
------------------------------------------------------------
BM_MoveItJacobian        558 ns          558 ns      1251060
BM_KDLJacobian          1013 ns         1014 ns       686766
```

This change adds a dependency from `robot_state` to `kdl_parser`, but `robot_core` already depends on KDL, so I think this should be ok?

### Checklist


sjahr (approved):
Nice, thank you! I think the kdl_parser as test dependency should be fine :+1:

henningkayser (approved):

PR #2381, opened 2023-09-23
[Servo] Set static parameters as `read-only`

### LINKED ISSUE #2367: [Servo] Set paramters that will/should not be updated during runtime to `read-only`
There are parameters in the Servo config that are only used during initialization and some that should not be updated during runtime.
To make this explicit to the users, it will be best to set them as `read-only`.

Some examples of the above parameters are :

```
monitored_planning_scene_topic
status_topic
command_out_topic
command_out_type
cartesian_command_in_topic
joint_command_in_topic
joint_topic
thread_priority
incoming_command_timeout
publish_period
use_smoothing
smoothing_filter_plugin_name
```

ibrahiminfinite:
### Description

Fixes #2367 

Many parameters in servo should not be changed during runtime.
This PR sets such parameters as read-only through the generate_parameter_library.
The parameters have also been grouped in a more meaningful manner.

Ping: @sea-bass  @AndyZe

### DIFF
--- a/moveit_ros/moveit_servo/config/servo_parameters.yaml
+++ b/moveit_ros/moveit_servo/config/servo_parameters.yaml
@@ -25,6 +25,7 @@ servo:
 
   move_group_name: {
     type: string,
+    read_only: true,
     description: "The name of the moveit move_group of your robot \
                   This parameter does not have a default value and \
                   must be passed to the node during launch time."
@@ -70,7 +71,6 @@ servo:
 
   command_in_type: {
     type: string,
-    read_only: true,
     default_value: "unitless",
     description: "The unit of the incoming command. \
                   unitless commands are in the range [-1:1] as if from joystick \
@@ -100,7 +100,6 @@ servo:
 
   incoming_command_timeout: {
     type: double,
-    read_only: true,
     default_value: 0.1,
     description: "Commands will be discarded if it is older than the timeout.\
                    Important because ROS may drop some messages."
@@ -307,13 +306,13 @@ servo:
   halt_all_joints_in_joint_mode: {
     type: bool,
     default_value: true,
-    description: "Halt all joints in joint mode"
+    description: "Halt all joints in joint mode, else halt only the joints at their limit"
   }
 
   halt_all_joints_in_cartesian_mode: {
     type: bool,
     default_value: true,
-    description: "Halt all joints in cartesian mode"
+    description: "Halt all joints in cartesian mode, else halt only the joints at their limit"
   }
 
   joint_limit_margin: {
--- a/moveit_ros/moveit_servo/config/servo_parameters.yaml
+++ b/moveit_ros/moveit_servo/config/servo_parameters.yaml
@@ -1,5 +1,28 @@
 servo:
-################################ ROBOT SPECIFIC CONFIG #############################
+################################ GENERAL CONFIG ################################
+
+  thread_priority: {
+    type: int,
+    read_only: true,
+    default_value: 40,
+    description: "This value is used when configuring the servo loop thread to use SCHED_FIFO scheduling \
+                  We use a slightly lower priority than the ros2_control default in order to reduce jitter \
+                  Reference: https://man7.org/linux/man-pages/man2/sched_setparam.2.html",
+    validation: {
+      gt_eq<>: 0
+    }
+  }
+
+  publish_period: {
+    type: double,
+    read_only: true,
+    default_value: 0.034,
+    description: " 1 / (Nominal publish rate) [seconds]",
+    validation: {
+      gt<>: 0.0
+    }
+  }
+
   move_group_name: {
     type: string,
     description: "The name of the moveit move_group of your robot \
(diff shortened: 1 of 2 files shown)


moveit_ros/moveit_servo/config/servo_parameters.yaml
```diff
@@ -22,58 +45,32 @@ servo:
...
   joint_command_in_topic: {
     type: string,
+    read_only: true,
     default_value: "~/delta_joint_cmds",
     description: "The topic on which servo will receive the joint jog commands"
   }
 
   command_in_type: {
     type: string,
+    read_only: true,
```
sea-bass, line 73: Why should this be read only? Far as I know, this is just an if statement in the code so pretty easy to tune at runtime

moveit_ros/moveit_servo/config/servo_parameters.yaml
```diff
@@ -127,6 +98,16 @@ servo:
     }
 
 
+  incoming_command_timeout: {
+    type: double,
+    read_only: true,
```
sea-bass, line 103: Same here. Seems this is super easy to tune at runtime given it's in a single condition

moveit_ros/moveit_servo/config/servo_parameters.yaml
```diff
@@ -135,18 +116,51 @@ servo:
...
+    description: "The topic to which the status will be published"
   }
 
-  publish_period: {
-    type: double,
-    default_value: 0.034,
-    description: " 1 / (Nominal publish rate) [seconds]",
+  command_out_topic: {
+    type: string,
+    read_only: true,
```
sea-bass, line 151: This one will definitely need to change when we enable changing planning groups. I'm OK if it's read only in this PR but we will very likely need to undo that later.

moveit_ros/moveit_servo/config/servo_parameters.yaml
```diff
@@ -1,5 +1,28 @@
...
+    default_value: 0.034,
+    description: " 1 / (Nominal publish rate) [seconds]",
+    validation: {
+      gt<>: 0.0
+    }
+  }
+
   move_group_name: {
     type: string,
     description: "The name of the moveit move_group of your robot \
```
AndyZe, line 28: Not possible to change at this time

```suggestion
    type: string,
    read_only: true,
    description: "The name of the moveit move_group of your robot \
```

moveit_ros/moveit_servo/config/servo_parameters.yaml
```diff
@@ -191,18 +210,60 @@ servo:
...
+    type: double,
+    default_value: 0.02,
+    description: "Start decelerating when a collision is this far [m]",
+    validation: {
+      gt<>: 0.0
+    }
+  }
+
+  collision_check_rate: {
+    type: double,
```
AndyZe, line 256: If you look at `collision_monitor.cpp`, this is only initialized once. Can't be changed at runtime.

```suggestion
  collision_check_rate: {
    type: double,
    read_only: true,
```

moveit_ros/moveit_servo/config/servo_parameters.yaml
```diff
@@ -191,18 +210,60 @@ servo:
...
+    default_value: 0.01,
+    description: "Start decelerating when a self-collision is this far [m]",
+    validation: {
+      gt<>: 0.0
+    }
+  }
+
+  scene_collision_proximity_threshold: {
+    type: double,
```
AndyZe, line 247: If you look at collision_monitor.cpp, this is involved in variable initialization. Can't be changed at runtime.

```suggestion
    type: double,
    read_only: true,
```

moveit_ros/moveit_servo/config/servo_parameters.yaml
```diff
@@ -191,18 +210,60 @@ servo:
...
+
+  check_collisions: {
     type: bool,
     default_value: true,
-    description: "Halt all joints in cartesian mode"
+    description: "If true, servo will check for collision using the planning scene monitor."
+  }
+
+  self_collision_proximity_threshold: {
+    type: double,
```
AndyZe, line 238: If you look at collision_monitor.cpp, this is involved in variable initialization. Can't be changed at runtime.

```suggestion
    type: double,
    read_only: true,
```

moveit_ros/moveit_servo/config/servo_parameters.yaml
```diff
@@ -235,73 +296,37 @@ servo:
...
-    }
-  }
+############################### JOINT LIMITING #################################
 
-  check_collisions: {
+  halt_all_joints_in_joint_mode: {
     type: bool,
     default_value: true,
-    description: "If true, servo will check for collision using the planning scene monitor."
+    description: "Halt all joints in joint mode"
```
AndyZe, line 310:
```suggestion
    description: "Halt all joints in joint mode, else halt only the joints at their limit"
```

moveit_ros/moveit_servo/config/servo_parameters.yaml
```diff
@@ -235,73 +296,37 @@ servo:
...
-    default_value: 10.0,
-    description: "[Hz] Collision-checking can easily bog down a CPU if done too often. \
-                  Collision checking begins slowing down when nearer than a specified distance.",
-    validation: {
-      gt_eq<>: 0.0
-    }
+  halt_all_joints_in_cartesian_mode: {
+    type: bool,
+    default_value: true,
+    description: "Halt all joints in cartesian mode"
```
AndyZe, line 316:
```suggestion
    description: "Halt all joints in cartesian mode, else halt only the joints at their limit"
```

moveit_ros/moveit_servo/config/servo_parameters.yaml
ibrahiminfinite: Ah yes, this shouldn't be read-only,

moveit_ros/moveit_servo/config/servo_parameters.yaml
ibrahiminfinite: Yeah, will keep it read-only for now.

moveit_ros/moveit_servo/config/servo_parameters.yaml
ibrahiminfinite: Actually, the collision check parameters can be updated, since it is initialized locally in the method called in the thread.
The collision checking thread is stopped when Servo is paused. So it is possible to pause Servo, update the params.
When unpaused, collision checking thread will start again with the new params.

That said, the situation is similar to the one we faced with planning group, we lack a mechanism to properly reject updates at the moment, if Servo is not paused.

moveit_ros/moveit_servo/config/servo_parameters.yaml
```diff
@@ -135,18 +115,51 @@ servo:
...
+  }
+
+  command_out_type: {
+    type: string,
+    read_only: true,
+    default_value: "trajectory_msgs/JointTrajectory",
+    description: "The type of command that servo will publish",
     validation: {
-      gt<>: 0.0
+      one_of<>: [["trajectory_msgs/JointTrajectory", "std_msgs/Float64MultiArray"]]
```
sea-bass, line 162: nice catch here!

moveit_ros/moveit_servo/config/servo_parameters.yaml
sea-bass: Yes and no... there a few lines at the very top of `CollisionMonitor::checkCollisions()` that do not change:
https://github.com/ros-planning/moveit2/blob/main/moveit_ros/moveit_servo/src/collision_monitor.cpp#L90-L91

Whereas these do react live at runtime:
https://github.com/ros-planning/moveit2/blob/main/moveit_ros/moveit_servo/src/collision_monitor.cpp#L132-L134

So I guess we have a choice of either moving the first lines in the loop (keeping the `-log(0.001)` as a constant), or making the whole thing nontunable. I prefer the former!

AndyZe: making them tunable is fine with me. Seems like it shouldn't hurt performance much

ibrahiminfinite: Yeah, will move the two variable initializations into the loop then.
That should make things nice.

ibrahiminfinite: Not sure if the change should go in this PR though since this one is about making params `read-only`

sea-bass: I think it's small enough (and technically a related bug fix) that you can just throw it in here!

sea-bass (approved):

PR #2389, opened 2023-09-26
Make getJacobian simpler and faster

marioprats:
### Description

This change makes `getJacobian` a little bit more readable (hopefully!), while also making it 2x faster, by removing string comparisons where possible and other little optimizations.
The jacobian benchmark shows a ~2x improvement with respect to HEAD (>500ns), and ~4x with respect to the KDL Jacobian, for the panda robot:

```
------------------------------------------------------------
Benchmark                  Time             CPU   Iterations
------------------------------------------------------------
BM_MoveItJacobian        248 ns          249 ns      2867707
BM_KDLJacobian          1099 ns         1100 ns       634762
```

The recently added Jacobian tests for the different types of joints are also passing.
Also manually tested this with Servo, which I know uses the MoveIt! Jacobian.

Note: `a_pose_b` is a notation I have used in the past to denote a transformation matrix to convert points in `b` to points in `a`, e.g:  `point_a = a_pose_b * point_b`. Not sure if there's another established notation if MoveIt!. Happy to change to whatever makes more sense.

### Checklist

### DIFF
--- a/moveit_core/robot_state/src/robot_state.cpp
+++ b/moveit_core/robot_state/src/robot_state.cpp
@@ -851,7 +851,7 @@ void RobotState::updateStateWithLinkAt(const LinkModel* link, const Eigen::Isome
 
 const LinkModel* RobotState::getRigidlyConnectedParentLinkModel(const std::string& frame) const
 {
-  const moveit::core::LinkModel* link{ nullptr };
+  const LinkModel* link{ nullptr };
 
   size_t idx = 0;
   if ((idx = frame.find('/')) != std::string::npos)
@@ -1005,7 +1005,7 @@ double RobotState::distance(const RobotState& other, const JointModelGroup* join
 
 void RobotState::interpolate(const RobotState& to, double t, RobotState& state) const
 {
-  moveit::core::checkInterpolationParamBounds(LOGGER, t);
+  checkInterpolationParamBounds(LOGGER, t);
   robot_model_->interpolate(getVariablePositions(), to.getVariablePositions(), t, state.getVariablePositions());
 
   memset(state.dirty_joint_transforms_, 1, state.robot_model_->getJointModelCount() * sizeof(unsigned char));
@@ -1014,7 +1014,7 @@ void RobotState::interpolate(const RobotState& to, double t, RobotState& state) 
 
 void RobotState::interpolate(const RobotState& to, double t, RobotState& state, const JointModelGroup* joint_group) const
 {
-  moveit::core::checkInterpolationParamBounds(LOGGER, t);
+  checkInterpolationParamBounds(LOGGER, t);
   const std::vector<const JointModel*>& jm = joint_group->getActiveJointModels();
   for (const JointModel* joint : jm)
   {
@@ -1066,7 +1066,7 @@ void RobotState::attachBody(const std::string& id, const Eigen::Isometry3d& pose
                             const std::vector<shapes::ShapeConstPtr>& shapes,
                             const EigenSTL::vector_Isometry3d& shape_poses, const std::set<std::string>& touch_links,
                             const std::string& link, const trajectory_msgs::msg::JointTrajectory& detach_posture,
-                            const moveit::core::FixedTransformsMap& subframe_poses)
+                            const FixedTransformsMap& subframe_poses)
 {
   attachBody(std::make_unique<AttachedBody>(robot_model_->getLinkModel(link), id, pose, shapes, shape_poses,
                                             touch_links, detach_posture, subframe_poses));
@@ -1349,11 +1349,9 @@ bool RobotState::getJacobian(const JointModelGroup* group, const LinkModel* link
                              const Eigen::Vector3d& reference_point_position, Eigen::MatrixXd& jacobian,
                              bool use_quaternion_representation) const
 {
-  using LinkModel = moveit::core::LinkModel;
-  using JointModel = moveit::core::JointModel;
-
   assert(checkLinkTransforms());
 
+  // Check that the group is a chain, contains 'link' and has joint models.
   if (!group->isChain())
   {
     RCLCPP_ERROR(LOGGER, "The group '%s' is not a chain. Cannot compute Jacobian.", group->getName().c_str());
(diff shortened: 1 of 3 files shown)


moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -1357,91 +1357,78 @@ bool RobotState::getJacobian(const JointModelGroup* group, const LinkModel* link
...
 
-  Eigen::Vector3d joint_axis;
-  Eigen::Isometry3d joint_transform;
+  Eigen::Isometry3d root_pose_tip = root_pose_world * getGlobalLinkTransform(link);
+  Eigen::Vector3d tip_point = root_pose_tip * reference_point_position;
 
-  while (link)
+  std::size_t active_joints = group->getActiveJointModels().size();
+  int i = 0;
+  for (std::size_t joint = 0; joint < active_joints; joint++)
```
sea-bass, line 1389:
```suggestion
  for (std::size_t joint = 0; joint < active_joints; ++joint)

File: moveit_core/robot_state/src/robot_state.cpp [hunk 2]
```diff
@@ -1357,91 +1357,78 @@ bool RobotState::getJacobian(const JointModelGroup* group, const LinkModel* link
     return false;
   }
 
-  if (!group->isLinkUpdated(link->getName()))
+  const std::set<const moveit::core::LinkModel*>& descendant_links = group->getUpdatedLinkModelsSet();
```
sea-bass, line 1360: Is it worth creating an alias for `std::set<const moveit::core::LinkModel*>` for readability?

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -1357,91 +1357,78 @@ bool RobotState::getJacobian(const JointModelGroup* group, const LinkModel* link
...
+  }
+  const moveit::core::JointModel* root_joint_model = group->getJointModels().front();
   const moveit::core::LinkModel* root_link_model = root_joint_model->getParentLinkModel();
-  // getGlobalLinkTransform() returns a valid isometry by contract
-  Eigen::Isometry3d reference_transform =
+
+  Eigen::Isometry3d root_pose_world =
       root_link_model ? getGlobalLinkTransform(root_link_model).inverse() : Eigen::Isometry3d::Identity();
   int rows = use_quaternion_representation ? 7 : 6;
   int columns = group->getVariableCount();
```
sea-bass, line 1381: not your doing, but I think these can be `const`?

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -1357,91 +1357,78 @@ bool RobotState::getJacobian(const JointModelGroup* group, const LinkModel* link
...
-  // RCLCPP_DEBUG(LOGGER, "Point from reference origin expressed in world coordinates: %f %f %f",
-  //         point_transform.x(),
-  //         point_transform.y(),
-  //         point_transform.z());
+  jacobian.resize(rows, columns);
 
-  Eigen::Vector3d joint_axis;
-  Eigen::Isometry3d joint_transform;
+  Eigen::Isometry3d root_pose_tip = root_pose_world * getGlobalLinkTransform(link);
+  Eigen::Vector3d tip_point = root_pose_tip * reference_point_position;
```
sea-bass, line 1385: I also think these can be `const`

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -1357,91 +1357,78 @@ bool RobotState::getJacobian(const JointModelGroup* group, const LinkModel* link
...
     RCLCPP_ERROR(LOGGER, "Link name '%s' does not exist in the chain '%s' or is not a child for this chain",
                  link->getName().c_str(), group->getName().c_str());
     return false;
   }
 
-  const moveit::core::JointModel* root_joint_model = group->getJointModels()[0];  // group->getJointRoots()[0];
+  const std::vector<const moveit::core::JointModel*>& joint_models = group->getActiveJointModels();
+  if (joint_models.empty())
+  {
+    RCLCPP_ERROR(LOGGER, "The group '%s' doesn't contain any JointModel's. Cannot compute Jacobian.",
```
sea-bass, line 1371:
```suggestion
    RCLCPP_ERROR(LOGGER, "The group '%s' doesn't contain any joint models. Cannot compute Jacobian.",

File: moveit_core/robot_state/src/robot_state.cpp [hunk 6]
```diff
@@ -1357,91 +1357,78 @@ bool RobotState::getJacobian(const JointModelGroup* group, const LinkModel* link
...
+                 group->getName().c_str());
+    return false;
+  }
+  const moveit::core::JointModel* root_joint_model = group->getJointModels().front();
   const moveit::core::LinkModel* root_link_model = root_joint_model->getParentLinkModel();
-  // getGlobalLinkTransform() returns a valid isometry by contract
-  Eigen::Isometry3d reference_transform =
+
+  Eigen::Isometry3d root_pose_world =
```
sea-bass, line 1378: this can also be const?

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -1357,91 +1357,78 @@ bool RobotState::getJacobian(const JointModelGroup* group, const LinkModel* link
...
-        jacobian.block<3, 1>(0, joint_index + 1) = jacobian.block<3, 1>(0, joint_index + 1) + joint_axis;
-        joint_axis = joint_transform * Eigen::Vector3d(0.0, 0.0, 1.0);
-        jacobian.block<3, 1>(0, joint_index + 2) = jacobian.block<3, 1>(0, joint_index + 2) +
-                                                   joint_axis.cross(point_transform - joint_transform.translation());
-        jacobian.block<3, 1>(3, joint_index + 2) = jacobian.block<3, 1>(3, joint_index + 2) + joint_axis;
-      }
-      else
-        RCLCPP_ERROR(LOGGER, "Unknown type of joint in Jacobian computation");
+      Eigen::Vector3d axis_wrt_origin =
+          origin_pose_link.linear() * static_cast<const moveit::core::RevoluteJointModel*>(joint_model)->getAxis();
```
sea-bass, line 1398: const?

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -1357,91 +1357,78 @@ bool RobotState::getJacobian(const JointModelGroup* group, const LinkModel* link
...
+      jacobian.block<3, 1>(0, i) = axis_wrt_origin.cross(tip_point - origin_pose_link.translation());
+      jacobian.block<3, 1>(3, i) = axis_wrt_origin;
+    }
+    else if (joint_model->getType() == moveit::core::JointModel::PRISMATIC)
+    {
+      Eigen::Vector3d axis_wrt_origin =
+          origin_pose_link.linear() * static_cast<const moveit::core::PrismaticJointModel*>(joint_model)->getAxis();
```
sea-bass, line 1405: const?

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -1357,91 +1357,78 @@ bool RobotState::getJacobian(const JointModelGroup* group, const LinkModel* link
...
+    {
+      jacobian.block<3, 1>(0, i) = origin_pose_link.linear() * Eigen::Vector3d(1.0, 0.0, 0.0);
+      jacobian.block<3, 1>(0, i + 1) = origin_pose_link.linear() * Eigen::Vector3d(0.0, 1.0, 0.0);
+      jacobian.block<3, 1>(0, i + 2) =
+          (origin_pose_link.linear() * Eigen::Vector3d(0.0, 0.0, 1.0)).cross(tip_point - origin_pose_link.translation());
+      jacobian.block<3, 1>(3, i + 2) = origin_pose_link.linear() * Eigen::Vector3d(0.0, 0.0, 1.0);
+    }
+    else
+    {
+      RCLCPP_ERROR(LOGGER, "Unknown type of joint in Jacobian computation");
```
sea-bass, line 1419: Should we return `false` in here instead of proceeding?

sea-bass (commented):
Very cool!

moveit_core/robot_state/src/robot_state.cpp
marioprats: created a small local alias for moveit::core, so it reads:
`std::set<const LinkModel*>`

I think it's good to see the explicit type here, otherwise could just use `auto`.

moveit_core/robot_state/src/robot_state.cpp
marioprats: Sorry for all the missing const!
I'm not used to use const on local variables, but I see this is encouraged at PickNik, from this review and others. I'll start using it.

moveit_core/robot_state/src/robot_state.cpp
marioprats: Yes, I think so. Changed.

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -1349,6 +1349,9 @@ bool RobotState::getJacobian(const JointModelGroup* group, const LinkModel* link
                              const Eigen::Vector3d& reference_point_position, Eigen::MatrixXd& jacobian,
                              bool use_quaternion_representation) const
 {
+  using LinkModel = moveit::core::LinkModel;
```
sjahr, line 1352: Is this necessary? I think this already is in the moveit core namespace?

sjahr (commented):
Do you mind adding some comments in front of the main blocks to explain what is happening there? I think that will help maintaining this in the future

moveit_core/robot_state/src/robot_state.cpp
marioprats: ah, good catch!
removed the unnecessary namespace qualification here and in other places in this file.

marioprats:
Thanks for the reviews.
While going again through the code adding comments I realized of an issue when the joint group origin doesn't match the URDF origin. Added a test for it and should be ok now.

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -1351,97 +1351,93 @@ bool RobotState::getJacobian(const JointModelGroup* group, const LinkModel* link
...
-    //         link_state->getGlobalLinkTransform().translation().z());
-    // RCLCPP_DEBUG(LOGGER, "Joint: %s",link_state->getParentJointState()->getName().c_str());
+    RCLCPP_ERROR(LOGGER, "The group '%s' doesn't contain any joint models. Cannot compute Jacobian.",
+                 group->getName().c_str());
+    return false;
+  }
 
-    const JointModel* pjm = link->getParentJointModel();
-    if (pjm->getVariableCount() > 0)
+  // Get the link model of the group root link, and it's inverted pose with respect to the RobotModel (URDF) root,
```
sea-bass, line 1377:
```suggestion
  // Get the link model of the group root link, and its inverted pose with respect to the RobotModel (URDF) root,

sea-bass (approved):

sjahr (approved):

PR #2396, opened 2023-09-29
Enable using a subgroup of the move group in servo

sjahr:
### Description

This PR enables using a subgroup of the move_group at runtime without the need to stop servo or switch controllers.

- [x] Make subgroup -> move group joint index mapping more efficient
- [x] Disable joint jogging for joints that aren't part of active subgroup

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/config/servo_parameters.yaml
+++ b/moveit_ros/moveit_servo/config/servo_parameters.yaml
@@ -49,7 +49,7 @@ servo:
   active_subgroup: {
     type: string,
     default_value: "",
-    description: "This parameter can be used to switch online to actuating a sub-joint group of the move group with servo. \
+    description: "This parameter can be used to switch online to actuating a subgroup of the move group. \
     If it is empty, the full move group is actuated."
   }
 
--- a/moveit_ros/moveit_servo/config/servo_parameters.yaml
+++ b/moveit_ros/moveit_servo/config/servo_parameters.yaml
@@ -46,8 +46,14 @@ servo:
                   must be passed to the node during launch time."
   }
 
-############################# INCOMING COMMAND SETTINGS ########################
+  active_subgroup: {
+    type: string,
+    default_value: "",
+    description: "This parameter can be used to switch online to actuating a sub-joint group of the move group with servo. \
+    If it is empty, the full move group is actuated."
+  }
 
+############################# INCOMING COMMAND SETTINGS ########################
   pose_command_in_topic: {
     type: string,
     read_only: true,
--- a/moveit_ros/moveit_servo/include/moveit_servo/utils/command.hpp
+++ b/moveit_ros/moveit_servo/include/moveit_servo/utils/command.hpp
@@ -55,7 +55,7 @@ namespace moveit_servo
  * @param command The joint jog command.
  * @param robot_state_ The current robot state as obtained from PlanningSceneMonitor.
  * @param servo_params The servo parameters.
- * @param joint_name_group_index_map Mapping between sub group joint name and move group joint vector position
+ * @param joint_name_group_index_map Mapping between joint subgroup name and move group joint vector position.
  * @return The status and joint position change required (delta).
  */
 JointDeltaResult jointDeltaFromJointJog(const JointJogCommand& command, const moveit::core::RobotStatePtr& robot_state,
@@ -67,7 +67,7 @@ JointDeltaResult jointDeltaFromJointJog(const JointJogCommand& command, const mo
  * @param command The twist command.
  * @param robot_state_ The current robot state as obtained from PlanningSceneMonitor.
  * @param servo_params The servo parameters.
- * @param joint_name_group_index_map Mapping between sub group joint name and move group joint vector position
+ * @param joint_name_group_index_map Mapping between joint subgroup name and move group joint vector position.
  * @return The status and joint position change required (delta).
  */
 JointDeltaResult jointDeltaFromTwist(const TwistCommand& command, const moveit::core::RobotStatePtr& robot_state,
@@ -91,7 +91,7 @@ JointDeltaResult jointDeltaFromPose(const PoseCommand& command, const moveit::co
  * @param cartesian_position_delta The change in Cartesian position.
  * @param robot_state_ The current robot state as obtained from PlanningSceneMonitor.
  * @param servo_params The servo parameters.
- * @param joint_name_group_index_map Mapping between sub group joint name and move group joint vector position
+ * @param joint_name_group_index_map Mapping between joint subgroup name and move group joint vector position.
  * @return The status and joint position change required (delta).
  */
 JointDeltaResult jointDeltaFromIK(const Eigen::VectorXd& cartesian_position_delta,
--- a/moveit_ros/moveit_servo/include/moveit_servo/utils/command.hpp
+++ b/moveit_ros/moveit_servo/include/moveit_servo/utils/command.hpp
@@ -55,39 +55,47 @@ namespace moveit_servo
  * @param command The joint jog command.
  * @param robot_state_ The current robot state as obtained from PlanningSceneMonitor.
  * @param servo_params The servo parameters.
+ * @param joint_name_group_index_map Mapping between sub group joint name and move group joint vector position
  * @return The status and joint position change required (delta).
  */
 JointDeltaResult jointDeltaFromJointJog(const JointJogCommand& command, const moveit::core::RobotStatePtr& robot_state,
-                                        const servo::Params& servo_params);
+                                        const servo::Params& servo_params,
+                                        const JointNameToMoveGroupIndexMap& joint_name_group_index_map);
 
 /**
  * \brief Compute the change in joint position for the given twist command.
  * @param command The twist command.
  * @param robot_state_ The current robot state as obtained from PlanningSceneMonitor.
  * @param servo_params The servo parameters.
+ * @param joint_name_group_index_map Mapping between sub group joint name and move group joint vector position
  * @return The status and joint position change required (delta).
  */
 JointDeltaResult jointDeltaFromTwist(const TwistCommand& command, const moveit::core::RobotStatePtr& robot_state,
-                                     const servo::Params& servo_params);
+                                     const servo::Params& servo_params,
+                                     const JointNameToMoveGroupIndexMap& joint_name_group_index_map);
 
 /**
  * \brief Compute the change in joint position for the given pose command.
  * @param command The pose command.
  * @param robot_state_ The current robot state as obtained from PlanningSceneMonitor.
  * @param servo_params The servo parameters.
+ * @param joint_name_group_index_map Mapping between sub group joint name and move group joint vector position
  * @return The status and joint position change required (delta).
  */
 JointDeltaResult jointDeltaFromPose(const PoseCommand& command, const moveit::core::RobotStatePtr& robot_state,
-                                    const servo::Params& servo_params);
+                                    const servo::Params& servo_params,
+                                    const JointNameToMoveGroupIndexMap& joint_name_group_index_map);
 
 /**
  * \brief Computes the required change in joint angles for given Cartesian change, using the robot's IK solver.
  * @param cartesian_position_delta The change in Cartesian position.
  * @param robot_state_ The current robot state as obtained from PlanningSceneMonitor.
  * @param servo_params The servo parameters.
+ * @param joint_name_group_index_map Mapping between sub group joint name and move group joint vector position
  * @return The status and joint position change required (delta).
  */
 JointDeltaResult jointDeltaFromIK(const Eigen::VectorXd& cartesian_position_delta,
-                                  const moveit::core::RobotStatePtr& robot_state, const servo::Params& servo_params);
+                                  const moveit::core::RobotStatePtr& robot_state, const servo::Params& servo_params,
+                                  const JointNameToMoveGroupIndexMap& joint_name_group_index_map);
 
 }  // namespace moveit_servo
--- a/moveit_ros/moveit_servo/src/servo.cpp
+++ b/moveit_ros/moveit_servo/src/servo.cpp
@@ -229,6 +229,33 @@ bool Servo::updateParams()
                                        << std::to_string(params.override_velocity_scaling_factor));
       }
 
+      // If necessary, create new subgroup map
+      if (!params.active_subgroup.empty() && params.active_subgroup != params.move_group_name &&
+          joint_name_to_index_maps_.count(params.active_subgroup) == 0)
+      {
+        const auto& move_group_joint_names = planning_scene_monitor_->getRobotModel()
+                                                 ->getJointModelGroup(params.move_group_name)
+                                                 ->getActiveJointModelNames();
+        const auto& subgroup_joint_names = planning_scene_monitor_->getRobotModel()
+                                               ->getJointModelGroup(params.active_subgroup)
+                                               ->getActiveJointModelNames();
+
+        JointNameToMoveGroupIndexMap new_map;
+        // For each joint name of the subgroup calculate the index in the move group joint vector
+        for (const auto& joint_name : subgroup_joint_names)
+        {
+          // Find sub group joint name in move group joint names
+          const auto move_group_iterator =
+              std::find(move_group_joint_names.cbegin(), move_group_joint_names.cend(), joint_name);
+          // Calculate position and add a new mapping of joint name to move group joint vector position
+          new_map.insert(std::make_pair<std::string, std::size_t>(
+              std::string(joint_name), std::distance(move_group_joint_names.cbegin(), move_group_iterator)));
+        }
+        // Add new joint name to index map to existing maps
+        joint_name_to_index_maps_.insert(std::make_pair<std::string, JointNameToMoveGroupIndexMap>(
+            std::string(params.active_subgroup), std::move(new_map)));
+      }
+
       servo_params_ = params;
       return true;
     }
@@ -309,6 +336,12 @@ KinematicState Servo::haltJoints(const std::vector<int>& joints_to_halt, const K
 
 Eigen::VectorXd Servo::jointDeltaFromCommand(const ServoInput& command, const moveit::core::RobotStatePtr& robot_state)
 {
+  // Determine joint_name_group_index_map, if no subgroup is active, the map is empty
+  const auto& joint_name_group_index_map =
+      (!servo_params_.active_subgroup.empty() && servo_params_.active_subgroup != servo_params_.move_group_name) ?
+          joint_name_to_index_maps_.at(servo_params_.active_subgroup) :
+          JointNameToMoveGroupIndexMap();
+
   const int num_joints =
       robot_state->getJointModelGroup(servo_params_.move_group_name)->getActiveJointModelNames().size();
   Eigen::VectorXd joint_position_deltas(num_joints);
@@ -321,7 +354,8 @@ Eigen::VectorXd Servo::jointDeltaFromCommand(const ServoInput& command, const mo
   {
     if (expected_type == CommandType::JOINT_JOG)
     {
-      delta_result = jointDeltaFromJointJog(std::get<JointJogCommand>(command), robot_state, servo_params_);
+      delta_result = jointDeltaFromJointJog(std::get<JointJogCommand>(command), robot_state, servo_params_,
+                                            joint_name_group_index_map);
       servo_status_ = delta_result.first;
     }
     else if (expected_type == CommandType::TWIST)
@@ -329,7 +363,8 @@ Eigen::VectorXd Servo::jointDeltaFromCommand(const ServoInput& command, const mo
       try
       {
         const TwistCommand command_in_planning_frame = toPlanningFrame(std::get<TwistCommand>(command));
-        delta_result = jointDeltaFromTwist(command_in_planning_frame, robot_state, servo_params_);
+        delta_result =
+            jointDeltaFromTwist(command_in_planning_frame, robot_state, servo_params_, joint_name_group_index_map);
         servo_status_ = delta_result.first;
       }
       catch (tf2::TransformException& ex)
@@ -343,7 +378,8 @@ Eigen::VectorXd Servo::jointDeltaFromCommand(const ServoInput& command, const mo
       try
       {
         const PoseCommand command_in_planning_frame = toPlanningFrame(std::get<PoseCommand>(command));
-        delta_result = jointDeltaFromPose(command_in_planning_frame, robot_state, servo_params_);
+        delta_result =
+            jointDeltaFromPose(command_in_planning_frame, robot_state, servo_params_, joint_name_group_index_map);
         servo_status_ = delta_result.first;
       }
       catch (tf2::TransformException& ex)
--- a/moveit_ros/moveit_servo/src/servo.cpp
+++ b/moveit_ros/moveit_servo/src/servo.cpp
@@ -248,7 +248,6 @@ bool Servo::validateParams(const servo::Params& servo_params) const
 bool Servo::updateParams()
 {
   bool params_updated = false;
-  
   if (servo_param_listener_->is_old(servo_params_))
   {
     const auto params = servo_param_listener_->get_params();
--- a/moveit_ros/moveit_servo/src/servo.cpp
+++ b/moveit_ros/moveit_servo/src/servo.cpp
@@ -159,12 +159,13 @@ void Servo::setCollisionChecking(const bool check_collision)
 
 bool Servo::validateParams(const servo::Params& servo_params) const
 {
+  bool params_valid = true;
   auto robot_state = planning_scene_monitor_->getStateMonitor()->getCurrentState();
   auto joint_model_group = robot_state->getJointModelGroup(servo_params.move_group_name);
   if (joint_model_group == nullptr)
   {
     RCLCPP_ERROR_STREAM(LOGGER, "Invalid move group name: `" << servo_params.move_group_name << '`');
-    return false;
+    params_valid = false;
   }
 
   if (servo_params.hard_stop_singularity_threshold <= servo_params.lower_singularity_threshold)
@@ -172,7 +173,7 @@ bool Servo::validateParams(const servo::Params& servo_params) const
     RCLCPP_ERROR(LOGGER, "Parameter 'hard_stop_singularity_threshold' "
                          "should be greater than 'lower_singularity_threshold.' "
                          "Check the parameters YAML file used to launch this node.");
-    return false;
+    params_valid = false;
   }
 
   if (!servo_params.publish_joint_positions && !servo_params.publish_joint_velocities &&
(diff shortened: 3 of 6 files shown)


moveit_ros/moveit_servo/config/servo_parameters.yaml
```diff
@@ -46,8 +46,14 @@ servo:
                   must be passed to the node during launch time."
   }
 
-############################# INCOMING COMMAND SETTINGS ########################
+  active_subgroup: {
+    type: string,
+    default_value: "",
+    description: "This parameter can be used to switch online to actuating a sub-joint group of the move group with servo. \
```
sea-bass, line 52:
```suggestion
description: "This parameter can be used to switch online to actuating a subgroup of the move group. \

File: moveit_ros/moveit_servo/src/utils/command.cpp [hunk 2]
```diff
@@ -272,6 +273,24 @@ JointDeltaResult jointDeltaFromIK(const Eigen::VectorXd& cartesian_position_delt
     delta_theta = pseudo_inverse * cartesian_position_delta;
   }
 
+  if (!servo_params.active_subgroup.empty())
+  {
+    // Create full delta vector and add delta theta's at actuated joints
```
sea-bass, line 278:
```suggestion
    // Create full delta vector and add delta theta values only at actuated joints

File: moveit_ros/moveit_servo/src/utils/command.cpp [hunk 3]
```diff
@@ -272,6 +273,24 @@ JointDeltaResult jointDeltaFromIK(const Eigen::VectorXd& cartesian_position_delt
...
+        robot_state->getJointModelGroup(servo_params.move_group_name)->getActiveJointModelNames();
+    const auto& subgroup_joint_names =
+        robot_state->getJointModelGroup(servo_params.active_subgroup)->getActiveJointModelNames();
+
+    Eigen::VectorXd move_group_delta_theta = Eigen::VectorXd::Zero(move_group_joint_names.size());
+    for (size_t index = 0; index < subgroup_joint_names.size(); index++)
+    {
+      const auto move_group_iterator =
+          find(move_group_joint_names.cbegin(), move_group_joint_names.cend(), subgroup_joint_names.at(index));
+      move_group_delta_theta[std::distance(move_group_joint_names.cbegin(), move_group_iterator)] = delta_theta[index];
```
sea-bass, line 289: I can see this adding a lot of slowness to the regular running loop. I wonder if we should instead build an unordered map either at initialization (go through all the subgroups?) and/or at parameter switching time so it could let us look up indices in O(1) time in the loop?

... or if there is otherwise already another way of doing this with `RobotModel`?

sea-bass (commented):
Thanks for getting on this! I think your idea of having a fixed move group and only allowing the changing of subgroups has reduced scope enough to be useful and also not overly complicated.

I have one meaningful inline comment about searching for indices (see below).

My other main comment is, I know @AndyZe, @ibrahiminfinite, and I were discussing the risks of immediately switching planning groups and having joints added/removed be commanded with sudden velocity changes. I think because of the subgroup assumption, this may be OK, but it's worth thinking about it: if, e.g., a joint was moving with a nonzero velocity and we suddenly change to a subgroup without the joint, will this cause any issue or will it slow down the joint using the scalings provided?

moveit_ros/moveit_servo/src/utils/command.cpp
ibrahiminfinite: > at parameter switching time 

This option looks better since we are in a free function.
If we create map at initialization, it would likely need to be passed through multiple functions.

moveit_ros/moveit_servo/src/utils/command.cpp
```diff
@@ -272,6 +273,24 @@ JointDeltaResult jointDeltaFromIK(const Eigen::VectorXd& cartesian_position_delt
...
 
+  if (!servo_params.active_subgroup.empty())
+  {
+    // Create full delta vector and add delta theta's at actuated joints
+    const auto& move_group_joint_names =
+        robot_state->getJointModelGroup(servo_params.move_group_name)->getActiveJointModelNames();
+    const auto& subgroup_joint_names =
+        robot_state->getJointModelGroup(servo_params.active_subgroup)->getActiveJointModelNames();
+
+    Eigen::VectorXd move_group_delta_theta = Eigen::VectorXd::Zero(move_group_joint_names.size());
```
ibrahiminfinite, line 284: Instead of creating a new vector here, can we just set the required values in `delta_theta` to zero ?

sjahr: Sadly no, because they have different sizes.
move_group_delta_theta.size() = Total number of joints of the whole move group
delta_theta.size() = Number of joints in the subgroup

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -160,22 +159,20 @@ void Servo::setCollisionChecking(const bool check_collision)
...
 {
-  bool params_valid = true;
-
   auto robot_state = planning_scene_monitor_->getStateMonitor()->getCurrentState();
   auto joint_model_group = robot_state->getJointModelGroup(servo_params.move_group_name);
   if (joint_model_group == nullptr)
   {
     RCLCPP_ERROR_STREAM(LOGGER, "Invalid move group name: `" << servo_params.move_group_name << '`');
-    params_valid = false;
+    return false;
```
AndyZe, line 167: Just to offer an explanation for why it was like this, Ibrahim was told that a function should only return in one place. I kind of like this new style a bit better myself. Context-dependent, I guess.

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -194,51 +191,52 @@ bool Servo::validateParams(const servo::Params& servo_params) const
...
+                 "The value '%s' Parameter 'active_subgroup' does not name a valid subgroup of joint group '%s'.",
+                 servo_params.active_subgroup.c_str(), servo_params.move_group_name.c_str());
+    return false;
   }
 
-  return params_valid;
+  return true;
 }
 
 bool Servo::updateParams()
```
AndyZe, line 217: Need to make sure Servo is paused before doing this for a couple reasons:

- Don't want to have these operations in the control loop
- I think it's good from a safety perspective to stop the other joints beside this subgroup. Pausing should do that

AndyZe (commented):
Is there an urgent need for this feature? Seems like a bit of a stretch.

What happens to the other joints when you start actuating the subgroup only? Are they still filtered?

AndyZe:
Does this change the dimension of the final command to ros2_control? That would be another issue

ibrahiminfinite:
> Does this change the dimension of the final command to ros2_control? That would be another issue when the controller topic type is `std_msgs/Float64MultiArray` (It's just a dumb array without joint names.)

From what I understand, the new logic does not change the dimension of the final command, it just sets the `delta_theta` for joints not in the desired `sub_group` to zero.

moveit_ros/moveit_servo/src/servo.cpp
sjahr: I just liked that better, but if that was a discussed before I'll revert that change :+1:

sjahr:
@AndyZe my changes are meant to enable servoing only a subgroup at runtime without the need to pause servo and/or switch controllers. If a subgroup is enabled I only use this subgroup to calculate the delta_thetas and command a 0 delta theta to the other joints of the whole move_group. From a robot perspective nothing should change except that some joints (the ones that aren't part of the subgroup) get commands to hold their current position.

moveit_ros/moveit_servo/src/utils/command.cpp
sjahr: I moved this find operation into the param update function so it is only called once and only the first time a group is added. We could even think about moving it into the constructor :thinking: and just create a map for all existing joint groups

sjahr: Done :heavy_check_mark:

sjahr:
@ibrahiminfinite Any clue why the tests are failing?

```
    [ros2_control_node-1] terminate called after throwing an instance of 'std::runtime_error'
    [ros2_control_node-1]   what():  can't compare times with different time sources
    [ros2_control_node-1] Stack trace (most recent call last) in thread 23189:
    [ros2_control_node-1] #13   Object "", at 0xffffffffffffffff, in 
    [ros2_control_node-1] #12   Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7fde71106bf3, in __clone
    [ros2_control_node-1] #11   Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7fde71075ac2, in 
    [ros2_control_node-1] #10   Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30", at 0x7fde71307252, in 
    [spawner-2] [INFO] [1696511567.840236022] [spawner_joint_state_broadcaster]: [NON-XML-CHAR-0x1B][92mConfigured and activated [NON-XML-CHAR-0x1B][1mjoint_state_broadcaster[NON-XML-CHAR-0x1B][0m
    [ros2_control_node-1] #9    Object "/opt/ros/rolling/lib/controller_manager/ros2_control_node", at 0x5614c0025a33, in 
    [ros2_control_node-1] #8    Object "/opt/ros/rolling/lib/libcontroller_manager.so", at 0x7fde7176af3c, in controller_manager::ControllerManager::update(rclcpp::Time const&, rclcpp::Duration const&)
    [ros2_control_node-1] #7    Object "/opt/ros/rolling/lib/librclcpp.so", at 0x7fde7155bb09, in 
    [ros2_control_node-1] #6    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30", at 0x7fde712d94d7, in __cxa_throw
    [ros2_control_node-1] #5    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30", at 0x7fde712d9276, in std::terminate()
    [ros2_control_node-1] #4    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30", at 0x7fde712d920b, in 
    [ros2_control_node-1] #3    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30", at 0x7fde712cdb9d, in 
    [ros2_control_node-1] #2    Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7fde710097f2, in abort
    [ros2_control_node-1] #1    Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7fde71023475, in raise
    [ros2_control_node-1] #0    Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7fde710779fc, in pthread_kill
    [ros2_control_node-1] Aborted (Signal sent by tkill() 23157 0)
Error: ROR] [ros2_control_node-1]: process has died [pid 23157, exit code -6, cmd '/opt/ros/rolling/lib/controller_manager/ros2_control_node --ros-args --params-file /tmp/launch_params_uphjjf6h --params-file /home/runner/work/moveit2/moveit2/.work/upstream_ws/install/moveit_resources_panda_moveit_config/share/moveit_resources_panda_moveit_config/config/ros2_controllers.yaml'].
```

moveit_ros/moveit_servo/include/moveit_servo/utils/command.hpp
```diff
@@ -55,39 +55,47 @@ namespace moveit_servo
  * @param command The joint jog command.
  * @param robot_state_ The current robot state as obtained from PlanningSceneMonitor.
  * @param servo_params The servo parameters.
+ * @param joint_name_group_index_map Mapping between sub group joint name and move group joint vector position
```
sea-bass, line 58:
```suggestion
 * @param joint_name_group_index_map Mapping between joint subgroup name and move group joint vector position.

File: moveit_ros/moveit_servo/include/moveit_servo/utils/command.hpp [hunk 8]
```diff
@@ -55,39 +55,47 @@ namespace moveit_servo
...
-                                        const servo::Params& servo_params);
+                                        const servo::Params& servo_params,
+                                        const JointNameToMoveGroupIndexMap& joint_name_group_index_map);
 
 /**
  * \brief Compute the change in joint position for the given twist command.
  * @param command The twist command.
  * @param robot_state_ The current robot state as obtained from PlanningSceneMonitor.
  * @param servo_params The servo parameters.
+ * @param joint_name_group_index_map Mapping between sub group joint name and move group joint vector position
```
sea-bass, line 70:
```suggestion
 * @param joint_name_group_index_map Mapping between joint subgroup name and move group joint vector position.

File: moveit_ros/moveit_servo/include/moveit_servo/utils/command.hpp [hunk 9]
```diff
@@ -55,39 +55,47 @@ namespace moveit_servo
...
-                                    const servo::Params& servo_params);
+                                    const servo::Params& servo_params,
+                                    const JointNameToMoveGroupIndexMap& joint_name_group_index_map);
 
 /**
  * \brief Computes the required change in joint angles for given Cartesian change, using the robot's IK solver.
  * @param cartesian_position_delta The change in Cartesian position.
  * @param robot_state_ The current robot state as obtained from PlanningSceneMonitor.
  * @param servo_params The servo parameters.
+ * @param joint_name_group_index_map Mapping between sub group joint name and move group joint vector position
```
sea-bass, line 94:
```suggestion
 * @param joint_name_group_index_map Mapping between joint subgroup name and move group joint vector position.

File: moveit_ros/moveit_servo/src/servo.cpp [hunk 10]
```diff
@@ -117,8 +116,38 @@ Servo::Servo(const rclcpp::Node::SharedPtr& node, std::shared_ptr<const servo::P
...
+      continue;
+    }
+    const auto& subgroup_joint_names =
+        planning_scene_monitor_->getRobotModel()->getJointModelGroup(sub_group_name)->getActiveJointModelNames();
+
+    JointNameToMoveGroupIndexMap new_map;
+    // For each joint name of the subgroup calculate the index in the move group joint vector
+    for (const auto& joint_name : subgroup_joint_names)
+    {
+      // Find sub group joint name in move group joint names
```
sea-bass, line 139:
```suggestion
      // Find subgroup joint name in move group joint names

File: moveit_ros/moveit_servo/src/servo.cpp [hunk 11]
```diff
@@ -205,40 +233,41 @@ bool Servo::validateParams(const servo::Params& servo_params) const
...
       }
-      if (params.move_group_name != servo_params_.move_group_name)
-      {
-        RCLCPP_INFO_STREAM(LOGGER, "Move group changed from " << servo_params_.move_group_name << " to "
-                                                              << params.move_group_name);
-      }
 
       servo_params_ = params;
-      params_updated = true;
+      return true;
```
sea-bass, line 263: Per the other recommendation to go back to the former style of returning just at the end.

```suggestion
      params_updated = true;

File: moveit_ros/moveit_servo/src/servo.cpp [hunk 12]
```diff
@@ -205,40 +233,41 @@ bool Servo::validateParams(const servo::Params& servo_params) const
...
     }
     else
     {
       RCLCPP_WARN_STREAM(LOGGER, "Parameters will not be updated.");
     }
   }
-  return params_updated;
+  return false;
```
sea-bass, line 270:
```suggestion
  return params_updated;

File: moveit_ros/moveit_servo/src/utils/command.cpp [hunk 13]
```diff
@@ -42,17 +42,49 @@
...
-}
+
+/**
+ * @brief Helper function to create a move group deltas vector from a sub group deltas vector. A delta vector for the
+ * whole move group is created and all entries zeroed. The elements of the subgroup deltas vector are copied into the
+ * correct element of the bigger move group delta vector.
+ * @param sub_group_deltas Set of command deltas for a subgroup of the move group actuated by servo
+ * @param robot_state Current robot state
+ * @param servo_params Servo params
+ * @param joint_name_group_index_map Mapping between sub group joint name and move group joint vector position
```
sea-bass, line 53:
```suggestion
 * @param joint_name_group_index_map Mapping between joint subgroup name and move group joint vector position.

**Review by sea-bass — COMMENTED** [review]
I tested this on our robot on linear actuator example and it works beautifully!

Once you address @AndyZe's comment that servo should be paused for this parameter switch to kick in, should be good to go from my end.

**Comment by sjahr** [comment]
> @ibrahiminfinite Any clue why the tests are failing?
> 
> ```
>     [ros2_control_node-1] terminate called after throwing an instance of 'std::runtime_error'
>     [ros2_control_node-1]   what():  can't compare times with different time sources
>     [ros2_control_node-1] Stack trace (most recent call last) in thread 23189:
>     [ros2_control_node-1] #13   Object "", at 0xffffffffffffffff, in 
>     [ros2_control_node-1] #12   Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7fde71106bf3, in __clone
>     [ros2_control_node-1] #11   Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7fde71075ac2, in 
>     [ros2_control_node-1] #10   Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30", at 0x7fde71307252, in 
>     [spawner-2] [INFO] [1696511567.840236022] [spawner_joint_state_broadcaster]: [NON-XML-CHAR-0x1B][92mConfigured and activated [NON-XML-CHAR-0x1B][1mjoint_state_broadcaster[NON-XML-CHAR-0x1B][0m
>     [ros2_control_node-1] #9    Object "/opt/ros/rolling/lib/controller_manager/ros2_control_node", at 0x5614c0025a33, in 
>     [ros2_control_node-1] #8    Object "/opt/ros/rolling/lib/libcontroller_manager.so", at 0x7fde7176af3c, in controller_manager::ControllerManager::update(rclcpp::Time const&, rclcpp::Duration const&)
>     [ros2_control_node-1] #7    Object "/opt/ros/rolling/lib/librclcpp.so", at 0x7fde7155bb09, in 
>     [ros2_control_node-1] #6    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30", at 0x7fde712d94d7, in __cxa_throw
>     [ros2_control_node-1] #5    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30", at 0x7fde712d9276, in std::terminate()
>     [ros2_control_node-1] #4    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30", at 0x7fde712d920b, in 
>     [ros2_control_node-1] #3    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30", at 0x7fde712cdb9d, in 
>     [ros2_control_node-1] #2    Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7fde710097f2, in abort
>     [ros2_control_node-1] #1    Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7fde71023475, in raise
>     [ros2_control_node-1] #0    Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7fde710779fc, in pthread_kill
>     [ros2_control_node-1] Aborted (Signal sent by tkill() 23157 0)
> Error: ROR] [ros2_control_node-1]: process has died [pid 23157, exit code -6, cmd '/opt/ros/rolling/lib/controller_manager/ros2_control_node --ros-args --params-file /tmp/launch_params_uphjjf6h --params-file /home/runner/work/moveit2/moveit2/.work/upstream_ws/install/moveit_resources_panda_moveit_config/share/moveit_resources_panda_moveit_config/config/ros2_controllers.yaml'].
> ```

@ibrahiminfinite FYI: The [CI failure](https://github.com/ros-planning/moveit2/actions/runs/6419387321/job/17429114006?pr=2396) was caused by a bug in an upstream dependency (https://github.com/ros-controls/ros2_control/pull/1127) and it already got fixed on their end with a new minor release https://github.com/ros-controls/ros2_control/compare/3.19.0...3.19.1, so our CI will be working again once the docker images got updated (Sebastian C just triggered that :pray:)

File: moveit_ros/moveit_servo/src/servo.cpp [hunk 14]
```diff
@@ -205,40 +233,41 @@ bool Servo::validateParams(const servo::Params& servo_params) const
...
+  }
+
   return params_valid;
 }
 
 bool Servo::updateParams()
 {
-  bool params_updated = false;
-
   if (servo_param_listener_->is_old(servo_params_))
```
sjahr, line 250:
```suggestion
  bool params_updated = false;
  
  if (servo_param_listener_->is_old(servo_params_))
```

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -205,30 +233,33 @@ bool Servo::validateParams(const servo::Params& servo_params) const
...
+  }
+
   return params_valid;
 }
 
 bool Servo::updateParams()
 {
   bool params_updated = false;
-
+
```
sjahr, line 251:
```suggestion

```

sea-bass:
It does seem there is another failure now, though... 😔

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -205,30 +233,32 @@ bool Servo::validateParams(const servo::Params& servo_params) const
...
   bool params_updated = false;
-
   if (servo_param_listener_->is_old(servo_params_))
   {
-    auto params = servo_param_listener_->get_params();
+    const auto params = servo_param_listener_->get_params();
 
-    const bool params_valid = validateParams(params);
-    if (params_valid)
+    if (validateParams(params))
```
AndyZe, line 255: It's a bit of an issue here that `validateParams()` is expensive. It doesn't really belong in a control loop. My suggestion here would be:

1. Have an `initialParamValidation()` function that gets called once and does a full validation.
2. Have a `runtimeParamValidation()` function which only checks the things that can change, from here:  https://github.com/ros-planning/moveit2/blob/main/moveit_ros/moveit_servo/config/servo_parameters.yaml

Preferably no string comparison or other expensive operations in `runtimeParamValidation()`.

AndyZe: If we need to trim down the list of parameters that can change so they can be validated in realtime, that's fine with me. Open to other ideas.

AndyZe: Also, simply not validating the parameters at all once the loop has started is a pretty good option.

sjahr:
Now this unittest is failing which is not related to these changes (it fails in #2288 and [main](https://github.com/ros-planning/moveit2/actions/runs/6434266730/job/17473109368) too).

```bash
    [moveit_servo_utils_test-5] [INFO] [1696613355.885158439] [moveit_robot_model.robot_model]: Loading robot model 'panda'...
    [moveit_servo_utils_test-5] /home/runner/work/moveit2/moveit2/.work/target_ws/src/moveit2/moveit_ros/moveit_servo/tests/test_utils.cpp:162: Failure
    [moveit_servo_utils_test-5] Expected equality of these values:
    [moveit_servo_utils_test-5]   scaling_result.second
    [moveit_servo_utils_test-5]     Which is: 1-byte object <02>
    [moveit_servo_utils_test-5]   moveit_servo::StatusCode::NO_WARNING
    [moveit_servo_utils_test-5]     Which is: 1-byte object <00>
```
@ibrahiminfinite Do you have any idea why this is happening? I don't see any immediate reason for the flakiness but I am probably missing something

ibrahiminfinite:
> Now this unittest is failing which is not related to these changes (it fails in #2288 and [main](https://github.com/ros-planning/moveit2/actions/runs/6434266730/job/17473109368) too).
> 
> ```shell
>     [moveit_servo_utils_test-5] [INFO] [1696613355.885158439] [moveit_robot_model.robot_model]: Loading robot model 'panda'...
>     [moveit_servo_utils_test-5] /home/runner/work/moveit2/moveit2/.work/target_ws/src/moveit2/moveit_ros/moveit_servo/tests/test_utils.cpp:162: Failure
>     [moveit_servo_utils_test-5] Expected equality of these values:
>     [moveit_servo_utils_test-5]   scaling_result.second
>     [moveit_servo_utils_test-5]     Which is: 1-byte object <02>
>     [moveit_servo_utils_test-5]   moveit_servo::StatusCode::NO_WARNING
>     [moveit_servo_utils_test-5]     Which is: 1-byte object <00>
> ```
> 
> @ibrahiminfinite Do you have any idea why this is happening? I don't see any immediate reason for the flakiness but I am probably missing something

That is weird!
I see no reason for it to fail, the test is also very straightforward.
Also the 2 other related tests (Hating and leaving singularity) are passing.

moveit_ros/moveit_servo/src/servo.cpp
ibrahiminfinite: `validateParams()`  is only called within `updateParams()` when a parameter has actually changed.
https://github.com/ros-planning/moveit2/blob/739afb649f5e248cb7d049bb35348a57fe5e5287/moveit_ros/moveit_servo/src/servo.cpp#L215
That should be fine right?

ibrahiminfinite: > * Have an `initialParamValidation()` function that gets called once and does a full validation.
> * Have a `runtimeParamValidation()` function which only checks the things that can change, from here:

But I like this idea better

sjahr:
That was my impression too, thanks for your quick answer! Also when I run the test locally, it succeeds :thinking:

sea-bass:
> That was my impression too, thanks for your quick answer! Also when I run the test locally, it succeeds 🤔

Maybe the change in condition number is right at the threshold of zero:

```cpp
  if (next_condition_number <= current_condition_number)
  {
    vector_towards_singularity *= -1;
  }
```

I would print the above values in the test to see how small they are.

You could consider maybe picking a bigger cartesian delta and/or singularity step scale in the servo params, or choosing an initial state with higher manipulability. It's OK to use bigger than default numbers here!

moveit_ros/moveit_servo/src/servo.cpp
AndyZe: Made a new issue for this.

https://github.com/ros-planning/moveit2/issues/2412

AndyZe (approved):

sea-bass:
I think I know what's wrong with the unit tests. (@ibrahiminfinite)

I ran these locally, printing output, and have been noticing their values are not consistent between runs. Turns out, when you do

```
robot_state->setJointGroupActivePositions();
```

This only sets the joint values but not the link transforms used in `getJacobian()` ... so you need to also call

```
robot_state->updateLinkTransforms();
```

When I made this change, my results are consistent, but also leads to new test failures because for many of the tests the condition number actually evaluates to `inf` -- so something in the calculation is not exactly going correctly.

Up to you guys if you want to look at this for this PR, or in a follow on.

sea-bass:
I think I may have fixed it in https://github.com/ros-planning/moveit2/pull/2414?

sea-bass (approved):

PR #2399, opened 2023-09-29
Use find_package for fcl

tylerjw:
### Description

The version of fcl we use has a find module. We should just use it.

See https://github.com/ros-planning/moveit2/issues/2391


ChrisThrasher (approved):
LGTM!

PR #2405, opened 2023-10-03
[Servo] Update Smoothing plugin API

ibrahiminfinite:
### Description

This PR updates the smoothing plugins to accept `Eigen` vectors instead of `std` vectors and lets us avoid having to create `Eigen::Map`s.  The only reason `moveit_servo::KinematicState` used `std::vector` was due to the smoothing plugin.
Additionally the API of the smoothing plugins has been updated in preparation for the `Ruckig` smoothing plugin.

### DIFF
--- a/moveit_core/online_signal_smoothing/CMakeLists.txt
+++ b/moveit_core/online_signal_smoothing/CMakeLists.txt
@@ -15,6 +15,7 @@ set_target_properties(moveit_smoothing_base PROPERTIES VERSION
 )
 ament_target_dependencies(moveit_smoothing_base
   rclcpp
+  tf2_eigen
 )
 
 # Plugin implementations
--- a/moveit_ros/moveit_servo/src/servo.cpp
+++ b/moveit_ros/moveit_servo/src/servo.cpp
@@ -103,8 +103,7 @@ Servo::Servo(const rclcpp::Node::SharedPtr& node, std::shared_ptr<const servo::P
     // Load the smoothing plugin
     if (servo_params_.use_smoothing)
     {
-      // setSmoothingPlugin();
-      RCLCPP_WARN(LOGGER, "smoothing plugin loaded");
+      setSmoothingPlugin();
     }
     else
     {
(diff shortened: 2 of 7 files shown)


AndyZe:
Ready for a review @ibrahiminfinite?

ibrahiminfinite:
Will be finishing up today.
It's mostly done and everything works, just a clang-tidy type issue is left to fix.

moveit_ros/moveit_servo/src/utils/common.cpp
```diff
@@ -121,13 +121,36 @@ trajectory_msgs::msg::JointTrajectory composeTrajectoryMessage(const servo::Para
 
   trajectory_msgs::msg::JointTrajectoryPoint point;
```
AndyZe, line 122: You could make this static so it's not re-created every time.

```suggestion
  static trajectory_msgs::msg::JointTrajectoryPoint point;
```

moveit_core/online_signal_smoothing/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ set_target_properties(moveit_smoothing_base PROPERTIES VERSION
 )
 ament_target_dependencies(moveit_smoothing_base
   rclcpp
+  tf2_eigen
```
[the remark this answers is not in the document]
AndyZe: Suggest deleting this line, replacing with:

`find_package(Eigen3 REQUIRED)`

```
include_directories(SYSTEM
  ${EIGEN3_INCLUDE_DIRS}
```
  
```suggestion
```

The reason is, there is nothing needing tf2 here.

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -459,36 +453,36 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
...
+    target_state.positions = current_state.positions + joint_position_delta;
 
     // TODO : apply filtering to the velocity instead of position
     // Apply smoothing to the positions if a smoother was provided.
     // Update filter state and apply filtering in position domain
     if (smoother_)
     {
-      smoother_->reset(current_state.positions);
-      smoother_->doSmoothing(target_state.positions);
+      smoother_->reset(current_state.positions, current_state.velocities, current_state.accelerations);
```
AndyZe, line 463: Not related to this PR, but does it even make sense to call `reset()` here? It seems like we should only do that after the arm has halted.

If you agree, let's open an issue

moveit_ros/moveit_servo/src/utils/common.cpp
```diff
@@ -119,15 +119,38 @@ trajectory_msgs::msg::JointTrajectory composeTrajectoryMessage(const servo::Para
   joint_trajectory.header.frame_id = servo_params.planning_frame;
   joint_trajectory.joint_names = joint_state.joint_names;
 
-  trajectory_msgs::msg::JointTrajectoryPoint point;
+  static trajectory_msgs::msg::JointTrajectoryPoint point;
   point.time_from_start = rclcpp::Duration::from_seconds(servo_params.publish_period);
+  const size_t num_joints = joint_state.joint_names.size();
+  point.positions.resize(num_joints);
+  point.velocities.resize(num_joints);
+  point.accelerations.resize(num_joints);
```
AndyZe, line 127:
```suggestion
  const size_t num_joints = joint_state.joint_names.size();
  if (point.positions.size() != num_joints)
  {
    point.positions.resize(num_joints);
    point.velocities.resize(num_joints);
    point.accelerations.resize(num_joints);
  }
```

moveit_ros/moveit_servo/src/utils/common.cpp
```diff
@@ -119,15 +119,38 @@ trajectory_msgs::msg::JointTrajectory composeTrajectoryMessage(const servo::Para
...
+      point.velocities[i] = joint_state.velocities[i];
+    }
+  }
+  if (servo_params.publish_joint_accelerations)
+  {
+    for (size_t i = 0; i < num_joints; i++)
+    {
+      point.accelerations[i] = joint_state.accelerations[i];
+    }
+  }
```
AndyZe, line 153: Can reduce the number of for-loops here:

```suggestion
    for (size_t i = 0; i < num_joints; i++)
    {
      if (servo_params.publish_joint_positions)
      {
        point.positions[i] = joint_state.positions[i];
      }
      if (servo_params.publish_joint_velocities)
      {
        point.velocities[i] = joint_state.velocities[i];
      }
      if (servo_params.publish_joint_accelerations)
      {
        point.accelerations[i] = joint_state.accelerations[i];
      }
    }
```

moveit_ros/moveit_servo/src/utils/common.cpp
```diff
@@ -137,13 +160,21 @@ std_msgs::msg::Float64MultiArray composeMultiArrayMessage(const servo::Params& s
...
-    multi_array.data = joint_state.positions;
+    for (size_t i = 0; i < num_joints; i++)
+    {
+      multi_array.data[i] = joint_state.positions[i];
+    }
   }
   else if (servo_params.publish_joint_velocities)
   {
-    multi_array.data = joint_state.velocities;
+    for (size_t i = 0; i < num_joints; i++)
```
AndyZe, line 174: Can combine the for-loops here, too, so you just have 1 for-loop

moveit_ros/moveit_servo/src/utils/common.cpp
ibrahiminfinite: I had considered putting  them in a single loop ,but from a performance point , I think it is better when we don't have conditionals inside the loop,

moveit_ros/moveit_servo/src/utils/common.cpp
ibrahiminfinite: I left out the check here so that the memory for the vectors are always pre-allocated rather than dynamically on "push_back"

moveit_ros/moveit_servo/src/servo.cpp
ibrahiminfinite: Since the current state of the joints might be different than the  `target_state` passed to `doSmoothing` in previous iteration, wouldn't calling `reset` at each iteration  be useful to avoid discrepancies?

firas-elkak:
في اثنين، 16 أكتوبر، 2023 في 2:23 م، كتب codecov[bot] <
***@***.***>:

AndyZe:
@ibrahiminfinite can i have push access to your fork? Wanted to attempt some cleanup

moveit_ros/moveit_servo/src/utils/common.cpp
AndyZe: But the `point` variable is static now? I still think this suggestion makes sense.

ibrahiminfinite: Ah! 
That makes sense , would only have to resize once since it is static unless joint group changes.

moveit_ros/moveit_servo/src/servo.cpp
AndyZe: Yeah, i don't think this is correct. Looking at the code for `reset()` it does this:

```
  previous_measurements_[0] = data;
  previous_measurements_[1] = data;
  previous_filtered_measurement_ = data;
```

So it completely wipes out all filter information. It's like there's no filtering at all. My bad for not catching this before.

AndyZe:
Replaced by #2470

PR #2411, opened 2023-10-07
[Python] Add RetimeTrajectory to RobotTrajectory

JensVanhooydonck:
### Description

This adds the RetimeTrajectory to the RobotTrajectory binding. This was available in the ROS1 movegroup python interface. This feature allows to plan a trajectory and change the trajectory speeds (velocity and acceleration values).

Currently the two algoritms inside trajectory_processing where added: Ruckig Trajectory Smoothing and Time Optimal Trajectory Generation.

### Checklist

### DIFF
--- a/moveit_py/src/moveit/moveit_core/robot_trajectory/robot_trajectory.cpp
+++ b/moveit_py/src/moveit/moveit_core/robot_trajectory/robot_trajectory.cpp
@@ -36,8 +36,7 @@
 
 #include "robot_trajectory.h"
 #include <moveit_py/moveit_py_utils/ros_msg_typecasters.h>
-#include <moveit/trajectory_processing/ruckig_traj_smoothing.h>
-#include <moveit/trajectory_processing/time_optimal_trajectory_generation.h>
+#include <moveit/trajectory_processing/trajectory_tools.h>
 
 namespace moveit_py
 {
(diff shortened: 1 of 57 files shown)


moveit_py/src/moveit/moveit_core/robot_trajectory/robot_trajectory.cpp
```diff
@@ -57,6 +59,34 @@ set_robot_trajectory_msg(const std::shared_ptr<robot_trajectory::RobotTrajectory
...
 
+bool retimeTrajectory(std::shared_ptr<robot_trajectory::RobotTrajectory>& self, const std::string& algorithm,
+                      const double& velocity_scaling_factor, const double& acceleration_scaling_factor)
+{
+  bool success = false;
+  // Do the actual retiming
+  if (algorithm == "ruckig_smoothing")
+  {
+    trajectory_processing::RuckigSmoothing time_param;
+    success = time_param.applySmoothing(*self, velocity_scaling_factor, acceleration_scaling_factor, false, 0.01);
```
sjahr, line 70: `mitigate_overshoot` and `overshoot_threshold` are not configurable with this API

moveit_py/src/moveit/moveit_core/robot_trajectory/robot_trajectory.h
```diff
@@ -55,6 +56,9 @@ robot_trajectory::RobotTrajectory
 set_robot_trajectory_msg(const std::shared_ptr<robot_trajectory::RobotTrajectory>& robot_trajectory,
                          const moveit::core::RobotState& robot_state, const moveit_msgs::msg::RobotTrajectory& msg);
 
+bool retimeTrajectory(std::shared_ptr<robot_trajectory::RobotTrajectory>& self, const std::string& algorithm,
```
sjahr, line 59: Please add documentation for the functions you're creating

```cpp
  /**
   * \brief ...
   * \param [in] ...
   * ...
   * \return
   */
```

sjahr (changes requested):
Thank you for your contribution! While I like the idea of having this functionality I think there a are some things that should be reconsidered:
- Rather than having a single `retimeTrajectory()` function that does not expose specific paramters of the individual algorithms + you can give it a non-existing algorithm name, I'd prefer having two more specific functions for example:

```cpp
addTOTGTimeParameterization(std::shared_ptr<robot_trajectory::RobotTrajectory>& self, const std::string& algorithm,
                      const double& velocity_scaling_factor, const double& acceleration_scaling_factor);
                      
applyRuckigSmoothing(std::shared_ptr<robot_trajectory::RobotTrajectory>& self, const std::string& algorithm,
                      const double& velocity_scaling_factor, const double& acceleration_scaling_factor, const bool mitigate_overshoot = false,
                                     const double overshoot_threshold = 0.01);
```
- These functions shouldn't be implemented in the python bindings package but as free functions in `moveit_core/trajectory_processing` (to avoid a circular dependency between RobotTrajectory and the time parameterization algorithms) e.g. [here](https://github.com/ros-planning/moveit2/blob/main/moveit_core/trajectory_processing/include/moveit/trajectory_processing/trajectory_tools.h) 

```

File: moveit_py/src/moveit/moveit_core/robot_trajectory/robot_trajectory.cpp [hunk 3]
```diff
@@ -134,6 +164,19 @@ void init_robot_trajectory(py::module& m)
            Returns:
                list of float: The duration from previous of each waypoint in the trajectory.
            )")
+      .def("retime_trajectory", &moveit_py::bind_robot_trajectory::retimeTrajectory, py::arg("algorithm"),
```
v4hn, line 167: why would you call a method on a class `RobotTrajectory` `retime_trajectory` instead of simply `retime`?
In moveit, the reason for the full name is that it's a member of the `MoveGroupInterface` and not the trajectory.

v4hn (commented):
I would suggest to either

moveit_py/src/moveit/moveit_core/robot_trajectory/robot_trajectory.h
JensVanhooydonck: Added documentation for all functions in trajectory_tools.h

sjahr (approved):
Looks good, thank you!

v4hn (approved):
Looks good assuming it's tested.

JensVanhooydonck:
> Looks good assuming it's tested. 

Tested both functions already in my python script.

PR #2417, opened 2023-10-10
Handle unsupported position constraints in OMPL

henningkayser:
OMPL constrained planning assumes that all position constraints have three dimensions, meaning that they are represented by a BOX bounding volume. If another shape is used (like a SPHERE from moveit_core/kinematic_constraints/utils.hpp), the constraint adapter implementation will produce a segfault because of the lack of dimensions. This fix prevents this by checking for the required BOX type.


moveit_planners/ompl/ompl_interface/src/detail/ompl_constraints.cpp
```diff
@@ -364,48 +364,61 @@ ompl::base::ConstraintPtr createOMPLConstraints(const moveit::core::RobotModelCo
...
   {
-    BaseConstraintPtr pos_con;
-    if (constraints.name == "use_equality_constraints")
+    if (constraints.position_constraints.size() > 1)
+    {
+      RCLCPP_WARN(LOGGER, "Only a single position constraint is supported. Using the first one.");
+    }
+
+    const auto& primitives = constraints.position_constraints.at(0).constraint_region.primitives;
+    if (primitives.empty() || primitives.at(0).type != shape_msgs::msg::SolidPrimitive::BOX)
```
sea-bass, line 381: Based on how the `BoxConstraint` and `EqualityPositionConstraint` parsing works, I think we maybe also want a warning if the number of primitives is greater than 1?

moveit_planners/ompl/ompl_interface/src/detail/ompl_constraints.cpp
```diff
@@ -364,48 +364,61 @@ ompl::base::ConstraintPtr createOMPLConstraints(const moveit::core::RobotModelCo
...
     {
-      pos_con = std::make_shared<EqualityPositionConstraint>(robot_model, group, num_dofs);
+      RCLCPP_ERROR(LOGGER, "Unable to plan with the requested position constraint. "
+                           "Only BOX primitive shapes are supported as constraint region.");
```
sea-bass, line 384: Did you want this case to continue on, or to return `nullptr`?

henningkayser: My intention was to leave ompl_constraints empty, so that line 416 kicks in

sjahr (approved):

PR #2418, opened 2023-10-10
Use constraint distance instead of bool validity in STOMP cost function

henningkayser:
This enables STOMP to optimize on the more precise distance gradient of a constraint set instead of the very imprecise const penalty.

### DIFF
--- a/moveit_planners/stomp/include/stomp_moveit/cost_functions.hpp
+++ b/moveit_planners/stomp/include/stomp_moveit/cost_functions.hpp
@@ -49,7 +49,7 @@
 
 namespace stomp_moveit
 {
-// Decides if the given state position vector is valid or not - example use cases are collision or constraint checking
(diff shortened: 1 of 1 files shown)


moveit_planners/stomp/include/stomp_moveit/cost_functions.hpp
```diff
@@ -50,14 +50,14 @@
 namespace stomp_moveit
 {
 // Decides if the given state position vector is valid or not - example use cases are collision or constraint checking
-using StateValidatorFn = std::function<bool(const Eigen::VectorXd& state_positions)>;
+using StateValidatorFn = std::function<double(const Eigen::VectorXd& state_positions)>;
```
sjahr, line 53: Comment above needs to be updated

sjahr (approved):

PR #2420, opened 2023-10-10
Enable iron build jobs in CI

henningkayser:
I thought we had done this already for main, but obviously this was not the case. Iron should also be enforced for PRs.


sea-bass (approved):

henningkayser:
> Seems like the [failing test](https://github.com/ros-planning/moveit2/blob/main/moveit_setup_assistant/moveit_setup_srdf_plugins/test/test_srdf.cpp) is doing `rclcpp::init()` and creating an `rclcpp::Logger` instance even though neither is being used.
> 
> This can be removed for this test, but there likely will be more instances like these.

My take is that we might want to rerun this on a fresh set of iron images.

sea-bass:
Sounds good. And actually, that failing test does use `rclcpp` under the hood in an included test fixture, it just wasn't immediately visible from only looking at the file.

sea-bass:
@henningkayser I saw this sitting around and updated / re-ran CI under the assumption that it should work now.

One question though: I notice that with this PR both humble and iron also add a `-testing` image, but the Jazzy CI does not. Should this also add that?

sea-bass:
Iron's about to go EOL, so it's probably not worth putting this in at this point.

Issue #2408, opened 2023-10-06
Extend Planning Pipeline to allow calling multiple planners in sequence


sjahr:
## Motiviation
_In order to_ get better solutions
_As a_ user
_I want to_ to seed an optimizing planner with a solution from RRTConnect

_In order to_ configure the pipeline easily
_As a_ user
_I want to_ determine the order in which certain pipeline components are being called

## Current implementation and limitations
- A planner plugin + a bunch of plan request adapters are configured and the call order is figured out by some logic of a class `PlanningRequestAdapterChain` + inside the plugins 
- It is currently not possible in MoveIt2 to chain planners and the CHOMP/ STOMP post processing adapters don't seem to work
- PlanRequestAdapters are not taken into account for the configured maximum planning time
- The component call order is [very un-intuitive](https://github.com/ros-planning/moveit2/blob/739afb649f5e248cb7d049bb35348a57fe5e5287/moveit_core/planning_request_adapter/src/planning_request_adapter.cpp#L103) and [apparently prone to errors](https://github.com/ros-planning/moveit2/pull/2234)

## Proposed refactoring
- [x] Determine pipeline order with configuration in yaml #2429 
  - [x] Distinguish between PlanRequestAdapters (Pre-processing) and PlanResponseAdapters (Post-processing)
  - [x] Remove all logic from plugins that messes with the call order
  - [x] Call PlanRequestAdapter vector, Planner vector and PlanResponseAdapters vector in the correct order
- [ ] Call multiple planners in sequence https://github.com/ros-planning/moveit2/pull/2457
  - [x] Enable loading multiple planner plugins
  - [x] Call planner plugins in load-order
  - [x] Remove planner PlanRequestAdapters (CHOMP, STOMP)
  - [x] Make sure, planner plugins use reference trajectory if available

MarqRazz:
@sjahr what is the status here? Could you please update the milestone and link any pending PR's?

sjahr:
Both PRs need to be merged in order to close this issue. I've linked the second one, so this is considered done once that issue is merged

v4hn:
You spent a lot of work on this, and it's definitely a good idea to get rid of the planning request adapter recursive call structure.

The main shortcoming I keep encountering with the recursive PRA structure is that it does not allow support for anytime planners that yield (over time improving) solutions continuously. It's a very nice feature of the tesseract GUI, for example.
[...]

PR #2429, opened 2023-10-11
[Planning Pipeline Refactoring] #1 Simplify Adapter - Planner chain

sjahr:
## Description
Addresses #2408 
Requires https://github.com/ros-planning/moveit_msgs/pull/170

Related PRs:
- https://github.com/ros-planning/moveit_resources/pull/189
- https://github.com/ros-planning/moveit2_tutorials/pull/793
### How to test
Use https://github.com/ros-planning/moveit2_tutorials/pull/793 and https://github.com/ros-planning/moveit_resources/pull/189 and run 

```
ros2 launch moveit2_tutorials move_group.launch.py
```
and
```
ros2 launch moveit2_tutorials motion_planning_pipeline_tutorial.launch.py
```

Expected output
```bash
[motion_planning_pipeline_tutorial-1] [INFO] [1699454408.777050834] [motion_planning_pipeline_tutorial.remote_control]: Waiting to continue: Press 'next' in the RvizVisualToolsGui window to continue the demo
[motion_planning_pipeline_tutorial-1] [INFO] [1699454410.087785146] [motion_planning_pipeline_tutorial.remote_control]: ... continuing
[motion_planning_pipeline_tutorial-1] [INFO] [1699454410.087971200] [motion_planning_pipeline_tutorial]: Calling PlanningRequestAdapter 'ResolveConstraintFrames'
[motion_planning_pipeline_tutorial-1] [INFO] [1699454410.088272213] [motion_planning_pipeline_tutorial]: Calling PlanningRequestAdapter 'ValidateWorkspaceBounds'
[motion_planning_pipeline_tutorial-1] [WARN] [1699454410.088287442] [motion_planning_pipeline_tutorial.validate_workspace_bounds]: It looks like the planning volume was not specified. Using default values.
[motion_planning_pipeline_tutorial-1] [INFO] [1699454410.088442316] [motion_planning_pipeline_tutorial]: Calling PlanningRequestAdapter 'CheckStartStateBounds'
[motion_planning_pipeline_tutorial-1] [INFO] [1699454410.088596429] [motion_planning_pipeline_tutorial]: Calling PlanningRequestAdapter 'CheckStartStateCollision'
[motion_planning_pipeline_tutorial-1] [INFO] [1699454410.089114285] [moveit.ompl_planning.model_based_planning_context]: Planner configuration 'panda_arm[RRTConnectkConfigDefault]' will use planner 'geometric::RRTConnect'. Additional configuration parameters will be set when the planner is constructed.
[motion_planning_pipeline_tutorial-1] [INFO] [1699454410.124115820] [motion_planning_pipeline_tutorial]: Calling PlanningResponseAdapter 'AddTimeOptimalParameterization'
[motion_planning_pipeline_tutorial-1] [WARN] [1699454410.124140126] [moveit_trajectory_processing.time_optimal_trajectory_generation]: Invalid max_velocity_scaling_factor 0.000000 specified, defaulting to 1.000000 instead.
[motion_planning_pipeline_tutorial-1] [WARN] [1699454410.124149954] [moveit_trajectory_processing.time_optimal_trajectory_generation]: Invalid max_acceleration_scaling_factor 0.000000 specified, defaulting to 1.000000 instead.
[motion_planning_pipeline_tutorial-1] [INFO] [1699454410.125490595] [motion_planning_pipeline_tutorial]: Calling PlanningResponseAdapter 'ValidateSolution'
[motion_planning_pipeline_tutorial-1] [INFO] [1699454410.125781979] [motion_planning_pipeline_tutorial]: Calling PlanningResponseAdapter 'DisplayMotionPath'
[motion_planning_pipeline_tutorial-1] [INFO] [1699454410.125976980] [motion_planning_pipeline]: Visualizing the trajectory
```
### Changes
- Simplify pipeline component call chain to a sequence of two vectors + a planner
- Disallow adapters to call the planner
-  Remove FixStartStatePathConstraints :arrow_right: The problem of connecting a start state to the start of a constrained motion should be solved with MTC today 
- Add `PipelineState` publisher which publishes the intermediate state between the pipeline stages. 
  - To test it, set the `progress_topic` some topic name you'd like to see the intermediate the intermediate results published to before the planning pipeline instances are constructed. The parameter is evaluated during the construction of a planning pipeline instance
 - Make the moveit error code the sole source to report success or failure
   - `moveit::core::MoveItStatus PlanningRequestAdapter::adapt(..` Plan request adapters return a status containing error information
   - PlannerContext `virtual void solve(MotionPlanResponse& res) = 0;` uses the response to communicate success or failure
   - PlanRequestAdapter `virtual void adapt(..., planning_interface::MotionPlanResponse& res)` also uses the response for communication
   

### TODOs
- [x] Write Unit tests
- [ ] Create humble release + branch for moveit_resources
- [x] Update setup assistant
- [x] Add instructions for migration guide


moveit_core/CMakeLists.txt
```diff
@@ -59,7 +59,7 @@ add_subdirectory(kinematics_metrics)
 add_subdirectory(macros)
 add_subdirectory(online_signal_smoothing)
 add_subdirectory(planning_interface)
-add_subdirectory(planning_request_adapter)
+#add_subdirectory(planning_request_adapter)
```
sjahr, line 62:
```suggestion
```

moveit_core/CMakeLists.txt
```diff
@@ -115,7 +115,7 @@ install(
     moveit_kinematics_metrics
     moveit_macros
     moveit_planning_interface
-    moveit_planning_request_adapter
+    #moveit_planning_request_adapter
```
sjahr, line 118:
```suggestion
```

moveit_ros/planning/planning_pipeline/res/planning_pipeline_parameters.yaml
```diff
@@ -5,7 +5,12 @@ planning_pipeline_parameters:
     default_value: "UNKNOWN",
   }
   request_adapters: {
-    type: string,
+    type: string_array,
+    description: "Names of the planning request adapter plugins (plugin names separated by space).",
```
sjahr, line 9:
```suggestion
    description: "Names of the planning request adapter plugins  (Order in chain is similar to array order).",
```

moveit_ros/planning/planning_pipeline/res/planning_pipeline_parameters.yaml
```diff
@@ -5,7 +5,12 @@ planning_pipeline_parameters:
...
+    default_value: [],
+  }
+  response_adapters: {
+    type: string_array,
     description: "Names of the planning request adapter plugins (plugin names separated by space).",
```
sjahr, line 14:
```suggestion
    description: "Names of the planning request adapter plugins  (Order in chain is similar to array order).",
```

Abishalini (commented):
I briefly read through this PR and had some cursory suggestions. I will give this another read at the end of the day for a more thorough review.

moveit_ros/planning/planning_pipeline/res/planning_pipeline_parameters.yaml
```diff
@@ -5,7 +5,12 @@ planning_pipeline_parameters:
     default_value: "UNKNOWN",
   }
   request_adapters: {
```
MarqRazz, line 7: Are these parameters something that can change at runtime with dynamic params? If so do we have to worry about someone changing the list while we are processing a request?

MarqRazz (commented):
I was able to build and run the examples and it works as expected. Great work @sjahr!

moveit_ros/planning/introspection/src/list_planning_adapter_plugins.cpp
```diff
@@ -65,7 +65,7 @@ int main(int argc, char** argv)
   for (const std::string& adapter_plugin_name : classes)
   {
     std::cout << " \t " << adapter_plugin_name << '\n';
```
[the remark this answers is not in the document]
sjahr: I'd even think we should just delete this

moveit_ros/planning/planning_response_adapter_plugins/res/default_plan_response_adapter_params.yaml
```diff
@@ -0,0 +1,17 @@
+# This files contains the parameters for all of MoveIt's default PlanResponseAdapters
```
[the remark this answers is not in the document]
sjahr: Good thought, we could add another namespace to differentiate between the plugins in the `default_plan_response_adapter_parameters` namespace. What do you think?

moveit_ros/planning/planning_response_adapter_plugins/res/default_plan_response_adapter_params.yaml
```diff
@@ -0,0 +1,17 @@
...
+  resample_dt: {
+    type: double,
+    description: "AddTimeOptimalParameterization: Time steps between two adjacent waypoints of the parameterized trajectory. The trajectory is re-sampled from the original path.",
+    default_value: 0.1,
+  }
+  min_angle_change: {
+    type: double,
+    description: "AddTimeOptimalParameterization: Minimum joint value change to consider two waypoints unique.",
+    default_value: 0.001,
+  }
```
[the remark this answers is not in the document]
sjahr: No, because that is set by the motion plan request and not by a parameter. To remove this warning, it needs to be configured in the request.

moveit_ros/planning/planning_pipeline/include/moveit/planning_pipeline/planning_pipeline.h
```diff
@@ -32,32 +32,108 @@
...
+    if (adapter)
+    {
+      adapter->initialize(node, parameter_namespace);
+      plugin_vector.push_back(std::move(adapter));
+      RCLCPP_INFO(node->get_logger(), "Loaded adapter '%s'", plugin_name.c_str());
+    }
+    else
+    {
+      RCLCPP_WARN(node->get_logger(), "Failed to load adapter '%s'", plugin_name.c_str());
+    }
```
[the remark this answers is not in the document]
sjahr: I guess that depends on the implementation but the warning is misleading. I renamed it

moveit_ros/planning/planning_pipeline/res/planning_pipeline_parameters.yaml
sjahr: No they are only loaded once during initialization and then they don't change

moveit_ros/planning/planning_response_adapter_plugins/res/default_plan_response_adapter_params.yaml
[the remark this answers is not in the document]
MarqRazz: Yes I would like if the parameters were namespaced.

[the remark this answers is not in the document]
MarqRazz: I changed this params file as you suggested and got the error below but after updating add_time_optimal_parameterization to `params.totg.xxx` it compiled an ran for me.

```
/root/moveit2_ws/src/moveit2/moveit_ros/planning/planning_response_adapter_plugins/src/add_time_optimal_parameterization.cpp:72:51: error: ‘const struct default_plan_response_adapter_parameters::Params’ has no member named ‘path_tolerance’
   72 |       TimeOptimalTrajectoryGeneration totg(params.path_tolerance, params.resample_dt, params.min_angle_change);
      |                                                   ^~~~~~~~~~~~~~
/root/moveit2_ws/src/moveit2/moveit_ros/planning/planning_response_adapter_plugins/src/add_time_optimal_parameterization.cpp:72:74: error: ‘const struct default_plan_response_adapter_parameters::Params’ has no member named ‘resample_dt’
   72 |       TimeOptimalTrajectoryGeneration totg(params.path_tolerance, params.resample_dt, params.min_angle_change);
      |                                                                          ^~~~~~~~~~~
/root/moveit2_ws/src/moveit2/moveit_ros/planning/planning_response_adapter_plugins/src/add_time_optimal_parameterization.cpp:72:94: error: ‘const struct default_plan_response_adapter_parameters::Params’ has no member named ‘min_angle_change’
   72 |       TimeOptimalTrajectoryGeneration totg(params.path_tolerance, params.resample_dt, params.min_angle_change);
      |                                                                                              ^~~~~~~~~~~~~~~~
```

moveit_ros/planning/planning_response_adapter_plugins/res/default_plan_response_adapter_params.yaml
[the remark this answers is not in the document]
MarqRazz: Do you think it's worth updating the tutorials in PR 793 to set these values and suppress the warnings?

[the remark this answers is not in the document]
sjahr: Yes!

moveit_ros/planning/planning_response_adapter_plugins/res/default_plan_response_adapter_params.yaml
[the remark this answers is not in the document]
sjahr: Ok, I'll try again. I had an error caused by the code generation but maybe I messed something up

moveit_ros/planning/planning_pipeline/include/moveit/planning_pipeline/planning_pipeline.h
```diff
@@ -32,32 +32,106 @@
...
+  }
+  // Return true after all adapters were successfully applied
+  return true;
+}
+
+/**
+ * \brief Helper function template to load a vector of plugins
+ *
+ * \tparam TPluginClass Plugin class type, PlanningRequestAdapter or PlanningResponseAdapter
```
MarqRazz, line 92: Should this be `\tparam`? Don't we have a standard format for comments? Lots of the other files look like they follow standard doxygen.

```
  /** \brief ....
   *  @param ...
   *  @param ...
   */
```

moveit_ros/planning/default_response_adapters_plugin_description.xml
```diff
@@ -0,0 +1,15 @@
...
+
+  <class name="default_planner_response_adapters/AddTimeOptimalParameterization" type="default_planner_response_adapters::AddTimeOptimalParameterization" base_class_type="planning_interface::PlanningResponseAdapter">
+    <description>
+      Time parameterization using Time-Optimal Trajectory Generation for Path Following With Bounded Acceleration and Velocity (Kunz and Stilman, RSS 2008). Assumes starting and ending at rest. Not jerk limited.
+    </description>
+  </class>
+
+  <class name="default_planner_response_adapters/AddRuckigTrajectorySmoothing" type="default_planner_response_adapters::AddRuckigTrajectorySmoothing" base_class_type="planning_interface::PlanningResponseAdapter">
+    <description>
+      Adds jerk-limited trajectory smoothing via ruckig. Best used after a time parameterization algorithm such as TimeOptimalParameterization.
```
MarqRazz, line 11: Ruckig

MarqRazz (commented):
I assume you are still updating the dependencies because right now I get this error...
```
[motion_planning_pipeline_tutorial-1] terminate called after throwing an instance of 'rclcpp::ParameterTypeException'
[motion_planning_pipeline_tutorial-1]   what():  expected [string_array] got [string]
```

moveit_ros/planning/planning_pipeline/include/moveit/planning_pipeline/planning_pipeline.h
sjahr: `\tparam` highlights that this is the template param and not and input argument

sjahr:
> ros2 launch moveit2_tutorials motion_planning_pipeline_tutorial.launch.py

@MarqRazz I cannot reproduce that issue. Are you using  https://github.com/ros-planning/moveit_resources/pull/189 and https://github.com/ros-planning/moveit2_tutorials/pull/793? Maybe you need to wipe your workspace and rebuild it :thinking:

moveit_ros/planning/default_response_adapters_plugin_description.xml
sjahr, line 11:
```suggestion
      Adds jerk-limited trajectory smoothing via Ruckig. Best used after a time parameterization algorithm such as TimeOptimalParameterization.
```

MarqRazz:
> > ros2 launch moveit2_tutorials motion_planning_pipeline_tutorial.launch.py
> 
> @MarqRazz I cannot reproduce that issue. Are you using [ros-planning/moveit_resources#189](https://github.com/ros-planning/moveit_resources/pull/189) and [ros-planning/moveit2_tutorials#793](https://github.com/ros-planning/moveit2_tutorials/pull/793)? Maybe you need to wipe your workspace and rebuild it 🤔

So I just did a clean build and it is still failing. Below is what my workspace looks like. I'm testing this in Humble so maybe that is causing issues???
```
=== ./moveit2 (git) ===
On branch pr-refactor_pipeline_chain
Your branch is up to date with 'sebj/pr-refactor_pipeline_chain'.
nothing to commit, working tree clean
=== ./moveit2_tutorials (git) ===
On branch pr-update_for_refactroing_planning_pipeline
Your branch is up to date with 'origin/pr-update_for_refactroing_planning_pipeline'.
nothing to commit, working tree clean
=== ./moveit_msgs (git) ===
On branch ros2
Your branch is up to date with 'origin/ros2'.
nothing to commit, working tree clean
=== ./moveit_resources (git) ===
On branch pr-update_pipeline_planner_configs
Your branch is up to date with 'origin/pr-update_pipeline_planner_configs'.
nothing to commit, working tree clean
=== ./moveit_task_constructor (git) ===
On branch ros2
Your branch is up to date with 'origin/ros2'.
nothing to commit, working tree clean
=== ./moveit_visual_tools (git) ===
On branch ros2
Your branch is up to date with 'origin/ros2'.
nothing to commit, working tree clean
=== ./pick_ik (git) ===
On branch main
Your branch is up to date with 'origin/main'.
nothing to commit, working tree clean
=== ./ros2_kortex (git) ===
On branch main
Your branch is up to date with 'origin/main'.
nothing to commit, working tree clean
=== ./rviz_visual_tools (git) ===
On branch ros2
Your branch is up to date with 'origin/ros2'.
nothing to commit, working tree clean
```

MarqRazz (approved):
I'm not sure what I changed other than restarting my computer and the crash is now gone. I tested with Cyclone and FastDDS, I thought the RMW might be causing issues, and everything worked as expected.

sjahr:
> I'm not sure what I changed other than restarting my computer and the crash is now gone. I tested with Cyclone and FastDDS, I thought the RMW might be causing issues, and everything worked as expected.

Glad it worked! I have not tested it on humble but on rolling, so good to know it works on humble too. Thanks!

moveit_ros/planning/planning_request_adapter_plugins/src/fix_workspace_bounds.cpp
```diff
@@ -65,9 +65,9 @@ class FixWorkspaceBounds : public planning_request_adapter::PlanningRequestAdapt
     return "Fix Workspace Bounds";
   }
 
-  bool adaptAndPlan(const PlannerFn& planner, const planning_scene::PlanningSceneConstPtr& planning_scene,
-                    const planning_interface::MotionPlanRequest& req,
-                    planning_interface::MotionPlanResponse& res) const override
+  bool adapt(const planning_scene::PlanningSceneConstPtr& /*planning_scene*/,
+             planning_interface::MotionPlanRequest& req,
+             const planning_interface::MotionPlanResponse& /*res*/) const override
```
sea-bass, line 70: This one seems like it wouldn't need the planner, since it is just mutating the workspace bounds. ✅

sea-bass (commented):
I have 2 high level comments from reading this.

### 1: Removing the planner function from `adaptAndPlan()`?
At a glance, it seems okay to remove this and simply have the adapters mutate the original request/response... but it does seem weird to me that replacing a call to the planner's solve with simply `true` will keep exactly the same functionality.

... but maybe it does work out? It seems like the previous "adapter chain" implementation would basically call the adapters recursively so it likely has the same effect as what you changed it to, but way more confusing than your cleanup!

### 2. Discussions on `PlanRequestAdapter` vs. `PlanResponseAdapter`
We spoke offline about the existence of the separate classes, and how you've had to create a templated function to run `adapt()` on both types, plus a lot of copypasta. I do like that one implementation holds the request const and the other holds the response const, though!

Another alternative would be to have a single `PlanAdapter` base class with empty implementations of, e.g., `adaptRequest()` and `adaptResponse()` methods that can be separately overridden. This way, you can have a single adapter that works as both request/response adapter, and also gets rid of code duplication and the need for templates.

Don't have a strong opinion here, but as we discussed I wanted to float the option?

sea-bass:
Testing with the Kinova config from the tutorials, I just found a bug with the default Pilz config that gets loaded by `MoveItConfigsBuilder`. If you remove the following line, it does what we want:

https://github.com/ros-planning/moveit2/blob/main/moveit_configs_utils/default_configs/pilz_industrial_motion_planner_planning.yaml#L2

sjahr:
@sea-bass Thanks for your review! Regarding your high-level comments:

> 1: Removing the planner function from adaptAndPlan()? ...

My understanding is that once we get to the point where the planner functions is called, i within the plugin we need to assume the happy-path in order to forward the request changes to the planner plugin call within the pipeline. That's why I simply return true here. This way the accumulated modifications of the motion planning request will be given to the planner plugin itself. Personally, I am unsure about the usefulness (and correctness) of most of the existing default request adapters and I'd be in favor of removing or refactoring them but to avoid scope creep I did not start with that in this PR.

> 2. Discussions on PlanRequestAdapter vs. PlanResponseAdapter ...

As we discussed offline, this is one of the parts were I am unsure about the design decision I made and I appreciate opinions and suggestions in this review! The way I implemented it has the downside of introducing some copypasta but we have two distinguishable classes/ types for pre- and post processing. I think having a single `PlanAdapter` class might tempt people who write custom plugins to create messy plugin classes that contain a broad variety of non-reusable pre- and post-processing logic.  Also we'd require some kind of function that checks if an adapter loaded as a request adapters actually is a request adapter (Although a `validateInput()` function for all plugins to check if the base assumptions of that algorithm are satisfied might actually be a good idea!).  An middle way I considered was to have a parent `PlanAdapter` class from which the Request and ResponseAdapter classes inherit but that additional inheritance also increases the complexity.

sjahr:
> I might have an issue with this:
> 
> > 
> 
> Remove FixStartStatePathConstraints ➡️ The problem of connecting a start state to the start of a constrained motion should be solved with MTC today
> 
> I certainly don't want MTC to be mandatory. Can you explain more what the issue is?

@AndyZe Sorry I should have explained this in more detail. In general, MTC won't be mandatory but I think it should for the use case that the removed adapter solves. `FixStartStatePathConstraints` solves a specific use-case with a not intuitive solution ([description from the tutorials](https://moveit.picknik.ai/main/doc/concepts/motion_planning.html#fixstartstatepathconstraints)):

```
This adapter is applied when the start state for a motion plan does not obey the specified path constraints. 
It will attempt to plan a path between the current configuration of the robot to a new location where
the path constraint is obeyed. The new location will serve as the start state for planning.
```

So the adapter checks if the start state satisfies the path constraints [[1]](https://github.com/ros-planning/moveit2/blob/87e3f9169ac9e179bd9b95014c13d319e25e853b/moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_path_constraints.cpp#L86) and if it doesn't it creates a solution that partially satisfies the path constraints. First, a trajectory from the start state to a state that satisfies the path constraints is planned [[2]](https://github.com/ros-planning/moveit2/blob/87e3f9169ac9e179bd9b95014c13d319e25e853b/moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_path_constraints.cpp#L93) (Note: This subsolution does not satisfy the path constraints). Second, from the end state of that first subsolution another trajectory is planned to the original goal state from the planning request [[3]](https://github.com/ros-planning/moveit2/blob/87e3f9169ac9e179bd9b95014c13d319e25e853b/moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_path_constraints.cpp#L107) (Note: That solution satisfies the path constraints). These two subsolutions are then concatenated to a full solution which is returned as solution by the planning pipeline [[[4]](https://github.com/ros-planning/moveit2/blob/87e3f9169ac9e179bd9b95014c13d319e25e853b/moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_path_constraints.cpp#L130)] (Note: the overall solution does NOT satisfy the path constraints).

With the refactorings of this PR, this functionality is not possible anymore, because the adapters cannot call the planner anymore, but more importantly, I think this behavior is un-intuitive for the user. A seemingly valid solution is returned, that violates the path constraints and that contains a detour to some state that does :thinking:. Sure, somewhere in the logging output are warnings about that, but overall as I user I'd expect a planning pipeline to return failure if no solutions can be created that satisfies the constraints completely.

I think it worked before because path constraints are not used very often but nonetheless, the adapter is part of every default config and might lead to weird& unexpected behavior in case someone wants to use path constraints.

moveit_ros/planning/planning_pipeline/include/moveit/planning_pipeline/planning_pipeline.h
```diff
@@ -32,32 +32,107 @@
...
+ *
+ * \tparam TAdapter Adapter type, PlanningRequestAdapter or PlanningResponseAdapter
+ * \param adapter_vector Adapters to be applied to request or response
+ * \param planning_scene Planning context
+ * \param req Motion plan request that describes the planning problem
+ * \param res Motion plan response to which the solution is written
+ * \param logger ROS logger
+ * \return true if all adapters are applied successfully
+ */
+template <class TAdapter>
```
henningkayser, line 68: I would prefer an API without templates for this, if possible. As already discussed, the adapter interfaces for requests and responses should probably not be the same.

moveit_ros/planning/planning_pipeline/include/moveit/planning_pipeline/planning_pipeline.h
```diff
@@ -32,32 +32,107 @@
...
+                                    const planning_scene::PlanningSceneConstPtr& planning_scene,
+                                    planning_interface::MotionPlanRequest& req,
+                                    planning_interface::MotionPlanResponse& res, const rclcpp::Logger& logger)
+{
+  for (const auto& adapter : adapter_vector)
+  {
+    assert(adapter);
+    RCLCPP_INFO(logger, "Calling PlanningAdapter '%s'", adapter->getDescription().c_str());
+    // If adapter does not succeed, break chain and return false
+    if (!adapter->adapt(planning_scene, req, res))
```
henningkayser, line 79: Here, we should keep track of the individual changes and errors as we go, allowing the caller to introspect what's going wrong where.

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_bounds.cpp
```diff
@@ -181,34 +180,14 @@ class FixStartStateBounds : public planning_request_adapter::PlanningRequestAdap
...
-      planning_interface::MotionPlanRequest req2 = req;
-      moveit::core::robotStateToRobotStateMsg(start_state, req2.start_state);
-      solved = planner(planning_scene, req2, res);
+      RCLCPP_WARN(logger_, "Changing start state.");
+      moveit::core::robotStateToRobotStateMsg(start_state, req.start_state);
     }
-    else
-      solved = planner(planning_scene, req, res);
 
-    // re-add the prefix state, if it was constructed
```
henningkayser, line 195: without the block below, this adapter is somewhat useless (not that I'm arguing that it is overwhelmingly useful otherwise). I would suggest replacing these "hacks" with validity checks that throw an error if the precondition is not met - here start state out of bounds -> fail and tell the user

moveit_core/planning_interface/include/moveit/planning_interface/planning_interface.h
```diff
@@ -39,16 +39,12 @@
 #include <moveit/macros/class_forward.h>
 #include <moveit/planning_interface/planning_request.h>
 #include <moveit/planning_interface/planning_response.h>
```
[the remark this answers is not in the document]
sjahr: It is done in planning_scene.h. Why shouldn't the planning-interface class depend on planning_scene, have I missed a circular dependency that is introduced by this?

moveit_core/planning_interface/include/moveit/planning_interface/planning_response_adapter.h
```diff
@@ -0,0 +1,75 @@
...
+  /** \brief Get a description of this adapter
+   *  @return Returns a short string that identifies the planning response adapter
+   */
+  [[nodiscard]] virtual std::string getDescription() const = 0;
+
+  /** @brief Adapt the planning response
+   *  @param planning_scene Representation of the environment for the planning
+   *  @param req Motion planning request with a set of constraints
+   *  @param res Reference to which the generated motion plan response is written to
+   *  @return True if response got generated correctly */
```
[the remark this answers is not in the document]
sjahr: I think this is necessary, so the response adapter can compare/ evaluate the solutions against the problem.

moveit_ros/planning/default_request_adapters_plugin_description.xml
```diff
@@ -0,0 +1,27 @@
+<library path="moveit_default_planning_request_adapter_plugins">
+
+  <class name="default_planner_request_adapters/FixStartStateBounds" type="default_planner_request_adapters::FixStartStateBounds" base_class_type="planning_interface::PlanningRequestAdapter">
```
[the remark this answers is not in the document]
sjahr: I've used for namespaces etc. `default_planning_*_adapter_plugins` and shortened the filenames to `default_*_adapter_plugins`. I think omitting the `planning` from the filenames makes sense to make them a bit shorter (`default_request_adapters_plugin_description.xml` is long enough :sweat_smile:).

moveit_ros/planning/planning_pipeline/include/moveit/planning_pipeline/planning_pipeline.h
sjahr: I removed it :+1:

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_bounds.cpp
sjahr: I updated all the adapters.

moveit_core/planning_interface/include/moveit/planning_interface/planning_interface.h
```diff
@@ -119,11 +115,11 @@ class PlanningContext
 
   /** \brief Solve the motion planning problem and store the result in \e res. This function should not clear data
    * structures before computing. The constructor and clear() do that. */
-  virtual bool solve(MotionPlanResponse& res) = 0;
+  virtual void solve(MotionPlanResponse& res) = 0;
 
   /** \brief Solve the motion planning problem and store the detailed result in \e res. This function should not clear
    * data structures before computing. The constructor and clear() do that. */
-  virtual bool solve(MotionPlanDetailedResponse& res) = 0;
+  virtual void solve(MotionPlanDetailedResponse& res) = 0;
```
sjahr, line 122: @henningkayser I am wondering if we should get rid of this. OMPL is the only planner that uses this interface (and has distinguishable processing steps as far as I know) and in the planning pipeline we don't use it anyways. Rather than enforcing this interface on all planners we should consider a custom solution for OMPL

moveit_core/utils/include/moveit/utils/moveit_status.h
```diff
@@ -1,7 +1,7 @@
 /*********************************************************************
  * Software License Agreement (BSD License)
  *
- *  Copyright (c) 2012, Willow Garage, Inc.
+ *  Copyright (c) 2021, PickNik Inc.
```
sjahr, line 4:
```suggestion
 *  Copyright (c) 2023, PickNik Inc.
```

moveit_core/utils/include/moveit/utils/moveit_status.h
```diff
@@ -32,35 +32,31 @@
...
 
-#include <moveit/planning_request_adapter/planning_request_adapter.h>
-#include <class_loader/class_loader.hpp>
+#include <string>
 
-namespace default_planner_request_adapters
+namespace moveit
 {
-/** @brief Empty adapter that just calls the planner. */
-class Empty : public planning_request_adapter::PlanningRequestAdapter
```
sjahr, line 45: This diff doesn't make much sense since this is a new file with completely different content :shrug:. We could consider using this MoveItStatus in other places as well and/or add it to the motion plan response since it is more verbose than a plain error code.

MIGRATION.md
```diff
@@ -3,19 +3,37 @@
...
+-     default_planner_request_adapters/FixStartStateCollision
+-     default_planner_request_adapters/FixStartStatePathConstraints
++ # The order of the elements in the adapter corresponds to the order they are processed by the motion planning pipeline.
++ request_adapters:
++   - default_planner_request_adapters/ResolveConstraintFrames
++   - default_planner_request_adapters/FixWorkspaceBounds
++   - default_planner_request_adapters/FixStartStateBounds
++   - default_planner_request_adapters/FixStartStateCollision
++ response_adapters:
++   - default_planner_response_adapters/AddTimeOptimalParameterization
```
sea-bass, line 22: These should change to `default_plannING_request_adapters` and `default_plannING_response_adapters` as well.

Also I think `FixWorkspaceBounds` is now `ValidateWorkspaceBounds`.

sea-bass (commented):
Glanced through all the code updates and all looks good -- just have some very small things.

I tested again on the tutorials just to be safe as well. Found a stack trace in the tutorials with:

```
[move_group-4] [FATAL] [1699537089.467628231] [move_group]: Exception while loading planning adapter plugin 'default_planner_request_adapters/CheckForStackedConstraints': According to the loaded plugin descriptions the class default_planner_request_adapters/CheckForStackedConstraints with base class type planning_interface::PlanningRequestAdapter does not exist. Declared types are  default_planning_request_adapters/CheckForStackedConstraints default_planning_request_adapters/CheckStartStateBounds default_planning_request_adapters/CheckStartStateCollision default_planning_request_adapters/ResolveConstraintFrames default_planning_request_adapters/ValidateWorkspaceBounds
[move_group-4] terminate called after throwing an instance of 'pluginlib::LibraryLoadException'
```

I think we have to change all config instances of `default_planner_request_adapters` to `default_plannING_request_adapters` in the related PRs. And also change things like `FixWorkspaceBounds` to `ValidateWorkspaceBounds`. My diffs were:

image

One thing to note: It's not just `moveit_resources` that needs to be updated for the tutorials to work. We also need to do this for the [`ros2_kortex`](https://github.com/Kinovarobotics/ros2_kortex) repo, which I've since updated myself. Here's a PR: https://github.com/Kinovarobotics/ros2_kortex/pull/187

Maybe just include this as well in this PR description somewhere so we don't lose track?

MIGRATION.md
```diff
@@ -3,19 +3,37 @@
...
+```
+- [2021] `lockSceneRead()` and `lockSceneWrite()` are now protected member functions, for internal use only. To lock the planning scene, use LockedPlanningSceneRO or LockedPlanningSceneRW:
 ```
       planning_scene_monitor::LockedPlanningSceneRO ls(planning_scene_monitor);
       moveit::core::RobotModelConstPtr model = ls->getRobotModel();
 ```
-- ServoServer was renamed to ServoNode
-- `CollisionObject` messages are now defined with a `Pose`, and shapes and subframes are defined relative to the object's pose. This makes it easier to place objects with subframes and multiple shapes in the scene. This causes several changes:
+- [2021] ServoServer was renamed to ServoNode
```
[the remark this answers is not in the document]
sjahr: Maybe :sweat_smile: I feel like this file is not well maintained

moveit_core/utils/include/moveit/utils/moveit_status.h
```diff
@@ -32,35 +32,31 @@
...
+{
+  // Error code - encodes the overall reason for failure
+  MoveItErrorCode error_code;
+  // A message to provide additional information
+  std::string message;
+  // Name of the component that created the status
+  std::string source;
 
-  void initialize(const rclcpp::Node::SharedPtr& /* node */, const std::string& /* parameter_namespace */) override
+  MoveItStatus(const int code = 0 /*UNDEFINED*/, const std::string& message = std::string(),
```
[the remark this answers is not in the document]
sjahr: :point_up: Undefined does not exist (yet). I'll open a PR for it

moveit_ros/planning/planning_pipeline/res/planning_pipeline_parameters.yaml
```diff
@@ -2,10 +2,23 @@ planning_pipeline_parameters:
...
+    description: "Names of the planning request adapter plugins (plugin names separated by space). The order of the elements in the adapter corresponds to the order they are processed by the motion planning pipeline.",
+    read_only: true,
+    default_value: [],
+  }
+  progress_topic: {
     type: string,
-    description: "Names of the planning request adapter plugins (plugin names separated by space).",
-    default_value: "",
+    description: "For every stage of the planning pipeline a progress message is published to this topic. An empty string disables the publisher.",
+    default_value: "pipeline_state",
```
sjahr, line 23: Should this be disabled by default?

Abishalini: I would like it to be enabled by default. Otherwise, new users might find it hard to figure out such a feature exists.

moveit_core/utils/include/moveit/utils/moveit_status.h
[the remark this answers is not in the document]
sjahr: https://github.com/ros-planning/moveit_msgs/pull/172

moveit_core/planning_interface/include/moveit/planning_interface/planning_request_adapter.h
```diff
@@ -0,0 +1,75 @@
...
+/** @brief Concept in MoveIt which can be used to modify the planning problem(pre-processing) in a planning pipeline.
+ * PlanningRequestAdapters enable adjusting to or validating a planning problem for a subsequent planning algorithm.
+ */
+class PlanningRequestAdapter
+{
+public:
+  /** @brief Initialize parameters using the passed Node and parameter namespace.
+   *  @param node Node instance used by the adapter
+   *  @param parameter_namespace Parameter namespace for adapter
+      @details If no initialization is needed, simply implement as empty */
```
sea-bass, line 60: I think you updated the code, but the comment should be updated too to reflect that if no initialization is needed, a default empty implementation already is provided.

moveit_core/planning_interface/include/moveit/planning_interface/planning_response_adapter.h
```diff
@@ -0,0 +1,74 @@
...
+ * PlanningResponseAdapters enable using for example smoothing and trajectory generation algorithms in sequence to
+ * produce a trajectory.
+ */
+class PlanningResponseAdapter
+{
+public:
+  /** @brief Initialize parameters using the passed Node and parameter namespace.
+   *  @param node Node instance used by the adapter
+   *  @param parameter_namespace Parameter namespace for adapter
+      @details If no initialization is needed, simply implement as empty */
```
sea-bass, line 58: I think you updated the code, but the comment should be updated too to reflect that if no initialization is needed, a default empty implementation already is provided.

moveit_core/utils/include/moveit/utils/moveit_status.h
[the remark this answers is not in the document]
sea-bass: Cool, I approved that PR so feel free to make the change here afterwards

moveit_ros/planning/planning_request_adapter_plugins/src/resolve_constraint_frames.cpp
```diff
@@ -32,52 +32,51 @@
  *  POSSIBILITY OF SUCH DAMAGE.
  *********************************************************************/
 
-/* Author: Robert Haschke */
+/* Author: Robert Haschke
+   Desc: This adapter changes the link_name field of a constraint from an object's (sub-)frame name to the name of the
+   robot link, that object is attached to. Transforming the frame names is necessary because the frames of an attached
+   object are not know to a planner.
```
sea-bass, line 38: typos

```suggestion
   Desc: This adapter changes the link_name field of a constraint from an object's (sub-)frame name to the name of the
   robot link, that the object is attached to. Transforming the frame names is necessary because the frames of an attached
   object are not known to a planner.

**Review by sea-bass — COMMENTED** [review]
I can only see small things with comments at this point. Once your small PR to `moveit_msgs` is merged and you make that change here, LGTM

File: moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp [hunk 29]
```diff
@@ -204,172 +212,77 @@ bool planning_pipeline::PlanningPipeline::generatePlan(const planning_scene::Pla
...
-      solved = context ? context->solve(res) : false;
+          planner_instance_->getPlanningContext(planning_scene, mutable_request, res.error_code);
+      if (context)
+      {
+        RCLCPP_ERROR(node_->get_logger(),
+                     "Failed to create PlanningContext for planner '%s'. Aborting planning pipeline.",
+                     planner_instance_->getDescription().c_str());
+        res.error_code = moveit::core::MoveItErrorCode::PLANNING_FAILED;
+      }
+      context->solve(res);
```
Abishalini, line 256: I was testing this with MoveIt Studio and I see this error even when the plan is solved successfully.
I think this should be `if (!context) { ... }`

Abishalini:
Should we consider merging https://github.com/ros-planning/moveit_msgs/pull/171 and remove the MoveItStatus struct?

moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
sjahr: Good catch thanks!

sea-bass (approved):
Let's do this! 🚢

Issue #2376, opened 2023-09-19
Fix loggers to publish to /rosout


henningkayser:
**Is your feature request related to a problem? Please describe.**
Currently, basically none of our loggers are associated with a ros node. The pattern we decided [here](https://github.com/ros-planning/moveit2/blob/main/doc/MIGRATION_GUIDE.md) will create detached loggers for which no rosout publishers are initiated. See https://github.com/ros2/rclcpp/issues/1694#issuecomment-1453492628

**Describe the solution you'd like**
I'd like to initialize all loggers from rclcpp nodes, possibly implement child nodes wherever we have child logger namespaces.

If we do that, we'll have the issue that rclcpp::Node needs to be passed to all classes and functions that require logging, requiring many additional API changes and extensions. Question is, can we find a solution that allows us to hide this dependency (possibly like console_bridge), and even pick different logger backends?

tylerjw:
Another thing we should consider changing is [enabling the logging service to enable changing the log level via ros service](https://docs.ros.org/en/rolling/Tutorials/Demos/Logging-and-logger-configuration.html#logger-level-configuration-externally).

tylerjw:
When investigating if there is a way we can send logs to rosout that are not associated with nodes I found this quote in a Discourse thread from 2018:

> Because of the way ROS 2 associates DDS concepts with ROS Nodes, there didn’t seem to be a clean way to setup a general topic on a process that wasn’t associated with an existing Node. Since a process can also contain multiple ROS Nodes, we didn’t think it would be good design to just pick one to have everything published to.

Source: https://discourse.ros.org/t/ros2-logging/6469?page=2

tylerjw:
To fix the easy things first and explore the patterns we want to use, I'm starting making PRs from main functions. I'm working on the warehouse package in moveit_ros and will submit PRs. Here is the list of PRs with notes.

- #2432 - simplest case, only a main
- #2433 - main and free functions, passing logger as parameter

tylerjw:
Here is a new approach: #2445

tylerjw:
Here we go with more complexities.

In humble child loggers `node->get_logger().get_child("child");` under a node logger are not published to `/rosout`. Here are references with the details:

- https://github.com/ros2/rclpy/issues/1131
- https://github.com/ros2/rcl/pull/921
- https://github.com/ros2/rclpy/pull/1084

[...]

PR #2433, opened 2023-10-11
Use node logging in warehouse import_from_text

tylerjw:
### Description

Similar to #2432 but slightly more complex.

In this one we are only dealing with free functions that log, so they don't need a copy of the logger. For this reason, I'm passing in the logger as a reference argument.

Continuing work on https://github.com/ros-planning/moveit2/issues/2376 refactoring.


tylerjw:
Closing in favor of #2445

PR #2434, opened 2023-10-11
Use node logging in warehouse initialize_demo_db

tylerjw:
### Description

The simplest case, only a main. #2376


tylerjw:
Closing in favor of #2445

PR #2445, opened 2023-10-12
Node logger through singleton (warehouse)

tylerjw:
### Description

This creates a function in moveit utils for getting the logger and demonstrates using it by converting the whole warehouse package.


moveit_core/utils/include/moveit/utils/logger.hpp
```diff
@@ -0,0 +1,51 @@
...
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Tyler Weaver */
+
+#pragma once
+
+#include <rclcpp/rclcpp.hpp>
```
Abishalini, line 39:
```suggestion
#include <rclcpp/logger.hpp>
```

moveit_ros/warehouse/src/broadcast.cpp
```diff
@@ -73,7 +75,7 @@ int main(int argc, char** argv)
   node_options.allow_undeclared_parameters(true);
   node_options.automatically_declare_parameters_from_overrides(true);
   rclcpp::Node::SharedPtr node = rclcpp::Node::make_shared("publish_warehouse_data", node_options);
-  const auto logger = node->get_logger();
+  moveit::get_logger_mut() = node->get_logger();
```
Abishalini, line 78: I'm not following this. 
`moveit::get_logger_mut()` returns this static logger right?
`static rclcpp::Logger logger = rclcpp::get_logger("moveit");`

How are we able to assign value?

tylerjw: This works because get_logger_mut returns a reference to the static variable. Usually, it would be really bad to return a reference (or pointer) to a local variable, but in this case, it is static, which means it has a global lifetime. The first time the function is called, it gets set to the return of that get_logger function from ros, which is a logger that is not associated with a node; we can then assign to the static from the node logger.

I will write some tests today to verify this works as I think it does.

Abishalini: What will happen if there are multiple nodes modifying the logger? Would there be a case where the wrong logger instance is used?

tylerjw: If multiple nodes are created, and this function gets called multiple times, it will have the value of the second one. I don't know of any-place in moveit where that would happen, but it is one of the downsides to this approach.

tylerjw:
@clalancette here is my current attempt to get rosout logging working in moveit without changing API everywhere passing around node or logger objects.

Note that this is just changing one library but if we decide to go this way I'm going to apply this to the whole repo.

If you have a chance to look at this I'd really appreciate it. I know you said there might be some way to hook up rosout logging without nodes? Do you have an example somewhere I could read?

moveit_core/utils/src/logger.cpp
```diff
@@ -0,0 +1,77 @@
...
+  // Request for backport (rclpy issue) - https://github.com/ros2/rclpy/issues/1131
+  // MoveIt PR that added this - https://github.com/ros-planning/moveit2/pull/2445
+#if !RCLCPP_VERSION_GTE(21, 0, 3)
+  static std::vector<std::shared_ptr<rclcpp::Node>> child_nodes;
+  std::string ns = get_logger().get_name();
+  child_nodes.push_back(std::make_shared<rclcpp::Node>(name, ns));
+#endif
+
+  auto logger = get_logger_mut().get_child(name);
+  return logger;
```
tylerjw, line 67:
```suggestion
  return get_logger_mut().get_child(name);
```

moveit_core/utils/src/logger.cpp
```diff
@@ -0,0 +1,77 @@
...
+rclcpp::Logger make_child_logger(const std::string& name)
+{
+  // On versions of ROS older than Iron we need to create a node for each child logger
+  // Remove once Humble is EOL
+  // References:
+  // Use parent logger (rcl PR) - https://github.com/ros2/rcl/pull/921
+  // Request for backport (rclpy issue) - https://github.com/ros2/rclpy/issues/1131
+  // MoveIt PR that added this - https://github.com/ros-planning/moveit2/pull/2445
+#if !RCLCPP_VERSION_GTE(21, 0, 3)
+  static std::vector<std::shared_ptr<rclcpp::Node>> child_nodes;
```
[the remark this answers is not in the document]
tylerjw: This is a good suggestion, I'll do this.

moveit_core/utils/include/moveit/utils/logger.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+namespace moveit
+{
+
+// Function for getting a reference to a logger object kept in a global variable.
+// Use get_logger_mut to set the logger to a node logger.
+const rclcpp::Logger& get_logger();
+
+// Function for getting a child logger. In Humble this also creates a node.
+// Do no use this in place as it will create a new logger each time,
```
henningkayser, line 50:
```suggestion
// Do not use this in place as it will create a new logger each time,
```

henningkayser (approved):

PR #2449, opened 2023-10-13
CI job for building docker images for tutorial on main

tylerjw:
### Description

Here is the docker job we talked about yesterday. This should build a docker image on the main branch in CI for rolling, humble, and iron and push to the github container registry on pushes to main. The docker tag will be `ghcr.io/ros-planning/moveit2:main-tutorial-${ROS_DISTRO}`. I made it a separate workflow from the docker one because I want this to also run in PRs as a CI job and separate from the normal CI one because I only want it to run on pushes to the main branch.


MarqRazz (approved):
I locally built this image on my machine and it works! Thanks @tylerjw!

PR #2451, opened 2023-10-13
Rename tutorial docker

tylerjw:
### Description

main-tutorial-humble-tutorial-source -> main-humble-tutorial-source

### DIFF
--- a/.docker/tutorial-source/Dockerfile
+++ b/.docker/tutorial-source/Dockerfile
@@ -1,6 +1,6 @@
 # syntax = docker/dockerfile:1.3
 
-# ghcr.io/ros-planning/moveit2:main-tutorial-${ROS_DISTRO}
+# ghcr.io/ros-planning/moveit2:main-${ROS_DISTRO}-tutorial-source
 # Source build of the repos file from the tutorail site
 
 ARG ROS_DISTRO=rolling
--- a/.github/workflows/tutorial_docker.yaml
+++ b/.github/workflows/tutorial_docker.yaml
@@ -21,8 +21,8 @@ jobs:
       packages: write
       contents: read
     env:
-      GH_IMAGE: ghcr.io/ros-planning/moveit2:main-tutorial-${{ matrix.ROS_DISTRO }}-${{ github.job }}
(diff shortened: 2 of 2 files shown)


.docker/tutorial-source/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # syntax = docker/dockerfile:1.3
 
-# ghcr.io/ros-planning/moveit2:main-tutorial-${ROS_DISTRO}
+# ghcr.io/ros-planning/moveit2:main-${ROS_DISTRO}-tutorial-source
```
MarqRazz, line 3: Before this change the name was `ghcr.io/ros-planning/moveit2:main-tutorial-humble-tutorial-source`

Reading this I would expect the new name to be `ghcr.io/ros-planning/moveit2:main-humble-tutorial-source-tutorial-source` or am I missing something?

tylerjw: this is just a comment about what it is, ROS_DISTRO just expands to the ros distro name

.github/workflows/tutorial_docker.yaml
```diff
@@ -21,8 +21,8 @@ jobs:
       packages: write
       contents: read
     env:
-      GH_IMAGE: ghcr.io/ros-planning/moveit2:main-tutorial-${{ matrix.ROS_DISTRO }}-${{ github.job }}
-      DH_IMAGE: moveit/moveit2:main-tutorial-${{ matrix.ROS_DISTRO }}-${{ github.job }}
+      GH_IMAGE: ghcr.io/ros-planning/moveit2:main-${{ matrix.ROS_DISTRO }}-${{ github.job }}
+      DH_IMAGE: moveit/moveit2:main-${{ matrix.ROS_DISTRO }}-${{ github.job }}
```
tylerjw, line 25: @MarqRazz here is the real technical change.

ROS_DISTRO expands to `humble`, `rolling`, or `iron`
github.job expands to `tutorial-source`

so a full name will in the form `main-humble-tutorial-source`

.docker/tutorial-source/Dockerfile
MarqRazz: I know the code only appends the ROS_DISTRO but where does it append the extra `-tutorial-source` at the end?

MarqRazz (approved):

PR #2453, opened 2023-10-14
Do not pass and return simple types by const ref

### LINKED ISSUE #859: Do not pass and return simple types by const ref

Shobuj-Paul:
### Description

Fixes #859 
Continues the work from #2119 

### Checklist

### DIFF
--- a/moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/bullet_integration/basic_types.h
+++ b/moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/bullet_integration/basic_types.h
@@ -59,7 +59,7 @@ struct ContactTestData
 {
   EIGEN_MAKE_ALIGNED_OPERATOR_NEW
 
-  ContactTestData(const std::vector<std::string>& active, const double& contact_distance,
+  ContactTestData(const std::vector<std::string>& active, const double contact_distance,
                   collision_detection::CollisionResult& res, const collision_detection::CollisionRequest& req)
     : active(active), contact_distance(contact_distance), res(res), req(req), done(false), pair_done(false)
   {
@@ -68,7 +68,7 @@ struct ContactTestData
   const std::vector<std::string>& active;
 
   /** \brief If after a positive broadphase check the distance is below this threshold, a contact is added. */
-  const double& contact_distance;
+  const double contact_distance;
 
   collision_detection::CollisionResult& res;
   const collision_detection::CollisionRequest& req;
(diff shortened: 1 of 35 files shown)


Shobuj-Paul:
@tylerjw Please check if this is fine.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/bullet_integration/basic_types.h
```diff
@@ -68,7 +68,7 @@ struct ContactTestData
   const std::vector<std::string>& active;
 
   /** \brief If after a positive broadphase check the distance is below this threshold, a contact is added. */
-  const double& contact_distance;
```
ChrisThrasher, line 71: This has the the potential to change semantics since a reference to this double is prone to be changed by the calling code.

moveit_ros/perception/mesh_filter/src/gl_renderer.cpp
```diff
@@ -246,12 +246,12 @@ const GLuint& mesh_filter::GLRenderer::getProgramID() const
   return program_;
 }
 
-const double& mesh_filter::GLRenderer::getNearClippingDistance() const
+double mesh_filter::GLRenderer::getNearClippingDistance() const
 {
   return near_;
 }
 
-const double& mesh_filter::GLRenderer::getFarClippingDistance() const
```
ChrisThrasher, line 254: This also has the potential to change semantics and can potentially be an API break if such a change happens in a public function. It's valid to store this return value as a `const&` but that becomes invalid after this transformation. This is not a big problem per se but such things ought to be considered.

ChrisThrasher (commented):
In general this PR does good stuff. Changing `const double&` to `double` in function declarations is a good change. Performance wise it makes no difference since doubles are 64 bits and we're all using 64 bit computers, but it's certainly a bit odd to see these fundamentally types used via a const ref. Likewise returning a const ref to a built in type is weird and doesn't solve any problems, generally speaking. It's good to avoid such things.

Let's be careful though. By changing `const double&` to `double` in a function _definition_ we now are opening up the ability to mutate data was previously immutable which means some amount of intent is lost and it's a bit easier for someone to make potentially invalid changes to the code in the future. While adding `const` to a non-reference parameter is meaningless in a function declaration, it does carry meaning in the function definition.

Also when changing `const double&` return values to `double` we run the risk of API breaks if users are storing the return value in another `const double&`. Let's keep that in mind in case any public functions are being changed.

```cpp
const double& value = someMoveItFunction(); // This breaks when the return type is not longer a reference
```

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/bullet_integration/basic_types.h
Shobuj-Paul: Reverted back.

ChrisThrasher (commented):
No more comments to make. Dismissing the review request.

tylerjw (approved):

PR #2454, opened 2023-10-14
Use std::optional instead of nullptr checking

### LINKED ISSUE #2131: Replace the nullpointer checking in the Planning Scene with std::optional
Please ping @henningkayser or @tylerjw before starting this issue! It's just an idea.

Stuff like [this](https://github.com/ros-planning/moveit2/blob/acbf3770763e58e4f86ce0dea251a50cd4987fe6/moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h#L1035):

`std::unique_ptr<ObjectTypeMap> object_types_;`

```
  if (!object_types_)
  {
    ObjectTypeMap kc;
    parent_->getKnownObjectTypes(kc);
    object_types_ = ObjectTypeMap(kc);
  }
  else
  {
    ...
```

Ought to be replaced with std::optional<ObjectTypeMap>.

#2132 is a good start at this that seems like it could be extended for:

  scene_transforms_, robot_state_,  acm_, object_types_

Shobuj-Paul:
### Description

Wrapped scene_transforms_, robot_state_, acm_ and object_types_ in std::optional to handle missing data instead of checking for nullptr.
Fixes #2131 

### Checklist

### DIFF
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -1273,7 +1273,7 @@ bool PlanningScene::setPlanningSceneMsg(const moveit_msgs::msg::PlanningScene& s
 
   if (parent_)
     decoupleParent();
-  
+
   object_types_.reset();
   scene_transforms_->setTransforms(scene_msg.fixed_frame_transforms);
   setCurrentState(scene_msg.robot_state);
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -1191,7 +1191,7 @@ void PlanningScene::decoupleParent()
   {
     ObjectTypeMap kc;
     parent_->getKnownObjectTypes(kc);
-    object_types_.value() = std::make_unique<ObjectTypeMap>(kc);
+    object_types_.value_or(std::make_unique<ObjectTypeMap>(kc));
   }
   else
   {
@@ -1273,6 +1273,7 @@ bool PlanningScene::setPlanningSceneMsg(const moveit_msgs::msg::PlanningScene& s
 
   if (parent_)
     decoupleParent();
+  
   object_types_.reset();
   scene_transforms_->setTransforms(scene_msg.fixed_frame_transforms);
   setCurrentState(scene_msg.robot_state);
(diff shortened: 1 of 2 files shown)


moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -1262,7 +1262,7 @@ bool PlanningScene::setPlanningSceneMsg(const moveit_msgs::msg::PlanningScene& s
   if (parent_)
     decoupleParent();
 
-  object_types_.reset();
+  object_types_.value().reset();
```
sjahr, line 1265: Needs `has_value()` guard

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -1926,25 +1926,25 @@ const object_recognition_msgs::msg::ObjectType& PlanningScene::getObjectType(con
 
 void PlanningScene::setObjectType(const std::string& object_id, const object_recognition_msgs::msg::ObjectType& type)
 {
-  if (!object_types_)
-    object_types_ = std::make_unique<ObjectTypeMap>();
-  (*object_types_)[object_id] = type;
+  if (!object_types_.value())
```
sjahr, line 1929:
```suggestion
  if (!object_types_.has_value())
```

sjahr (commented):
Thanks for tackling this! Something is still not working since the tests are complaining. Can you fix that?

Shobuj-Paul:
> Thanks for tackling this! Something is still not working since the tests are complaining. Can you fix that?

The build is completing but the tests fail, working on sorting it out.

henningkayser (commented):
Most of these pointers are only pointers because they are optional. If we use std::optional, we should switch members to values as long as the API is only exposing references.

Shobuj-Paul:
> Most of these pointers are only pointers because they are optional. If we use std::optional, we should switch members to values as long as the API is only exposing references.

I have changed object_types_, robot_state_ and acm_ to values.
scene_transforms_ is giving some issues, here were are trying to assign it a SceneTransforms object, but here the object has been defined as the parent class Transforms object. The pointer implementation seems to be casting it, but I don't understand the reason behind doing this.

Shobuj-Paul:
@sjahr 
Could you review this PR?

sjahr (approved):
I like this change :+1: Thanks and sorry for the late response

PR #2455, opened 2023-10-16
Map ompl's APPROXIMATE_SOLUTION -> TIMED_OUT / PLANNING_FAILED

rhaschke:
ompl's `APPROXIMATE_SOLUTION` is not suitable for actual execution. It just states that we [got closer to the goal](https://ompl.kavrakilab.org/classompl_1_1base_1_1ProblemDefinition.html#a612902cacc3b1a2bcc367a9e82739132):

> If the top found solution is approximate, does not actually reach the desired goal, but hopefully is closer to it

As the most prominent reason for an approximate solution is a timeout, also checking for this case.

Fixes https://github.com/ros-planning/moveit_task_constructor/issues/485

### DIFF
--- a/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
+++ b/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
@@ -860,7 +860,7 @@ const moveit_msgs::msg::MoveItErrorCodes ompl_interface::ModelBasedPlanningConte
     RCLCPP_DEBUG(LOGGER, "%s: Solving the planning problem once...", name_.c_str());
     ob::PlannerTerminationCondition ptc = constructPlannerTerminationCondition(timeout, start);
     registerTerminationCondition(ptc);
-    result.val = ompl_simple_setup_->solve(ptc) == ompl::base::PlannerStatus::EXACT_SOLUTION;
+    std::ignore = ompl_simple_setup_->solve(ptc);
     last_plan_time_ = ompl_simple_setup_->getLastPlanComputationTime();
     unregisterTerminationCondition();
     // fill the result status code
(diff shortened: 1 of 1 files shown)


rhaschke:
Interestingly the issue was fixed in the MoveIt1 backport (https://github.com/ros-planning/moveit/pull/3257#pullrequestreview-1163396768), but then not back-ported to MoveIt2 again!

moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
```diff
@@ -1000,8 +1000,17 @@ int32_t ompl_interface::ModelBasedPlanningContext::logPlannerStatus(const og::Si
       result = moveit_msgs::msg::MoveItErrorCodes::TIMED_OUT;
       break;
     case ompl::base::PlannerStatus::APPROXIMATE_SOLUTION:
-      RCLCPP_WARN(LOGGER, "Solution is approximate");
-      result = moveit_msgs::msg::MoveItErrorCodes::SUCCESS;
+      // timeout is a common reason for APPROXIMATE_SOLUTION
+      if (ompl_simple_setup->getLastPlanComputationTime() > request_.allowed_planning_time)
+      {
+        RCLCPP_WARN(LOGGER, "Planning timed out");
```
Abishalini, line 1006: Nit: Can we print the planning timeout value here along with the `Planning timed out` message?
```suggestion
        RCLCPP_WARN(LOGGER, "Planning timed out. Allowed planning time is currently set to %f", request_.allowed_planning_time
```

Abishalini (approved):
Thanks for this fix! 
I tested with a setup that often executed partial solutions and this fixes it. I had to increase the planning time in my setup.

henningkayser (approved):

leander2189:
Thanks for the fix!!

What is the procedure to get this PR merged also into **humble** branch? I recall there is a bot that does that, but have no idea how to invoke it...

rhaschke:
@mergifyio backport humble

rhaschke:
> What is the procedure to backport? I recall there is a bot that does that, but have no idea how to invoke it...

Usually, it's done by providing the label `backport-humble` (before merging), which I forgot to do...
However, one can manually trigger a backport by providing a comment (see https://github.com/ros-planning/moveit2/pull/2455#issuecomment-1768315627). I will merge the backport if CI succeeds.

Note, that we will need a new release to actually have these changes deployed.

leander2189:
> Note, that we will need a new release to actually have these changes deployed.

Yes, I understand. Thank you

leander2189:
Just to note that these changes are still not deployed in latest binary version from apt repos (2.5.5-1jammy.20240217.122953).

PR #2456, opened 2023-10-16
Port #3464 from MoveIt1

rhaschke:
https://github.com/ros-planning/moveit/pull/3464 fixed an issue in the numbering of added path indices.
This PR ports this fix to MoveIt2. As the `added_path_index` vector was moved into the `MotionPlanResponse` struct in #2285, I had to choose a slightly different approach.
The first commit ports the (failing) unit test.
The second commit fixes the bug (and the test).

### DIFF
--- a/moveit_core/planning_request_adapter/test/test_planning_request_adapter_chain.cpp
+++ b/moveit_core/planning_request_adapter/test/test_planning_request_adapter_chain.cpp
@@ -1,0 +1,204 @@
+/*********************************************************************
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2023, Willow Garage, Inc.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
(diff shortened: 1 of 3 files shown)


moveit_core/planning_request_adapter/test/test_planning_request_adapter_chain.cpp
```diff
@@ -0,0 +1,204 @@
...
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Hugo Laloge */
+
+#include <gmock/gmock.h>
+#include <moveit/utils/robot_model_test_utils.h>
+#include <moveit/planning_request_adapter/planning_request_adapter.h>
+
+class AppendingPlanningRequestAdapter final : public planning_request_adapter::PlanningRequestAdapter
```
sjahr, line 41: Would you mind adding brief documentation for the classes and the test in this file?

sjahr (approved):
Thanks for porting this over! Just a small comment not blocking comment

moveit_core/planning_request_adapter/test/test_planning_request_adapter_chain.cpp
rhaschke: I don't have time to do this. Maybe, @Hugal31 can do so?
On the other hand, these classes are rather trivial and the names already state their semantics quite well.

PR #2482, opened 2023-10-24
Use node logging in moveit_ros

tylerjw:
### Description

I updated the logger util library with some changes

I removed the assigning into the return of a function. It looks weird and I replaced it with a `set_logger` function.

I added more static state so if `get_logger` is called before `set_logger` it will still work as it will create a `moveit` node. There might be a better way to do this. I did this so child loggers created without creating the base node would work. I added a test for the child logger without first assigning the node logger.

### DIFF
--- a/moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
+++ b/moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
@@ -42,21 +42,22 @@
 #include <rclcpp/parameter_value.hpp>
 #include <rclcpp/time.hpp>
 #include <typeinfo>
+#include <moveit/utils/logger.hpp>
 
 namespace robot_model_loader
 {
-static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_ros.robot_model_loader");
 
 RobotModelLoader::RobotModelLoader(const rclcpp::Node::SharedPtr& node, const std::string& robot_description,
                                    bool load_kinematics_solvers)
-  : node_(node)
+  : node_(node), logger_(moveit::makeChildLogger("robot_model_loader"))
 {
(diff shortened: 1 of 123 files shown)


tylerjw:
#2376

moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
```diff
@@ -75,23 +76,23 @@ RobotModelLoader::~RobotModelLoader()
 
 namespace
 {
-bool canSpecifyPosition(const moveit::core::JointModel* jmodel, const unsigned int index)
+bool canSpecifyPosition(const moveit::core::JointModel* jmodel, const unsigned int index, const rclcpp::Logger& logger)
```
henningkayser, line 79: ouch

henningkayser (approved):
Very clean refactoring. The changes are much less intrusive than I thought. Please update https://github.com/ros-planning/moveit2/blob/main/doc/MIGRATION_GUIDE.md before merging.

moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
tylerjw: this is one of the trade-offs of carrying around state everywhere for logging. for the child loggers you have to pass them around (either by putting them in shared state in a class or by using function parameters).

tylerjw: For free functions that are public I think we should just use the global node logger and users can use `{function}` to filter for those logs.

For functions like this one that are in the private namespace of cpp file, adding parameters to it has no affect on the public api and seems fine to me.

PR #2507, opened 2023-11-01
Remove boost lexical cast

### LINKED ISSUE #2088: Replace boost::lexical_cast in OMPL inteface
While looking into enabling AnytimePathShortening I encountered an error when I tried to use a ROS 2 param like this: `hybridize: true`. Boost lexical cast does not seem to be capable of casting the string 'true' to a boolean value which is why initializing the parameter like this leads to a runtime error [here](https://github.com/ros-planning/moveit2/blob/0e54c83f0d618b32b786541f5c686d5ac6d36e4b/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp#L389). The immediate fix is to use `hybridize: 1` instead.

The reason why `lexical_cast` is required at all is that all parameters (independent of the type) for a specific planner should be stored in a container, so we cast all parameters to string [here](https://github.com/ros-planning/moveit2/blob/0e54c83f0d618b32b786541f5c686d5ac6d36e4b/moveit_planners/ompl/ompl_interface/src/ompl_interface.cpp#L151) and transform them back into the original type with `lexical_cast`. Instead of using a parameter map <std::string, std::string> I'd propose to use map<std::string, [std::any](https://en.cppreference.com/w/cpp/utility/any)>. This way we would keep the generic parameter structure to be planner agnostic but get rid of boost.
@tylerjw @henningkayser What do you think?

Shobuj-Paul:
### Description

Used std::any as data type in std::map to store different params without casting to string to avoid using lexical_cast.
Fixes #2088 

### Checklist

### DIFF
--- a/moveit_core/planning_interface/include/moveit/planning_interface/planning_interface.h
+++ b/moveit_core/planning_interface/include/moveit/planning_interface/planning_interface.h
@@ -42,6 +42,7 @@
 #include <rclcpp/node.hpp>
 #include <string>
 #include <map>
+#include <any>
 #include <rclcpp/rclcpp.hpp>
 
 namespace planning_scene
@@ -69,7 +70,7 @@ struct PlannerConfigurationSettings
   std::string name;
 
   /** \brief Key-value pairs of settings that get passed to the planning algorithm */
-  std::map<std::string, std::string> config;
+  std::map<std::string, std::any> config;
 };
 
 /** \brief Map from PlannerConfigurationSettings.name to PlannerConfigurationSettings */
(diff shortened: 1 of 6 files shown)


Shobuj-Paul:
Tried doing as was mentioned in the issue, but faced a few problems. Specifically [this](https://github.com/Shobuj-Paul/moveit2/blob/b7be81474859d1ea60a8237fc58bf9b6daf56961/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp#L422) function `setParams()` is in the core OMPL library, hence I cannot change its argument data type. Please suggest how to proceed.

Shobuj-Paul:
This solution is causing more problems than it is solving, looking for a different approach for the corresponding issue.

PR #2511, opened 2023-11-06
Add distance to servo collision checker requests

sea-bass:
We found a bug today where the collision distances were not actually being returned by servo's collision checker. So basically, the distance scaling was never working and you could servo right into a binary "is colliding" state.

For reference, these parameters were on before the refactor: https://github.com/ros-planning/moveit2/blob/humble/moveit_ros/moveit_servo/src/collision_check.cpp#L62-L64

@ibrahiminfinite

### DIFF
--- a/moveit_ros/moveit_servo/src/collision_monitor.cpp
+++ b/moveit_ros/moveit_servo/src/collision_monitor.cpp
@@ -50,6 +50,13 @@ CollisionMonitor::CollisionMonitor(const planning_scene_monitor::PlanningSceneMo
   , planning_scene_monitor_(planning_scene_monitor)
   , collision_velocity_scale_(collision_velocity_scale)
 {
+  scene_collision_request_.distance = true;
+  scene_collision_request_.contacts = true;
+  scene_collision_request_.group_name = servo_params.move_group_name;
+
+  self_collision_request_.distance = true;
+  self_collision_request_.contacts = true;
+  self_collision_request_.group_name = servo_params.move_group_name;
 }
 
 void CollisionMonitor::start()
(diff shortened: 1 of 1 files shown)


moveit_ros/moveit_servo/src/collision_monitor.cpp
```diff
@@ -50,6 +50,13 @@ CollisionMonitor::CollisionMonitor(const planning_scene_monitor::PlanningSceneMo
   , planning_scene_monitor_(planning_scene_monitor)
   , collision_velocity_scale_(collision_velocity_scale)
 {
+  scene_collision_request_.distance = true;
+  scene_collision_request_.contacts = true;
```
AndyZe, line 54: I believe `contacts` is optional unless you want to print the colliding pair (please test, though)

```suggestion
  scene_collision_request_.contacts = false;
```

moveit_ros/moveit_servo/src/collision_monitor.cpp
```diff
@@ -50,6 +50,13 @@ CollisionMonitor::CollisionMonitor(const planning_scene_monitor::PlanningSceneMo
...
+  scene_collision_request_.group_name = servo_params.move_group_name;
+
+  self_collision_request_.distance = true;
+  self_collision_request_.contacts = true;
```
AndyZe, line 58:
```suggestion
  self_collision_request_.contacts = false;
```

moveit_ros/moveit_servo/src/collision_monitor.cpp
sea-bass: it is optional (it actually works without) -- I put it there because it was also on in Humble, but I can get rid of it if you prefer.

sea-bass: I will say it will be useful if we ever try to implement something that uses the contact pairs to determine whether the velocity is towards/away from collision... but not needed for now.

AndyZe (approved):

ibrahiminfinite:
Oh no!
I just remembered I haven't added tests for collision checking. I had planned to add it but must have forgot.

ibrahiminfinite:
But I also remember collision  scaling actually working since I had tested it manually and the robot did slow down and stop without colliding.

sea-bass:
The stopping part did work, but the distance always returned some large max value because it was not being returned. So basically the scaling factor would go directly from 1.0 to 0.0 once two links were in actual collision.

Issue #2412, opened 2023-10-07
Move `validateParams()` out of the Servo control loop


AndyZe:
As mentioned here:   https://github.com/ros-planning/moveit2/pull/2396#discussion_r1349465881

`validateParams()` is expensive. It doesn't belong in a control loop. Some options are:

- Move the parameter loading out of the Servo class completely
- Only validate parameters that have changed at runtime. Not sure if this is possible
- Not validate parameters at all after initialization. Assume the user knows what they're doing

ibrahiminfinite:
> Not validate parameters at all after initialization. Assume the user knows what they're doing

I prefer this way if it is acceptable, can avoid some complexity as well.

PR #2515, opened 2023-11-08
[Servo] Change planning frame to base frame of active joint subgroup

sea-bass:
### Description

This PR seeks to solve an issue in which changing the `active_subgroup` parameter still enforces using the `servo_params.planning_frame` planning frame... which is not necessarily the correct frame for all subgroups.

So instead, we're grabbing the planning frame (for twist/pose commands) to always be the base frame of the IK solver and doing the frame conversions from there. This fixes the immediate issue, but the implementation leaves much to be desired.


moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -369,10 +373,21 @@ Eigen::VectorXd Servo::jointDeltaFromCommand(const ServoInput& command, const mo
...
       try
       {
-        const TwistCommand command_in_planning_frame = toPlanningFrame(std::get<TwistCommand>(command));
-        delta_result =
-            jointDeltaFromTwist(command_in_planning_frame, robot_state, servo_params_, joint_name_group_index_map);
-        servo_status_ = delta_result.first;
+        const auto ik_solver = active_joint_model_group->getSolverInstance();
+        if (ik_solver)
+        {
+          const auto planning_frame = ik_solver->getBaseFrame();
```
sea-bass, line 379: Is there a better way to get the base link of a joint model group without having to grab the IK solver like this?

marioprats: Can't you do this?

```
const JointModel* root_joint_model = group->getJointModels().front();
const LinkModel* root_link_model = root_joint_model->getParentLinkModel();
```

But now I wonder if you rather need the IK solver base frame (i.e. your code as is), since the underlying IK solver could be working on a different frame, not necessarily the group root?
If that's the case, maybe create a `getIKSolverBaseFrame()` function that can be called here and below?

sea-bass: Thanks!

This raises a follow up question that, if these changes are the right thing for changing subgroups, can we actually get rid of the planning frame parameter and have servo always do its own internal conversions to active group's IK base frame?

marioprats: Hmm.. looking deeper into Servo I got more confused about what the 'planning frame' is, but also how this fix works, since the IK base frame for the manipulator group is the one tilted 45 deg. right? so I would expect this to still move the wrong way.
Another q: the command has a `frame_id`. Shouldn't that be all that is needed to move in the same frame (e.g. `world`), independently of the active group? Sorry I have more questions than answers!

sea-bass: You should have more questions than answers, I think there's a lot here that needs to be disentangled for clarity.

Basically, Servo has this rule that enforces that all commands passed into it are expressed w.r.t. a common frame -- this is the "planning frame" we speak of.

All we're trying to do, perhaps in an overly complex way, is the following:

* Receive a Cartesian command w.r.t. some arbitrary frame ID
* Convert it to a reference frame that agrees with the IK solver (this is our planning frame)
* ???
* Profit

sea-bass: Agreed, as I was thinking about this today as well.

I think this parameter was carryover from before and we can probably remove it if this is how we're going to handle things going forward.

sea-bass: We can probably also get rid of the `ee_frame` parameter if that's the case... given these descriptions:

```
  planning_frame: {
    type: string,
    description: "The name of moveit planning frame.\
                  This parameter does not have a default value and \
                  must be passed to the node during launch time."
  }

  ee_frame: {
    type: string,
    description: "The name of end effector frame of your robot, \
                  this should also be the IK tip frame of your IK solver. \
                  This parameter does not have a default value and \
                  must be passed to the node during launch time."
  }
```

AndyZe: I think the answer is, I didn't know about retrieving the IK solver base frame when this was coded 6-years-or-so ago. Automating it sounds nice.

BTW please tag myself and @ibrahiminfinite on Servo PR's

sea-bass: Glad to have your vote of confidence to help automate this, I wanted to ask you but didn't want to bug you guys too much with more servo issues until this was out of draft and in a direction we were sure we wanted to take.

ibrahiminfinite: I remember having some confusing discussions about these frames during the rewrite.
Also agree on the direction of removing the `planning_frame` and `ee_frame` parameters and fetching the right **planning frame**   internally in Servo. 
We already have the frame information REQUIRED from the user in the `frame_id` attribute of the command.

sea-bass: Was able to remove them!

Sadly on MoveIt Studio we rely on `moveit_params.ee_frame` to send a command... but that's our problem to go figure out now :)

sea-bass: Hmm... I just saw some other piece of code that reminded me of the fact that a planning group can have multiple tip frames. So can we really get rid of `moveit_params.ee_frame` after all?

marioprats: I think we can, at least the way it is now, because `ee_frame` doesn't seem to be used anywhere. The IK computation is done for `ik_solver->getTipFrame()`, so it will always use the first tip.
Longer term what about having Servo expose a `/configure` service, where the user could set the planning group and tip frame?

sea-bass: Given all the struggles we've had trying to make sure that parameters stay consistent w.r.t. each other, I quite like the idea of dedicated services that touch these interrelated parameters like planning group + ee frame.

Also, having this `configure()` service means that configuring can implicitly do the task of pausing servo, changing values, and then unpausing it.

Also, I guess to be fair this code was calling `getTipFrame()` already, and also there is an error logged when you call `getTipFrame()` on a planning group with multiple tips.

So maybe this is just a known limitation for now and we proceed as is?

moveit_ros/moveit_servo/demos/cpp_interface/demo_pose.cpp
```diff
@@ -88,9 +88,9 @@ int main(int argc, char* argv[])
 
   // The dynamically updated target pose.
   PoseCommand target_pose;
-  target_pose.frame_id = servo_params.planning_frame;
+  target_pose.frame_id = "panda_link0";
```
marioprats, line 91: maybe constexpr base_frame and tip_frame at the top of the file?

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
```diff
@@ -104,9 +104,11 @@ class Servo
   std::string getStatusMessage() const;
 
   /**
-   * \brief Returns the end effector pose in planning frame
+   * \brief Returns the current pose of a specified target frame.
+   * @param target_frame The name of the target frame.
+   * @return The global transform of the specified frame.
    */
-  Eigen::Isometry3d getEndEffectorPose() const;
+  Eigen::Isometry3d getCurrentPose(const std::string& target_frame) const;
```
marioprats, line 111: Given that the library is not using this, I would remove it from the API and move into the tests where this is needed.

moveit_ros/moveit_servo/include/moveit_servo/utils/common.hpp
```diff
@@ -54,6 +54,23 @@
 
 namespace moveit_servo
 {
+/**
+ * \brief Get the base frame of the current active joint group or subgroup's IK solver.
+ * @param robot_state A pointer to the current robot state.
+ * @param group_name The currently active joint group name.
+ * @return The IK solver base frame, if one exists, otherwise std::nullopt.
+ */
+std::optional<std::string> getIKSolverBaseFrame(const moveit::core::RobotStatePtr& robot_state,
```
marioprats, line 63: Can we avoid the pointer here and below and just do `const moveit::core::RobotState &robot_state`?

marioprats (commented):
This looks good to me. A couple of minor comments that can be done in follow-up changes too, since they are not really related to this PR changes.

moveit_ros/moveit_servo/demos/cpp_interface/demo_pose.cpp
sea-bass: done

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
sea-bass: Good call -- done!

moveit_ros/moveit_servo/include/moveit_servo/utils/common.hpp
sea-bass: I think the robot state which is extracted from

```
moveit::core::RobotStatePtr CurrentStateMonitor::getCurrentState() const
```

already returns in pointer form, so is it fine to keep as is?

marioprats: That's fine if this is only used in this context. If we ever put it somewhere more general I think we should just take the const&.
`shared_ptr` is viral, it easily spreads everywhere. If this takes a `shared_ptr`, it can't really can't be used unless you have a `shared_ptr` already. Whereas if this takes a `const&`, you can call it directly with a `RobotState` object, but also with a `shared_ptr` or `unique_ptr` by derefencing, so it's more usable.

marioprats (approved):

moveit_ros/moveit_servo/src/servo.cpp
AndyZe: A service to update the parameters might be a good idea b/c it would fix this issue:  https://github.com/ros-planning/moveit2/issues/2412

i.e. do the parameter validation in a callback in a different thread

Hate to change the API so drastically, so soon, though. Have to think carefully about this.

sea-bass: I think servo has changed enough with the refactor since we kept running into limitations trying to keep the API as stable as possible with the old version. It might be beneficial to do all the big changes as soon as possible in hopes that we get to a somewhat more stable API before people actually start using this new version.

That said, we should try do all this in a way that minimizes our need to keep twiddling the interfaces going forward.

moveit_ros/moveit_servo/demos/cpp_interface/demo_pose.cpp
```diff
@@ -51,6 +51,12 @@
 using moveit::getLogger;
 using namespace moveit_servo;
 
+namespace
+{
+constexpr auto K_BASE_FRAME = "panda_link0";
+constexpr auto K_TIP_FRAME = "panda_link8";
```
sjahr, line 57: Not in this PR, but we should move this demo to the tutorials. What do you think?

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -497,29 +501,42 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
...
-    return tf2::transformToEigen(planning_scene_monitor_->getTFClient()->lookupTransform(
-        servo_params_.planning_frame, command_frame, rclcpp::Time(0)));
+    try
+    {
+      return tf2::transformToEigen(
+          planning_scene_monitor_->getTFClient()->lookupTransform(planning_frame, command_frame, rclcpp::Time(0)));
+    }
+    catch (tf2::TransformException& ex)
+    {
+      return std::nullopt;
```
sjahr, line 522: Maybe print `ex.what()` with RCLCPP_ERROR here

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -561,13 +578,19 @@ TwistCommand Servo::toPlanningFrame(const TwistCommand& command) const
...
 
-PoseCommand Servo::toPlanningFrame(const PoseCommand& command) const
+std::optional<PoseCommand> Servo::toPlanningFrame(const PoseCommand& command, const std::string& planning_frame) const
 {
-  return PoseCommand{ servo_params_.planning_frame,
-                      getPlanningToCommandFrameTransform(command.frame_id) * command.pose };
+  const auto planning_to_command_tf_maybe = getPlanningToCommandFrameTransform(command.frame_id, planning_frame);
+  if (!planning_to_command_tf_maybe)
+  {
+    return std::nullopt;
```
sjahr, line 589: Print warning or error?

moveit_ros/moveit_servo/src/utils/command.cpp
```diff
@@ -250,8 +250,7 @@ JointDeltaResult jointDeltaFromPose(const PoseCommand& command, const moveit::co
     }
     if (!is_planning_frame)
     {
-      RCLCPP_WARN_STREAM(getLogger(),
-                         "Command frame is: " << command.frame_id << " expected: " << servo_params.planning_frame);
+      RCLCPP_WARN_STREAM(getLogger(), "Command frame is: " << command.frame_id << " expected: " << planning_frame);
```
sjahr, line 253: Isn't this an error? What is the resulting joint_position_delta in that case?

sjahr (commented):
Nice work! Most of my comments are around more error messages and documentations. Besides that I think it is good to go :+1:

moveit_ros/moveit_servo/demos/cpp_interface/demo_pose.cpp
sea-bass: I think we made the conscious decision to keep a few simple demos in the core repo itself during development of this project, so I'd rather keep as is until/unless we decide to do something sweeping with the whole repo and the way we do demos.

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -88,36 +88,23 @@ Servo::Servo(const rclcpp::Node::SharedPtr& node, std::shared_ptr<const servo::P
...
-    // Load the smoothing plugin
-    if (servo_params_.use_smoothing)
-    {
-      setSmoothingPlugin();
-    }
-    else
-    {
-      RCLCPP_WARN(logger_, "No smoothing plugin loaded");
-    }
+    RCLCPP_WARN(logger_, "No smoothing plugin loaded");
```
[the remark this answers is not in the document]
sea-bass: I don't think it's unrecoverable -- you can run servo without a smoothing plugin if you want, and this is just a warning that you're on your own re: jerky behavior.

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -367,32 +350,53 @@ Eigen::VectorXd Servo::jointDeltaFromCommand(const ServoInput& command, const mo
...
     {
-      try
+      const auto planning_frame_maybe = getIKSolverBaseFrame(robot_state, active_subgroup_name);
+      if (planning_frame_maybe.has_value())
       {
-        const TwistCommand command_in_planning_frame = toPlanningFrame(std::get<TwistCommand>(command));
-        delta_result =
-            jointDeltaFromTwist(command_in_planning_frame, robot_state, servo_params_, joint_name_group_index_map);
-        servo_status_ = delta_result.first;
+        const auto& planning_frame = *planning_frame_maybe;
```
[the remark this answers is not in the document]
sea-bass: You can, but this is standard to make it less awkward in follow-on calls. And because it's being extracted by reference, it's a pretty inexpensive operation.

moveit_ros/moveit_servo/src/utils/command.cpp
sea-bass: Yeah, nothing happens in this case (delta is zero) so it probably should be an error.

moveit_ros/moveit_servo/demos/cpp_interface/demo_pose.cpp
AndyZe: I'm with sea-bass here

sjahr: That's fine with me :+1:

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -367,32 +350,53 @@ Eigen::VectorXd Servo::jointDeltaFromCommand(const ServoInput& command, const mo
     }
     else if (expected_type == CommandType::TWIST)
     {
-      try
+      const auto planning_frame_maybe = getIKSolverBaseFrame(robot_state, active_subgroup_name);
```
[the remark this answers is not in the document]
sea-bass: Done!

moveit_ros/moveit_servo/src/servo.cpp
sea-bass: Done!

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -497,29 +501,42 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
...
 
-  if (command.frame_id != servo_params_.planning_frame)
+  if (command.frame_id != planning_frame)
   {
     // Look up the transform between the planning and command frames.
-    const auto planning_to_command_tf = getPlanningToCommandFrameTransform(command.frame_id);
+    const auto planning_to_command_tf_maybe = getPlanningToCommandFrameTransform(command.frame_id, planning_frame);
+    if (!planning_to_command_tf_maybe.has_value())
+    {
+      return std::nullopt;
```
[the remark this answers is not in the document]
sea-bass: This will be directly related to the error from the previous comment, so I don't want to be redundant here since I just added an error to the above

moveit_ros/moveit_servo/src/servo.cpp
sea-bass: Same as above, this is captured in the error from `getPlanningToCommandFrameTransform()`

sjahr (approved):

sea-bass:
@AndyZe we're cutting a release tomorrow so I'm going to merge this in now. If you have any follow-on comments related to this PR, let's do a new issue/PR and take it from there?

PR #2517, opened 2023-11-08
Fix threading issue for collision velocity scaling in MoveIt Servo

eholum:
### Description

Fixes a thread safety issues in the collision monitor in MoveIt Servo. Basically, the shared atomic double `collision_velocity_scale_` was being reset to `1.0` at the start of each collision check loop, which would sometimes be picked up in the `Servo::getNextJointState`, so the robot ends up moving even though the scene is in collision. E.g.,

```
[component_container_mt-7] ERH TEMP: Servo::getNextJointState, collision_velocity_scale_=1                                                                                                    
[component_container_mt-7] ERH TEMP: CollisionMonitor::checkCollisions, collision_velocity_scale_=1      
```

To fix this, we should use a local variable to compute the new scaling factor, and only set the shared variable at the end of the collision processing loop.

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/src/collision_monitor.cpp
+++ b/moveit_ros/moveit_servo/src/collision_monitor.cpp
@@ -95,9 +95,6 @@ void CollisionMonitor::checkCollisions()
     const double self_velocity_scale_coefficient{ log_val / servo_params_.self_collision_proximity_threshold };
     const double scene_velocity_scale_coefficient{ log_val / servo_params_.scene_collision_proximity_threshold };
 
-    // Use a local variable to prevent access issues between the collision monitor and the servo loop
-    auto new_collision_velocity_scale = 1.0;
-
     if (servo_params_.check_collisions)
     {
       // Fetch latest robot state.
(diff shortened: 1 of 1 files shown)


moveit_ros/moveit_servo/src/collision_monitor.cpp
```diff
@@ -95,8 +95,8 @@ void CollisionMonitor::checkCollisions()
     const double self_velocity_scale_coefficient{ log_val / servo_params_.self_collision_proximity_threshold };
     const double scene_velocity_scale_coefficient{ log_val / servo_params_.scene_collision_proximity_threshold };
 
-    // Reset the scale on every iteration.
-    collision_velocity_scale_ = 1.0;
+    // Use a local variable to prevent access issues between the collision monitor and the servo loop
+    auto new_collision_velocity_scale = 1.0;
 
     if (servo_params_.check_collisions)
```
sea-bass, line 101: Another alternative is to do away entirely with the local variable, and simply set the collision scale to 0.0 in the `else` of this statement. Either/or

sea-bass:
@ibrahiminfinite another one! I think I might have caused this one out of negligence in some of my earlier "fixes".

moveit_ros/moveit_servo/src/collision_monitor.cpp
eholum: Yeah I think it's 6 of one, half dozen of the other. 

In this instance I'll lean towards the extra variable to make in painfully obvious that it's a shared variable and it shouldn't have intermediate values set while it's being computed. It wasn't immediately obvious to me how `collision_velocity_scale_` works until you explained it to me, and this way I think it's more clear to new devs working on servo.

eholum: In that spirit I updated the comment to make it even more obvious...

eholum: Ignore my previous comments here it is without a local variable.

sea-bass (approved):
Thank you!

AndyZe (approved):

PR #2536, opened 2023-11-20
[PSM] Add proccess Collision Object to PSM and request planning scene to moveit py to allow syncing of mutliple PSM

JensVanhooydonck:
Add proccess collision object and proccess attached collision object to PlanningSceneMonitor and add request planning scene to moveit py to allow syncing of multiple planning scene monitors.

### Description

When altering a(n)  (attached-)collision object in the planning scene by using the applyCollisionObject or processAttachedCollisionObject function on the planning scene object, this does not trigger an updated in the PSM.

When using the same logic in PSM we can alter a collsiion object directly, and sync this to other planning scene monitors.

To make this complete I also added the request planning scene function to the MoveitPy interface

### Checklist


JensVanhooydonck:
Know issues when trying to sync multiple planning scene:
 - When using an acm, this does not allow for diff planning scene messages and triggers a full planning scene each time. Which results in an infinity loop when subscribing to a master PSM.

peterdavidfagan (approved):
These changes look good to me as they enable a more expressive Python API. 

@henningkayser are you ok with the changes made to the `planning_scene_monitor` cpp code? They should be mostly inconsequential from what I can see, as they simply split a callback into a method which is called by the callback and the original callback (changes to collision and attached collision object callbacks on the psm). The purpose of this is to create  a method for processing collision objects directly in the Python API essentially exposing the callback functionality as a Python method.

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
```diff
@@ -780,6 +780,42 @@ bool PlanningSceneMonitor::newPlanningSceneMessage(const moveit_msgs::msg::Plann
...
+      std::unique_lock<std::shared_mutex> ulock(scene_update_mutex_);
+      last_update_time_ = rclcpp::Clock().now();
+      if (!scene_->processAttachedCollisionObjectMsg(*object))
+        return false;
+    }
+    triggerSceneUpdateEvent(UPDATE_GEOMETRY);
+    RCLCPP_INFO(logger_, "Published update attached");
+    return true;
+  }
+  return false;
```
sjahr, line 816: Small nitpick. I like it a bit better if "correct" return is done at the end.
```suggestion
  if (!scene_)
  {
    return false;
  }

  updateFrameTransforms();
  {
    std::unique_lock<std::shared_mutex> ulock(scene_update_mutex_);
    last_update_time_ = rclcpp::Clock().now();
    if (!scene_->processAttachedCollisionObjectMsg(*object))
      return false;
  }
  triggerSceneUpdateEvent(UPDATE_GEOMETRY);
  RCLCPP_INFO(logger_, "Published update attached");
  return true;
```

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
```diff
@@ -807,31 +843,12 @@ void PlanningSceneMonitor::newPlanningSceneWorldCallback(
...
-    if (!scene_->processCollisionObjectMsg(*obj))
-      return;
-  }
-  triggerSceneUpdateEvent(UPDATE_GEOMETRY);
+  processCollisionObjectMsg(obj);
 }
 
 void PlanningSceneMonitor::attachObjectCallback(const moveit_msgs::msg::AttachedCollisionObject::ConstSharedPtr& obj)
 {
-  if (scene_)
```
sjahr, line 825: I think it would be even nicer to completely get rid of the callback member functions and just call the new function with a lambda when the services are initialized

sjahr (commented):
Thanks for your contribution @JensVanhooydonck! Just a small comment and then I think this is good to go

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
JensVanhooydonck: Shall I just remove the callback functions and change the call of these functions from 
```cpp
  collision_object_subscriber_ = pnode_->create_subscription<moveit_msgs::msg::CollisionObject>(
        collision_objects_topic, rclcpp::SystemDefaultsQoS(),
        [this](const moveit_msgs::msg::CollisionObject::ConstSharedPtr& obj) { return collisionObjectCallback(obj); });
    RCLCPP_INFO(logger_, "Listening to '%s'", collision_objects_topic.c_str());
```

to

```cpp
  collision_object_subscriber_ = pnode_->create_subscription<moveit_msgs::msg::CollisionObject>(
        collision_objects_topic, rclcpp::SystemDefaultsQoS(),
        [this](const moveit_msgs::msg::CollisionObject::ConstSharedPtr& obj) { return processCollisionObjectMsg(obj); });
    RCLCPP_INFO(logger_, "Listening to '%s'", collision_objects_topic.c_str());
```

And the same for the AttachedCollisionObjects?

These functions do not seem to be used somewhere else. I've added this to make sure there would not be any breaking changes.

sjahr: Yes :+1: They aren't used anywhere else and if they only call another function we don't need them. I think the lambda needs be
```diff
-   [this](const moveit_msgs::msg::CollisionObject::ConstSharedPtr& obj) { return processCollisionObjectMsg(obj); }
+   [this](const moveit_msgs::msg::CollisionObject::ConstSharedPtr& obj) { processCollisionObjectMsg(obj); }
```
because the `create_subscription` expects a function argument that returns void but just use whatever compiles

sjahr (approved):

PR #2537, opened 2023-11-20
Ensure to reset the smoothing plugin when resuming Servo

sea-bass:
This fixes another Servo issue we found when resuming Servo after moving through another means, e.g., by running the motion planner to move to a completely different waypoint.

With this fix, the robot shouldn't teleport in these situations.

@ibrahiminfinite

### DIFF
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
@@ -125,7 +125,19 @@ public:
    * @param The last commanded joint states.
    * @return The next state stepping towards the required halting state.
    */
-  std::pair<bool, KinematicState> smoothHalt(const KinematicState& halt_state) const;
+  std::pair<bool, KinematicState> smoothHalt(const KinematicState& halt_state);
+
+  /**
+   * \brief Updates the smoother, if one exists, by filtering with a new state.
+   * @param state The state used to update the smoothing plugin.
+   */
+  void updateSmoother(KinematicState& state);
+
+  /**
+   * \brief Resets the smoother, if one exists, to a specified state.
+   * @param state The desired state to reset the smoother to.
+   */
+  void resetSmoother(const KinematicState& state);
 
 private:
   /**
@@ -215,6 +227,7 @@ private:
   std::atomic<double> collision_velocity_scale_ = 1.0;
   std::unique_ptr<CollisionMonitor> collision_monitor_;
 
+  // Pointer to the (optional) smoothing plugin.
   pluginlib::UniquePtr<online_signal_smoothing::SmoothingBaseClass> smoother_ = nullptr;
 
   // Map between joint subgroup names and corresponding joint name - move group indices map
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
@@ -128,16 +128,16 @@ public:
   std::pair<bool, KinematicState> smoothHalt(const KinematicState& halt_state);
 
   /**
-   * \brief Updates the smoother, if one exists, by filtering with a new state.
-   * @param state The state used to update the smoothing plugin.
+   * \brief Applies smoothing to an input state, if a smoothing plugin is set.
+   * @param state The state to be updated by the smoothing plugin.
    */
-  void updateSmoother(KinematicState& state);
+  void doSmoothing(KinematicState& state);
 
   /**
-   * \brief Resets the smoother, if one exists, to a specified state.
-   * @param state The desired state to reset the smoother to.
+   * \brief Resets the smoothing plugin, if set, to a specified state.
+   * @param state The desired state to reset the smoothing plugin to.
    */
-  void resetSmoother(const KinematicState& state);
+  void resetSmoothing(const KinematicState& state);
 
 private:
   /**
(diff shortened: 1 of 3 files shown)


moveit_ros/moveit_servo/src/servo_node.cpp
```diff
@@ -162,6 +162,9 @@ void ServoNode::pauseServo(const std::shared_ptr<std_srvs::srv::SetBool::Request
   }
   else
   {
+    // Reset the smoother with the robot's current state in case the robot moved between pausing and unpausing.
+    servo_->resetSmoother(servo_->getCurrentRobotState());
```
sea-bass, line 166: this is the actual fix, the rest is reducing code duplication

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -630,16 +640,12 @@ std::pair<bool, KinematicState> Servo::smoothHalt(const KinematicState& halt_sta
...
 
+  updateSmoother(target_state);
+
   // If all velocities are near zero, robot has decelerated to a stop.
   stopped =
       (std::accumulate(target_state.velocities.begin(), target_state.velocities.end(), 0.0) <= STOPPED_VELOCITY_EPS);
 
-  if (smoother_)
-  {
-    smoother_->reset(current_state.positions, current_state.velocities, current_state.accelerations);
```
sea-bass, line 639: Not sure why `reset()` was being called here, but I think this was a bug?

moveit_ros/moveit_servo/src/servo_node.cpp
marioprats: Is `pauseServo` always called when switching to another mode? in other words, is it possible to move the arm with planners, while servo is active?

sea-bass: Yes, this is possible because both the planning pipelines and servo are separate nodes and can be blasting out data to the same ros2_controller.

However, in most applications (MoveIt Studio very much included), we never have unpaused Servo and motion planning + execution happening at the same time.

I'd call this case "user error".

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
```diff
@@ -125,7 +125,19 @@ class Servo
...
    * @return The next state stepping towards the required halting state.
    */
-  std::pair<bool, KinematicState> smoothHalt(const KinematicState& halt_state) const;
+  std::pair<bool, KinematicState> smoothHalt(const KinematicState& halt_state);
+
+  /**
+   * \brief Updates the smoother, if one exists, by filtering with a new state.
+   * @param state The state used to update the smoothing plugin.
+   */
+  void updateSmoother(KinematicState& state);
```
marioprats, line 134: what about `doSmoothing` as in the smoother function it calls? or `applySmoothing`?
`updateSmoother` makes me think it's changing the underlying smoothing algorithm.

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
```diff
@@ -125,7 +125,19 @@ class Servo
    * @param The last commanded joint states.
    * @return The next state stepping towards the required halting state.
    */
-  std::pair<bool, KinematicState> smoothHalt(const KinematicState& halt_state) const;
+  std::pair<bool, KinematicState> smoothHalt(const KinematicState& halt_state);
+
+  /**
+   * \brief Updates the smoother, if one exists, by filtering with a new state.
```
marioprats, line 131: Maybe reword this to emphasize that this is applying smoothing to `state` and updating it accordingly.

marioprats (approved):
LGTM with minor comments

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
```diff
@@ -215,6 +227,7 @@ class Servo
   std::atomic<double> collision_velocity_scale_ = 1.0;
   std::unique_ptr<CollisionMonitor> collision_monitor_;
 
+  // Pointer to the (optional) smoothing plugin.
```
AndyZe, line 230: You could probably use std::optional here. Different issue, I suppose

AndyZe (changes requested):
> this is the actual fix, the rest is reducing code duplication

I don't think this is true, LOC went up by 22 on a 2-line bug fix. Suggest reverting the new functions `doSmoothing()` and `resetSmoothing()` b/c they just add an extra layer. I think the smoother API is pretty clear already

sea-bass:
> > this is the actual fix, the rest is reducing code duplication
> 
> I don't think this is true, LOC went up by 22 on a 2-line bug fix. Suggest reverting the new functions `doSmoothing()` and `resetSmoothing()` b/c they just add an extra layer. I think the smoother API is pretty clear already

Going by raw LOC, sure, more lines were added. However, the 2 line bug fix would have been 5 lines without this and would have copy pasted a bunch of code that already existed elsewhere:

```cpp
const auto current_state = getCurrentRobotState();
if (smoother_)
{
  smoother_->reset(current_state.positions, current_state.velocities, current_state.accelerations);
}
```

Also, I needed a public function so that the servo node could reset the smoother... meaning getting rid of at least the `resetSmoothing` function would require me to change `smoother_` to be a public member. Which is not ideal.

AndyZe:
OK. I didn't realize this would involve making the member var public

AndyZe (approved):

PR #2547, opened 2023-11-22
Update ROS Snapshots GPG key

stephanie-eng:
As per https://github.com/osrf/docker_images/issues/697


Abishalini (approved):

mikaelarguedas:
FYI the ROS images on dockerhub have been rebuilt with the updated GPG key.
More details at https://github.com/osrf/docker_images/issues/697#issuecomment-1856463789

stephanie-eng:
Thanks!
In this case the script is using a 4-year-old image ([personalrobotics/ros-openrave](https://hub.docker.com/layers/personalrobotics/ros-openrave/latest/images/sha256-19a7abfb8b6c3a98cf70a38fed2481e76618c1f7fba552e309a02bd34798678a?context=explore)), so unless that image is updated as well, we'll still need to add the key in this way.

PR #2549, opened 2023-11-24
Remember original color of objects in planning scene

sea-bass:
This PR adds a mechanism for the planning scene to remember objects' original colors when first added, which helps restore colors to the original when attaching/detaching in RViz.

See the video below. Prior to this PR, the blue object would have turned green when being placed.

https://github.com/ros-planning/moveit2/assets/4603398/ba72dc59-5490-426f-a75e-b765a00880fa

### DIFF
--- a/moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
+++ b/moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
@@ -53,6 +53,7 @@
 #include <octomap_msgs/msg/octomap_with_pose.hpp>
 #include <memory>
 #include <functional>
+#include <optional>
 #include <thread>
 #include <variant>
 #include <rclcpp/rclcpp.hpp>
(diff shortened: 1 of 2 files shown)


moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
```diff
@@ -742,6 +743,7 @@ class MOVEIT_PLANNING_SCENE_EXPORT PlanningScene : public std::enable_shared_fro
   bool hasObjectColor(const std::string& id) const;
 
   const std_msgs::msg::ColorRGBA& getObjectColor(const std::string& id) const;
+  std::optional<std_msgs::msg::ColorRGBA> getOriginalObjectColor(const std::string& id) const;
```
sjahr, line 746: Can you add /** \brief ... */ documentation here?

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -1621,6 +1622,11 @@ bool PlanningScene::processAttachedCollisionObjectMsg(const moveit_msgs::msg::At
         const Eigen::Isometry3d& pose = attached_body->getGlobalPose();
         world_->addToObject(name, pose, attached_body->getShapes(), attached_body->getShapePoses());
         world_->setSubframesOfObject(name, attached_body->getSubframes());
+        const auto original_object_color = getOriginalObjectColor(name);
+        if (original_object_color.has_value())
```
sjahr, line 1626: Can you add a small explanation here. Which color has the object if it does not have an original color?

sea-bass: Done!

sjahr (approved):

PR #2555, opened 2023-11-27
Don't assume gripper controller for single joint control in MoveIt Setup Assistant

forrest-rm:
### Description

The MoveIt Setup Assistant assumed any controller with a single joint was a gripper controller. This created a bug in the generated ros2_controllers yaml with parameter `joint` instead of `joints`.


sjahr (approved):

moveit_setup_assistant/moveit_setup_controllers/src/ros2_controllers_config.cpp
```diff
@@ -192,7 +192,7 @@ bool ROS2ControllersConfig::GeneratedControllersConfig::writeYaml(YAML::Emitter&
       emitter << YAML::Value;
       emitter << YAML::BeginMap;
       {
-        if (ci.joints_.size() != 1)
+        if (ci.joints_.size() != 1 || ci.type_ != "position_controllers/GripperActionController")
```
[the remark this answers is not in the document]
forrest-rm: Only checking for "GripperActionController" would be simpler, and either option should probably work. I was going with the minimal change without making too many assumptions, but you know the controller parameters better.

moveit_setup_assistant/moveit_setup_controllers/src/ros2_controllers_config.cpp
```diff
@@ -192,7 +192,7 @@ bool ROS2ControllersConfig::GeneratedControllersConfig::writeYaml(YAML::Emitter&
       emitter << YAML::Value;
       emitter << YAML::BeginMap;
       {
-        if (ci.joints_.size() != 1)
+        if (ci.type_ != "position_controllers/GripperActionController")
```
mikeferguson, line 195: While the diff will be slightly larger - maybe swap these statements so that we aren't working in negatives?
```
if (ci.type_ == "position_controllers/GripperActionController")
{
  emitter << YAML::Key << "joint" << YAML::Value << ci.joints_[0];
}
else
{
  emitter << YAML::Key << "joints" << YAML::Value << ci.joints_;
}
```

forrest-rm: The default behavior is to use the "joints" parameter as only the gripper controller uses the singular "joint". Either putting the "default" or "positive" clause first would work, but I think I'll leave it as is for now.

PR #2406, opened 2023-10-04
(moveit_py) Add Trajectory Execution Manager

MatthijsBurgh:
### Description

Add Trajectory Execution Manager.

Note: I have not been able to test the correct behaviour of all added functions. Also I skipped some functions, which are shown as comments.

### Checklist
- [x] **Required by CI**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)
- [x] Extend the tutorials / documentation [reference](http://moveit.ros.org/documentation/contributing/)
- [x] Document API changes relevant to the user in the [MIGRATION.md](https://github.com/ros-planning/moveit2/blob/main/MIGRATION.md) notes
- [x] Create tests, which fail without this PR [reference](https://moveit.picknik.ai/humble/doc/examples/tests/tests_tutorial.html)
- [x] Include a screenshot if changing a GUI
- [x] While waiting for someone to review your request, please help review [another open pull request](https://github.com/ros-planning/moveit2/pulls) to support the maintainers

[//]: # "You can expect a response from a maintainer within 7 days. If you haven't heard anything by then, feel free to ping the thread. Thank you!"

PR #2563, opened 2023-12-01
Fix trajectory execution manager comments for docs builds

sea-bass:
Some of the new additions to the `moveit_py` package in https://github.com/ros-planning/moveit2/pull/2406 caused docs build failures because it's not happy with underscores.

Anyways, the original comments shouldn't point to a private member variable, but rather the ROS parameter that it is set from.

Confirmed this fixes docs builds on my end.


Abishalini (approved):
Thanks for the fix! LGTM

PR #2566, opened 2023-12-01
Fix warning and cleanup unneeded placeholders

sjahr:
### Description

Cleanups + remove wrong print statement

### Checklist


moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -135,8 +135,6 @@ class Ros2ControlManager : public moveit_controller_manager::MoveItControllerMan
     // Skip if controller stamp is too new for new discovery, enforce update if force==true
     if (!force && ((node_->now() - controllers_stamp_) < CONTROLLER_INFORMATION_VALIDITY_AGE))
     {
-      RCLCPP_WARN_STREAM(LOGGER, "Controller information from " << list_controllers_service_->get_service_name()
-                                                                << " is out of date, skipping discovery");
```
Abishalini, line 139: Why is this being deleted?

sjahr: That if statement throttles the discovery rate. The warning itself is wrong because the data is too new for new discovery and not out of date. Furthermore, I think that this is not a thing the user needs to be warned about, so I removed it completely since it was noise in my log

tylerjw (approved):

PR #2574, opened 2023-12-04
(moveit_py) Python Bindings for MoveGroupInterface

m-elwin:
### Description

Implements preliminary python bindings for the C++ [MoveGroupInterface](https://github.com/ros-planning/moveit2/blob/iron/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h)

Only the minimum needed to create a python version of the [Your First C++ MoveIt Project Tutorial](https://moveit.picknik.ai/main/doc/tutorials/your_first_project/your_first_project.html), is currently implemented.

A draft of a python version of the tutorial can be found [here](https://github.com/m-elwin/moveit2_tutorials/tree/py_movegroupinterface/doc/tutorials/your_first_project_py).

For example,  `ros2 launch moveit2_tutorials demo.launch.py`. Then, with this PR,
running the following python script will move end-effector to a target pose:

```python
import moveit
from moveit import MoveGroupInterface
from geometry_msgs.msg import Pose, Quaternion, Point

# Creates a node called "hello_moveit_interface" 
# (in C++ with a separate executor) and connects to the "manipulator" move_group
move_group_interface = MoveGroupInterface("hello_moveit_interface", "manipulator")

# set the desired end-effector target pose
target_pose = Pose(quaternion=Quaternion(w=1),
               position=Point(x=0.28, y=-0.2, z=0.5))
move_group_interface.setPoseTarget(target_pose)

# Plan the motion
(result, plan) = move_group_interface.plan()

# Execute the motion if planning was successful
if result.val == result.SUCCESS:
    move_group_interface.execute(plan)
else:
    print("Planning Failed")
```

### Checklist

### DIFF
--- a/moveit_py/CMakeLists.txt
+++ b/moveit_py/CMakeLists.txt
@@ -64,10 +64,12 @@ pybind11_add_module(planning
     src/moveit/planning.cpp
     src/moveit/moveit_ros/moveit_cpp/moveit_cpp.cpp
     src/moveit/moveit_ros/moveit_cpp/planning_component.cpp
+    src/moveit/moveit_ros/move_group_interface/move_group_interface.cpp
     src/moveit/moveit_ros/planning_scene_monitor/planning_scene_monitor.cpp
     src/moveit/moveit_ros/trajectory_execution_manager/trajectory_execution_manager.cpp
 )
 target_link_libraries(planning PRIVATE moveit_ros_planning::moveit_cpp
+    moveit_ros_planning_interface::moveit_move_group_interface
     moveit_ros_planning::moveit_planning_scene_monitor
     moveit_ros_planning::moveit_trajectory_execution_manager
     moveit_core::moveit_utils
--- a/moveit_py/src/moveit/moveit_ros/move_group_interface/move_group_interface.cpp
+++ b/moveit_py/src/moveit/moveit_ros/move_group_interface/move_group_interface.cpp
@@ -1,0 +1,107 @@
+/*********************************************************************
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2023, Matthew Elwin
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of PickNik Inc. nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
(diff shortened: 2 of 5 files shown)


m-elwin:
Re-based off the latest main branch.

sjahr (changes requested):
Thank you for this contribution!
I am not sure, if it makes sense to expose the move_group_interface with the python bindings. We'd have a python wrapper around the cpp wrapper of the ros interface of moveit which is probably too many layers. I think an approach such as the one taken [here](https://github.com/AndrejOrsula/pymoveit2) where the python code uses the move_group ros api directly might be more appropriate although I am not sure if something like this needs to live in the main repository. Please let me know if you have a good rationale for the MoveGroupInterface python bindings, I am not seeing at the moment!

m-elwin:
Thanks for the feedback @sjahr.

There are two primary motivations behind why I decided to work on this:
1. Keep the beginner python moveit2 experience as close as possible to that of the beginner C++ experience.
2. Enable every C++ moveit tutorial to have a direct python equivalent and help ensure that what can be done in C++ can also be done in python with similar code.

Here's the main motivating example: The [Your First C++ MoveIt Project](https://moveit.picknik.ai/main/doc/tutorials/your_first_project/your_first_project.html) 

I wrote an equivalent python tutorial [Your First Python MoveIt Project](https://github.com/m-elwin/moveit2_tutorials/blob/py_movegroupinterface/doc/tutorials/your_first_project_py/your_first_project_py.rst) (which relies on this PR).

The python tutorial could not be written with existing `moveit_py`, because it would require setting up moveit parameters in a launchfile. Nor could it be written with `pymoveit2` because the python user would need to deal with executors, callback groups, and threads (as far as I can tell). Either way, the alternative approaches would have resulted in a python tutorial that was arguably more complicated than the C++ equivalent (which seems wrong).

It would definitely be possible to do all this with a third-party python wrapper for the ROS API (such as `pymoveit2`) but there are a few disadvantages of this approach that I can see:
1. These API's being third-party makes them seem less official and less integrated and supported.
2. There is potentially a lot of duplicated code in maintaining a separate pure-python API wrapper, especially because `moveit_py` already provides the infrastructure for making python bindings.
3. It is easier to ensure feature-parity and similar usage patterns between C++ and python by using bindings rather than a separate ROS API wrapper.


I agree that there are advantages to the approach `pymoveit2` takes, including avoiding "wrapping the wrapper", but I think of both approaches as complimentary rather than mutually exclusive and targeted at slightly different users and use-cases.

MatthijsBurgh:
I think I agree with @sjahr.

The move group interface is some cpp code to talk to ROS API of moveit. I don't think we should wrap that with pybind. I do think it is a good idea to create a move group interface for python, but written in python.

If that library is close to the cpp interfaces, it could live in this repo, maybe even be part of moveit_py.

m-elwin:
Thanks for the feedback @MatthijsBurgh.

It seems like the main reluctance here is that MoveGroupInterface is already a wrapper around the ROS API so this is "wrapping a wrapper".

My concern is that MoveGroupInterface does not seem to be just a simple ROS API wrapper: it uses threads, futures,  and matrix math,all of which a pure python approach needs to re-implement and maintain. And the existing python API wrappers (such as pymoveit2) are non-trivial.

Anyway, thanks for taking the time to consider this. I'd like to help contribute to making the python and C++ moveit experiences similar regardless of the outcome of this PR, so I'm looking for feedback on how to proceed. 

Is the pybind11 approach in this PR ruled out completely? If so I'll close the PR (and can make a new one removing the currently unused python wrapping code in the C++ MoveGroupInterface).

If there is a possibility of acceptance down the road, I'm happy to attempt writing a full wrapper implementation with tests and tutorials to prove the concept further.

sjahr:
@m-elwin We really appreciate your effort here! 

> Is the pybind11 approach in this PR ruled out completely (seems like it might be)? If so I'll close the PR (and can make a new one removing the currently unused python wrapping code in the C++ MoveGroupInterface).

Maybe this would be a good step forward. Furthermore, you could try writing a minimal python move_group interface (e.g. as a tutorial for moveit and we'll see where to go from there). If you're interested you could join the public maintainer meeting later this month to discuss any further ideas

MatthijsBurgh:
We should close this PR

MatthijsBurgh:
We should close this PR

MatthijsBurgh:
@sjahr please close this one

rhaschke:
Closing. See https://github.com/moveit/moveit_task_constructor/issues/647#issuecomment-2619945410 for a proposed roadmap.

PR #2580, opened 2023-12-06
Refactor MoveIt Simple Controller Manager to use generate param library 

### LINKED ISSUE #2390: Q: How to configure the goal_time_tolerance?
Is it possible to configure the `goal_time_tolerance`?

As the only code I have found which refers to it is:
https://github.com/ros-planning/moveit2/blob/c8a0eb0db885c4ee0c8c4e04304ec39e8d760f59/moveit_plugins/moveit_simple_controller_manager/src/follow_joint_trajectory_controller_handle.cpp#L108C7-L108C21

Abishalini:
### Description

Also resolves #2390 

`generate_parameter_library` currently doesn't support nested maps in the YAML - https://github.com/PickNikRobotics/generate_parameter_library/issues/151
So I haven't implemented the `goal_tolerance` and `path_tolerance` parameters in this PR. But we can now set the `goal_time_tolerance` in the YAML file for `follow_joint_trajectory_controller`.

I will add a TODO/open another issue to implement the nested map parameters.


moveit_plugins/moveit_simple_controller_manager/src/moveit_simple_controller_manager_parameters.yaml
```diff
@@ -0,0 +1,103 @@
...
+    }
+    make_default: {
+      type: bool,
+      default_value: false,
+      description: "This param was originally named default. Can't use default with generate_param_lib as default is a C++ keyword. Marking a controller as
+       default makes MoveIt prefer this controller when multiple options are available.",
+      validation: {
+        not_empty<>: []
+      }
+    }
```
Abishalini, line 35: I also had to rename this parameter. Originally, it was called `default` but default is a C++ keyword and won't work with how generate param library initializes variables to store the parameter values 
```
/home/abishalini/ws_moveit2/build/moveit_simple_controller_manager/moveit_simple_controller_manager_parameters/include/moveit_simple_controller_manager_parameters.hpp:73:18: error: expected unqualified-id before ‘default’
   73 |             bool default = false;
      |                  ^~~~~~~
```

sjahr: Do we need to update any config files in the moveit_resources repos or studio?

moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/follow_joint_trajectory_controller_handle.h
```diff
@@ -63,6 +63,11 @@ class FollowJointTrajectoryControllerHandle
 
   // TODO(JafarAbdi): Revise parameter lookup
```
sjahr, line 64: I think this comment can be removed or updated

moveit_plugins/moveit_simple_controller_manager/src/follow_joint_trajectory_controller_handle.cpp
```diff
@@ -97,6 +98,19 @@ bool FollowJointTrajectoryControllerHandle::sendTrajectory(const moveit_msgs::ms
   return true;
 }
 
+void FollowJointTrajectoryControllerHandle::setPathTolerance(double /*path_tolerance*/)
+{
```
sjahr, line 102: Let's print a warning that they're not doing anything at the moment. Why do we need them in the first place?

sjahr (commented):
Nice thanks! Just a couple of comments and then it is ready to be merged

TomCC7:
Seems like [this PR](https://github.com/PickNikRobotics/generate_parameter_library/pull/166) resolves the issue of nested map parameters

henningkayser:
@Abishalini do you think you could complete this to get it merged?

PR #2587, opened 2023-12-07
Pass along move_group_capabilities parameters

forrest-rm:
The move_group_capabilities is currently unused in launches.py even though it is listed as a parameter of MoveItConfigs.

This fix is necessary to support MoveIt Task Constructor (MTC). In the move_group.lauch.py generated by the MoveIt Setup Assistant, the following line should be added when needed by MTC:
`moveit_config.move_group_capabilities = {"capabilities": "move_group/ExecuteTaskSolutionCapability"}`

I am open to suggestions if there's a better way of handling the move_group_capabilities parameter or to add the previous note as a comment in the auto generated launch scripts.

This was tested on the main and humble branches.


moveit_configs_utils/moveit_configs_utils/launches.py
```diff
@@ -228,6 +228,7 @@ def generate_move_group_launch(moveit_config):
     move_group_params = [
         moveit_config.to_dict(),
```
henningkayser, line 229: This line should already include the capabilities. Could you verify this?

forrest-rm: The capabilities parameter is not currently included in the `to_dict()` function:
https://github.com/ros-planning/moveit2/blob/8d2eb900d222f6609ee2588f6f5ebb97986635bd/moveit_configs_utils/moveit_configs_utils/moveit_configs_builder.py#L119C5-L119C23

I tried adding the capabilities parameter  to the `to_dict()` function but that didn't seem to work. There seems that's we need to added it to `move_group_params` after `move_group_configuration`, but I'm not sure why.

forrest-rm: @henningkayser, did you have additional thoughts on this PR or are you able to approve it?

bmagyar (approved):

moveit_configs_utils/moveit_configs_utils/launches.py
henningkayser: sorry for late response... I think params override each other, and move_group_configuration provides empty capabilities by default (lines 200 , 214). So I think the best solution would be to set the default to moveit_config.move_group_capabilities in line 200. That way, the launch file would still allow using the launch argument if necessary.

sjahr (approved):

moveit_configs_utils/moveit_configs_utils/launches.py
sjahr: @forrest-rm You're right. Do you mind opening a brief issue so we can track fixing this in a more appropriate way in the future?

forrest-rm:
@sjahr, I created Issue #2694 to track a more integrated solution.

Is it possible to also back-port this to the humble and iron branches? I have tested this on humble.

PR #2591, opened 2023-12-08
Adding planning_frame and ee_frame to panda_simulated_config.yaml

Nils-ChristianIseke:
When implementing moveit servo for my own robot. The servo node crashed because planning_frame and ee_frame were not defined. This was a bit of a pain to debug because the error did not show up in the terminal log when the servo node was started as a node component (this may also be fixed).

Disclaimer: I have not had time to test if this works with the example provided. So either wait until I do, or test it on your machine before approving.

### DIFF
--- a/moveit_ros/moveit_servo/config/panda_simulated_config.yaml
+++ b/moveit_ros/moveit_servo/config/panda_simulated_config.yaml
@@ -37,6 +37,8 @@ is_primary_planning_scene_monitor: true
 
 ## MoveIt properties
 move_group_name:  panda_arm  # Often 'manipulator' or 'arm'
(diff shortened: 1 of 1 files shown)


sea-bass (changes requested):
I think in the latest version of MoveIt Servo, we removed these parameters. See https://github.com/ros-planning/moveit2/pull/2515.

So, I think maybe you are using this configuration on an older commit of this repository?

Nils-ChristianIseke:
@sea-bass i am using the docker image moveit/moveit2:rolling-release but should have used moveit/moveit2:rolling-source instead, right?

Issue #2275, opened 2023-07-27
Break out parameters for `servo publish rate` and `trajectory point duration` for improved servoing via position control 


nbbrooks:
# Summary
Certain robot arms that only have a position control interface (e.g. no velocity control support) experience non-smooth and noisy (bad noises from gearbox) servo performance.

My current hypothesis is that this is related to using the same time parameter for both [publish period loop](https://github.com/ros-planning/moveit2/blob/main/moveit_ros/moveit_servo/src/servo_calcs.cpp#L314) and the [published JointTrajectoryPoint duration](https://github.com/ros-planning/moveit2/blob/main/moveit_ros/moveit_servo/src/servo_calcs.cpp#L195). Servoing via a position controller could cause the below behavior, where the arm's onboard control works aggressively works to de-accelerate and be at 0 velocity at the end of the JointTrajectoryPoint duration, then it immediately receives a new JointTrajectoryPoint telling it to accelerate again.

![timing_current](https://github.com/ros-planning/moveit2/assets/5314817/6a2200c5-2a9b-4372-a7dd-932b10293ba9)

# Proposed change
We should try making these parameters separate, so we do the following and test if it resolves the poor performance on certain hardware. We should also enforce that your trajectory point duration equal to or shorter than your publish duration.

![timing_proposed](https://github.com/ros-planning/moveit2/assets/5314817/bf5c4978-ace7-4f64-8d5c-d9c85ab3c73f)

# Describe alternatives you've considered
- Make no changes and document requiresments for arm manufacturers around either position control filtering techniques and/or requiring a velocity control interface.
- Make no changes and instead use onboard jogging services from these arm manufacturers (if they exist).
  - We lose our collision checking here.

pac48:
I believe ros2_cotrol  allows for setting endpoint velocities even if the robot is not velocity controlled. The default interpolation is cubic for a reference with position and velocity values. I ran a quick simulation to show how interpolation affects the spline's shape. 
 
![interp](https://github.com/ros-planning/moveit2/assets/28868750/e34f827b-d0b0-41ec-836d-08bec09df8f7)

The velocity profile @nbbrooks highlighted occurs when the endpoint velocities of the reference are 0. But I believe if they are set to non-zero you won't have the "noise" part.  This needs further validation, however.

nbbrooks:
Thanks @AndyZe - glad your intuition follows here.

Mostly curious here, but do you remember what ended up being the Gz issue you referenced in #1812 ?

[...]

PR #2594, opened 2023-12-11
Add command queue to servo to account for latency

pac48:
There is an issue where some robots experience non-smooth and noisy servo performance https://github.com/ros-planning/moveit2/issues/2275. Currently, servo sends trajectory messages to the ros2_control JTC one waypoint at a time. However, This causes issues because when the JTC receives a new message, it replaces the old one entirely. My proposal to solve this issue is to send more than one waypoint at a time and make sure everything is time-synchronized. Below is an example diagram.   

buffer

When messages are sent from servo to the JTC, there is an inherent noisy delay that should be accounted for. One strategy is to use a buffer that contains all generated servo commands from the current time to the current time plus the maximum expected latency. This range is shown in red. Any commands that are put in the buffer are considered "committed", meaning they will be executed by the JTC. For any time after the red region, there is still an opportunity to change commands. This is illustrated with the green trajectory.        
 
I recorded and plotted the commands received by the JTC with the current servo implementation (left) vs. this PR (right) using the 'demo_pose.cpp' example in the servo package.     

image

### DIFF
--- a/moveit_ros/moveit_servo/config/servo_parameters.yaml
+++ b/moveit_ros/moveit_servo/config/servo_parameters.yaml
@@ -23,6 +23,16 @@ servo:
     }
   }
 
+  latency: {
+    type: double,
+    read_only: true,
+    default_value: 0.1,
+    description: " max expected latency between generating a servo command and the controller receiving it [seconds]",
+    validation: {
+      gt<>: 0.0
+    }
+  }
+
   move_group_name: {
     type: string,
     read_only: true,
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
@@ -54,6 +54,7 @@
 #include <tf2_ros/transform_listener.h>
 #include <variant>
 #include <rclcpp/logger.hpp>
+#include <queue>
 
 namespace moveit_servo
 {
@@ -78,6 +79,13 @@ public:
    * @return The required joint state.
    */
   KinematicState getNextJointState(const ServoInput& command);
+
+  /**
+   * \brief Create a trajectory message.
+   * @return The trajectory message.
+   */
+  trajectory_msgs::msg::JointTrajectory createTrajectoryMessage();
+
 
   /**
    * \brief Set the type of incoming servo command.
@@ -180,11 +188,6 @@ private:
   Eigen::VectorXd jointDeltaFromCommand(const ServoInput& command, const moveit::core::RobotStatePtr& robot_state);
 
   /**
-   * \brief Updates data depending on joint model group
-   */
-  void updateJointModelGroup();
-
-  /**
    * \brief Validate the servo parameters
    * @param servo_params The servo parameters
    * @return True if parameters are valid, else False
@@ -232,6 +235,10 @@ private:
 
   // Map between joint subgroup names and corresponding joint name - move group indices map
   std::unordered_map<std::string, JointNameToMoveGroupIndexMap> joint_name_to_index_maps_;
+
+  // keep track of previously generated joint commands for constructing trajectory messages
+  std::deque<KinematicState> committed_commands_;
+
 };
 
 }  // namespace moveit_servo
--- a/moveit_ros/moveit_servo/include/moveit_servo/utils/datatypes.hpp
+++ b/moveit_ros/moveit_servo/include/moveit_servo/utils/datatypes.hpp
@@ -115,13 +115,14 @@ struct KinematicState
 {
   std::vector<std::string> joint_names;
   Eigen::VectorXd positions, velocities, accelerations;
+  rclcpp::Time time;
 
   KinematicState(const int num_joints)
   {
     joint_names.resize(num_joints);
-    positions.resize(num_joints);
-    velocities.resize(num_joints);
-    accelerations.resize(num_joints);
+    positions = Eigen::VectorXd::Zero(num_joints);
+    velocities = Eigen::VectorXd::Zero(num_joints);
+    accelerations = Eigen::VectorXd::Zero(num_joints);
   }
 
   KinematicState()
--- a/moveit_ros/moveit_servo/src/servo.cpp
+++ b/moveit_ros/moveit_servo/src/servo.cpp
@@ -100,6 +100,7 @@ Servo::Servo(const rclcpp::Node::SharedPtr& node, std::shared_ptr<const servo::P
   }
 
   // Create the collision checker and start collision checking.
+  // TODO if servo_params_.check_collisions == false, should CollisionMonitor even be created?
   collision_monitor_ =
       std::make_unique<CollisionMonitor>(planning_scene_monitor_, servo_params_, std::ref(collision_velocity_scale_));
   collision_monitor_->start();
@@ -244,6 +245,14 @@ bool Servo::validateParams(const servo::Params& servo_params) const
     RCLCPP_ERROR(logger_,
                  "The value '%s' Parameter 'active_subgroup' does not name a valid subgroup of joint group '%s'.",
                  servo_params.active_subgroup.c_str(), servo_params.move_group_name.c_str());
+    params_valid = false;
+  }
+
+  if (servo_params.latency < servo_params.publish_period)
+  {
+    RCLCPP_ERROR(logger_,
+                 "The publish period (%f sec) parameter must be less than latency parameter (%f sec).",
+                 servo_params.publish_period, servo_params.latency);
     params_valid = false;
   }
 
@@ -392,7 +401,7 @@ Eigen::VectorXd Servo::jointDeltaFromCommand(const ServoInput& command, const mo
     }
     else if (expected_type == CommandType::POSE)
     {
-      // Transform the twist command to the planning frame, which is the base frame of the active subgroup's IK solver,
+      // Transform the pose command to the planning frame, which is the base frame of the active subgroup's IK solver,
       // before applying it. The end effector frame is also extracted as the tip frame of the IK solver.
       // Additionally verify there is an IK solver, and that the transformation is successful.
       const auto planning_frame_maybe = getIKSolverBaseFrame(robot_state, active_subgroup_name);
@@ -456,9 +465,22 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
   KinematicState current_state(num_joints), target_state(num_joints);
   target_state.joint_names = joint_names;
 
-  // Copy current kinematic data from RobotState.
-  robot_state->copyJointGroupPositions(joint_model_group, current_state.positions);
-  robot_state->copyJointGroupVelocities(joint_model_group, current_state.velocities);
+  // remove commands not yet committed
+  auto cur_time = node_->now();
+  while (!committed_commands_.empty() && committed_commands_.back().time > (cur_time + rclcpp::Duration::from_seconds(servo_params_.latency))){
+    committed_commands_.pop_back();
+  }
+  if (!committed_commands_.empty()){
+    current_state = committed_commands_.back();
+  } else {
+    // Copy current kinematic data from RobotState.
+      robot_state->copyJointGroupPositions(joint_model_group, current_state.positions);
+      robot_state->copyJointGroupVelocities(joint_model_group, current_state.velocities);
+      current_state.joint_names = joint_names;
+      current_state.time = cur_time;
+      committed_commands_.push_back(current_state);
+  }
+
 
   // Compute the change in joint position due to the incoming command
   Eigen::VectorXd joint_position_delta = jointDeltaFromCommand(command, robot_state);
@@ -513,8 +535,64 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
     }
   }
 
+  // remove old commands
+  cur_time = node_->now();
+  while (!committed_commands_.empty() && committed_commands_.front().time < (cur_time - rclcpp::Duration::from_seconds(2*servo_params_.latency))){
+    committed_commands_.pop_front();
+  }
+
+  // add next committed command
+  target_state.time = node_->now() + rclcpp::Duration::from_seconds(servo_params_.latency);
+  committed_commands_.push_back(target_state);
+
+  // add end command stop point in case of large delay
+  auto dt = 2*(servo_params_.latency);
+  auto end_state = target_state;
+  for (auto i = 0ul; i < num_joints; i++) {
+    end_state.positions[i] = target_state.positions[i] + target_state.velocities[i]*dt;
+    end_state.velocities[i] = 0;
+    end_state.accelerations[i] = 0;
+    end_state.time = target_state.time + rclcpp::Duration::from_seconds(dt);
+    }
+  committed_commands_.push_back(end_state);
+
   return target_state;
 }
+
+trajectory_msgs::msg::JointTrajectory Servo::createTrajectoryMessage() {
+  // TODO should have realtime capabilities
+  trajectory_msgs::msg::JointTrajectory joint_trajectory;
+
+  // remove old commands
+  auto cur_time = node_->now();
+  while (!committed_commands_.empty() && committed_commands_.front().time < (cur_time-rclcpp::Duration::from_seconds(2*servo_params_.latency))){
+    committed_commands_.pop_front();
+  }
+  if (committed_commands_.empty()) {
+    RCLCPP_ERROR(node_->get_logger(), "Command queue is empty when creating trajectory, the newest command is already in the past!");
+    return joint_trajectory;
+  }
+  joint_trajectory.joint_names = committed_commands_.front().joint_names;
+  joint_trajectory.points.reserve(committed_commands_.size());
+  joint_trajectory.header.stamp = committed_commands_.front().time;
+  for (const auto& state : committed_commands_) {
+    trajectory_msgs::msg::JointTrajectoryPoint point;
+    size_t num_joints = state.positions.size();
+    point.positions.reserve(num_joints);
+    point.velocities.reserve(num_joints);
+    for (const auto& pos : state.positions) {
+      point.positions.emplace_back(pos);
+    }
+    for (const auto& vel : state.velocities) {
+      point.velocities.emplace_back(vel);
+    }
+    point.time_from_start = state.time - joint_trajectory.header.stamp;
+    joint_trajectory.points.emplace_back(point);
+  }
+
+  return joint_trajectory;
+}
+
 
 std::optional<Eigen::Isometry3d> Servo::getPlanningToCommandFrameTransform(const std::string& command_frame,
                                                                            const std::string& planning_frame) const
--- a/moveit_ros/moveit_servo/src/servo_node.cpp
+++ b/moveit_ros/moveit_servo/src/servo_node.cpp
@@ -332,7 +332,7 @@ void ServoNode::servoLoop()
     {
       if (servo_params_.command_out_type == "trajectory_msgs/JointTrajectory")
       {
-        trajectory_publisher_->publish(composeTrajectoryMessage(servo_->getParams(), next_joint_state.value()));
+        trajectory_publisher_->publish(servo_->createTrajectoryMessage());
       }
       else if (servo_params_.command_out_type == "std_msgs/Float64MultiArray")
       {
--- a/moveit_ros/moveit_servo/src/utils/common.cpp
+++ b/moveit_ros/moveit_servo/src/utils/common.cpp
@@ -43,6 +43,7 @@ namespace
 {
 // The threshold above which `override_velocity_scaling_factor` will be used instead of computing the scaling from joint bounds.
 const double SCALING_OVERRIDE_THRESHOLD = 0.01;
+const double MIN_ANGLE_THRESHOLD = 1E-16;
 }  // namespace
 
 namespace moveit_servo
@@ -115,10 +116,12 @@ geometry_msgs::msg::Pose poseFromCartesianDelta(const Eigen::VectorXd& delta_x,
 
   // Get a transformation matrix with desired orientation change
   Eigen::Isometry3d tf_rot_delta(Eigen::Isometry3d::Identity());
-  const Eigen::Quaterniond q = Eigen::AngleAxisd(delta_x[3], Eigen::Vector3d::UnitX()) *
-                               Eigen::AngleAxisd(delta_x[4], Eigen::Vector3d::UnitY()) *
-                               Eigen::AngleAxisd(delta_x[5], Eigen::Vector3d::UnitZ());
-  tf_rot_delta.rotate(q);
+  Eigen::Vector3d rot_vec = delta_x.block<3, 1>(3, 0, 3, 1);
+  double angle = rot_vec.norm();
+  if (angle > MIN_ANGLE_THRESHOLD) {
+    const Eigen::Quaterniond q(Eigen::AngleAxisd(angle, rot_vec / angle).toRotationMatrix());
+    tf_rot_delta.rotate(q);
+  }
 
   // Find the new tip link position without newly applied rotation
   const Eigen::Isometry3d tf_no_new_rot = tf_pos_delta * base_to_tip_frame_transform;


moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -513,9 +538,75 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
...
+    committed_commands_.pop_front();
+  }
+
+  // add next committed command
+  target_state.time = node_->now() + rclcpp::Duration::from_seconds(servo_params_.latency);
+  committed_commands_.push_back(target_state);
+
+  // add end command stop point in case of large delay
+  auto dt = 2 * (servo_params_.latency);
+  auto end_state = target_state;
```
AndyZe, line 555: So this is like an added halt state, to help with JTC interpolation? That makes sense. I bet that's the main reason for the smoother performance.

I would tweak the wording a little bit here. Can discuss later, if my understanding is correct.

moveit_ros/moveit_servo/config/servo_parameters.yaml
```diff
@@ -23,6 +23,16 @@ servo:
     }
   }
 
+  latency: {
```
sea-bass, line 26: I'd give this a more specific name than just "latency".

moveit_ros/moveit_servo/config/servo_parameters.yaml
```diff
@@ -23,6 +23,16 @@ servo:
...
+    type: double,
+    read_only: true,
+    default_value: 0.1,
+    description: " max expected latency between generating a servo command and the controller receiving it [seconds]",
```
sea-bass, line 30:
```suggestion
    description: "Maximum expected latency between generating a servo command and the controller receiving it [seconds].",

File: moveit_ros/moveit_servo/src/servo.cpp [hunk 4]
```diff
@@ -100,6 +100,7 @@ Servo::Servo(const rclcpp::Node::SharedPtr& node, std::shared_ptr<const servo::P
   }
 
   // Create the collision checker and start collision checking.
+  // TODO if servo_params_.check_collisions == false, should CollisionMonitor even be created?
```
sea-bass, line 103: Yes, because it's runtime tunable so it may be enabled later

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -513,9 +538,75 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
...
+  }
+
+  // add next committed command
+  target_state.time = node_->now() + rclcpp::Duration::from_seconds(servo_params_.latency);
+  committed_commands_.push_back(target_state);
+
+  // add end command stop point in case of large delay
+  auto dt = 2 * (servo_params_.latency);
+  auto end_state = target_state;
+  for (int i = 0; i < num_joints; i++)
```
sea-bass, line 556:
```suggestion
  for (int i = 0; i < num_joints; ++i)

**Review by AndyZe — COMMENTED** [review]
I think you could get rid of the complexity of the rolling window and simply always keep 2 waypoints in `committed_commands_`. I bet the majority of the performance gain comes from the `latency` parameter and the stop command point.

The latency parameter is identical to #886 but the stop command point is a good idea.

File: moveit_ros/moveit_servo/src/servo.cpp [hunk 6]
```diff
@@ -513,9 +538,75 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
     }
   }
 
+  // remove old commands
+  cur_time = node_->now();
+  while (!committed_commands_.empty() &&
+         committed_commands_.front().time < (cur_time - rclcpp::Duration::from_seconds(2 * servo_params_.latency)))
```
sea-bass, line 544: You're doing the `2 * servo_params_.latency` calculation in here, though `dt` is defined below. Maybe just define that earlier on and keep reusing the variable?

AndyZe:
Adding that `halt position` waypoint will likely have strange effects if the position/vel/accel JTC interfaces are active, since that increases the spline interpolation order --> potential for large swings during interpolation. It should work well if just the position interface is active, though.

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
```diff
@@ -79,6 +80,12 @@ class Servo
    */
   KinematicState getNextJointState(const ServoInput& command);
 
+  /**
+   * \brief Create a trajectory message.
+   * @return The trajectory message.
+   */
+  trajectory_msgs::msg::JointTrajectory createTrajectoryMessage();
```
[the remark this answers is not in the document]
sea-bass: Also, there already is a `composeTrajectoryMessage` utility function in `common.cpp`. Consider modifying that one rather than making a new one?

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -513,9 +538,75 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
...
+    for (const auto& pos : state.positions)
+    {
+      point.positions.emplace_back(pos);
+    }
+    for (const auto& vel : state.velocities)
+    {
+      point.velocities.emplace_back(vel);
+    }
+    point.time_from_start = state.time - joint_trajectory.header.stamp;
+    joint_trajectory.points.emplace_back(point);
```
sea-bass, line 604: Refer to the `composeTrajectoryMessage()` function in `common.cpp`, but I believe there are parameters such as `servo_params.publish_joint_positions/velocities/accelerations` that control what data to actually pack into the `JointTrajectory` message. This should be consistent with that.

moveit_ros/moveit_servo/src/utils/common.cpp
```diff
@@ -43,6 +43,7 @@ namespace
 {
 // The threshold above which `override_velocity_scaling_factor` will be used instead of computing the scaling from joint bounds.
 const double SCALING_OVERRIDE_THRESHOLD = 0.01;
+const double MIN_ANGLE_THRESHOLD = 1E-16;
```
sea-bass, line 46: Add a comment to this parameter.

Also, can these two things be converted to `constexpr` instead?

moveit_ros/moveit_servo/src/servo.cpp
AndyZe: We might need to think more about this. There is a filteredHalt() function for coming to a stop, gently. This would be an abrupt halt.

I'm also not sure it's necessary. If the JTC position interface is active, it will automatically stop.

Further realizing that this code path should never be executed anyway, unless the latency is extreme.

AndyZe (changes requested):
Need to rethink the `end_state`

moveit_ros/moveit_servo/config/servo_parameters.yaml
pac48: Okay, I changed it to `max_expected_latency`

pac48:
> Need to rethink the `end_state`

The one problem I see with using the `haltJoints()` function is that it seems like it needs to be called in the servo loop. The only scenario when `end_state` is even a factor is when the servo loop dies or is severely delayed. So it won't be easy to use. 

One option is to actually call `haltJoints()` multiple times to generate a sequence of gradually slowing points, instead of a single one with zero velocity. The drawback is that it will increase the size of the trajectory message and slow down the servo loop.

>  Further realizing that this code path should never be executed anyway, unless the latency is extreme.       
Another option is to not explicitly account for this case because the low-level controller will likely have a stopping mechanism. But I think it's safer to have it just in case. @AndyZe Do you have a suggestion to handle the edge case of extreme latency?

AndyZe:
I guess the halt point is fine.

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -456,9 +464,27 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
...
+  }
+  else
+  {
+    // Copy current kinematic data from RobotState.
+    robot_state->copyJointGroupPositions(joint_model_group, current_state.positions);
+    robot_state->copyJointGroupVelocities(joint_model_group, current_state.velocities);
+    current_state.joint_names = joint_names;
+    current_state.time = cur_time;
+    committed_commands_.push_back(current_state);
+  }
```
AndyZe, line 487: Can you please recreate the graphs using a `committed_commands_` of size 3 always? I think the rolling window is not necessary, as mentioned before.

In a nutshell, `committed_commands_` will be:

```
committed_commands_.clear();
...
committed_commands_.push_back(current_state);
...
committed_commands_.push_back(target_state);
...
committed_commands_.push_back(end_state);
```

Curious to see how these graphs look after that change.

graphs

AndyZe: Also, this won't be resizing all the time --> better for realtime performance.

pac48: >  Can you please recreate the graphs using a committed_commands_ of size 3 always? 

It makes the motion a little smoother than before, but there is still substantial noise. 

Screenshot from 2023-12-11 17-06-15
Screenshot from 2023-12-11 17-05-31

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
```diff
@@ -232,6 +234,9 @@ class Servo
 
   // Map between joint subgroup names and corresponding joint name - move group indices map
   std::unordered_map<std::string, JointNameToMoveGroupIndexMap> joint_name_to_index_maps_;
+
+  // keep track of previously generated joint commands for constructing trajectory messages
+  std::deque<KinematicState> committed_commands_;
```
AndyZe, line 239:
```suggestion
  // rolling window of joint commands
  std::deque<KinematicState> joint_cmd_rolling_window_;
```

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -513,9 +539,87 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
...
 
+  // remove old commands
+  cur_time = node_->now();
+  const auto active_time_window = rclcpp::Duration::from_seconds(2 * servo_params_.max_expected_latency);
+  while (!committed_commands_.empty() && committed_commands_.front().time < (cur_time - active_time_window))
+  {
+    committed_commands_.pop_front();
+  }
+
+  // add next committed command
```
AndyZe, line 550:
```suggestion
  // add next command
```

moveit_ros/moveit_servo/src/utils/common.cpp
```diff
@@ -115,10 +117,13 @@ geometry_msgs::msg::Pose poseFromCartesianDelta(const Eigen::VectorXd& delta_x,
 
   // Get a transformation matrix with desired orientation change
   Eigen::Isometry3d tf_rot_delta(Eigen::Isometry3d::Identity());
-  const Eigen::Quaterniond q = Eigen::AngleAxisd(delta_x[3], Eigen::Vector3d::UnitX()) *
-                               Eigen::AngleAxisd(delta_x[4], Eigen::Vector3d::UnitY()) *
-                               Eigen::AngleAxisd(delta_x[5], Eigen::Vector3d::UnitZ());
-  tf_rot_delta.rotate(q);
+  Eigen::Vector3d rot_vec = delta_x.block<3, 1>(3, 0, 3, 1);
```
AndyZe, line 120: `delta_x` is a vector so I don't think the element you want starts at (3,0).

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -456,9 +464,27 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
...
-  robot_state->copyJointGroupPositions(joint_model_group, current_state.positions);
-  robot_state->copyJointGroupVelocities(joint_model_group, current_state.velocities);
+  // remove commands not yet committed
+  auto cur_time = node_->now();
+  while (!committed_commands_.empty() &&
+         committed_commands_.back().time >
+             (cur_time + rclcpp::Duration::from_seconds(servo_params_.max_expected_latency)))
+  {
+    committed_commands_.pop_back();
+  }
```
AndyZe, line 474: Basically this is pushing out the halt waypoint, correct? Seems brittle to do it based on time. I'd suggest to always pop one command each iteration.

```suggestion
  // remove the previous halt command
  auto cur_time = node_->now();
  if (!committed_commands_.empty())
  {
    committed_commands_.pop_back();
  }
```

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
AndyZe: Not a fan of the "committed" terminology. I think people normally think of this as a rolling window.

pac48:
Between @sea-bass comments about making the change more functional, e.g. passing the rolling window in, and @AndyZe concerns about performance, I think it is best to factor out the command queueing logic and have the calling code maintain it. I'll move the logic into the `servo_node`. It's better that way.  Also, I'll make the name changes @AndyZe suggested,

AndyZe:
Well, props for figuring out this long-standing issue!

pac48:
> Well, props for figuring out this long-standing issue!

Here is a result of running the `realtime_servo` demo in the tutorials. The new implementation is shown in the top plots and the old in the bottom plots. The left and right columns are position and position deltas as a function of time.
   
image

moveit_core/online_signal_smoothing/include/moveit/online_signal_smoothing/acceleration_filter.h
```diff
@@ -0,0 +1,96 @@
...
+   * @param accelerations (unused)
+   * @return True if reset was successful
+   */
+  bool reset(const Eigen::VectorXd& positions, const Eigen::VectorXd& velocities,
+             const Eigen::VectorXd& accelerations) override;
+
+private:
+  rclcpp::Node::SharedPtr node_;
+  long num_joints_;
+  std::array<std::pair<Eigen::VectorXd, std::optional<rclcpp::Time>>, 2> last_positions_;
```
sea-bass, line 89: I think this array requires some more explanation, as I got a little confused about whether its size had to do with the joints vs. something else.

As I now understand, we need a buffer of 2 previous positions to actually limit the acceleration, right? If so, explain that and also what the 0th and 1st indices correspond to.

Also, please include in the comments that the pair is the actual position and its corresponding time, if one is available. I know this part is clear(ish) in the actual type, but just to be thorough.

moveit_core/online_signal_smoothing/src/acceleration_filter.cpp
```diff
@@ -0,0 +1,145 @@
...
+  else if (!last_positions_[1].second.has_value())
+  {
+    positions = last_positions_[1].first;
+    last_positions_[1].second = node_->now();
+  }
+  else
+  {
+    rclcpp::Time cur_time = node_->now();
+    double dt_1 = (last_positions_[1].second.value() - last_positions_[0].second.value()).seconds();
+    double dt_2 = (cur_time - last_positions_[1].second.value()).seconds();
```
sea-bass, line 100: I think these 3 can all be `const`

moveit_core/online_signal_smoothing/src/acceleration_filter.cpp
```diff
@@ -0,0 +1,145 @@
...
+    {
+      RCLCPP_ERROR_THROTTLE(node_->get_logger(), *node_->get_clock(), 1000,
+                            "The maximum allowed time between filter updates exceeded! The `doSmoothing` method must "
+                            "be called at a higher rate.");
+      return false;
+    }
+
+    prev_velocity_ = (last_positions_[1].first - last_positions_[0].first) / dt_1;
+    cur_velocity_ = (positions - last_positions_[1].first) / dt_2;
+    cur_acceleration = (cur_velocity_ - prev_velocity_) / dt_2;
```
sea-bass, line 111: We're not checking that `dt_1` and `dt_2` are greater than zero, so we run a risk of division by zero here

moveit_core/online_signal_smoothing/src/acceleration_filter.cpp
```diff
@@ -0,0 +1,145 @@
...
+    double scale = 1.0;
+    for (long i = 0; i < num_joints_; i++)
```
sea-bass, line 113:
```suggestion
    for (size_t i = 0; i < num_joints_; ++i)

File: moveit_core/online_signal_smoothing/src/acceleration_filter.cpp [hunk 19]
```diff
@@ -0,0 +1,145 @@
...
+                                  cur_acceleration[i]);
+    }
+    cur_acceleration *= scale;
+    cur_velocity_ = prev_velocity_ + cur_acceleration * dt_2;
+    positions = last_positions_[1].first + cur_velocity_ * dt_2;
+
+    last_positions_[0] = last_positions_[1];
+    last_positions_[1].first = positions;
+    last_positions_[1].second = cur_time;
+  }
```
sea-bass, line 125: This whole block of code, which has the actual math, could use some more comments to describe what's going on.

moveit_core/online_signal_smoothing/src/acceleration_parameters.yaml
```diff
@@ -0,0 +1,17 @@
+online_signal_smoothing:
+  joint_acceleration_limit: {
+    type: double,
+    description: "The unsigned joint acceleration limit used for smoothing",
```
sea-bass, line 4: Add units (m/s^2) to the description

moveit_core/online_signal_smoothing/src/acceleration_parameters.yaml
```diff
@@ -0,0 +1,17 @@
...
+    default_value: 5.0,
+    validation: {
+      gt<>: 0.0
+    }
+  }
+  update_timeout: {
+    type: double,
+    description: "The allowable time between updates such that the filter state is considered valid",
```
sea-bass, line 12: Add units (seconds) to the description here.

sea-bass (commented):
Took another pass!

I think this all looks good, but have 2 high level comments:

1. CI is failing, so that needs to be resolved as well
2. I am a bit concerned that, while this fixes a lot of issues, it makes the "hello world" demo examples look much more complicated. I would hope that we could try absorb more of the sliding window updating in the user-facing API so it doesn't make the usability more complicated?

AndyZe:
I would accept the PR if you revert to 9e8c1a2e0 then address my last few comments.

I think it's OK to keep this new rolling window in the Servo instance. std::deque allocates memory in chunks so the dynamic memory allocation issue really wouldn't be significant, almost certainly.

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
pac48: I changed all references of "committed" to "joint_cmd_rolling_window"

moveit_ros/moveit_servo/demos/cpp_interface/demo_joint_jog.cpp
```diff
@@ -87,10 +94,19 @@ int main(int argc, char* argv[])
...
+  joint_cmd_rolling_window.push_back(current_state);
+
   RCLCPP_INFO_STREAM(demo_node->get_logger(), servo.getStatusMessage());
   while (rclcpp::ok())
   {
-    const KinematicState joint_state = servo.getNextJointState(joint_jog);
+    auto last_commanded_state = joint_cmd_rolling_window.back();
+    robot_state->setJointGroupPositions(joint_model_group, last_commanded_state.positions);
+    KinematicState joint_state = servo.getNextJointState(robot_state, joint_jog);
+    joint_state.time = demo_node->now() + rclcpp::Duration::from_seconds(servo_params.max_expected_latency);
```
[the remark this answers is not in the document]
pac48: I simplified the demos. You don't need to add joint states manually. One reason it might seem more complex now is that you have to set the robot state yourself. This makes the interface more functional of course at the cost of more effect on the caller's part.

pac48:
> I would accept the PR if you revert to [9e8c1a2](https://github.com/ros-planning/moveit2/commit/9e8c1a2e04561afda6fd98074ef368960fb6f0b1) then address my last few comments.
> 
> I think it's OK to keep this new rolling window in the Servo instance. std::deque allocates memory in chunks so the dynamic memory allocation issue really wouldn't be significant, almost certainly.
> 
> Not sure the weekend changes are moving in the right direction.

I reverted `set last_commanded_state_ in ...` commit. 

> Basically this is pushing out the halt waypoint, correct? Seems brittle to do it based on time. I'd suggest to always pop one command each iteration.

I removed the "halt point" from the sliding window completely. It is now appended to the trajectory message when it is created in `composeTrajectoryMessage`.

>  I think you could get rid of the complexity of the rolling window and simply always keep 2 waypoints in committed_commands_

I think having more than 2 waypoints is beneficial when the lag parameter `max_expected_latency` is larger than the update period. For example, if I wanted to servo the robot in a rectangular pattern, the robot needs to change direction rapidly at the corners. If the latency was .1 seconds, then the robot would have trouble taking the sharp turn. Adding more points to the window, say 10 points over point .1 seconds, allows for faster changes. There is an upper bound on the queue size though, so we could use a vector with preallocated memory to make things more efficient if needed.

moveit_ros/moveit_servo/include/moveit_servo/utils/datatypes.hpp
```diff
@@ -115,13 +115,14 @@ struct KinematicState
 {
   std::vector<std::string> joint_names;
   Eigen::VectorXd positions, velocities, accelerations;
+  rclcpp::Time time;
```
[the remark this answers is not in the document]
pac48: renamed to time_stamp

moveit_ros/moveit_servo/config/panda_simulated_config.yaml
```diff
@@ -26,8 +27,9 @@ publish_joint_velocities: true
 publish_joint_accelerations: false
 
 ## Plugins for smoothing outgoing commands
-use_smoothing: true
-smoothing_filter_plugin_name: "online_signal_smoothing::ButterworthFilterPlugin"
+use_smoothing: false
+smoothing_filter_plugin_name: "online_signal_smoothing::ButterworthFilter"
+joint_acceleration_limit: 15.0
```
sea-bass, line 32: Why this `joint_acceleration_limit` vs. using the actual moveit params for max joint accelerations?

And I think it's also not reasonable for the "platform agnostic" case to have a single acceleration limit for all possible joints. Especially in the prismatic + revolute combination case.

I think the "correct" thing to do is launching servo with a different `joint_limits.yaml` parameter file than the move group, if you want to use different values.

moveit_ros/moveit_servo/demos/cpp_interface/demo_joint_jog.cpp
```diff
@@ -75,10 +75,17 @@ int main(int argc, char* argv[])
...
+
+  // Get the robot state and joint model group info.
+  const moveit::core::JointModelGroup* joint_model_group =
+      robot_state->getJointModelGroup(servo_params.move_group_name);
+
   // Set the command type for servo.
   servo.setCommandType(CommandType::JOINT_JOG);
   // JointJog command that moves only the 7th joint at +1.0 rad/s
-  JointJogCommand joint_jog{ { "panda_joint7" }, { 1.0 } };
+  JointJogCommand joint_jog{ { "panda_joint7" }, { 0.4 } };
```
sea-bass, line 88: Update the comment above this line since that still reads +1.0 rad/s

moveit_ros/moveit_servo/demos/cpp_interface/demo_joint_jog.cpp
```diff
@@ -102,7 +113,13 @@ int main(int argc, char* argv[])
     }
     else if (status != StatusCode::INVALID)
     {
-      trajectory_outgoing_cmd_pub->publish(composeTrajectoryMessage(servo_params, joint_state));
+      updateSlidingWindow(joint_state, joint_cmd_rolling_window, servo_params.max_expected_latency);
+      if (auto msg = composeTrajectoryMessage(servo_params, joint_cmd_rolling_window))
```
sea-bass, line 117: Can this be `const`?

moveit_ros/moveit_servo/demos/cpp_interface/demo_pose.cpp
```diff
@@ -81,87 +81,93 @@ int main(int argc, char* argv[])
...
 
-    command_rate.sleep();
+    // get next servo command
+    joint_state = servo.getNextJointState(robot_state, target_pose);
+    joint_state.time_stamp = demo_node->now() + rclcpp::Duration::from_seconds(servo_params.max_expected_latency);
+    StatusCode status = servo.getStatus();
+    if (status != StatusCode::INVALID)
+    {
+      updateSlidingWindow(joint_state, joint_cmd_rolling_window, servo_params.max_expected_latency);
+      if (auto msg = composeTrajectoryMessage(servo_params, joint_cmd_rolling_window))
```
sea-bass, line 159: same here, can this be `const`?

moveit_ros/moveit_servo/demos/cpp_interface/demo_twist.cpp
```diff
@@ -77,24 +77,35 @@ int main(int argc, char* argv[])
...
+  const moveit::core::JointModelGroup* joint_model_group =
+      robot_state->getJointModelGroup(servo_params.move_group_name);
+
   // Set the command type for servo.
   servo.setCommandType(CommandType::TWIST);
 
   // Move end effector in the +z direction at 10 cm/s
   // while turning around z axis in the +ve direction at 0.5 rad/s
-  TwistCommand target_twist{ "panda_link0", { 0.0, 0.0, 0.1, 0.0, 0.0, 0.5 } };
+  TwistCommand target_twist{ "panda_link0", { 0.0, 0.0, 0.05, 0.0, 0.0, 0.4 } };
```
sea-bass, line 92: again, update the numbers in the comment above this line... or remove them.

I generally think it's a bad idea to copy-paste literal numbers in comments for this reason, so I'm pro removing.

moveit_ros/moveit_servo/demos/cpp_interface/demo_twist.cpp
```diff
@@ -106,7 +117,13 @@ int main(int argc, char* argv[])
     }
     else if (status != StatusCode::INVALID)
     {
-      trajectory_outgoing_cmd_pub->publish(composeTrajectoryMessage(servo_params, joint_state));
+      updateSlidingWindow(joint_state, joint_cmd_rolling_window, servo_params.max_expected_latency);
+      if (auto msg = composeTrajectoryMessage(servo_params, joint_cmd_rolling_window))
```
sea-bass, line 121: const?

moveit_ros/moveit_servo/demos/cpp_interface/demo_twist.cpp
```diff
@@ -106,7 +117,13 @@ int main(int argc, char* argv[])
...
+      {
+        trajectory_outgoing_cmd_pub->publish(msg.value());
+      }
+      auto last_commanded_state = joint_cmd_rolling_window.back();
```
sea-bass, line 125: const?

moveit_ros/moveit_servo/demos/cpp_interface/demo_pose.cpp
```diff
@@ -81,87 +81,93 @@ int main(int argc, char* argv[])
...
+      {
+        trajectory_outgoing_cmd_pub->publish(msg.value());
+      }
+      auto last_commanded_state = joint_cmd_rolling_window.back();
```
sea-bass, line 163: const?

moveit_ros/moveit_servo/demos/cpp_interface/demo_joint_jog.cpp
```diff
@@ -102,7 +113,13 @@ int main(int argc, char* argv[])
...
+      {
+        trajectory_outgoing_cmd_pub->publish(msg.value());
+      }
+      auto last_commanded_state = joint_cmd_rolling_window.back();
```
sea-bass, line 121: const?

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
```diff
@@ -74,10 +75,11 @@ class Servo
 
   /**
    * \brief Computes the joint state required to follow the given command.
+   * @param a RobotStatePtr instance used for calculating the joint state.
```
sea-bass, line 78: The second thing in these has to be the actual variable name

```suggestion
   * @param robot_state a RobotStatePtr instance used for calculating the joint state.

File: moveit_ros/moveit_servo/include/moveit_servo/servo.hpp [hunk 34]
```diff
@@ -175,15 +177,11 @@ class Servo
   /**
    * \brief Compute the change in joint position required to follow the received command.
    * @param command The incoming servo command.
+   * @param a RobotStatePtr instance used for calculating the command.
```
sea-bass, line 180: same here

```suggestion
   * @param robot_state a RobotStatePtr instance used for calculating the command.

File: moveit_ros/moveit_servo/include/moveit_servo/utils/common.hpp [hunk 35]
```diff
@@ -182,4 +194,12 @@ PoseCommand poseFromPoseStamped(const geometry_msgs::msg::PoseStamped& msg);
  */
 planning_scene_monitor::PlanningSceneMonitorPtr createPlanningSceneMonitor(const rclcpp::Node::SharedPtr& node,
                                                                            const servo::Params& servo_params);
+
+/**
+ * \brief Extract the state from a RobotStatePtr instance.
+ * @param The RobotStatePtr instance.
```
sea-bass, line 200:
```suggestion
 * @param robot_state The RobotStatePtr instance.
 * @param move_group_name The name of the planning group.

File: moveit_ros/moveit_servo/src/servo.cpp [hunk 36]
```diff
@@ -532,9 +539,24 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
     {
       RCLCPP_DEBUG_STREAM(logger_, "Joint velocity limit scaling applied by a factor of " << joint_limit_scale);
     }
-
     target_state.velocities *= joint_limit_scale;
 
+    // Scale down the acceleration based on joint acceleration limit if applicable.
+    // TODO move to function: ApplyAccelerationLimit(...)
```
sea-bass, line 545: Did you mean to do this TODO for this PR? I'm fine either way, but I'd say either implement this in this PR or remove the comment.

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -532,9 +539,24 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
...
+    Eigen::VectorXd accel = (target_state.velocities - current_state.velocities) / (servo_params_.publish_period);
+    double min_scaling_factor = 1.0;
+    for (double val : accel)
```
sea-bass, line 548: Const. Also, the names `val` and `accel` aren't descriptive enough -- maybe call the vector `target_accelerations` and then do

```
for (const double accel : target_accelerations)
```

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -532,9 +539,24 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
...
+    {
+      // Find the ratio of clamped velocity to original velocity
+      const auto bounded_acc =
+          std::clamp(val, -servo_params_.joint_acceleration_limit, servo_params_.joint_acceleration_limit);
+      if (std::abs(val) > 0.0)
```
sea-bass, line 553: I feel in this case we can just do `if (val != 0.0)` then.

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -649,39 +670,38 @@ std::optional<PoseCommand> Servo::toPlanningFrame(const PoseCommand& command, co
...
 
 std::pair<bool, KinematicState> Servo::smoothHalt(const KinematicState& halt_state)
 {
   bool stopped = false;
   auto target_state = halt_state;
-  const KinematicState current_state = getCurrentRobotState();
 
-  const size_t num_joints = current_state.joint_names.size();
+  // scale target velocities according to acceleration limits
+  double max_speed = target_state.velocities.cwiseAbs().maxCoeff();
```
sea-bass, line 682: const

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -649,39 +670,38 @@ std::optional<PoseCommand> Servo::toPlanningFrame(const PoseCommand& command, co
...
+  double scale = 1.0;
+  if (max_speed > 0)
+  {
+    double new_speed = std::max(0.0, max_speed - servo_params_.joint_acceleration_limit * servo_params_.publish_period);
```
sea-bass, line 686: const

moveit_ros/moveit_servo/src/servo_node.cpp
```diff
@@ -214,12 +214,12 @@ std::optional<KinematicState> ServoNode::processJointJogCommand()
...
   {
     JointJogCommand command{ latest_joint_jog_.joint_names, latest_joint_jog_.velocities };
-    next_joint_state = servo_->getNextJointState(command);
+    next_joint_state = servo_->getNextJointState(robot_state, command);
   }
   else
   {
     auto result = servo_->smoothHalt(last_commanded_state_);
-    new_joint_jog_msg_ = result.first;
+    new_joint_jog_msg_ = !result.first;
```
sea-bass, line 222: Was this a previous bug? Could you explain why this has a `!` in front of it now?

moveit_ros/moveit_servo/src/servo_node.cpp
```diff
@@ -246,12 +246,12 @@ std::optional<KinematicState> ServoNode::processTwistCommand()
...
                                                latest_twist_.twist.angular.y, latest_twist_.twist.angular.z };
     const TwistCommand command{ latest_twist_.header.frame_id, velocities };
-    next_joint_state = servo_->getNextJointState(command);
+    next_joint_state = servo_->getNextJointState(robot_state, command);
   }
   else
   {
     auto result = servo_->smoothHalt(last_commanded_state_);
-    new_twist_msg_ = result.first;
+    new_twist_msg_ = !result.first;
```
sea-bass, line 254: same here re: the previous comment

moveit_ros/moveit_servo/src/servo_node.cpp
```diff
@@ -275,12 +275,12 @@ std::optional<KinematicState> ServoNode::processPoseCommand()
...
   {
     const PoseCommand command = poseFromPoseStamped(latest_pose_);
-    next_joint_state = servo_->getNextJointState(command);
+    next_joint_state = servo_->getNextJointState(robot_state, command);
   }
   else
   {
     auto result = servo_->smoothHalt(last_commanded_state_);
-    new_pose_msg_ = result.first;
+    new_pose_msg_ = !result.first;
```
sea-bass, line 283: and here too

moveit_ros/moveit_servo/src/servo_node.cpp
```diff
@@ -297,6 +297,22 @@ void ServoNode::servoLoop()
   std::optional<KinematicState> next_joint_state = std::nullopt;
   rclcpp::WallRate servo_frequency(1 / servo_params_.publish_period);
 
+  // wait for first robot joint state update
+  auto servo_node_start = node_->now();
```
sea-bass, line 301: const

moveit_ros/moveit_servo/src/utils/common.cpp
```diff
@@ -42,7 +42,12 @@
 namespace
 {
 // The threshold above which `override_velocity_scaling_factor` will be used instead of computing the scaling from joint bounds.
-const double SCALING_OVERRIDE_THRESHOLD = 0.01;
+constexpr double SCALING_OVERRIDE_THRESHOLD = 0.01;
+// The angle threshold in radians below which a rotation will be considered the identity.
+constexpr double MIN_ANGLE_THRESHOLD = 1E-16;
+// A minimum of 3 points are used to help with interpolation when creating a trjectory messages.
```
sea-bass, line 48: typo

```suggestion
// A minimum of 3 points are used to help with interpolation when creating trajectory messages.

File: moveit_ros/moveit_servo/src/utils/common.cpp [hunk 46]
```diff
@@ -42,7 +42,12 @@
...
+constexpr int NUM_POINTS_FOR_TRAJ_MSG = 3;
```
sea-bass, line 49: maybe call it `MIN_POINTS_FOR_TRAJ_MSG`?

moveit_ros/moveit_servo/src/utils/common.cpp
```diff
@@ -115,10 +120,13 @@ geometry_msgs::msg::Pose poseFromCartesianDelta(const Eigen::VectorXd& delta_x,
 
   // Get a transformation matrix with desired orientation change
   Eigen::Isometry3d tf_rot_delta(Eigen::Isometry3d::Identity());
-  const Eigen::Quaterniond q = Eigen::AngleAxisd(delta_x[3], Eigen::Vector3d::UnitX()) *
-                               Eigen::AngleAxisd(delta_x[4], Eigen::Vector3d::UnitY()) *
-                               Eigen::AngleAxisd(delta_x[5], Eigen::Vector3d::UnitZ());
-  tf_rot_delta.rotate(q);
+  Eigen::Vector3d rot_vec = delta_x.block<3, 1>(3, 0, 3, 1);
+  double angle = rot_vec.norm();
```
sea-bass, line 124: const

moveit_ros/moveit_servo/src/utils/common.cpp
```diff
@@ -138,53 +146,115 @@ geometry_msgs::msg::Pose poseFromCartesianDelta(const Eigen::VectorXd& delta_x,
...
-  if (point.positions.size() != num_joints)
+  joint_trajectory.joint_names = joint_cmd_rolling_window.front().joint_names;
+  joint_trajectory.points.reserve(joint_cmd_rolling_window.size() + 1);
+  joint_trajectory.header.stamp = joint_cmd_rolling_window.front().time_stamp;
+
+  auto add_point = [servo_params](trajectory_msgs::msg::JointTrajectory& joint_trajectory, const KinematicState& state) {
+    trajectory_msgs::msg::JointTrajectoryPoint point;
+    size_t num_joints = state.positions.size();
+    point.positions.reserve(num_joints);
+    point.velocities.reserve(num_joints);
```
sea-bass, line 167: should also reserve `points.accelerations` right?

moveit_ros/moveit_servo/src/utils/common.cpp
```diff
@@ -138,53 +146,115 @@ geometry_msgs::msg::Pose poseFromCartesianDelta(const Eigen::VectorXd& delta_x,
...
+    auto& last_state = joint_cmd_rolling_window[num_points - 1];
+    auto& second_last_state = joint_cmd_rolling_window[num_points - 2];
+
+    Eigen::VectorXd direction_1 = second_last_state.positions - last_state.positions;
+    Eigen::VectorXd direction_2 = next_joint_state.positions - last_state.positions;
+    Eigen::VectorXd signs = (direction_1.array().sign() - direction_2.array().sign()).round();
+    for (long i = 0; i < last_state.velocities.size(); ++i)
     {
-      point.accelerations[i] = joint_state.accelerations[i];
+      if (abs(signs[i]) < 1.5)
```
sea-bass, line 232: What's this 1.5? Is this basically checking that there are more than 1 sign flip? I think this value is fine, but may be worth a quick comment

moveit_ros/moveit_servo/src/utils/common.cpp
```diff
@@ -138,53 +146,115 @@ geometry_msgs::msg::Pose poseFromCartesianDelta(const Eigen::VectorXd& delta_x,
...
+        auto velocity_1 = (next_joint_state.positions[i] - last_state.positions[i]) / delta_time_1;
+        auto velocity_2 = (last_state.positions[i] - second_last_state.positions[i]) / delta_time_2;
+        if (std::abs(velocity_1) < std::abs(velocity_2))
+        {
+          last_state.velocities[i] = velocity_1;
+        }
+        else
+        {
+          last_state.velocities[i] = velocity_2;
+        }
```
sea-bass, line 250:
```suggestion
       last_state.velocities[i] = (std::abs(velocity_1) < std::abs(velocity_2)) ? velocity_1 : velocity_2;
```

moveit_ros/moveit_servo/src/utils/common.cpp
```diff
@@ -138,53 +146,115 @@ geometry_msgs::msg::Pose poseFromCartesianDelta(const Eigen::VectorXd& delta_x,
...
+      {
+        // direction changed
+        last_state.velocities[i] = 0;
+      }
+      else
+      {
+        double delta_time_1 = (next_joint_state.time_stamp - last_state.time_stamp).seconds();
+        double delta_time_2 = (last_state.time_stamp - second_last_state.time_stamp).seconds();
+        auto velocity_1 = (next_joint_state.positions[i] - last_state.positions[i]) / delta_time_1;
+        auto velocity_2 = (last_state.positions[i] - second_last_state.positions[i]) / delta_time_2;
```
sea-bass, line 242: these can all be const

sea-bass (commented):
Awesome, things are looking like they got simplified without the accel filter.

To that point, I am not sure this new `joint_acceleration_limit` parameter is the way to go. Minimally, this should be a per-joint limit like the joint limit margins, but ideally we can use the actual joint acceleration limits from the robot model itself? I believe MoveIt requires these now for the motion planning side anyway.

moveit_ros/moveit_servo/config/panda_simulated_config.yaml
AndyZe: ^agree with that. Let's not add a new parameter.

The big-picture plan is to add a Ruckig smoothing plugin that would take care of acceleration limits and replace the ButterworthFilter.

moveit_ros/moveit_servo/src/servo.cpp
AndyZe: I'd rather get rid of the acceleration limiting here completely. IMO it belongs as an online_signal_smoothing plugin to replace the ButterworthFilter. It should be a drop-in replacement with no code changes required in the Servo pkg.

moveit_ros/moveit_servo/config/panda_simulated_config.yaml
pac48: This is not a limit imposed by the hardware. The acceleration limit here acts as a smoother. I think it is important to distinguish between desired acceleration/deacceleration and hardware limit values. You rarely want to go at the hardware limits unless you are about to crash. The name is probably not clear enough.

Getting Ruckig smoothing integrated sounds like a good next step. In the meantime, we need a mechanism to enhance servo's acceleration/deacceleration for trajectory interpolation. The current implementation introduces overshoot when combined with trajectory interpolation.

sea-bass: >  You rarely want to go at the hardware limits unless you are about to crash.

The same is true for URDF joint limits. Usually they dictate "max desired acceleration for planned motions". Which I think is why @AndyZe and I are taking this stance.

moveit_ros/moveit_servo/demos/cpp_interface/demo_joint_jog.cpp
pac48: Are yo referring to `msg`?

sea-bass: Yeah, sorry -- and the "same" is for the twist/pose demo files too.

moveit_ros/moveit_servo/src/servo.cpp
sea-bass: This is what @pac48 had in previous versions of this PR, right?

moveit_ros/moveit_servo/src/utils/common.cpp
pac48: I added a comment explaining

moveit_ros/moveit_servo/src/utils/common.cpp
pac48: This is much better!

moveit_ros/moveit_servo/config/panda_simulated_config.yaml
pac48: Does the URDF even specify acceleration? It has a field for effort, but that is N-m not m/s^2.

sea-bass: It sadly does not... it's why MoveIt has its own joint limits file that gets parsed in the config parser:

https://github.com/ros-planning/moveit_resources/blob/ros2/panda_moveit_config/config/joint_limits.yaml

pac48: I added parsing to extract values from Moveit's internal representation. Additionally, I added a linear scalar parameter for acceleration limits.

pac48:
I made a short video showing servo on the Kinova gen3. I was pushing it to the limits but was still able to generate smooth and fast motions. There is some shaking, but that is mainly from the table the robot is on and the robot itself since it is moving fast. 
https://drive.google.com/file/d/1OxJkykwgWPxXGaNZr1yZIT3yich_kcQ8/view

AndyZe:
OK, but the motion was also smooth at commit 55c76bc4, right?

I really don't think acceleration smoothing should go in Servo. Frankly, the acceleration smoothing algorithm in this PR looks crude compared to Ruckig and I wouldn't want to run it in series with Ruckig.

pac48:
> OK, but the motion was also smooth at commit 55c76bc, right?

It is actually smoother now because I made some changes to avoid overshooting in trajectory interpolation. Also, the acceleration limiting does help. Here are some plots. You can compare them with the original plots from the first post.    

Screenshot from 2023-12-21 15-02-25
Screenshot from 2023-12-21 15-02-31

pac48:
> I really don't think acceleration smoothing should go in Servo. Frankly, the acceleration smoothing algorithm in this PR looks crude compared to Ruckig and I wouldn't want to run it in series with Ruckig.

The acceleration limiting is a parameter that can be turned off, so it should not be a problem for your use case. I think most users will interface with servo via the topics interface, so I don't see the harm. However, it's possible to factor the acceleration limiting out of core servo and into `servo_node`.

sea-bass:
@AndyZe I think these fixes have shown clear improvements to the status quo for position controlled robots -- both in the accompanying plots and videos in this PR. So even if not perfect, it's definitely a step up.l and we want to roll this out onto our Kinova robots ASAP.

What do you think would be a good state that retains these fixes, but also leaves the door open for possible improvements.

For example, do you think we could possibly merge something close-ish to what is here, and then make a follow up issue to prove out Ruckig as a smoothing plugin? And if that all goes well, we can roll back the then-unnecessary acceleration smoothing code?

I spoke to @pac48 earlier about some issues with the current acceleration scaling math (which I'll comment in more detail later), but otherwise I can't think of anything else that wouldn't be a nontrivial amount of work.

moveit_ros/moveit_servo/config/panda_simulated_config.yaml
sea-bass: Gotcha, then this line should now be removed from this file.

moveit_ros/moveit_servo/config/servo_parameters.yaml
```diff
@@ -87,6 +98,11 @@ servo:
       default_value: 0.5,
       description: "Max joint angular/linear velocity. Only used for joint commands on joint_command_in_topic."
     }
+    joint_acceleration: {
+      type: double,
+      default_value: 1.0,
+      description: "Scale the maximum allowable joint acceleration specified in the joint limits configuration."
```
sea-bass, line 104: Can you specify the units here? Assuming `rad/s^2`?

Also if you compare with the other scale parameters, notice those define a maximum velocity whereas this one defines a scale? Make sure it's consistent.

moveit_ros/moveit_servo/config/test_config_panda.yaml
```diff
@@ -29,6 +30,7 @@ publish_joint_accelerations: false
 ## Plugins for smoothing outgoing commands
 use_smoothing: false
 smoothing_filter_plugin_name: "online_signal_smoothing::ButterworthFilterPlugin"
+joint_acceleration_limit: 500.0
```
sea-bass, line 33: again, remove this line in favor of `joint_acceleration` up top.

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
```diff
@@ -232,6 +230,9 @@ class Servo
 
   // Map between joint subgroup names and corresponding joint name - move group indices map
   std::unordered_map<std::string, JointNameToMoveGroupIndexMap> joint_name_to_index_maps_;
+
+  // joint limit acceleration for most restricted joint
+  double joint_acceleration_limit = 0.0;
```
sea-bass, line 235: I don't think we need this variable, will comment in the part where this is calculated.

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -135,6 +135,23 @@ Servo::Servo(const rclcpp::Node::SharedPtr& node, std::shared_ptr<const servo::P
...
+  for (const auto& bound : joint_bounds)
+  {
+    for (const auto& b1 : *bound)
+    {
+      double max_acceleration = servo_params_.scale.joint_acceleration *
+                                std::max(std::abs(b1.min_acceleration_), std::abs(b1.min_acceleration_));
+      joint_acceleration_limit =
+          joint_acceleration_limit < 0 ? max_acceleration : std::min(joint_acceleration_limit, max_acceleration);
+    }
+  }
```
sea-bass, line 153: Few comments here:

First off, these acceleration joint bounds I don't think are correct for a couple reasons.
* You don't just want to take the lowest value of all the joints; rather, you want to look at the un-limited target command and determine the limiting joint such that scaling down meets all the joint limits.
* This means that this shouldn't be computed as a member variable of servo, but just  with local variables at every iteration.
* If the scale parameter is acting as a true scale factor value, we should disallow exceeding the joint limits, meaning the scale always needs to be in the range `(0..1]`.
* If this is not acting as a scale factor, but as an actual acceleration in rad/s^2, then similarly you want to scale down all the computed target accelerations and scale them down to the value provided... UNLESS they exceed the actual joint's limits.
* For better variable naming, see what is done for velocity scaling: https://github.com/ros-planning/moveit2/blob/main/moveit_ros/moveit_servo/src/utils/common.cpp#L324-L338

As such, I think this should be moved to a utility in a place similar to what I linked above, and ensure it's unit tested.

---

Secondly, what does it mean when the robot does not have any max accelerations provided? This again boils down to a similar question as above:

* If the scale parameter is acting as a true scale factor, then we need to check if the joint HAS acceleration bounds (there are variables/functions for this in the joint bounds), and if not then don't limit acceleration.
* If this is not acting as a scale factor, but as an actual acceleration in rad/s^2 ... then we just take the parameter as max acceleration and move on with scaling.

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -135,6 +135,23 @@ Servo::Servo(const rclcpp::Node::SharedPtr& node, std::shared_ptr<const servo::P
...
+      robot_state->getJointModelGroup(servo_params_.move_group_name);
+  const std::vector<std::string> joint_names = joint_model_group->getActiveJointModelNames();
+  const moveit::core::JointBoundsVector joint_bounds = joint_model_group->getActiveJointModelsBounds();
+  joint_acceleration_limit = -1.0;
+  for (const auto& bound : joint_bounds)
+  {
+    for (const auto& b1 : *bound)
+    {
+      double max_acceleration = servo_params_.scale.joint_acceleration *
+                                std::max(std::abs(b1.min_acceleration_), std::abs(b1.min_acceleration_));
```
sea-bass, line 149: even if this were the correct math, I think this is a typo:

```suggestion
                                std::max(std::abs(b1.min_acceleration_), std::abs(b1.max_acceleration_));
```

... but again, this would of course depend on whether the commanded acceleration is in the positive or negative direction, so I think the logic should be flipped:

* First check whether the joint needs to move in the positive or negative direction
* THEN pick either min or max acceleration

moveit_ros/moveit_servo/src/utils/common.cpp
```diff
@@ -138,53 +146,111 @@ geometry_msgs::msg::Pose poseFromCartesianDelta(const Eigen::VectorXd& delta_x,
...
+
+    Eigen::VectorXd direction_1 = second_last_state.positions - last_state.positions;
+    Eigen::VectorXd direction_2 = next_joint_state.positions - last_state.positions;
+    Eigen::VectorXd signs = (direction_1.array().sign() - direction_2.array().sign()).round();
+    for (long i = 0; i < last_state.velocities.size(); ++i)
     {
-      point.accelerations[i] = joint_state.accelerations[i];
+      // check if the direction have changed. `signs` will either have -2, +2 meaning a flat line, -1, 1 meaning a
+      // rotated L shape, or 0 meaning a v-shape.
+      if (abs(signs[i]) == 0.0)
```
sea-bass, line 235: If you're just checking equality to 0.0, no need for the `abs` here

sea-bass (commented):
Formulated my thoughts here on the acceleration scaling math. I'm OK with the general structure of the code right now, but there are 3 main things to consider:

* Is this parameter a scaling factor of max acceleration, or a max acceleration in rad/s^2 that competes with the actual joint acceleration limits in the robot model?
* What to do when the robot has no acceleration limits specified? Depends on the previous question.
* The math for finding the minimum acceleration isn't a fixed calculation up front, but rather a calculation that needs to happen at every iteration based on the provided command.

Also, with all these changes seems like the actual integration tests are failing as there are some comparison operators on the final state of the robot that are no longer being met. Please check those, noting that this is separate from the actual segfaults we're seeing in the `humble_ci` job.

AndyZe:
I'll give this a good review by EOD Tuesday. I think I've been doing a decent job staying up-to-date considering the PR has been completely rewritten like 3 times  ;)   It takes some time to give a quality review.

moveit_ros/moveit_servo/config/servo_parameters.yaml
pac48: I removed all uses of the `joint_acceleration_limit` in favor of taking values from the `robot_state`

moveit_ros/moveit_servo/config/test_config_panda.yaml
pac48: Done

moveit_ros/moveit_servo/include/moveit_servo/servo.hpp
pac48: Removed

moveit_ros/moveit_servo/src/servo.cpp
pac48: > First off, these acceleration joint bounds 

Thank you for pointing this out. What I did was make a new function called `jointLimitAccelerationScalingFactor` which is like `jointLimitVelocityScalingFactor` except that it acts on acceleration instead of velocity.  Unfortunately, it is a lot of copied code. There are two ways to refactor that I can see. Have both methods call an internal method that 1) takes lambda that calculates the scaling factor for either velocity or acceleration or 2) takes a boolean flag and has an `if else` to either do velocity or acceleration.

> If the scale parameter is acting as a true scale factor value, we should disallow exceeding the joint limits, meaning the scale always needs to be in the range (0..1].    

We need to rethink scaling. I think a unitless factor makes more sense than an actual limit in m/s^2. I think we should have a separate parameter for scaling the dynamic joint limits. This would apply to both velocity and acceleration. None of the existing parameters actually scale these limits, it's either scaling the command or scaling the velocity to be a certain amount less than the limit. 

> Secondly, what does it mean when the robot does not have any max accelerations provided?

I think if the robot has no maximum acceleration, then we should either 1) not apply the acceleration scaling or 2) throw an exception. I prefer option 1), especially since the velocity scaling already works that way.

moveit_ros/moveit_servo/src/servo.cpp
pac48: This is now handled in the new `jointLimitAccelerationScalingFactor` method.

moveit_ros/moveit_servo/src/utils/common.cpp
pac48: Removed the `abs`

moveit_ros/moveit_servo/src/servo.cpp
pac48: I also added a test for `jointLimitAccelerationScalingFactor`

sea-bass:
> I'll give this a good review by EOD Tuesday. I think I've been doing a decent job staying up-to-date considering the PR has been completely rewritten like 3 times ;) It takes some time to give a quality review.

Of course! It wasn't so much about time urgency as it was about finding a good path forward that would get the PR to a mergeable state. Just wanted to make sure we can give @pac48 feedback so it's not a 4th+ complete rewrite.

moveit_ros/moveit_servo/src/servo.cpp
sea-bass: I like option 1 for both of these.

I was recently told that having a boolean parameter for a function to do two different things means you should have 2 separate functions... or make it generic via things like templates/lambdas

moveit_ros/moveit_servo/src/servo_node.cpp
pac48: Yes, I believe that was a bug. `smoothHalt` returns true if the joints are stopped. Until the joints are actually stopped, `new_joint_jog_msg_` should remain true so that `smoothHalt` keeps getting called. At some point, the joints will be stopped, which means `!result.first` will be false, `new_joint_jog_msg_` will be false, and the `processJointJogCommand` will stop getting called.

moveit_ros/moveit_servo/src/servo_node.cpp
pac48: See previous response

moveit_ros/moveit_servo/src/servo_node.cpp
pac48: See previous response

moveit_ros/moveit_servo/src/utils/common.cpp
pac48: Agree, much better name

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -532,9 +538,21 @@ KinematicState Servo::getNextJointState(const ServoInput& command)
     {
       RCLCPP_DEBUG_STREAM(logger_, "Joint velocity limit scaling applied by a factor of " << joint_limit_scale);
     }
-
     target_state.velocities *= joint_limit_scale;
```
AndyZe, line 541: Naming, to be consistent...

```suggestion
    target_state.velocities *= joint_velocity_limit_scale;
```

moveit_ros/moveit_servo/config/panda_simulated_config.yaml
```diff
@@ -6,7 +6,8 @@
 # override_velocity_scaling_factor = 0.0 # valid range [0.0:1.0]
```
AndyZe, line 6:
```suggestion
# override_velocity_scaling_factor = 0.0 # valid range [0.0:1.0], 0.0 disables it
# override_acceleration_scaling_factor = 0.0 # valid range [0.0:1.0], 0.0 disables it
```

moveit_ros/moveit_servo/config/panda_simulated_config.yaml
```diff
@@ -26,7 +27,7 @@ publish_joint_velocities: true
 publish_joint_accelerations: false
 
 ## Plugins for smoothing outgoing commands
-use_smoothing: true
+use_smoothing: false
```
AndyZe, line 30: This is going to default to no smoothing at all, since `override_acceleration_scaling_factor` is zero too. Prob not what we want. Suggest reverting. Were your tests done with these settings?

moveit_ros/moveit_servo/src/utils/common.cpp
```diff
@@ -319,27 +387,53 @@ double jointLimitVelocityScalingFactor(const Eigen::VectorXd& velocities,
...
-      if (variable_bound.velocity_bounded_ && target_vel != 0.0)
-      {
-        // Find the ratio of clamped velocity to original velocity
-        const auto bounded_vel = std::clamp(target_vel, variable_bound.min_velocity_, variable_bound.max_velocity_);
-        const auto joint_scaling_factor = bounded_vel / target_vel;
-        min_scaling_factor = std::min(min_scaling_factor, joint_scaling_factor);
-      }
+      const auto& joint_target = joint_values(idx);
+      double joint_scaling_factor = calculate_joint_scaling_factor(variable_bound, joint_target);
+      min_scaling_factor = std::min(min_scaling_factor, joint_scaling_factor);
```
AndyZe, line 398: With velocities it makes sense to slow all of the velocities down by the same ratio. That keeps the robot on the intended path.

It doesn't really make sense with accelerations; it will cause the robot to go off the path. However, Ruckig would handle this well.

AndyZe (commented):
I'd be cool with merging after any of these options:

- get rid of `override_acceleration_scaling_factor`. It has a bad smell to it.
- Move `override_acceleration_scaling_factor` to an `online_signal_smoothing` plugin so knowledgeable people can accept the path-following bug if they want to. It seems like a good tradeoff on your Kinova for example.

As an alternative to this new acceleration smoothing, you probably could just increase the filter coefficient here:  https://github.com/ros-planning/moveit2/blob/main/moveit_core/online_signal_smoothing/src/butterworth_parameters.yaml

moveit_ros/moveit_servo/config/panda_simulated_config.yaml
pac48: Yes, all of the hardware testing I did was with `ButterworthFilterPlugin` smoothing turned off and acceleration limits enabled. I think depending on the task, it makes sense to have it enabled. We can enable it by default.

pac48:
> I'd be cool with merging after any of these options:
> 
> * get rid of `override_acceleration_scaling_factor`. It has a bad smell to it.
> * Move `override_acceleration_scaling_factor` to an `online_signal_smoothing` plugin so knowledgeable people can accept the path-following bug if they want to. It seems like a good tradeoff on your Kinova for example.

Okay, I removed the `override_acceleration_scaling_factor` parameter.  

> 
> As an alternative to this new acceleration smoothing, you probably could just increase the filter coefficient here: https://github.com/ros-planning/moveit2/blob/main/moveit_core/online_signal_smoothing/src/butterworth_parameters.yaml

The one problem I have seen with increasing the filter coefficient is that it alters the velocity of the robot. For example, if it is commanded to go `0.25 rad/s` it might actually go `0.10 rad/s`. 

I think we need a parameter to enable/disable the acceleration limits. I'm just not sure where that parameter should be. We could add a new parameter in `servo_parameters.yaml` called `apply_acceleration_limits` that can be true/false.

moveit_ros/moveit_servo/src/utils/common.cpp
pac48: When you say "path" are you referring to a pre-defined/pre-calculated trajectory? For example, maybe you want to move the robot in a square pattern, so some trajectory optimizer/planner creates the trajectory and then the waypoints are fed into Servo?

AndyZe: What i really mean is direction. When this acceleration gets integrated it will change the direction.

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -649,39 +667,41 @@ std::optional<PoseCommand> Servo::toPlanningFrame(const PoseCommand& command, co
...
+  // Get joint bounds from robot state for current move group.
+  const moveit::core::JointModelGroup* joint_model_group =
+      robot_state->getJointModelGroup(servo_params_.move_group_name);
+  const moveit::core::JointBoundsVector joint_bounds = joint_model_group->getActiveJointModelsBounds();
+
+  // apply scaling to target velocity based on robot's limits
+  Eigen::VectorXd acceleration = -target_state.velocities / servo_params_.publish_period;
+  const double joint_acceleration_limit_scale = jointLimitAccelerationScalingFactor(acceleration, joint_bounds, 1.0);
+  target_state.velocities += joint_acceleration_limit_scale * acceleration * servo_params_.publish_period;
+
```
AndyZe, line 688: This acceleration scaling is what I was trying to get you to remove with the previous comments. The ones about changing direction and `override_acceleration_scaling_factor`.

```suggestion
```

AndyZe (commented):
Almost there, in my opinion

AndyZe:
> The one problem I have seen with increasing the filter coefficient is that it alters the velocity of the robot.

If we filtered velocities instead of positions --> problem solved. There's an issue about that.

moveit_ros/moveit_servo/src/utils/common.cpp
```diff
@@ -138,53 +146,112 @@ geometry_msgs::msg::Pose poseFromCartesianDelta(const Eigen::VectorXd& delta_x,
...
-    {
-      point.velocities[i] = joint_state.velocities[i];
-    }
+    joint_cmd_rolling_window.pop_back();
   }
-  if (servo_params.publish_joint_accelerations)
+
+  // update velocity: the velocity has the potential to dramatically influence interpolation of splines and causes large
+  // overshooting. To alleviate this effect, the target velocity will be set to zero if the motion changes direction,
+  // otherwise, it will calculate the forward and backward finite difference velocities and choose the minimum.
```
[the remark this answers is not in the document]
pac48: No, there isn't a reference algorithm for this. I would prefer to have a spline method like PCHIP that doesn't have so much overshoot like regular cubic splines. Setting the velocity boundary conditions does help with overshooting. I think this could use some work. 

I think we can do better here. One problem is that Euler integration is being used to calculate the position and velocity of the next state given the current state and command. Since the integration isn't completely accurate, it can affect the boundary conditions. However, if we use a constant acceleration, then we can use the analytical integral solution. If the analytical solution is used, then we probably won't need the special checks in this method.

moveit_ros/moveit_servo/src/servo.cpp
pac48: Okay, I see. One problem is that the robot will be commanded to zero velocity immediately if we don't limit the acceleration. I wouldn't call that a smooth hault. The smoothing might help make the stop more gradual. But what benefit is there for gradual slow down via smoothing (e.g. butterworth filtering) vs. acceleration limited? One benefit of acceleration limited is that it will slow down as quickly as possible while not violating the hardware limits.

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -284,6 +284,15 @@ bool Servo::validateParams(const servo::Params& servo_params) const
     params_valid = false;
   }
 
+  if (servo_params.max_expected_latency / 3 < servo_params.publish_period)
```
sea-bass, line 287: This `3` in here is a magic number, but I'm assuming this is related to you wanting 3 points in your trajectory, right? Is this the same as `MIN_POINTS_FOR_TRAJ_MSG` you defined in `common.cpp`?

If that's not the case, you can probably leave in the literal here (especially since it's also in the error message), but definitely add a comment explaining.

moveit_ros/moveit_servo/src/servo.cpp
sea-bass: I think @AndyZe's comment about direction switching is valid, and I hadn't thought of that since I was thinking about the full trajectory planning case where scaling down acceleration + scaling up time accelerating works equivalently to the unscaled case, as far as the space swept by the robot motion.

Is there any way to do the same within the current servo framework now that we have the rolling window? Like, time scaling https://www2.ece.ohio-state.edu/~zhang/RoboticsClass/docs/LN9_TrajectoryGeneration.pdf

sea-bass (commented):
I see @AndyZe's point about scaling acceleration doing weird things to the direction being tracked, and left a comment inline asking if there is a way to "time scale" the bundled up trajectory in a way that is more correct.

The other comment about trying velocity filtering with the Butterworth plugin is something we could explore? I tried with these changes and it made my Gazebo simulation do horrible things, but that could have been a combination of Gazebo's physics and some other issues that may have been resolved since (including in this PR). I guess it's worth a shot to try it and get some data.

moveit_ros/moveit_servo/src/servo.cpp
pac48: I think it's correct to relate the 3 here to `MIN_POINTS_FOR_TRAJ_MSG`. Do you think I should move `MIN_POINTS_FOR_TRAJ_MSG` to `moveit_servo/utils/common.hpp` so it can be accessed?

moveit_ros/moveit_servo/src/servo.cpp
AndyZe: Suggest moving the acceleration smoothing to a new PR so we can merge this one.

Again, I think an acceleration-smoothing plugin in `online_signal_smoothing` is the way to go.

pac48: @sea-bass I like the idea of generating a path and then time scaling to enforce dynamic constraints. The biggest problem is that a twice-differentiable path is needed. We would need to fill in the path between the waypoints with a smooth function. One option is to use circular arcs like done in [here](https://roboticsproceedings.org/rss08/p27.pdf) section IV. 

For the full trajectory planning case, deviation from the path found by the global planner is detrimental because those are the points that are collision checked, so it could result in a collision. However, since servo is constantly running collision checking, this is not a big concern. Maybe this graphic will help clearify.

Screenshot from 2023-12-28 10-48-42

In this scenario, the blue dots represent the commands the user sends to the robot. The plot on the left shows the user commanding a sharp 90 degree angle. On the right, the same situation is shown but with acceleration limiting applied. Because the command violates the robot's acceleration limits, the commands are modified (shown in red). In this case, the red dots are points that are being collision checked, so colliding due to deviating from the reference should not be an issue. However, if we knew we had the entire trajectory ahead of time, then we could do time parameterization in such as way that no limits are violated and the path is followed exactly.

Does anyone have thoughts in this?

pac48: > Suggest moving the acceleration smoothing to a new PR so we can merge this one.
> 
> Again, I think an acceleration-smoothing plugin in `online_signal_smoothing` is the way to go.

Maybe that is the best option.

pac48:
> If we filtered velocities instead of positions --> problem solved. There's an issue about that.

@AndyZe  If we filter in velocity, won't we run into the same issue of the robot not moving in the direction commanded? That is the main criticism of acceleration limiting.

AndyZe:
True. Well, it's an optional plugin. Hopefully Ruckig will come along and solve all of our issues. Happy to help make that happen

pac48:
> True. Well, it's an optional plugin. Hopefully Ruckig will come along and solve all of our issues. Happy to help make that happen

Okay great. I removed the acceleration-based limits from this PR. I will follow up with an `online_signal_smoothing` optional plugin in a future PR if needed.

AndyZe (approved):
yay!

moveit_ros/moveit_servo/src/servo.cpp
sea-bass: I think so!

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -649,39 +657,34 @@ std::optional<PoseCommand> Servo::toPlanningFrame(const PoseCommand& command, co
...
+  // set target velocity
+  target_state.velocities *= 0.0;
+
+  // apply smoothing: this will change target position/velocity to make slow down gradual
+  doSmoothing(target_state);
+
+  // scale velocity in case of obstacle
+  target_state.velocities *= collision_velocity_scale_;
+
+  for (size_t i = 0; i < (size_t)halt_state.positions.size(); ++i)
```
sea-bass, line 677: Do you need to cast the size here? If so, consider using `static_cast` instead.

pac48: It’s only needed because of the warning from comparing signed with unsigned numbers. I think I’d prefer to make the indexing variable a long instead.

sea-bass (approved):

sjahr:
@pac48 Can you address the conflicts? Then we should merge this, nice work!

sea-bass:
> @pac48 Can you address the conflicts? Then we should merge this, nice work!

I just addressed them, they were simple and from my recent PR, so let's do this!

Issue #2548, opened 2023-11-23
Servo node allows the robot to collide with the Octomap


amalnanavati:
### Description

I have a MoveGroup node running, and its planning scene has a couple of mesh and geometric shapes and an Octomap. I also have a Servo node running, which is configured to not be the primary planning scene monitor (because MoveGroup is). 

Both MoveGroup and Servo register collisions between the robot and mesh/geometric shapes in the planning scene, as expected. However, only MoveGroup registers collisions between the robot and the Octomap; **Servo does not register collisions with the Octomap**. More specifically, I can use Servo to move the robot directly in collision with the Octomap; MoveGroup (as visualized in RVIZ) recognizes that the robot is in collision, but Servo does not.

### Your environment
* ROS Distro: Humble
* OS Version: 22.04
* Source or Binary build? Source
* If binary, which release version? 2.5.5-1jammy.20230920.090831
* If source, which branch? N/A
* Which RMW (Fast DDS or Cyclone DDS)? Cyclone DDS

### Steps to reproduce
I can provide a working demo if need be. However, because I have traced through the code and know where the issue is (see below), I instead opt for a high-level description of the steps I followed. Let me know if you'd like an example.

1. Launch a MoveGroup node and a servo node configured to **not** be the primary planning scene monitor.
2. Populate the octomap with obstacles within the robot's reach.
3. Using keyboard teleop, servo the robot into the obstacles.

### Expected behaviour
The Servo node should decelerate and stop before the robot comes into contact with the Octomap obstacles.

### Actual behaviour
The Servo node allows us to move the robot directly into the Octomap obstacles.

### Suggested Solution

Looking through the code, I believe the issue is as follows:

[...]

PR #2601, opened 2023-12-13
Make `moveit_servo` listen to Octomap updates

amalnanavati:
### Description

As explained in #2548, the current `moveit_servo` does not respect Octomap collisions. Responding to that, this PR makes the following one-line changes:

1. [Addressing the issue] Start the world geometry monitor when initializing the servo node.

(Instead of enabling the world geometry monitor by default, I'm happy to change it to being controlled by a boolean parameter. Maintainers, let me know if you'd prefer that.)

### Testing

I have tested this code on our local robot setup in ROS2 Humble, where a `move_group` node is the primary planning scene monitor, and a `servo_node` subscribes to its updates over the "monitored_planning_scene" topic. As expected, with these changes the robot treats octomap collisions as any other collision in the planning scene, and without these changes it ignores octomap collisions.

Importantly, I have **not** tested the code in cases where the `servo_node` is the primary planning scene monitor. I imagine it should work as long as developers pass in the save octomap and `sensors_3d.yaml` configuration parameters as they would to a `move_group` with an octomap.

### Checklist
- [N/A] Extend the tutorials / documentation [reference](http://moveit.ros.org/documentation/contributing/)
- [N/A] Include a screenshot if changing a GUI


sjahr (commented):
Thanks! Please address the comment above

> Please target the main branch for development, we will backport the changes to humble for you if approved and if they don't break API.

This looks fine to me. @AndyZe is there any reason not to activate the WorldGeometryMonitor?

AndyZe (approved):

amalnanavati:
Done; I rebased onto `main`. Note that I haven't been able to test the version on `main` (since I don't use ROS2 Rolling), but given that it is a one-line change and worked on Humble, and you both have looked it over, I imagine it should be fine.

sjahr (approved):

sjahr:
Thanks for your contribution!

amalnanavati:
@sjahr any idea what might be causing `humble-ci` to fail? Looking at the [logs](https://github.com/ros-planning/moveit2/actions/runs/7199811625/job/19615742681?pr=2601#step:12:9077), it seems like `ros2_control_node` had a segmentation fault, although none of my changes should impact that.

The only error relevant to my change is that no sensor is configured for the Octomap (log [here](https://github.com/ros-planning/moveit2/actions/runs/7199811625/job/19615742681?pr=2601#step:12:9117)) although I believe that is a non-fatal error.

The only other pointer I have is that in Humble, the correct line is `planning_scene_monitor_->startWorldGeometryMonitor();` whereas in Rolling it is `planning_scene_monitor->startWorldGeometryMonitor();` (no underscore), which may impact the automatic backport.

sjahr:
I don't think that error is related to your change. I'll merge the PR

Issue #2604, opened 2023-12-14
Motion planning with multiple path constraints takes tens of seconds


medvedevigorek:
### Description

`PlanningComponent::plan()` method may takes tens of seconds to complete while the motion plan request has `planning_time` set to smaller value than actual execution time. This seems to be happening only in presence of multiple path constraints, with no path constraints the execution time is within the `planning_time`. 

### Your environment
* ROS Distro: Iron
* OS Version: e.g. Ubuntu 22.04
* Source or Binary build? source 
* If binary, which release version?
* If source, which branch? Iron (tag 2.8.0)
* Which RMW (Fast DDS or Cyclone DDS)? default rmw in Iron

### Steps to reproduce
The issue is reproducible by having 2 or more path constraints defined. We initially noticed such behavior in our application and could easily reproduce it by modifying `MoveItCpp` example by including path constraints into one of the planning requests. 

### Expected behaviour
If `planning_time` is set then `plan()` method shouldn't take more defined planning time. 

### Actual behaviour
The execution of `plan()` method may take tens of seconds. We observed up to 50-60 seconds. 

### Backtrace or Console output

```
[moveit_cpp_tutorial-4] [WARN] [1702518092.514120827] [kdl_parser]: The root link panda_link0 has an inertia specified in the URDF, but KDL does not support a root link with an inertia.  As a workaround, you can add an extra dummy link to your URDF.
[moveit_cpp_tutorial-4] [INFO] [1702518092.514153702] [moveit_kinematics_base.kinematics_base]: Joint weights for group 'panda_arm': 1 1 1 1 1 1 1
[moveit_cpp_tutorial-4] [INFO] [1702518092.514964550] [moveit.ompl_planning.model_based_planning_context]: Planner configuration 'panda_arm[PRMstarkConfigDefault]' will use planner 'geometric::PRMstar'. Additional configuration parameters will be set when the planner is constructed.
[moveit_cpp_tutorial-4] [INFO] [1702518092.525809750] [moveit.ompl_planning.model_based_planning_context]: panda_arm[PRMstarkConfigDefault]: Allocating specialized state sampler for state space
[moveit_cpp_tutorial-4] [INFO] [1702518092.525865763] [moveit.ompl_planning.model_based_planning_context]: panda_arm[PRMstarkConfigDefault]: Allocating specialized state sampler for state space
[...]

PR #2608, opened 2023-12-15
Invoke OMPL debug print only when debug logging is enabled

medvedevigorek:
### Description

A call to `ompl::base::SimpleSetup::print()` was added by #1330 to collect OMPL debug information and later log it at debug level however the actual call happens regardless of log level and as described in #2604 the call may take tens of seconds to complete hence delaying motion planning. 

The proper approach would be to call `print()` method only if log level is set to DEBUG to minimize the impact in production deployments. It's still going to be a problem when DEBUG logging is enabled though. 

As for validation, I used a modified MoveItCpp tutorial by adding path constraints and verifying that there is no additional delay in motion planning at INFO level.


sjahr (approved):
Thank you! You just need to [run pre-commit to fix the formatting](https://moveit.ros.org/documentation/contributing/code/#pre-commit-formatting-checks) and we can merge it when that last CI check passed

tylerjw (approved):

PR #2627, opened 2023-12-21
[Servo] Make listening to octomap updates optional

sea-bass:
Following from https://github.com/ros-planning/moveit2/pull/2601, the [`RCLCPP_ERROR` log message when no 3D sensor plugins are available](https://github.com/ros-planning/moveit2/blob/main/moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor_middleware_handle.cpp#L91) has been loud -- so I put this addition behind a new boolean parameter called `check_octomap_collisions`.

@amalnanavati sharing with you since you made this original PR and would appreciate your input behind this change.

In the process, I also cleaned up a few things I saw.

### DIFF
--- a/moveit_ros/moveit_servo/config/panda_simulated_config.yaml
+++ b/moveit_ros/moveit_servo/config/panda_simulated_config.yaml
@@ -33,8 +33,8 @@ smoothing_filter_plugin_name: "online_signal_smoothing::ButterworthFilterPlugin"
 # which other nodes can use as a source for information about the planning environment.
 # NOTE: If a different node in your system is responsible for the "primary" planning scene instance (e.g. the MoveGroup node),
 # then is_primary_planning_scene_monitor needs to be set to false.
-check_octomap_collisions: false
 is_primary_planning_scene_monitor: true
+check_octomap_collisions: false  # Check collision against the octomap (if a 3D sensor plugin is available)
 
 ## MoveIt properties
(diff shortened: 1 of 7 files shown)


moveit_ros/moveit_servo/config/panda_simulated_config.yaml
```diff
@@ -33,26 +33,27 @@ smoothing_filter_plugin_name: "online_signal_smoothing::ButterworthFilterPlugin"
 # which other nodes can use as a source for information about the planning environment.
 # NOTE: If a different node in your system is responsible for the "primary" planning scene instance (e.g. the MoveGroup node),
 # then is_primary_planning_scene_monitor needs to be set to false.
+check_octomap_collisions: false
 is_primary_planning_scene_monitor: true
```
sea-bass, line 37:
```suggestion
is_primary_planning_scene_monitor: true
check_octomap_collisions: false  # Check collision against the octomap (if a 3D sensor plugin is available)
```

amalnanavati (approved):
I looked through all the changes and they look good -- thanks for making them! Requesting a humble backport please :)

sea-bass:
> I looked through all the changes and they look good -- thanks for making them! Requesting a humble backport please :)

Thanks! I unfortunately don't think Humble has the new MoveIt Servo refactor, so the backport may actually be completely different code in this case...

sjahr (approved):

PR #2628, opened 2023-12-21
Fix benchmarks

rhaschke:
I had a closer look at the benchmarks presented in #2546 today. I noticed that the increasing number of states was not allocated simultaneously, but sequentially (thus actually simply freeing and re-allocating the same memory blocks).
This PR changes this to true parallel allocation, which mimics the real behavior in a planner more closely. 
This _and_ the additional changes of #2617 yield a much stronger relative slow-down than what was shown in #2546:

benchmark                               | ROS1     | ROS2     | increase
----------------------------------------|----------|----------|--------
RobotStateBenchmark/construct/100       | 0.019 ms | 0.020 ms | +5 %
RobotStateBenchmark/construct/1000      | 1.20 ms  | 1.25 ms  | +4.1 %
RobotStateBenchmark/construct/10000     | 15.6 ms  | 16.0 ms  | +2.5 %
RobotStateBenchmark/copyConstruct/100   | 0.018 ms | 0.020 ms | +11.1 %
RobotStateBenchmark/copyConstruct/1000  | 1.18 ms  | 0.197 ms |
RobotStateBenchmark/copyConstruct/10000 | 15.2 ms  | 16.3 ms  | +7.2 %
RobotStateBenchmark/update/10           | 0.007 ms | 0.008 ms | +14.2 %
RobotStateBenchmark/update/100          | 0.068 ms | 0.086 ms | +26.4 %
RobotStateBenchmark/update/1000         | 0.710 ms | 0.885 ms | +24.6 %
RobotStateBenchmark/update/10000        | 10.4 ms  | 11.7 ms  | +12.5 %

Not only allocations are affected, but primarily also state updates, which occur much more frequently.
There is also a noticeable non-linearity when scaling up the amount of parallel states, which is probably due to more frequent cache misses.


moveit_core/robot_state/test/robot_state_benchmark.cpp
```diff
@@ -66,312 +61,279 @@ Eigen::Isometry3d createTestIsometry()
...
+  void SetUp(const ::benchmark::State& /*state*/) override
   {
-    st.SkipWithError("The planning group doesn't exist.");
-    return;
-  }
-
-  for (auto _ : st)
-  {
-    for (int i = 0; i < n_states; i++)
+    if (rcutils_logging_set_logger_level("moveit_robot_model.robot_model", RCUTILS_LOG_SEVERITY_WARN) != RCUTILS_RET_OK)
```
marioprats, line 174: Looks like there's something missing inside this block?

moveit_core/robot_state/test/robot_state_benchmark.cpp
```diff
@@ -66,312 +61,279 @@ Eigen::Isometry3d createTestIsometry()
...
+    std::vector<moveit::core::RobotState> states;
+    states.reserve(num);
+    for (size_t i = 0; i < num; i++)
+      states.emplace_back(state);
+    return states;
+  }
 
-  // Make sure the group exists, otherwise exit early with an error.
-  if (!robot_model->hasJointModelGroup(PANDA_TEST_GROUP))
+  std::vector<size_t> randomPermudation(size_t num)
```
marioprats, line 198:
```suggestion
  std::vector<size_t> randomPermutation(size_t num)
```

moveit_core/robot_state/test/robot_state_benchmark.cpp
```diff
@@ -66,312 +61,279 @@ Eigen::Isometry3d createTestIsometry()
...
   {
-    st.SkipWithError("The planning group doesn't exist.");
-    return;
+    std::vector<size_t> result;
+    result.reserve(num);
+    for (size_t i = 0; i < num; i++)
+      result.push_back(i);
```
marioprats, line 203: I believe we can do this?

```
std::vector<size_t> result(num);
std::iota(result.begin(), result.end(), 0);
```

moveit_core/robot_state/test/robot_state_benchmark.cpp
```diff
@@ -66,312 +61,279 @@ Eigen::Isometry3d createTestIsometry()
...
-    }
+    auto states = constructStates(st.range(0), state);
+    benchmark::DoNotOptimize(states);
+    benchmark::ClobberMemory();
   }
 }
 
-// Benchmark time to call `setToRandomPositions` and `getGlobalLinkTransform` on a RobotState.
-static void robotStateForwardKinematics(benchmark::State& st)
+// Benchmark time to call `setToRandomPositions` and `update` on RobotState.
```
marioprats, line 241: Should we keep the `getGlobalLinkTransform` version, in addition to `update`?
`update` calls `getGlobalLinkTransform` + `updateCollisionBodyTransforms`, so it does something extra.
Kinematics solvers for instance don't always need the collision body transforms, and use `getGlobalLinkTransform` instead, which should be slightly faster.

moveit_core/robot_state/test/robot_state_benchmark.cpp
```diff
@@ -66,312 +61,279 @@ Eigen::Isometry3d createTestIsometry()
...
+  std::vector<size_t> randomPermudation(size_t num)
   {
-    st.SkipWithError("The planning group doesn't exist.");
-    return;
+    std::vector<size_t> result;
+    result.reserve(num);
+    for (size_t i = 0; i < num; i++)
+      result.push_back(i);
+
+    std::random_device random_device;
```
marioprats, line 205: What do you think about using a deterministic random sequence instead?
Not sure if it will change anything, but at least it would create more deterministic conditions.

moveit_core/robot_state/test/robot_state_benchmark.cpp
```diff
@@ -66,312 +61,279 @@ Eigen::Isometry3d createTestIsometry()
...
   }
 
-  KDL::Chain kdl_chain;
-  if (!kdl_tree.getChain(jmg->getJointModels().front()->getParentLinkModel()->getName(),
-                         jmg->getLinkModelNames().back(), kdl_chain))
-  {
-    st.SkipWithError("Can't create KDL Chain.");
-    return;
-  }
+  KDL::TreeJntToJacSolver jacobian_solver(kdl_tree);
```
marioprats, line 298: Curious on the difference between `TreeJntToJacSolver` and `ChainJntToJacSolver`?
Is it a different implementation, or actually calling the same implementation underneath?

moveit_core/robot_state/test/robot_state_benchmark.cpp
```diff
@@ -66,312 +61,279 @@ Eigen::Isometry3d createTestIsometry()
...
-
-BENCHMARK(inverseIsometry3d)->Arg(MATRIX_OPS_N_ITERATIONS)->Unit(benchmark::kMillisecond);
-BENCHMARK(inverseAffineIsometry)->Arg(MATRIX_OPS_N_ITERATIONS)->Unit(benchmark::kMillisecond);
-BENCHMARK(inverseAffine)->Arg(MATRIX_OPS_N_ITERATIONS)->Unit(benchmark::kMillisecond);
-BENCHMARK(inverseMatrix4d)->Arg(MATRIX_OPS_N_ITERATIONS)->Unit(benchmark::kMillisecond);
-
-BENCHMARK(robotStateConstruct)->RangeMultiplier(10)->Range(100, 10000)->Unit(benchmark::kMillisecond);
-BENCHMARK(robotStateCopy)->RangeMultiplier(10)->Range(100, 10000)->Unit(benchmark::kMillisecond);
-BENCHMARK(robotStateUpdate)->RangeMultiplier(10)->Range(10, 1000)->Unit(benchmark::kMillisecond);
-BENCHMARK(robotStateForwardKinematics)->RangeMultiplier(10)->Range(10, 1000)->Unit(benchmark::kMillisecond);
```
marioprats, line 374: See my other comment. I think it still makes sense to keep this one, but maybe rename to `robotStateGetGlobalLinkTransform`, since that's what it calls?

marioprats (commented):
Hey Robert, thanks for making the benchmark more realistic, and cleaning it up, this is a great change.
I don't think it changes much about performance. We reported a slow down of 4x in https://github.com/ros-planning/moveit2/pull/2546, which is worse than what we see here. I understand it's a few more ms in the end, but we think it's a worth price to pay for additional safety. We also think we can recover these loses in many other ways, like your own https://github.com/ros-planning/moveit/pull/3548, or what we did with Jacobians in https://github.com/ros-planning/moveit2/pull/2389, and many other improvements we can do. We should be able to have both performant and bullet-proof core types. But to reach that point we need good benchmarks and tests, so it's great you made this benchmark better and already using it to drive improvements. Thank you!

moveit_core/robot_state/test/robot_state_benchmark.cpp
```diff
@@ -66,312 +61,279 @@ Eigen::Isometry3d createTestIsometry()
...
+  Eigen::Affine3d result;
   for (auto _ : st)
   {
-    for (int i = 0; i < n_iters; ++i)
-    {
-      Eigen::Affine3d result;
-      benchmark::DoNotOptimize(result = isometry.affine() * isometry.matrix());
-      benchmark::ClobberMemory();
-    }
+    benchmark::DoNotOptimize(result.affine().noalias() = isometry.affine() * isometry.matrix());
```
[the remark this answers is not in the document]
rhaschke: In my old code, multiple iterations were required for some statistical significance. However, the Google benchmark lib automatically iterates a to-be-timed code snipped as long as needed to reach statistical significance. Hence, it's not needed to be done manually anymore.

moveit_core/robot_state/test/robot_state_benchmark.cpp
```diff
@@ -66,312 +61,279 @@ Eigen::Isometry3d createTestIsometry()
...
-BENCHMARK(robotStateUpdate)->RangeMultiplier(10)->Range(10, 1000)->Unit(benchmark::kMillisecond);
-BENCHMARK(robotStateForwardKinematics)->RangeMultiplier(10)->Range(10, 1000)->Unit(benchmark::kMillisecond);
-
-BENCHMARK(moveItJacobian);
-BENCHMARK(kdlJacobian);
+BENCHMARK(multiplyAffineTimesMatrixNoAlias);
+BENCHMARK(multiplyMatrixTimesMatrixNoAlias);
+BENCHMARK(multiplyIsometryTimesIsometryNoAlias);
+BENCHMARK(multiplyMatrixTimesMatrix);
+BENCHMARK(multiplyIsometryTimesIsometry);
```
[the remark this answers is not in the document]
rhaschke: I just tested different variants. It's quite obvious that NoAlias versions are faster.

moveit_core/robot_state/test/robot_state_benchmark.cpp
```diff
@@ -38,21 +38,16 @@
...
+#include <kdl/treejnttojacsolver.hpp>
 #include <moveit/robot_model/robot_model.h>
 #include <moveit/robot_state/robot_state.h>
 #include <moveit/utils/robot_model_test_utils.h>
-#include <random_numbers/random_numbers.h>
 
 // Robot and planning group for benchmarks.
 constexpr char PANDA_TEST_ROBOT[] = "panda";
 constexpr char PANDA_TEST_GROUP[] = "panda_arm";
-constexpr char PR2_TEST_ROBOT[] = "pr2";
```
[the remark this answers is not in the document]
rhaschke: Feel free to change. From my point of view it doesn't matter whether calling `update()` more often or having a longer-running `update()` call.

moveit_core/robot_state/test/robot_state_benchmark.cpp
```diff
@@ -66,312 +61,279 @@ Eigen::Isometry3d createTestIsometry()
 }  // namespace
 
 // Benchmark time to multiply an Eigen::Affine3d with an Eigen::Matrix4d.
-static void multiplyAffineTimesMatrix(benchmark::State& st)
+static void multiplyAffineTimesMatrixNoAlias(benchmark::State& st)
```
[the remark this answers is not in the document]
rhaschke: The NoAlias versions just use Eigen's `.noalias()` modifier, allowing to write the result of matrix multiplication directly into the result matrix instead of using an intermediate temporary (which is the default).

moveit_core/robot_state/test/robot_state_benchmark.cpp
```diff
@@ -66,312 +61,279 @@ Eigen::Isometry3d createTestIsometry()
...
 static void inverseMatrix4d(benchmark::State& st)
 {
-  int n_iters = st.range(0);
-  Eigen::Isometry3d isometry = createTestIsometry();
-  Eigen::Affine3d affine;
-  affine.matrix() = isometry.matrix();
-
+  Eigen::Affine3d affine = createTestIsometry();
+  Eigen::Affine3d result;
   for (auto _ : st)
```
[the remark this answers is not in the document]
rhaschke: Yes, I think that there is some compiler optimization going on. However, I didn't manage to fix that using various variants of `DoNotOptimize`. Finally I had to move on to other stuff...

moveit_core/robot_state/test/robot_state_benchmark.cpp
rhaschke: No. The if condition was just required to silent a compiler warning about unused function results. I don't care about the result.

moveit_core/robot_state/test/robot_state_benchmark.cpp
rhaschke: Thanks for pointing this out. I couldn't find the corresponding function when looking for it...

moveit_core/robot_state/test/robot_state_benchmark.cpp
rhaschke: I consider `update()` simply a real-world test for FK updates: we just want to run many transform updates - no matter being link or collision transforms.

moveit_core/robot_state/test/robot_state_benchmark.cpp
rhaschke: These are two different classes implementing the same code essentially. The central difference is the size of the Jacobian: tree considers columns for all joints in the tree, chain only those in the chain.

moveit_core/robot_state/test/robot_state_benchmark.cpp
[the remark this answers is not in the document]
marioprats: It also includes different joint types, and a more complex tree structure in general, so I think it makes sense to keep the PR2 test. But I'm happy to change this in a follow-up PR.

moveit_core/robot_state/test/robot_state_benchmark.cpp
marioprats: I see, then let's use `std::ignore` or a cast to `void` to express intent.
```suggestion
    std::ignore = rcutils_logging_set_logger_level("moveit_robot_model.robot_model", RCUTILS_LOG_SEVERITY_WARN);
```

moveit_core/robot_state/test/robot_state_benchmark.cpp
marioprats: That makes sense, but we have code that doesn't care about collision-body transforms, like IK solvers, or Servo (the IK part). They just call `getGlobalLinkTransform`, which updates only the link transforms. I would like to keep that version so that we have a direct measure of how much time that's taking.

moveit_core/robot_state/test/robot_state_benchmark.cpp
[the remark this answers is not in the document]
marioprats: Good to know about the `.noalias()`. Unfortunately the default behavior is much more frequent in code, so it would be great to keep both versions of the three multiply benchmarks.

marioprats (changes requested):

moveit_core/robot_state/test/robot_state_benchmark.cpp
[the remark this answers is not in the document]
rhaschke: The internal loop doesn't contribute anything when using Google's benchmark lib,
which does exactly that: performing the same operation in a loop for good statistics.
I can imagine that affine inverse is slower, because it cannot optimize memory allocation (needs to skip entries of last, fixed matrix row).

moveit_core/robot_state/test/robot_state_benchmark.cpp
rhaschke: I think that doesn't make a difference.

rhaschke:
Is there anything more to do for this to get merged, @marioprats?

marioprats (approved):

henningkayser (approved):

Issue #2554, opened 2023-11-27
last waypoint from STOMP solution doesn't always match goal


patrickKXMD:
### Description
this is a really annoying problem, the last waypoint from STOMP solution  doesn't always match the `goal_state`

I added four lines here for debug
https://github.com/ros-planning/moveit2/blob/c2292a7054904c25851b6af78abb6aca8a5852ae/moveit_planners/stomp/src/stomp_moveit_planning_context.cpp#L68C1-L71C4
```cpp
  if (!input_trajectory || input_trajectory->empty())
  {
    success = stomp->solve(getPositions(start_state, joints), getPositions(goal_state, joints), waypoints);
    // new lines
    auto goal_positions = get_positions(goal_state, joints);
    auto last_waypoint = waypoints.col(waypoints.cols() - 1);
    RCLCPP_WARN(LOGGER, "TEST Goal position: %.12f, %.12f, %.12f, %.12f, %.12f, %.12f", goal_positions[0], goal_positions[1], goal_positions[2], goal_positions[3], goal_positions[4], goal_positions[5]);
    RCLCPP_WARN(LOGGER, "TEST Last waypoint: %.12f, %.12f, %.12f, %.12f, %.12f, %.12f", last_waypoint[0], last_waypoint[1], last_waypoint[2], last_waypoint[3], last_waypoint[4], last_waypoint[5]);
  }
```

and here is the output

sometimes they are identical which is good.
>[stomp_moveit]: TEST Goal position: 0.234587246836, 0.992731932008, -2.384229104531, -0.235384185478, -2.907002753126, -0.000018023283
[stomp_moveit]: TEST Last waypoint: 0.234587246836, 0.992731932008, -2.384229104531, -0.235384185478, -2.907002753126, -0.000018023283
[stomp_moveit]: TEST Goal position: 1.560993985891, 0.992730570867, -2.384225161300, -0.235366751959, -1.580596014100, -0.000002689587
[stomp_moveit]: TEST Last waypoint: 1.560993985891, 0.992730570867, -2.384225161300, -0.235366751959, -1.580596014100, -0.000002689587

sometimes they are different
>[stomp_moveit]: TEST Goal position: 0.975131363602, 0.760374503146, -0.912535765952, 1.468677947059, -2.166458636384, -0.000005143243
[stomp_moveit]: TEST Last waypoint: 0.975118794469, 0.760292204555, -0.912173806359, 1.469008654541, -2.166498134031, 0.000181114966
[...]

PR #2631, opened 2023-12-22
Fix penalty calculation in constraints cost function in STOMP

patrickKXMD:
### Description

@henningkayser 
I mentioned this issue here: https://github.com/ros-planning/moveit2/issues/2554#issuecomment-1866178090
I can't plan in STOMP with joint constrants.

I think the penalty result for constraints cost function should be `0.0` when `ConstraintEvaluationResult.satisfied` is `true`

### Checklist

### DIFF
--- a/moveit_planners/stomp/include/stomp_moveit/cost_functions.hpp
+++ b/moveit_planners/stomp/include/stomp_moveit/cost_functions.hpp
@@ -212,7 +212,8 @@ CostFn getConstraintsCostFunction(const std::shared_ptr<const planning_scene::Pl
     setJointPositions(positions, joints, state);
     state.update();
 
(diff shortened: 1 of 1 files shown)


henningkayser (commented):
I am a little bit conflicted about this change. I agree, that states that satisfy the constraints should not render states or trajectories invalid. At the same time, it would be great if we could still continue optimizing for constraints within the space. This change would probably result in many solutions that just barely satisfy the constraints since they stop optimizing at the constraint boundaries. Considering, that the current cost function categorizes non-zero costs as invalid, I'll approve this change. But at the same time I would be very interested in follow-ups that would support explicit cost thresholding to catch this case instead.
Can you add a NOTE referencing this issue? https://github.com/ros-planning/moveit2/issues/2658

PR #2635, opened 2023-12-23
[Humble] Remove ros2_control 2.13 source build

sea-bass:
Turns out we were cloning a pretty old version of `ros2_control` here -- 2.13 when it's now at 2.36.

Maybe this will be the thing to fix the `humble_ci` job on `main` which uses the generated Docker image from this branch since https://github.com/ros-planning/moveit2/pull/2284?

### DIFF
--- a/moveit2.repos
+++ b/moveit2.repos
@@ -9,7 +9,3 @@ repositories:
     type: git
     url: https://github.com/ros-planning/moveit_resources.git
     version: humble
-  ros2_control:
-    type: git
-    url: https://github.com/ros-controls/ros2_control.git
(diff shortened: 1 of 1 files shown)


Abishalini (approved):

sea-bass:
I think the source CI is failing because it's pulling from the `ci-testing` image before this PR lands... so it's missing some `ros2_control` stuff.

And ikfast just, has been failing.

Can we force merge this?

sjahr:
@tylerjw @henningkayser Can you force merge this?

henningkayser (approved):

PR #2653, opened 2024-01-18
Attempt to use SCHED_FIFO for Servo regardless of RT kernel

stephanie-eng:
### Description

This PR changes Servo to always try to use SCHED_FIFO, regardless of presence of a RT kernel. SCHED_FIFO can be used on a non-realtime kernel, and using it will boost the priority above any SCHED_OTHER tasks. While the non-real-time kernel isn't optimized for low jitter, setting SCHED_FIFO will still help to minimize scheduling latency. 

To use SCHED_FIFO, it is still required for users have the proper permissions to allow setting of real-time priorities.

To show the difference this change makes on a non-RT kernel, I ran some tests on my laptop:
These are histograms of scheduling latencies I measured running Servo at 100 Hz (10 ms deadline) for 100 seconds (for a total of 10 000 scheduling latency measurements in the histogram)

Wakeup Latency Using SCHED_OTHER (non-RT kernel)

Wakeup Latency Using SCHED_FIFO (non-RT kernel)

In the top histogram, you can see the max wakeup latency in a short 100 second test was about 8.5 ms, which is almost a whole Servo period! A longer duration test may reveal complete deadline misses. However, when using SCHED_FIFO, the scheduling latency is much better controlled.


moveit_ros/moveit_servo/src/servo_node.cpp
```diff
@@ -75,19 +75,18 @@ ServoNode::ServoNode(const rclcpp::NodeOptions& options)
...
-    }
-    else
-    {
-      RCLCPP_WARN_STREAM(node_->get_logger(), "Could not enable FIFO RT scheduling policy.");
-    }
+    RCLCPP_INFO_STREAM(node_->get_logger(), "Enabled SCHED_FIO and higher thread priority.");
   }
   else
+  {
+    RCLCPP_WARN_STREAM(node_->get_logger(), "Could not enable FIFO RT scheduling policy.");
```
AndyZe, line 85: Maybe make this warning sound more friendly since it's not a big deal

```suggestion
    RCLCPP_WARN_STREAM(node_->get_logger(), "Could not enable FIFO RT scheduling policy. Continuing with the default.");
```

sjahr (changes requested):
I like this change but how is the scheduler changed back for a user after they ran a `servo node`? As a user I wouldn't expect a ROS node to automatically mess with my system settings. Is this call reverted by a reboot or at shutdown of the node?

stephanie-eng:
> I like this change but how is the scheduler changed back for a user after they ran a `servo node`? As a user I wouldn't expect a ROS node to automatically mess with my system settings. Is this call reverted by a reboot or at shutdown of the node?

This isn't really what I'd call a system setting. The scheduler is not set for the user, but on a process. In fact, you probably have some processes running with the FIFO scheduler on your computer now:

```ps -e -o pid,class,rtprio,comm ```

I see a line like this (amongst many other processes, I had to scroll)
```   1230 FF      50 irq/200-AudioDS```
The first column is PID, second is scheduler (FF = FIFO), third is priority (50 in my case), and last is the command. It looks like this is related to audio, and should indeed use a real-time scheduler or else we'd risk hearing cracks in our audio.

This change sets the scheduler for the [`servo_loop_` thread](https://github.com/ros-planning/moveit2/blob/cba873a899e6057c48e4ee07c49c4327b24c7a7d/moveit_ros/moveit_servo/src/servo_node.cpp#L148). When this [thread is joined](https://github.com/ros-planning/moveit2/blob/cba873a899e6057c48e4ee07c49c4327b24c7a7d/moveit_ros/moveit_servo/src/servo_node.cpp#L57), this change is "undone" insofar as there will be no more FIFO-scheduled servo thread (or any servo thread at all), so there are no "lasting" effects on the user.

TL;DR: the scheduler is set on the process level, not the user level, so the ROS node isn't messing with your system settings, it is messing with its own process settings. There's nothing to "revert" by rebooting.

sea-bass (approved):

AndyZe (approved):

sjahr (approved):
Awesome explanation, thank you! I did not know that

PR #2654, opened 2024-01-22
avoid a relative jump threshold of 0.0

marioprats:
The Cartesian path capability request can have a `jump_threshold` of zero. A 'relative' jump of 0.0 is not allowed. Therefore we need to catch that case and disable the jump threshold if needed.
This became an issue after the changes to the JumpThreshold API.

Fixes https://github.com/ros-planning/moveit2_tutorials/issues/864


sjahr (approved):
Makes sense, thanks for fixing this!

Issue #2325, opened 2023-08-23
getLinkModelNamesWithCollisionGeometry() does not include base_link of robot


MarqRazz:
### Description

I am working on a robot that includes a linear actuator at the base of a UR5. When testing a pick and place MTC task I noticed that I was getting plans that included collision between the `base_link` of the linear actuator group and my attached collision object.
![collision_issue](https://user-images.githubusercontent.com/25058794/226439198-498defc0-f9fe-4ea1-9997-b338800f123f.gif)

In my robots srdf file I have the following groups included to allow motion planning with the arm + actuator at the same time:
```
    <group name="linear_actuator">
        <chain base_link="linear_actuator_base_link" tip_link="linear_actuator_carriage_link"/>
    </group>
    <group name="manipulator">
        <chain base_link="base_link" tip_link="tool0"/>
    </group>
    <group name="linear_actuator_and_manipulator">
        <group name="linear_actuator" />
        <group name="manipulator" />        
    </group>
```

In my MTC task I am getting the collision links of the motion planning group to allow and exclude collision between different groups with the following function:
```
const auto& manipulator_group_collision_link_model_names = task
                                                    ->getRobotModel()
                                                    ->getJointModelGroup("linear_actuator_and_manipulator")
                                                    ->getLinkModelNamesWithCollisionGeometry();
```

But when I inspect this list the `linear_actuator_base_link` is not part of it.
```
Links with Collision Geometry: [linear_actuator_carriage_link, base_link_inertia, shoulder_link, upper_arm_link, forearm_link, wrist_1_link, wrist_2_link, wrist_3_link]
```

Debugging the construction of the group I noticed that all `base_link`s of a `group` are not included in the `JointModelGroup`. This kinda makes sense because the base is not a joint but the pair of links that make the first joint should include it's collision geometry.

PR #2660, opened 2024-01-24
Fix getLinkModelNamesWithCollisionGeometry to include the base link

MarqRazz:
### Description
I was working on an MTC task a while ago where I was manually modifying the planning scene. In my program I was calling  `getLinkModelNamesWithCollisionGeometry()` on the `JointModelGroup` to get the links of the manipulator that had collision geometry and noticed that the base link of the group is never included. This caused planning problems like this while I was debugging...
collision_issue

To test this change out I modified the moveit2_tutorial `Robot Model and Robot State` where I asked it to print the results of `getLinkModelNamesWithCollisionGeometry` and with this new change it now includes `panda_link0` :1st_place_medal: 
```
RCLCPP_INFO(LOGGER, "Joint model group `panda_arm` collision links:");
std::vector<std::string> collision_link_model_names = joint_model_group->getLinkModelNamesWithCollisionGeometry();
for (const auto& name : collision_link_model_names)
{
    RCLCPP_INFO(LOGGER, "Link with collision: %s", name.c_str());
}

Output:
Joint model group `panda_arm` collision links:
  Link with collision: panda_link0
  Link with collision: panda_link1
  Link with collision: panda_link2
  Link with collision: panda_link3
  Link with collision: panda_link4
  Link with collision: panda_link5
  Link with collision: panda_link6
  Link with collision: panda_link7
```

Very small nitpick but I have also noticed that in Rviz the base link is not highlighted for the goal pose which shows that it is not include in the group when using the motion planning widget.
image

### Checklist

### DIFF
--- a/moveit_core/robot_model/src/joint_model_group.cpp
+++ b/moveit_core/robot_model/src/joint_model_group.cpp
@@ -229,6 +229,13 @@ JointModelGroup::JointModelGroup(const std::string& group_name, const srdf::Mode
   {
     link_model_map_[link_model->getName()] = link_model;
     link_model_name_vector_.push_back(link_model->getName());
+    // if this is the first link of the group with geometry (for example `base_link`) it should included
+    if(link_model_with_geometry_vector_.empty() && !link_model->getParentLinkModel()->getShapes().empty())
+    {        
+      link_model_with_geometry_vector_.push_back(link_model->getParentLinkModel());
+      link_model_with_geometry_name_vector_.push_back(link_model->getParentLinkModel()->getName());
+    }
+    // all child links with collision geometry should also be included
     if (!link_model->getShapes().empty())
     {
       link_model_with_geometry_vector_.push_back(link_model);
--- a/moveit_core/robot_model/src/joint_model_group.cpp
+++ b/moveit_core/robot_model/src/joint_model_group.cpp
@@ -230,8 +230,8 @@ JointModelGroup::JointModelGroup(const std::string& group_name, const srdf::Mode
     link_model_map_[link_model->getName()] = link_model;
     link_model_name_vector_.push_back(link_model->getName());
     // if this is the first link of the group with geometry (for example `base_link`) it should included
-    if(link_model_with_geometry_vector_.empty() && !link_model->getParentLinkModel()->getShapes().empty())
-    {        
+    if (link_model_with_geometry_vector_.empty() && !link_model->getParentLinkModel()->getShapes().empty())
+    {
       link_model_with_geometry_vector_.push_back(link_model->getParentLinkModel());
       link_model_with_geometry_name_vector_.push_back(link_model->getParentLinkModel()->getName());
     }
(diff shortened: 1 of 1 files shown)


MarqRazz:
Closes https://github.com/ros-planning/moveit2/issues/2325

sjahr (approved):
Makes sense to me, thanks for fixing it!

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -229,6 +229,14 @@ JointModelGroup::JointModelGroup(const std::string& group_name, const srdf::Mode
   {
     link_model_map_[link_model->getName()] = link_model;
     link_model_name_vector_.push_back(link_model->getName());
+    // if this is the first link of the group with a valid parent and includes geometry (for example `base_link`) it should included
+    if (link_model_with_geometry_vector_.empty() && link_model->getParentLinkModel() &&
```
simonschmeisser, line 233: I'm not sure this is the correct approach. There can be more than one discrete subtree in a group, this would only consider the first one.

We discussed a similar issue I observed in https://github.com/ros-planning/moveit/issues/3570

I think the previous version of the code is correct for `is_chain_ == false`  (a list of `<link> or <joint>` tags in SRDF) but not for the case of `<chain>` where the root link is included but not parent joint.

simonschmeisser:
Could you please add your SRDF?

moveit_core/robot_model/src/joint_model_group.cpp
sjahr: Good point, thanks for pointing that out! Do you think we should revert this change for now or just add a check for `is_chain_`?

simonschmeisser: Hmm, this is getting surprisingly complicated ... `is_chain_` is not passed to the constructor but *guessed* about near the end of the constructor. They might be different things altogether?

According to the code in `RobotModel` a chain as specified in SRDF could also go from `finger_1_tip` to `finger_2_tip` with the root being somewhere in the middle (eg the `wrist`).

Maybe a list of links should also get passed in to the constructor.

I'm also not yet sure how this relates to collision checking and what the problem actually was.

sjahr:
> Could you please add your SRDF?

Unfortunately, I don't have access to it

MarqRazz:
[Here](https://github.com/ros-planning/moveit_resources/blob/ros2/panda_moveit_config/config/panda_arm.xacro#L9-L11) is the SRDF for the Panda arm.

Prior to this PR if you were to request the link models with collision geometry 
`std::vector<std::string> collision_link_model_names = joint_model_group->getLinkModelNamesWithCollisionGeometry();`
it would not include the base frame `panda_link0` which you can see is part of the `panda_arm` joint model group. I was only trying to fix this function call to return a valid list. In my original use case with the robot on a rail (which was also build with <chain>) I was manually enabling and disabling collisions given the result from  `getLinkModelNamesWithCollisionGeometry` which is why I noticed the issue.

simonschmeisser:
Good to know that the collision checking failure was caused by workarounds. I tested a bit with a similar setup today (MoveIt1) but couldn't notice anything broken.

I think this fix adds regressions if:
* group is specified via <links>, then the parent should obviously be excluded
* group is specified via <joints>, then the convention as documented is to include only the child links
* group is chain but ^ shaped, then some link will be duplicated in the list

And does not fix
* more than one "vertical" chain is specified, then only one will be corrected

Most likely the list of links should be passed into the constructor and the special casing for <chain> be done in RobotModel

PR #2662, opened 2024-01-26
Add parameter api integration test

sjahr:
### Description

Unittest to check if all parameters are correctly loaded when a move_group is launched

### Checklist


moveit_ros/move_group/test/move_group_test_node.cpp
```diff
@@ -0,0 +1,102 @@
...
+  std::shared_ptr<rclcpp::Node> test_node_;
+
+  // Executor and a thread to run the executor.
+  rclcpp::Executor::SharedPtr executor_;
+  std::thread executor_thread_;
+};
+
+TEST_F(MoveGroupFixture, testParamAPI)
+{
+  rclcpp::sleep_for(std::chrono::seconds(5));
```
sjahr, line 79:
```suggestion
```

moveit_ros/move_group/test/move_group_test_node.cpp
```diff
@@ -0,0 +1,101 @@
...
+  // THEN it is a defined in the note
+  for (const auto& param_name : move_group_test::PARAMETER_NAME_LIST)
+  {
+    const auto param_exists = test_node_->has_parameter(param_name);
+    if (!param_exists)
+    {
+      RCLCPP_ERROR(test_node_->get_logger(), "Parameter %s doesn't exists", param_name.c_str());
+    }
+    EXPECT_TRUE(param_exists);
+    rclcpp::sleep_for(std::chrono::seconds(1));
```
sjahr, line 91: For some reason, the shorter this sleep is, the fewer parameters are found by `->has_parameter` :disappointed:

moveit_ros/move_group/CMakeLists.txt
```diff
@@ -103,7 +103,6 @@ pluginlib_export_plugin_description_file(moveit_ros_move_group default_capabilit
 
 if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
-
   # These don't pass yet, disable them for now
```
sjahr, line 106:
```suggestion

  # These don't pass yet, disable them for now
```

moveit_ros/planning_interface/test/move_group_api_test.cpp
```diff
@@ -0,0 +1,105 @@
...
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Sebastian Jahr
+   Description: Integrations tests for the move_group interface
```
henningkayser, line 36:
```suggestion
   Description: Integration tests for the move_group interface
```
Can you describe the type of integration tests and motivation for them here? Are we really adding more?

moveit_ros/planning_interface/test/launch/move_group_api.test.py
```diff
@@ -0,0 +1,139 @@
...
+        output="screen",
+    )
+
+    joint_state_broadcaster_spawner = launch_ros.actions.Node(
+        package="controller_manager",
+        executable="spawner",
+        arguments=[
+            "joint_state_broadcaster",
+            "--controller-manager-timeout",
+            "300",
```
henningkayser, line 58: Is this seconds? Seems pretty long if it is

moveit_ros/planning_interface/package.xml
```diff
@@ -43,6 +43,9 @@
   <test_depend>moveit_resources_panda_moveit_config</test_depend>
   <test_depend>moveit_simple_controller_manager</test_depend>
   <test_depend>moveit_planners_ompl</test_depend>
+  <test_depend>moveit_planners_chomp</test_depend>
+  <test_depend>moveit_planners_stomp</test_depend>
+  <test_depend>pilz_industrial_motion_planner</test_depend>
```
henningkayser, line 48: so this adds a circular test dependency, even though the planning interface is not listed here. I think we should have a higher-level package for running integration tests for that purpose.

moveit_ros/planning_interface/test/launch/move_group_api.test.py
sjahr: Good point, let's see if it works without it

sjahr, line 58:
```suggestion
```

moveit_ros/planning_interface/package.xml
sjahr: I moved the test into its own package. At the moment it is part of moveit_ros but we can move it somewhere else too

moveit_ros/tests/src/move_group_api_test.cpp
```diff
@@ -0,0 +1,105 @@
...
+  }
+  ASSERT_TRUE(reach_param_client);
+
+  // GIVEN a running move_group
+  // WHEN a parameter name from parameter API checked
+  // THEN that parameter exists
+  for (const auto& param_name : move_group_test::PARAMETER_NAME_LIST)
+  {
+    bool param_exists = false;
+    if (params_client->wait_for_service(std::chrono::seconds(1)))
```
tylerjw, line 79: Is this going to cause it to wait 1 second for each parameter in the case where this fails? That will make this test take forever. You already have this outside the for loop so maybe consider dropping it here and assume that if we've connected before to the parameter client we are connected here too.

sjahr: I've experienced issues when I did not use this. My guess is that the other node couldn't handle getting so many service requests in little time. This fixes this as we're waiting for the service node to be ready to handle a request once it has answered another one.

sjahr, line 79:
```suggestion
    if (params_client->wait_for_service(std::chrono::seconds(0.1)))
```

sjahr: Let's try if the test still works if we drastically decrease the time

moveit_ros/tests/src/move_group_api_test.cpp
```diff
@@ -0,0 +1,105 @@
...
+  }
+  ASSERT_TRUE(reach_param_client);
+
+  // GIVEN a running move_group
+  // WHEN a parameter name from parameter API checked
+  // THEN that parameter exists
+  for (const auto& param_name : move_group_test::PARAMETER_NAME_LIST)
+  {
+    bool param_exists = false;
+    if (params_client->wait_for_service(std::chrono::seconds(0.1)))
```
sjahr, line 79:
```suggestion
    if (params_client->wait_for_service(std::chrono::milliseconds(1)))
```

moveit_ros/tests/src/move_group_api_test.cpp
tylerjw: Looking at the API it of the parameter thing it looks like we could ask for multiple parameters at once. Maybe it makes sense to just ask for all of them in one shot?

PR #2669, opened 2024-01-30
Fix bug in GetUrdfService move_group capability 

sjahr:
### Description

The capability did not take into account that it is possible to have link elements that close with "/>". This PR fixes that

### Checklist


EzraBrooks:
Reading and writing an XML document with substringing and string concatenation seems very fragile. Using something like TinyXML2 would avoid parsing mistakes like this in the future.

sjahr:
> Reading and writing an XML document with substringing and string concatenation seems very fragile. Using something like TinyXML2 would avoid parsing mistakes like this in the future.

I tried tinyxml2 first but I had a really hard time using it (very little documentation and bad error messaging) and it did not succeed in the end. If we allocate more time for this, I could give it another try but for now I think we should fix this.

moveit_ros/move_group/src/default_capabilities/get_group_urdf_capability.cpp
```diff
@@ -99,32 +100,56 @@ void GetUrdfService::initialize()
...
         for (const auto& link_name : link_names)
         {
-          const auto start = full_urdf_string.find("<link name=\"" + link_name);
+          const auto start = full_urdf_string.find("<link name=\"" + link_name + "\"");
           auto substring = full_urdf_string.substr(start, full_urdf_string.size() - start);
-          res->urdf_string += substring.substr(0, substring.find(LINK_ELEMENT_CLOSING) + LINK_ELEMENT_CLOSING.size());
+
+          // Link elements can be closed either by "/>" or "</link>" so we need to consider both cases
+          auto const substring_without_opening = substring.substr(1, substring.size() - 2);
+          auto const general_opening_pos_a = substring_without_opening.find("<");
```
sjahr, line 115:
```suggestion
          auto const general_opening_pos_a = substring_without_opening.find('<');
```

henningkayser:
Isn't there a way to implement the same thing just using urdfdom? It seems very odd to me that we have to run manual xml string parsing and editing.

sjahr:
> urdfdom

You can export the parsed model to a TiXmlDoment type (tinyxml format, not tinyxml2!) but not to a string :/

marioprats:
This PR LGTM since it's fixing a bug in code that is already there.
Maybe merge this in but open an issue to improve it using `urdfdom` or existing xml parsing libraries, for when we get to it?

tylerjw (approved):

PR #2673, opened 2024-01-31
Add servo solver

sjahr:
### Description

Ports over the servo-based local planner from the [UR welding demo](https://github.com/PickNikRobotics/UR10e_welding_demo)
### Checklist


moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/include/moveit/local_constraint_solver_plugins/servo_solver.hpp
```diff
@@ -0,0 +1,102 @@
...
+{
+/**
+ * @brief Local solver plugin that utilizes moveit_servo as a local planner
+ *
+ */
+class ServoSolver : public moveit::hybrid_planning::LocalConstraintSolverInterface
+{
+public:
+  /**
+   * @brief Initializes moveit servo
```
sjahr, line 58: > Maybe you could add a small description to node. For example, I am unsure if the node needs to be spinning to be used, but a comment such as "node is used to run servo's ROS callbacks". Maybe that would be more clear.

moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/servo_solver.cpp
```diff
@@ -0,0 +1,174 @@
...
+  // TF planning_frame -> current EE
+  Eigen::Isometry3d current_pose = current_state->getFrameTransform(solver_parameters_.reference_frame);
+  // TF planning -> target EE
+  Eigen::Isometry3d target_pose = target_state.getFrameTransform(solver_parameters_.reference_frame);
+
+  // current EE -> planning frame * planning frame -> target EE
+  Eigen::Isometry3d diff_pose = current_pose.inverse() * target_pose;
+  Eigen::AngleAxisd axis_angle(diff_pose.linear());
+
+  constexpr double fixed_trans_vel = 0.05;
```
sjahr, line 123: > Do you think it would make sense to make these into parameters?

moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/servo_solver.cpp
```diff
@@ -0,0 +1,174 @@
...
+      diff_pose.translation().z() * trans_gain, axis_angle.axis().x() * axis_angle.angle() * rot_gain,
+      axis_angle.axis().y() * axis_angle.angle() * rot_gain, axis_angle.axis().z() * axis_angle.angle() * rot_gain }
+  };
+
+  std::optional<trajectory_msgs::msg::JointTrajectory> trajectory_msg;
+  // Create servo commands until a trajectory message can be generated
+  while (!trajectory_msg)
+  {
+    // Calculate next servo command
+    moveit_servo::KinematicState joint_state = servo_->getNextJointState(current_state, target_twist);
```
sjahr, line 144: > This point may cause issues if not handled with caution. Servo has a parameter called publish_period which is the time difference used for numerical integration. So if you ask servo to move at a velocity of 1 m/s and publish_period was .1 sec, then servo will return a new position that correspondence to a .1 m offset from the current position. So you should consider setting a faction of local trajectory length. For example, if you are supposed to arrive at your next point in .3 sec, then the update period should be .3/n, where n is the number of points you want in the command queue. I would recommend n=3.

moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/servo_solver.cpp
```diff
@@ -0,0 +1,174 @@
...
+  local_trajectory.getRobotTrajectoryMsg(robot_command);
+
+  if (robot_command.joint_trajectory.points.empty())
+  {
+    feedback_result.feedback = std::string("Reference trajectory does not contain any points");
+    return feedback_result;
+  }
+
+  // Get current state
+  const auto current_state = planning_scene_monitor_->getStateMonitor()->getCurrentState();
```
sjahr, line 106: > If you are certain that the current state is actually current, then I would add it into the command queue as the first state.

moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/servo_solver.cpp
```diff
@@ -0,0 +1,174 @@
...
+  moveit_servo::TwistCommand target_twist{
+    solver_parameters_.reference_frame,
+    { diff_pose.translation().x() * trans_gain, diff_pose.translation().y() * trans_gain,
+      diff_pose.translation().z() * trans_gain, axis_angle.axis().x() * axis_angle.angle() * rot_gain,
+      axis_angle.axis().y() * axis_angle.angle() * rot_gain, axis_angle.axis().z() * axis_angle.angle() * rot_gain }
+  };
+
+  std::optional<trajectory_msgs::msg::JointTrajectory> trajectory_msg;
+  // Create servo commands until a trajectory message can be generated
+  while (!trajectory_msg)
```
sjahr, line 141: > Before beginning the loop, delete any values in the command queue that are in than 1 update period in the past. So if the local planner is called at 100 Hz, then keep values as long as they are not older than 0.01 seconds.

sjahr (commented):
Porting over @pac48's review from https://github.com/sjahr/moveit2/pull/16

> Overall looks good! I added a few comments in the main local planner solve function on how to possibly use servo's command queue more effectively. One other thing, it may be too much work to set up, but this draft PR to ros2_control shows how to add data_tamer to log the values sent to the hardware interface https://github.com/ros-controls/ros2_control/pull/1255. If the command queue is setup correctly, you should see very smooth motion.

moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/forward_trajectory.cpp
```diff
@@ -36,6 +36,7 @@
 #include <moveit/local_planner/feedback_types.h>
 #include <moveit/planning_scene/planning_scene.h>
 #include <moveit/robot_state/conversions.h>
+#include <forward_trajectory_parameters.hpp>
```
henningkayser, line 39: Is it time to do a repo-wide refactoring for .hpp file type extensions?

moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/servo_solver.cpp
```diff
@@ -0,0 +1,174 @@
...
+  const std::string param_namespace = "moveit_servo";
+  const std::shared_ptr<const servo::ParamListener> servo_param_listener =
+      std::make_shared<const servo::ParamListener>(node, param_namespace);
+  servo_parameters_ = servo_param_listener->get_params();
+
+  // Create Servo and start it
+  servo_ = std::make_unique<moveit_servo::Servo>(node_, servo_param_listener, planning_scene_monitor_);
+
+  // Use for debugging
+  // twist_cmd_pub_ = node_->create_publisher<geometry_msgs::msg::TwistStamped>("~/delta_twist_cmds", 10);
```
henningkayser, line 76: Could you either remove this or make it a debug option?

moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/servo_solver.cpp
```diff
@@ -0,0 +1,174 @@
...
+  {
+    feedback_result.feedback = std::string("Reference trajectory does not contain any points");
+    return feedback_result;
+  }
+
+  // Get current state
+  const auto current_state = planning_scene_monitor_->getStateMonitor()->getCurrentState();
+
+  // Create goal state
+  moveit::core::RobotState target_state(local_trajectory.getRobotModel());
```
henningkayser, line 109: Better set to default values or copy the current state

moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/servo_solver.cpp
sjahr: I think it might not be the current state anymore since the robot is moving while local planning takes place. What do you think is the tolerance there?

moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/forward_trajectory.cpp
sjahr: Yes!

moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/servo_solver.cpp
sjahr: I am setting the (hopefully) correct publish_period in the initialize() function (see [diff](https://github.com/ros-planning/moveit2/pull/2673/commits/96160dab8d68bd565af4ceb19cbfc54e3b4f51cc)). Do you mind validating this?

pac48: It looks correct. The only issue I have with it now is that `getNextJointState` keeps getting called until `trajectory_msg` is a value. if for some reason  `moveit_servo::composeTrajectoryMessage` is changed to only return once 4 values are in the trajectory, then this would half break and be hard to detect. All that to say, I think you should add a check on the trajectory message to make sure it has length 3.

moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/servo_solver.cpp
```diff
@@ -0,0 +1,200 @@
...
+    return false;
+  }
+  node_->set_parameter(rclcpp::Parameter("publish_period", servo_publish_rate));
+  return true;
+}
+
+bool ServoSolver::reset()
+{
+  RCLCPP_INFO(getLogger(), "Reset Servo Solver");
+  joint_cmd_rolling_window_.clear();
```
pac48, line 96: Does this get called on every iteration of of local planning or is it once per global plan?

moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/servo_solver.cpp
pac48: This is correct as long as this loop always gets called three times. I think you should replace the whole loop with a for loop of length 3 to make sure this is the case. Then check at the end to make sure the message is not empty.

pac48 (commented):
@sjahr  I added a few more clarifying comments

moveit_ros/hybrid_planning/local_planner/local_constraint_solver_plugins/src/servo_solver.cpp
sjahr: Only once per global plan. Would it make sense to reset the rolling window when the reference pose changes?

pac48: That is interesting. I thought it may have been called on each iteration. In that case, won't the `joint_cmd_rolling_window_` queue grow indefinitely as `solve` continues to be called? In that case, what I would do is remove old commands from `joint_cmd_rolling_window_` in the `solve` method. Then I think you can set the timestep of servo queal to the time step of the local planner.

sjahr:
Closing this due to not having time to work further on this

PR #2681, opened 2024-02-12
Get robot description from topic in GetUrdfService

Abishalini:
### Description

If robot description is not passed as a parameter to the move_group node, this service will fail.
The robot description will now be acquired from the `robot_description` topic

### DIFF
--- a/moveit_ros/move_group/src/default_capabilities/get_group_urdf_capability.cpp
+++ b/moveit_ros/move_group/src/default_capabilities/get_group_urdf_capability.cpp
@@ -63,7 +63,7 @@ GetUrdfService::GetUrdfService() : MoveGroupCapability("get_group_urdf")
 void GetUrdfService::initialize()
 {
   robot_description_subscriber_ = context_->moveit_cpp_->getNode()->create_subscription<std_msgs::msg::String>(
-      "robot_description", rclcpp::SystemDefaultsQoS(),
+      "robot_description", rclcpp::QoS(1).transient_local().reliable(),
       [this](const std_msgs::msg::String::ConstSharedPtr& msg) { return robotDescriptionSubscriberCallback(msg); });
 
   get_urdf_service_ = context_->moveit_cpp_->getNode()->create_service<moveit_msgs::srv::GetGroupUrdf>(
--- a/moveit_ros/move_group/src/default_capabilities/get_group_urdf_capability.cpp
+++ b/moveit_ros/move_group/src/default_capabilities/get_group_urdf_capability.cpp
@@ -56,7 +56,11 @@ const auto ROBOT_ELEMENT_CLOSING = std::string("</robot>");
 const auto GENERAL_ELEMENT_CLOSING = std::string("/>");
 }  // namespace
 
-GetUrdfService::GetUrdfService() : MoveGroupCapability("get_group_urdf")
+GetUrdfService::GetUrdfService()
+  : MoveGroupCapability("get_group_urdf")
+  , robot_description_subscriber_{ context_->moveit_cpp_->getNode()->create_subscription<std_msgs::msg::String>(
+        "robot_description", rclcpp::SystemDefaultsQoS(),
+        [this](const std_msgs::msg::String::ConstSharedPtr& msg) { return robotDescriptionSubscriberCallback(msg); }) }
 {
 }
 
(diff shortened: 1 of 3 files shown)


moveit_ros/move_group/src/default_capabilities/get_group_urdf_capability.cpp
```diff
@@ -56,7 +56,11 @@ const auto ROBOT_ELEMENT_CLOSING = std::string("</robot>");
 const auto GENERAL_ELEMENT_CLOSING = std::string("/>");
 }  // namespace
 
-GetUrdfService::GetUrdfService() : MoveGroupCapability("get_group_urdf")
+GetUrdfService::GetUrdfService()
+  : MoveGroupCapability("get_group_urdf")
+  , robot_description_subscriber_{ context_->moveit_cpp_->getNode()->create_subscription<std_msgs::msg::String>(
+        "robot_description", rclcpp::SystemDefaultsQoS(),
```
tylerjw, line 62:
```suggestion
        "robot_description", rclcpp::QoS(1).transient_local().reliable(),
```

tylerjw (approved):
This all looks good except for the QoS on the subscriber. Set it to transient local to make sure you get a copy of the message if the subscriber starts after the publisher.

moveit_ros/move_group/src/default_capabilities/get_group_urdf_capability.cpp
MarqRazz: I thought we discussed getting `robot_description` from the robot state publisher, so I expected this to be namespaced with that node. This reads as if its getting it from the `moveit_cpp` node. What am I missing here?

tylerjw: I think it is just using the moveit_cpp node to create the subscription. The description as far as I can tell is published by the robot state publisher to a unamespaced topic: `/robot_description` unless you put the robot state publisher in a namespace which we don't right now.

moveit_ros/move_group/src/default_capabilities/get_group_urdf_capability.cpp
```diff
@@ -62,6 +62,10 @@ GetUrdfService::GetUrdfService() : MoveGroupCapability("get_group_urdf")
 
 void GetUrdfService::initialize()
 {
+  robot_description_subscriber_ = context_->moveit_cpp_->getNode()->create_subscription<std_msgs::msg::String>(
```
MarqRazz, line 65: Should we make sure the topic `robot_description` is available and wait for a given amount of time if not?

Abishalini: I tried using [this rclcpp function](https://github.com/ros2/rclcpp/blob/foxy/rclcpp/include/rclcpp/wait_for_message.hpp#L39-L44) to wait for a certain time until we get a message in the `robot_description` topic. But this causes a timeout error when using this service. Not sure I'm using it wrong.

I could add a while loop with a timer to check if the `full_urdf_string_` variable is set. 

Do you have a better solution to this?

tylerjw: I thought about this a bit more and because you have a moveit_cpp you should be able to get the URDF string like this and not need to manually handle the subscription.

If you add a function to RDFLoader to get the `urdf_string_` value it contains. Maybe name it `getUrdfString() -> std::string`. Then you can do this:

```context_->moveit_cpp_->getPlanningSceneMonitor()->getRobotModelLoader()->getRDFLoader()->getUrdfString();```

I know it is ugly but if you have a moveit_cpp and it is initialized you should be able to get the string this way too.

File: moveit_ros/move_group/src/default_capabilities/get_group_urdf_capability.cpp [hunk 3]
```diff
@@ -41,11 +41,16 @@
 #include <moveit/utils/logger.hpp>
 #include <urdf_parser/urdf_parser.h>
 
+#include "rclcpp/wait_for_message.hpp"
```
tylerjw: You should be able to drop this now, right?

tylerjw (approved):

tylerjw:
It's sort of related; do you mind reviewing this? https://github.com/ros-planning/moveit2/pull/2682

PR #2682, opened 2024-02-13
srdf publisher node

tylerjw:
- Sort lists in moveit_ros/planning CMake and package.xml
- srdf_publisher_node

### Description

Similar to how [robot_state_publisher](https://github.com/ros/robot_state_publisher/tree/rolling) publishes a string to `/robot_description` for the URDF this publishes to `/robot_description_semantic` controlled by a parameter of the same name.

MoveIt already subscribes to the topic `/robot_description_semantic,` so this node could be useful in a system using MoveIt.


moveit_ros/planning/srdf_publisher_node/src/srdf_publisher_node.cpp
```diff
@@ -0,0 +1,95 @@
...
+public:
+  SrdfPublisher(const rclcpp::NodeOptions& options) : rclcpp::Node("srdf_publisher", options)
+  {
+    srdf_publisher_ = this->create_publisher<std_msgs::msg::String>("robot_description_semantic",
+                                                                    rclcpp::QoS(1).transient_local().reliable());
+    on_set_parameters_handle_ =
+        this->add_on_set_parameters_callback([this](const std::vector<rclcpp::Parameter>& parameters) {
+          for (auto const& parameter : parameters)
+          {
+            if (parameter.get_name() == "robot_description_semantic")
```
moriarty, line 72: Can there be multiple `robot_description_semantic` parameter? or should if the get_name( ) == then should this function return?

moveit_ros/planning/srdf_publisher_node/src/srdf_publisher_node.cpp
```diff
@@ -0,0 +1,95 @@
...
+            {
+              std_msgs::msg::String msg;
+              msg.data = parameter.get_value<std::string>();
+              srdf_publisher_->publish(msg);
+            }
+          }
+
+          rcl_interfaces::msg::SetParametersResult result;
+          result.successful = true;
+          return result;
```
moriarty, line 82: is this callback always suppose to return success or should it return success in the if, and false otherwise?

tylerjw: I'm adding a comment about this. In humble the only callback we have for parameters is the validation one. You may ask, why don't I validate this is a valid srdf? I considered that until I realized the srdf parser requires a urdf as a parameter which I don't want to make this node get the urdf just to validate the srdf. This node is just a relay anyway and if you use it to set an invalid srdf the nodes that depend on it will blow up.

After humble is eol we can use add_post_set_parameters_callback which is only in rolling now. I'm adding a comment explaining this.

moveit_ros/planning/srdf_publisher_node/src/srdf_publisher_node.cpp
tylerjw: Adding a break here instead of a return because we need to return success through the result for now.

moriarty (approved):
thanks for the clarification comment

PR #2683, opened 2024-02-13
CMake format and lint in pre-commit

tylerjw:
- cmake-format
- cmake_lint
- Drop ament_lint_cmake

### Description

I think it is a much better developer experience having cmake consistently formatted and linted in pre-commit than in ament.


sjahr (commented):
Awesome, thank you! For some reason, this change leads to a huge amount of failures like this:
```
- moveit_core.cpplint whitespace/line_length [2] (/home/runner/work/moveit2/moveit2/.work/target_ws/src/moveit2/moveit_core/collision_detection_bullet/src/bullet_integration/bullet_utils.cpp:226)
  <<< failure message
    Lines should be <= 100 characters long
```
Do you know why this is happening? Once it is fixed, we can merge this

tylerjw:
Here is the list of linters that come from ament_lint. Now that we are dropping cmake-lint and using it in pre-commit none of these make much sense for us so I'm going to drop all of them.

https://github.com/ament/ament_lint/blob/rolling/ament_lint_common/doc/index.rst

sjahr (approved):

PR #2684, opened 2024-02-13
Fix xacro args loading issue

alexnavtt:
In certain cases, the xacro args provided to `moveit_setup_assistant` were not loaded properly by `moveit_configs_builder`. This was due to an incorrect indentation in `moveit_configs_builder.py` which had the xacro args loaded only if `self.__urdf_package` was `None` which seemed incorrect to me. Now it depends only on the existence of the urdf config file.


tylerjw:
Please run pre-commit to format your python code with black.

alexnavtt:
Also, as an aside, xacro arg loading was introduced for Iron, but not for Humble. Is there a reason for that or did it simply fall through the cracks?

alexnavtt:
Sure thing. I'll wait until this PR is merged then I'll cherry pick the two commits to get it in. However, one of the CI tests seems to be failing and I don't know how to fix that. Any suggestions?

sjahr:
> Sure thing. I'll wait until this PR is merged then I'll cherry pick the two commits to get it in. However, one of the CI tests seems to be failing and I don't know how to fix that. Any suggestions?

Thanks, this test failure is not caused by these changes but the test is sometimes flacky. I'll merge main into this PR one last time and then we can merge it

alexnavtt:
Still have checks failing which I don't believe are on my end

PR #2689, opened 2024-02-15
Update moveit::core::error_msg_to_string in Humble

### LINKED ISSUE #2688: moveit::core::error_code_to_string broken in Humble
### Description

`moveit::core::error_code_to_string` was not updated to match the updated __moveit_msgs/msg/MoveItErrorCodes.msg__ definition. For example, an invalid start state gives -26 (START_STATE_INVALID), but that is not a checked value in the switch statement.It also has no default case, which causes the program to crash if you have a case not accounted for (like START_STATE_INVALID). This is fixed in Iron onwards.

### Your environment
* ROS Distro: Humble
* OS Version: Ubuntu 22.04
* Binary - recent apt install

### Steps to reproduce
Have plan start state in collision and call 
`moveit_cpp::PlanningComponent::PlanSolution plan_solution = planning_component.plan(plan_parameters_);` 
When the plan fails, call `moveit::core::error_code_to_string(plan_solution.error_code)`

### Expected behaviour
Get a string describing the error code

### Actual behaviour
Crash the program (throw std::bad_alloc)

reidchristopher:
### Description

Update moveit::core::error_msg_to_string to match updates to moveit_msgs/msg/MoveItErrorCodes.msg. Simply pulled down the file from Iron.

Fixes #2688 

### DIFF
--- a/moveit_core/utils/include/moveit/utils/moveit_error_code.h
+++ b/moveit_core/utils/include/moveit/utils/moveit_error_code.h
@@ -47,11 +47,7 @@ namespace core
 class MoveItErrorCode : public moveit_msgs::msg::MoveItErrorCodes
 {
 public:
-  MoveItErrorCode()
-  {
-    val = 0;
-  }
-  MoveItErrorCode(int code)
+  MoveItErrorCode(int code = 0)
   {
     val = code;
   }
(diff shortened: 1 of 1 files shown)


sjahr (approved):
Thanks for providing this fix!

henningkayser (approved):

PR #2690, opened 2024-02-16
Return all colliding links from getCollidingLinks function in Planning Scene

Abishalini:
### Description

Currently, this function only returns robot links in collision. If the robot is in collision with another link in the environment, it would be helpful to get the environment link as part of this function. 

TODO
- [ ] Is it better for this function to return `std::vector<std::pair<std::string, std::string>>` instead of `std::vector<std::string>`

### Checklist

### DIFF
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -511,10 +511,8 @@ void PlanningScene::getCollidingLinks(std::vector<std::string>& links, const mov
   {
     for (const collision_detection::Contact& contact : it->second)
     {
-      if (contact.body_type_1 == collision_detection::BodyTypes::ROBOT_LINK)
-        links.push_back(contact.body_name_1);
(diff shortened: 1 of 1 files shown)


sjahr (commented):
:thinking: I guess the intend of this function was to return only a list of robot link names. Not sure how this information alone is helpful and I agree that the collision pair information would be more useful ... There seems to be a function available that returns this information already:  https://github.com/ros-planning/moveit2/blob/e9b214ccb7a1f99732a5489d4d031980cf4c09ff/moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h#L522C8-L522C25

Abishalini:
@sjahr I don't know how I missed that function. I'll check if that works well for my bug fix and close this PR accordingly.

sjahr:
> @sjahr I don't know how I missed that function. I'll check if that works well for my bug fix and close this PR accordingly.

Sounds good! Don't worry about it, it's a very convoluted class :see_no_evil:

Abishalini:
Closing this since the function @sjahr mentioned works for my case.

PR #2691, opened 2024-02-16
CPP Best Practices

### LINKED ISSUE #852: Standardize comment style
https://github.com/cpp-best-practices/cppbestpractices/blob/master/03-Style.md#comments

### LINKED ISSUE #1928: Fix overriding binary install
Overriding is when you build a package from a source that is also installed as a binary. Currently, this does not work in MoveIt because of the way header files are found. The problem is that the binary header is found before the source header when building and if the source one is newer and has changes the compile will break in surprising ways.

#1901 is a good example of what needs to be done everywhere in moveit. 

Here are the colcon docs on overriding: https://colcon.readthedocs.io/en/released/user/overriding-packages.html#how-to-make-it-easier-for-your-users-to-override

Tasks:
- [ ] Search for cmake files that are missing this package directory
- [ ] Make PR fixing the issue

BhavdeepSinghNijhawan:
Fixes #852
Fixes #1928
### Checklist


sjahr (changes requested):
This PR doesn't build and doesn't address the mentioned issues. Why should the install directory be changed? Closing this PR.
@BhavdeepSinghNijhawan you've opened a bunch of PRs that looked like you were not sure what you're actually doing. Please follow these [guidelines for contributing to moveit](https://moveit.ros.org/documentation/contributing/pullrequests/). Furthermore, please
- Make sure your code is building locally before submitting a PR
- Add a proper verbose description of your changes to the PR.
- If you're unsure or need help, start with trying to fix a single issue with one PR and ask questions in that PR so we can help you. Opening multiple not working PRs is considered rude.

Issue #2581, opened 2023-12-06
[Planning Pipeline] Make planning pipeline respect the max. planning time


sjahr:
Planning time is currently checked in the planner plugin under the assumption that this is the only process time in the pipeline that consumes a meaningful amount of time. This assumption was wrong and got even more invalidated with the recent refactoring that enables chaining planners.

Planning time should be measured and tracked at the pipeline level [when the planning algorithms are called ](https://github.com/ros-planning/moveit2/blob/f596b9df89c07805f5b911936486ac9f165543b0/moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp#L265)

TomCC7:
Hi @sjahr, I'm interested in working on this issue and have several questions regarding the requirements:

1. Should we take the request and response adapter processing time into account?
2. For chained planners, how should we distribute the `allowed_planning_time` among all planners? It seems to be not optimal if we just let the previous planner use up all the available time. Maybe the planners can share the available time evenly?

For implementation details, it seems like we can directly modify the `allowed_planning_time` before passing it into the planner [here](https://github.com/ros-planning/moveit2/blob/f596b9df89c07805f5b911936486ac9f165543b0/moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp#L299) so that we don't need to add extra parameters to the planning request. Does this solution sound viable to you?

TomCC7:
Thanks for replying! I'll work on this then

PR #2692, opened 2024-02-18
WIP: Make planning pipeline respect the allowed_planning_time

TomCC7:
### Description
@sjahr WIP PR to solve #2581. Several notes:
1. Modified `allowed_planning_time` directly to change the timeout of each planner
2. Currently the timeout for each planner is naively evenly distributed as discussed in the issue.
3. Haven't thoroughly tested the modification. I wonder if is there an easy way to do this.

### Checklist

### DIFF
--- a/moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
+++ b/moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
@@ -37,6 +37,9 @@
 #include <moveit/planning_pipeline/planning_pipeline.h>
 #include <fmt/format.h>
 #include <moveit/utils/logger.hpp>
+#include <chrono>
+#include <ratio>
+#include <stdexcept>
 
 namespace
 {
@@ -267,10 +270,14 @@ bool PlanningPipeline::generatePlan(const planning_scene::PlanningSceneConstPtr&
   // ---------------------------------
   // Solve the motion planning problem
   // ---------------------------------
-
   planning_interface::MotionPlanRequest mutable_request = req;
   try
   {
+    using clock = std::chrono::system_clock;
+    const auto timeout_error = std::runtime_error("allowed_planning_time exceeded");
+    const auto plan_start_time = clock::now();
+    const double allowed_planning_time = req.allowed_planning_time;
+
     // Call plan request adapter chain
     for (const auto& req_adapter : planning_request_adapter_vector_)
     {
@@ -288,11 +295,24 @@ bool PlanningPipeline::generatePlan(const planning_scene::PlanningSceneConstPtr&
                      req_adapter->getDescription().c_str(), status.message.c_str());
         break;
       }
+
+      // check for timeout
+      if (std::chrono::duration<double>(clock::now() - plan_start_time).count() >= allowed_planning_time)
+      {
+        throw timeout_error;
+      }
     }
+
+    // modify planner request to notice plugins of their corresponding max_allowed_time
+    // NOTE: currently just evenly distributing the remaining time
+    const double max_single_planner_time = std::chrono::duration<double>(clock::now() - plan_start_time).count() /
+                                           pipeline_parameters_.planning_plugins.size();
+    mutable_request.allowed_planning_time = max_single_planner_time;
 
     // Call planners
     for (const auto& planner_name : pipeline_parameters_.planning_plugins)
     {
+      auto planner_start_time = clock::now();
       const auto& planner = planner_map_.at(planner_name);
       // Update reference trajectory with latest solution (if available)
       if (res.trajectory)
@@ -323,6 +343,13 @@ bool PlanningPipeline::generatePlan(const planning_scene::PlanningSceneConstPtr&
         RCLCPP_ERROR(node_->get_logger(), "Planner '%s' failed", planner->getDescription().c_str());
         break;
       }
+
+      // TODO: should this be optional since the plugins already checked for timeout?
+      // check for timeout
+      if (std::chrono::duration<double>(clock::now() - planner_start_time).count() >= max_single_planner_time)
+      {
+        throw timeout_error;
+      }
     }
 
     // Call plan response adapter chain
@@ -341,6 +368,12 @@ bool PlanningPipeline::generatePlan(const planning_scene::PlanningSceneConstPtr&
           RCLCPP_ERROR(node_->get_logger(), "PlanningResponseAdapter '%s' failed",
                        res_adapter->getDescription().c_str());
           break;
+        }
+
+        // check for timeout
+        if (std::chrono::duration<double>(clock::now() - plan_start_time).count() >= allowed_planning_time)
+        {
+          throw timeout_error;
         }
       }
     }


moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
```diff
@@ -288,11 +295,24 @@ bool PlanningPipeline::generatePlan(const planning_scene::PlanningSceneConstPtr&
                      req_adapter->getDescription().c_str(), status.message.c_str());
         break;
       }
+
+      // check for timeout
+      if (std::chrono::duration<double>(clock::now() - plan_start_time).count() >= allowed_planning_time)
+      {
+        throw timeout_error;
```
sjahr, line 302: I don't think that a timeout is a reason to throw. Instead, you should return an unsuccessful MotionPlanResponse with the [timeout error code](https://github.com/ros-planning/moveit_msgs/blob/9756797dcbc293c7cae66541a7bc0c7acec85892/msg/MoveItErrorCodes.msg#L12).
For example:
```cpp
        RCLCPP_ERROR(node_->get_logger(),
                     "Planning pipeline couldn't solve the motion planning problem within the allowed planning time (%lf s). Last element called '%s'",req_adapter->getDescription().c_str()) ;
        res.error_code = moveit::core::MoveItErrorCode::TIMED_OUT;
        break;
```

moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
```diff
@@ -288,11 +295,24 @@ bool PlanningPipeline::generatePlan(const planning_scene::PlanningSceneConstPtr&
                      req_adapter->getDescription().c_str(), status.message.c_str());
         break;
       }
+
+      // check for timeout
```
sjahr, line 299: Add a note here that it is assumed that the request adapters only consume a very small fraction of the allowed planning time and thus they don't get a budget assigned. The timeout check is mainly a sanity check

moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
```diff
@@ -323,6 +343,13 @@ bool PlanningPipeline::generatePlan(const planning_scene::PlanningSceneConstPtr&
         RCLCPP_ERROR(node_->get_logger(), "Planner '%s' failed", planner->getDescription().c_str());
         break;
       }
+
+      // TODO: should this be optional since the plugins already checked for timeout?
+      // check for timeout
+      if (std::chrono::duration<double>(clock::now() - planner_start_time).count() >= max_single_planner_time)
```
sjahr, line 349: Maybe I would even make sense to use the same check here as for the adapters
```cpp
if (std::chrono::duration<double>(clock::now() - plan_start_time).count() >= allowed_planning_time)
```
I think we might not care too much about the individual elements exceeding their allocated individual budget (for now) but for the overall planning process to not exceed the `allowed_planning_time` :thinking:

sjahr (commented):
Nice work, thank you!

Regarding your comments:

> Modified allowed_planning_time directly to change the timeout of each planner

That's fine :+1: 

> Currently the timeout for each planner is naively evenly distributed as discussed in the issue.

:+1:  Sounds good for now

> Haven't thoroughly tested the modification. I wonder if is there an easy way to do this. I saw a tutorial [here](https://moveit.picknik.ai/main/doc/examples/tests/tests_tutorial.html) but it seems outdated.

You could extend the [existing unit tests](https://github.com/ros-planning/moveit2/blob/main/moveit_ros/planning/planning_pipeline/test/planning_pipeline_tests.cpp)

TomCC7:
Thanks for the review! Based on your feedback I would propose that we don't enforce and check for timeout after each step but only modify the corresponding remaining time to guide the planners and adapters to follow the time limit. 

We can do a single check after response adapters finished and modify the error code accordingly. In this way even if the timeout is reached the function will also finish so that users can still get the planning result. Does that sound good?

moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
sea-bass: Agreed. Especially because some planners will take longer, so splitting the timeout evenly seems risky.

sjahr:
> Thanks for the review! Based on your feedback I would propose that we don't enforce and check for timeout after each step but only modify the corresponding remaining time to guide the planners and adapters to follow the time limit.
> 
> We can do a single check after response adapters finished and modify the error code accordingly. In this way even if the timeout is reached the function will also finish so that users can still get the planning result. Does that sound good?

The check in every iteration makes sense to me, so I wouldn't change that. If the first out of ten pipeline elements uses the full time budget, there is no point in running the other elements before reporting the error. I think we should always check if the full time budget is exhausted but we don't care so much if a planner uses the allocated fraction of the budget or a bit more as long as the overall planning time remains within the budget

TomCC7:
@sjahr sorry for the late update. I revised my implementation based on your feedback. Here's the update:
1. after checking the timeout a `moveit::core::MoveItErrorCode::TIMED_OUT` error code is sent, which will be catched by error code checks later
2. now the unused time by previous planner will be automatically released for later planners to use
3. added timeout check in the unit test files and verified that it's working

I think if you are good with the current time distribution strategy this pull request is ready for review and merge!

moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
```diff
@@ -267,19 +270,32 @@ bool PlanningPipeline::generatePlan(const planning_scene::PlanningSceneConstPtr&
...
+
     // Call plan request adapter chain
     for (const auto& req_adapter : planning_request_adapter_vector_)
     {
       assert(req_adapter);
       RCLCPP_INFO(node_->get_logger(), "Calling PlanningRequestAdapter '%s'", req_adapter->getDescription().c_str());
       const auto status = req_adapter->adapt(planning_scene, mutable_request);
       res.error_code = status.val;
+      std::string message = status.message;
+
```
sjahr, line 288:
```suggestion
```

moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
```diff
@@ -267,19 +270,32 @@ bool PlanningPipeline::generatePlan(const planning_scene::PlanningSceneConstPtr&
...
       // Publish progress
       publishPipelineState(mutable_request, res, req_adapter->getDescription());
+
+      // check for timeout (mainly for sanity check since adapters are fast)
+      if (std::chrono::duration<double>(clock::now() - plan_start_time).count() >= allowed_planning_time)
+      {
+        message = "failed to finish within " + std::to_string(allowed_planning_time) + "s";
```
sjahr, line 295:
```suggestion
        status.message = "Failed to finish planning within allowed planning time ( " + std::to_string(allowed_planning_time) + "s )";
```

sjahr: No need to write short error messages. Be as precise and long as necessary :wink: (If you accept this change you'll still need to format the code correctly with pre-commit)

moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
```diff
@@ -291,9 +307,15 @@ bool PlanningPipeline::generatePlan(const planning_scene::PlanningSceneConstPtr&
     }
 
     // Call planners
-    for (const auto& planner_name : pipeline_parameters_.planning_plugins)
+    for (size_t i = 0; i < pipeline_parameters_.planning_plugins.size(); i++)
     {
-      const auto& planner = planner_map_.at(planner_name);
+      // modify planner request to notice plugins of their corresponding max_allowed_time
+      // NOTE: currently just evenly distributing the remaining time among the remaining planners
+      double max_single_planner_time = std::chrono::duration<double>(clock::now() - plan_start_time).count() /
```
sjahr, line 314: I don't understand this time calculation. Can you explain it? Why are we giving the planner the time since the planning started as time budget? Maybe I am missing something. Also can you make this const?

moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
```diff
@@ -291,9 +307,15 @@ bool PlanningPipeline::generatePlan(const planning_scene::PlanningSceneConstPtr&
...
+                                       (pipeline_parameters_.planning_plugins.size() - i);
+      mutable_request.allowed_planning_time = max_single_planner_time;
```
sjahr, line 316: On second thought, no need to create `max_single_planner_time` just use `mutable_request.allowed_planning_time`

moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
```diff
@@ -291,9 +307,15 @@ bool PlanningPipeline::generatePlan(const planning_scene::PlanningSceneConstPtr&
...
+
+      const auto& planner = planner_map_.at(pipeline_parameters_.planning_plugins[i]);
```
sjahr, line 318:
```suggestion
      const auto& planner = planner_map_.at(pipeline_parameters_.planning_plugins.at(i));
```

moveit_ros/planning/planning_pipeline/test/planning_pipeline_test_plugins.cpp
```diff
@@ -38,6 +38,7 @@
 #include <moveit/planning_interface/planning_interface.h>
 #include <moveit/planning_interface/planning_request_adapter.h>
 #include <moveit/planning_interface/planning_response_adapter.h>
+#include <moveit/utils/moveit_error_code.h>
```
sjahr, line 41: Why is this necessary?

moveit_ros/planning/planning_pipeline/test/planning_pipeline_tests.cpp
```diff
@@ -115,6 +117,34 @@ TEST_F(TestPlanningPipeline, NoPlannerPluginConfigured)
                std::runtime_error);
 }
 
+TEST_F(TestPlanningPipeline, TestTimeout)
```
sjahr, line 120: Awesome that you've extended the unit-tests :heart:

moveit_ros/planning/planning_pipeline/test/planning_pipeline_tests.cpp
```diff
@@ -115,6 +117,34 @@ TEST_F(TestPlanningPipeline, NoPlannerPluginConfigured)
...
+  planning_interface::MotionPlanRequest motion_plan_request;
+  const auto planning_scene_ptr = std::make_shared<planning_scene::PlanningScene>(robot_model_);
+
+  // timeout by request adapter
+  motion_plan_request.allowed_planning_time = 0.1;
+  EXPECT_FALSE(pipeline_ptr_->generatePlan(planning_scene_ptr, motion_plan_request, motion_plan_response));
+  EXPECT_EQ(motion_plan_response.error_code, moveit::core::MoveItErrorCode::TIMED_OUT);
+
+  // timeout by planner
+  motion_plan_request.allowed_planning_time = 1;
```
sjahr, line 138:
```suggestion
  motion_plan_request.allowed_planning_time = 1.0;
```

moveit_ros/planning/planning_pipeline/test/planning_pipeline_tests.cpp
```diff
@@ -115,6 +117,34 @@ TEST_F(TestPlanningPipeline, NoPlannerPluginConfigured)
...
+{
+  // construct pipline
+  EXPECT_NO_THROW(pipeline_ptr_ = std::make_shared<planning_pipeline::PlanningPipeline>(
+                      robot_model_, node_, "", PLANNER_PLUGINS, REQUEST_ADAPTERS, RESPONSE_ADAPTERS));
+
+  // WHEN generatePlan is called
+  // THEN A successful response is created
+  planning_interface::MotionPlanResponse motion_plan_response;
+  planning_interface::MotionPlanRequest motion_plan_request;
+  const auto planning_scene_ptr = std::make_shared<planning_scene::PlanningScene>(robot_model_);
```
sjahr, line 130: I think the successful case is already covered by earlier tests, so you can remove it here

moveit_ros/planning/planning_pipeline/test/planning_pipeline_tests.cpp
```diff
@@ -115,6 +117,34 @@ TEST_F(TestPlanningPipeline, NoPlannerPluginConfigured)
...
+
+  // timeout by planner
+  motion_plan_request.allowed_planning_time = 1;
+  EXPECT_FALSE(pipeline_ptr_->generatePlan(planning_scene_ptr, motion_plan_request, motion_plan_response));
+  EXPECT_EQ(motion_plan_response.error_code, moveit::core::MoveItErrorCode::TIMED_OUT);
+
+  // passes
+  motion_plan_request.allowed_planning_time = 10;
+  EXPECT_TRUE(pipeline_ptr_->generatePlan(planning_scene_ptr, motion_plan_request, motion_plan_response));
+  EXPECT_EQ(motion_plan_response.error_code, moveit::core::MoveItErrorCode::SUCCESS);
```
sjahr, line 145: No need to rest this a third time, remove the passed tests

moveit_ros/planning/planning_pipeline/test/planning_pipeline_tests.cpp
```diff
@@ -115,6 +117,34 @@ TEST_F(TestPlanningPipeline, NoPlannerPluginConfigured)
...
+  EXPECT_NO_THROW(pipeline_ptr_ = std::make_shared<planning_pipeline::PlanningPipeline>(
+                      robot_model_, node_, "", PLANNER_PLUGINS, REQUEST_ADAPTERS, RESPONSE_ADAPTERS));
+
+  // WHEN generatePlan is called
+  // THEN A successful response is created
+  planning_interface::MotionPlanResponse motion_plan_response;
+  planning_interface::MotionPlanRequest motion_plan_request;
+  const auto planning_scene_ptr = std::make_shared<planning_scene::PlanningScene>(robot_model_);
+
+  // timeout by request adapter
```
sjahr, line 132: Can you also the the response adapter?

sjahr (commented):
Thanks, I reviewed your changes

moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
TomCC7: sorry made a mistake, I changed it to
```
mutable_request.allowed_planning_time =
          (allowed_planning_time - std::chrono::duration<double>(clock::now() - plan_start_time).count()) /
          (pipeline_parameters_.planning_plugins.size() - i);
```

moveit_ros/planning/planning_pipeline/test/planning_pipeline_test_plugins.cpp
TomCC7: removed

moveit_ros/planning/planning_pipeline/test/planning_pipeline_tests.cpp
```diff
@@ -119,7 +119,7 @@ TEST_F(TestPlanningPipeline, NoPlannerPluginConfigured)
 
 TEST_F(TestPlanningPipeline, TestTimeout)
 {
-  // construct pipline
+  // construct pipeline
```
TomCC7, line 122: didn't know that checker also checks for spelling lol

TomCC7:
hi @sjahr, just checking in, is there anything else I need to do on this PR?

moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
```diff
@@ -317,6 +339,12 @@ bool PlanningPipeline::generatePlan(const planning_scene::PlanningSceneConstPtr&
       context->solve(res);
       publishPipelineState(mutable_request, res, planner->getDescription());
 
+      // check for overall timeout
+      if (std::chrono::duration<double>(clock::now() - plan_start_time).count() >= allowed_planning_time)
```
henningkayser, line 343: Maybe I'm missing it, but it looks like we will hit a timeout here if the planners use up the allowed planning time. Optimizing planners usually do that, so I think we should either ignore response adapters or use a separate budget

moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
```diff
@@ -335,6 +363,13 @@ bool PlanningPipeline::generatePlan(const planning_scene::PlanningSceneConstPtr&
         RCLCPP_INFO(node_->get_logger(), "Calling PlanningResponseAdapter '%s'", res_adapter->getDescription().c_str());
         res_adapter->adapt(planning_scene, mutable_request, res);
         publishPipelineState(mutable_request, res, res_adapter->getDescription());
+
+        // check for timeout
```
henningkayser, line 367: Maybe I'm missing it, but it looks like we will hit a timeout here if the planners use up the allowed planning time. Optimizing planners usually do that, so I think we should either ignore response adapters or use a separate budget

TomCC7: Yeah I think the ultimate solution to this is to give user more control over the timeout. Maybe let them decide timeout of each section (and each planner) separately? But that would require changing the `MotionPlanRequest` message so I'm not sure. What's your suggestion on this? Should we make the PR more complex or just ignore the response adapters for now?

sjahr: Can we ensure for now that a fraction of the time will be reserved for the planners? Something like 100ms in any case?

TomCC7: But in that case the planner might still use up all the reserved time. Or are you saying that giving the request adapter at least 100ms?

henningkayser: the response adapter should have a reserved time budget which is deducted from the planner's allowed planning time.

TomCC7: hi, I add an extra check to ensure that the response adapter has 10ms left in the latest commit

EzraBrooks:
@TomCC7, I apologize for the delay, we've been pretty absorbed with bugfixes for multi-arm and mobile manipulator configurations lately.

@sjahr @henningkayser is there anything else we need changed on this PR before we give it a last round of testing and merge it?

PR #2701, opened 2024-02-21
Update test_robot_trajectory.cpp

### LINKED ISSUE #1728: Use rclcpp::Duration instead of double in RobotTrajectory
Strong types for time should be used where possible.

BhavdeepSinghNijhawan:
Fixes #1728


sjahr (changes requested):
This does not fix the issue but randomly changes some type in the tests. Not even sure if this compiles given that these values are passed to functions that expect double. Please consider changing the interfaces to address the issue properly. I'll close this PR for now

Issue #871, opened 2021-12-01
Char is a char, string is a string


tylerjw:

flip360pranjit:
I will work on this.

prsndp:
@tylerjw  Could I work on this issue? if it is possible please assign me.

tylerjw:
Maybe the two of you can work together or you can pick another one?

sameergupta4873:
@tylerjw I would like to work on this issue

TanmayAgarwal123:
I would love to work on this project, can you assign me
Regards,
Tanmay Agarwal

TaMaN2031A:
I'm working on it now inshallah.

yaswanth1701:
Hi @tylerjw, it seems there are still a few places left where changes related to this issue are required. Can I work on this issue?

PR #2705, opened 2024-02-23
Modified gitignore to avoid tracking .DS_Store MacOS files and also r…

utkarshanand140:
Fixes a part of #871 
Replaced unnecessary strings with characters to make compilation faster. 
Also modified the .gitignore file to avoid tracking .DS_Store files which are often generated when working on MacOS systems.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,3 +76,6 @@ _autosummary
 # source dependencies (see moveit2.repos)
 moveit_msgs
 moveit_resources
+
+# ignore macos generated files
+.DS_Store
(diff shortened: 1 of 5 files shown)


sjahr (commented):
Thanks for these changes. Can you try to fix it everywhere rather than partially? There was an abandoned PR that attempted the same you could pick up and bring over the finish line

utkarshanand140:
> Thanks for these changes. Can you try to fix it everywhere rather than partially? There was an abandoned PR that attempted the same you could pick up and bring over the finish line

Yes sure, I'll be happy to complete it.

EzraBrooks (changes requested):
Please break these changes out into two PRs - I'd be happy to approve and merge the `.gitignore` change separately from the code changes.

EzraBrooks:
The bots are now conversing with each other..

I'm closing this PR since it seems unlikely at this point that the feedback will be addressed.

PR #2723, opened 2024-03-04
Do not overwrite the error code in planWithSinglePipeline

galou:
Return the `MotionPlanResponse` as-is, also in case of error.

### Description

The error code was forced-set to `FAILURE` in case of error, preventing the caller to get the correct information in case of error.

### Checklist
- [X] Document API changes relevant to the user in the [MIGRATION.md](https://github.com/ros-planning/moveit2/blob/main/MIGRATION.md) notes

### DIFF
--- a/moveit_ros/planning/planning_pipeline_interfaces/src/planning_pipeline_interfaces.cpp
+++ b/moveit_ros/planning/planning_pipeline_interfaces/src/planning_pipeline_interfaces.cpp
@@ -38,7 +38,6 @@
 #include <moveit/utils/logger.hpp>
 
 #include <thread>
-#include <tuple>  // std::ignore.
 
 namespace moveit
 {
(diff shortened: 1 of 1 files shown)


galou:
Should I write a note about this change in [MIGRATION.md](https://github.com/ros-planning/moveit2/blob/main/MIGRATION.md)?

sjahr (commented):
I think it might be possible that generatePlan returns false without properly setting the error_code. What do you think about only overwritting the error code if it is SUCCESS or UNDEFINED? That way nothing is overwritten + we're not ignoring an error

galou:
I did the changes to consider the case the `generatePlan()` return false but do not set the error code

This appears to me to be an API design issue since the failure-or-success piece of information must be set at two places.

sjahr:
> I did the changes to consider the case the `generatePlan()` return false but do not set the error code
> 
> This appears to me to be an API design issue since the failure-or-success piece of information must be set at two places.

Thank you! You're right, maybe this could be updated to return the response in the future to avoid duplicate error codes

sjahr (approved):

PR #2725, opened 2024-03-04
Do not overwrite the error code with OMPL interface

galou:
### Description

In case of failure, set the error code to the one returned by the planning pipeline's `solve` method rather than overwriting it with `PLANNING_FAILED`.

### Checklist

### DIFF
--- a/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
+++ b/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
@@ -802,12 +802,10 @@ void ModelBasedPlanningContext::solve(planning_interface::MotionPlanResponse& re
 void ModelBasedPlanningContext::solve(planning_interface::MotionPlanDetailedResponse& res)
 {
   res.planner_id = request_.planner_id;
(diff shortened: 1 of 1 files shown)


galou:
Should I document the changes in `MIGRATION.md`?

sjahr (commented):
Thanks! I think this can be simplified even further. How about
```diff
void ModelBasedPlanningContext::solve(planning_interface::MotionPlanDetailedResponse& res)
{
  res.error_code = solve(request_.allowed_planning_time, request_.num_planning_attempts);
}
```
My understanding is that the whole if statement is not necessary or am I missing something :thinking:?

galou:
Simplification done.

By the way, what uses `solve(planning_interface::MotionPlanDetailedResponse&)`?

sjahr (approved):
Thanks!

sjahr:
@galou Looks like the PR needs to be rebased on the latest version of main but otherwise it is good to go :+1:

galou:
I rebased and squashed. I hope that it's ok that the conversation may lose context.

galou:
To really see the effects of this PR, we have to wait for https://github.com/ompl/ompl/pull/1147 to be merged.

PR #2727, opened 2024-03-05
Print links in collision

Abishalini:
### Description

The CheckStartStateCollision adapter does return failure when links are in collision but it currently does not say which links are in collision. This PR adds the links in collision to the MoveItErrorCode message so it can help with debugging.

### DIFF
--- a/moveit_ros/planning/planning_request_adapter_plugins/src/check_start_state_collision.cpp
+++ b/moveit_ros/planning/planning_request_adapter_plugins/src/check_start_state_collision.cpp
@@ -78,9 +78,6 @@ public:
     // TODO(sjahr): Would verbose make sense?
     planning_scene->checkCollision(creq, cres, start_state);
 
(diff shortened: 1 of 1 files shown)


moveit_ros/planning/planning_request_adapter_plugins/src/check_start_state_collision.cpp
```diff
@@ -85,8 +85,18 @@ class CheckStartStateCollision : public planning_interface::PlanningRequestAdapt
     }
     else
     {
+      collision_detection::CollisionResult::ContactMap contacts;
+      planning_scene->getCollidingPairs(contacts);
+
+      std::string contact_information = std::to_string(contacts.size()) + " contact(s) detected : ";
```
sjahr, line 91:
```suggestion
      const std::string contact_information = std::to_string(contacts.size()) + " contact(s) detected : ";
```

sjahr (approved):
This will be helpful for debugging, thanks!

PR #2740, opened 2024-03-13
Enable mdof trajectory execution

henningkayser:
This implements preparative steps to enable mobile base trajectory execution using ros2_control JTC.

In particular, these changes convert mdof trajectories into joint trajectories using the local variables which in turn can be supported by ros2_control already.

An example for a functioning setup with the Stretch robot + JTC + chained DiffDrive is provided in https://github.com/PickNikRobotics/stretch_ros/pull/48


moveit_ros/move_group/src/default_capabilities/execute_trajectory_action_capability.cpp
```diff
@@ -106,8 +106,29 @@ void MoveGroupExecuteTrajectoryAction::executePath(const std::shared_ptr<ExecTra
 {
   RCLCPP_INFO(getLogger(), "Execution request received");
 
+  // Convert mdof trajectory to RobotTrajectory
```
henningkayser, line 109: This conversion block needs to be added to the Move action as well. Alternatively, we could bring the conversion into the TEM to reduce some redundancy, however that would probably require even more API changes.

moveit_core/robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -707,4 +707,120 @@ std::optional<double> waypointDensity(const RobotTrajectory& trajectory)
   return std::nullopt;
 }
 
+std::optional<trajectory_msgs::msg::JointTrajectory> toJointTrajectory(const RobotTrajectory& trajectory,
```
henningkayser, line 710: I would still like to consolidate this with the existing getMessage() API

moveit_core/robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -707,4 +707,120 @@ std::optional<double> waypointDensity(const RobotTrajectory& trajectory)
...
+                                                                       bool include_mdof_joints,
+                                                                       const std::vector<std::string>& joint_filter)
+{
+  const auto group = trajectory.getGroup();
+  const auto robot_model = trajectory.getRobotModel();
```
sjahr, line 715: Can this be a nullptr?

moveit_core/robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -707,4 +707,120 @@ std::optional<double> waypointDensity(const RobotTrajectory& trajectory)
...
+    if (!joint_filter.empty() &&
+        std::find(joint_filter.begin(), joint_filter.end(), active_joint->getName()) == joint_filter.end())
+      continue;
+
+    if (active_joint->getVariableCount() == 1)
+    {
+      onedof.push_back(active_joint);
+      joint_trajectory.joint_names.push_back(active_joint->getName());
+    }
+    else if (include_mdof_joints)
```
sjahr, line 738: Is this option crucial? Maybe `include_mdof_joints` could be removed from the interface and the user can just choose to ignore mdof in the data structure

moveit_core/robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -707,4 +707,120 @@ std::optional<double> waypointDensity(const RobotTrajectory& trajectory)
...
+          joint_trajectory.points[i].accelerations.push_back(
+              waypoint.getVariableAcceleration(onedof[j]->getFirstVariableIndex()));
+        }
+        if (waypoint.hasEffort())
+        {
+          joint_trajectory.points[i].effort.push_back(waypoint.getVariableEffort(onedof[j]->getFirstVariableIndex()));
+        }
+      }
+      // clear velocities if we have an incomplete specification
+      if (joint_trajectory.points[i].velocities.size() != onedof.size())
```
sjahr, line 788: Maybe print and error message here, since this probably indicates a bug somewhere

moveit_core/robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -707,4 +707,120 @@ std::optional<double> waypointDensity(const RobotTrajectory& trajectory)
...
+        joint_trajectory.points[i].effort.clear();
+    }
+
+    if (!mdof.empty())
+    {
+      for (const auto joint : mdof)
+      {
+        // Add variable placeholders
+        const std::vector<std::string> names = joint->getVariableNames();
+        joint_trajectory.points[i].positions.reserve(joint_trajectory.points[i].positions.size() + names.size());
```
sjahr, line 804: I probably need to re-read this tomorrow, but the information whether a joint is a multidof joint is lost in the joint_trajectory message right?

moveit_ros/move_group/src/default_capabilities/execute_trajectory_action_capability.cpp
sjahr: Maybe we can move it into TEM as some follow-up tasks? That component needs to be cleaned-up heavily anyways

sjahr (commented):
Overall, I like this change, nice work!

moveit_ros/move_group/src/default_capabilities/execute_trajectory_action_capability.cpp
henningkayser: done

moveit_core/robot_trajectory/src/robot_trajectory.cpp
henningkayser: This knowledge is preserved by the specific naming of the joint. mdof joints have local variables that will be appended with a "/". So a planar `position` joint will produce `position/x`, `position/y`, `position/theta`. A floating joint has seven local variables https://github.com/ros-planning/moveit2/blob/main/moveit_core/robot_model/src/floating_joint_model.cpp#L66. This way we can technically encode any kind of "special" joint in MoveIt, even if it's not properly supported by URDF.

moveit_core/robot_trajectory/src/robot_trajectory.cpp
henningkayser: I think this is crucial if we want to merge this with the original message getter. It would be very surprising for those that expect a multi-dof trajectory.

moveit_core/robot_trajectory/src/robot_trajectory.cpp
henningkayser: Opened issue here https://github.com/ros-planning/moveit2/issues/2773

henningkayser:
I've put some of the open requests into #2773, since they seem out of scope of the new free function to me. I'm happy to provide API cleanup after mdof support has been completed.

moveit_core/robot_trajectory/test/test_robot_trajectory.cpp
```diff
@@ -634,6 +634,31 @@ TEST_F(OneRobot, UnwindFromState)
   }
 }
 
+TEST_F(OneRobot, MultiDofTrajectoryToJointStates)
```
sjahr, line 637: Nit: Do you mind adding a test description here, something like GIVEN, WHEN, THEN

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -51,6 +52,7 @@ static const double DEFAULT_CONTROLLER_GOAL_DURATION_MARGIN = 0.5;  // allow 0.5
                                                                     // after scaling)
 static const double DEFAULT_CONTROLLER_GOAL_DURATION_SCALING =
     1.1;  // allow the execution of a trajectory to take more time than expected (scaled by a value > 1)
+static const bool DEFAULT_CONTROL_MULTI_DOF_JOINT_VARIABLES = false;
```
sjahr, line 55: Do we need this as a static const variable, can't we treat it like the default value of `wait_for_trajectory_completion_`?

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -200,6 +203,7 @@ void TrajectoryExecutionManager::initialize()
   controller_mgr_node_->get_parameter("trajectory_execution.allowed_goal_duration_margin",
                                       allowed_goal_duration_margin_);
   controller_mgr_node_->get_parameter("trajectory_execution.allowed_start_tolerance", allowed_start_tolerance_);
+  controller_mgr_node_->get_parameter("trajectory_execution.control_multi_dof_joint_variables", control_multi_dof_joint_variables_);
```
sjahr, line 206: Future work but if we switch to generate params lib, we could define a default param value and simplify this a bit

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -350,8 +354,45 @@ bool TrajectoryExecutionManager::push(const moveit_msgs::msg::RobotTrajectory& t
     return false;
   }
 
+  // Optionally, convert multi dof waypoints to joint states and replace trajectory for execution
+  std::optional<moveit_msgs::msg::RobotTrajectory> replaced_trajectory;
+  if (control_multi_dof_joint_variables_ && !trajectory.multi_dof_joint_trajectory.points.empty())
+  {
+    // We convert the trajectory message into a RobotTrajectory first,
```
sjahr, line 361: Would it be worth making this a free function in case we might need it outside of the trajectory execution manager?

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -350,8 +354,45 @@ bool TrajectoryExecutionManager::push(const moveit_msgs::msg::RobotTrajectory& t
...
+    // Check success of conversion
+    // This should never happen when using MoveIt's interfaces, but users can pass anything into TEM::push() directly
+    if (!joint_trajectory.has_value())
+    {
+      RCLCPP_ERROR(logger_, "Failed to convert multi-DOF trajectory to joint trajectory, aborting execution!");
+      return false;
+    }
+
+    // Create a new robot trajectory message that only contains the combined joint trajectory
+    RCLCPP_INFO(logger_, "Successfully converted multi-DOF trajectory to joint trajectory for execution.");
```
sjahr, line 389: DEBUG?

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -704,12 +745,12 @@ bool TrajectoryExecutionManager::distributeTrajectory(const moveit_msgs::msg::Ro
...
   for (const std::string& joint_name : trajectory.joint_trajectory.joint_names)
   {
-    const moveit::core::JointModel* jm = robot_model_->getJointModel(joint_name);
+    const moveit::core::JointModel* jm = robot_model_->getJointOfVariable(joint_name);
     if (jm)
     {
       if (jm->isPassive() || jm->getMimic() != nullptr || jm->getType() == moveit::core::JointModel::FIXED)
         continue;
-      actuated_joints_single.insert(jm->getName());
+      actuated_joints_single.insert(joint_name);
```
sjahr, line 753: :sweat_smile:

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -863,30 +904,45 @@ bool TrajectoryExecutionManager::validate(const TrajectoryExecutionContext& cont
...
+        if (reference_state.distance(*current_state, joint) > allowed_start_tolerance_)
         {
           RCLCPP_ERROR(logger_,
-                       "\nInvalid Trajectory: start point deviates from current robot state more than %g"
-                       "\njoint '%s': expected: %g, current: %g",
-                       allowed_start_tolerance_, joint_names[i].c_str(), traj_position, cur_position);
+                       "Invalid Trajectory: start point deviates from current robot state more than %g at joint '%s'."
+                       "\nEnable DEBUG for detailed state info.",
+                       allowed_start_tolerance_, joint->getName().c_str());
+          // TODO(henningkayser): print state info, or treat multi-dof joint separately using the
```
sjahr, line 935: Can you create an issue to track this?

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
Abishalini: @henningkayser Why do we need this parameter? Can control multi-DOF joints be true by default since this is the only way to get multi-DFO execution working?

henningkayser: Because by default multi-dof trajectories should be controlled as multi-dof trajectories :man_shrugging: . Converting to joint variables is a feature that we use in a specific use case with ros2_control, so I would not set this as new default.

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
henningkayser: fixed

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
henningkayser: fixed

moveit_core/robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -707,4 +707,120 @@ std::optional<double> waypointDensity(const RobotTrajectory& trajectory)
...
+      }
+    }
+  }
+
+  if (!joint_trajectory.joint_names.empty())
+  {
+    joint_trajectory.header.frame_id = robot_model->getModelFrame();
+    joint_trajectory.header.stamp = rclcpp::Time(0, 0, RCL_ROS_TIME);
+    joint_trajectory.points.resize(trajectory.getWayPointCount());
+  }
```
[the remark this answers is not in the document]
henningkayser: with the updated logic, I would just keep it the way it is probably

moveit_core/robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -707,4 +707,120 @@ std::optional<double> waypointDensity(const RobotTrajectory& trajectory)
...
+
+  trajectory_msgs::msg::JointTrajectory joint_trajectory;
+  std::vector<const moveit::core::JointModel*> onedof;
+  std::vector<const moveit::core::JointModel*> mdof;
+
+  for (const moveit::core::JointModel* active_joint : jnts)
+  {
+    // only consider joints listed in joint_filter
+    if (!joint_filter.empty() &&
+        std::find(joint_filter.begin(), joint_filter.end(), active_joint->getName()) == joint_filter.end())
```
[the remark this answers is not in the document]
henningkayser: no, that's done at another point when no controller is found (intersections are empty in most cases)

moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
```diff
@@ -410,4 +410,13 @@ std::ostream& operator<<(std::ostream& out, const RobotTrajectory& trajectory);
 /// or nullopt if it is not possible to calculate the density
 [[nodiscard]] std::optional<double> waypointDensity(const RobotTrajectory& trajectory);
 
+/// \brief Converts a RobotTrajectory to a JointTrajectory message
+//  \param[in] trajectory Given robot trajectory
+//  \param[in] include_mdof_joints Treat Multi-DOF variables as joints, e.g. position/x position/y position/theta
```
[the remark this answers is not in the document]
henningkayser: hmm.. isn't that already very clear by the return type and by the variable naming (include=false should mean that it's ignored)

EzraBrooks (approved):
This is working in Pro and the CI failure appears to be a flaky test.

I will re-run the CI here a few times to see if it passes, but otherwise I propose we open another PR to temporarily disable the flaky test.

Issue #2749, opened 2024-03-18
KDL parameters fail to build on latest version of `generate_parameter_library`


sea-bass:
### Description

When building the KDL plugin in the `moveit_kinematics` package, I see build errors because the generated header file from the parameters places two `.` characters in a row, which is invalid C++.

This stems from the parameter map defined in https://github.com/ros-planning/moveit2/blob/main/moveit_kinematics/kdl_kinematics_plugin/src/kdl_kinematics_parameters.yaml#L8-L16

I believe these were caused by this recent PR to generate_parameter_library: https://github.com/PickNikRobotics/generate_parameter_library/pull/166. I filed issue https://github.com/PickNikRobotics/generate_parameter_library/issues/182 in that repo as well.

### Your environment
* ROS Distro: Humble
* OS Version: e.g. Ubuntu 22.04
* Source or Binary build? Source, from `main`
* Which RMW (Fast DDS or Cyclone DDS)? Any

### Backtrace or Console output

```
--- stderr: moveit_kinematics
  In file included from /home/runner/work/moveit2/moveit2/.work/target_ws/src/moveit2/moveit_kinematics/kdl_kinematics_plugin/include/moveit/kdl_kinematics_plugin/kdl_kinematics_plugin.h:42,
                   from /home/runner/work/moveit2/moveit2/.work/target_ws/src/moveit2/moveit_kinematics/kdl_kinematics_plugin/src/kdl_kinematics_plugin.cpp:37:
  /home/runner/work/moveit2/moveit2/.work/target_ws/build/moveit_kinematics/kdl_kinematics_plugin/kdl_kinematics_parameters/include/kdl_kinematics_parameters.hpp: In member function ‘void kdl_kinematics::ParamListener::refresh_dynamic_parameters()’:
  /home/runner/work/moveit2/moveit2/.work/target_ws/build/moveit_kinematics/kdl_kinematics_plugin/kdl_kinematics_parameters/include/kdl_kinematics_parameters.hpp:146:36: error: expected unqualified-id before ‘.’ token
    146 |       auto& entry = updated_params..joints_map[value_1];
        |                                    ^
[...]

PR #2750, opened 2024-03-18
Change KDL IK plugin joint weight parameters specification

sea-bass:
### Description

This is a potential fix for #2749.

It changes the KDL parameters introduced in https://github.com/ros-planning/moveit2/pull/1671 from:

```yaml
panda_arm:
  kinematics_solver: "kdl_kinematics_plugin/KDLKinematicsPlugin"
  kinematics_solver_search_resolution: 0.005
  kinematics_solver_timeout: 0.05

  joints: ["joint1", "joint2", "joint3", "joint4", "joint5", "joint6"]
  # Default weight for joints is 1.0 
  joint1:
    weight: 2.0
  joint5:
    weight: 1.5
```

to 

```yaml
panda_arm:
  kinematics_solver: "kdl_kinematics_plugin/KDLKinematicsPlugin"
  kinematics_solver_search_resolution: 0.005
  kinematics_solver_timeout: 0.05

  joints: ["joint1", "joint2", "joint3", "joint4", "joint5", "joint6"]
  # Default weight for joints is 1.0 
  weights:
    joint1:
      value: 2.0
    joint5:
      value: 1.5
```

### Checklist

### DIFF
--- a/moveit_kinematics/kdl_kinematics_plugin/src/kdl_kinematics_parameters.yaml
+++ b/moveit_kinematics/kdl_kinematics_plugin/src/kdl_kinematics_parameters.yaml
@@ -5,15 +5,16 @@ kdl_kinematics:
     description: "Joints names to assign weights",
   }
 
-  __map_joints:
-    weight: {
-      type: double,
-      default_value: 1.0,
-      description: "Joint weight",
-      validation: {
-        gt<>: [ 0.0 ]
+  weights:
+    __map_joints:
+      value: {
+        type: double,
+        default_value: 1.0,
+        description: "Joint weight",
+        validation: {
+          gt<>: [ 0.0 ]
+        }
       }
-    }
 
   max_solver_iterations: {
     type: int,
--- a/moveit_kinematics/kdl_kinematics_plugin/src/kdl_kinematics_plugin.cpp
+++ b/moveit_kinematics/kdl_kinematics_plugin/src/kdl_kinematics_plugin.cpp
@@ -95,11 +95,10 @@ void KDLKinematicsPlugin::getJointWeights()
   joint_weights_ = std::vector<double>(joint_names.size(), 1.0);
 
   // Check if joint weight is assigned in kinematics YAML
-  // Loop through map (key: joint name and value: Struct with a weight member variable)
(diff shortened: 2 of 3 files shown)


sea-bass:
Ugh, seems like MoveIt Setup Assistant has some failures due to the addition of the nonzero velocity here: https://github.com/ros-planning/moveit_resources/pull/198/files

sjahr (approved):
Do we need to update some moveit_resoures repositories?

sea-bass:
> Do we need to update some moveit_resources packages?

Actually, no, because none of the example configs even sets joint weights in the first place!

But ler's not merge this in yet, as we could also opt to fix generate_parameter_library

rhaschke:
Why don't you simply use a map from joint name to weight (instead of an additional indirection layer):
```yaml
  weights:
    joint1: 2.0
    joint5: 1.5
```

sea-bass:
> Why don't you simply use a map from joint name to weight (instead of an additional indirection layer):
> 
> ```yaml
>   weights:
>     joint1: 2.0
>     joint5: 1.5
> ```

That does sound good, but I'm not sure if `generate_parameter_library` supports doing this at the moment. Looking at the examples: https://github.com/PickNikRobotics/generate_parameter_library?tab=readme-ov-file#mapped-parameters, seems like there has to be a field underneath the mapped parameters.

rhaschke:
> I'm not sure if `generate_parameter_library` supports doing this at the moment.

If it doesn't, you probably should fix that library. A library implementation shouldn't impose constraints on the kind of parameter structures available.

Abishalini:
If humble CI jobs pass, would it make sense to merge this while a fix to `generate_param_library` is being worked on?
This would unblock other PRs in the repo

sea-bass:
Seems there is interest in fixing CI for now, so will open up https://github.com/ros-planning/moveit2/issues/2756 and we'll make sure we get that in for the next MoveIt 2 release.

sea-bass:
Update: I think I fixed `generate_parameter_library` to be able to keep the old syntax: https://github.com/PickNikRobotics/generate_parameter_library/pull/185

gavanderhoorn:
@sea-bass: could the branch be deleted?

Issue #2614, opened 2023-12-15
Multiple OMPL constraints are not possible with 6-DOF arms


medvedevigorek:
### Description

Having 2 path constraints with OMPL constrained state space lead to `Space is over constrained!` exception and as the result failed motion planning. 
```
[moveit_cpp_tutorial-4] [ERROR] [1702624020.869948730] [moveit.ros_planning.planning_pipeline]: Exception caught: ompl::base::Constraint(): Space is over constrained!'
```

This was discovered with UR5e robotic arm while exactly the same logic works with Panda arm (the one used in the examples). 

Note using either one of the same path constraints at a time works fine with both arms.

### Your environment
* ROS Distro: Iron
* OS Version: Ubuntu 22.04
* Source or Binary build? Source
* If binary, which release version?
* If source, which branch? tag 2.8.0
* Which RMW (Fast DDS or Cyclone DDS)? Default RMW in Iron

### Steps to reproduce
We have modified MoveItCpp example to use OMPL constraints, defined 2 path constraints and tested with UR5e and Panda arms. 

### Expected behaviour
The planning should not fail with the exception but complete with either solution found or not. 

### Actual behaviour
Planning fails with `Space is over constrained!` exception

medvedevigorek:
Here is some additional things we came across after a quick look into the issue: 

* `ompl::base::Constraint` takes ambient dimension as well as dimension of function output and calculates manifold dimension by subtracting function dimension from ambient dimension. As defined [here](https://github.com/ompl/ompl/blob/main/src/ompl/base/Constraint.h#L87)
* `ompl_interace::BaseConstraint` uses 3 as function dimenstion ([here](https://github.com/ros-planning/moveit2/blob/main/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/detail/ompl_constraints.h#L127
) and [here](https://github.com/ros-planning/moveit2/blob/main/moveit_planners/ompl/ompl_interface/src/detail/ompl_constraints.cpp#L132)) 
* The ambient dimension is [defined](https://github.com/ros-planning/moveit2/blob/main/moveit_planners/ompl/ompl_interface/src/detail/ompl_constraints.cpp#L376) as number of joints in Joint Model Group, so for Panda arm it's going to be 7, for UR5 - 6
* So for a single constraint the manifold dimension of Panda arm is 4 and UR5 arm is 3
[...]

PR #2775, opened 2024-03-28
Planners ompl constraints overconstrain

BerendPijnenburg:
### Description

When creating OMPL constraints the AmbientDimension often gets overconstraint as the coDimension of the constraint is set to  3 even when you constrict less than 3 dimensions.

Current implementation uses the bounds class to never give an error for a dimension if that dimension is set using -1 but still "constrains" that dimension by setting the coDimension of the ompl::base::constraint to 3.
 this causes you to be unable to use both a position and orientation constraint with <7 DOF robots as an AmbientDimension of 6 - 3 - 3 = 0 . #2614 

## Changes
- I removed the derivative function of the bounds class as the sign can already be obtained from the penalty function.
- I Consolidated the BoxConstraint and EqualityConstraint.
- I changed the coDimension of the box and orientation constraints. Now a dimension of a position constraint can be unconstrained if the respective dimension of the constraint region is less than 0 or equal to infinity.

- A dimension of an orientation constraint can be unconstrained if the absolute axis tolerance of the orientation is less than 0 or equal to infinity.

- I removed the init function, it doesn't make sense to have an uninitialized constraint
- I removed i removed the functions: calcError and calcErrorJacobian as their implementation belongs in function and jacobian

### Checklist

### DIFF
--- a/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/detail/ompl_constraints.h
+++ b/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/detail/ompl_constraints.h
@@ -68,30 +68,25 @@ class Bounds
 {
 public:
   Bounds();
-  Bounds(const std::vector<double>& lower, const std::vector<double>& upper);
   /** \brief Distance to region inside bounds
    *
    * Distance of a given value outside the bounds, zero inside the bounds.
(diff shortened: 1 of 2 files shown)


jih189:
Does the changes fix the problem?

sjahr:
@BerendPijnenburg Is this still a draft or ready for review?

Issue #2714, opened 2024-02-28
Remove topic name parameters in favor of const topic names


sjahr:
In multiple places of moveit2 (for example [moveit_cpp](https://github.com/ros-planning/moveit2/blob/5887eb0e2eb8f6803d85fac458b11122fe495449/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h#L65)), topic names can be configured with a parameter. This is not necessary because remapping topics on launch file level is trivial and the parameter API can be simplified by removing these parameters. Instead of having the topic names configurable with a ROS 2 parameters we should just use a const name in the cpp code.

yash-011:
first issue was good

Gaurang-1402:
Hi @sjahr  I have made a PR to the moveit2 repo that resolves this issue: https://github.com/ros-planning/moveit2/pull/2777

I'm curious to hear your feedback and improvise!

PR #2777, opened 2024-03-31
Removing topic name parameter configuration and replacing with const strings

Gaurang-1402:
### Description

This PR introduces changes to streamline the setup process for the `moveit_cpp::PlanningSceneMonitorOptions` and `moveit_servo::ServoParameters` by utilizing constant strings for topic names instead of parameter-based configurations. The changes ensure that topic names are now consistent and predefined. This modification also implies that users who need customized topic names will now have to use topic remapping in their launch files or at the ROS node level.

These updates are in direct response to the issue discussed in #2714 

### Checklist
In the tutorials, I see some files where changes need to be made to remove topic names from parameters, for example here: 
https://github.com/ros-planning/moveit2_tutorials/blob/main/doc/examples/moveit_cpp/config/moveit_cpp.yaml
I'm confused about the sequence of changes though since these are 2 separate repos. Do I make these changes before this PR is merged, do I make 2 simultaneous PR or do I raise an issue on that repo? Some guidance would be helpful to proceed on this point.
Done included in the commit
As this PR stands, there are no scenarios where the code fails without this PR because it is an API enhancement. Yet, if needed, I can create tests- please let me know.
Not Applicable

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -3,23 +3,27 @@
 API changes in MoveIt releases
 
 ## ROS Rolling
+
 - `lockSceneRead()` and `lockSceneWrite()` are now protected member functions, for internal use only. To lock the planning scene, use LockedPlanningSceneRO or LockedPlanningSceneRW:
+
 ```
       planning_scene_monitor::LockedPlanningSceneRO ls(planning_scene_monitor);
       moveit::core::RobotModelConstPtr model = ls->getRobotModel();
 ```
+
 - ServoServer was renamed to ServoNode
 - `CollisionObject` messages are now defined with a `Pose`, and shapes and subframes are defined relative to the object's pose. This makes it easier to place objects with subframes and multiple shapes in the scene. This causes several changes:
-    - `getFrameTransform()` now returns this pose instead of the first shape's pose.
-    - The Rviz plugin's manipulation tab now uses the object's pose instead of the shape pose to evaluate if object's are in the region of interest.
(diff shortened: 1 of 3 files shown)


moveit_ros/moveit_servo/src/servo_parameters.cpp
```diff
@@ -271,12 +264,11 @@ ServoParameters ServoParameters::get(const std::string& ns,
 
   // ROS Parameters
   parameters.use_gazebo = node_parameters->get_parameter(ns + ".use_gazebo").as_bool();
-  parameters.status_topic = node_parameters->get_parameter(ns + ".status_topic").as_string();
+  parameters.status_topic = "/servo_server/status";
```
sjahr, line 267: Can you make these constants in the anonymous namespace above like Logger and then assign the constant to the parameters member? Something like

```
namespace {
const std::string SERVO_STATUS_TOPIC = "/servo_server/status";
}
...
parameters.status_topic = SERVO_STATUS_TOPIC;
```

MIGRATION.md
```diff
@@ -44,29 +48,69 @@ API changes in MoveIt releases
 
 ## ROS Melodic
 
-- Migration to ``tf2`` API.
```
sjahr, line 47: Why are these changes necessary?

sjahr (commented):
Nice work! Just small comments and then this is good to go

sjahr:
Closing this due to inactivity. Feel free to re-open if you want to continue working on it again

PR #2790, opened 2024-04-15
CI: Allow non-interactive installation of rmw-connextdds

rhaschke:
When installing `rmw-connextdds` within a humble docker image, e.g. via rosdep, this fails, because interactive acceptance of the license is required despite common flags like `--yes`:
https://github.com/ros-planning/moveit_task_constructor/issues/558
https://github.com/ros-planning/moveit_task_constructor/actions/runs/8680352798/job/23800710624#step:4:177

Declaring the environment variable `RTI_NC_LICENSE_ACCEPTED=yes` resolves that issue:
https://github.com/ros-planning/moveit_task_constructor/actions/runs/8680352798

PR #2791, opened 2024-04-15
CI: Fix building of ikfast plugins

rhaschke:
Ignore missing authentication for Indigo packages using `--force-yes`.
Fixes https://github.com/ros-planning/moveit2/pull/2790#issuecomment-2056545176


sea-bass (approved):
Assuming other stuff unrelated to this are causing CI to fail. I see you have other PRs that possible address this too, thanks!

PR #2828, opened 2024-05-13
Fix trajectory display.

kpwelsh:
### Description

The RViz plugin for displaying planned trajectories handles elapsed time incorrectly. Probably sometime in the last few years the dt passed to `Update` was switched from "s" to "ns," and this plugin wasn't updated. 

Also, the plugin doesn't show the last pose in the trajectory. When we are on the last pose in the trajectory, we call `displaying_trajectory_message_->getWayPointDurationFromPrevious(current_state_ + 1);` to determine how long to display the pose for. This returns 0 and we accordingly display it for 0s. 

Instead, I chose to put in a constant 1s display time (scaled by the user's trajectory playback rate) at the end of each trajectory so it will pause at the end briefly before looping back to the start, or disappearing.

### Checklist

### DIFF
--- a/moveit_ros/visualization/rviz_plugin_render_tools/src/trajectory_visualization.cpp
+++ b/moveit_ros/visualization/rviz_plugin_render_tools/src/trajectory_visualization.cpp
@@ -56,7 +56,7 @@
 #include <rviz_default_plugins/robot/robot_link.hpp>
 #include <rviz_common/window_manager_interface.hpp>
 
-#include <string> 
+#include <string>
 
 using namespace std::placeholders;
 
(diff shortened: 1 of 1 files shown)


henningkayser (commented):
Thank you! Can you retarget this change to 'main'? That is what we use as development head. We can backport any bugfixes to humble or other stable branches.

riv-mjohnson:
https://github.com/moveit/moveit2/pull/3567 is relevant to this PR (I raised https://github.com/ros2/rviz/pull/1533 to make the time be in std::chrono::nanoseconds, and https://github.com/moveit/moveit2/pull/3567 updates MoveIt to take advantage of those changes)

PR #2851, opened 2024-05-24
Jazzy CI and build status

henningkayser:

PR #2855, opened 2024-05-26
TEMP: Remove ROS' octomap after 'rosdep install'

rhaschke:
Quick, temporary solution for https://github.com/moveit/moveit2/pull/2851#issuecomment-2132370683.

### DIFF
--- a/.docker/ci/Dockerfile
+++ b/.docker/ci/Dockerfile
@@ -40,6 +40,8 @@ RUN \
     rosdep update && \
     DEBIAN_FRONTEND=noninteractive \
     rosdep install -y --from-paths src --ignore-src --rosdistro ${ROS_DISTRO} --as-root=apt:false && \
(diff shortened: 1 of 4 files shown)


henningkayser:
I pulled this into #2851 for now to complete the CI fixes, even though this should really only be a short-term solution.
If possible, we should not prevent users to have the ROS package installed. I'm thinking about solving this by setting a version filter in the find_package call. Unfortunately, we can not declare the liboctomap-dev dependency in the package.xml, but at least we know that it must be available since it must be provided along with fcl.

rhaschke:
> At least we know that it (octomap) must be available since it must be provided along with fcl.

That's true for `moveit_core`, but not for `geometric_shapes`!

I explicitly filed a new PR for this change to not mangle it as part of #2851. This will allow to revert the change more easily in the future.

PR #2858, opened 2024-05-29
Copy planning scene predicates in the copy constructor

galou:
### Description

The child planning scene should behave as its parent, so the state- and motion- feasibility predicates should be copied as well.

The copy of the collision update callback is treated in a separate commit because I'm not quite sure how to do it properly. I guess that the cloned planning scene should not call the collision update callback to avoid confusion.

Often in the planning scene implementation, the child planning scene uses members of its parent when it doesn't have its own but this is not the case for the predicates so I did copy them.

### Checklist


henningkayser:
@galou thank you! Can you update your branch to main? CI should succeed now.

sjahr (approved):

PR #2861, opened 2024-06-05
Add utility functions to get limits and trajectory message

sjahr:
### Description

### Checklist

### DIFF
--- a/moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
+++ b/moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
@@ -591,7 +591,7 @@ public:
    *
    * @return std::pair<Eigen::VectorXd, Eigen::VectorXd> Containing the joint limits
    */
-  [[nodiscard]] std::pair<Eigen::VectorXd, Eigen::VectorXd> getLowerAndUpperLimits();
+  [[nodiscard]] std::pair<Eigen::VectorXd, Eigen::VectorXd> getLowerAndUpperLimits() const;
 
   /**
    * @brief Gets the pair of maximum joint velocities/accelerations for a given group. Asserts that the group contains
@@ -599,7 +599,7 @@ public:
    *
    * @return std::pair<Eigen::VectorXd, Eigen::VectorXd> Containing the velocity and acceleration limits
    */
-  [[nodiscard]] std::pair<Eigen::VectorXd, Eigen::VectorXd> getMaxVelocitiesAndAccelerationBounds();
+  [[nodiscard]] std::pair<Eigen::VectorXd, Eigen::VectorXd> getMaxVelocitiesAndAccelerationBounds() const;
 
 protected:
   /** \brief Update the variable values for the state of a group with respect to the mimic joints. This only updates
--- a/moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
+++ b/moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
@@ -587,7 +587,7 @@ public:
                                    std::vector<size_t>& joint_bijection) const;
 
   /**
-   * @brief Get the lower and upper position limits of a given joint group.
+   * @brief Get the lower and upper position limits of all active variables in the group.
    *
    * @return std::pair<Eigen::VectorXd, Eigen::VectorXd> Containing the joint limits
    */
@@ -596,7 +596,7 @@ public:
   /**
    * @brief Gets the pair of maximum joint velocities/accelerations for a given group. Asserts that the group contains
    * only single-variable joints,
-   *
+   * @details In case of asymmetric velocity or acceleration limits, this function will return the most limiting component.
    * @return std::pair<Eigen::VectorXd, Eigen::VectorXd> Containing the velocity and acceleration limits
    */
   [[nodiscard]] std::pair<Eigen::VectorXd, Eigen::VectorXd> getMaxVelocitiesAndAccelerationBounds() const;
--- a/moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
+++ b/moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
@@ -599,7 +599,7 @@ public:
    *
    * @return std::pair<Eigen::VectorXd, Eigen::VectorXd> Containing the velocity and acceleration limits
    */
-  [[nodiscard]] std::pair<Eigen::VectorXd, Eigen::VectorXd> getMaximumVelocitiesAndAccelerations();
+  [[nodiscard]] std::pair<Eigen::VectorXd, Eigen::VectorXd> getMaxVelocitiesAndAccelerationBounds();
 
 protected:
   /** \brief Update the variable values for the state of a group with respect to the mimic joints. This only updates
--- a/moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
+++ b/moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
@@ -754,5 +754,24 @@ protected:
   /** \brief The names of the default states specified for this group in the SRDF */
   std::vector<std::string> default_states_names_;
 };
+
+/**
+ * @brief Get the lower and upper position limits of a given joint group.
+ *
+ * @param group Joint model group from which the limits are read
+ * @return std::pair<Eigen::VectorXd, Eigen::VectorXd> Containing the joint limits
+ */
+[[nodiscard]] std::pair<Eigen::VectorXd, Eigen::VectorXd>
+getLowerAndUpperLimits(const moveit::core::JointModelGroup& group);
+
+/**
+ * @brief Gets the pair of maximum joint velocities/accelerations for a given group. Asserts that the group contains
+ * only single-variable joints,
+ *
+ * @param group Joint model group from which the limits are read
+ * @return std::pair<Eigen::VectorXd, Eigen::VectorXd> Containing the velocity and acceleration limits
+ */
+[[nodiscard]] std::pair<Eigen::VectorXd, Eigen::VectorXd>
+getMaximumVelocitiesAndAccelerations(const moveit::core::JointModelGroup& group);
 }  // namespace core
 }  // namespace moveit
--- a/moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
+++ b/moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
@@ -586,6 +586,21 @@ public:
   bool computeJointVariableIndices(const std::vector<std::string>& joint_names,
                                    std::vector<size_t>& joint_bijection) const;
 
+  /**
+   * @brief Get the lower and upper position limits of a given joint group.
+   *
+   * @return std::pair<Eigen::VectorXd, Eigen::VectorXd> Containing the joint limits
+   */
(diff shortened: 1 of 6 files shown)


moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -1310,4 +1310,33 @@ double TimeOptimalTrajectoryGeneration::verifyScalingFactor(const double request
...
   return scaling_factor;
 }
+
+trajectory_msgs::msg::JointTrajectory createTrajectoryMessage(const std::vector<std::string>& joint_names,
+                                                              const trajectory_processing::Trajectory& trajectory,
+                                                              const int sampling_rate)
+{
+  trajectory_msgs::msg::JointTrajectory trajectory_msg;
+  trajectory_msg.joint_names = joint_names;
+  assert(sampling_rate > 0);
```
Abishalini, line 1320: Should we add an error message here in case the condition fails here?

moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
```diff
@@ -189,6 +189,13 @@ class Trajectory
   mutable std::list<TrajectoryStep>::const_iterator cached_trajectory_segment_;
 };
 
+/**
+ * @brief Converts a `trajectory_processing::Trajectory` into a `JointTrajectory` message with a given sampling rate.
+ */
+[[nodiscard]] trajectory_msgs::msg::JointTrajectory
+createTrajectoryMessage(const std::vector<std::string>& joint_names,
+                        const trajectory_processing::Trajectory& trajectory, const int sampling_rate);
```
Abishalini, line 197: Can this free function be moved to `trajectory_tool.h` file so that its easier to find and not get lost among other class declarations.

moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
```diff
@@ -754,5 +754,24 @@ class JointModelGroup
...
+
+/**
+ * @brief Gets the pair of maximum joint velocities/accelerations for a given group. Asserts that the group contains
+ * only single-variable joints,
+ *
+ * @param group Joint model group from which the limits are read
+ * @return std::pair<Eigen::VectorXd, Eigen::VectorXd> Containing the velocity and acceleration limits
+ */
+[[nodiscard]] std::pair<Eigen::VectorXd, Eigen::VectorXd>
+getMaximumVelocitiesAndAccelerations(const moveit::core::JointModelGroup& group);
```
Abishalini, line 775: Would adding these functions to the `JointModelGroup` as member functions make sense?

marioprats: I think that makes sense, to be more consistent with the current design / implementation.

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -831,5 +831,41 @@ bool JointModelGroup::isValidVelocityMove(const double* from_joint_pose, const d
...
   return true;
 }
+
+std::pair<Eigen::VectorXd, Eigen::VectorXd> getLowerAndUpperLimits(const moveit::core::JointModelGroup& group)
+{
+  // Get the group joints lower/upper position limits.
+  Eigen::VectorXd lower_limits(group.getActiveVariableCount());
+  Eigen::VectorXd upper_limits(group.getActiveVariableCount());
+  const moveit::core::JointBoundsVector& group_bounds = group.getActiveJointModelsBounds();
+  int joint_index = 0;
```
marioprats, line 841: Would `variable_index` be more correct?

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -831,5 +831,41 @@ bool JointModelGroup::isValidVelocityMove(const double* from_joint_pose, const d
...
+{
+  const std::size_t n_dofs = group.getActiveVariableCount();
+  Eigen::VectorXd max_joint_velocities = Eigen::VectorXd::Constant(n_dofs, 0.0);
+  Eigen::VectorXd max_joint_accelerations = Eigen::VectorXd::Constant(n_dofs, 0.0);
+  const moveit::core::JointBoundsVector& bounds = group.getActiveJointModelsBounds();
+  assert(bounds.size() == n_dofs);
+  for (std::size_t i = 0; i < bounds.size(); ++i)
+  {
+    assert(bounds[i]->size() == 1);  // single-variable joints.
+    max_joint_velocities[i] = std::min(-bounds[i]->at(0).min_velocity_, bounds[i]->at(0).max_velocity_);
```
marioprats, line 865: If max. velocities or accelerations are not symmetric (e.g. [-1, 2]), this function will return the most limiting element (i.e. 1). Maybe worth clarifying that in the doc in the header, and maybe rename the function to `getMostLimitingSymmetricVelocitiesAndAccelerations`, which I realize it's probably too long...

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -831,5 +831,41 @@ bool JointModelGroup::isValidVelocityMove(const double* from_joint_pose, const d
...
+}
+
+std::pair<Eigen::VectorXd, Eigen::VectorXd>
+getMaximumVelocitiesAndAccelerations(const moveit::core::JointModelGroup& group)
+{
+  const std::size_t n_dofs = group.getActiveVariableCount();
+  Eigen::VectorXd max_joint_velocities = Eigen::VectorXd::Constant(n_dofs, 0.0);
+  Eigen::VectorXd max_joint_accelerations = Eigen::VectorXd::Constant(n_dofs, 0.0);
+  const moveit::core::JointBoundsVector& bounds = group.getActiveJointModelsBounds();
+  assert(bounds.size() == n_dofs);
```
marioprats, line 861: I think the assert here would be ok if this was an internal function and something else guarantees this condition is true.
But since this is a user-facing function, we probably want to return a proper error, or just implement the case for multi-dof joints.
Otherwise it can happen for instance that people are using this function without issues in Release mode where asserts don't have effect, and suddenly start getting an exception in Debug mode.

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
marioprats: I think so, since this is a public function that anyone can call, we should check for invalid inputs and return a proper error.
In Release mode the assert wouldn't run and we could have undefined behavior below because of a division by zero.

moveit_core/robot_model/src/joint_model_group.cpp
sjahr: I like joint index better

moveit_core/robot_model/src/joint_model_group.cpp
sjahr: I named it getMaxVelocitiesAndAccelerationBounds what do you think about that?

moveit_core/robot_model/src/joint_model_group.cpp
marioprats: The reason why I'm proposing `variable_index` is because a joint can have multiple variables and this function iterates over those variables, not over joints.

moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
```diff
@@ -586,6 +586,21 @@ class JointModelGroup
   bool computeJointVariableIndices(const std::vector<std::string>& joint_names,
                                    std::vector<size_t>& joint_bijection) const;
 
+  /**
+   * @brief Get the lower and upper position limits of a given joint group.
```
marioprats, line 590:
```suggestion
   * @brief Get the lower and upper position limits of all active variables in the group.
```

moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
```diff
@@ -586,6 +586,21 @@ class JointModelGroup
...
+   *
+   * @return std::pair<Eigen::VectorXd, Eigen::VectorXd> Containing the joint limits
```
marioprats, line 592:
```suggestion
   * @return std::pair<Eigen::VectorXd, Eigen::VectorXd> Containing the lower and upper joint limits for all active variables.
```

moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
```diff
@@ -586,6 +586,21 @@ class JointModelGroup
...
+   */
+  [[nodiscard]] std::pair<Eigen::VectorXd, Eigen::VectorXd> getLowerAndUpperLimits();
+
+  /**
+   * @brief Gets the pair of maximum joint velocities/accelerations for a given group. Asserts that the group contains
+   * only single-variable joints,
+   *
```
marioprats, line 599:
```suggestion
   * @details In case of asymmetric velocity or acceleration limits, this function will return the most limiting component.
```

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -831,5 +831,52 @@ bool JointModelGroup::isValidVelocityMove(const double* from_joint_pose, const d
...
+
+std::pair<Eigen::VectorXd, Eigen::VectorXd> JointModelGroup::getMaxVelocitiesAndAccelerationBounds()
+{
+  Eigen::VectorXd max_joint_velocities = Eigen::VectorXd::Constant(active_variable_count_, 0.0);
+  Eigen::VectorXd max_joint_accelerations = Eigen::VectorXd::Constant(active_variable_count_, 0.0);
+  if (active_joint_models_bounds_.size() != active_variable_count_)
+  {
+    // TODO(sjahr) Support multiple variables
+    RCLCPP_ERROR(getLogger(), "Number of active joint models does not match number of active joint model bounds. "
+                              "Returning bound vectors.with zeros");
```
marioprats, line 861:
```suggestion
                              "Returning bound vectors with zeros");
```

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -831,5 +831,52 @@ bool JointModelGroup::isValidVelocityMove(const double* from_joint_pose, const d
...
+    return { max_joint_velocities, max_joint_accelerations };
+  }
+  for (std::size_t i = 0; i < active_joint_models_bounds_.size(); ++i)
+  {
+    if (active_joint_models_bounds_[i]->size() == 1)
```
marioprats, line 866: Shouldn't this be != 1?

moveit_core/robot_model/src/joint_model_group.cpp
```diff
@@ -831,5 +831,52 @@ bool JointModelGroup::isValidVelocityMove(const double* from_joint_pose, const d
...
+    {
+      RCLCPP_ERROR(
+          getLogger(),
+          "Joint '%li' has '%li' bounds and only joints with one bounds set are supported by this function. In "
+          "the returned vector this element will be 0.",
```
marioprats, line 871: Is this true? will it be zero or will it be just not included? Maybe catch this case with the check above and return all zeros?

moveit_core/robot_model/src/joint_model_group.cpp
sjahr: I'll rename it

marioprats (approved):

Abishalini (approved):

Issue #2859, opened 2024-06-03
Incorrect Implementations inside MoveGroupInterface for waiting response from server


CihatAltiparmak:
### Description

Hello everyone,
Firstly, thanks for this great project.  I want to inform you about some bad practices in move_group_interface . There occur some functionality problem in  
the part of `plan` method of MoveGroupInterface
https://github.com/moveit/moveit2/blob/672e0ecd48f471fe05057f81e6233272a692a156/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp#L691-L710 ,

the part of `move` method of MoveGroupInterface
https://github.com/moveit/moveit2/blob/672e0ecd48f471fe05057f81e6233272a692a156/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp#L767-L782

and maybe this part of `execute` method of MoveGroupInterface
https://github.com/moveit/moveit2/blob/672e0ecd48f471fe05057f81e6233272a692a156/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp#L831-L850

I also want to make a critic about node spinning inside separate thread in move_group_interface. For users who don't know node starts to spin in separate thread when move group interface is created, this logic maybe can be problem. I think we need better implementation for this part as well.

https://github.com/moveit/moveit2/blob/672e0ecd48f471fe05057f81e6233272a692a156/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp#L117-L120

Problem is that there is a possibility that request is sent by client but is not received forever by server, especially move_group_server. This can bring about move_group_interface to get stuck inside while loop and plan method not to finish its own job.

In addition, these implementations normally work for rmw_fastrtps and rmw_cyclonedds, but get stuck with rmw_zenoh.

If you want, i can deal with it and can try to come up with better implementation.

### Your environment
* ROS Distro: Rolling
* OS Version: Ubuntu 22.04
* Source Build
* Branch : main
* RMW : rmw_zenoh

### Steps to reproduce
Firstly, run zenoh router (be sure that [rmw_zenoh_cpp](https://github.com/ros2/rmw_zenoh) is built at your workspace)

```shell
export RMW_IMPLEMENTATION=rmw_zenoh_cpp
ros2 run rmw_zenoh_cpp rmw_zenohd
```

[...]

PR #2863, opened 2024-06-07
Reimplemented server response-waiting functionalities inside the methods of move_group_interface

CihatAltiparmak:
### Description

Firstly, this pull request is created in order to solve this issue [#2859](tg://search_hashtag?hashtag=2859) . I still have to apply some test such that this PR works for move2_tutorials and so on. I already have tested this PR on rviz but i need to try this part more to see if any corruption occurs for this PR.

I haven't tested computeCartesianMethod, getPlannerParams, setPlannerParams(acutally there is a posibility that i tested getPlannerParam and setPlannerParam implicitly while testing rviz and getInterfaceDescriptions, or else rviz gets stuck related to node spining)

I have removed callback_thread_ because it's not necessary to yield CPU continuously during move_group_interface to work. Instead of this, i just made a implementation for spinning only when client send a request from move_group_interface. I also have added private node for move_group_interface to handle its client jobs.

I also have to give anonymous node name to pnode_ . I will work on this as well.

I need some review from variable naming to logic. 

### Checklist


CihatAltiparmak:
Hello, it seems ready, but not ready for merging, i need a second eye. I only spin now when the methods like plan, execute etc. are called instead of spinning during lifetime of MoveGroupInterface instance. But I want to ask a question.

* With my modification, This instance(move_group_interface) cannot be used concurrently like this because of both of two threads will spin separately. In old implementation, it seems below pseudo code works. Should we also support to be able to implement codes like below?
```
move_group_interface = MoveGroupInterface(...)
thread1[move_group_interface] = {move_group_interface.plan()}
thread2[move_group_interface] = {move_group_interface.plan()}

thread1.start()
thread2.start()
```

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
```diff
@@ -833,13 +876,24 @@ class MoveGroupInterface::MoveGroupInterfaceImpl
     goal.controller_names = controllers;
 
     auto goal_handle_future = execute_action_client_->async_send_goal(goal, send_goal_opts);
+    if (callback_executor_.spin_until_future_complete(goal_handle_future,
+                                                      server_timeout.to_chrono<std::chrono::duration<double>>()) !=
+        rclcpp::FutureReturnCode::SUCCESS)
+    {
+      RCLCPP_ERROR(logger_, "MoveGroupInterface::execute() get sent call failed :(");
```
sjahr, line 883: You should make this error message more descriptive. Try to explain what failed and why

sjahr: You'll probably need to update all error messages like this

sjahr (commented):
Thanks! How have you tested this? With the move_group tutorial?

Issue #2846, opened 2024-05-23
Jazzy Support and Release


henningkayser:
[ROS 2 Jazzy is being released these days](https://docs.ros.org/en/jazzy/Releases/Release-Jazzy-Jalisco.html#release-timeline), and we are working on adding support for it in MoveIt, pushing for having the Source (and soon Debian) release version completed. The plan is to enable Noble CI jobs in all repositories, testing Jazzy along with Rolling, and applying the necessary remaining fixes. Thanks to @rhaschke a first step to fixing Noble support has been already provided with https://github.com/moveit/moveit2/pull/2793. The remaining steps for completion are:

- [x] Fix MoveIt docker images - new Jazzy and Rolling images are now available
- [x] Add Jazzy build job to MoveIt and dependencies https://github.com/moveit/moveit2/pull/2851
- [ ] Fix dependencies
    - [ ] Tutorials are failing, mainly because of a transitive dependency to the outdated gazebo-ros2-control package, see  https://github.com/Kinovarobotics/ros2_kortex/pull/159
    - [x] Octomap source build failed:  https://github.com/moveit/moveit2/actions/runs/9227203194/job/25388623434
           `CMake Error in CMakeLists.txt:
    IMPORTED_LOCATION or IMPORTED_IMPLIB not set for imported target "octomap"
    configuration "Debug".`
    - [x] Octomap ABI conflict: liboctomap-dev 1.9.7 (from fcl) and ros-*-octomap 1.10.0 (see https://github.com/moveit/moveit2/pull/2851#issuecomment-2132370683), continued in #2862
- [x] Release geometric_shapes: https://github.com/moveit/geometric_shapes/pull/242 
- [x] Release moveit_msgs: https://github.com/moveit/moveit_msgs/pull/178
- [x] Update build instructions on website
- [x] Cut release for Rolling and Jazzy - most binaries are already available in Jazzy, it's only blocked because of the failing moveit_ros_benchmarks build. https://github.com/moveit/moveit2/pull/2867

rhaschke:
What fixes do you have in mind for the MoveIt docker images, @henningkayser?
With #2793, I thought I have fixed all open issues. I just noticed that - due to the ros-planning org migration, dockerhub doesn't yet receive image updates. That's tackled in #2847.

henningkayser:
Jazzy: https://github.com/ros/rosdistro/pull/41645
Rolling: https://github.com/ros/rosdistro/pull/41643

[...]

PR #2867, opened 2024-06-13
Release 2.10.0 (Rolling, Jazzy) - DON'T MERGE

henningkayser:
Reopen of #2846  - don't merge, only fast forward

The tutorials are still expected to fail unfortunately. We will have to fix the kinova runtime dependency for that. All the other jobs should succeed.


.github/workflows/ci.yaml
```diff
@@ -18,16 +18,18 @@ jobs:
       matrix:
         env:
           - IMAGE: rolling-ci
-            CCOV: true
+            CCOV: false # Disabled: https://github.com/moveit/moveit2/issues/2866
```
rhaschke, line 21: Please comment out instead of setting `CCOV: false`. As the image is identical to the next job, there is no benefit in running the job here as well.

moveit2.repos
```diff
@@ -1,14 +1,14 @@
 repositories:
   # Keep moveit_* repos here because they are released with moveit
+  geometric_shapes:
+    type: git
+    url: https://github.com/moveit/geometric_shapes.git
+    version: ros2
```
rhaschke, line 6: Isn't geometric_shapes released already?

rhaschke (commented):
Looks good. I have a few remarks only:

moveit2.repos
henningkayser: yes, but on rolling it makes sense to always test against the latest version to detect upstream breakages immediately. The build farm already takes care of testing against binaries. All source install instructions suggest installing the source dependencies in the repos files as well. The scenario of building moveit with binaries only is not recommended or supported.

.github/workflows/ci.yaml
henningkayser: Hm, don't you think there is an advantage testing gcc separately from clang+clang-tidy? I assume that most users run exactly this setup, so I would prefer keeping it in.

moveit2.repos
rhaschke: Ok. Agreed.

.github/workflows/ci.yaml
rhaschke: Ok. Sure.

PR #2874, opened 2024-06-19
added possibility to optionally optimize time during compute cartesian path

Duke194:
### Description

Added the ability to opt out of time optimization during Cartesian path computation.

This change requires a corresponding update in https://github.com/moveit/moveit_msgs/pull/179, which is related to this PR.

This feature is necessary for our project to plan trajectories with multiple waypoints and turning points. We have tested the planning with and without this feature. Without this feature, time optimization removes some essential points, resulting in incorrect trajectories. With this feature enabled, the trajectory is planned correctly without any issues.

### DIFF
--- a/moveit_ros/move_group/src/default_capabilities/cartesian_path_service_capability.cpp
+++ b/moveit_ros/move_group/src/default_capabilities/cartesian_path_service_capability.cpp
@@ -176,12 +176,18 @@ bool MoveGroupCartesianPathService::computeService(
 
           // time trajectory
           // \todo optionally compute timing to move the eef with constant speed
-          trajectory_processing::TimeOptimalTrajectoryGeneration time_param;
-          time_param.computeTimeStamps(rt, 1.0);
+          if (req->time_optimize)
+          {
+            auto orig_point_count = rt.getWayPointCount();
+            trajectory_processing::TimeOptimalTrajectoryGeneration time_param;
+            time_param.computeTimeStamps(rt, 1.0);
+            RCLCPP_INFO(LOGGER, "Optimized trajectory with %u points to %u points", (unsigned int)orig_point_count,
+                        (unsigned int)rt.getWayPointCount());
+          }
 
           rt.getRobotTrajectoryMsg(res->solution);
           RCLCPP_INFO(LOGGER, "Computed Cartesian path with %u points (followed %lf%% of requested trajectory)",
-                      (unsigned int)traj.size(), res->fraction * 100.0);
+                      (unsigned int)rt.getWayPointCount(), res->fraction * 100.0);
           if (display_computed_paths_ && rt.getWayPointCount() > 0)
           {
             moveit_msgs::msg::DisplayTrajectory disp;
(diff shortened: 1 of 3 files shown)


Duke194:
Ping.
Could someone please review/merge this?

Duke194:
Someone there ? ping

moveit_ros/move_group/src/default_capabilities/cartesian_path_service_capability.cpp
```diff
@@ -176,12 +176,18 @@ bool MoveGroupCartesianPathService::computeService(
 
           // time trajectory
           // \todo optionally compute timing to move the eef with constant speed
-          trajectory_processing::TimeOptimalTrajectoryGeneration time_param;
-          time_param.computeTimeStamps(rt, 1.0);
+          if (req->time_optimize)
+          {
+            auto orig_point_count = rt.getWayPointCount();
```
sjahr, line 181:
```suggestion
            const auto orig_point_count = rt.getWayPointCount();
```

moveit_ros/move_group/src/default_capabilities/cartesian_path_service_capability.cpp
```diff
@@ -176,12 +176,18 @@ bool MoveGroupCartesianPathService::computeService(
...
+            trajectory_processing::TimeOptimalTrajectoryGeneration time_param;
+            time_param.computeTimeStamps(rt, 1.0);
+            RCLCPP_INFO(LOGGER, "Optimized trajectory with %u points to %u points", (unsigned int)orig_point_count,
```
sjahr, line 184:
```suggestion
            RCLCPP_DEBUG(LOGGER, "Optimized trajectory with %u points to %u points", (unsigned int)orig_point_count,
```

sjahr (commented):
This makes sense to me, thanks! Rather than calling it time_optimize, I'd propose calling the option `compute_time_stamps` or `generate_trajectory` to highlight that if this option is false, the output of the capability is just a path an not a trajectory. What do you think?

rhaschke:
Instead of using a simple boolean flag, did you considered to specify the time parameterization plugin via a string?
This would allow to change the time parameterization method as well as disable time parameterization (if empty).

Duke194:
> This makes sense to me, thanks! Rather than calling it time_optimize, I'd propose calling the option `compute_time_stamps` or `generate_trajectory` to highlight that if this option is false, the output of the capability is just a path an not a trajectory. What do you think?


I see your point. But I think we should then go for `generate_trajectory`, I think it fits best.

Duke194:
> Instead of using a simple boolean flag, did you considered to specify the time parameterization plugin via a string? This would allow to change the time parameterization method as well as disable time parameterization (if empty).

I didn't considered that because for our specific use case it is sufficient to get a path out of this capability. We do the time optimization aka. trajectory generation afterwards ourselves. I do not know if the parametrization needs to be configurable. Probably depends on the use case.

sjahr:
@Duke194 Just realized your targeting the humble branch with your PR. Do you mind changing it to main? We'll backport it to humble with a github bot afterwards immediately. I think this is ready to be merged once you addressed my review

rhaschke:
I suggest implementing the more generic selection of time parameterization suggested in https://github.com/moveit/moveit2/pull/2874#issuecomment-2210411652.

Duke194:
> @Duke194 Just realized your targeting the humble branch with your PR. Do you mind changing it to main? We'll backport it to humble with a github bot afterwards immediately. I think this is ready to be merged once you addressed my review

I'll do it on main this week.

PR #2875, opened 2024-06-20
Added 'const' to setRobotStateFromPoint declaration parameters

ybnelan:
issue: https://github.com/moveit/moveit2/issues/879
I made the integer parameter in the setRobotStateFromPoint function declaration a constant.


sjahr:
Thanks for the PR but I think you'd need to update the function definition too in order to make this work. Given the size of the PR and how relevant this change is, I'll close the PR for now. If you'd like to contribute to moveit2 in a meaningful way, I'd recommend picking up one of the [good first issues](https://github.com/moveit/moveit2/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22)

PR #2876, opened 2024-06-21
PSM: Correctly handle full planning scene message

sjahr:
[Port #3610](https://github.com/moveit/moveit/pull/3610/files)

### Description

### Checklist

### DIFF
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -1262,6 +1262,7 @@ bool PlanningScene::setPlanningSceneDiffMsg(const moveit_msgs::msg::PlanningScen
 
 bool PlanningScene::setPlanningSceneMsg(const moveit_msgs::msg::PlanningScene& scene_msg)
 {
+  assert(scene_msg.is_diff == false);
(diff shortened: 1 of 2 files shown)


rhaschke (approved):

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -1262,6 +1262,7 @@ bool PlanningScene::setPlanningSceneDiffMsg(const moveit_msgs::msg::PlanningScen
 
 bool PlanningScene::setPlanningSceneMsg(const moveit_msgs::msg::PlanningScene& scene_msg)
 {
+  assert(scene_msg.is_diff == false);
```
marioprats, line 1265: Is this method public-facing? if so, do we want to throw an assertion here?
Since this function returns a `bool` already, we could also return `false` and print out an error message, which would also work in release builds

rhaschke: The assertion is just a reminder, that this method should not be called with planning scene diffs. If it is, it is probably a logical error in the code, causing some issues down the road.

PR #2886, opened 2024-06-26
Replaced gazebo_ros_control by gazebo_ros2_control

vonunwerth:
### Description

When using the <exec-depend> tags generated in the package.xml for the use of Gazebo Classic, rosdep failed because it could not find the package `gazebo_ros_control`.

`ERROR: the following packages/stacks could not have their rosdep keys resolved
to system dependencies:
...: Cannot locate rosdep definition for [gazebo_ros_control]
`

I changed the template for the package.xml to include [gazebo_ros2_control](https://github.com/ros-controls/gazebo_ros2_control)

It could also be useful to include a second commented paragraph containing `gz_ros2_control` for Gazebo.

### Checklist


scastro-bdai (commented):
Gazebo classic is no longer supported in Jazzy and onwards... so if anything, this should be `gz_ros2_control`. But we'd need to know if Gazebo Sim even works with these tools...

sjahr:
@vonunwerth Thanks for that PR, I'll close this for now because of Sebastian's comment + moveit2 shouldn't depend on it anyways

Issue #541, opened 2021-07-07
joint limit loading and overwriting behavior 


brianyany:
A few things: 
1. joint_limits.yaml not loaded in the demo apps and not clear how exactly would one go about loading it. 
2. When joint_limits.yaml is not loaded, in robot_model_loader.cpp, the configure function would attempts to load the limits but fail (which means there's no pos, vel, or acc limit). Then the call at line 226 to setVariableBounds would overwrite the joint limits read from the urdf file. This behavior seems strange: shouldn't the joint_model retain the limits read from the urdf file? 
3. When joint_limits are not properly loaded and close to singularity, position/velocity value would explode..

JafarAbdi:
> 1. joint_limits.yaml not loaded in the demo apps and not clear how exactly would one go about loading it.

This [PR](https://github.com/ros-planning/moveit2/pull/544) add how to load the joint limits

> 2. When joint_limits.yaml is not loaded, in robot_model_loader.cpp, the configure function would attempts to load the limits but fail (which means there's no pos, vel, or acc limit). Then the call at line 226 to setVariableBounds would overwrite the joint limits read from the urdf file. This behavior seems strange: shouldn't the joint_model retain the limits read from the urdf file?

The joint_limit is initially set to the values from the urdf in [L125](https://github.com/ros-planning/moveit2/blob/main/moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp#L125), and will only be overridden if `joint_limits.yaml` is loaded

> 3. When joint_limits are not properly loaded and close to singularity, position/velocity value would explode..

Do you mind sharing a reproducible example .?

JafarAbdi:
> Have to disagree with the point you made about 2 since I am observing the limit values got overwritten without loading joint_limits.yaml. If you add a print for bounds.velocity_bounded_ at line 765 of servo_calcs.cpp. You will notice that bounds.velocity_bounded_ evaluates to false. However, if you comment out the call to setVariableBounds at line 225 of robot_model_loader.cpp, bounds.velocity_bounded_ evaluates to true in servo_calcs.cpp. This means the joint limits from urdf got overwritten by the unsuccessful attempt to load limit values from joint_limits.yaml. (Running one of the servo demos)

You're right, can you test this [PR](https://github.com/ros-planning/moveit2/pull/546) and see if it fixes the bug for you .?

brianyany:
[...]

PR #2910, opened 2024-07-15
Fix constrained-based planning / PoseModelStateSpace

rhaschke:
This is a forward port of https://github.com/moveit/moveit/pull/3615

* Fix constrained-based planning

Constrained-based planning enables the PoseModelStateSpace by default, meaning that state samples are created by sampling in SE(3) and then performing IK. This state space frequently caused issues, eventually leading to the introduction of the ROS parameter enforce_joint_model_state_space in PR #541 as a work-around.

A more detailed analysis shows that the state space's interpolation approach is the culprit - performing interpolation in Cartesian space. If the two interpolation ends are close in Cartesian but distant in joint space, the interpolated state (obtained from IK) can be far away from both start and end (particularly for redundant arms), while its Cartesian pose is naturally close to both interpolation states. This will lead to consecutive waypoints being close in Cartesian space but distant in joint space. If such a trajectory is executed, the controller will interpolate in joint space, leading to a large and unchecked motion.

To avoid this issue, interpolation should be performed in joint space during planning as well. The only drawback of this change is that trajectories are less linear in Cartesian space. However, as the trajectory is random-sampled anyway, this shouldn't be a big issue.

* PoseModelStateSpace: Use joint-space distance

Configurations may be similar in Cartesian EE pose, but very far apart in joint space. What actually matters when connecting states is the joint-space distance.

### DIFF
--- a/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/parameterization/work_space/pose_model_state_space.h
+++ b/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/parameterization/work_space/pose_model_state_space.h
@@ -145,6 +145,5 @@ private:
   };
 
   std::vector<PoseComponent> poses_;
-  double jump_factor_;
 };
 }  // namespace ompl_interface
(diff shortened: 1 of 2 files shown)


sjahr (approved):
Thanks for porting this over!

PR #2916, opened 2024-07-17
New implementation for computeCartesianPath()

rhaschke:
This is a port of https://github.com/moveit/moveit/pull/3618

The Cartesian interpolator uses a JumpThreshold parameter to detect and reject a switch between IK solution branches leading to larger joint-space jumps. However, this approach is rather fragile: It is difficult to find a suitable threshold parameter and the approach requires a long-enough trajectory to work (at least 10 waypoints).
Particularly if the last requirement is not met, joint-space jumps can still occur unnoticed leading to a dangerous execution of an unchecked, large-volume trajectory.

This PR proposes a fundamentally different way to check for joint-space jumps, which works robustly and locally for any two adjacent waypoints: If the two waypoints are distant in joint-space, their half-way interpolation state will typically lead to a very strong deviation from the straight-line Cartesian path between them. This PR measures this deviation and rejects a trajectory if it exceeds a "precision" threshold.
Actually, the approach (recursively) introduces more waypoints in between to improve the precision of the linear Cartesian path. Thus, the method doesn't need to be provided with a good estimate of the `MaxEEFStep` parameter, it adjusts the step size dynamically to meet the given precision requirement. (This statement only holds if IK succeeds at all. In order to evaluate IK on a fine-grained grid and detect obstacles, one still needs the `MaxEEFStep` parameter.)

The following plots illustrate the new behavior. They all use the extremely coarse `MaxEEFStep` > 1m, resulting in a single initial step and then adjust the precision to introduce additional waypoints:
| 1 step, 0.3m precision | 1 step, collision not detected! |
|---|---|
| 1 step | collision |

| 1 + 1 steps, 0.1m precision | 1 + 3 steps, 0.025m precision |
|---|---|
| 2 steps | 4 steps |

I kept but deprecated the old JumpThreshold-based methods. I'm not sure how to best migrate the MoveGroup interface. To expose the new interface, we would need to introduce the `CartesianPrecision` argument as a message.
For now, I just dropped the JumpThreshold argument and resort to the default precision, which is 1mm. Given a reasonable `MaxEEFStep`, this should perfectly resolve the jump issues. I think, fiddling with the new parameter shouldn't be necessary in most cases.


moveit_core/robot_state/include/moveit/robot_state/cartesian_interpolator.h
```diff
@@ -44,6 +44,14 @@ namespace moveit
 {
 namespace core
 {
+/** Struct defining linear and rotational precision */
+struct CartesianPrecision
+{
+  double translational = 0.001;
+  double rotational = 0.01;
+  double max_resolution = 1e-5;
```
sea-bass, line 52: Can you add brief docstrings to each of these, which include units?

I can make sense of `translational` and `rotational`, but without reading the rest of the code I don't know what `max_resolution` means.

moveit_core/robot_state/include/moveit/robot_state/cartesian_interpolator.h
```diff
@@ -144,6 +152,75 @@ class CartesianInterpolator
...
+     The resulting joint values are stored in the vector \e traj, one by one. The interpolation distance in
+     Cartesian space between consecutive points on the resulting path is specified in the \e MaxEEFStep struct which
+     provides two fields: translation and rotation. If a \e validCallback is specified, this is passed to the internal
+     call to setFromIK(). In case of IK failure, the computation of the path stops and the value returned corresponds to
+     the distance that was achieved and for which corresponding states were added to the path.
+
+     The struct CartesianPrecision specifies the precision to which the path should follow the Cartesian straight line.
+     If the deviation at the mid point of two consecutive waypoints is larger than the specified precision, another waypoint
+     will be inserted at that mid point. The precision is specified separately for translation and rotation.
+     The maximal resolution to consider (in percentage of the total path length) is specified by max_resolution.
```
sea-bass, line 169: So is it percentage or a fraction? Given its default value (`1e-5`) I assume fraction?

moveit_core/robot_state/include/moveit/robot_state/cartesian_interpolator.h
```diff
@@ -144,6 +152,75 @@ class CartesianInterpolator
...
+      const RobotState* start_state, const JointModelGroup* group, std::vector<std::shared_ptr<RobotState>>& traj,
+      const LinkModel* link, const Eigen::Vector3d& translation, bool global_reference_frame,
+      const MaxEEFStep& max_step, const CartesianPrecision& precision,
+      const GroupStateValidityCallbackFn& validCallback = GroupStateValidityCallbackFn(),
+      const kinematics::KinematicsQueryOptions& options = kinematics::KinematicsQueryOptions(),
+      const kinematics::KinematicsBase::IKCostFn& cost_function = kinematics::KinematicsBase::IKCostFn());
+
+  /** \brief Compute the sequence of joint values that correspond to a straight Cartesian path, for a particular link.
+
+     In contrast to the previous function, the translation vector is specified as a (unit) direction vector and
```
sea-bass, line 181: Does "(unit)" mean that no matter what vector you pass in, it gets normalized first? Maybe worth saying.

moveit_core/robot_state/include/moveit/robot_state/cartesian_interpolator.h
```diff
@@ -144,6 +152,75 @@ class CartesianInterpolator
...
+                                precision, validCallback, options, cost_function);
+  }
+
+  /** \brief Compute the sequence of joint values that correspond to a straight Cartesian path, for a particular frame.
+
+     In contrast to the previous function, the Cartesian path is specified as a target frame to be reached (\e target)
+     for a virtual frame attached to the robot \e link with the given \e link_offset.
+     The target frame is assumed to be specified either w.r.t. to the global reference frame or the virtual link frame.
+     This function returns the fraction (0..1) of path that was achieved. All other comments from the previous function apply. */
+  static Percentage computeCartesianPath(
```
sea-bass, line 201: This one may be legacy, but again confusing that the comment says we return a fraction, but the output data type is called `Percentage`.

moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
```diff
@@ -756,9 +756,16 @@ class MOVEIT_MOVE_GROUP_INTERFACE_EXPORT MoveGroupInterface
       Return a value that is between 0.0 and 1.0 indicating the fraction of the path achieved as described by the
      waypoints.
       Return -1.0 in case of error. */
+  [[deprecated("Drop jump_threshold")]] double  //
```
sea-bass, line 759: Replace with your better worded deprecation message from before:

```
[[deprecated("Replace JumpThreshold with CartesianPrecision")]]
```

moveit_core/robot_state/include/moveit/robot_state/cartesian_interpolator.h
rhaschke: No, the direction vector is simply multiplied by the distance to yield the final translation vector.
As this can be directly seen from the inline implementation, I don't see the need for explaining this in a comment.

moveit_core/robot_state/include/moveit/robot_state/cartesian_interpolator.h
rhaschke: Fractions are measured in percentage.

moveit_core/robot_state/include/moveit/robot_state/cartesian_interpolator.h
rhaschke: If you mean that fraction and percentage are different by a factor of 100, you are right: a fraction (values between 0 and 1) is meant here.

moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
rhaschke: There is no CartesianPrecision argument needed. Hence, a different message.

moveit_core/robot_state/include/moveit/robot_state/cartesian_interpolator.h
sea-bass: Exactly what I meant!

moveit_core/robot_state/include/moveit/robot_state/cartesian_interpolator.h
sea-bass: 🤷🏻

moveit_core/robot_state/include/moveit/robot_state/cartesian_interpolator.h
sea-bass: This is one of those "if you know what you're doing, you will pass in a unit vector" situations. But not necessary to change, I guess.

rhaschke: If you really know what you are doing, it is completely fine to pass a non-unit vector here ;-)
In that case the distance argument, just doesn't determine the travelled distance.

rhaschke: Usually, you want to pass a unit vector.

sjahr (approved):
Cool, looks good thank you!

DaniGarciaLopez:
Thanks for the contribution @rhaschke! Any plans to backport this to humble?

rhaschke:
No. This PR involves a big API change.

PR #2922, opened 2024-07-19
Fix deprecation warning

rhaschke:
Address the [newly failing rolling build](https://github.com/moveit/moveit2/actions/runs/10008006352/job/27663837261). They deprecated the use of `rmw_qos_profile_t` in favour of `rclcpp:QoS`.

Unfortunately, fixing the rolling build now invalidates the humble and jazzy builds.
Is there a compile-time version check available to handle this?


sjahr:
@rhaschke Thanks for the fix! Tyler solved it with pre-processor directives for the move_group_interface [here](https://github.com/moveit/moveit2/commit/44821258a71c54a1b2381d8e664dfcdac79c40fa). Maybe that would fix it here too?

sjahr (approved):
Awesome, thanks!

PR #2943, opened 2024-08-01
Deduplicate joint trajectory points in Pilz Move Group Sequence capability

sea-bass:
### Description

I was unable to test https://github.com/moveit/moveit2_tutorials/pull/917 successfully on Rolling/main because there was an issue with sending duplicate time points to joint trajectory controllers with the Pilz move group sequence capability.

This PR fixes that, but I am not entirely sure this is the most efficient fix, as it would be preferable to package the trajectory correctly in the first place. But... this works.

### Checklist

Closes https://github.com/moveit/moveit2/issues/2741


stephanie-eng (approved):
I don't entirely agree that we should fix this at this level, but I'd be loathe to let perfect be the enemy of the good here, and a band aid fix is better than broken.

Could you add a comment or a TODO note specifically calling this out, though, such that in the future, if the duplicate points issue is "properly" addressed, it's clear that this de-duplicate step can be removed?

moveit_planners/pilz_industrial_motion_planner/src/command_list_manager.cpp
```diff
@@ -112,7 +112,24 @@ RobotTrajCont CommandListManager::solve(const planning_scene::PlanningSceneConst
                               // therefore: "i-1".
                               (i > 0 ? radii.at(i - 1) : 0.));
   }
-  return plan_comp_builder_.build();
+
+  const auto res_vec = plan_comp_builder_.build();
+
+  // De-duplicate trajectory points with the same time value.
+  // This is necessary since some controllers do not allow times that are not monotonically increasing.
```
sea-bass, line 119: // De-duplicate trajectory points with the same time value.
  // This is necessary since some controllers do not allow times that are not monotonically increasing.
  // TODO: Ideally, we would not need this code if the trajectory segments were created without
  // duplicate time points in the first place. Leaving this note to revisit this with a more principled fix.

stephanie-eng (approved):

sjahr (approved):

PR #2949, opened 2024-08-05
Add parameter `default_planning_scene` to load planning scene geometry on startup

pac48:
### Description

This PR adds a new parameter to the `move_group` node that allows users to specify a planning scene file to load on startup. The parameter is a string containing the absolute file path. The parameter can be set with either `$(find-pkg-share pkg-wth-planning-file)` for XML or  `ament_index_python.packages.get_package_share_directory` in Python to facilitate specifying the path. If the parameter is not specified, the `move_group` node will behave as it previously did. If there is an error loading the file, the node will exit with a fatal error.

### Checklist
(https://github.com/moveit/moveit2/blob/main/MIGRATION.md) notes

### DIFF
--- a/moveit_ros/move_group/src/move_group.cpp
+++ b/moveit_ros/move_group/src/move_group.cpp
@@ -34,17 +34,20 @@
 
 /* Author: Ioan Sucan */
 
-#include <moveit/moveit_cpp/moveit_cpp.h>
-#include <moveit/planning_scene_monitor/planning_scene_monitor.h>
-#include <tf2_ros/transform_listener.h>
-#include <moveit/move_group/move_group_capability.h>
-#include <moveit/trajectory_execution_manager/trajectory_execution_manager.h>
+#include <memory>
+#include <bits/fs_path.h>
+#include <filesystem>
+#include <set>
+
 #include <boost/tokenizer.hpp>
 #include <moveit/macros/console_colors.h>
+#include <moveit/move_group/move_group_capability.h>
 #include <moveit/move_group/move_group_context.h>
-#include <memory>
-#include <set>
+#include <moveit/moveit_cpp/moveit_cpp.h>
+#include <moveit/planning_scene_monitor/planning_scene_monitor.h>
+#include <moveit/trajectory_execution_manager/trajectory_execution_manager.h>
 #include <moveit/utils/logger.hpp>
+#include <tf2_ros/transform_listener.h>
 
 static const std::string ROBOT_DESCRIPTION =
     "robot_description";  // name of the robot description (a param name, so it can be changed externally)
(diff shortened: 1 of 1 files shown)


moveit_ros/move_group/src/move_group.cpp
```diff
@@ -287,9 +288,25 @@ int main(int argc, char** argv)
...
-
-  if (planning_scene_monitor->getPlanningScene())
+  if (auto ps = planning_scene_monitor->getPlanningScene())
   {
+    if (nh->has_parameter("default_planning_scene"))
+    {
+      std::string path = nh->get_parameter("default_planning_scene").as_string();
+      std::fstream file_stream;
+      file_stream.open(path, std::fstream::in);
+      if (!file_stream.is_open() || !ps->loadGeometryFromStream(file_stream))
```
sjahr, line 298: I think this is too dangerous because you're not locking the planning scene when writing to it (see https://github.com/moveit/moveit2/blob/19d4a2818bca7024305cbf5be5c3bca4186351d5/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h#L170).
You can get thread safe access like this:
```cpp
{
  planning_scene_monitor::LockedPlanningSceneRW ps = planning_display_->getPlanningSceneRW();
  if (ps)
     FANCY MODIFICATIONS
} // End scope to release lock
```

moveit_ros/move_group/src/move_group.cpp
```diff
@@ -290,8 +290,26 @@ int main(int argc, char** argv)
   const auto moveit_cpp = std::make_shared<moveit_cpp::MoveItCpp>(nh, moveit_cpp_options);
   const auto planning_scene_monitor = moveit_cpp->getPlanningSceneMonitorNonConst();
 
-  if (planning_scene_monitor->getPlanningScene())
+  if (auto ps = planning_scene_monitor->getPlanningScene())
```
sjahr, line 293: Given that `ps` is never used, would make sense to revert this change?

moveit_ros/move_group/src/move_group.cpp
```diff
@@ -290,8 +290,26 @@ int main(int argc, char** argv)
...
   {
+    if (nh->has_parameter("default_planning_scene"))
```
sjahr, line 295: I am a bit concerned about tying this to only the existence of the parameter. How about checking whether or not the parameter is set.
```cpp
const auto default_planning_scene_file = std::string();
nh->get_parameter_or("default_planning_scene", default_planning_scene_file, std::string());
// Maybe add function to validate the string pattern if it actually representing  path
if(validateFilePath(default_planning_scene_file))
{
...
}


```

sjahr (commented):
I think it is a useful addition but it could be hardened a bit to make it harder to misuse. Also can you add the PR description as a documentation comment on top of this function call? At the moment there is nothing explaining what this is doing or how it is supposed to be used

PR #2954, opened 2024-08-07
Ensure the robot state is up-to-date before Servoing

AndyZe:
### Description

Previously we didn't wait to get a current robot state before servoing began. That caused the initial joint angles to be all-zero, which caused a small jump at the beginning of motion. This is a fix.

Test with:

`ros2 launch moveit_servo demo_ros_api.launch.py`

`ros2 run moveit_servo servo_keyboard_input`

### DIFF
--- a/moveit_ros/moveit_servo/src/servo.cpp
+++ b/moveit_ros/moveit_servo/src/servo.cpp
@@ -643,8 +643,21 @@ std::optional<PoseCommand> Servo::toPlanningFrame(const PoseCommand& command, co
   return PoseCommand{ planning_frame, planning_to_command_tf * command.pose };
 }
 
-KinematicState Servo::getCurrentRobotState() const
+KinematicState Servo::getCurrentRobotState(bool block_for_current_state) const
 {
+  if (block_for_current_state)
+  {
+    bool have_current_state = false;
+    while (rclcpp::ok() && !have_current_state)
+    {
+      have_current_state = planning_scene_monitor_->getStateMonitor()->waitForCurrentState(
+          rclcpp::Clock(RCL_ROS_TIME).now(), ROBOT_STATE_WAIT_TIME /* s */);
+      if (!have_current_state)
+      {
+        RCLCPP_WARN(logger_, "Waiting for the current state");
+      }
+    }
+  }
   moveit::core::RobotStatePtr robot_state = planning_scene_monitor_->getStateMonitor()->getCurrentState();
   return extractRobotState(robot_state, servo_params_.move_group_name);
 }
--- a/moveit_ros/moveit_servo/src/servo.cpp
+++ b/moveit_ros/moveit_servo/src/servo.cpp
@@ -158,7 +158,6 @@ void Servo::setSmoothingPlugin()
     RCLCPP_ERROR(logger_, "Smoothing plugin could not be initialized");
     std::exit(EXIT_FAILURE);
   }
-  resetSmoothing(getCurrentRobotState());
 }
 
 void Servo::doSmoothing(KinematicState& state)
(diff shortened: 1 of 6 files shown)


moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -158,7 +158,6 @@ void Servo::setSmoothingPlugin()
     RCLCPP_ERROR(logger_, "Smoothing plugin could not be initialized");
     std::exit(EXIT_FAILURE);
   }
-  resetSmoothing(getCurrentRobotState());
```
AndyZe, line 161: The positions from `getCurrentRobotState()` were all-zeros since this call occurred early and we didn't wait for a valid robot state.

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -548,8 +544,8 @@ KinematicState Servo::getNextJointState(const moveit::core::RobotStatePtr& robot
     }
   }
 
-  // Update internal state of filter with final calculated command.
-  resetSmoothing(target_state);
+  // Apply smoothing to the positions if a smoother was provided.
+  doSmoothing(target_state);
```
AndyZe, line 548: Calling resetSmoothing() in the loop was a bug. It's not compatible with the Butterworth filter b/c it wipes out the history of smoothed positions. See https://github.com/moveit/moveit2/blob/f30d51b2cef111ca92c89ced8776960d6e3b2470/moveit_core/online_signal_smoothing/src/butterworth_filter.cpp#L147

```
  for (size_t joint_idx = 0; joint_idx < num_positions; ++joint_idx)
  {
    position_filters_.at(joint_idx).reset(positions[joint_idx]);
  }
```

moveit_ros/moveit_servo/src/servo_node.cpp
```diff
@@ -303,6 +303,8 @@ void ServoNode::servoLoop()
   }
   KinematicState current_state = servo_->getCurrentRobotState();
   last_commanded_state_ = current_state;
+  // Ensure the filter is up to date
+  servo_->resetSmoothing(current_state);
```
AndyZe, line 307: Now the initial joint state is retrieved here. A few lines above, we wait for the up-to-date robot state.

sea-bass (commented):
Fixes look good at a glance!

One thing to watch out for, from my own experience, is what happens if you pause/stop servo, move the robot through some other free-space motion planner, and then resume servo. (Maybe @pac48 and @sjahr can test with MoveIt Pro and double check?)

Far as I remember, this was one reason for the liberal sprinkling of `resetSmoothing()`, but I don't doubt this PR fixes that. I just want to make sure it wasn't pruned too much to affect the interaction with other motions.

Essentially, there are 2 alternatives:
1. Keep the smoothing going even when servo is stopped,
2. Ensure that transitioning servo from stopped to not stopped triggers a reset to the latest state when restarting.

There are merits to both, but just wanted to bring it up and have people test this case.

moveit_ros/moveit_servo/src/servo.cpp
AndyZe: @pac48 can you do this reset at the end of the `doSmoothing()` method of your filter?

AndyZe:
> One thing to watch out for, from my own experience, is what happens if you pause/stop servo, move the robot through some other free-space motion planner, and then resume servo. (Maybe @pac48 and @sjahr can test with MoveIt Pro and double check?)

I had to add one line to fix this. Good catch. Here's a video of several cycles of Servo -> pause -> joint command -> unpause -> Servo again


[servo_pausing.webm](https://github.com/user-attachments/assets/a034b733-4b8d-4124-a700-5baf64cc0474)

sea-bass (approved):
Awesome, thanks for that test!

sjahr (approved):
Nice work, thanks!

PR #2961, opened 2024-08-10
Fix Pilz blending times... the right way

sea-bass:
### Description

Found the issue!

Turns out that Pilz's `appendWithStrictTimeIncrease()` function had a single if-statement that encompassed two conditions:
1. If trajectory is empty
2. If the end state of one segment is not equal to the start state of the second segment

In both cases, the whole trajectory segment was appended with a `dt=0`.

The problem is that in case 1, it's appropriate to add a `dt=0` (as the trajectory is empty, the first point should be at `t=0`). In case 2, we needed to apply a non-zero sample time offset dictated by the actual `dt` of that trajectory's first waypoint.

Closes https://github.com/moveit/moveit2/issues/2945

### Checklist

### DIFF
--- a/moveit_planners/pilz_industrial_motion_planner/src/trajectory_blender_transition_window.cpp
+++ b/moveit_planners/pilz_industrial_motion_planner/src/trajectory_blender_transition_window.cpp
@@ -70,7 +70,7 @@ bool pilz_industrial_motion_planner::TrajectoryBlenderTransitionWindow::blend(
   std::size_t second_intersection_index;
   if (!searchIntersectionPoints(req, first_intersection_index, second_intersection_index))
   {
-    RCLCPP_ERROR(getLogger(), "Blend radius to large.");
+    RCLCPP_ERROR(getLogger(), "Blend radius too large.");
     res.error_code.val = moveit_msgs::msg::MoveItErrorCodes::INVALID_MOTION_PLAN;
     return false;
   }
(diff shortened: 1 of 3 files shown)


moveit_planners/pilz_industrial_motion_planner/src/trajectory_blender_transition_window.cpp
```diff
@@ -70,7 +70,7 @@ bool pilz_industrial_motion_planner::TrajectoryBlenderTransitionWindow::blend(
   std::size_t second_intersection_index;
   if (!searchIntersectionPoints(req, first_intersection_index, second_intersection_index))
   {
-    RCLCPP_ERROR(getLogger(), "Blend radius to large.");
+    RCLCPP_ERROR(getLogger(), "Blend radius too large.");
```
sea-bass, line 73: freebie typo fix (:

sea-bass:
@ct2034 want to give this a sanity check?

stephanie-eng (approved):
LGTM!

Could you verify that this prevents the duplicate timestamp issue you saw in the tutorial? I could never reproduce it for whatever reason, so would be good to know if it works for you.

sea-bass:
> Could you verify that this prevents the duplicate timestamp issue you saw in the tutorial? I could never reproduce it for whatever reason, so would be good to know if it works for you.

Oh, it for sure resolves this issue. It's how I've been testing it.

In fact, the previous PR squashed duplicate time points, but this one actually handles it correctly and ensures the two points are spaced accordingly.

sea-bass:
```
error: unknown warning option '-Wno-maybe-uninitialized'; did you mean '-Wno-uninitialized'? [clang-diagnostic-unknown-warning-option]
```

anyone know what this is?

https://github.com/moveit/moveit2/actions/runs/10333028649/job/28604931968?pr=2961#step:11:8033

rhaschke (approved):

PR #2962, opened 2024-08-10
Small fixes to flaky MoveIt Servo integration tests

sea-bass:
### Description

I've been seeing servo integration tests flake every now and then, and found it improved when removing the node before trying to cancel the executor and join its spinning thread.

I also changed some incorrect `/` characters in the integration test launch files' static transform publisher. We were failing to get a proper state from the planning scene because the virtual joint was flagged as missing. 

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/tests/servo_ros_fixture.hpp
+++ b/moveit_ros/moveit_servo/tests/servo_ros_fixture.hpp
@@ -93,9 +93,10 @@ protected:
     switch_input_client_ =
         servo_test_node_->create_client<moveit_msgs::srv::ServoCommandType>("/servo_node/switch_command_type");
 
+    waitForService();
+
     executor_.add_node(servo_test_node_);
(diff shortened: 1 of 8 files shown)


moveit_ros/moveit_servo/tests/servo_ros_fixture.hpp
```diff
@@ -96,13 +95,14 @@ class ServoRosFixture : public testing::Test
...
-    executor_->add_node(servo_test_node_);
-    executor_thread_ = std::thread([this]() { executor_->spin(); });
+    executor_.add_node(servo_test_node_);
+    executor_thread_ = std::thread([this]() { executor_.spin(); });
   }
 
   void TearDown() override
   {
-    executor_->cancel();
+    executor_.remove_node(servo_test_node_);
```
AndyZe, line 104: Seems like you shouldn't have to do this, but no harm done I guess.

AndyZe (approved):

PR #2963, opened 2024-08-10
Fixed segmentation fault in depth_image_octomap_updater

CihatAltiparmak:
### Description

The relavant issue:  https://github.com/moveit/moveit2/issues/2964

### Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -31,8 +31,10 @@ jobs:
           - IMAGE: jazzy-ci
             ROS_DISTRO: jazzy
     env:
+      # TODO(andyz): When this clang-tidy issue is fixed, remove -Wno-unknown-warning-option
+      # https://stackoverflow.com/a/41673702
       CXXFLAGS: >-
-        -Wall -Wextra -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls
+        -Wall -Wextra -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls -Wno-unknown-warning-option
       CLANG_TIDY_ARGS: --fix --fix-errors --format-style=file
       DOCKER_IMAGE: moveit/moveit2:${{ matrix.env.IMAGE }}
       UPSTREAM_WORKSPACE: >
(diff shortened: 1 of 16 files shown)


rhaschke (changes requested):
Please always explain your changes! 
This looks like a great data type mess: the data array of type `uint8` should store floats (`TYPE_32FC1`), but is passed as a `double` pointer to `getModelDepth()`, which might be the root cause of the segfault.
Now, you allocate an additional `float` array (which takes extra time and memory), still pass it as a `double` pointer (messing up the data) and finally copy the float data as unsigned short into the uint8 slots using some weird transformation. As the `uint8` array is to be interpreted as a float array, this is definitely still wrong.

CihatAltiparmak:
> Now, you allocate an additional `float` array (which takes extra time and memory), still pass it as a `double` pointer (messing up the data) and finally copy the float data as unsigned short into the uint8 slots using some weird transformation.

It actually comes from https://github.com/moveit/moveit2/blob/aece55c3ebe4a88bb9c09b85fa12479b94690c0b/moveit_ros/perception/depth_image_octomap_updater/src/depth_image_octomap_updater.cpp#L468-L492

> Have a look at the [MoveIt1 implementation of that snippet](https://github.com/moveit/moveit/blob/61afd978498291e5113115766d2505c1172bf651/moveit_ros/perception/depth_image_octomap_updater/src/depth_image_octomap_updater.cpp#L410-L411), which seems to use the correct types.

Hmm, thanks, i will give a try converting `getModelDepth(double)` to `getModelDepth(float)`.

rhaschke:
The referenced code looks wrong as well.

CihatAltiparmak:
@rhaschke Should we remove this parts too, what do you think @rhaschke ?

https://github.com/moveit/moveit2/blob/aece55c3ebe4a88bb9c09b85fa12479b94690c0b/moveit_ros/perception/depth_image_octomap_updater/src/depth_image_octomap_updater.cpp#L486-L490

rhaschke (approved):
This looks better now. Now idea, why #2343 blindly converted all float args to double.

rhaschke:
> Should we remove this parts too (Moveit1 has same implementation)?

No. This code explicitly returns an unsigned short array, i.e. converting float meters into int millimeters.

sjahr:
> This looks better now. Now idea, why #2343 blindly converted all float args to double.

Thanks for referencing moveit1. The reason for the conversation is: https://github.com/cpp-best-practices/cppbestpractices/blob/master/08-Considering_Performance.md#prefer-double-to-float-but-test-first but apparently we ignore the `test_first` part :man_facepalming:

CihatAltiparmak:
I have surfed through the PR you mentioned and i also discovered that the evil is also here ( https://github.com/moveit/moveit2/pull/1578 ) I can be necessary to revert some float-to-double conversions

PR #2966, opened 2024-08-12
Update the "format" workflow version of ubuntu

AndyZe:
### Description

If you search the `.github` folder, you'll see that `ubuntu-latest` is used almost everywhere. And that makes good sense. Switch `ubuntu-22.04` to `ubuntu-latest` in this one last place.

### DIFF
--- a/.github/workflows/format.yaml
+++ b/.github/workflows/format.yaml
@@ -14,7 +14,7 @@ on:
 jobs:
   pre-commit:
     name: Format
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


sea-bass (approved):

rhaschke (commented):
Having Ubuntu 22.04 here is for a reason: while most other workflows run a docker container and thus can run on any Ubuntu distro, this one directly runs in the given Ubuntu distro.
Using ubuntu-latest here will break the workflow in future, when latest refers to a future distro, which doesn't ship clang-format-14 anymore.

PR #2967, opened 2024-08-12
Unknown compiler options for clang shouldn't be show-stoppers

AndyZe:
### Description

This is a workaround for a CI error that has been cropping up once in awhile recently:

>    Error while processing /home/runner/work/moveit2/moveit2/.work/target_ws/src/moveit2/moveit_ros/moveit_servo/src/utils/common.cpp.
  error: unknown warning option '-Wno-maybe-uninitialized'; did you mean '-Wno-uninitialized'? [clang-diagnostic-unknown-warning-option]
  
 From https://github.com/moveit/moveit2/actions/runs/10335417126/job/28609933300 for one example
  
 I don't think there's actually a potentially uninitialized variable there, at least not a new one in this PR. I think what happened is, `clang-tidy` was cached previously for this `cpp` file. That PR happened to touch that `cpp` file, which caused `clang-tidy` to re-run for it. And something about `clang-tidy` has changed since the last time it ran.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -31,8 +31,10 @@ jobs:
           - IMAGE: jazzy-ci
             ROS_DISTRO: jazzy
     env:
+      # TODO(andyz): When this clang-tidy issue is fixed, remove -Wno-unknown-warning-option
+      # https://stackoverflow.com/a/41673702
       CXXFLAGS: >-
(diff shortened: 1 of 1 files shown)


sea-bass (commented):
I additionally searched for this issue and found this relevant post: https://stackoverflow.com/a/41673702

So this change makes sense to me, but would be nice for someone with more software chops than me to give it the final approval.

rhaschke (approved):
Maybe add a comment to remove that option in the future if clang-tidy is fixed.

sea-bass (approved):

PR #2969, opened 2024-08-12
Never return nullptr silently

sjahr:
### Description

- Create some warning/ error message before returning a nullptr for better debugability
- Some minor documentation improvement

### Checklist

### DIFF
--- a/moveit_core/robot_model/src/robot_model.cpp
+++ b/moveit_core/robot_model/src/robot_model.cpp
@@ -1357,21 +1357,21 @@ LinkModel* RobotModel::getLinkModel(const std::string& name, bool* has_link)
   {
     *has_link = false;  // Report failure via argument
   }
-  else
-  {  // Otherwise print error
-    RCLCPP_ERROR(getLogger(), "Link '%s' not found in model '%s'", name.c_str(), model_name_.c_str());
-  }
+  RCLCPP_ERROR(getLogger(), "Link '%s' not found in model '%s'", name.c_str(), model_name_.c_str());
   return nullptr;
 }
 
 const LinkModel* RobotModel::getRigidlyConnectedParentLinkModel(const LinkModel* link)
 {
   if (!link)
+  {
+    RCLCPP_ERROR(get_logger(), "Cannot determine rigidly connected parent link because input link is nullptr");
     return link;
+  }
   const moveit::core::LinkModel* parent_link = link->getParentLinkModel();
   const moveit::core::JointModel* joint = link->getParentJointModel();
 
-  while (parent_link && joint->getType() == moveit::core::JointModel::FIXED)
+  while (parent_link && joint && joint->getType() == moveit::core::JointModel::FIXED)
   {
     link = parent_link;
     joint = link->getParentJointModel();
--- a/moveit_core/robot_state/src/robot_state.cpp
+++ b/moveit_core/robot_state/src/robot_state.cpp
@@ -519,6 +519,8 @@ const double* RobotState::getJointPositions(const JointModel* joint) const
 {
   if (joint->getVariableCount() == 0)
   {
+    RCLCPP_ERROR(getLogger(), "Cannot get joint positions for joint '%s' because it has no variables",
+                 joint->getName().c_str());
     return nullptr;
   }
   return &position_.at(joint->getFirstVariableIndex());
@@ -528,6 +530,8 @@ const double* RobotState::getJointVelocities(const JointModel* joint) const
 {
   if (joint->getVariableCount() == 0)
   {
+    RCLCPP_ERROR(getLogger(), "Cannot get joint velocities for joint '%s' because it has no variables",
+                 joint->getName().c_str());
     return nullptr;
   }
   return &velocity_.at(joint->getFirstVariableIndex());
@@ -537,6 +541,8 @@ const double* RobotState::getJointAccelerations(const JointModel* joint) const
 {
   if (joint->getVariableCount() == 0)
   {
+    RCLCPP_ERROR(getLogger(), "Cannot get joint accelerations for joint '%s' because it has no variables",
+                 joint->getName().c_str());
     return nullptr;
   }
   return &effort_or_acceleration_.at(joint->getFirstVariableIndex());
@@ -546,6 +552,8 @@ const double* RobotState::getJointEffort(const JointModel* joint) const
 {
   if (joint->getVariableCount() == 0)
   {
+    RCLCPP_ERROR(getLogger(), "Cannot get joint effort for joint '%s' because it has no variables",
+                 joint->getName().c_str());
     return nullptr;
   }
   return &effort_or_acceleration_.at(joint->getFirstVariableIndex());
@@ -911,24 +919,43 @@ const LinkModel* RobotState::getRigidlyConnectedParentLinkModel(const std::strin
 {
   const LinkModel* link{ nullptr };
 
+  // Resolve the rigidly connected parent link for the given frame
   size_t idx = 0;
   if ((idx = frame.find('/')) != std::string::npos)
-  {  // resolve sub frame
+  {  // Check if the frame has a subframe
     std::string object{ frame.substr(0, idx) };
     if (!hasAttachedBody(object))
+    {
+      // If the attached body is not found, log an error and return nullptr
+      RCLCPP_ERROR(getLogger(),
+                   "Cannot find rigidly connected parent link for frame '%s' because there is no attached body.",
+                   frame.c_str());
       return nullptr;
+    }
     auto body{ getAttachedBody(object) };
     if (!body->hasSubframeTransform(frame))
+    {
+      // If the subframe transform is not found, log an error and return nullptr
+      RCLCPP_ERROR(getLogger(),
+                   "Cannot find rigidly connected parent link for frame '%s' because the transformation to the parent "
+                   "link is unknown.",
+                   frame.c_str());
       return nullptr;
+    }
     link = body->getAttachedLink();
   }
   else if (hasAttachedBody(frame))
   {
+    // If the frame is an attached body, get the attached link
     link = getAttachedBody(frame)->getAttachedLink();
   }
   else if (getRobotModel()->hasLinkModel(frame))
+  {
+    // If the frame is a link in the robot model, get the link model
     link = getLinkModel(frame);
+  }
 
+  // Return the rigidly connected parent link model for the given frame
   return getRobotModel()->getRigidlyConnectedParentLinkModel(link);
 }
 
--- a/moveit_core/collision_detection_bullet/src/bullet_integration/contact_checker_common.cpp
+++ b/moveit_core/collision_detection_bullet/src/bullet_integration/contact_checker_common.cpp
@@ -116,7 +116,7 @@ collision_detection::Contact* processResult(ContactTestData& cdata, collision_de
 
     return &(dr.back());
   }
-
+  RCLCPP_ERROR(getLogger(), "Result cannot be processed. Returning nullptr, this should never happen!");
   return nullptr;
 }
 
--- a/moveit_planners/pilz_industrial_motion_planner/src/trajectory_generator_circ.cpp
+++ b/moveit_planners/pilz_industrial_motion_planner/src/trajectory_generator_circ.cpp
@@ -268,6 +268,7 @@ std::unique_ptr<KDL::Path> TrajectoryGeneratorCIRC::setPathCIRC(const MotionPlan
     throw CenterPointDifferentRadius(os.str());
   }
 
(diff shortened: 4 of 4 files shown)


moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -911,24 +919,43 @@ const LinkModel* RobotState::getRigidlyConnectedParentLinkModel(const std::strin
...
+      // If the attached body is not found, log an error and return nullptr
+      RCLCPP_ERROR(getLogger(),
+                   "Cannot find rigidly connected parent link for frame '%s' because there is no attached body.",
+                   frame.c_str());
       return nullptr;
+    }
     auto body{ getAttachedBody(object) };
     if (!body->hasSubframeTransform(frame))
+    {
+      // If the subframe transform is not found, log an error and return nullptr
```
sjahr, line 938:
```suggestion
```

moveit_core/robot_model/src/robot_model.cpp
```diff
@@ -1357,21 +1357,21 @@ LinkModel* RobotModel::getLinkModel(const std::string& name, bool* has_link)
...
 
 const LinkModel* RobotModel::getRigidlyConnectedParentLinkModel(const LinkModel* link)
 {
   if (!link)
+  {
+    RCLCPP_ERROR(getLogger(), "Cannot determine rigidly connected parent link because input link is nullptr");
     return link;
+  }
   const moveit::core::LinkModel* parent_link = link->getParentLinkModel();
   const moveit::core::JointModel* joint = link->getParentJointModel();
```
rhaschke, line 1372: A valid link always has a parent joint.

moveit_core/robot_model/src/robot_model.cpp
```diff
@@ -1357,21 +1357,21 @@ LinkModel* RobotModel::getLinkModel(const std::string& name, bool* has_link)
   {
     *has_link = false;  // Report failure via argument
   }
-  else
-  {  // Otherwise print error
-    RCLCPP_ERROR(getLogger(), "Link '%s' not found in model '%s'", name.c_str(), model_name_.c_str());
-  }
```
rhaschke, line 1363: The error message was printed only if the caller didn't provide the has_link argument to handle error checking itself. Probably, now you will get duplicates of the error message.

rhaschke (commented):
Issueing an error message before returning a nullptr is not always desired: It might be perfectly possible that the caller correctly handles the nullptr result and issues an error itself or handles the error condition gracefully in some other fashion.
In that case, the error message would be spurious.
At most, you should issue a debug message.

moveit_core/robot_model/src/robot_model.cpp
```diff
@@ -1357,21 +1357,21 @@ LinkModel* RobotModel::getLinkModel(const std::string& name, bool* has_link)
...
   if (!link)
+  {
+    RCLCPP_ERROR(getLogger(), "Cannot determine rigidly connected parent link because input link is nullptr");
     return link;
+  }
   const moveit::core::LinkModel* parent_link = link->getParentLinkModel();
   const moveit::core::JointModel* joint = link->getParentJointModel();
 
-  while (parent_link && joint->getType() == moveit::core::JointModel::FIXED)
+  while (parent_link && joint && joint->getType() == moveit::core::JointModel::FIXED)
```
sjahr, line 1374:
```suggestion
  while (parent_link && joint->getType() == moveit::core::JointModel::FIXED)
```

marioprats:
I agree with Robert.
The functions returning `nullptr` are indicative of an error, which the caller should handle.
A better solution would be to introduce a good error propagation mechanism, like `tl::expected` or `absl::Status`. Then we could return an error with a string, and let the caller decide what to do.
In the meanwhile I think a log message makes sense, but I would make it a debug message.

moveit_core/robot_model/src/robot_model.cpp
sjahr: I deprecated the argument, this pattern is just ugly and prone to errors

moveit_core/robot_model/src/robot_model.cpp
```diff
@@ -1331,7 +1331,8 @@ JointModel* RobotModel::getJointModel(const std::string& name)
 
 const LinkModel* RobotModel::getLinkModel(const std::string& name, bool* has_link) const
 {
-  return const_cast<RobotModel*>(this)->getLinkModel(name, has_link);
+  *has_link = false;
```
rhaschke, line 1334: This will segfault if `has_link` is NULL!

moveit_core/robot_model/src/robot_model.cpp
```diff
@@ -1347,27 +1348,29 @@ const LinkModel* RobotModel::getLinkModel(size_t index) const
...
 {
-  if (has_link)
-    *has_link = true;  // Start out optimistic
   LinkModelMap::const_iterator it = link_model_map_.find(name);
   if (it != link_model_map_.end())
-    return it->second;
-
-  if (has_link)
   {
-    *has_link = false;  // Report failure via argument
```
rhaschke, line 1358: `has_link` was explicitly used to suppress the error message here and issue an error by the caller.

moveit_core/robot_model/src/robot_model.cpp
```diff
@@ -1347,27 +1348,29 @@ const LinkModel* RobotModel::getLinkModel(size_t index) const
...
+    return it->second;
   }
+  RCLCPP_ERROR(getLogger(), "Link '%s' not found in model '%s'", name.c_str(), model_name_.c_str());
   return nullptr;
 }
 
+const LinkModel* getLinkModel(const std::string& link, bool* has_link) const
+{
+  RCLCPP_ERROR(getLogger(), "'has_link' argument is deprecated. Use getLinkModel(const std::string& name) instead.");
+  *has_link = false;
```
rhaschke, line 1363: segfault (see above)

moveit_core/robot_model/src/robot_model.cpp
```diff
@@ -1347,27 +1348,29 @@ const LinkModel* RobotModel::getLinkModel(size_t index) const
...
-    RCLCPP_ERROR(getLogger(), "Link '%s' not found in model '%s'", name.c_str(), model_name_.c_str());
+    return it->second;
   }
+  RCLCPP_ERROR(getLogger(), "Link '%s' not found in model '%s'", name.c_str(), model_name_.c_str());
   return nullptr;
 }
 
+const LinkModel* getLinkModel(const std::string& link, bool* has_link) const
+{
+  RCLCPP_ERROR(getLogger(), "'has_link' argument is deprecated. Use getLinkModel(const std::string& name) instead.");
```
rhaschke, line 1362: Issue a deprecation warning at compile-time only!

rhaschke (changes requested):
The last commit (57a85c88fd02728c18e24c282940cb7bce1ac823) should be handled in a separate PR. This is a fundamental change of the API and logic, which should be well motivated. I don't yet see the "ugliness" of the current API.
You should better address the reviews and change error messages to debug messages.

sjahr:
> The last commit ([57a85c8](https://github.com/moveit/moveit2/commit/57a85c88fd02728c18e24c282940cb7bce1ac823)) should be handled in a separate PR. This is a fundamental change of the API and logic, which should be well motivated. I don't yet see the "ugliness" of the current API. You should better address the reviews and change error messages to debug messages.

You're right, that derailed to initial intention of the PR too much, sorry. I reverted it and updated the log level to debug in most cases. Let me know what you think. 

>  I don't yet see the "ugliness" of the current API.

Maybe isn't the right term but rather outdated. I think we shouldn't pass raw boolean pointers anywhere in modern C++ but as you pointed out correctly, this PR is not the place to try changing that.

moveit_core/robot_model/src/robot_model.cpp
```diff
@@ -1367,7 +1367,10 @@ LinkModel* RobotModel::getLinkModel(const std::string& name, bool* has_link)
 const LinkModel* RobotModel::getRigidlyConnectedParentLinkModel(const LinkModel* link)
 {
   if (!link)
+  {
+    RCLCPP_ERROR(getLogger(), "Cannot determine rigidly connected parent link because input link is nullptr");
```
rhaschke, line 1371: The nullptr is correctly handled by callers. Change to debug level:
```suggestion
    RCLCPP_DEBUG(getLogger().get_child("getRigidlyConnectedParentLink", "Link is NULL");
```

rhaschke (commented):
See comments below.

rhaschke:
Sorry, I missed some closing parentheses.

moveit_core/robot_model/src/robot_model.cpp
```diff
@@ -1367,7 +1367,10 @@ LinkModel* RobotModel::getLinkModel(const std::string& name, bool* has_link)
 const LinkModel* RobotModel::getRigidlyConnectedParentLinkModel(const LinkModel* link)
 {
   if (!link)
+  {
+    RCLCPP_DEBUG(getLogger().get_child("getRigidlyConnectedParentLink", "Link is NULL");
```
sjahr, line 1371:
```suggestion
    RCLCPP_DEBUG(getLogger().get_child("getRigidlyConnectedParentLink", "Link is NULL"));
```

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -911,24 +919,38 @@ const LinkModel* RobotState::getRigidlyConnectedParentLinkModel(const std::strin
...
+    {
+      RCLCPP_DEBUG(getLogger().get_child("getRigidlyConnectedParentLink",
+                   "Attached object '%s' for frame '%s' does not exist.", object.c_str(), frame.c_str());
       return nullptr;
+    }
     auto body{ getAttachedBody(object) };
     if (!body->hasSubframeTransform(frame))
+    {
+      RCLCPP_DEBUG(getLogger().get_child("getRigidlyConnectedParentLink",
+                   "Body '%s' does not have subframe '%s', object.c_str(), frame.c_str());
```
sjahr, line 937:
```suggestion
                   "Body '%s' does not have subframe '%s'", object.c_str(), frame.c_str());
```

moveit_core/robot_model/src/robot_model.cpp
```diff
@@ -1367,7 +1367,10 @@ LinkModel* RobotModel::getLinkModel(const std::string& name, bool* has_link)
 const LinkModel* RobotModel::getRigidlyConnectedParentLinkModel(const LinkModel* link)
 {
   if (!link)
+  {
+    RCLCPP_DEBUG(getLogger().get_child("getRigidlyConnectedParentLink", "Link is NULL"));
```
rhaschke, line 1371:
```suggestion
    RCLCPP_DEBUG(getLogger().get_child("getRigidlyConnectedParentLink"), "Link is NULL");
```

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -911,24 +919,38 @@ const LinkModel* RobotState::getRigidlyConnectedParentLinkModel(const std::strin
...
+  // Resolve the rigidly connected parent link for the given frame
   size_t idx = 0;
   if ((idx = frame.find('/')) != std::string::npos)
-  {  // resolve sub frame
+  {  // Check if the frame has a subframe
     std::string object{ frame.substr(0, idx) };
     if (!hasAttachedBody(object))
+    {
+      RCLCPP_DEBUG(getLogger().get_child("getRigidlyConnectedParentLink",
+                   "Attached object '%s' for frame '%s' does not exist.", object.c_str(), frame.c_str());
```
rhaschke, line 930:
```suggestion
      RCLCPP_DEBUG(getLogger().get_child("getRigidlyConnectedParentLink"),
                   "Attached object '%s' for frame '%s' does not exist.", object.c_str(), frame.c_str());
```

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -911,24 +919,38 @@ const LinkModel* RobotState::getRigidlyConnectedParentLinkModel(const std::strin
...
       return nullptr;
+    }
     auto body{ getAttachedBody(object) };
     if (!body->hasSubframeTransform(frame))
+    {
+      RCLCPP_DEBUG(getLogger().get_child("getRigidlyConnectedParentLink",
+                   "Body '%s' does not have subframe '%s'", object.c_str(), frame.c_str());
```
rhaschke, line 937:
```suggestion
      RCLCPP_DEBUG(getLogger().get_child("getRigidlyConnectedParentLink"),
                   "Body '%s' does not have subframe '%s'", object.c_str(), frame.c_str());
```

rhaschke (commented):
Fix parentheses. Still unchecked. I just did that in the Web GUI.

henningkayser:
@sjahr unfortunately this needs a new iteration because I merged #2985

PR #2985, opened 2024-08-29
Fix RobotState::getRigidlyConnectedParentLinkModel()

rhaschke:
Simplify function using `getFrameInfo()` to yield the link corresponding to the given frame.

The order of frame resolution was wrong here:
If a _link_ frame containing a _slash_ was given, the code was expecting an attached body with a subframe. As these didn't exist, the function falsely returned `NULL`.

@sjahr: Maybe this is the error you recently observed? Note, that the function was already fixed in MoveIt1.


henningkayser:
Answering for @sjahr, this really looks like the issue we have been experiencing. This is a really good code simplification as well. Thank you!

henningkayser (approved):

henningkayser:
The backport should have been triggered.... Not sure what went wrong here. But we can also call mergify manually. Let me check

henningkayser:
@Mergifyio backport humble

PR #3003, opened 2024-09-16
Update urdf/model.h -> urdf/model.hpp

rhaschke:
The old header is marked obsolete.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -34,7 +34,7 @@ jobs:
       # TODO(andyz): When this clang-tidy issue is fixed, remove -Wno-unknown-warning-option
       # https://stackoverflow.com/a/41673702
       CXXFLAGS: >-
-        -Wall -Wextra -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls -Wno-unknown-warning-option
+        -Wall -Wextra -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls -Wno-unknown-warning-option -Wno-cpp
       CLANG_TIDY_ARGS: --fix --fix-errors --format-style=file
       DOCKER_IMAGE: moveit/moveit2:${{ matrix.env.IMAGE }}
       UPSTREAM_WORKSPACE: >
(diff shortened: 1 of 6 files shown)


sea-bass (approved):

sea-bass (approved):
This is actually fine.

HOWEVER, CI is still failing because `srdfdom` requires the same update, and a release: https://github.com/moveit/srdfdom/pull/127

sea-bass:
BTW for reference, seems the deprecation warning came in with a linting PR: https://github.com/ros2/urdf/pull/39

sea-bass:
@sjahr shared this PR with me as an idea, which we could use here since MoveIt definitely can incur dependencies on `rclcpp`: https://github.com/moveit/moveit2/pull/2978

sjahr:
@rhaschke I took your PR and added SebC's proposed changes. :crossed_fingers: this fixes CI

rhaschke:
@sjahr: I added srdfdom to the upstream workspace, which should fix CI.

rhaschke:
Looks like workspace overlay doesn't work correctly: Although srdfdom is part of the upstream_ws, MoveIt still finds srdfdom headers in /opt/ros... Same issue in #3026.

mikeferguson:
I think there isn't anything we can do at the moment to get CI working - however, for the future, we can fix srdfdom to install it's headers slightly differently (into include/${PACKAGE_NAME} as noted in https://github.com/ros2/ros2/issues/1150) so that overlays could work - this is what most of the core ROS packages do to work around the underlay/overlay issue

sea-bass:
Worst case @JafarAbdi and I already put in a new release of srdfdom for the next sync, so... yeah.

sjahr:
https://github.com/moveit/moveit2/pull/3026#pullrequestreview-2361148434
> Do we need the selective include here in the moveit2 repo? As far as I understood, that was only necessary for srdfdom, which only has a single branch for ros2. Here, we have different branches for rolling and older releases.

The main branch is supposed to support multiple versions of ROS 2 because it acts as a "dev" branch that contains the latest (breaking) features and the distribution branches contain a stable version that is branched and only receives bug fixes (@tylerjw wrote a [nice article](https://moveit.ai/moveit%202/ros/2023/05/31/balancing-stability-and-development.html) about this in the moveit blog a while ago). Without the selective include we'd break support for Humble and Iron and personally, I'd be in favor of adding the selective includes at least for a year. Feel free to take what you need from https://github.com/moveit/moveit2/pull/3026

sea-bass (approved):
Let's get this in until `srdfdom` is released.

Can we have an issue or TODO reminder to re-enable this compiler warning though?

PR #3020, opened 2024-10-06
Fix Cartesian interpolation

rhaschke:
This is a follow-up on #2916, fixing a bug reported in https://github.com/moveit/moveit/issues/3651.

### DIFF
--- a/moveit_core/robot_state/src/cartesian_interpolator.cpp
+++ b/moveit_core/robot_state/src/cartesian_interpolator.cpp
@@ -292,7 +292,9 @@ CartesianInterpolator::Percentage CartesianInterpolator::computeCartesianPath(
     traj.insert(traj.end(), start, waypoint_traj.end());
 
     if (fabs(wp_percentage_solved - 1.0) < std::numeric_limits<double>::epsilon())
+    {
       percentage_solved = static_cast<double>(i + 1) / static_cast<double>(waypoints.size());
+    }
     else
     {
       percentage_solved += wp_percentage_solved / static_cast<double>(waypoints.size());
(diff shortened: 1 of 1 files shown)


sea-bass (approved):

sea-bass:
The build failure on Rolling here looks quite strange... also seemingly unrelated to this PR. Could this be a build caching thing?

rhaschke:
> The build failure on Rolling here looks quite strange... also seemingly unrelated to this PR. Could this be a build caching thing?

I had similar issues in #3003, which were fixed by deleting the upstream cache. However, here, the caches were not found anyway. So, I'm not sure why this fails for one rolling build but not the other.
It's definitely unrelated to the PR. I suggest giving it a try and merge.

rhaschke:
libsrdfdom doesn't link against tinyxml2:
```
ldd /opt/ros/rolling/lib/libsrdfdom.so.2.0.4
	linux-vdso.so.1 (0x00007ffd7d343000)
	libconsole_bridge.so.1.0 => /lib/x86_64-linux-gnu/libconsole_bridge.so.1.0 (0x000071a3689a5000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x000071a368728000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x000071a3686fb000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x000071a3684e9000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x000071a368400000)
```
Probably this wasn't noticed before, because the plugins (removed in #3024) pulled in this dependency.

rhaschke:
I have no idea why this PR still fails (for an unrelated reason), while the main branch works again.
`srdfdom/model.h` should be missing in both or none case.

Issue #3005, opened 2024-09-18
ServoCppFixture.PoseTest is flaky


rhaschke:
### Description

failure: https://github.com/moveit/moveit2/actions/runs/10915496106/job/30295360965
success: https://github.com/moveit/moveit2/actions/runs/10915496106

### ROS Distro

Rolling

### OS and version

Ubuntu 24.04

### Source or binary build?

Source

### If source, which branch?

main

### Which RMW are you using?

None

### Steps to Reproduce

run CI actions

### Expected behavior

tests should always succeed

### Actual behavior

sometimes the test fails

sea-bass:
The "bad" status code for this test indicates that the robot is halting for collision.

I guess picking a safer target pose could be helpful, but I think using a full IK solver instead of something simpler/more conducive to online control like a Jacobian pseudoinverse could also be a cause.

sea-bass:
Attempting something in https://github.com/moveit/moveit2/pull/3007 -- but unclear if this will help.

sea-bass:
> Attempting something in #3007 -- but unclear if this will help.

Just caught the flake again in this PR. Sadly, it does not fix it. Let's keep this issue open.

rhaschke:
I just saw another test failure: https://github.com/moveit/moveit2/actions/runs/11203600643/job/31141038313?pr=3020

mikeferguson:
Copying the text here - since the github CI runs eventually disappear: 

```
    ======================================================================
    FAIL: test_gtest_pass (moveit_servo.TestGTestProcessPostShutdown.test_gtest_pass)
    ----------------------------------------------------------------------
    Traceback (most recent call last):
      File "/home/runner/work/moveit2/moveit2/.work/target_ws/src/moveit2/moveit_ros/moveit_servo/tests/launch/servo_cpp_integration.test.py", line 133, in test_gtest_pass
        launch_testing.asserts.assertExitCodes(proc_info, process=servo_gtest)
      File "/opt/ros/jazzy/lib/python3.12/site-packages/launch_testing/asserts/assert_exit_codes.py", line 62, in assertExitCodes
        assert info.returncode in allowable_exit_codes, 'Proc {} exited with code {}'.format(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    AssertionError: Proc moveit_servo_cpp_integration_test-5 exited with code 1
    
    ----------------------------------------------------------------------
    Ran 1 test in 0.001s
    
[...]

PR #3024, opened 2024-10-09
Remove plugins from export set

pac48:
### Description

I think the plugin that are intended to be loaded by classloader or pluginlib should not be exported as the same exportset as the libraries that are intended to be dynamically linked. I ran into an issue where `libmoveit_ompl_planner_plugin.so` is being dynamically linked to `libmoveit_default_planning_response_adapter_plugins.so`, which breaks `MultiLibraryClassLoader`.

### DIFF
--- a/moveit_ros/planning/CMakeLists.txt
+++ b/moveit_ros/planning/CMakeLists.txt
@@ -49,8 +49,6 @@ set(THIS_PACKAGE_LIBRARIES
     moveit_collision_plugin_loader
     moveit_constraint_sampler_manager_loader
     moveit_cpp
-    moveit_default_planning_request_adapter_plugins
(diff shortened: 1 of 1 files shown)


sjahr (approved):
Makes sense, thanks!

pac48:
Looks like a servo test failed. Is it flaky?

sea-bass:
> Makes sense, thanks!

Yes. #3005 ... I couldn't figure out why after many attempts

sea-bass:
> Looks like a servo test failed. Is it flaky?

Yes. #3005 ... I couldn't figure out why after many attempts.

sjahr:
Not a blocker for this PR, just waited for CI to pass to be sure this PR doesn't introduce new issues :sweat_smile:

PR #3027, opened 2024-10-11
set filter state when no commands

pac48:
WIP


henrygerardmoore:
This fixes servo jumping into or out of collision very suddenly. If you servo into collision then try to servo out, you will note that the arm jumps, seemingly randomly, either out of collision or into collision (very dangerously on hardware). We should add a test for this to make sure there aren't regressions in this.

sjahr (approved):

mikeferguson:
Retriggered CI - same failure as another recent job - ticketed in #3045

PR #3032, opened 2024-10-15
Obtain time from node.

Jelmerdw:
### Description

If I use the moveit_servo servo_node in combination with Gazebo simulation, the `moveit_ros.current_state_monitor` complains that it doesn't receive robot state with recent timestamp. I launch all nodes with the 'use_sim_time' set to True, however the current_state_monitor still seems to use wall time. A similar problem was explained in this issue, although not for moveit_servo:

https://github.com/moveit/moveit2/issues/2906

My solution is to change the code from `rclcpp::Clock(RCL_ROS_TIME).now()` to `node_->now()`, such that the time is obtained from the node, which is launched with 'use_sim_time' set to True, so that it gives the sim time. This fixes my problem for the moveit_servo node, but will probably not fix the problem in the issue I mentioned.

This is my first contribution. I would like to hear feedback about this suggested change!

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/src/servo.cpp
+++ b/moveit_ros/moveit_servo/src/servo.cpp
@@ -648,8 +648,8 @@ KinematicState Servo::getCurrentRobotState(bool block_for_current_state) const
     bool have_current_state = false;
     while (rclcpp::ok() && !have_current_state)
     {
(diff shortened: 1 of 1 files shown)


sea-bass (approved):
This looks good to me -- thank you for this fix!

PR #3043, opened 2024-10-27
Switch to get for Dict lookup to prevent KeyError

brendandburns:
### Description

If you run the default generated configs for MoveIt on a headless system where `DISPLAY` isn't defined in the environment, this code crashes with a `KeyError`. Switch to use `dict.get(DISPLAY, "")` which returns `""` instead of
throwing.

### Checklist


mikeferguson (approved):
Thanks - I've actually run into this issue previously

PR #3064, opened 2024-11-05
Fix createTrajectoryMessage

pac48:


dyackzan (approved):

rhaschke (changes requested):
The inequality was correct here: If we reserve space for `n_samples`, `sample` needs to iterate from `0` to `n_samples-1`. Please provide a comment justifying this change.

pac48:
@rhaschke The reason this is needed is because the last point in the trajectory is not sampled when the frequency doesn't perfectly align with the trajectory length. For example, if we have a trajectory over the time interval [0, 0.9999] and a frequency of 2 Hz, we want 3 total time points: (0, 0.5, 0.9999). The current version would calculate `n_samples=2` and the generated time points would be (0.0, 0.5). 

This issue becomes evident when controlling a robot with velocity command interfaces. The trajectories generated are intended to end with zero velocity, but they do not due to the aforementioned issue.

I can see how this change can be implemented in a more concise way to minimize confusion. It may be better to round `static_cast<int>(trajectory.getDuration() / time_step)` to the nearest whole number to handle the case I mentioned.

sea-bass (approved):

sea-bass:
I've done this same math recently on `moveit_drake` and @pac48 's fix looks correct.

https://github.com/moveit/moveit_drake/blob/154e32b8804d240606bc10d56d13a870bb8c43b7/src/ktopt_planning_context.cpp#L185

gavanderhoorn:
Regardless of whether or not it's a correct fix, PRs without a proper motivation should really not be approved 1 minute after they are submitted.

It's clear you are in the same company, so likely discussed this approach and both were OK with it, but that's not the point.

MoveIt is still an OSS project as far as we know -- well, these parts are anyway -- so let's try to adhere to the agreed upon workflow.

rhaschke (approved):
This looks much better. Thanks.

PR #3069, opened 2024-11-08
Fix jacobian calculation

sjahr:
### Description

@marioprats and I encountered an error in the Jacobian calculation for a link that is not the tip link. In that case all joint in the chain after the reference link don't have any impact on it's motion and thus, the related Jacobian entries need to zero. This PR fixes that + adds a unittest to catch this in the future.

### Checklist

### DIFF
--- a/moveit_core/robot_state/src/robot_state.cpp
+++ b/moveit_core/robot_state/src/robot_state.cpp
@@ -1496,19 +1496,25 @@ bool RobotState::getJacobian(const JointModelGroup* group, const LinkModel* link
   const int rows = use_quaternion_representation ? 7 : 6;
   const int columns = group->getVariableCount();
   jacobian.resize(rows, columns);
+  jacobian.setZero();
 
   // Get the tip pose with respect to the group root link. Append the user-requested offset 'reference_point_position'.
   const Eigen::Isometry3d root_pose_tip = root_pose_world * getGlobalLinkTransform(link);
   const Eigen::Vector3d tip_point = root_pose_tip * reference_point_position;
 
-  // Here we iterate over all the group active joints, and compute how much each of them contribute to the Cartesian
-  // displacement at the tip. So we build the Jacobian incrementally joint by joint.
-  std::size_t active_joints = group->getActiveJointModels().size();
+  // Initialize the column index of the Jacobian matrix.
   int i = 0;
-  for (std::size_t joint = 0; joint < active_joints; ++joint)
+
+  // Here we iterate over all the group active joints, and compute how much each of them contribute to the Cartesian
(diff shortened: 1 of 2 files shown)


marioprats (approved):
LGTM, and thanks for adding to the test!

sea-bass (dismissed):
LGTM -- so these values were formerly undefined? Good catch.

(something something use Pinocchio instead)

sea-bass:
I really need to look at CI results before reviewing... seems like this fails on Jazzy/Rolling for some reason?

sjahr:
> I really need to look at CI results before reviewing... seems like this fails on Jazzy/Rolling for some reason?

Wow, this is bad :see_no_evil: Why would it fail on rolling and iron but not on Humble. There is not ROS supposed to be involved in core

sea-bass:
> Why would it fail on rolling and iron but not on Humble. There is not ROS supposed to be involved in core

No, but new Ubuntu version might mean newer dependencies like Eigen etc.

Probably best to repro these locally on a 24.04 setup

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -1496,19 +1496,25 @@ bool RobotState::getJacobian(const JointModelGroup* group, const LinkModel* link
   const int rows = use_quaternion_representation ? 7 : 6;
   const int columns = group->getVariableCount();
```
sea-bass, line 1497: ah! columns should not be the GROUP's variable count. Should be the entire robot?

sjahr:
Very dump mistake, let's see if this fixes it. Interestingly, this error did not seem to happen in older Eigen versions :thinking:

sea-bass (approved):
Oh --  it was the test! Nice.

rhaschke (commented):
The new code implicitly assumes a linear kinematic chain. It does not work for a tree, like two arms or multi-fingered hands.

sea-bass:
> The new code implicitly assumes a linear kinematic chain. It does not work for a tree, like two arms or multi-fingered hands.

Good catch! This would also be a good unit test to add.

sea-bass:
> > The new code implicitly assumes a linear kinematic chain. It does not work for a tree, like two arms or multi-fingered hands.
> 
> Good catch! This would also be a good unit test to add.

Update: I just tried a simple test case to see what happened, and there is a specific error that tells you getting Jacobians from non-chain groups is already not supported. So I guess we should make a separate issue to support this.

https://github.com/moveit/moveit2/blob/4c28a74343e1d8470689cae9e23eb30c91bdcb4d/moveit_core/robot_state/src/robot_state.cpp#L1467-L1472

Also... Pinocchio, anyone? :)

PR #3070, opened 2024-11-08
Enhancement/ports moveit 3522

rr-tom-noble:
### Description

Ports https://github.com/moveit/moveit/pull/3522 to moveit2. Should fix https://github.com/moveit/moveit2/issues/2857

### Checklist
- [x] **Required by CI**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)
- [x] Extend the tutorials / documentation [reference](http://moveit.ros.org/documentation/contributing/) (fixes bug)
- [x] Document API changes relevant to the user in the [MIGRATION.md](https://github.com/moveit/moveit2/blob/main/MIGRATION.md) notes (fixes bug)
- [ ] Create tests, which fail without this PR [reference](https://moveit.picknik.ai/humble/doc/examples/tests/tests_tutorial.html)
- [x] Include a screenshot if changing a GUI (No GUI changes)
- [ ] While waiting for someone to review your request, please help review [another open pull request](https://github.com/moveit/moveit2/pulls) to support the maintainers

[//]: # "You can expect a response from a maintainer within 7 days. If you haven't heard anything by then, feel free to ping the thread. Thank you!"

PR #3076, opened 2024-11-09
Fixes humble backport issues (#3070)

TSNoble:
### Description

Backport fixes for #3070


TSNoble:
@sjahr I'm unsure why a planning scene monitor test failed as I've only touched pilz. Would it be possible to rerun the pipeline in case it's sporadic?

sea-bass:
Looks good. Seems to also be needed for the Iron backport PR, by the way!

PR #3088, opened 2024-11-14
Add use_padding flag + deprecate checkCollisionUnpadded() functions

sjahr:
### Description

I wanted to perform a padded self-collision check with a robot state and realized that this is not possible with the current API. So I decided to enable it + clean-up the checkCollision API of the planning scene a bit. This PR:
- Adds use_padded_collision_environment & use_padded_self_collision flags to choose the environment to be used for collision checks
- Clean up & update checkCollision(): Move all implementations into header make transform update more consistent
- Deprecate checkCollisionUnpadded() since they become redundant with the new flags
- Clean up & update checkSelfCollision(): Move all implementations into header make transform update more consistent

### Checklist

### DIFF
--- a/moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
+++ b/moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
@@ -151,10 +151,10 @@ struct CollisionRequest
   std::string group_name = "";
 
   /** \brief If true, use padded collision environment */
-  bool use_padded_collision_environment = true;
+  bool pad_environment_collisions = true;
 
   /** \brief If true, do self collision check with padded robot links */
-  bool use_padded_self_collision = false;
+  bool pad_self_collisions = false;
 
   /** \brief If true, compute proximity distance */
   bool distance = false;
--- a/moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
+++ b/moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
@@ -150,6 +150,12 @@ struct CollisionRequest
    * are included. */
   std::string group_name = "";
 
+  /** \brief If true, use padded collision environment */
+  bool use_padded_collision_environment = true;
+
+  /** \brief If true, do self collision check with padded robot links */
+  bool use_padded_self_collision = false;
+
   /** \brief If true, compute proximity distance */
   bool distance = false;
 
--- a/moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
+++ b/moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
@@ -347,19 +347,12 @@ public:
   void checkCollision(const collision_detection::CollisionRequest& req, collision_detection::CollisionResult& res);
 
   /** \brief Check whether the current state is in collision. The current state is expected to be updated. */
-  void checkCollision(const collision_detection::CollisionRequest& req, collision_detection::CollisionResult& res) const
-  {
-    checkCollision(req, res, getCurrentState());
-  }
+  void checkCollision(const collision_detection::CollisionRequest& req, collision_detection::CollisionResult& res) const;
 
   /** \brief Check whether a specified state (\e robot_state) is in collision. This variant of the function takes
       a non-const \e robot_state and calls updateCollisionBodyTransforms() on it. */
   void checkCollision(const collision_detection::CollisionRequest& req, collision_detection::CollisionResult& res,
-                      moveit::core::RobotState& robot_state) const
-  {
-    robot_state.updateCollisionBodyTransforms();
-    checkCollision(req, res, static_cast<const moveit::core::RobotState&>(robot_state));
-  }
+                      moveit::core::RobotState& robot_state) const;
 
   /** \brief Check whether a specified state (\e robot_state) is in collision. The collision transforms of \e
    * robot_state are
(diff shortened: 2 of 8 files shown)


moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
```diff
@@ -150,6 +150,12 @@ struct CollisionRequest
    * are included. */
   std::string group_name = "";
 
+  /** \brief If true, use padded collision environment */
+  bool use_padded_collision_environment = true;
```
sea-bass, line 154: Should these names be more consistent with each other?

Like `pad_environment_collisions` and `pad_self_collisions`?

Or `use_padded_environment_collisions` and `use_padded_self_collisions`?

moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
```diff
@@ -372,114 +365,80 @@ class MOVEIT_PLANNING_SCENE_EXPORT PlanningScene : public std::enable_shared_fro
...
-    checkCollision(req, res, static_cast<const moveit::core::RobotState&>(robot_state), acm);
-  }
+                      const collision_detection::AllowedCollisionMatrix& acm) const;
 
   /** \brief Check whether a specified state (\e robot_state) is in collision, with respect to a given
       allowed collision matrix (\e acm). */
   void checkCollision(const collision_detection::CollisionRequest& req, collision_detection::CollisionResult& res,
                       const moveit::core::RobotState& robot_state,
-                      const collision_detection::AllowedCollisionMatrix& acm) const;
+                      const collision_detection::AllowedCollisionMatrix& acm, bool validate_transforms = true) const;
```
sea-bass, line 374: Add this to docstring

moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
sea-bass: Also, is it intentional for the defaults to be different? I assume the defaults maintain the old behavior?

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -411,58 +417,147 @@ void PlanningScene::checkCollision(const collision_detection::CollisionRequest&
...
 
-  // do self-collision checking with the unpadded version of the robot
-  if (!res.collision || (req.contacts && res.contacts.size() < req.max_contacts))
+void PlanningScene::checkSelfCollision(const collision_detection::CollisionRequest& req,
+                                       collision_detection::CollisionResult& res)
+{
+  checkSelfCollision(req, res, getCurrentStateNonConst());
+}
+
+/** \brief Check whether the current state is in self collision */
```
sea-bass, line 518: This is a .cpp file, no need for a docstring here, right?

sea-bass (commented):
Looks good, just small comments. But my first one on variable naming is probably the more important.

Also, this change is probably worth adding to the corresponding CHANGELOG.rst

moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
sjahr: Yes, exactly that's why I made them different. At the moment, we're using by default padding for environment collision checks and no padding for self-collision checks

sea-bass:
Can you also add to the CHANGELOG.rst?

sjahr:
> Can you also add to the CHANGELOG.rst?

@sea-bass Sorry, missed that but now it's done :heavy_check_mark:

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -411,58 +417,146 @@ void PlanningScene::checkCollision(const collision_detection::CollisionRequest&
...
+    RCLCPP_WARN(getLogger(), "Robot state has dirty collision body transforms. Please update the collision body "
+                             "transforms before checking collision.");
+  }
   // check collision with the world using the padded version
-  getCollisionEnv()->checkRobotCollision(req, res, robot_state, acm);
+  req.pad_environment_collisions ? getCollisionEnv()->checkRobotCollision(req, res, robot_state, acm) :
+                                   getCollisionEnvUnpadded()->checkRobotCollision(req, res, robot_state, acm);
+
+  // Return early if a collision was found or the maximum number of allowed contacts is exceeded
+  if (res.collision || (req.contacts && res.contacts.size() >= req.max_contacts))
```
marioprats, line 447: I don't fully understand this check.
If there's a collision, it returns right away.
If there's no collision, wouldn't `res.contacts.size()` be zero?
So if there's no env collision, but `req.contacts==true` and `req.max_contacts==0`, wouldn't this return without checking self-collisions?

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -411,58 +417,146 @@ void PlanningScene::checkCollision(const collision_detection::CollisionRequest&
...
 }
 
 void PlanningScene::checkCollision(const collision_detection::CollisionRequest& req,
                                    collision_detection::CollisionResult& res,
                                    const moveit::core::RobotState& robot_state,
-                                   const collision_detection::AllowedCollisionMatrix& acm) const
+                                   const collision_detection::AllowedCollisionMatrix& acm,
+                                   bool validate_transforms) const
 {
+  if (validate_transforms && robot_state.dirtyCollisionBodyTransforms())
```
marioprats, line 437: Do we really need `validate_transforms` if it's just to show a warning? i.e. it's not making this function fail in any way.
And `dirtyCollisionBodyTransforms` should be a cheap check anyway since it's just checking an internal member variable of the RobotState.

moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
```diff
@@ -372,114 +365,80 @@ class MOVEIT_PLANNING_SCENE_EXPORT PlanningScene : public std::enable_shared_fro
...
-                      const collision_detection::AllowedCollisionMatrix& acm) const
-  {
-    robot_state.updateCollisionBodyTransforms();
-    checkCollision(req, res, static_cast<const moveit::core::RobotState&>(robot_state), acm);
-  }
+                      const collision_detection::AllowedCollisionMatrix& acm) const;
 
   /** \brief Check whether a specified state (\e robot_state) is in collision, with respect to a given
-      allowed collision matrix (\e acm). */
+      allowed collision matrix (\e acm). If validate_transforms is set true, the given robot state is checked for dirty transforms. */
```
marioprats, line 371: See my other comment, I find this API confusing since `validate_transforms` is not making this function fail or succeed.

marioprats (commented):
Thanks Sebastian, nice cleanup!
Just a couple of questions.

moveit_core/planning_scene/src/planning_scene.cpp
sjahr: :thinking: Good point, would it be better to check self-collisions first because that should always be evaluated? I did not change that part so I am not sure if there is a good reason why it is the other way around.

moveit_core/planning_scene/src/planning_scene.cpp
sjahr: I don't like about this function that you could silently do the wrong thing if you're not update the collision transforms :thinking: However, I wanted to give the option to disable the warning with this flag (e.g. if you run this in a fast loop). On second thought, it might be a better idea to warn on default and the user needs to actively disable the warning. Alternatively,  we could do a bigger change and return an error message with the collision result. I am hesitant to do this in this PR though to get this change in first

moveit_core/planning_scene/src/planning_scene.cpp
marioprats: Looking again, I think the previous check is correct, we either have to fix this negated version, or use the previous

sea-bass:
Sorry @sjahr -- I was putting together a Humble release and realized it's not CHANGELOG.rst that needs modification, as that is put together by `catkin_generate_changelog`.

Could you move that stuff to `doc/MIGRATION_GUIDE.md`?

Also changes LGTM, but will let @marioprats have the approval since he still had some open questions.

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -438,7 +438,7 @@ void PlanningScene::checkCollision(const collision_detection::CollisionRequest&
                                    getCollisionEnvUnpadded()->checkRobotCollision(req, res, robot_state, acm);
 
   // Return early if a collision was found or the maximum number of allowed contacts is exceeded
```
marioprats, line 440:
```
// Return early if a collision was found and the number of contacts found already exceed `req.max_contacts`, if `req.contacts` is enabled.
```

marioprats (approved):
Thanks Sebastian, just a nit to update the comment, since the PR changes the check to return early

sea-bass (approved):

marioprats (approved):

PR #3126, opened 2024-11-22
[Humble] Add sleep in setup of flaky PSM test fixture

sea-bass:
### Description

This is the Humble version of https://github.com/moveit/moveit2/pull/3124, in which the test was never disabled and continued to be flaky.

See my log of successful retries: https://github.com/moveit/moveit2/actions/runs/11966024737?pr=3126

### Checklist


TSNoble:
@sea-bass 

Could we replace the sleep with waitForService call(s)? Additionally, could we keep the issue open, since this is more of a workaround than a fix of the race condition?

sjahr:
> @sea-bass
> 
> Could we replace the sleep with waitForService call(s)? Additionally, could we keep the issue open, since this is more of a workaround than a fix of the race condition?

:+1: Let's merge this workaround for now and keep the issue open until it is properly solved

sjahr (approved):

PR #3137, opened 2024-11-28
Test fixes to `geometric_shapes` in CI

sea-bass:
### Description

### Checklist

### DIFF
--- a/moveit2.repos
+++ b/moveit2.repos
@@ -10,7 +10,7 @@ repositories:
     version: ros2
   moveit_resources:
     type: git
-    url: https://github.com/moveit/moveit_resources.git
+    url: https://github.com/rhaschke/geometric_shapes.git
     version: ros2
   srdfdom:
     type: git


rhaschke:
Note that https://github.com/moveit/geometric_shapes/pull/253 only takes effect after a new release or if the binary package (and its headers in /opt/ros/*/include/geometric_shapes/*) are removed.

sea-bass:
> Note that [moveit/geometric_shapes#253](https://github.com/moveit/geometric_shapes/pull/253) only takes effect after a new release or if the binary package (and its headers in /opt/ros/_/include/geometric_shapes/_) are removed.

I see. So should we pin to the "hack" commit for now, and cut a release of `geometric_shapes` for the next sync?

rhaschke:
We just need to remove the installed binary package.

sea-bass:
Seems that alone didn't quite work, unless there's some caching that is persisting

sea-bass:
Nah, it just does not have binary installs to remove, it seems

```
  Reading state information... Done
  E: Unable to locate package ros-.*-geometric-shapes
  E: Couldn't find any package by glob 'ros-.*-geometric-shapes'
```

https://github.com/moveit/moveit2/actions/runs/12072620084/job/33666989895?pr=3137#step:11:261

rhaschke:
For some reason, the glob didn't work. Using the exact name should work now.

rhaschke:
And of course we need `-y` to avoid apt waiting for user input...

rhaschke:
This works. Reverting change to moveit2.repos...

rhaschke (approved):
This can be merged. After releasing and syncing geometric_shapes, the commit should be reverted.

sea-bass:
> For some reason, the glob didn't work. Using the exact name should work now.

I think apt might use `*` instead of `.*` for its glob, but anyways I like specifying it with the ROS distro environment variable better.

rhaschke:
> I think apt might use `*` instead of `.*` for its glob

No, surprisingly all glob/regex variants didn't work.

PR #3142, opened 2024-11-29
Do not perform most Jazzy CI jobs on main branch

sea-bass:
### Description

This is a complementary PR to https://github.com/moveit/moveit2/pull/3141.

Jazzy Docker and tutorial images will now be built off the `jazzy` branch.

### Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -28,8 +28,6 @@ jobs:
             ROS_DISTRO: humble
           - IMAGE: jazzy-ci
             ROS_DISTRO: jazzy
-          - IMAGE: jazzy-ci-testing
-            ROS_DISTRO: jazzy
     env:
       # TODO(andyz): When this clang-tidy issue is fixed, remove -Wno-unknown-warning-option
       # https://stackoverflow.com/a/41673702
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ROS_DISTRO: [rolling, jazzy]
+        ROS_DISTRO: [rolling]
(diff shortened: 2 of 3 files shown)


mikeferguson (approved):

.github/workflows/ci.yaml
```diff
@@ -28,8 +28,6 @@ jobs:
             ROS_DISTRO: humble
           - IMAGE: jazzy-ci
             ROS_DISTRO: jazzy
-          - IMAGE: jazzy-ci-testing
-            ROS_DISTRO: jazzy
```
mikeferguson, line 32: Wait - should we leave this part? So that we are still testing main branch against Jazzy?

sea-bass: I was wondering about this -- what is the difference between `jazzy-ci` and `jazzy-ci-testing`? Because I just removed the latter

sea-bass: Ah it's for the Docker images

Regular builds a docker `FROM ros:${ROS_DISTRO}-ros-base`

Testing builds it `FROM osrf/ros2:testing`

... so it seems like the testing repo should only really be happening on rolling if I understand correctly?

mikeferguson: Ah yes, I saw the humble-ci above, but missed that there was still a jazzy-ci above. All good.

sea-bass: It does raise the question of adding a `rolling-ci-testing` job though?

mikeferguson: I'm guessing the thought is that rolling-testing is just too volatile and things would be breaking all the time rather than every sync?

sea-bass: Yeah, I guess we can always add jobs back as needed. Thanks for checking!

PR #3143, opened 2024-11-30
Fix logic in CheckStartStateBounds adapter

### LINKED ISSUE #2924: CheckStartStateBounds does not support REVOLUTE and NOT Continuous type joints
### Description
CheckStartStateBounds does not support REVOLUTE and NOT Continuous type joints

### Your environment
* ROS Distro: Jazzy
* OS Version: e.g. Ubuntu 24.04
* Binary build 2.10.0

### Steps to reproduce
- Set the initial posture of the actual robot to a position beyond the urdf limit.
- Press “plan” button in rviz to plan the trajectory.

### Expected behaviour
CheckStartStateBounds is not SUCCESS. And fixed to joint limit in case fix_start_state is true.

### Actual behaviour
CheckStartStateBounds is SUCCESS.

### Details

In https://github.com/moveit/moveit2/pull/2429, CheckStartStateBounds is refactored.
The logic has also been changed so that the check is skipped for non-infinite rotation joints, which are used in most arm joints.

https://github.com/moveit/moveit2/blob/a7fe0df4c565db15710581bdcba1d33d70aa0ddb/moveit_ros/planning/planning_request_adapter_plugins/src/check_start_state_bounds.cpp#L105-L120

sea-bass:
### Description

This PR aims to fix the logic in the `CheckStartStateBounds` adapter, as @chama1176 pointed out.

It also adds some basic unit tests to prevent further breakages and maintain expected behavior.

Closes #2924

### Checklist

### DIFF
--- a/moveit_ros/planning/planning_request_adapter_plugins/src/check_start_state_bounds.cpp
+++ b/moveit_ros/planning/planning_request_adapter_plugins/src/check_start_state_bounds.cpp
@@ -188,8 +188,9 @@ public:
     }
     else if (params.fix_start_state && should_fix_state)
     {
-      status.message = std::string("Changing start state.");
-      RCLCPP_WARN(logger_, status.message.c_str());
+      constexpr auto msg_string = "Changing start state.";
+      status.message = msg_string;
+      RCLCPP_WARN(logger_, msg_string);
       moveit::core::robotStateToRobotStateMsg(start_state, req.start_state);
     }
     return status;
--- a/moveit_ros/planning/planning_request_adapter_plugins/src/check_start_state_bounds.cpp
+++ b/moveit_ros/planning/planning_request_adapter_plugins/src/check_start_state_bounds.cpp
@@ -188,7 +188,7 @@ public:
     }
     else if (params.fix_start_state && should_fix_state)
     {
-      constexpr auto msg_string = "Changing start state.";
+      constexpr auto msg_string = "Normalized start state.";
       status.message = msg_string;
       RCLCPP_WARN(logger_, msg_string);
       moveit::core::robotStateToRobotStateMsg(start_state, req.start_state);
--- a/moveit_ros/planning/planning_request_adapter_plugins/src/check_start_state_bounds.cpp
+++ b/moveit_ros/planning/planning_request_adapter_plugins/src/check_start_state_bounds.cpp
@@ -94,7 +94,8 @@ public:
     // Read parameters
     const auto params = param_listener_->get_params();
 
-    bool changed_req = false;
+    bool should_fix_state = false;
+    bool is_out_of_bounds = false;
     for (const moveit::core::JointModel* jmodel : jmodels)
     {
       // Check if we have a revolute, continuous joint. If we do, then we only need to make sure
@@ -113,7 +114,7 @@ public:
             double after = start_state.getJointPositions(jmodel)[0];
             if (fabs(initial - after) > std::numeric_limits<double>::epsilon())
             {
-              changed_req = true;
+              should_fix_state |= true;
             }
           }
           break;
@@ -126,7 +127,7 @@ public:
           if (static_cast<const moveit::core::PlanarJointModel*>(jmodel)->normalizeRotation(copy))
           {
             start_state.setJointPositions(jmodel, copy);
-            changed_req = true;
+            should_fix_state |= true;
           }
           break;
         }
@@ -138,7 +139,7 @@ public:
           if (static_cast<const moveit::core::FloatingJointModel*>(jmodel)->normalizeRotation(copy))
           {
             start_state.setJointPositions(jmodel, copy);
-            changed_req = true;
+            should_fix_state |= true;
           }
           break;
         }
@@ -151,6 +152,8 @@ public:
       // Check the joint against its bounds.
       if (!start_state.satisfiesBounds(jmodel))
       {
+        is_out_of_bounds |= true;
+
         std::stringstream joint_values;
         std::stringstream joint_bounds_low;
         std::stringstream joint_bounds_hi;
@@ -173,26 +176,22 @@ public:
       }
     }
 
-    // If we made any changes, consider using them.
-    if (params.fix_start_state && changed_req)
-    {
-      RCLCPP_WARN(logger_, "Changing start state.");
-      moveit::core::robotStateToRobotStateMsg(start_state, req.start_state);
-      return moveit::core::MoveItErrorCode(moveit_msgs::msg::MoveItErrorCodes::SUCCESS, std::string(""),
-                                           getDescription());
-    }
-
+    // Package up the adapter result, changing the state if applicable.
     auto status = moveit::core::MoveItErrorCode();
-    if (!changed_req)
-    {
-      status.val = moveit_msgs::msg::MoveItErrorCodes::SUCCESS;
-    }
-    else
+    status.source = getDescription();
+    status.val = moveit_msgs::msg::MoveItErrorCodes::SUCCESS;
+
+    if (is_out_of_bounds || (!params.fix_start_state && should_fix_state))
     {
       status.val = moveit_msgs::msg::MoveItErrorCodes::START_STATE_INVALID;
       status.message = std::string("Start state out of bounds.");
     }
-    status.source = getDescription();
+    else if (params.fix_start_state && should_fix_state)
+    {
+      status.message = std::string("Changing start state.");
+      RCLCPP_WARN(logger_, status.message.c_str());
+      moveit::core::robotStateToRobotStateMsg(start_state, req.start_state);
+    }
     return status;
   }
 
--- a/moveit_ros/planning/planning_request_adapter_plugins/src/check_start_state_bounds.cpp
+++ b/moveit_ros/planning/planning_request_adapter_plugins/src/check_start_state_bounds.cpp
@@ -115,9 +115,8 @@ public:
             {
               changed_req = true;
             }
-            break;
           }
-          [[fallthrough]];  // If the revolute joint is non-continuous, we want to enter default case.
+          break;
         }
         // Normalize yaw; no offset needs to be remembered
         case moveit::core::JointModel::PLANAR:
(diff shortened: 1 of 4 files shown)


sea-bass:
@sjahr I also want to check whether it's intentional that we're not checking the bounds for planar/floating joints at all?

Is this because URDF also doesn't specify limits for this joint type?

moveit_ros/planning/planning_request_adapter_plugins/src/check_start_state_bounds.cpp
```diff
@@ -115,8 +115,9 @@ class CheckStartStateBounds : public planning_interface::PlanningRequestAdapter
             {
               changed_req = true;
             }
+            break;
           }
-          break;
+          [[fallthrough]];  // If the revolute joint is non-continuous, we want to enter default case.
```
mikeferguson, line 120: don't we need to re-order this so that case REVOLUTE is immediately before the 'default'?

mikeferguson:
Posted this in the issue ticket - but:

The original logic was:

- for each joint model:
  - if REVOLUTE & continuous, call enforceBounds() [OK]
  - else
     - if PLANAR - normalizeRotation [OK]
     - else FLOATING - normalizeRotation [OK]
- read parameters
- for each joint model:
   - apply the satisfiesBounds logic

I think we actually just need to pull the "default" case outside the switch statement and do it for all joints?

mikeferguson:
The fact that this is broken, also makes me think we should really have a test for this... since it seems pretty important to not violate the joint limits in a motion planning software...

moveit_ros/planning/planning_request_adapter_plugins/src/check_start_state_bounds.cpp
sea-bass: Yeah, I think so.

For some reason I thought C++ kept checking the other cases, and not execute them all.

sea-bass:
> The fact that this is broken, also makes me think we should really have a test for this... since it seems pretty important to not violate the joint limits in a motion planning software...

I also didn't look at this... and there are no tests for any of the adapters? Oof. Yeah, I'd like to add one.

sea-bass:
Hah, yeah. It shouldn't be that bad to create a planning scene from e.g. a PR2 or Panda and add some quick tests for the "revolute joint out of bounds" case.

It's more the other paths like continuous joints that will likely require making more robot models.

I'm internally debating how much to bite off here...

mikeferguson:
PR2 has continuous joints in the forearm/wrist roll and it looks like the SRDF in moveit_resources has a planar "world_joint"

sea-bass:
I added some tests and the logic is... more broken than we thought.

@sjahr @mikeferguson do you guys mind lending me a hand w.r.t. what the expected behavior should be?

moveit_ros/planning/planning_request_adapter_plugins/src/check_start_state_bounds.cpp
```diff
@@ -144,33 +144,36 @@ class CheckStartStateBounds : public planning_interface::PlanningRequestAdapter
...
-                         "the range [%s], [%s].",
-                         jmodel->getName().c_str(), joint_values.str().c_str(), joint_bounds_low.str().c_str(),
-                         joint_bounds_hi.str().c_str());
-          }
+          break;  // do nothing
+        }
+      }
+
+      // Check the joint against its bounds.
+      if (!start_state.satisfiesBounds(jmodel))
```
sea-bass, line 152: Should this condition set `changed_req` to `true` and actually update the start_state variable to be within bounds using `enforceBounds()`?

The old code had this so long as the state was within some specified parameter threshold, which has since been removed.

mikeferguson: There was a discussion in https://github.com/moveit/moveit2/issues/2849

@sjahr stated "In main I replaced the FixStartState adapter in https://github.com/moveit/moveit2/pull/2429 with a simple [CheckStartStateCollision](https://github.com/moveit/moveit2/blob/main/moveit_ros/planning/planning_request_adapter_plugins/src/check_start_state_collision.cpp) adapter. My rationale is that an adapter that introduces small jumps is un-intuitive and might lead to undesired behavior, especially given that is was used almost everywhere as a default. I think the decision about what to do when the start state is in collision should be made in a higher level logic than the planning pipeline."

mikeferguson:
So, my understanding (taking into account the notes in 2849) is that, the following should happen:

 * REVOLUTE:
   * continuous - make sure the value is inside the limits (usually +/-pi) - if fix_start_state then basically wrap it, else fail
   * non-continuous - make sure the value is inside the limits - never fix it though (this is the change in behavior from the old days)
 * PLANAR: make sure yaw is normalized - if fix_start_state then fix it, else fail.
 * FLOATING: make sure quaternion rotation is normalized - if fix_start_state then fix it, else fail.

So then, the logic really is:

 - first loop through all joints, check PLANAR, FLOATING, continuous REVOLUTE - figure out how to fix them if needed, then mark 'changed_req' (by the way, this used to be "change_req" before the refactor - not sure why the "d" was added - it's not really past tense when it is set...)
 - important note: when PLANAR, FLOATING or continuous REVOLUTE joints are "change_req" we actually update the `start_state` variable.
 - The call to `start_state.satisifesBounds(jmodel)` will then find non-continuous joints that are invalid (we already updated `start_state` with the fixes above). It looks like we need to add a proper exit here and fail if this isn't true. There is also a huge print statement here - we shouldn't duplicate that, so I think the logic should actually be: `if (!start_state.satisfiesBounds(jmodel) || (change_req && !params.fix_start_state)) { // print and fail }` to catch the case that we need to fix the start state for PLANAR, FLOATING or continuous REVOLUTE.
 - Once we've made it past that now larger if statement, we can simply return success, we're all set...

sea-bass:
That's about in line with what I was thinking. Meaning the doc in the parameters YAML also needs to fixed.

Will wait til Sebastian confirms this behavior and will clean up / fill out the test suite as well.

Thanks!

moveit_ros/planning/planning_request_adapter_plugins/test/test_check_start_state_bounds.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+  const auto result = adapter_->adapt(planning_scene_, request);
+  EXPECT_EQ(result.val, moveit_msgs::msg::MoveItErrorCodes::SUCCESS);
+  EXPECT_EQ(result.message, "Changing start state.");
+
+  // Validate that the start state in the request was actually changed.
+  const auto& joint_names = request.start_state.joint_state.name;
+  const size_t joint_idx =
+      std::find(joint_names.begin(), joint_names.end(), "r_forearm_roll_joint") - joint_names.begin();
+  EXPECT_NEAR(request.start_state.joint_state.position[joint_idx], -0.530965, 1.0e-4);
```
sea-bass, line 156: this is

```python
>>> import math
>>> math.pi - math.fmod(100.0, math.pi)
0.5309649148733797
```

moveit_ros/planning/planning_request_adapter_plugins/src/check_start_state_bounds.cpp
```diff
@@ -138,58 +139,60 @@ class CheckStartStateBounds : public planning_interface::PlanningRequestAdapter
...
+
+    if (is_out_of_bounds || (!params.fix_start_state && should_fix_state))
     {
       status.val = moveit_msgs::msg::MoveItErrorCodes::START_STATE_INVALID;
       status.message = std::string("Start state out of bounds.");
     }
-    status.source = getDescription();
+    else if (params.fix_start_state && should_fix_state)
+    {
+      constexpr auto msg_string = "Changing start state.";
```
mikeferguson, line 191: perhaps this should really be "Normalized start state" since that's all we actually change these days?

mikeferguson (approved):

moveit_ros/planning/planning_request_adapter_plugins/src/check_start_state_bounds.cpp
sjahr: I'd still state that, what do you think?

moveit_ros/planning/planning_request_adapter_plugins/src/check_start_state_bounds.cpp
```diff
@@ -138,58 +139,60 @@ class CheckStartStateBounds : public planning_interface::PlanningRequestAdapter
...
+      status.message = msg_string;
+      RCLCPP_WARN(logger_, msg_string);
```
sjahr, line 193: Is it duplicate to print + send the status message?

sjahr (approved):
Thanks for fixing this (and sorry for breaking it :see_no_evil:). My comment's aren't blocking the merge. I think this looks good

moveit_ros/planning/planning_request_adapter_plugins/src/check_start_state_bounds.cpp
sea-bass: Is that because the message is printed by the planning pipeline on failures?

mikeferguson: Most people won't see the status message, but will see the logging. I suppose the status message might be useful for some automated response in downstream user code?

sea-bass: No, I mean it's because the logging happens on the return of the planning adapter here: https://github.com/moveit/moveit2/blob/e7872ebf92146bb80930ec98cf6c181e8384e99c/moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp#L284-L290

sea-bass: ... although in this case it actually won't double print because this warning happens alongside a `SUCCESS` status code. So I think we do want to keep it here.

moveit_ros/planning/planning_request_adapter_plugins/src/check_start_state_bounds.cpp
sea-bass: This is fine!

sjahr (approved):

sea-bass:
Uhh... why is a bunch of new stuff failing suddenly?

Will we ever get some peace in this repo? Lol

sea-bass:
I deleted some GHA caches and the merqe queue jobs seem to be working now. 🤷🏻

mikeferguson (approved):

PR #3144, opened 2024-12-01
Allow attached objects in move_group's /compute_ik pose frames

ANogin:
### Description

In the existing `move_group` implementation, the pose frame in `PositionIKRequest`'s `pose_stamped` and `pose_stamped_vector` were converted to robot's root frame using TF2, which meant that the calls to `/compute_ik` service would fail with `FRAME_TRANSFORM_FAILURE` when one tried to use an attached object frame (or subframe) there. This changes the code to use `RobotState::getFrameTransform` to convert those poses to the robot's root frame instead, enabling the use of attached object frames/subframes.

### Checklist

### DIFF
--- a/moveit_ros/move_group/src/default_capabilities/kinematics_service_capability.cpp
+++ b/moveit_ros/move_group/src/default_capabilities/kinematics_service_capability.cpp
@@ -91,7 +91,6 @@ void MoveGroupKinematicsService::computeIK(moveit_msgs::msg::PositionIKRequest& 
     {
       moveit::core::robotStateMsgToRobotState(req.robot_state, rs);
     }
-    const std::string& default_frame = context_->planning_scene_monitor_->getRobotModel()->getModelFrame();
 
     if (req.pose_stamped_vector.empty() || req.pose_stamped_vector.size() == 1)
     {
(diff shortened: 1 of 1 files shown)


sea-bass (commented):
This overall looks good, but I am concerned that it breaks the behavior where the target pose does not specify a frame ID.

As seen in https://github.com/moveit/moveit2/blob/f82cdcdccf471dd25550578098c4bb7e074a5535/moveit_ros/move_group/src/move_group_capability.cpp#L203, the `performTransform()` function will fall back to using that default frame if the header has no frame id.

It might be beneficial to reinstate this behavior, and maybe also the previpus condition where frame ID being equal to the model frame resulting in a no-op transform.

sea-bass:
Also, does `RobotState::getFrameTransform()` actually support getting frames not in the robot model, like available as external frames in the TF tree?

This may also be another major difference between the existing and proposed implementations, which would also break for any non-robot model reference frames.

ANogin:
> Also, does `RobotState::getFrameTransform()` actually support getting frames not in the robot model, like available as external frames in the TF tree?
> 
> This may also be another major difference between the existing and proposed implementations, which would also break for any non-robot model reference frames.

Good question - I intend to find some time to look into this, but do not expect to have time for that in the very near future, so if someone else is willing to investigate, it would be great.

Issue #2909, opened 2024-07-15
 Pilz LIN planner deviates from linear speed when rotation is involved in a move


gaspatxo:
### Description

My goal is to control the linear velocity at which my end effector moves throughout a move sequence. I am using the Pilz `LIN` planner and its multi-segment capability.

The speed is respected when movements do not include rotation, being the `max_trans_vel` (in `pilz_cartesianl_limits.yaml`) multiplied by `maxVelocityScalingFactor`.

However, when rotation is involved in a move, it deviates from the specified:

![pilz_lin_planner_velocity_test-0-lowres](https://github.com/user-attachments/assets/7fb46d46-1ba2-45f2-92ef-5344fcb6802f)

`pilz_cartesianl_limits.yaml`:
```yaml
# Cartesian limits for the Pilz planner
max_acc: &max_acc 3.0

cartesian_limits:
  max_trans_vel: 0.4
  max_rot_vel: 3.0
  max_trans_acc: *max_acc
  max_trans_dec: *max_acc
```

### Your environment

ROS2 Humble on Ubuntu 22.04 using moveit2 binary humble install with cyclone DDS

### Steps to reproduce

1. Install moveit tutorials
2. Activate the `/move_group_sequence` capability for the panda arm [#1281](https://github.com/moveit/moveit2/issues/1248#issuecomment-1350773802)
3. Run my test file

```cpp
#include <memory>
#include <string>
using std::string;
#include <fstream>
#include <iostream>
#include <math.h>
#include <sstream>
#include <unistd.h>
#include <vector>

#include "moveit/move_group_interface/move_group_interface.h"
#include "moveit/planning_scene_interface/planning_scene_interface.h"
#include "moveit/robot_model/robot_model.h"
#include "moveit/robot_state/conversions.h"
#include "moveit/robot_state/robot_state.h"
#include "moveit/robot_trajectory/robot_trajectory.h"
#include "moveit/trajectory_processing/iterative_time_parameterization.h"

using moveit::planning_interface::MoveGroupInterface;
#include "moveit_msgs/action/move_group_sequence.hpp"
#include "moveit_msgs/msg/motion_sequence_request.hpp"
using moveit_msgs::action::MoveGroupSequence;

#include "rclcpp/rclcpp.hpp"
using namespace rclcpp;
#include "rclcpp_action/rclcpp_action.hpp"
#include "tf2/LinearMath/Quaternion.h"
using tf2::Quaternion;

[...]

PR #3158, opened 2024-12-11
Humble

vivekcdavid:
The solution for solving the deviation as described in  #2909.

 The issue is that the Pilz LIN planner deviates from target linear speed when rotation is involved in a move. The issue seems to arise from two factors within Pilz, both related to KDL::Path_line. 

1) While generating the path using KDL::Path_line a parameter called eqradius is used. The use of Eqradius prevents very fast angular motions but as a side effect the motion deviated from the specified linear speed. *
*solution** : set eqradius to zero if the motion should only consider the specified linear TCP speed

2) Even after setting eqradius to zero we observed a small deviation in the time take for motion, this was pin-pointed to the use of trapezoidal velocity profile. 
**solution** : By using a rectangular velocity profile this issue seems to be solved.

It would be nice if these changes can be used if the use-case requires that the time for motion has to be accurate. Now I have permanently changed the code but maybe we could force Pilz to use rectangular profile when acceleration limit is not specified as suggested by @sea-bass ?

### DIFF
--- a/moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/trajectory_generator.h
+++ b/moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/trajectory_generator.h
@@ -136,8 +136,16 @@ protected:
    * The trap profile returns uses the longer distance of translational and
    * rotational motion.
    */
-  std::unique_ptr<KDL::VelocityProfile> cartesianTrapVelocityProfile(const double& max_velocity_scaling_factor,
+    std::unique_ptr<KDL::VelocityProfile> cartesianTrapVelocityProfile(const double& max_velocity_scaling_factor,
                                                                      const double& max_acceleration_scaling_factor,
+                                                                     const std::unique_ptr<KDL::Path>& path) const;
+    /**
+   * @brief build cartesian velocity profile for the path
+   *
+   * Uses the path to get the cartesian length from start to goal.
+   * The rect profile returns a uniform velocity profile without considering the acceleration
+   */    
+   std::unique_ptr<KDL::VelocityProfile> cartesianRectVelocityProfile(const double& max_velocity_scaling_factor,
                                                                      const std::unique_ptr<KDL::Path>& path) const;
 
 private:
--- a/moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/trajectory_generator.h
+++ b/moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/trajectory_generator.h
@@ -279,7 +279,7 @@ private:
 protected:
   const moveit::core::RobotModelConstPtr robot_model_;
   const pilz_industrial_motion_planner::LimitsContainer planner_limits_;
-  static constexpr double MIN_SCALING_FACTOR{ 0.0001 };
+  static constexpr double MIN_SCALING_FACTOR{ 0.0000001 };
   static constexpr double MAX_SCALING_FACTOR{ 1. };
   static constexpr double VELOCITY_TOLERANCE{ 1e-8 };
   const std::unique_ptr<rclcpp::Clock> clock_;
--- a/moveit_planners/pilz_industrial_motion_planner/src/trajectory_generator.cpp
+++ b/moveit_planners/pilz_industrial_motion_planner/src/trajectory_generator.cpp
@@ -41,6 +41,7 @@
 #include <boost/range/combine.hpp>
 
 #include <kdl/velocityprofile_trap.hpp>
+#include <kdl/velocityprofile_rect.hpp>
 #include <moveit/robot_state/conversions.h>
 
 #include <pilz_industrial_motion_planner/limits_container.h>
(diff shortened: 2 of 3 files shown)


sea-bass:
> Please target the `main` branch for development, we will backport the changes to humble for you if approved and if they don't break API.

Thanks @vivekcdavid! As this comment says, all new changes should go through to the `main` branch and backported to `humble`. Consider rebasing your changes on `main` and retargeting that branch here.

I'll still leave a few comments for now in case you want to go at those too.

moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/trajectory_generator.h
```diff
@@ -136,9 +136,17 @@ class TrajectoryGenerator
    * The trap profile returns uses the longer distance of translational and
    * rotational motion.
    */
-  std::unique_ptr<KDL::VelocityProfile> cartesianTrapVelocityProfile(const double& max_velocity_scaling_factor,
+    std::unique_ptr<KDL::VelocityProfile> cartesianTrapVelocityProfile(const double& max_velocity_scaling_factor,
```
sea-bass, line 139: this indentation may break the formatting -- accidental change?

moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/trajectory_generator.h
```diff
@@ -271,7 +279,7 @@ class TrajectoryGenerator
 protected:
   const moveit::core::RobotModelConstPtr robot_model_;
   const pilz_industrial_motion_planner::LimitsContainer planner_limits_;
-  static constexpr double MIN_SCALING_FACTOR{ 0.0001 };
+  static constexpr double MIN_SCALING_FACTOR{ 0.0000001 };
```
sea-bass, line 282: This may end up changing default behavior -- but as a question to you, did you find this made a difference in other cases?

Also minor nitpick, with all the zeros consider using scientific notation like `1e-7`.

moveit_planners/pilz_industrial_motion_planner/src/trajectory_generator.cpp
```diff
@@ -300,6 +301,37 @@ TrajectoryGenerator::cartesianTrapVelocityProfile(const double& max_velocity_sca
...
+  //std::unique_ptr<KDL::VelocityProfile_Rectangular> vp_trans(new KDL::VelocityProfile_Rectangular());
+  std::unique_ptr<KDL::VelocityProfile> vp_trans = std::make_unique<KDL::VelocityProfile_Rectangular>(
+      max_velocity_scaling_factor * planner_limits_.getCartesianLimits().getMaxTranslationalVelocity());
+  //std::unique_ptr<KDL::VelocityProfile> vp_trans = std::make_unique<KDL::VelocityProfile_Trap>(
+  //    max_velocity_scaling_factor * planner_limits_.getCartesianLimits().getMaxTranslationalVelocity(),
+  //    max_acceleration_scaling_factor * planner_limits_.getCartesianLimits().getMaxTranslationalAcceleration());
+
+  if (path->PathLength() > std::numeric_limits<double>::epsilon())  // avoid division by zero
+  {
+    //vp_trans->SetProfile(0, path->PathLength());
```
sea-bass, line 318: make sure all these commented out lines are removed

moveit_planners/pilz_industrial_motion_planner/src/trajectory_generator_lin.cpp
```diff
@@ -154,10 +154,12 @@ void TrajectoryGeneratorLIN::plan(const planning_scene::PlanningSceneConstPtr& s
   // create Cartesian path for lin
   std::unique_ptr<KDL::Path> path(setPathLIN(plan_info.start_pose, plan_info.goal_pose));
 
-  // create velocity profile
+  // create velocity profile using trap profile
```
sea-bass, line 157: should be rect?

moveit_planners/pilz_industrial_motion_planner/src/trajectory_generator_lin.cpp
```diff
@@ -154,10 +154,12 @@ void TrajectoryGeneratorLIN::plan(const planning_scene::PlanningSceneConstPtr& s
...
+  //std::unique_ptr<KDL::VelocityProfile> vp(
+  //    cartesianTrapVelocityProfile(req.max_velocity_scaling_factor, req.max_acceleration_scaling_factor, path));
+  // create velocity profile using rect profile
```
sea-bass, line 160: remove these?

moveit_planners/pilz_industrial_motion_planner/src/trajectory_generator_lin.cpp
```diff
@@ -154,10 +154,12 @@ void TrajectoryGeneratorLIN::plan(const planning_scene::PlanningSceneConstPtr& s
...
   std::unique_ptr<KDL::VelocityProfile> vp(
-      cartesianTrapVelocityProfile(req.max_velocity_scaling_factor, req.max_acceleration_scaling_factor, path));
-
+        cartesianRectVelocityProfile(req.max_velocity_scaling_factor, path));
```
sea-bass, line 162: This cannot be just changed, it has to be a runtime parameter configurable by the user, whose default still retains trapezoidal for compatibility.

moveit_planners/pilz_industrial_motion_planner/src/trajectory_generator_lin.cpp
```diff
@@ -185,8 +187,9 @@ std::unique_ptr<KDL::Path> TrajectoryGeneratorLIN::setPathLIN(const Eigen::Affin
   KDL::Frame kdl_start_pose, kdl_goal_pose;
   tf2::transformEigenToKDL(start_pose, kdl_start_pose);
   tf2::transformEigenToKDL(goal_pose, kdl_goal_pose);
-  double eqradius = planner_limits_.getCartesianLimits().getMaxTranslationalVelocity() /
-                    planner_limits_.getCartesianLimits().getMaxRotationalVelocity();
+  //double eqradius = planner_limits_.getCartesianLimits().getMaxTranslationalVelocity() /
+  //                  planner_limits_.getCartesianLimits().getMaxRotationalVelocity();
+  double eqradius = 0.0;
```
sea-bass, line 192: same here -- probably whatever parameter is used to switch from trapezoidal to rectangle can also zero out the `eq_radius`.

moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/trajectory_generator.h
vivekcdavid: I didn't see any difference in behavior so will put it back to the default value for the time being

gaspatxo:
hey @sea-bass,

Me and @vivekcdavid have been thinking what the best way forwards is.

We first considered adding the interpretetion of acceleration=0 on a trajectory profile
implies the usage of the rectiliniar profile (as if acceleration=∞). But adding a zero case can easily introduce math bugs.

Our next and current idea is adding a new flag variable in the `MotionPlanRequest` after filling it out with `constructMotionPlanRequest`. Then we would interpret that in the [`trajectory_planner_lin.cpp`](https://github.com/vivekcdavid/moveit2_raven2/blob/main/moveit_planners/pilz_industrial_motion_planner/src/trajectory_generator_lin.cpp) to use a rectangular profile.

Please give us your thoughts in the idea, what are we missing, what would be better for us to implement.

sea-bass:
@gaspatxo I might suggest instead of modifying the request, to make a boolean parameter to either use trapezoidal or rectangular velocity profiles. Since this is specific to one planner type, it would bloat the planning request definition to add this there.

If you (or users in general) need to mix and match both profiles, you can launch 2 separate Pilz planner instances with the different parameters.

What do you think?

gaspatxo:
Where would we add this parameter? we are currently making use of pilz planner over the `move_group_interface`, modifying that function would modify it for all users which Im guessing it is not a good idea.

Are you suggesting we stop using the `move_group_interface` and directly call the pilz_planner?

sea-bass:
> Where would we add this parameter?

For example, the cartesian limits are currently defined as parameters here: https://github.com/moveit/moveit2/blob/b47e302d5652c0f9ee06bffad749c1793690bb26/moveit_planners/pilz_industrial_motion_planner/src/cartesian_limits_parameters.yaml#L1

And are read in here: https://github.com/moveit/moveit2/blob/b47e302d5652c0f9ee06bffad749c1793690bb26/moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/pilz_industrial_motion_planner.hpp#L138

It is kind of annoying that these parameters only refer to the Cartesian limits, since it feels out of place to add what we've been talking about there.

But maybe this could come with a slight rename of these parameters, or you could add a completely new parameter definition file?

> Are you suggesting we stop using the `move_group_interface` and directly call the pilz_planner?

Not at all! These would be configuration time parameters that would be set at launch time. So basically they are server-side parameters, but when you use move group interface you're still calling out to that server to plan.

Issue #3161, opened 2024-12-13
[pilz_industrial_motion_planner] velocity drop to near 0 during a  plan_sequence_path service generated motion


wuisky:
### Description

Hi, I have a issue when executing a trajectory generated by pilz_industrial_motion_planner plan_sequence_path service.
Here is the motion generated by plan_sequence_path. There are two MotionSequenceItems in my MotionSequenceRequest(go to right 0.3[m], then go to down 0.3[m]. blend radius is 0.2[m]).
![Image](https://github.com/user-attachments/assets/1b8af11f-3e67-42e6-abfd-ab4001c4ac39)

And i noticed that velocity dropped to near zero two times during the motion(at the time transition start and end?)
![Image](https://github.com/user-attachments/assets/d6019389-d782-4756-850b-454e52599967)
that causes an unsmooth motion.

### ROS Distro

Humble

### OS and version

ubuntu 22.04

### Source or binary build?

Source

### If source, which branch?

humble

### Which RMW are you using?

FastRTPS

## pilz config file
```yaml
planning_plugin: pilz_industrial_motion_planner/CommandPlanner
default_planner_config: PTP
request_adapters: >-
  default_planner_request_adapters/FixWorkspaceBounds
  default_planner_request_adapters/FixStartStateBounds
  default_planner_request_adapters/FixStartStateCollision
  default_planner_request_adapters/FixStartStatePathConstraints
response_adapters:
  - default_planning_response_adapters/ValidateSolution
  - default_planning_response_adapters/DisplayMotionPath
capabilities: >-
    pilz_industrial_motion_planner/MoveGroupSequenceAction
    pilz_industrial_motion_planner/MoveGroupSequenceService
```
## pilz cartesian_limits:
```yaml
cartesian_limits:
  max_trans_vel: 1.0
  max_trans_acc: 2.25
  max_trans_dec: -5.0
  max_rot_vel: 1.57
```

## c++ code snippet
```c++
moveit_msgs::msg::MotionSequenceItem createMotionSequenceItem(
  const std::string group_name,
  const std::string link_name,
  const geometry_msgs::msg::Pose target_pose,
  const std::string planner_id,
  const double blend_radius)
{
  moveit_msgs::msg::MotionSequenceItem item;
  moveit_msgs::msg::Constraints constraints;
  moveit_msgs::msg::PositionConstraint position_constraint;
  position_constraint.header.frame_id = "world";
  position_constraint.link_name = link_name;
[...]

PR #3169, opened 2024-12-17
rm unnessary time offset in pilz

wuisky:
### Description

A workaround of #3161 

### Checklist

### DIFF
--- a/moveit_planners/pilz_industrial_motion_planner/src/plan_components_builder.cpp
+++ b/moveit_planners/pilz_industrial_motion_planner/src/plan_components_builder.cpp
@@ -54,6 +54,11 @@ std::vector<robot_trajectory::RobotTrajectoryPtr> PlanComponentsBuilder::build()
 void PlanComponentsBuilder::appendWithStrictTimeIncrease(robot_trajectory::RobotTrajectory& result,
                                                          const robot_trajectory::RobotTrajectory& source)
 {
+  if (!source.getWayPointCount())
+  {
+    return;
+  }
+
   if (result.empty())
   {
     result.append(source, 0.0);
@@ -62,7 +67,14 @@ void PlanComponentsBuilder::appendWithStrictTimeIncrease(robot_trajectory::Robot
   if (!pilz_industrial_motion_planner::isRobotStateEqual(result.getLastWayPoint(), source.getFirstWayPoint(),
                                                          result.getGroupName(), ROBOT_STATE_EQUALITY_EPSILON))
   {
-    result.append(source, source.getWayPointDurationFromStart(0));
+    if (source.getWayPointDurationFromPrevious(0) == 0)
+    {
+      result.append(source, source.getWayPointDurationFromStart(0));
+    }
+    else
+    {
+      result.append(source, 0.0);
+    }
     return;
   }
 


moveit_planners/pilz_industrial_motion_planner/src/plan_components_builder.cpp
```diff
@@ -62,7 +67,14 @@ void PlanComponentsBuilder::appendWithStrictTimeIncrease(robot_trajectory::Robot
   if (!pilz_industrial_motion_planner::isRobotStateEqual(result.getLastWayPoint(), source.getFirstWayPoint(),
                                                          result.getGroupName(), ROBOT_STATE_EQUALITY_EPSILON))
   {
-    result.append(source, source.getWayPointDurationFromStart(0));
+    if (source.getWayPointDurationFromPrevious(0) == 0)
+    {
+      result.append(source, source.getWayPointDurationFromStart(0));
+    }
+    else
```
sea-bass, line 74: Can this be instead added to the condition in line 62?

```cpp
if (result.empty() || source.getWayPointDurationFromPrevious(0) > 0.0) {
    result.append(source, 0.0);
    return;
}
```

It may make this logic a bit easier to follow

sea-bass: Alternatively, do things still work out if instead of returning early in the IF case here, you simply let the function return at the very end?

sea-bass: Hmm, you're right. Maybe we shouldn't do that.

Let me test with this and get back to you.

sea-bass: OK I tested on [this tutorial example](https://moveit.picknik.ai/main/doc/how_to_guides/pilz_industrial_motion_planner/pilz_industrial_motion_planner.html#sequence-of-multiple-segments) and unfortunately got this error:

```
[ros2_control_node-5] [INFO] [1734435722.902666137] [panda_arm_controller]: Received new action goal
[ros2_control_node-5] [ERROR] [1734435722.902766306] [panda_arm_controller]: Time between points 27 and 28 is not strictly increasing, it is 2.700000 and 2.700000 respectively
[move_group-4] [INFO] [1734435722.903129429] [moveit.simple_controller_manager.follow_joint_trajectory_controller_handle]: panda_arm_controller started execution
[move_group-4] [WARN] [1734435722.903173172] [moveit.simple_controller_manager.follow_joint_trajectory_controller_handle]: Goal request rejected
```

In this example, both duration from start and previous are `0.1`, so the command `result.append(source, 0.0);` will not work out.

What are the duration from start and previous in your case? Wonder if there is some other special case happening on your end?

sea-bass: The thing is, that on Humble, `ros2_control`'s joint trajectory controller actually allowed duplicate waypoints to be at the same time and it would deconflict them. In later versions, errors like the above show up.

So appending waypoints whose state is not equal, with zero time offset, is not something we should be doing.

sea-bass: I actually don't know what changed between Humble and later versions that having these duplicate time points ends up working... since these lines of code have been there since Humble:

https://github.com/ros-controls/ros2_controllers/blob/0438a1e886386035295da90c3e3dc0436217ddcc/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L1444-L1452

sea-bass: Oh INTERESTING! It seems that the Humble branch does not have this commit: https://github.com/moveit/moveit2/pull/1813

which may be why it works there, but not on later versions.

wuisky: Yeah, after I include #1821 my results look good!
I think I'll  just close my PR and waiting for #1821 to be merged to humble.
Thanks for your help.

sea-bass: Thanks for confirming! Will do this now.

PR #3183, opened 2024-12-30
Remove ACM entries when removing collision objects

JafarAbdi:
### Description

When removing collision objects from a planning scene, we need to also remove their corresponding entries from the allowed collision matrix. Previously, these entries would remain in the ACM even after the object was removed, which could cause inconsistencies if an object with the same name was added later.

### Checklist

### DIFF
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -1466,7 +1466,6 @@ void PlanningScene::removeAllCollisionObjects()
       world_->removeObject(object_id);
       removeObjectColor(object_id);
       removeObjectType(object_id);
-      // TODO: How about octomap?
       getAllowedCollisionMatrixNonConst().removeEntry(object_id);
     }
   }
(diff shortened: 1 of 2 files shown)


sea-bass (approved):
Nice -- thanks @JafarAbdi!

Marking these for Humble/Jazzy backport as well.

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -1465,6 +1466,8 @@ void PlanningScene::removeAllCollisionObjects()
       world_->removeObject(object_id);
       removeObjectColor(object_id);
       removeObjectType(object_id);
+      // TODO: How about octomap?
```
sea-bass, line 1469: worth a separate issue here?

moveit_core/planning_scene/test/test_planning_scene.cpp
```diff
@@ -590,6 +590,61 @@ TEST(PlanningScene, RobotStateDiffBug)
   }
 }
 
+TEST(PlanningScene, ACMBug)
```
sea-bass, line 593: optional: rather than calling the test `ACMBug`, how about something more descriptive like `UpdateACMAfterObjectRemoval` or so?

JafarAbdi: I updated the name. Thanks!

sea-bass (approved):

moveit_core/planning_scene/src/planning_scene.cpp
JafarAbdi: After thinking about it, I think we shouldn't handle it since there is a reserved name for it and user can't add an object with the same name

Issue #3189, opened 2025-01-03
Conditionally install launch directory in generated MoveIt config packages


filippo-bosi:
### Description

When generating a MoveIt configuration package using the MoveIt Setup Assistant, the **CMakeLists.txt** file always attempts to install the launch directory, even if it does not exist. This can cause the installation process to fail when no launch files are created in the setup assistant.

### Proposed Solution
Add a condition to check if the launch directory exists before attempting to install it. This ensures that the installation process does not fail when the launch directory is not present.

- CMakeLists.txt:
```
// filepath: /ros2_humble/src/moveit2/moveit_setup_assistant/moveit_setup_framework/templates/CMakeLists.txt
// ...existing code...
if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/launch")
  install(DIRECTORY launch DESTINATION share/${PROJECT_NAME}
    PATTERN "setup_assistant.launch" EXCLUDE)
endif()
// ...existing code...
```

### ROS Distro

Humble

### OS and version

Ubuntu 22.04

### Source or binary build?

Source

### If source, which branch?

humble

### Which RMW are you using?

None

### Steps to Reproduce

1. Generate a new MoveIt configuration package using the Setup Assistant.
2. Do not create any launch files in the wizard.
3. Attempt to build and install the package using `colcon build`

### Expected behavior

The installation process should complete successfully, even if the launch directory is not present.

### Actual behavior

The installation process fails because the launch directory does not exist.

PR #3191, opened 2025-01-03
Fix: Conditionally install launch directory in generated MoveIt config package

filippo-bosi:
### Overview

This pull request addresses issue #3189 where the `CMakeLists.txt` file always attempts to install the `launch` directory, even if it does not exist. This can cause the installation process to fail when no launch files are created in the setup assistant.

### Changes Made

1. **CMakeLists.txt**:
   - Added a condition to check if the `launch` directory exists before attempting to install it. This ensures that the installation process does not fail when the `launch` directory is not present.

```cmake
  // filepath: /ros2_humble/src/moveit2/moveit_setup_assistant/moveit_setup_framework/templates/CMakeLists.txt
  // ...existing code...
  if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/launch")
    install(
      DIRECTORY launch
      DESTINATION share/${PROJECT_NAME}
      PATTERN "setup_assistant.launch" EXCLUDE)
  endif()
  // ...existing code...

### DIFF
--- a/moveit_setup_assistant/moveit_setup_framework/templates/CMakeLists.txt
+++ b/moveit_setup_assistant/moveit_setup_framework/templates/CMakeLists.txt
@@ -5,9 +5,12 @@ find_package(ament_cmake REQUIRED)
 
 ament_package()
 
-install(
-  DIRECTORY launch
-  DESTINATION share/${PROJECT_NAME}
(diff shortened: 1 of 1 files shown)


sea-bass (approved):
Looks good!

PR #3198, opened 2025-01-04
[moveit_ros] fix race condition when stopping trajectory execution

patrickKXMD:
### Description

It is not thread-safe calling `stopExecution` or publishing a `"stop"` to `trajectory_execution_event` while trajectory executing. Sometimes it crashes, here's a sample

> Program terminated with signal SIGSEGV, Segmentation fault.
> 
> #0  0x0000ffffa367ac48 in __gnu_cxx::__normal_iterator<moveit_msgs::msg::RobotTrajectory_<std::allocator<void> > const*, std::vector<moveit_msgs::msg::RobotTrajectory_<std::allocator<void> >, std::allocator<moveit_msgs::msg::RobotTrajectory_<std::allocator<void> > > > >::__normal_iterator (__i=<optimized out>, this=<optimized out>)
>     at /usr/include/c++/11/bits/stl_iterator.h:1028
> b1028	      : _M_current(__i) { }
> [Current thread is 1 (Thread 0xffff1e7cf120 (LWP 4402))]
> (gdb) 
> (gdb) 
> (gdb) bt
> #0  0x0000ffffa367ac48 in __gnu_cxx::__normal_iterator<moveit_msgs::msg::RobotTrajectory_<std::allocator<void> > const*, std::vector<moveit_msgs::msg::RobotTrajectory_<std::allocator<void> >, std::allocator<moveit_msgs::msg::RobotTrajectory_<std::allocator<void> > > > >::__normal_iterator (__i=<optimized out>, this=<optimized out>)
>     at /usr/include/c++/11/bits/stl_iterator.h:1028
> #1  std::vector<moveit_msgs::msg::RobotTrajectory_<std::allocator<void> >, std::allocator<moveit_msgs::msg::RobotTrajectory_<std::allocator<void> > > >::end (
>     this=<optimized out>) at /usr/include/c++/11/bits/stl_vector.h:839
> #2  trajectory_execution_manager::TrajectoryExecutionManager::waitForRobotToStop (this=this@entry=0xaaaaeed29f20, context=..., wait_time=wait_time@entry=1)
>     at /workspace/source/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp:1592
> #3  0x0000ffffa36829d4 in trajectory_execution_manager::TrajectoryExecutionManager::executeThread(std::function<void (moveit_controller_manager::ExecutionStatus const&)> const&, std::function<void (unsigned long)> const&, bool) (this=0xaaaaeed29f20, callback=..., part_callback=..., auto_clear=<optimized out>) at /usr/include/c++/11/bits/stl_vector.h:1043
> #4  0x0000ffffba7431fc in ?? () from /lib/aarch64-linux-gnu/libstdc++.so.6
> #5  0x0000ffffbb92d5c8 in start_thread (arg=0x0) at ./nptl/pthread_create.c:442
> #6  0x0000ffffbb995edc in thread_start () at ../sysdeps/unix/sysv/linux/aarch64/clone.S:79

This CL introduces a lock for interactions with `trajectories_` and can fix this issue.

### Checklist


moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -1271,9 +1271,11 @@ void TrajectoryExecutionManager::clear()
 {
   if (execution_complete_)
   {
+    execution_trajectories_mutex_.lock();
```
sea-bass, line 1274: You could consider using a scoped lock instead.

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -1312,7 +1314,23 @@ void TrajectoryExecutionManager::executeThread(const ExecutionCompleteCallback&
 
   // only report that execution finished successfully when the robot actually stopped moving
   if (last_execution_status_ == moveit_controller_manager::ExecutionStatus::SUCCEEDED)
-    waitForRobotToStop(*trajectories_[i - 1]);
+  {
+    execution_trajectories_mutex_.lock();
```
sea-bass, line 1318: Same here re: scoped lock

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -1312,7 +1314,23 @@ void TrajectoryExecutionManager::executeThread(const ExecutionCompleteCallback&
...
+    if (!execution_complete_ && trajectories_.size() > 0)
```
sea-bass, line 1319: Should this check that the trajectory size is greater than or equal to `i` instead?

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -1589,9 +1608,8 @@ bool TrajectoryExecutionManager::waitForRobotToStop(const TrajectoryExecutionCon
 
     // check for motion in effected joints of execution context
     bool moved = false;
-    for (const auto& trajectory : context.trajectory_parts_)
+    (const auto& joint_names : joint_names_per_traj)
```
sea-bass, line 1611: If you already have access to `context.trajectory_parts_` here, maybe consider locking and computing the joint names per traj in this function?

This means you wouldn't have to change the function interface in this case.

sea-bass (commented):
Thanks!

In addition to looking at the CI failures since something isn't building, left a few suggestions.

Rolling is failing now until https://github.com/moveit/moveit2/pull/3197, but the rest should pass.

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
sjahr: :point_up:

sjahr (commented):
Thanks, once the remaining comment is addressed and the PR that updates the deprecated headers is merged I think this is good to go

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -1312,7 +1313,23 @@ void TrajectoryExecutionManager::executeThread(const ExecutionCompleteCallback&
 
   // only report that execution finished successfully when the robot actually stopped moving
   if (last_execution_status_ == moveit_controller_manager::ExecutionStatus::SUCCEEDED)
-    waitForRobotToStop(*trajectories_[i - 1]);
+  {
+    execution_trajectories_mutex_.lock();
```
rhaschke, line 1317: Use existing `execution_state_mutex_` instead:
```suggestion
    std::scoped_lock slock(execution_state_mutex_);
```

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -1312,7 +1313,23 @@ void TrajectoryExecutionManager::executeThread(const ExecutionCompleteCallback&
...
+    if (!execution_complete_ && trajectories_.size() >= i)
```
rhaschke, line 1318: Can the condition `trajectories_.size() >= i` become false at all? That's only possible if clear() is called in between.
As we hold the `execution_state_mutex_` this cannot happen anymore.
```suggestion
    if (!execution_complete_)
```

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -1312,7 +1313,23 @@ void TrajectoryExecutionManager::executeThread(const ExecutionCompleteCallback&
...
+      std::vector<std::vector<std::string>> joint_names_per_traj;
+      for (const auto& trajectory : context.trajectory_parts_)
+      {
+        joint_names_per_traj.push_back(trajectory.joint_trajectory.joint_names);
+      }
+      execution_trajectories_mutex_.unlock();
+      waitForRobotToStop(joint_names_per_traj);
+    }
+    else
+      execution_trajectories_mutex_.unlock();
```
rhaschke, line 1330: Obsolete when using scoped_lock.
```suggestion
    else
      execution_trajectories_mutex_.unlock();
```

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -1271,6 +1271,7 @@ void TrajectoryExecutionManager::clear()
 {
   if (execution_complete_)
   {
+    std::scoped_lock slock(execution_trajectories_mutex_);
```
rhaschke, line 1274:
```suggestion
    std::scoped_lock slock(execution_trajectories_mutex_);
```

rhaschke (commented):
We might avoid introducing a new mutex and reuse `execution_state_mutex_`.
In all locations calling `clear()`, the flag `execution_complete_` is True:
- [`TrajectoryExecutionManager::stopExecution`](https://github.com/moveit/moveit2/blob/c3a2edc88702416ad0661d06b3035a79cdf4c8b2/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp#L1190)
- [`TrajectoryExecutionManager::execute`](https://github.com/moveit/moveit2/blob/c3a2edc88702416ad0661d06b3035a79cdf4c8b2/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp#L1251): `execution_complete_` is _still_ true when calling `clear()`
- [`TrajectoryExecutionManager::executeThread`](https://github.com/moveit/moveit2/blob/c3a2edc88702416ad0661d06b3035a79cdf4c8b2/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp#L1321)

However, currently `clear()` is a public method. Thus, calling it directly will face the race condition.
I suggest changing its visibility to protected or even private.

sea-bass (approved):
This looks really clean now -- as long as this still fixes your tests, looks great to me. Thanks!

rhaschke (changes requested):
If you follow my suggestion https://github.com/moveit/moveit2/pull/3198#pullrequestreview-2531845347, you should also change visibility of `clear()` to `protected` or `private`!

Also, as a safe-guard for future changes, you should add an assertion to `clear()`, e.g. performing the following change to the line:
https://github.com/moveit/moveit2/blob/80c7f612bcaa45264ff8a5483f99325c7e5689f2/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp#L1279
```diff
-    RCLCPP_ERROR(logger_, "Cannot push a new trajectory while another is being executed");
+    RCLCPP_FATAL(logger_, "Expecting execution_complete_ to be true!");
```

patrickKXMD:
> If you follow my suggestion [#3198 (review)](https://github.com/moveit/moveit2/pull/3198#pullrequestreview-2531845347), you should also change visibility of `clear()` to `protected` or `private`!
> 
> Also, as a safe-guard for future changes, you should add an assertion to `clear()`, e.g. performing the following change to the line:
> 
> https://github.com/moveit/moveit2/blob/80c7f612bcaa45264ff8a5483f99325c7e5689f2/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp#L1279
> 
> ```diff
> -    RCLCPP_ERROR(logger_, "Cannot push a new trajectory while another is being executed");
> +    RCLCPP_FATAL(logger_, "Expecting execution_complete_ to be true!");
> ```

Hi @rhaschke,
Thank you for your suggestion, I added an assertion in `clear`.

But I have concern for changing visibility of `clear` now. It has been public for a long time, changing it might break someone's code.
In fact moveit2 uses it's public visibility too. 
e.g.
https://github.com/moveit/moveit2/blob/75286c7822e5f2d9babbb41b9a0ffaddb0d54808/moveit_ros/move_group/src/default_capabilities/execute_trajectory_action_capability.cpp#L109

https://github.com/moveit/moveit2/blob/75286c7822e5f2d9babbb41b9a0ffaddb0d54808/moveit_ros/planning/plan_execution/src/plan_execution.cpp#L405

https://github.com/moveit/moveit2/blob/75286c7822e5f2d9babbb41b9a0ffaddb0d54808/moveit_py/src/moveit/moveit_ros/trajectory_execution_manager/trajectory_execution_manager.cpp#L203-L206

rhaschke:
All these calls to `clear()` should be replaced by `stopExecution(true)`.

patrickKXMD:
> All these calls to `clear()` should be replaced by `stopExecution(true)`.

Hi I opened a new pr #3226 for this

rhaschke (approved):
I suggest to merge #3226 as part of this PR.

PR #3205, opened 2025-01-07
Don't destroy objects on attach

MarqRazz:
### Description

When an object is added to the scene and immediately attached to the gripper it gets marked as [collision_detection::World::DESTROY](https://github.com/moveit/moveit2/blob/main/moveit_core/planning_scene/src/planning_scene.cpp#L372) in the `world_diff_`. When this happens the [ACM is modified](https://github.com/moveit/moveit2/blob/main/moveit_core/planning_scene/src/planning_scene.cpp#L377) and the attached object is not allowed to touch anything. 99% of the time this does not cause issues when executing a trajectory because the [scene diff](https://github.com/moveit/moveit2/blob/main/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp#L502) is not applied until after the trajectory starts, but is still updating the ACM incorrectly. Sometimes the scene diff is applied before starting the trajectory and it fails after the arm starts moving because of [this check here](https://github.com/moveit/moveit2/blob/main/moveit_ros/planning/plan_execution/src/plan_execution.cpp#L443). This PR adds some logic to see if the object is being `ADD`ed as an attached collision object so that it is not `DESTROY`ed causing the ACM to incorrectly update.

### DIFF
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -371,10 +371,16 @@ void PlanningScene::pushDiffs(const PlanningScenePtr& scene)
     {
       if (it.second == collision_detection::World::DESTROY)
       {
-        scene->world_->removeObject(it.first);
-        scene->removeObjectColor(it.first);
-        scene->removeObjectType(it.first);
-        scene->getAllowedCollisionMatrixNonConst().removeEntry(it.first);
+        moveit_msgs::msg::AttachedCollisionObject aco;
+        scene->getAttachedCollisionObjectMsg(aco, it.first);
+        // if object became attached, it should not be recorded as DESTROY here
+        if (aco.object.operation == moveit_msgs::msg::CollisionObject::REMOVE)
+        {
+          scene->world_->removeObject(it.first);
+          scene->removeObjectColor(it.first);
+          scene->removeObjectType(it.first);
+          scene->getAllowedCollisionMatrixNonConst().removeEntry(it.first);
+        }
       }
       else
       {
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -371,14 +371,12 @@ void PlanningScene::pushDiffs(const PlanningScenePtr& scene)
     {
       if (it.second == collision_detection::World::DESTROY)
       {
-        moveit_msgs::msg::AttachedCollisionObject aco;
-        scene->getAttachedCollisionObjectMsg(aco, it.first);
-        // if object became attached, it should not be recorded as DESTROY here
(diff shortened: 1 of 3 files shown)


moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -371,10 +371,16 @@ void PlanningScene::pushDiffs(const PlanningScenePtr& scene)
...
+        moveit_msgs::msg::AttachedCollisionObject aco;
+        scene->getAttachedCollisionObjectMsg(aco, it.first);
+        // if object became attached, it should not be recorded as DESTROY here
+        if (aco.object.operation == moveit_msgs::msg::CollisionObject::REMOVE)
+        {
+          scene->world_->removeObject(it.first);
+          scene->removeObjectColor(it.first);
+          scene->removeObjectType(it.first);
+          scene->getAllowedCollisionMatrixNonConst().removeEntry(it.first);
+        }
```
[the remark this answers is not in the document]
MarqRazz: Thanks for the suggestion @rhaschke!

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -374,7 +374,11 @@ void PlanningScene::pushDiffs(const PlanningScenePtr& scene)
         scene->world_->removeObject(it.first);
         scene->removeObjectColor(it.first);
         scene->removeObjectType(it.first);
-        scene->getAllowedCollisionMatrixNonConst().removeEntry(it.first);
+        // if object is attached, it should not should not be removed from the ACM
```
rhaschke, line 377:
```suggestion
        // if object is attached, it should not be removed from the ACM
```

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -374,7 +374,11 @@ void PlanningScene::pushDiffs(const PlanningScenePtr& scene)
...
+        if (!getCurrentState().hasAttachedBody(it.first))
```
rhaschke, line 378: Sorry, this was wrong in my initial suggestion:
```suggestion
        if (!scene->getCurrentState().hasAttachedBody(it.first))
```

JafarAbdi (commented):
Thanks for the fix! Could you please extend the test case in test_planning_scene.cpp? Otherwise lgtm

MarqRazz:
> Thanks for the fix! Could you please extend the test case in test_planning_scene.cpp? Otherwise lgtm

Thanks for the suggestion, let me see what I can do.

MarqRazz:
@JafarAbdi how does the test extension look?

JafarAbdi (approved):
Thanks!!

sjahr (approved):
Thanks!

PR #3206, opened 2025-01-07
Simplify scene update that does not include new robot_state

MarqRazz:
### Description

Updates changes applied in #3187 to match the [back port to MoveIt1](https://github.com/moveit/moveit/pull/3683).

### DIFF
--- a/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
+++ b/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
@@ -740,21 +740,9 @@ bool PlanningSceneMonitor::newPlanningSceneMessage(const moveit_msgs::msg::Plann
 
     if (!scene.is_diff && parent_scene_)
     {
-      // if the scene does not contain a robot state then use the latest known values for the robot state
-      // so that we are not updating the scene with stale values.
-      if (scene.robot_state.joint_state.name.empty())
-      {
-        // if we have a valid current_state_monitor_ with complete state use the latest data otherwise try
-        // to use the maintained (diff) scene_ which may fall back to the parent_scene_ data and not update.
-        if (current_state_monitor_ && current_state_monitor_->haveCompleteState())
-        {
-          parent_scene_->setCurrentState(*current_state_monitor_->getCurrentState());
-        }
-        else
-        {
-          parent_scene_->setCurrentState(scene_->getCurrentState());
-        }
-      }
+      // If there is no new robot_state, transfer RobotState from current scene to parent scene
+      if (scene.robot_state.joint_state.name.empty() && scene.robot_state.multi_dof_joint_state.joint_names.empty())
+        parent_scene_->setCurrentState(scene_->getCurrentState());
       // clear maintained (diff) scene_ and set the full new scene in parent_scene_ instead
       scene_->clearDiffs();
       result = parent_scene_->setPlanningSceneMsg(scene);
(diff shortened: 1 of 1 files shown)


rhaschke (approved):
I assume you verified that this simplification still fixes your original issue.

JafarAbdi (approved):

MarqRazz:
Yes I verified that this does fix my issue. FYI I do see small differences between the diff `scene_` and `CSM`, for example
`[move_group.moveit.moveit.ros.planning_scene_monitor ]: CSM: 0.814504, diff scene_: 0.812823`, but it is so small it is not effecting my motion plans like it was before.

rhaschke:
Let's wait for https://github.com/moveit/moveit/pull/3683 before merging this.

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
```diff
@@ -740,21 +740,9 @@ bool PlanningSceneMonitor::newPlanningSceneMessage(const moveit_msgs::msg::Plann
...
-        {
-          parent_scene_->setCurrentState(*current_state_monitor_->getCurrentState());
-        }
-        else
-        {
-          parent_scene_->setCurrentState(scene_->getCurrentState());
-        }
-      }
+      // If there is no new robot_state, transfer RobotState from current scene to parent scene
+      if (scene.robot_state.joint_state.name.empty() && scene.robot_state.multi_dof_joint_state.joint_names.empty())
```
rhaschke, line 744:
```suggestion
      if (scene.robot_state.is_diff)
```

MarqRazz:
> I guess those are some randomly picked joint values? CSM updates are much more frequent than PS updates. If the robot gets to rest, they should become identical eventually.

That is showing values for joint_1 of my robot when `newPlanningSceneMessage` is called with `scene.is_diff=false` and an empty joint_state. At this point of my setup the robot motion is complete so it depends on the last time the planning_scene_monitor updated it's robot_state / published at it's 2Hz rate if will include this small change in position.

sea-bass (approved):
I think your commit invalidated the other approving reviews. Here's another!

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
```diff
@@ -740,21 +740,9 @@ bool PlanningSceneMonitor::newPlanningSceneMessage(const moveit_msgs::msg::Plann
...
-        {
-          parent_scene_->setCurrentState(*current_state_monitor_->getCurrentState());
-        }
-        else
-        {
-          parent_scene_->setCurrentState(scene_->getCurrentState());
-        }
-      }
+      // If there is no new robot_state, transfer RobotState from current scene to parent scene
+      if (scene.robot_state.is_diff)
```
MarqRazz, line 744: I considered using `robot_state.is_diff` as the flag but that means that the sender knows they are required to set it when when they create the new scene message. As you know [MTC](https://github.com/moveit/moveit_task_constructor/blob/a924b566fbe25ac8f4e3b5d70da2713e39d62637/capabilities/src/execute_task_solution_capability.cpp#L176) only does this to silence a warning.

It feels like checking joint_names.empty() guarantees that the scene is updated correctly.

rhaschke: If `robot_state.is_diff == False` the robot_state should include the full state! If it doesn't, the message is _malformed_, triggering this warning: https://github.com/moveit/moveit2/blob/80c7f612bcaa45264ff8a5483f99325c7e5689f2/moveit_core/robot_state/src/conversions.cpp#L377-L379

MTC explicitly sets `is_diff=True` to convey a valid message and _thus_ avoid the warning.
Did you observe any code that generates `is_diff == False` _and_ an empty JointState message?

MarqRazz: I have not, I'm only using MTC right now. It's good to hear it returns an error, not just a warning! Since that is the case I agree this is a good simplification :+1:

PR #3209, opened 2025-01-08
fix: OctoMap and Filtered_Cloud Not Updating During Movement Execution

MarcoMagriDev:
### Description

MutuallyExclusive callbackgroup has been added to `point_cloud_subscriber_` to allow updating octomap during robot motion. See https://github.com/moveit/moveit2/issues/2192

---
### Additional notes
-  I was not able to do the same fix to the `depth_image_octomap_updater` since `image_transport::create_camera_subscription` seems to not allow passing `SubscriptionOptions`.

- I noticed that there is a big computational time mismatch between the two approaches (depth/pointcloud). After digging in a bit i found that the main computational time difference comes from `lazy_free_space_updater` but unfortunately I have not time to fix it. I will use the pointcloud instead._

### DIFF
--- a/moveit_ros/perception/pointcloud_octomap_updater/include/moveit/pointcloud_octomap_updater/pointcloud_octomap_updater.hpp
+++ b/moveit_ros/perception/pointcloud_octomap_updater/include/moveit/pointcloud_octomap_updater/pointcloud_octomap_updater.hpp
@@ -37,6 +37,7 @@
 #pragma once
 
 #include <rclcpp/rclcpp.hpp>
+#include <rclcpp/callback_group.hpp>
 #include <rclcpp/version.h>
 #include <tf2_ros/transform_listener.h>
 #include <tf2_ros/message_filter.h>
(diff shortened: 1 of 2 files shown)


sea-bass (commented):
Thanks! To clarify, the fix is simply using a new callback group to process these subscriber callbacks separately from everything else, but the type of callback is less relevant in this case?

Anyways, just one very small comment for readability.

sea-bass:
Also seems you are right about the `image_transport` `CameraSubscriber` not having an input for `rclcpp::SubscriberOptions`.

I looked at https://github.com/ros-perception/image_common and it wouldn't be too hard to add by just copying how it's done for the `Subscriber`... but it is extra work.

sea-bass (approved):
Great, thank you!

I'm gonna request one more review on this one just in case.

henningkayser (approved):
Thank you! Do you think we should be switching to another API for depth image processing?

sjahr:
Thanks! Do you mind opening and issue for 

> I noticed that there is a big computational time mismatch between the two approaches (depth/pointcloud). After digging in a bit i found that the main computational time difference comes from lazy_free_space_updater but unfortunately I have not time to fix it. I will use the pointcloud instead._

PR #3225, opened 2025-01-10
Fix bad logic in Servo twist speed scaling

sea-bass:
### Description

I mistakenly added some logic to the twist mode of MoveIt Servo in https://github.com/moveit/moveit2/pull/3007 that did not need to be there.

When the speed commands are in real speed units, we should not touch their scaling value as documented!

### Checklist


sea-bass:
Wait no, I read the comment wrong. This is actually fine.

It says:

"Max linear velocity in m/s if using speed units or tracking a pose, else a scaling factor on the unitless commanded velocity. Used for Cartesian and pose tracking commands."

PR #3227, opened 2025-01-11
Remove plugins from export set

pac48:
### Description

Plugins that are intended to be loaded by classloader or pluginlib should not be in export sets that `ament_export_targets` is called on. This can cause plugins to fail to load.


sjahr (approved):
Makes sense, thanks!

PR #3233, opened 2025-01-12
Remove source builds of `geometric_shapes` and `srdfdom`

sea-bass:
### Description

We had added these a few months back to circumvent some breakages, but this should no longer be needed.

Q: Should we backport this to Jazzy? Because if so, `srdfdom` there is at 2.0.5 and not 2.0.7, so I think we may need to first release to there.

### Checklist

### DIFF
--- a/moveit2.repos
+++ b/moveit2.repos
@@ -1,9 +1,5 @@
 repositories:
   # Keep moveit_* repos here because they are released with moveit
-  geometric_shapes:
-    type: git
-    url: https://github.com/moveit/geometric_shapes.git
-    version: ros2
   moveit_msgs:
     type: git
     url: https://github.com/moveit/moveit_msgs.git
@@ -11,8 +7,4 @@ repositories:
   moveit_resources:
     type: git
     url: https://github.com/moveit/moveit_resources.git
-    version: ros2
-  srdfdom:
-    type: git
-    url: https://github.com/moveit/srdfdom.git
     version: ros2


sea-bass:
Bleh. This error seems familiar but I forget what it was.

```
geninfo: ERROR: unable to open /opt/ros/rolling/include/srdfdom/srdfdom/model.h: No such file or directory
  	(use "geninfo --ignore-errors source ..." to bypass this error)

sea-bass:
Maybe we hold off on this until https://github.com/moveit/geometric_shapes/pull/258 is released as well?

sea-bass:
> Bleh. This error seems familiar but I forget what it was.
> 
> ```
> geninfo: ERROR: unable to open /opt/ros/rolling/include/srdfdom/srdfdom/model.h: No such file or directory
>   	(use "geninfo --ignore-errors source ..." to bypass this error)
> ```
> 
> Update: Ended up ignoring it, and it makes CI pass, but unclear whether this is right.

This is annoying. Seems that the coverage tool has analyzed files in `geometric_shapes` that were previously cloned via vcs, but removing this is what causes the issue.

Is there a better way to clear out this state that doesn't lead to ignoring all missing source code errors?

mikeferguson (approved):

mikeferguson:
Since it's the code coverage tool, I think the only way to reset the cache of what it is comparing against is to merge the PR (when we merge, it will run the CI, and then that is the new baseline that all PRs are compared against)

PR #3251, opened 2025-01-20
Update robot state if time since last command exceeds timeout

KazuyaOguma18:
### Description

When starting both move_group and moveit_servo simultaneously, controlling the arm with move_group first and then performing jog operations with moveit_servo resulted in the jog operations being executed from the arm's initial pose, without considering the results of the previous control.

To address this issue, the robot's state is now updated synchronously, halting other processes if no control commands are received for a duration exceeding the incoming_command_timeout, which indicates that no jog operations are being executed.

### Checklist


sea-bass (commented):
Thank you for this!

I'm wondering, how is this logic different from what is already in e.g. https://github.com/moveit/moveit2/blob/2b9173b372b4e33559104d0afdb99f32b98eb01e/moveit_ros/moveit_servo/src/servo_node.cpp#L213-L229?

Would you be able to instead fix the existing logic instead of adding new lines of code?

KazuyaOguma18:
@sea-bass 
Thank you for reviewing!

> Would you be able to instead fix the existing logic instead of adding new lines of code?

Yes, I can. Is it correct to write the added code when command_stale in jog, twist and pose fuction is true?

sea-bass:
> Yes, I can. Is it correct to write the added code when command_stale in jog, twist and pose fuction is true?

Exactly! It would also be appreciated if you could write a unit test for this logic, to verify the fixes.

moveit_ros/moveit_servo/src/servo_node.cpp
```diff
@@ -414,4 +417,4 @@ void ServoNode::servoLoop()
 }  // namespace moveit_servo
 
 #include "rclcpp_components/register_node_macro.hpp"
-RCLCPP_COMPONENTS_REGISTER_NODE(moveit_servo::ServoNode)
+RCLCPP_COMPONENTS_REGISTER_NODE(moveit_servo::ServoNode)
```
sea-bass, line 420: Removing this newline will fail the formatting check.

sea-bass (commented):
Looking at these changes, I now wonder why it is not sufficient to get `last_commanded_state_` in here?

https://github.com/moveit/moveit2/blob/f821ad7a0fa92314fa0834b8360bf261e7d62a68/moveit_ros/moveit_servo/src/servo_node.cpp#L320

This logic is quite difficult to follow and full of hidden state, so I am likely missing something. 

I ask because maybe there is a better place to update this variable in one place instead of doing it repetitively and multiple times in a single loop iteration of the servo node?

sea-bass:
Ah, I think I see the issue now!

When I last ran into this case, it was because I was pausing servo, using move_group to command the robot, and then resuming servo before sending new commands.

If you do the same without pausing and unpausing servo (which does write to `last_commanded_state_`), then returning from a state of stale commands to new commands will indeed use old state.

I still think there are better ways to resolve this, maybe in the main servo loop instead of the individual `processXYZ()` functions. Maybe keeping around something that tracks whether the previous command was stale? What you would want is something like this:

```cpp
if (previous command was stale and new one is not)
{
    last_commanded_state_ = current state;
}
```

One way I can think of doing this is by making `last_commanded_state_` a `std::optional`, and when the servo commands go stale you set them to `std::nullopt`. So then this logic becomes

```cpp
if (!last_commanded_state_)
{
    last_commanded_state_ = current state;
}
```

What do you think?

sea-bass:
Follow up as I was reading the code on my computer instead of my phone...

Does the same issue get resolved if you update `last_commanded_state_` to the current state in this block of code?

https://github.com/moveit/moveit2/blob/23022683efc7e16b48a4de7c72e908f8bb38988d/moveit_ros/moveit_servo/src/servo_node.cpp#L348-L351

KazuyaOguma18:
> I still think there are better ways to resolve this, maybe in the main servo loop instead of the individual processXYZ() functions. Maybe keeping around something that tracks whether the previous command was stale? What you would want is something like this:

I also felt that updating last_command_state_ in many places would cause bugs, so I agree with using std::optional to determine valid and invalid values.
However, I need to add a way to deal with the case of nullopt in the input to functions that use last_command_state_.

> Does the same issue get resolved if you update last_commanded_state_ to the current state in this block of code?

Indeed, this line is similar to the current problem of not updating the robot.
What is curious is that this line is not updated with blocking.

KazuyaOguma18:
> I still think there are better ways to resolve this, maybe in the main servo loop instead of the individual processXYZ() functions. Maybe keeping around something that tracks whether the previous command was stale? What you would want is something like this:

I also felt that updating last_command_state_ in many places would cause bugs, so I agree with using std::optional to determine valid and invalid values.
However, I need to add a way to deal with the case of nullopt in the input to functions that use last_command_state_.

> Does the same issue get resolved if you update last_commanded_state_ to the current state in this block of code?

Indeed, this line is similar to the current problem of not updating the robot.
What is curious is that this line is not updated with blocking.

sea-bass:
> > Does the same issue get resolved if you update last_commanded_state_ to the current state in this block of code?
> 
> Indeed, this line is similar to the current problem of not updating the robot. What is curious is that this line is not updated with blocking.

Right. This is to prevent the servo loop from taking too long to compute, whereas on pause/unpause this happens separately.

Even with blocking set to false, could you try whether this still works fine for your use case?

If you want to guarantee an up-to-date current state, my recommendation would be call the "pause_servo" service, then motion plan with move_group, and finally unpause servo with the same service.

KazuyaOguma18:
> Does the same issue get resolved if you update last_commanded_state_ to the current state in this block of code?

I checked this, but even when using current_state = servo_->getCurrentRobotState(true /* block for current robot state */);, the robot's state was not updated to the latest.

When I checked the condition expression, the size of joint_cmd_rolling_window_ was around 20.

Upon further review, I found that the value of next_joint_state being nullopt indicates that the robot is in a stopped command state and that no next target pose exists. So, I tried updating the robot's state in a blocking manner under these conditions. What do you think?

sea-bass:
> Upon further review, I found that the value of next_joint_state being nullopt indicates that the robot is in a stopped command state and that no next target pose exists. So, I tried updating the robot's state in a blocking manner under these conditions. What do you think?

Thank you for checking!

I am worried that this blocking call will mean the overall servo loop might miss its frequency (which is not desired especially for real-time applications) -- so I think it will be good to avoid blocking during the main loop.

However, it seems the place you have now added the code already has access to a `current_state` variable.

So does this happen to work for your problem instead?

```cpp
if (!next_joint_state) {
  last_commanded_state_ = current_state_;
}
```

KazuyaOguma18:
Not using blocking processing does not mean that execution cannot be performed using the robot's latest state. I found that the issue was actually caused by `joint_cmd_rolling_window_` not containing the current posture at all.

Therefore, if a valid `next_joint_state` does not exist, I modified it to retrieve the latest robot state in a non-blocking manner using `last_commanded_state_` and `current_state`.

sea-bass (approved):
Wonderful! Thank you for iterating with me on this, @KazuyaOguma18. I think the changes now are very simple, and if they solve your original issue, I am happy to approve.

sjahr (approved):

PR #3265, opened 2025-01-24
Return if requested pause state is already active.

Jelmerdw:
### Description

When using the servo_node from moveit_servo, the service call to /servo_node/pause_servo hangs for a while (sometimes even up to a minute on my machine) when requesting pause = false while the pause status was already false. This behavior can be experienced by running the demo:

`ros2 launch moveit_servo demo_ros_api.launch.py`

And calling the service:

`ros2 service call /servo_node/pause_servo std_srvs/srv/SetBool '{data: false}'`

After a while, the following error is logged:

`[moveit_servo_demo_container.moveit.ros.move_group.collision_monitor]: Collision monitor could not be started`

To avoid this behavior, I propose this change to first check whether the requested pause state is different compared to the current pause state. If not, the response will be returned with the message that nothing is changed since the requested pause stated was already active.

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/src/servo_node.cpp
+++ b/moveit_ros/moveit_servo/src/servo_node.cpp
@@ -147,6 +147,13 @@ ServoNode::ServoNode(const rclcpp::NodeOptions& options)
 void ServoNode::pauseServo(const std::shared_ptr<std_srvs::srv::SetBool::Request>& request,
                            const std::shared_ptr<std_srvs::srv::SetBool::Response>& response)
 {
+  if (servo_paused_ == request->data)
+  {
+    RCLCPP_INFO(node_->get_logger(), "Requested pause state is already active.");
+    response->success = true;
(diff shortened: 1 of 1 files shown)


sea-bass (commented):
This is a good change, but also noting that https://github.com/moveit/moveit2/pull/3259 was recently merged which may help with this. Checking if you have this commit locally?

Besides that, just one small comment to reduce duplication.

Jelmerdw:
> This is a good change, but also noting that #3259 was recently merged which may help with this. Checking if you have this commit locally?

Thanks for this suggestion, I didn't have that commit locally indeed. Now I get the *Collision monitor could not be started* error immediately, which is good enough for me. 

I did apply you suggestion. Feel free to close this PR (since my actual problem was already solved by #3259), or to still merge it if you like to have this check on the servo pause state!

sea-bass (approved):

PR #3270, opened 2025-01-25
Enhanced documentation for composeMultiArrayMessage function

attu0:
Description
I have enhanced the documentation for the composeMultiArrayMessage function in common.cpp within the moveit_servo module.

Added a detailed Doxygen-compatible comment block for the function to clarify its purpose, parameters, return value, and usage.
The changes address the issue related to the lack of documentation for handling controllers that accept std_msgs::msg::Float64MultiArray messages.
This update improves developer experience and aligns with MoveIt 2's goal of becoming an industry-standard robotics framework.

Related Issue: https://github.com/moveit/moveit2/issues/3243

### DIFF
--- a/moveit_ros/moveit_servo/src/utils/common.cpp
+++ b/moveit_ros/moveit_servo/src/utils/common.cpp
@@ -255,6 +255,22 @@ void updateSlidingWindow(KinematicState& next_joint_state, std::deque<KinematicS
   joint_cmd_rolling_window.push_back(next_joint_state);
 }
 
+/**
+ * @brief Composes a Float64MultiArray message for controllers requiring Float64MultiArray input.
+ * 
+ * This function converts the given joint state into a Float64MultiArray message.
+ * 
+ * @param servo_params Configuration parameters for the servo, including message type selection.
+ * @param joint_state The current joint state to be converted into the message.
+ * @return A Float64MultiArray message containing the joint state data.
+ * 
+ * Example:
+ * @code
+ * auto message = composeMultiArrayMessage(servo_params, joint_state);
+ * controller.publish(message);
+ * @endcode
+ */
+
 std_msgs::msg::Float64MultiArray composeMultiArrayMessage(const servo::Params& servo_params,
                                                           const KinematicState& joint_state)
 {


moveit_ros/moveit_servo/src/utils/common.cpp
```diff
@@ -255,6 +255,22 @@ void updateSlidingWindow(KinematicState& next_joint_state, std::deque<KinematicS
   joint_cmd_rolling_window.push_back(next_joint_state);
 }
 
+/**
+ * @brief Composes a Float64MultiArray message for controllers requiring Float64MultiArray input.
+ * 
+ * This function converts the given joint state into a Float64MultiArray message.
+ * 
+ * @param servo_params Configuration parameters for the servo, including message type selection.
+ * @param joint_state The current joint state to be converted into the message.
```
AndyZe, line 264:
```suggestion
 * @param joint_state The joint state to be converted into the message.
```

AndyZe: Nitpick: usually this wouldn't be "current", it would be desired or target.

AndyZe (commented):
Looks great to me, just one little nitpick. I've been meaning to write some documentation myself for a while now

AndyZe (approved):

sea-bass (changes requested):
Thanks for this! There is already a docstring in the `common.hpp` file (which is where docstrings go instead of the `.cpp` files, unless the function is not in the header at all): https://github.com/moveit/moveit2/blob/524ca4f15bb449e1f8d25223d9725ca7868c32ee/moveit_ros/moveit_servo/include/moveit_servo/utils/common.hpp#L137-L144

Could you please take these improvements and combine them with the existing one?

moveit_ros/moveit_servo/src/utils/common.cpp
```diff
@@ -255,6 +255,22 @@ void updateSlidingWindow(KinematicState& next_joint_state, std::deque<KinematicS
   joint_cmd_rolling_window.push_back(next_joint_state);
 }
 
+/**
+ * @brief Composes a Float64MultiArray message for controllers requiring Float64MultiArray input.
+ * 
+ * This function converts the given joint state into a Float64MultiArray message.
```
sea-bass, line 261: these two sentences virtually say the same thing -- you probably can combine these into just a single line.

moveit_ros/moveit_servo/src/utils/common.cpp
```diff
@@ -255,6 +255,22 @@ void updateSlidingWindow(KinematicState& next_joint_state, std::deque<KinematicS
...
+ * 
+ * @param servo_params Configuration parameters for the servo, including message type selection.
+ * @param joint_state The joint state to be converted into the message.
+ * @return A Float64MultiArray message containing the joint state data.
+ * 
+ * Example:
+ * @code
+ * auto message = composeMultiArrayMessage(servo_params, joint_state);
+ * controller.publish(message);
```
sea-bass, line 270: It may not be obvious to most people that `controller` in this case is actually a ROS 2 publisher. Maybe call this variable `controller_pub` or something like that?

Also looking at other examples, it appears you may need to indent what is inside the `@code` / `@endcode`?

For example: https://github.com/moveit/moveit2/blob/524ca4f15bb449e1f8d25223d9725ca7868c32ee/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.hpp#L607-L610

Which renders to: https://moveit.picknik.ai/main/api/html/classplanning__scene__monitor_1_1LockedPlanningSceneRO.html#details

PR #3274, opened 2025-01-27
Use attached object colors as is in Rviz plugin

### LINKED ISSUE #3273: Rviz plugin overrides attached object colors incorrectly.
### Description

https://github.com/moveit/moveit2/blob/524ca4f15bb449e1f8d25223d9725ca7868c32ee/moveit_ros/visualization/rviz_plugin_render_tools/src/robot_state_visualization.cpp#L135-L139

changes color values in a way that does not make sense. Value are supposed to be from 0 to 1, so having max(1.0, something) makes them either always be 1.0, or be > 1.0.  This makes it impossible to get the attached objects to show up with the desired colors, or at least I have not figured out how to work around this unexpected code. I am guessing the code (added in 00a0686be7d45d331ef4b7e3b764a462d8a6c405 in 2013!) was supposed to have `min` instead of `max` (so that it's 1.05 * the specified value, but capped at 1.0), but a typo there went unnoticed for 12 years!

Should I replace "max" with "min"? Or just change leave the specified values alone?

### ROS Distro

Rolling

### OS and version

Ubuntu 22.04

### Source or binary build?

Source

_No response_

### If source, which branch?

main

### Which RMW are you using?

FastRTPS

### Steps to Reproduce

Use the `apply_planning_scene` service to define attached collision objects with custom colors, while displaying the robot in rviz

### Expected behavior

Colors reflect the requested color.

### Actual behavior

Colors are wrong.

_No response_

ANogin:
### Description

Fixes #3273. As explained in #3273, there is some weird code from 2013 that messes up the colors of attached objects in rviz, overriding the colors provided via the planning scene interface with potentially out of range values. This removes the erroneous code and just passes the colors as is.

### Checklist


sea-bass (approved):

rhaschke (approved):
I'm curious under which circumstances this bug manifested itself. I never noticed a fully-white attached body.
Anyway, given that fact, that nobody noticed that issue yet, we can safely use the original color.

rr-mark:
> I'm curious under which circumstances this bug manifested itself. I never noticed a fully-white attached body. Anyway, given that fact, that nobody noticed that issue yet, we can safely use the original color.

I think we were previously seeing these changes, but had a workaround to avoid them (see e.g. https://robotics.stackexchange.com/questions/113543/how-can-attached-collision-objects-be-recoloured-in-rviz2).

It's relatively hard to change the colour of an attached collision object away from default, so I'm not surprised not many people ran into this behaviour.

rhaschke:
Now, I'm even more surprised: as MoveIt1 used the same code, I'm wondering why it didn't display the attached object as white.

rr-mark:
> Now, I'm even more surprised: as MoveIt1 used the same code, I'm wondering why it didn't display the attached object as white.

The behaviour we saw before our workaround was that attached collision objects would be the default colour until recoloured, at which point they would be white for any valid colour input.

Our workaround is to have a separate node listening to moveit's planning scene, recolouring all the objects, and republishing to rviz. Hopefully that won't be needed after we start using this PR.

PR #3276, opened 2025-01-27
Support including the names of other attached objects in `touch_link`

### LINKED ISSUE #3275: Should be possible to specify two attached objects as allowed to touch.
### Is your feature request related to a problem? Please describe.

Currently, each `ArmttachedObject` can be specified with a `touch_links` array that makes it allowable for the object to collide with the given robot links. However, any pair of separate attached objects, unless they are attached to the same robot link (and therefore their relative position would never change) is always checked for collisions: https://github.com/moveit/moveit2/blob/524ca4f15bb449e1f8d25223d9725ca7868c32ee/moveit_core/collision_detection_fcl/src/collision_common.cpp#L158-L163

Imagine a use case of e.g. a robot with two manipulators, each with an object attached, where MoveIt IK and path planning are used to get the two objects to touch - right now this is impossible to do, at least in a straightforward way. Ideally, it should be possible to allow specific object/object collisions, similarly to how it is already possible to allow link/link and object/link collisions.

### Describe the solution you'd like

I think the most straightforward way would be to just overload `touch_links`, making it possible to also include names of other attached objects there. If at least one of the two objects contains the other's name in its `touch_link`, then the collision would be allowed. I am hoping (and plan to test in a bit) that only the code around the above-quoted lines would have to be modified.

### Describe alternatives you've considered

One downside of overloading the `touch_links` is that the field name would become a misleading. Also, I do not know whether it is possible to have an attached object with the same name as an existing link - if so, the proposed approach would clearly cause problems. 

An alternative would be to create a separate `touch_objects` paralleling `touch_links`. Of course that would require a bunch of API and message type changes, and would potentially be a lot more invasive.

Finally, it may be a bit safer to only allow collisions when both objects include the other one in their `touch_links`. However, this is likely to necessitate additional somewhat messy code on the part of the user of the API, keeping the two sides in sync every time a collision is allowed or disallowed, and my expectation is that the "collisions are allowed when at least one object names the other in `touch_links`" semantics would be easier to use than "each object names the other" semantics.

### Additional context

_No response_

ANogin:
### Description

Fixes #3275. See that RFE for motivation and discussion of pros/cons and alternative approaches.

### Checklist

### DIFF
--- a/moveit_core/collision_detection_fcl/src/collision_common.cpp
+++ b/moveit_core/collision_detection_fcl/src/collision_common.cpp
@@ -160,6 +160,21 @@ bool collisionCallback(fcl::CollisionObjectd* o1, fcl::CollisionObjectd* o2, voi
   {
     if (cd1->ptr.ab->getAttachedLink() == cd2->ptr.ab->getAttachedLink())
       always_allow_collision = true;
+    else
+    {
+      const std::set<std::string>& tl1 = cd1->ptr.ab->getTouchLinks();
+      const std::set<std::string>& tl2 = cd2->ptr.ab->getTouchLinks();
+      if (tl1.find(cd2->getID()) != tl1.end() || tl2.find(cd1->getID()) != tl2.end())
+      {
+        always_allow_collision = true;
+      }
+    }
+    if (always_allow_collision && cdata->req_->verbose)
+    {
+      RCLCPP_DEBUG(getLogger(),
+                   "Attached object '%s' is allowed to touch attached object '%s'. No contacts are computed.",
+                   cd2->getID().c_str(), cd1->getID().c_str());
+    }
   }
 
   // if collisions are always allowed, we are done


moveit_core/collision_detection_fcl/src/collision_common.cpp
```diff
@@ -160,6 +160,21 @@ bool collisionCallback(fcl::CollisionObjectd* o1, fcl::CollisionObjectd* o2, voi
   {
     if (cd1->ptr.ab->getAttachedLink() == cd2->ptr.ab->getAttachedLink())
       always_allow_collision = true;
+    else
+    {
+      const std::set<std::string>& tl1 = cd1->ptr.ab->getTouchLinks();
+      const std::set<std::string>& tl2 = cd2->ptr.ab->getTouchLinks();
+      if (tl1.find(cd2->getID()) != tl1.end() || tl2.find(cd1->getID()) != tl2.end())
```
sea-bass, line 167: You can now use `tl1.contains(cd2->getID())` and vice versa

moveit_core/collision_detection_fcl/src/collision_common.cpp
```diff
@@ -160,6 +160,21 @@ bool collisionCallback(fcl::CollisionObjectd* o1, fcl::CollisionObjectd* o2, voi
   {
     if (cd1->ptr.ab->getAttachedLink() == cd2->ptr.ab->getAttachedLink())
       always_allow_collision = true;
+    else
+    {
+      const std::set<std::string>& tl1 = cd1->ptr.ab->getTouchLinks();
```
sea-bass, line 165: A comment in this block may be helpful for those reading the code

sea-bass (commented):
Thank you! Just very minor suggestions from me.

Also, would you be willing to write a test for this?

moveit_core/collision_detection_fcl/src/collision_common.cpp
sea-bass: Ah, bummer. I think it may require adding c++20 support in the cmakelists.txt. Never mind then.

ANogin:
> Also, would you be willing to write a test for this?

Willing - yes. Have bandwidth some time soon to figure out how tests are set up and how to do this properly - quite unlikely, unfortunately... :(

moveit_core/collision_detection_fcl/src/collision_common.cpp
ANogin: Done

sea-bass (approved):

sjahr (approved):

PR #3280, opened 2025-01-28
Ports a6f8bb9 331381a 593a119 and 593a119 from moveit1

rr-mark:
### Description

 - Fixes #3279 
 - Reverts 1f23344
 - Ports a6f8bb9 331381a 593a119 and 4a3a312 from moveit1 (skipping changes to cartesian_interpolator.cpp, which has been completely re-written since those PRs).

Please back-port to humble when merged.

### Checklist
- [ ] **Required by CI**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)
- [ ] Extend the tutorials / documentation [reference](http://moveit.ros.org/documentation/contributing/)
- [ ] Document API changes relevant to the user in the [MIGRATION.md](https://github.com/moveit/moveit2/blob/main/MIGRATION.md) notes
- [ ] Create tests, which fail without this PR [reference](https://moveit.picknik.ai/humble/doc/examples/tests/tests_tutorial.html)
- [ ] Include a screenshot if changing a GUI
- [ ] While waiting for someone to review your request, please help review [another open pull request](https://github.com/moveit/moveit2/pulls) to support the maintainers

[//]: # "You can expect a response from a maintainer within 7 days. If you haven't heard anything by then, feel free to ping the thread. Thank you!"

PR #3284, opened 2025-01-30
Reverts #2985, Ports moveit #3388 #3470 #3539

### LINKED ISSUE #3279: `RobotState::getRigidlyConnectedParentLinkModel` and subframes: PRs not ported from moveit1.
`RobotState::getRigidlyConnectedParentLinkModel` fails when the frame is a subframe.

In moveit1, this is fixed by a6f8bb9 331381a 593a119 and 4a3a312.

It is not straightforward to patch these across, as Moveit2 has 1f23344 and f20265a.

Should 1f23344 and f20265a be reverted, and the four PRs from Moveit1 be patched across? Or are 1f23344 and f20265a still relevant after those four PRs? And if so, should they be backported to Moveit1?

rr-mark:
### Description

 - Fixes #3279 
 - As https://github.com/moveit/moveit2/pull/3280, but with history correctly preserved. (Also includes a few commits in the same PRs, which #3280 missed)
 - N.B. does not include changes to `cartesian_path_service_capability`, as that has changed significantly since the Moveit1 PRs were made.
 - Please back-port to Humble and Jazzy when merged

### Checklist

### DIFF
--- a/moveit_core/robot_model/include/moveit/robot_model/robot_model.hpp
+++ b/moveit_core/robot_model/include/moveit/robot_model/robot_model.hpp
@@ -256,6 +256,9 @@ public:
 
   /** \brief Get the latest link upwards the kinematic tree, which is only connected via fixed joints
    *
+   * If jmg is given, all links that are not active in this JMG are considered fixed.
+   * Otherwise only fixed joints are considered fixed.
+   *
    * This is useful, if the link should be warped to a specific pose using updateStateWithLinkAt().
    * As updateStateWithLinkAt() warps only the specified link and its descendants, you might not
    * achieve what you expect, if link is an abstract frame name. Considering the following example:
@@ -265,7 +268,8 @@ public:
    * what you went for. Instead, updateStateWithLinkAt(getRigidlyConnectedParentLinkModel(grasp_frame), ...)
    * will actually warp wrist (and all its descendants).
    */
-  static const moveit::core::LinkModel* getRigidlyConnectedParentLinkModel(const LinkModel* link);
+  static const moveit::core::LinkModel* getRigidlyConnectedParentLinkModel(const LinkModel* link,
+                                                                           const JointModelGroup* jmg = nullptr);
 
   /** \brief Get the array of links  */
   const std::vector<const LinkModel*>& getLinkModels() const
--- a/moveit_core/robot_model/include/moveit/robot_model/robot_model.hpp
+++ b/moveit_core/robot_model/include/moveit/robot_model/robot_model.hpp
@@ -269,7 +269,14 @@ public:
    * will actually warp wrist (and all its descendants).
    */
   static const moveit::core::LinkModel* getRigidlyConnectedParentLinkModel(const LinkModel* link,
+                                                                           Eigen::Isometry3d& transform,
                                                                            const JointModelGroup* jmg = nullptr);
+  static const moveit::core::LinkModel* getRigidlyConnectedParentLinkModel(const LinkModel* link,
+                                                                           const JointModelGroup* jmg = nullptr)
+  {
+    Eigen::Isometry3d unused;
+    return getRigidlyConnectedParentLinkModel(link, unused, jmg);
+  }
 
   /** \brief Get the array of links  */
   const std::vector<const LinkModel*>& getLinkModels() const
--- a/moveit_core/robot_model/include/moveit/robot_model/robot_model.hpp
+++ b/moveit_core/robot_model/include/moveit/robot_model/robot_model.hpp
@@ -268,11 +268,11 @@ public:
    * what you went for. Instead, updateStateWithLinkAt(getRigidlyConnectedParentLinkModel(grasp_frame), ...)
    * will actually warp wrist (and all its descendants).
    */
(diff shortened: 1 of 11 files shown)


moveit_core/robot_model/include/moveit/robot_model/robot_model.hpp
```diff
@@ -265,7 +268,13 @@ class RobotModel
    * what you went for. Instead, updateStateWithLinkAt(getRigidlyConnectedParentLinkModel(grasp_frame), ...)
    * will actually warp wrist (and all its descendants).
    */
-  static const moveit::core::LinkModel* getRigidlyConnectedParentLinkModel(const LinkModel* link);
+  static const LinkModel* getRigidlyConnectedParentLinkModel(const LinkModel* link, Eigen::Isometry3d& transform,
+                                                             const JointModelGroup* jmg = nullptr);
+  static const LinkModel* getRigidlyConnectedParentLinkModel(const LinkModel* link, const JointModelGroup* jmg = nullptr)
```
sjahr, line 273: Do we need these two functions? Wouldn't this be sufficient:

```
static const LinkModel* getRigidlyConnectedParentLinkModel(const LinkModel* link, Eigen::Isometry3d& transform = Eigen::Isometry3d(), const JointModelGroup* jmg = nullptr);
```
That way we don't break the existing API and don't add another function. Downside would be that you cannot call it with just two arguments but I think that is fine

moveit_core/robot_model/include/moveit/robot_model/robot_model.hpp
```diff
@@ -265,7 +268,13 @@ class RobotModel
    * what you went for. Instead, updateStateWithLinkAt(getRigidlyConnectedParentLinkModel(grasp_frame), ...)
    * will actually warp wrist (and all its descendants).
    */
-  static const moveit::core::LinkModel* getRigidlyConnectedParentLinkModel(const LinkModel* link);
+  static const LinkModel* getRigidlyConnectedParentLinkModel(const LinkModel* link, Eigen::Isometry3d& transform,
```
sjahr, line 271: Is the transform intended to be some implicit output? I don't like this API, especially given that there is no documentation around this provided. I think we need to add documentation of the parameters of this function and if a reference is passed as non-const to a function in order to be modified by it, I think it should be the last argument so that they are ordered, first inputs and second output argument.

moveit_core/robot_model/src/robot_model.cpp
```diff
@@ -1364,15 +1364,34 @@ LinkModel* RobotModel::getLinkModel(const std::string& name, bool* has_link)
...
   const moveit::core::LinkModel* parent_link = link->getParentLinkModel();
   const moveit::core::JointModel* joint = link->getParentJointModel();
+  decltype(jmg->getJointModels().cbegin()) begin{}, end{};
+  if (jmg)
+  {
+    begin = jmg->getJointModels().cbegin();
+    end = jmg->getJointModels().cend();
+  }
+
+  auto is_fixed_or_not_in_jmg = [begin, end](const JointModel* joint) {
```
sjahr, line 1383: Can you add a quick doc string here that explains what this does?

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -907,14 +907,50 @@ void RobotState::updateStateWithLinkAt(const LinkModel* link, const Eigen::Isome
...
+                                                                const moveit::core::JointModelGroup* jmg) const
 {
-  bool found;
   const LinkModel* link{ nullptr };
-  getFrameInfo(frame, link, found);
-  if (!found)
-    RCLCPP_ERROR(getLogger(), "Unable to find link for frame '%s'", frame.c_str());
-  return getRobotModel()->getRigidlyConnectedParentLinkModel(link);
+
+  if (getRobotModel()->hasLinkModel(frame))
```
sjahr, line 915: Same here, we need documentation around these complicated if else constructs

sjahr (commented):
Thanks for porting this over! I hate to ask but do you mind adding a bit documentation here? Moveit core is already complicated enough and these changes don't simplify it  :sweat_smile:

rr-mark:
> Thanks for porting this over! I hate to ask but do you mind adding a bit documentation here? Moveit core is already complicated enough and these changes don't simplify it 😅

Sure thing. None of this was originally my code, but I'll see what I can do later today.

sjahr:
> Sure thing. None of this was originally my code, but I'll see what I can do later today.

Yes, I am aware of this. Thanks a lot for doing it and let me know when you need help!

rr-mark:
It looks like the dubious `transform` arguments are only there for unit tests. I'm trying to work out how to calculate them separately, so they can be deleted.

rr-mark:
Actually, scratch that. Any alternate code to calculate the transform would only be used in the tests. I think I'm just going to delete these arguments.

sjahr (approved):
Thank you!

sjahr:
@rr-mark Thank's a lot for doing all the work here! Last request: Do you mind squashing all commits authored by only together? I am going to merge this PR with a merge commit to preserver the commits ported over from moveit1 and I want to keep the history on main clean.

PR #3294, opened 2025-02-01
fix: pointcloud_octomap_updater not cleaning objects at max_range

MarcoMagriDev:
Fix https://github.com/moveit/moveit2/issues/3286

### DIFF
--- a/moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
+++ b/moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
@@ -338,6 +338,7 @@ void PointCloudOctomapUpdater::cloudMsgCallback(const sensor_msgs::msg::PointClo
     /* compute the free cells along each ray that ends at a clipped cell */
     for (const octomap::OcTreeKey& clip_cell : clip_cells)
     {
+      free_cells.insert(clip_cell);
       if (tree_->computeRayKeys(sensor_origin, tree_->keyToCoord(clip_cell), key_ray_))
         free_cells.insert(key_ray_.begin(), key_ray_.end());
     }


sea-bass (approved):
This LGTM as I was discussing in the related issue -- would appreciate a second maintainer look at the write up in the issue and confirm whether this is okay.

mikeferguson:
## Question on Configuration

First, a question - since your screen grabs in the original ticket are in simulation, I'm guessing that your "max_range" parameter is the same in simulation as MoveIt (Is that correct?) and so simulation ends up using INF for any pixel that would be greater than "max_range" (INF would match [REP-117](https://ros.org/reps/rep-0117.html)). Hence, you never get clearing data.

In most setups, I think people are setting the "max_range" of the octomap to a range less than the actual maximum range that the sensor can see (for instance, if your sensor has 5m range, you still really don't usually need to render more than a 1m octomap since your arm can't reach beyond that). So, that's a likely reason people haven't run into serious issues with this configuration.

## This PR

Now onto the PR - I'm not convinced this *always* works - especially outside of simulation - basically what this does is:

 * Makes a very thin shell (1 octomap cell resolution - often just 5cm thick) just inside the max range and marks it as CLIPPED rather than OUTSIDE.
 * This means the very outer cell of the octomap is never marked as OCCUPIED ([since the vector X,Y,Z is normalized and then extended to length of max_range](https://github.com/moveit/moveit2/blob/0af5e0f45447e24f118b6210d5d5e7eb92868245/moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp#L300))

My concern around "always" working is that if the object quickly moves through that region of the shell in the time span in between successive point clouds you would still not clear it out, right?

## Alternate Approach

I think what you're actually after might be similar to the "inf_is_valid" parameter in nav2 costmaps - if a pixel comes back as INF, you still want to clear to the max range along that ray. You're currently getting this for a very small part of the visual space with this PR.

mikeferguson:
Also - just to clarify on my previous message, I'm not opposed to this fix being merged, I'm just point out that I think it might not entirely fix the underlying symptom (making something that is currently "deterministically bad" into something more "flaky bad").

MarcoMagriDev:
> ## Question on Configuration
> First, a question - since your screen grabs in the original ticket are in simulation, I'm guessing that your "max_range" parameter is the same in simulation as MoveIt (Is that correct?) and so simulation ends up using INF for any pixel that would be greater than "max_range" (INF would match [REP-117](https://ros.org/reps/rep-0117.html)). Hence, you never get clearing data.
> 

I’ve set the sensor’s max perception range to 10m, which is twice the max_range specified in the MoveIt sensor configuration (5m).

> In most setups, I think people are setting the "max_range" of the octomap to a range less than the actual maximum range that the sensor can see (for instance, if your sensor has 5m range, you still really don't usually need to render more than a 1m octomap since your arm can't reach beyond that). So, that's a likely reason people haven't run into serious issues with this configuration.
> 
Even if this relationship between the two max_range values is maintained, it doesn’t eliminate the possibility of getting inf values. These still occur whenever points are beyond the sensor's max perception range.

> ## This PR
> Now onto the PR - I'm not convinced this _always_ works - especially outside of simulation - basically what this does is:
> 
> * Makes a very thin shell (1 octomap cell resolution - often just 5cm thick) just inside the max range and marks it as CLIPPED rather than OUTSIDE.
> * This means the very outer cell of the octomap is never marked as OCCUPIED ([since the vector X,Y,Z is normalized and then extended to length of max_range](https://github.com/moveit/moveit2/blob/0af5e0f45447e24f118b6210d5d5e7eb92868245/moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp#L300))
> 

Yes, that’s the trade-off with this fix. The new effective max_range will be the original value minus the octomap resolution, meaning we lose essentially one octomap cell’s thickness. However, I believe this is a reasonable compromise to ensure proper clearing behavior.

> My concern around "always" working is that if the object quickly moves through that region of the shell in the time span in between successive point clouds you would still not clear it out, right?
> 

In my tests, everything is successfully cleared, even in the scenario you described. Additionally, I can confirm that the issue was present not only in simulation but also with real hardware, where it’s even easier to spot. Simply walking in and out of the max range, or moving tangentially to it, clearly shows that before the fix, the octomap isn't cleaned properly, whereas after the fix, it is.

> ## Alternate Approach
> I think what you're actually after might be similar to the "inf_is_valid" parameter in nav2 costmaps - if a pixel comes back as INF, you still want to clear to the max range along that ray. You're currently getting this for a very small part of the visual space with this PR.

Correct me if I am wrong but points with inf values are already marked as CLIP [here](https://github.com/moveit/moveit2/blob/d88d19fb2a1539f1dc21f37a15ad1ce1454b6400/moveit_ros/perception/point_containment_filter/src/shape_mask.cpp#L187) but still not cleaned up

mikeferguson:
So, maybe not entirely relevant to this discussion - but I'm not exactly sure how INF is handled then. We are only [filtering NAN values](https://github.com/moveit/moveit2/blob/0af5e0f45447e24f118b6210d5d5e7eb92868245/moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp#L287), but then CLIP values [take the norm of the vector and multiply by max_range](https://github.com/moveit/moveit2/blob/0af5e0f45447e24f118b6210d5d5e7eb92868245/moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp#L300) - the norm with INF shouldn't give the proper ray angle though?

One last thought after I walked through that code again - your original ticket noted that subtracting 0.1 worked, is it actually the case we don't need to change the limits of the shape mask, but rather need to [change the condition here](https://github.com/moveit/moveit2/blob/d88d19fb2a1539f1dc21f37a15ad1ce1454b6400/moveit_ros/perception/point_containment_filter/src/shape_mask.cpp#L187) to be inclusive of max_sensor_dist? In other words, change > to >= in ```if (d < min_sensor_dist || d >= max_sensor_dist)```.

MarcoMagriDev:
> One last thought after I walked through that code again - your original ticket noted that subtracting 0.1 worked, is it actually the case we don't need to change the limits of the shape mask, but rather need to [change the condition here](https://github.com/moveit/moveit2/blob/d88d19fb2a1539f1dc21f37a15ad1ce1454b6400/moveit_ros/perception/point_containment_filter/src/shape_mask.cpp#L187) to be inclusive of max_sensor_dist? In other words, change > to >= in `if (d < min_sensor_dist || d >= max_sensor_dist)`.

This was my first attempt, but unfortunately, it did not work as expected.

---

After digging in a bit more, I may have found the actual source of the issue and a proper fix.

The issue is related to the `computeRayKeys` method called [here](https://github.com/moveit/moveit2/blob/d88d19fb2a1539f1dc21f37a15ad1ce1454b6400/moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp#L341). 

As stated in the [documentation](https://octomap.github.io/octomap/doc/classoctomap_1_1OcTreeBaseImpl.html#ada3bfe3e6f0ca6804141aa6ee25fc4d2) : 

> Traces a ray from origin to end (**excluding**), returning an [OcTreeKey](https://octomap.github.io/octomap/doc/classoctomap_1_1OcTreeKey.html) of all nodes traversed by the beam.

This means that we are missing the final cell along the ray, which corresponds exactly to our `clip_cell` value.

To address this, I propose modifying the [for_loop](https://github.com/moveit/moveit2/blob/d88d19fb2a1539f1dc21f37a15ad1ce1454b6400/moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp#L339) as follows:

```cpp
for (const octomap::OcTreeKey& clip_cell : clip_cells)
{
  free_cells.insert(clip_cell);
  if (tree_->computeRayKeys(sensor_origin, tree_->keyToCoord(clip_cell), key_ray_))
    free_cells.insert(key_ray_.begin(), key_ray_.end());
}
```

This explicitly adds the key corresponding to the extreme point that `computeRayKeys` does not handle.

mikeferguson:
> This explicitly adds the key corresponding to the extreme point that `computeRayKeys` does not handle.

That actually makes very good sense! I was having a very hard time understanding exactly how the initially proposed solution was actually fixing the problem - but the observation that we can't clear the outer cell makes sense.

mikeferguson (approved):

mikeferguson:
@MarcoMagriDev thanks for working with us to get to the root of the problem!

sea-bass (approved):
Nice!

PR #3301, opened 2025-02-03
Fix Ros2ControlManager chained controller logic

pac48:
### Description

It looks like [this](https://github.com/moveit/moveit2/pull/785) PR might have broken the chained controller management logic. 
Previously, `claimed_interfaces` were being used to determine which joints each controller was associated with, but now `required_command_interfaces` is used. That seems reasonable, but it was not updated everywhere. Specifically, there is some logic to handle ros2_control's chained controllers that was not updated. This PR fixes that issue.


sea-bass (approved):

sjahr (approved):
Thanks!

PR #3309, opened 2025-02-04
Add allowed_start_tolerance_joints parameter (Port moveit#3287)

DaniGarciaLopez:
### Description

Allow to specify a per joint allowed start tolerance for the TrajectoryExecutionManager. Port from https://github.com/moveit/moveit/pull/3287 with some changes to be able to dynamically update its values at runtime.

### Example of usage
In `moveit_controllers.yaml`:
```yaml
trajectory_execution:
  allowed_execution_duration_scaling: 1.2
  allowed_goal_duration_margin: 0.5
  allowed_start_tolerance: 0.01
  allowed_start_tolerance_joints: {panda_joint3: 0.02, panda_joint5: 0.03, panda_finger_joint1: 0.005}

```
Joints that are not declared in `allowed_start_tolerance_joints` will use `allowed_start_tolerance` as the default.

### Some remarks
- ~~I would personally have named this parameter "allowed_start_tolerance_joints" to be next to "allowed_start_tolerance" when sorted alphabetically. However, to maintain continuity with the original PR, I kept it as is.~~
- I left the tests as they are because `test_execution_manager.cpp` is [not used at the moment](https://github.com/moveit/moveit2/blob/fbdd8c51cdeb9858e5c77da427694ef0f89b6246/moveit_ros/planning/trajectory_execution_manager/CMakeLists.txt#L28C1-L42C8), right?
- Why [skip negative tolerances](https://github.com/moveit/moveit/pull/3287/commits/0d58f8d4e268f25ec665503a0d818272e46ff649)? I think, rather than defaulting to `allowed_start_tolerance`, it would be more intuitive to use the absolute value in that case.

### Checklist

### DIFF
--- a/moveit_ros/planning/trajectory_execution_manager/include/moveit/trajectory_execution_manager/trajectory_execution_manager.hpp
+++ b/moveit_ros/planning/trajectory_execution_manager/include/moveit/trajectory_execution_manager/trajectory_execution_manager.hpp
@@ -293,7 +293,8 @@ private:
   void loadControllerParams();
 
   double getJointAllowedStartTolerance(std::string const& joint_name) const;
-  void updateJointsAllowedStartTolerance();
+  void setJointAllowedStartTolerance(std::string const& joint_name, double joint_start_tolerance);
+  void initializeJointsAllowedStartTolerance();
 
   // Name of this class for logging
   const std::string name_ = "trajectory_execution_manager";
--- a/moveit_ros/planning/trajectory_execution_manager/include/moveit/trajectory_execution_manager/trajectory_execution_manager.hpp
+++ b/moveit_ros/planning/trajectory_execution_manager/include/moveit/trajectory_execution_manager/trajectory_execution_manager.hpp
@@ -292,6 +292,9 @@ private:
 
   void loadControllerParams();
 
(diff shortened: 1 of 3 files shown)


mikeferguson:
> * I would personally have named this parameter `allowed_start_tolerance_joints` to be next to `allowed_start_tolerance` when sorted alphabetically. However, to maintain continuity with the original PR, I kept it as is.

There are so many parameter changes between ROS 1 and ROS 2 MoveIt that I wouldn't put very high value on "continuity with the original PR" - "allowed_start_tolerance_joints" does seem to be a much better name for this - I would further suggest that "allowed_start_tolerance" should actually be "allowed_start_tolerance_default" - to distinguish more clearly how it differs from the "_joints" parameter

sea-bass (approved):
This looks good to me as is!

Agree with @mikeferguson that you can pick whatever name you like for the new parameter, but would be nice to keep the name of the existing one to prevent people from having to update their configs.

That being said, comments in the parameter YAMLs and in the code go a long way in explaining these things!

Also, don't worry about the tutorial images failures. A bunch of packages got booted off Rolling because of a deprecated ros2 control header that needs to be updated, so nothing to do with these changes.

sea-bass (approved):

mikeferguson (approved):

PR #3327, opened 2025-02-06
fix: ensure attached objects update during motion execution

MarcoMagriDev:
### Description

- Check that attached objects in the monitored robot match those in the planned trajectory.
- If an object disappears from the monitored robot, remove it from the trajectory way-point robot_state. (avoid false positives, stop trajectories due to collisions between environment and objects that are not attached to the robot anymore)
- If an object is attached to the monitored robot but missing in the trajectory, add it to enable meaningful collision checking. (avoids disruptive condition in which even if the object has been added to the robot as attached collision object it is not considered for online collision checking)

---
#### Implications

This could impact a few different situations:
- Imagine you’re catching an object during motion, and an external node tells the planning scene to add the new attached collision object. Even though MoveIt knows about the new attached object, the collision check still only considers the attached objects that were there at the planning stage, so it misses the new one. This could lead to a situation where a collision is missed because the newly attached object isn’t part of the check.
- Similarly, dropping an object during motion may result in preempting a valid path due to a non existsent collision.


moveit_ros/planning/plan_execution/src/plan_execution.cpp
```diff
@@ -283,34 +283,70 @@ bool plan_execution::PlanExecution::isRemainingPathValid(const ExecutableMotionP
     collision_detection::CollisionRequest req;
     req.group_name = t.getGroupName();
     req.pad_environment_collisions = false;
+    auto getAttachedObjects = [](const moveit::core::RobotState& state) {
+      std::vector<const moveit::core::AttachedBody*> attached_bodies;
+      state.getAttachedBodies(attached_bodies);
+      std::map<std::string, const moveit::core::AttachedBody*> attached_objects;
```
sea-bass, line 289: Does it make sense to add this function to `RobotState` itself instead? Like a `getAttachedBodiedMap()` function?

moveit_ros/planning/plan_execution/src/plan_execution.cpp
```diff
@@ -283,34 +283,70 @@ bool plan_execution::PlanExecution::isRemainingPathValid(const ExecutableMotionP
...
+        attached_objects[ab->getName()] = ab;
+      }
+      return attached_objects;
+    };
+
+    moveit::core::RobotState state = plan.planning_scene->getCurrentState();
+    std::map<std::string, const moveit::core::AttachedBody*> current_attached_objects = getAttachedObjects(state);
     for (std::size_t i = std::max(path_segment.second - 1, 0); i < wpc; ++i)
     {
+      state = t.getWayPoint(i);
```
sea-bass, line 301: Thank you for this!

moveit_ros/planning/plan_execution/src/plan_execution.cpp
MarcoMagriDev: That sounds like a good idea! Adding a `getAttachedBodiesMap()` function to RobotState could make it more reusable and improve clarity. Would you like to propose this as a separate PR, or do you plan to integrate it into your current changes?

sea-bass: I was suggesting including this in your PR, if you would be up for it?

MarcoMagriDev: I have just created the `getAttachedBodies` version that returns a map instead of a vector and updated the PR accordingly. Let me know if this aligns with your ideas!

sea-bass: It looks good, though I would consider not having it call the other one since it seems inefficient to create a vector, then loop through that to create a map. Maybe just go straight by looping through and putting together the map?

moveit_ros/planning/plan_execution/src/plan_execution.cpp
sea-bass: Ah thanks for asking.

I guess I would be thinking about:
1. Does it slow things down to be potentially attaching/detaching objects at each waypoint during this check?
2. It's difficult for me to know what the user's intent would be on a case by case basis. Right now it seems like the current state is treated as the source of truth, but I'm unsure whether users want that in every case. I guess the previous implementation did the opposite and treated the attached objects from the pre-planned waypoint as the source of truth.

Is it worth maybe adding a flag to this function for whether one wants to prioritize current state vs. planned states? And elevate this up to the config/parameter level?

MarcoMagriDev: > 1. Does it slow things down to be potentially attaching/detaching objects at each waypoint during this check?

I think it depends on how frequently the planning scene changes. In a highly dynamic environment, this method would be called often, which could impact performance. Since attached objects are processed for all robot states associated with waypoints after the currently executed one, the worst-case scenario with the current implementation occurs when scene updates happen frequently at the start of the trajectory.

With the proposed simplification, this reduces to simply "high-frequency scene updates," as only one robot state from the trajectory is queried for attached objects, while the rest are assumed unchanged. 

On the other hand, if future plans include attached objects that are not consistently present throughout the motion, this simplification could introduce errors.

> 2. It's difficult for me to know what the user's intent would be on a case by case basis. Right now it seems like the current state is treated as the source of truth, but I'm unsure whether users want that in every case. I guess the previous implementation did the opposite and treated the attached objects from the pre-planned waypoint as the source of truth.
> 
> Is it worth maybe adding a flag to this function for whether one wants to prioritize current state vs. planned states? And elevate this up to the config/parameter level?

As for prioritizing the current state vs. planned states, since world objects are always considered in real-time rather than at planning time, I don’t see a strong reason to let the user choose the source of truth for attached collision objects. Keeping it consistent with real-time world objects could help prevent many erroneous situations.

moveit_ros/planning/plan_execution/src/plan_execution.cpp
MarcoMagriDev: You are absolutely right! I'll push a new commit to address this along with the 's' styling issue.

moveit_ros/planning/plan_execution/src/plan_execution.cpp
sea-bass: Agreed. But at least in your comments on 1., maybe we do need the toggle on using the attached objects only at the start vs. updating this frequently.

Basically, asking the user "do you want this simplification or not?"

MarcoMagriDev: I just pushed a new commit that automates this process without requiring user input.

Before entering the monitoring phase, it verifies whether the attached objects remain consistent throughout the trajectory. If they do, they are stored and later used by the `isRemainingPathValid` method without needing to be queried again.

If the attached objects change during the planned trajectory, the map is left empty, signaling `isRemainingPathValid` to query them at each waypoint.

I believe this approach is more robust than using a parameter, as it eliminates the possibility of misconfiguration by the user. What do you think?

moveit_ros/planning/plan_execution/src/plan_execution.cpp
```diff
@@ -430,6 +460,29 @@ moveit_msgs::msg::MoveItErrorCodes plan_execution::PlanExecution::executeAndMoni
   path_became_invalid_ = false;
   bool preempt_requested = false;
 
+  // Check that attached objects remain consistent throughout the trajectory and store them.
+  // This avoids querying the scene for attached objects at each waypoint whenever possible.
+  // If a change in attached objects is detected, they will be queried at each waypoint.
+  trajectory_attached_objects_.clear();
```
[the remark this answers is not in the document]
MarcoMagriDev: I have addressed your reviews, but I now have a doubt:

_Can we assume that there is only one `plan_components` with a specified trajectory?_

If not, the current check may not work as expected. Instead of filling `trajectory_attached_objects_`, I propose using `plan_components_attached_objects_`, which would be a vector of trajectory_attached_objects, where the index corresponds to the respective `plan_components`.  

This way, `isRemainingPathValid` can perform the check correctly.  

What do you think?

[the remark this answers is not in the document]
sea-bass: > Can we assume that there is only one plan_components with a specified trajectory?

Almost certainly not. I believe tools like MoveIt Task Constructor package up plans with multiple components for robots to execute.

So I guess your idea may be useful here.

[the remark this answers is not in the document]
MarcoMagriDev: Just added a new commit to address it

MarcoMagriDev:
Hey all, just checking in on this PR. It’s been open for a bit—let me know if I need to update anything to move it forward.

sjahr (approved):

Issue #3292, opened 2025-02-01
PILZ LIN samples in time instead of space


MarcoMagriDev:
### Description

The LIN pipeline of `pilz_industrial_motion_planner` samples IK solutions along the Cartesian path using a fixed sampling time (0.1 sec). Once the IK solutions are obtained, a joint-space interpolation is performed.

Since the Cartesian path is sampled based on time rather than space, the linearity of the resulting path degrades as the commanded motion speed increases.

For example, consider a motion of 1 meter:
- If commanded at 1 m/s, 10 Cartesian samples will be extracted along the path.
- If commanded at 5 m/s, only 2 samples will be extracted.

Since these points are then interpolated in joint space, the first motion will appear more linear than the second due to the higher number of samples along the Cartesian path.

### ROS Distro

Humble

### OS and version

Ubuntu 22.04

### Source or binary build?

Source

### If source, which branch?

humble

### Which RMW are you using?

CycloneDDS

### Steps to Reproduce

Command the same LIN motion at different speed and analyze resulting path linearity

### Expected behavior

The linearity of the generated path should remain independent of the commanded speed.

### Actual behavior

The linearity of the generated path is dependent of the commanded speed.

sea-bass:
It really does that? Weird.

While I think Pilz could definitely be rewritten to be like 10% of the amount of the code it currently is, that would take a while -- especially because of the blending functionality.

My suggestion would be to find a way to elevate that 0.1 seconds hard-coded value to an input parameter, thereby exposing a way to parameterize IK solve resolution by distance.

At least for the LIN profile, that would be an easy calculation based on linear and angular distances. For the CIRC profile, I guess it's also doable since there are formulas for arc lengths based on descriptions of a circle.

MarcoMagriDev:
> It really does that? Weird.
> 

Correct me if I am wrong but the value is hardcoded [here](https://github.com/moveit/moveit2/blob/23022683efc7e16b48a4de7c72e908f8bb38988d/moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/trajectory_generator.hpp#L110) and gets forwarded till [here](https://github.com/moveit/moveit2/blob/23022683efc7e16b48a4de7c72e908f8bb38988d/moveit_planners/pilz_industrial_motion_planner/src/trajectory_functions.cpp#L205) where IKs are performed for each time samples.

[...]

PR #3330, opened 2025-02-08
PILZ adjust cartesian interpolation

MarcoMagriDev:
### Description

#3292 

- [x] Added interpolation parameters  
- [x] Sampled the trajectory using a uniform `sampling_time` computed from the equivalent rectangular motion profile (see [this comment](https://github.com/moveit/moveit2/issues/3292#issuecomment-2629269300))  
- [x] Adjusted `sampling_time` so that all points are equally spaced, including the last one  
- [x] Modified `pilz_industrial_motion_planner::TrajectoryBlenderTransitionWindow` to no longer require `sampling_time` to be equal and uniform across the trajectories to be blended


moveit_planners/pilz_industrial_motion_planner/src/trajectory_functions.cpp
```diff
@@ -217,11 +217,13 @@ bool pilz_industrial_motion_planner::generateJointTrajectory(
...
-  for (double t_sample = 0.0; t_sample < trajectory.Duration() - epsilon; t_sample += sampling_time)
+  int num_samples = std::floor(trajectory.Duration() / sampling_time);
+  sampling_time = trajectory.Duration() / num_samples;
+  time_samples.reserve(num_samples);
+  for (int i = 0; i < num_samples; ++i)
   {
-    time_samples.push_back(t_sample);
+    time_samples.push_back(i * sampling_time);
   }
   time_samples.push_back(trajectory.Duration());
```
rr-mark, line 228: Is it worth making `i` run from `0` to `num_samples` inclusive, then deleting the `push_back(trajectory.Duration())` line?
```
  for (int i = 0; i <= num_samples; ++i)
  {
    time_samples.push_back(i * sampling_time);
  }
  ```

File: moveit_planners/pilz_industrial_motion_planner/src/trajectory_generator_circ.cpp [hunk 2]
```diff
@@ -215,9 +216,15 @@ void TrajectoryGeneratorCIRC::plan(const planning_scene::PlanningSceneConstPtr&
   moveit_msgs::msg::MoveItErrorCodes error_code;
   // sample the Cartesian trajectory and compute joint trajectory using inverse
   // kinematics
+  auto cartesian_limits = planner_limits_.getCartesianLimits();
+  auto st = std::min({ interpolation_params.max_sample_time,
```
rr-mark, line 220: Could this be `sampling_time`?

moveit_planners/pilz_industrial_motion_planner/src/trajectory_generator_lin.cpp
```diff
@@ -162,9 +163,15 @@ void TrajectoryGeneratorLIN::plan(const planning_scene::PlanningSceneConstPtr& s
   moveit_msgs::msg::MoveItErrorCodes error_code;
   // sample the Cartesian trajectory and compute joint trajectory using inverse
   // kinematics
+  auto cartesian_limits = planner_limits_.getCartesianLimits();
+  auto st = std::min({ interpolation_params.max_sample_time,
```
rr-mark, line 167: `sampling_time`

rr-mark (commented):
The approach here all looks sensible. I've left a few code review comments, and I'll try and find the time to look at those trajectory blender tests at some point this week.

MarcoMagriDev:
@rr-mark thanks for your quick review

I will answer to each review separately.

---

Note that I am still working (on another branch) trying to develop the non constant sampling time on the LIN trajectory. See [Problem 2](https://github.com/moveit/moveit2/issues/3292#issuecomment-2633020658).
The solution is not to change how velocity and acceleration accelerations are computed instead sampling the path at non constant sample time in order to ensure that points are distant enought for the numerical IK solver.
I am also adding some new tests for this PR

moveit_planners/pilz_industrial_motion_planner/src/trajectory_blender_transition_window.cpp
```diff
@@ -207,15 +195,74 @@ bool pilz_industrial_motion_planner::TrajectoryBlenderTransitionWindow::validate
...
+    double t1 = trajectory->getWayPointDurationFromStart(i);
+    double t2 = trajectory->getWayPointDurationFromStart(i + 1);
+
+    if (t1 <= time && t2 >= time)
+    {
+      idx_before = i;
+      idx_after = i + 1;
+      break;
+    }
+  }
```
rr-mark, line 215: Since `getWayPointDurationFromStart(i)` should monotonically increase with `i`, it should only be necessary to find the first value of `i` for which `getWayPointDurationFromStart(i) > time`.

For performance, it might also be worth passing in optional `start` and `end` indices to `interpolatePose`, so if we expect to call it multiple times in order we don't need to search the whole trajectory every time.

rr-mark (commented):
Since this PR is large, it might also be worth splitting it up - starting with a PR for just the trajectory blender.

moveit_planners/pilz_industrial_motion_planner/src/trajectory_blender_transition_window.cpp
```diff
@@ -207,15 +195,74 @@ bool pilz_industrial_motion_planner::TrajectoryBlenderTransitionWindow::validate
...
+  double lambda = (time - t1) / (t2 - t1);
+
+  // Linear interpolation for position
+  Eigen::Isometry3d interpolated_pose;
+  interpolated_pose.translation() = pose1.translation() + lambda * (pose2.translation() - pose1.translation());
+
+  // SLERP interpolation for rotation
+  Eigen::Quaterniond quat1(pose1.rotation());
+  Eigen::Quaterniond quat2(pose2.rotation());
+  interpolated_pose.linear() = quat1.slerp(lambda, quat2).toRotationMatrix();
```
rr-mark, line 243: It might be worth pulling out this code into a function to calculate `interpolated_pose = interpolate(pose1, pose2, lambda)`. Looks like this function would also be useful in `cartesian_interpolator.cpp`.

(This function might also already exist as part of KDL).

moveit_planners/pilz_industrial_motion_planner/src/trajectory_blender_transition_window.cpp
```diff
@@ -225,28 +272,42 @@ void pilz_industrial_motion_planner::TrajectoryBlenderTransitionWindow::blendTra
...
+  double blend_duration = d2 + align_time - t_start;
   pilz_industrial_motion_planner::CartesianTrajectoryPoint waypoint;
   blend_sample_pose2 = req.second_trajectory->getFirstWayPoint().getFrameTransform(req.link_name);
 
   // Pose on blending trajectory
   Eigen::Isometry3d blend_sample_pose;
-  for (std::size_t i = 0; i < blend_sample_num; ++i)
+
+  // Define an arbitrary small sample time to sample the blending trajectory
+  double sampling_time = 0.001;
```
rr-mark, line 283: How easy would it be to use your nice interpolation parameters instead of this arbitrary sampling time?

MarcoMagriDev: I like this approach since it also can helps ensure a minimum Cartesian distance similarly to what has been done [here](https://github.com/MarcoMagriDev/moveit2/blob/1af860df7fdacfab3a85c88a4de03b80ab4c087e/moveit_planners/pilz_industrial_motion_planner/src/trajectory_functions.cpp#L205) (which can be beneficial for numerical IK solvers). 

My only concern is that adopting this would make splitting up the PR no longer a valid option.

moveit_planners/pilz_industrial_motion_planner/src/trajectory_blender_transition_window.cpp
MarcoMagriDev: You are right, thanks. I'll check if KDL already provides a function for this. If not, I'll implement a utility method to handle it.

moveit_planners/pilz_industrial_motion_planner/src/trajectory_blender_transition_window.cpp
```diff
@@ -258,9 +319,21 @@ void pilz_industrial_motion_planner::TrajectoryBlenderTransitionWindow::blendTra
     Eigen::Quaterniond end_quat(blend_sample_pose2.rotation());
     blend_sample_pose.linear() = start_quat.slerp(alpha, end_quat).toRotationMatrix();
```
rr-mark, line 320: (The `interpolate` function above would also be useful here: `blend_sample_pose = interpolate(blend_sample_pose1, blend_sample_pose2, alpha)`.

moveit_planners/pilz_industrial_motion_planner/src/trajectory_blender_transition_window.cpp
rr-mark: Yeah, fair. I guess splitting the PR is nice to have but not necessary.

moveit_planners/pilz_industrial_motion_planner/src/trajectory_blender_transition_window.cpp
MarcoMagriDev: Unfortunately, KDL doesn’t seem to provide a built-in way to achieve this.  

Looking at `cartesian_interpolator.cpp`, I found that a common method could be useful [here](https://github.com/moveit/moveit/blob/7bc8da05dfb87ff1bc046d5d3d640eb0572ff0ec/moveit_core/robot_state/src/cartesian_interpolator.cpp#L171). However, I’m not sure if merging the two would be the best approach.

moveit_planners/pilz_industrial_motion_planner/src/trajectory_blender_transition_window.cpp
MarcoMagriDev: yep, we only need to decide which is the best place for that utility method

moveit_planners/pilz_industrial_motion_planner/src/trajectory_blender_transition_window.cpp
MarcoMagriDev: I just pushed a new version of this method incorporating your suggestions.  

Let me know if you spot any issues or areas for improvement!

moveit_planners/pilz_industrial_motion_planner/src/trajectory_blender_transition_window.cpp
MarcoMagriDev: For now, I’ve placed the method in `trajectory_functions`.  

Let’s wait for input from a maintainer to see if they prefer making this method accessible from another dependency package, such as `moveit_core`, so it can be used [here](https://github.com/moveit/moveit/blob/7bc8da05dfb87ff1bc046d5d3d640eb0572ff0ec/moveit_core/robot_state/src/cartesian_interpolator.cpp#L171).

PR #3332, opened 2025-02-10
Add logic to Ros2ControlManager to match ros2_control

pac48:
### Description

The logic for the chained controllers still does not match the expectation of ROS2 control. This PR makes a few changes:
1. Activation/deactivation is expected to be disjoint. For example, if controller B is a dependency of A (A chains to B) but controller B is also a dependency of C (B chains to B), then the switch from A->B to C->B would cause B to be in both the activation and deactivate list. This causes ROS2 control to through an error and reject the switch. The `simplifyControllerActivationDeactivation` function adds the logic needed to avoid this from happening.
2. The valid transitions for chained controllers depend asymmetrically on whether the controller is being started or stopped. If A chained to B, then when A is started, B should be automatically started. B is allowed to be started and stopped on its own if A is not running. However, if both A and B are running, B cannot be shut down alone, both must be shut down together or A alone.
3. Lastly, ROS 2 control should not be told to shut down controllers that are not running or start controllers that are already running. Doing so can cause an error. A check is added to only start controllers that are not already running.

### DIFF
--- a/moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
+++ b/moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
@@ -376,6 +376,11 @@ public:
   static void simplifyControllerActivationDeactivation(std::vector<std::string>& activate_controllers,
                                                        std::vector<std::string>& deactivate_controllers)
   {
+    // Activation/deactivation is expected to be disjoint. For example, if controller B is a dependency of A (A chains
+    // to B) but controller B is also a dependency of C (B chains to B), then the switch from A->B to C->B would cause B
+    // to be in both the activation and deactivate list. This causes ROS2 control to through an error and reject the
+    // switch. The simplifyControllerActivationDeactivation function adds the logic needed to avoid this from happening.
+
     // Convert vectors to sets for uniqueness
     std::unordered_set set1(activate_controllers.begin(), activate_controllers.end());
     std::unordered_set set2(deactivate_controllers.begin(), deactivate_controllers.end());
--- a/moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
+++ b/moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
@@ -88,37 +88,37 @@ std::string parseJointNameFromResource(const std::string& claimed_interface)
  * \detail Activation/deactivation is expected to be disjoint. For example, if controller B is a dependency of A (A
  * chains to B) but controller B is also a dependency of C (B chains to B), then the switch from A->B to C->B would
  * cause B to be in both the activation and deactivate list. This causes ROS2 control to through an error and reject the
- * switch. The simplifyControllerActivationDeactivation function adds the logic needed to avoid this from happening.
+ * switch. The deconflictControllerActivationLists function adds the logic needed to avoid this from happening.
  * @param[in] activate_controllers  controllers to activate
  * @param[in] deactivate_controllers  controllers to deactivate
  */
-void simplifyControllerActivationDeactivation(std::vector<std::string>& activate_controllers,
-                                              std::vector<std::string>& deactivate_controllers)
+void deconflictControllerActivationLists(std::vector<std::string>& activate_controllers,
+                                         std::vector<std::string>& deactivate_controllers)
 {
   // Convert vectors to sets for uniqueness
-  std::unordered_set set1(activate_controllers.begin(), activate_controllers.end());
-  std::unordered_set set2(deactivate_controllers.begin(), deactivate_controllers.end());
+  std::unordered_set activate_set(activate_controllers.begin(), activate_controllers.end());
+  std::unordered_set deactivate_set(deactivate_controllers.begin(), deactivate_controllers.end());
 
   // Find common elements
-  std::unordered_set<std::string> common;
-  for (const auto& str : set1)
+  std::unordered_set<std::string> common_controllers;
+  for (const auto& str : activate_set)
   {
-    if (set2.count(str))
+    if (deactivate_set.count(str))
     {
-      common.insert(str);
+      common_controllers.insert(str);
     }
   }
 
   // Remove common elements from both sets
-  for (const auto& str : common)
+  for (const auto& controller_name : common_controllers)
   {
-    set1.erase(str);
-    set2.erase(str);
+    activate_set.erase(controller_name);
+    deactivate_set.erase(controller_name);
   }
 
   // Convert sets back to vectors
-  activate_controllers.assign(set1.begin(), set1.end());
-  deactivate_controllers.assign(set2.begin(), set2.end());
+  activate_controllers.assign(activate_set.begin(), activate_set.end());
+  deactivate_controllers.assign(deactivate_set.begin(), deactivate_set.end());
 }
 
 MOVEIT_CLASS_FORWARD(Ros2ControlManager);  // Defines Ros2ControlManagerPtr, ConstPtr, WeakPtr... etc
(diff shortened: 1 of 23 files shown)


sjahr (commented):
How can this be tested / can we add very basic unittesting to make this code more robust (not blocking this PR)?

The test failures are cause by flacky unittests.

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -370,35 +373,91 @@ class Ros2ControlManager : public moveit_controller_manager::MoveItControllerMan
     return c;
   }
 
+  static void simplifyControllerActivationDeactivation(std::vector<std::string>& activate_controllers,
```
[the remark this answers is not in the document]
pac48: Added docs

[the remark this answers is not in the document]
griswaldbrooks: do you want to call this something like `removeCommon` or something to be more descriptive?

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -370,35 +373,96 @@ class Ros2ControlManager : public moveit_controller_manager::MoveItControllerMan
...
+                                                       std::vector<std::string>& deactivate_controllers)
+  {
+    // Activation/deactivation is expected to be disjoint. For example, if controller B is a dependency of A (A chains
+    // to B) but controller B is also a dependency of C (B chains to B), then the switch from A->B to C->B would cause B
+    // to be in both the activation and deactivate list. This causes ROS2 control to through an error and reject the
+    // switch. The simplifyControllerActivationDeactivation function adds the logic needed to avoid this from happening.
```
griswaldbrooks, line 382: The rest of this file seems to put this in the doxygen for the function

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -370,35 +373,96 @@ class Ros2ControlManager : public moveit_controller_manager::MoveItControllerMan
     return c;
   }
 
+  static void simplifyControllerActivationDeactivation(std::vector<std::string>& activate_controllers,
```
griswaldbrooks, line 376: Why is this a static method and not just a function?

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -370,35 +373,96 @@ class Ros2ControlManager : public moveit_controller_manager::MoveItControllerMan
...
                          const std::vector<std::string>& deactivate_base) override
   {
-    // add controller dependencies
-    std::vector<std::string> activate = activate_base;
-    std::vector<std::string> deactivate = deactivate_base;
-    for (auto controllers : { &activate, &deactivate })
-    {
-      auto queue = *controllers;
+    // add_all_dependencies traverses the provided dependency map and appends the results to the controllers vector.
+    auto add_all_dependencies = [](const std::unordered_map<std::string, std::vector<std::string>>& dependencies,
```
griswaldbrooks, line 421: Curious as to why this is a lambda and not a function?

pac48: I am using a lambda because I want to reuse this code but only in the scope of this method.

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
pac48: Added doxygen

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
pac48: Moved to function

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
griswaldbrooks: I apparently cannot resolve my own comments

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -373,32 +414,59 @@ class Ros2ControlManager : public moveit_controller_manager::MoveItControllerMan
...
+      auto queue = controllers;
       while (!queue.empty())
       {
         auto controller = queue.back();
-        controller.erase(0, 1);
+        if (controller.size() > 1 && controller[0] == '/')
+        {
+          // Remove leading / from controller name
+          controller.erase(0, 1);
+        }
```
griswaldbrooks, line 436: can you make this
```suggestion
        auto controller = queue.back();
        queue.pop_back();
        if (controller.size() > 1 && controller[0] == '/')
        {
          // Remove leading / from controller name
          controller.erase(0, 1);
        }
```
splitting them added noise to this which took me a minute to resolve

griswaldbrooks (changes requested):
Ok, done with comments about C++. I'm not qualified to review the functional aspects of this code.

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
pac48: Added this manually

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -83,6 +83,44 @@ std::string parseJointNameFromResource(const std::string& claimed_interface)
...
+/**
+ * \brief  Modified Activation/deactivation to conform to ROS 2 control expectations.
+ * \detail Activation/deactivation is expected to be disjoint. For example, if controller B is a dependency of A (A
+ * chains to B) but controller B is also a dependency of C (B chains to B), then the switch from A->B to C->B would
+ * cause B to be in both the activation and deactivate list. This causes ROS2 control to through an error and reject the
+ * switch. The simplifyControllerActivationDeactivation function adds the logic needed to avoid this from happening.
+ * @param[in] activate_controllers  controllers to activate
+ * @param[in] deactivate_controllers  controllers to deactivate
+ */
+void simplifyControllerActivationDeactivation(std::vector<std::string>& activate_controllers,
```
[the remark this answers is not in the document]
pac48: changed to deconflictControllerActivationLists

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -83,6 +83,44 @@ std::string parseJointNameFromResource(const std::string& claimed_interface)
...
+                                              std::vector<std::string>& deactivate_controllers)
+{
+  // Convert vectors to sets for uniqueness
+  std::unordered_set set1(activate_controllers.begin(), activate_controllers.end());
+  std::unordered_set set2(deactivate_controllers.begin(), deactivate_controllers.end());
```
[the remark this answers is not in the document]
pac48: Added suggestion

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -83,6 +83,44 @@ std::string parseJointNameFromResource(const std::string& claimed_interface)
...
+
+  // Find common elements
+  std::unordered_set<std::string> common;
+  for (const auto& str : set1)
```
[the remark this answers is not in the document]
pac48: Added

sea-bass (approved):
LGTM -- I'll wait until this one is merged before releasing Jazzy/Rolling (sadly I already released Humble like a day or 2 ago)

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
sea-bass: @mikeferguson with the lightning reflexes

sea-bass:
Huh, somehow there are clang-tidy fixes being suggested

sjahr (approved):

sea-bass (approved):

sea-bass:
@griswaldbrooks you may resolve your own comments now 🤣

griswaldbrooks:
How so?

sea-bass:
> How so?

Damn, you got double sniped. I'm sorry :(

PR #3335, opened 2025-02-11
Enable allowed_execution_duration_scaling and allowed_goal_duration_margin for each controller

DaniGarciaLopez:
### Description

Re-enabling the functionality to set `allowed_execution_duration_scaling` and `allowed_goal_duration_margin` for each controller, which has been commented out since moveit2 was created.

### Example of usage
In `moveit_controllers.yaml`:
```
trajectory_execution:
  allowed_execution_duration_scaling: 1.2 # Default global value for all controllers
  allowed_goal_duration_margin: 0.5 # Default global value for all controllers

...

moveit_simple_controller_manager:
  controller_names:
    - panda_arm_controller

  panda_arm_controller:
    type: FollowJointTrajectory
    allowed_execution_duration_scaling: 5.0 # Overrides the global value for this controller
    allowed_goal_duration_margin: 1.0 # Overrides the global value for this controller
    ...
```

### Checklist

### DIFF
--- a/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
+++ b/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
@@ -106,9 +106,6 @@ void TrajectoryExecutionManager::initialize()
   allowed_execution_duration_scaling_ = DEFAULT_CONTROLLER_GOAL_DURATION_SCALING;
   allowed_goal_duration_margin_ = DEFAULT_CONTROLLER_GOAL_DURATION_MARGIN;
 
-  // load controller-specific values for allowed_execution_duration_scaling and allowed_goal_duration_margin
-  loadControllerParams();
-
   // load the controller manager plugin
   try
   {
@@ -191,6 +188,10 @@ void TrajectoryExecutionManager::initialize()
 
   // other configuration steps
   reloadControllerInformation();
+
+  // load controller-specific values for allowed_execution_duration_scaling and allowed_goal_duration_margin
+  loadControllerParams();
+
   // The default callback group for rclcpp::Node is MutuallyExclusive which means we cannot call
   // receiveEvent while processing a different callback. To fix this we create a new callback group (the type is not
   // important since we only use it to process one callback) and associate event_topic_subscriber_ with this callback group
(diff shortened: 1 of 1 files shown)


moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -106,9 +106,6 @@ void TrajectoryExecutionManager::initialize()
   allowed_execution_duration_scaling_ = DEFAULT_CONTROLLER_GOAL_DURATION_SCALING;
   allowed_goal_duration_margin_ = DEFAULT_CONTROLLER_GOAL_DURATION_MARGIN;
 
-  // load controller-specific values for allowed_execution_duration_scaling and allowed_goal_duration_margin
-  loadControllerParams();
```
sjahr, line 110: Can you briefly explain why this is necessary?

sjahr (commented):
Thanks this looks good! Have you tested if it works and the margins are propagated correctly to the controller?

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -1841,25 +1842,24 @@ bool TrajectoryExecutionManager::ensureActiveControllers(const std::vector<std::
...
-  //       if (controller.hasMember("allowed_goal_duration_margin"))
-  //         controller_allowed_goal_duration_margin_[std::string(controller["name"])] =
-  //             controller["allowed_goal_duration_margin"];
-  //     }
-  //   }
-  // }
+  for (const auto& controller : known_controllers_)
+  {
+    const std::string& controller_name = controller.first;
+    for (const auto& controller_manager_name : controller_manager_loader_->getDeclaredClasses())
```
DaniGarciaLopez, line 1848: I'm not sure if this is actually needed to account for other plugins, like `moveit_ros_control_interface`. Maybe we could just safely simplify it hardcoding `moveit_simple_controller_manager`.

DaniGarciaLopez:
> Thanks this looks good! Have you tested if it works and the margins are propagated correctly to the controller?

Thanks for reviewing, @sjahr! As far as I could test, everything works as expected. However, I only tested using `moveit_simple_controller_manager`. I'm not sure if this would be needed for other plugins, as I don't use them in my configs. I've added one comment on the line which we could simplify if it's not needed.

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
sjahr: Ok thanks!

sjahr (approved):

sea-bass (approved):

PR #3341, opened 2025-02-12
Update current state even if servo is paused

pac48:
### Description

Servo was not updating the current robot state when it was paused. This can result in massive jumps when servo is paused and the robot is moved with another mechanism, then servo is reactivated.

### DIFF
--- a/moveit_ros/moveit_servo/src/servo_node.cpp
+++ b/moveit_ros/moveit_servo/src/servo_node.cpp
@@ -155,6 +155,7 @@ void ServoNode::pauseServo(const std::shared_ptr<std_srvs::srv::SetBool::Request
     response->message = message;
     return;
   }
+  std::lock_guard<std::mutex> lock_guard(lock_);
   servo_paused_ = request->data;
   response->success = (servo_paused_ == request->data);
   if (servo_paused_)
@@ -164,7 +165,6 @@ void ServoNode::pauseServo(const std::shared_ptr<std_srvs::srv::SetBool::Request
   }
   else
   {
-    std::lock_guard<std::mutex> lock_guard(lock_);
     // Reset the smoothing plugin with the robot's current state in case the robot moved between pausing and unpausing.
     last_commanded_state_ = servo_->getCurrentRobotState(true /* block for current robot state */);
     servo_->resetSmoothing(last_commanded_state_);


sea-bass (approved):

sea-bass:
Why can't this thing stay unbroken? Lol

moveit_ros/moveit_servo/src/servo_node.cpp
```diff
@@ -329,6 +329,7 @@ void ServoNode::servoLoop()
     // Skip processing if servoing is disabled.
     if (servo_paused_)
     {
+      current_state = servo_->getCurrentRobotState(true);
```
sea-bass, line 332: Actually, can you make this non-blocking by passing in `false`?

pac48: I think we should block actually. Otherwise, we run the risk of having an old robot state when servo resumes. Is there some reason why non-blocking would be better in this case?

sea-bass: Yes, in the main loop we should try not blocking.

However, the service callback that handles unpausing servo does have a blocking call for that reason, I think.

https://github.com/moveit/moveit2/blob/7504a2808f7c76d2e707f090340524df893d19f0/moveit_ros/moveit_servo/src/servo_node.cpp#L169

It just doesn't also update the `current_state_` data member. So maybe your fix should go here instead?

pac48: To be fair, this only blocks the main loop when it is explicitly disabled. I actually think we can remove this altogether. I think this issue here was due to 
```
  servo_paused_ = request->data;
  response->success = (servo_paused_ == request->data);
  if (servo_paused_)
  {
    servo_->setCollisionChecking(false);
    response->message = "Servoing disabled";
  }
  else
  {
    std::lock_guard<std::mutex> lock_guard(lock_);
    // Reset the smoothing plugin with the robot's current state in case the robot moved between pausing and unpausing.
    last_commanded_state_ = servo_->getCurrentRobotState(true /* block for current robot state */);
```

`servo_paused_` becomes unpaused but and the main loop grabs the mutex, this code gets stuck at the `lock_guard`. Then in the main loop, reaches `current_state = joint_cmd_rolling_window_.back();` and gets an outdated value. So I think moving the mutex above the assignment of `servo_paused_` would work.

sea-bass (approved):
I see. Thanks for looking into this again. If this resolves the issue, looks reasonable to me!

sjahr (approved):

PR #3351, opened 2025-02-15
Fix Servo JointJog Crash

mjforan:
### Description

Currently the realtime servo node crashes if a JointJog message is received without velocity commands. I added a check for this condition and a warning if the message contains displacement commands.

### Checklist
- [x] Create tests, which fail without this PR


moveit_ros/moveit_servo/tests/test_ros_integration.cpp
```diff
@@ -103,6 +103,32 @@ TEST_F(ServoRosFixture, testJointJog)
 
   ASSERT_GE(traj_count_, NUM_COMMANDS);
 
+  // Ensure deprecation warning is reported for displacements command
+  jog_cmd.displacements.push_back(1.0);
+  jog_cmd.header.stamp = servo_test_node_->now();
+  joint_jog_publisher->publish(jog_cmd);
+  rclcpp::sleep_for(std::chrono::milliseconds(100));
+
+  // TODO how to test logged warnings?
```
sea-bass, line 112: Not very easily unless you go down the whole dependency injection path... I would say this is quite an undertaking.

sea-bass (commented):
Thanks for finding this!

I think by putting all these checks at the ROS node level, it may be making servo even harder to test than it already is.

If I may offer a suggestion, a few lines down where you modified things there is a line that actually creates a `JointJogCommand` struct:

```cpp
JointJogCommand command{ latest_joint_jog_.joint_names, latest_joint_jog_.velocities };
```

As you have correctly noticed, this is just a plain struct and has no validation on the joint name vs. velocity inputs. I think what would make this stuff much more sustainable to test would be to create free functions that convert directly from the ROS `JointJog` message to the servo specific `JointJogCommand` type.

And these functions can have their own validation / failure handling with things like `std::optional` or `tl::expected`.

If you'd be up for giving this a try, it would definitely help to take some of the logic in the `servo_node.cpp` and encapsulate it a little better.

But I'll defer to you on how far you want to go here vs. just fix a bug and move on :)

moveit_ros/moveit_servo/tests/test_ros_integration.cpp
mjforan: Got this working by subscribing to logs on `/rosout`.

mjforan:
@sea-bass what do you think of this approach? Instead of adding extra functions on the struct initialization, I moved the check to the section where velocities are actually processed in `command.cpp`. This also protects against the case where a C++ user creates a `JointJogCommand` and alters the vector size afterwards.

moveit_ros/moveit_servo/tests/test_ros_integration.cpp
```diff
@@ -106,6 +106,43 @@ TEST_F(ServoRosFixture, testJointJog)
   moveit_servo::StatusCode status = status_;
   RCLCPP_INFO_STREAM(servo_test_node_->get_logger(), "Status after jointjog: " << static_cast<size_t>(status));
   ASSERT_EQ(status_, moveit_servo::StatusCode::NO_WARNING);
+
+  // Ensure deprecation warning is reported for displacements command
```
sea-bass, line 110: Not a deprecation warning, right? Just a regular one?

moveit_ros/moveit_servo/tests/servo_ros_fixture.hpp
```diff
@@ -145,6 +155,20 @@ class ServoRosFixture : public testing::Test
     }
   }
 
+  bool logContains(std::string logger_name, std::string msg)
```
sea-bass, line 158: Use `const std::string&` for both args

moveit_ros/moveit_servo/src/utils/command.cpp
```diff
@@ -122,6 +129,13 @@ JointDeltaResult jointDeltaFromJointJog(const JointJogCommand& command, const mo
   else
   {
     status = StatusCode::INVALID;
+    if (!command_valid)
+    {
+      RCLCPP_WARN_STREAM(getLogger(), "Invalid joint jog command. Each joint name must have one corresponding "
```
sea-bass, line 134: Should this be an ERROR instead?

moveit_ros/moveit_servo/src/utils/command.cpp
```diff
@@ -94,24 +94,31 @@ JointDeltaResult jointDeltaFromJointJog(const JointJogCommand& command, const mo
 
   velocities.setZero();
   bool names_valid = true;
-
-  for (size_t i = 0; i < command.names.size(); ++i)
+  bool command_valid = true;
```
sea-bass, line 97: Do we need both `command_valid` and `names_valid` as separate variables? Or can they both set the same variable to false?

moveit_ros/moveit_servo/src/utils/command.cpp
```diff
@@ -94,24 +94,31 @@ JointDeltaResult jointDeltaFromJointJog(const JointJogCommand& command, const mo
...
+  if (command.velocities.size() == command.names.size())
```
sea-bass, line 98: Might be better to avoid nesting to check whether the names/velocities are different size first, and return early if that's the case

moveit_ros/moveit_servo/src/utils/command.cpp
sea-bass: Also again, if you move these warnings to the place they happen when you actually set `command/names_valid`, you don't need more if statements here

sea-bass (commented):
This definitely is better! Still 2 comments though

1. I think the amount of conditionals/nesting can still be simplified significantly
2. I'm a bit worried about relying on a subscription to the `/rosout` topic for tests to pass, and may lead to (even more) flaky tests. Have you tried to use the `stress` command to stress your system and run this new test repeatedly?

moveit_ros/moveit_servo/tests/test_ros_integration.cpp
mjforan: According to the documentation, displacement commands were previously available in MoveIt. So I would say it counts as deprecation. Or should I change the message to "not yet supported"?

moveit_ros/moveit_servo/include/moveit_servo/utils/command.hpp
```diff
@@ -47,6 +47,10 @@
 #include <rclcpp/rclcpp.hpp>
 #include <tf2_eigen/tf2_eigen.hpp>
 
+#define COMMAND_ERROR(msg_stream)                                                                                      \
```
sea-bass, line 50: Would strongly advise against macros since it makes things hard to debug.

But just put the error + return lines in the code itself and it looks good!

mjforan:
>  I'm a bit worried about relying on a subscription to the `/rosout` topic for tests to pass, and may lead to (even more) flaky tests. Have you tried to use the `stress` command to stress your system and run this new test repeatedly?

I tried running with `stress` multiple times and saw no failures. The tests were already relying on a ROS topic (publishing commands) so I don't think my change significantly increases the flakiness.

sea-bass (commented):
Looks good -- thanks! Just one last comment from me.

Now that I see this all again, would it make sense to put the warning about displacements not being supported in `jointDeltaFromJointJog()` instead of in the ROS node itself?

mjforan:
The `JointJogCommand` struct does not have a displacements field so `command.cpp` is unaware of such input. If you think displacement functionality will be restored, we can add the struct field and move the warning.

mjforan:
You were right, CI is failing on the /rosout log check. Maybe I should add a longer delay before that check? Or is there something else unusual about the environment? So far I have only tested on Raspberry Pi 5 in Jazzy Docker.

sea-bass:
> You were right, CI is failing on the /rosout log check. Maybe I should add a longer delay before that check? Or is there something else unusual about the environment? So far I have only tested on Raspberry Pi 5 in Jazzy Docker.

Hah, never underestimate how crappy CI runners can be. Honestly, if it were up to me I'd just verify the value of the status coming out of the function and not worry about the logs.

Also, understood re: the displacement warning. 👍🏻

sea-bass (approved):

sea-bass:
> @sea-bass ready to merge?

Sorry, I thought I had asked other active maintainers for a 2nd approval. Tagged a few.

sea-bass (approved):
Thank you!

mjforan:
Thanks. Frustrating that the CI has different formatting rules from the colcon tests.

PR #3367, opened 2025-02-27
feat: add remapping argument to MoveItPy initialization

KazuyaOguma18:
### Description
Previously, the topics started by MoveItPy could not be remapped.
To address this, I added an argument for remapping, allowing remapping to be handled within Python scripts.

Ideally, I wanted MoveItPy to support remapping via the remappings option specified in the launch file. However, due to the design philosophy of rclpy, this seems difficult. As a compromise, I have ensured that remapping can at least be performed via an argument.

### Checklist

### DIFF
--- a/moveit_py/src/moveit/moveit_ros/moveit_cpp/moveit_cpp.cpp
+++ b/moveit_py/src/moveit/moveit_ros/moveit_cpp/moveit_cpp.cpp
@@ -64,7 +64,8 @@ void initMoveitPy(py::module& m)
 
       .def(py::init([](const std::string& node_name, const std::string& name_space,
                        const std::vector<std::string>& launch_params_filepaths, const py::object& config_dict,
-                       bool provide_planning_service, const std::optional<std::map<std::string, std::string>>& remappings) {
+                       bool provide_planning_service,
+                       const std::optional<std::map<std::string, std::string>>& remappings) {
              // This section is used to load the appropriate node parameters before spinning a moveit_cpp instance
              // Priority is given to parameters supplied directly via a config_dict, followed by launch parameters
              // and finally no supplied parameters.
@@ -89,10 +90,10 @@ void initMoveitPy(py::module& m)
 
              if (remappings.has_value())
              {
-               for (const auto &[key, value]: *remappings)
+               for (const auto& [key, value] : *remappings)
                {
-                  launch_arguments.push_back("--remap");
-                  launch_arguments.push_back(key + ":=" + value);
+                 launch_arguments.push_back("--remap");
+                 launch_arguments.push_back(key + ":=" + value);
                }
              }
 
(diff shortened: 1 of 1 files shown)


KazuyaOguma18:
Hi @sea-bass @henningkayser @peterdavidfagan ,  
I'm having trouble figuring out what's causing the CI failure.  
I've looked into it, but I haven't been able to pinpoint the issue.  

Would you be able to take a look when you have some time?  
Any insights would be greatly appreciated.  
Thanks!

moveit_py/src/moveit/moveit_ros/moveit_cpp/moveit_cpp.cpp
```diff
@@ -62,7 +64,7 @@ void initMoveitPy(py::module& m)
 
       .def(py::init([](const std::string& node_name, const std::string& name_space,
                        const std::vector<std::string>& launch_params_filepaths, const py::object& config_dict,
-                       bool provide_planning_service) {
+                       bool provide_planning_service, const py::object& remappings) {
```
rhaschke, line 67: Just use a std::map as argument here. This simplifies casting code below.
Wrapping the map into `std::optional` [allows to handle None values](https://github.com/pybind/pybind11/issues/1953#issuecomment-683409703).
```suggestion
                       bool provide_planning_service, const std::optional<std::map<std::string, std::string>>& remappings) {
```

moveit_py/src/moveit/moveit_ros/moveit_cpp/moveit_cpp.cpp
```diff
@@ -85,6 +87,21 @@ void initMoveitPy(py::module& m)
...
+                 if (item_tuple.size() != 2)
+                 {
+                   throw std::runtime_error("Remapping must be a list of tuples with 2 elements");
+                 }
+                 launch_arguments.push_back("--remap");
+                 launch_arguments.push_back(item_tuple[0].cast<std::string>() +
+                                            ":=" + item_tuple[1].cast<std::string>());
+               }
+             }
+
```
rhaschke, line 104:
```suggestion
             if (remappings.has_value())
             {
               for (const auto &[key, value]: remappings)
               {
                  launch_arguments.push_back("--remap");
                  launch_arguments.push_back(key + ":=" + value);
               }
             }
```

rhaschke (commented):
> Ideally, I wanted MoveItPy to support remapping via the remappings option specified in the launch file. However, due to the design philosophy of rclpy, this seems difficult.

Can you please elaborate on this? What's the difficulty?
I see this more as a design flaw of the `MoveItPy` constructor.
[`py_binding_tools`](https://github.com/moveit/py_binding_tools/blob/ros2/src/rclcpp.cpp) allows any arguments (including remapping args).
I always wanted to rewrite `MoveItPy` to use `py_binding_tools`, but never found time for it...

rhaschke (approved):

sea-bass (approved):

PR #3382, opened 2025-03-15
SERVICE_CALL_TIMEOUT = 1 second is harsh 🥵

ymollard:
### Description

For some reason, maybe over loaded network or CPU, listing the ros2 controllers of my robot is quite slow, a bit more than 1.0 second in average:

```bash
$ time ros2 control list_controllers
arm_controller          joint_trajectory_controller/JointTrajectoryController         active  
[...]
real    0m1.349s
user    0m0.658s
sys     0m0.144s
```
In consequence, the MoveIt ROS interface fails at executing the trajectory while the controller is actually ready to work.

I propose to increase the 1.0 second timeout to 10.0 seconds because 1.0 is really harsh, especially on slow computers.

Maybe a lower timeout would be more acceptable, as you prefer.

### DIFF
--- a/moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
+++ b/moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
@@ -57,7 +57,7 @@
 #include <moveit/utils/logger.hpp>
 
 static const rclcpp::Duration CONTROLLER_INFORMATION_VALIDITY_AGE = rclcpp::Duration::from_seconds(1.0);
-static const double SERVICE_CALL_TIMEOUT = 10.0;
+static const double SERVICE_CALL_TIMEOUT = 3.0;  // TODO: Create a ROS parameter allowing to customize this default timeout
 
 namespace moveit_ros_control_interface
 {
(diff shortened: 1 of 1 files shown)


sea-bass (approved):
I would say "why not?"

Even better (if you're up for it), since this manager contains a [ROS node](https://github.com/moveit/moveit2/blob/cdd577b1b13a0a425c3006056dd1494d6add6e58/moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp#L156), you could consider making this a ROS parameter so that users have the ability to control their timeouts. I can imagine in some cases if the system is this slow, it would be preferable to cause failures.

ymollard:
> if you're up for it

Not now ;) Can we maybe keep track of it in a CFC list somewhere?

sea-bass:
> Not now ;) Can we maybe keep track of it in a CFC list somewhere?

Yep, can just make a new Git issue for that.

In that case, would you mind reducing the timeout in this PR from 10 seconds to like 2 or 3, then? And just adding a TODO comment in the code for now?

ymollard:
Done @sea-bass

sea-bass:
Created https://github.com/moveit/moveit2/issues/3405 to track the parameter thing...

Issue #3405, opened 2025-03-30
Change hard-coded timeout in controller manager plugin to a ROS parameter


sea-bass:
### Is your feature request related to a problem? Please describe.

As documented in https://github.com/moveit/moveit2/pull/3382#pullrequestreview-2688278559, the `SERVICE_CALL_TIMEOUT` constant can be instead changed to a ROS parameter so users can tune the value for different applications.

### Describe the solution you'd like

Changing the constant to a parameter.

ashwinsnambiar:
Hi, can I start working on this?

sea-bass:
No need to ask for permission when it comes to OSS -- if you put up a PR, it will be reviewed and appreciated!

DhananjayKhandelwal:
Hello @sea-bass! Can you please guide me to a repository that includes a python launch file that uses the Ros2ControlManager plugin, so I can test the changes? Or something else that I can do, any tutorial to use the plugin, maybe. I've been unable to find much. Thanks!

sea-bass:
I'm not very familiar with this code, so I may not be the right person to ask in this case.

ashwinsnambiar:
I tested the implementation using the https://github.com/moveit/moveit_resources/tree/ros2/dual_arm_panda_moveit_config example moveit configuration by trying out the following values for the parameter in the `moveit_controllers.yaml`:

`controller_service_call_timeout: 0.00001` : Controllers did not connect
`controller_service_call_timeout: 1.0` : Controllers connected and path planning could be executed

Hope this is fine?

ashwinsnambiar:
I have made few additions in the moveit2_tutorial repo too. Please let me know if it is necessary. If so, any corrections to be made?
https://github.com/moveit/moveit2_tutorials/compare/main...ashwinsnambiar:moveit2_tutorials:main

mikeferguson:
Closed by #3419 (back ports are still in process, but I think it's safe to close this)

PR #3419, opened 2025-04-07
ROS Parameter for service call timeout for ros_control controllers

ashwinsnambiar:
### Description

Added a new ROS parameter controller_service_call_timeout for the ROS control interface managers, which can be specified in the moveit_controller yaml configuration file. I have tested it out using the https://github.com/moveit/moveit_resources/tree/ros2/dual_arm_panda_moveit_config example moveit configuration by trying out the following values for the parameter in the `moveit_controllers.yaml`:

`controller_service_call_timeout: 0.00001` :     Controllers did not connect
`controller_service_call_timeout: 1.0`         :     Controllers  connected and path planning could be executed

Closes issue #3405 

### Checklist

### DIFF
--- a/moveit_plugins/moveit_ros_control_interface/README.md
+++ b/moveit_plugins/moveit_ros_control_interface/README.md
@@ -22,10 +22,11 @@ In your MoveIt launch file (e.g. `ROBOT_moveit_config/launch/ROBOT_moveit_contro
 <param name="moveit_controller_manager" value="moveit_ros_control_interface::Ros2ControlManager" />
 ```
 
-Make sure to set the `ros_control_namespace` parameter to the namespace (without the /controller_manager/ part) of the ros_control-based node you like to interface.
+Make sure to set the `ros_control_namespace` parameter to the namespace (without the /controller_manager/ part) of the ros_control-based node you like to interface. Also, the timeout in seconds for the service call to the controller can be changed from the default value of 3, by setting the `controller_service_call_timeout` parameter. 
 If you are using `moveit_setup_assistant`, you can add it to `ROBOT_moveit_config/config/moveit_controllers.yaml`, e.g.:
 ```
 ros_control_namespace: /ROS_CONTROL_NODE
+controller_service_call_timeout: 5.0
 controller_list:
   - name: /ROS_CONTROL_NODE/position_trajectory_controller
     action_ns: follow_joint_trajectory
--- a/moveit_plugins/moveit_ros_control_interface/README.md
+++ b/moveit_plugins/moveit_ros_control_interface/README.md
@@ -23,7 +23,7 @@ In your MoveIt launch file (e.g. `ROBOT_moveit_config/launch/ROBOT_moveit_contro
 ```
 
 Make sure to set the `ros_control_namespace` parameter to the namespace (without the /controller_manager/ part) of the ros_control-based node you like to interface. 
-Also, the timeout in seconds for the service call to the controller can be changed from the default value of 3, by setting the `controller_service_call_timeout` parameter. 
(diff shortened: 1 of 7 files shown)


moveit_plugins/moveit_ros_control_interface/README.md
```diff
@@ -22,10 +22,11 @@ In your MoveIt launch file (e.g. `ROBOT_moveit_config/launch/ROBOT_moveit_contro
 <param name="moveit_controller_manager" value="moveit_ros_control_interface::Ros2ControlManager" />
 ```
 
-Make sure to set the `ros_control_namespace` parameter to the namespace (without the /controller_manager/ part) of the ros_control-based node you like to interface.
+Make sure to set the `ros_control_namespace` parameter to the namespace (without the /controller_manager/ part) of the ros_control-based node you like to interface. Also, the timeout in seconds for the service call to the controller can be changed from the default value of 3, by setting the `controller_service_call_timeout` parameter.
```
sea-bass, line 25: Can you split sentences into separate lines?

Also, would suggest not mentioning the actual default value of 3 in the event that the default changes and the docs go out of sync.

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -318,6 +317,16 @@ class Ros2ControlManager : public moveit_controller_manager::MoveItControllerMan
       }
     }
 
+    if (!node_->has_parameter("controller_service_call_timeout"))
+    {
+      service_call_timeout_ = node_->declare_parameter<double>("controller_service_call_timeout", 3.0);
```
sea-bass, line 322: This default should still be a named `constexpr` as before

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -133,6 +131,7 @@ MOVEIT_CLASS_FORWARD(Ros2ControlManager);  // Defines Ros2ControlManagerPtr, Con
 class Ros2ControlManager : public moveit_controller_manager::MoveItControllerManager
 {
   std::string ns_;
+  double service_call_timeout_;
```
sea-bass, line 134: Does it make sense to store this as an actual chrono duration so you don't need to convert it from double every time the service is called?

sea-bass (commented):
Thank you! See some minor comments, in addition to the formatter failures.

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
ashwinsnambiar: Yeah, that makes sense. It won't have to do the type conversion when checking if timeout has been exceeded everytime.

moveit_plugins/moveit_ros_control_interface/README.md
ashwinsnambiar: Done

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
ashwinsnambiar: Done

sea-bass (approved):

sea-bass:
Still some minor formatting issues it seems, but LGTM otherwise

sea-bass (approved):

ashwinsnambiar:
Hi @sea-bass, since the PR is approved and checks are passing, would you be able to merge it? Let me know if there's anything I should change or do further. Also, I had made few changes in the tutorials [here](https://github.com/moveit/moveit2_tutorials/compare/main...ashwinsnambiar:moveit2_tutorials:main).... Is that required?

mikeferguson (approved):
LGTM

PR #3426, opened 2025-04-10
Adds multi (joint) state validation service

riv-ljames:
### Description

The normal state validation service is not very fast if you want to check a lot of joint states. This is significantly faster.

Requires https://github.com/moveit/moveit_msgs/pull/186

Please could this be backported to humble

### Checklist


moveit_ros/move_group/src/default_capabilities/multi_state_validation_service_capability.cpp
```diff
@@ -0,0 +1,110 @@
...
+    const std::shared_ptr<moveit_msgs::srv::GetMultiStateValidity::Response>& res)
+{
+    res->valid = true;
+    planning_scene_monitor::LockedPlanningSceneRO ls(context_->planning_scene_monitor_);
+    
+    auto robot_state = req->robot_state;
+    moveit::core::RobotState rs = ls->getCurrentState();
+    moveit::core::robotStateMsgToRobotState(robot_state, rs);
+
+    for(int i=0; i<req->joint_states.size(); i++){
```
sea-bass, line 36: Use `++i` as in other loops

moveit_ros/move_group/src/default_capabilities/multi_state_validation_service_capability.cpp
```diff
@@ -0,0 +1,110 @@
...
+            // copy constraint results
+            res->constraint_result.resize(kres.size());
+            for (std::size_t k = 0; k < kres.size(); ++k)
+            {
+                res->constraint_result[k].result = kres[k].satisfied;
+                res->constraint_result[k].distance = kres[k].distance;
+            }
+        }
+
+        // No need to check any more joint states after the first invalid one
```
sea-bass, line 98: Not necessarily, depends on whether users want speed or a complete list of the states that were invalid.

You could consider adding an "early stop" boolean in the service request?

sea-bass (commented):
Looks great!

One question here: I assume a good bit of this code is copy pasted from the single state validation service, but done in a loop to prevent calling multiple services. Is there an opportunity to factor out the common code into a reusable utility function?

moveit_ros/move_group/src/default_capabilities/multi_state_validation_service_capability.cpp
riv-ljames: done

moveit_ros/move_group/src/default_capabilities/multi_state_validation_service_capability.cpp
riv-ljames: We currently dont have a use case for getting a complete list of invalid states.

The significant speedup of being able to check long series of joint states alone is the purpose of this service, and in our case knowing only the information about the first invalid state is ideal.

I cant quite think of a great use case for receiving a list of every invalid state. Maybe this can be expanded in the future when someone needs it?

riv-ljames:
@sea-bass ive tidied up the code and deduplicated the logic between the state validation class and the multi state validation class. Let me know what you think!

rr-mark:
Is this just waiting on https://github.com/moveit/moveit_msgs/pull/186 to be merged, before the tests will pass?

sjahr (approved):

Issue #3457, opened 2025-04-30
TOTG non-zero initial velocity


buckleytoby:
### Is your feature request related to a problem? Please describe.

Based off this issue in the ros1 moveit repo: https://github.com/moveit/moveit/issues/2538 

I created a fork here: https://github.com/NU-Haptics-Lab/moveit2/tree/iron-feature-initial-velocity

Should be done within a few days. The idea is you can use the initial position velocity vector to fit a circle, where the radius of curvature is a function of the max allowable acceleration.

Full proof to follow at some point.

### Describe the solution you'd like

if the TOTG input robot trajectory has non-zero initial state velocity, the output trajectory should maintain that initial velocity

PR #3460, opened 2025-05-02
TOTG non-zero initial velocity

buckleytoby:
### Description
See issue #3457.

### Checklist


mikeferguson:
The bot is correct - all PRs should target the main branch, they can then be backported. I will additionally note that Iron is actually EOL so we won't be back porting it there (this has the potential to still go into Jazzy, but probably not Humble since it has API changes).

PR #3478, opened 2025-05-24
Fix/controller yaml defaults

### LINKED ISSUE #3434: Fix joint trajectory controller duplicate key generation in setup assistant properly
### Description

See https://github.com/moveit/moveit2/pull/3316#issuecomment-2798165248 for more information.

The original fix (#2959) fixed key duplication in the event that the `action_ns` and `default` keys were specified in the UI, but seem to break the workflow if those text boxes are left blank.

There needs to be some if-condition (and hopefully tests) added to properly address both cases.

### ROS Distro

Rolling

### OS and version

Ubuntu 24.p4

### Source or binary build?

Source

_No response_

### If source, which branch?

_No response_

### Which RMW are you using?

None

### Steps to Reproduce

no u

### Expected behavior

Should work under both conditions outlined above.

### Actual behavior

Works only if the values are set in the UI, but not when blank.

_No response_

Eldgar:
### Fix
When the user skips “Edit Selected,” Setup Assistant dropped
`action_ns` and `default` from `moveit_controllers.yaml` (see #3434).
This patch adds fallback values:

  action_ns: follow_joint_trajectory  
  default: true

### Test
Adds `InjectedDefaultsWhenBlank` to `test_controllers.cpp` to verify
those keys are always present.

Fixes #3434


mikeferguson:
I've triggered the CI - looks like the linter isn't happy with a bit of the whitespace

rhaschke:
Instead of manually modifying the YAML generation, the default values should be simply defined when adding controller configs. I adapted the PR appropriately.

rhaschke (approved):

PR #3484, opened 2025-05-28
Check rclcpp::ok before sleeping for clean shutdown

MarqRazz:
### Description

In MTC [PR#684](https://github.com/moveit/moveit_task_constructor/pull/684) I found the move_group fails to shut down cleanly when preempting motion. This is because we are asking the node to sleep when the stack has been requested to shut down. To fix this I added a simple guard to make sure `rclcpp::ok()` before sleeping.

Here is a snippet of the stack trace when shutting down.
```
[move_group-5] #10   Source "/root/my_ws/src/moveit2/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp", line 549, in scenePublishingThread [0x7d8db9c1afaf]
[move_group-5]         546:       planning_scene_publisher_->publish(msg);
[move_group-5]         547:       if (is_full)
[move_group-5]         548:         RCLCPP_DEBUG(logger_, "Published full planning scene: '%s'", msg.name.c_str());
[move_group-5]       > 549:       rate.sleep();
[move_group-5]         550:     }
[move_group-5] #9    Object "/opt/ros/jazzy/lib/librclcpp.so", at 0x7d8db914c480, in rclcpp::Rate::sleep()
[move_group-5] #8    Object "/opt/ros/jazzy/lib/librclcpp.so", at 0x7d8db90a0238, in rclcpp::Clock::sleep_for(rclcpp::Duration, std::shared_ptr<rclcpp::Context>)
```

### Checklist

### DIFF
--- a/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
+++ b/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
@@ -546,7 +546,11 @@ void PlanningSceneMonitor::scenePublishingThread()
       planning_scene_publisher_->publish(msg);
       if (is_full)
         RCLCPP_DEBUG(logger_, "Published full planning scene: '%s'", msg.name.c_str());
-      rate.sleep();
+      // make sure the node is still running before sleeping otherwise it will crash on shutdown
+      if (!rclcpp::ok())
+        rclcpp::shutdown();
+      else
+        rate.sleep();
     }
   } while (publish_planning_scene_);
 }


moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
```diff
@@ -546,7 +546,11 @@ void PlanningSceneMonitor::scenePublishingThread()
       planning_scene_publisher_->publish(msg);
       if (is_full)
         RCLCPP_DEBUG(logger_, "Published full planning scene: '%s'", msg.name.c_str());
-      rate.sleep();
+      // make sure the node is still running before sleeping otherwise it will crash on shutdown
+      if (!rclcpp::ok())
+        rclcpp::shutdown();
+      else
+        rate.sleep();
```
rhaschke, line 553: I think, you don't need to trigger rclcpp::shutdown(). Just leave the loop.
```suggestion
      // finish thread on rclcpp shutdown (otherwise rate.sleep() will crash)
      if (!rclcpp::ok())
        break;
      rate.sleep();
```

rhaschke (commented):
Is this a new issue, i.e. a regression in rclcpp? If so, that should be reported upstream.

rhaschke (approved):

PR #3493, opened 2025-06-04
Qt6 compile fix

mosfet80:
### Description

the patch allows compilation with both qt5 and qt6

### Checklist


rhaschke:
Is this a backport of https://github.com/moveit/moveit/pull/3582? If so, why didn't you declare it like this?
If not, why not? The PR was already reviewed and we should not do work twice.

mosfet80:
> Is this a backport of [moveit/moveit#3582](https://github.com/moveit/moveit/pull/3582)? If so, why didn't you declare it like this? If not, why not? The PR was already reviewed and we should not do work twice.

In fact, if you compile with only Qt6, the build fails.
The patch fixes the problem of compiling Moveit with Qt 5 and 6 simultaneously. If I analyze the cmake file moveit_ros/visualization/motion_planning_rviz_plugin/CMakeLists.txt,
the Qt5 library is included even when using Qt6.
And, qt5_wrap_cpp is also used when using Qt6.

nbbrooks:
@mosfet80 Do you remember if there were specific packages that fail with the https://github.com/moveit/moveit/pull/3582 approach if you compile with only Qt6?

Would you mind isolating this fix to just the Qt6 changes and remove the ament_target_dependency changes for now? I am going to get https://github.com/moveit/moveit2/pull/3706 merged in which will not cause that deprecation to fail the CI job.

nbbrooks:
Thank you for the intial effort here - I picked it up in https://github.com/moveit/moveit2/pull/3707 

@rhaschke I took a look at the implementation in https://github.com/moveit/moveit/pull/3582 - on the cmake side of things our resolution needed to look different because rviz2 does not set rviz_QT_VERSION

PR #3514, opened 2025-06-27
moveit_ros/perception/pointcloud_octomap_updater: log missing parameters

szobov:
### Description

When the `sensor_3d.yaml` file with `PointCloudUpdater` is used, but some required parameters are missing, `moveit_group` node fails to start with the error log: `Failed to configure updater of type PointCloudUpdater`, which doesn't give any clues of what's actually failed.

This PR is aimed at providing more user-friendly log messages that explicitly tell what is missing, e.g.:
```
[move_group-2] [ERROR] [1750954248.813778881] [move_group]: Missing parameters under 'vision_perception_node': , 'max_update_rate', 'filtered_cloud_topic'
[move_group-2] [ERROR] [1750954248.814011991] [move_group.moveit.moveit.ros.occupancy_map_monitor]: Failed to configure updater of type PointCloudUpdater
```

The changes are done only in the `pointcloud_octomap_updater.cpp`, but if accepted, it can also be reused in `depth_image_octomap_updater.cpp`, which has the same behaviour.

### DIFF
--- a/moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
+++ b/moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
@@ -74,15 +74,38 @@ PointCloudOctomapUpdater::PointCloudOctomapUpdater()
 
 bool PointCloudOctomapUpdater::setParams(const std::string& name_space)
 {
+  auto check_required = [this, &name_space](const std::string& key, auto& target,
+                                            std::vector<std::string>& missing_keys) {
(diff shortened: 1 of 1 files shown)


szobov:
Hey @sjahr 
Thanks for the review and reply.
I checked the failing job on CI, but I was unable to identify whether it was caused by the changes I implemented.

Can you please point out if I need to make any changes to facilitate this merge?
Cheers!

szobov:
Hey @sjahr 
Let me know if I need to make any changes to make it merged.
Thanks!

PR #2994, opened 2024-09-09
Update ompl_interface with latest version of OMPL

MarqRazz:
### Description

I'm working on debugging some issues and require building MoveIt2 and OMPL from source. This PR updates the package `moveit_planners_ompl` to work with the latest changes made in OMPL and it's more modern approach to exporting the library but will require building it from source until a new release is made (I am working on getting it released and will keep this PR updated with the status). 

I was also having issues with OMPL but that package was recently updated to properly work with Colcon. [Here is a link](https://github.com/ompl/ompl/issues/1175) to the issue.

Big thanks to @JafarAbdi for helping me debug some of these issues and getting this updated! :1st_place_medal: 

### Checklist
- [ ] **Required by CI**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)
- [ ] Extend the tutorials / documentation [reference](http://moveit.ros.org/documentation/contributing/)
- [ ] Document API changes relevant to the user in the [MIGRATION.md](https://github.com/moveit/moveit2/blob/main/MIGRATION.md) notes
- [ ] Create tests, which fail without this PR [reference](https://moveit.picknik.ai/humble/doc/examples/tests/tests_tutorial.html)
- [ ] Include a screenshot if changing a GUI
- [ ] While waiting for someone to review your request, please help review [another open pull request](https://github.com/moveit/moveit2/pulls) to support the maintainers

[//]: # "You can expect a response from a maintainer within 7 days. If you haven't heard anything by then, feel free to ping the thread. Thank you!"

PR #3517, opened 2025-07-04
Support Docker siblings in IKFast plugin script

NoahR-ATC:
### Description

Adds support for running the IKFast plugin creation script `auto_create_ikfast_moveit_plugin.sh` from within a Docker container.

The script uses Docker containers itself, and was attempting to share files with the spawned containers using bind mounts. If you create your main container with a bind mount to the Docker socket (`/var/run/docker.sock`) then you are able to create sibling containers from inside, and thus would expect the script to execute correctly. The problem is that bind mounts always refer to the host-machine's paths, so if you try to run the script from inside a Docker container the mounts do not resolve correctly.

I solved this by switching to explicit `docker create` and `docker start` commands, and used `docker cp` to copy files in and out, since `docker cp` supports being run from inside a container. This works in-place and does not require any changes to documentation or processes.

### Checklist

### DIFF
--- a/moveit_kinematics/ikfast_kinematics_plugin/scripts/auto_create_ikfast_moveit_plugin.sh
+++ b/moveit_kinematics/ikfast_kinematics_plugin/scripts/auto_create_ikfast_moveit_plugin.sh
@@ -148,8 +148,15 @@ function create_dae_file {
       echo "Failed. Converting urdf to Collada (in docker)"
       build_docker_image
       cp "$INPUT" "$TMP_DIR/robot.urdf"
-      run_quiet docker run --rm --user $(id -u):$(id -g) -v $TMP_DIR:/input --workdir /input -e HOME=/input \
-             fixed-openrave:latest rosrun collada_urdf urdf_to_collada robot.urdf robot.dae
+      run_quiet docker create \
+        --workdir /input \
+        -e HOME=/input \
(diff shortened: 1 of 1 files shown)


moveit_kinematics/ikfast_kinematics_plugin/scripts/auto_create_ikfast_moveit_plugin.sh
```diff
@@ -171,9 +179,16 @@ EOF
    echo "Running $cmd"
 
    # run $cmd in docker as current user, outputting files to $TMP_DIR/.openrave
-   run_quiet docker run --rm --user $(id -u):$(id -g) \
-      -v $TMP_DIR:/input --workdir /input -e HOME=/input \
-      fixed-openrave:latest $cmd
+   run_quiet docker create \
+           --user  $(id -u):$(id -g) \
```
rhaschke, line 183: Same here:
```suggestion
```

NoahR-ATC:
Thanks for the investigation, I have applied your suggestions now and rebased in main. The CI tests are currently failing on my fork because of the same deprecation error a few other recent PRs have been getting (e.g. [this run for #2994](https://github.com/moveit/moveit2/actions/runs/17191606195/job/48825475857)). Seems like #3562 is trying to address this.

PR #3522, opened 2025-07-08
Get relative frame from robot state

### LINKED ISSUE #3521: `RobotState::setFromIK` sporadically transforms poses wrong
### Description

We are seeing a problem where sometimes a PILZ PTP plan to a pose target ends at the wrong pose. When the pose is "wrong", it is off by ~0.1mm (compared to an ~1m robot), and when it is "right" it is correct to many decimal places.

Debugging the problem, the issue seems to be the calls to `pose_parent_to_frame = getFrameTransform(pose_frame)` and `tip_parent_to_tip = getFrameTransform(solver_tip_frame)` in `RobotState::setFromIK`.
 - The two calls to `getFrameTransform` both call `updateLinkTransforms`.
 - This means the `RobotState` can update between the calls.
 - This means that `pose_parent_to_frame.inverse() * tip_parent_to_tip` is not the rigid transform from `pose_frame` to `solver_tip_frame`, because it includes the change from updating the robot state.

My proposed solution is to add an overload to `getFrameTransform` which takes a `source_frame_id` and `destination_frame_id`, rather than just one `frame_id`, and returns the transform from `source_frame_id` to `destination_frame_id`. Importantly, this overload would only call `updateLinkTransforms` once, before fetching both pose.

Does this sound like a plausible diagnosis of this error? Does my proposed solution seem sensible?

I'll raise a PR when I've done a bit more testing.

### ROS Distro

Humble

### OS and version

Ubuntu 22.04

### Source or binary build?

Source

_No response_

### If source, which branch?

cc3a82d8434ad23b297fb124c370eb13745456ed (humble plus a few minor and unrelated tweaks)

### Which RMW are you using?

None

### Steps to Reproduce

 - Set up a 6DOF robot arm
 - Attach a collision object to the tip of the arm
 - Perform a PILZ PTP plan, to a pose target specified using a subframe of the collision object.
 - Perform FK on the final joint state of the resulting trajectory, and check for self-consistency with the target pose.
 - Run this many times on a slow computer.

### Expected behavior

A PTP trajectory to a pose target should end at that pose (up to numeric error).

### Actual behavior

A PTP trajetory to a pose target sporadically ends at a slightly different pose to the target.

_No response_

riv-mjohnson:
### Description

- Fixes #3521 
- Adds overloads to `RobotState::getFrameTransform` which take destination and source frames, and only calls `updateLinkTransforms` once.
- Uses this new overload in `RobotState::setFromIK`, to calculate the static transform between `pose_frame` and `solver_tip_frame`, avoiding the possibility of `updateLinkTransforms` invalidating the calculation.

Can this be backported to Humble etc. when merged.

### Checklist

### DIFF
--- a/moveit_core/robot_state/include/moveit/robot_state/robot_state.hpp
+++ b/moveit_core/robot_state/include/moveit/robot_state/robot_state.hpp
@@ -1584,6 +1584,26 @@ public:
    * The returned transformation is always a valid isometry. */
   const Eigen::Isometry3d& getFrameTransform(const std::string& frame_id, bool* frame_found = nullptr) const;
 
+  /** \brief Get the transformation matrix from the frame identified by \e source_frame_id to the frame identified by \e
+   * destination_frame_id
+   *
+   * If source_frame_id or destination_frame_id was not found, \e frame_found is set to false and an identity transform
+   * is returned.
+   *
+   * The returned transformation is always a valid isometry. */
+  Eigen::Isometry3d getFrameTransform(const std::string& destination_frame_id, const std::string& source_frame_id,
+                                      bool* frames_found = nullptr);
+
+  /** \brief Get the transformation matrix from the frame identified by \e source_frame_id to the frame identified by \e
+   * destination_frame_id
+   *
+   * If source_frame_id or destination_frame_id was not found, \e frame_found is set to false and an identity transform
+   * is returned.
+   *
+   * The returned transformation is always a valid isometry. */
+  Eigen::Isometry3d getFrameTransform(const std::string& destination_frame_id, const std::string& source_frame_id,
+                                      bool* frames_found = nullptr) const;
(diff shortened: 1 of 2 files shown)


riv-mjohnson:
This doesn't seem to solve the problem. Since we already have `pose_parent`, is there a way to calculate the transform `pose_parent_from_pose_frame` directly, so we can go `pose_frame_from_solver_tip_frame = pose_frame_from_pose_parent.inverse() * tip_parent_from_solver_tip_frame`, which only uses rigid transforms?

Unhelpfully, the current code (before this PR) calls them `pose_parent_to_frame` and `tip_parent_to_tip`, but I don't think that's what is actually being calculated. I think `getFrameTransform(pose_frame)` returns `ik_base_frame_from_pose_frame`. Is this right?

Issue #3243, opened 2025-01-13
Completely lacking documentation on moveit servo


PietroValerio:
### Description

Completely lacking documentation, cpp api assumes that the controller used accepts messages of type` trajectory_msgs::msg::joint_trajectory`, what about if we have a controller only accepting `std_msgs/msg/Float64MultiArray`. Nothing inside the documentation covering this use case, you need to go digging inside the codebase to find that there is a function `composeMultiArrayMessage` that maybe can do the job, but you can't know it because is not documented.
You want to be the industry de-facto standard, before you should worry about fixing your documentation

### ROS Distro

Jazzy

### OS and version

24.04

### Source or binary build?

Source

### If source, which branch?

main

### Which RMW are you using?

None

### Steps to Reproduce

Watch moveit servo tutorial

### Expected behavior

Fully documented

### Actual behavior

Lacking documentation

mikeferguson:
This is an open source project, staffed by volunteers. PRs to improve documentation are welcome.

attu0:
could you check my pr and tell if any changes required

mikeferguson:
> could you check my pr and tell if any changes required

I'm not the right person to review moveit servo stuff - it's one of the modules I really haven't worked with at all

sea-bass:
> [@mikeferguson](https://github.com/mikeferguson) could you check my pr and tell if any changes required

There are already 2 other reviewers on that PR, with pending comments from a few days back. You should be able to proceed by looking at those existing comments.

vedbo:
Hey everyone! I've been working on documentation for `ros2_control` recently (Issue #417) and I'd love to help flesh out the MoveIt Servo documentation as well. It's such a core part of real-time control, so having solid 'how-to' guides would be a huge value-add. I'm planning to dive into the source and example configurations to see where the gaps are. I'd love to take this on as part of my GSoC 2026 preparation! My portfolio: https://vedbo.github.io/

PR #3543, opened 2025-08-01
Improve documentation for composeMultiArrayMessage to clarify usage

hihry:
### Description

Addresses #3243

This PR improves the documentation for the `composeMultiArrayMessage` function in `common.cpp` within the `moveit_servo` module.

- Added a Doxygen-compatible comment block to clarify the function’s purpose, parameters, return value, and usage.
- Helps developers understand how to use this function when working with controllers that accept `std_msgs::msg::Float64MultiArray` messages (rather than `trajectory_msgs::msg::JointTrajectory`).
- Enhances clarity and aligns with MoveIt 2’s commitment to being an accessible, industry-standard robotics framework.

This change is purely a documentation update and does not affect runtime behavior.

### DIFF
--- a/moveit_ros/moveit_servo/src/utils/common.cpp
+++ b/moveit_ros/moveit_servo/src/utils/common.cpp
@@ -254,7 +254,17 @@ void updateSlidingWindow(KinematicState& next_joint_state, std::deque<KinematicS
   // add next command
   joint_cmd_rolling_window.push_back(next_joint_state);
 }
-
+/**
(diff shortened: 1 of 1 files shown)


sea-bass (changes requested):
The header file already has the docstrings. Please add it to the one there.

sea-bass (approved):
Thanks! I was just checking this on some downtime since the same issue has come up before. Once an active maintainer approves, you should be all set.

nbbrooks (approved):

PR #3547, opened 2025-08-09
Initialize the namespace property for custom iMarkers in rviz

eholum:
### Description

MoveIt's iMarker doesn't initialize the marker namespace property, so that functionality is broken when you try to add other iMarkers alongside moveit's planning interface. Normally this just breaks the cursor, though in some cases we've seen rviz crash. I was under the impression that pluginlib might still load the default plugin for other markers, but that does not seem to be the case.

Regardless, I don't think there's a downside to providing the node abstraction to the namespace property object, so this PR just uncomments those lines.

Before (cursor gets stuck as a spinning wheel):

https://github.com/user-attachments/assets/753dc0bd-289f-47e4-8046-f22c667d8b5a

After:

https://github.com/user-attachments/assets/ccfcdb4d-5c47-47d4-a2e8-3f311f365d39

### Checklist

### DIFF
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/interactive_marker_display.cpp
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/interactive_marker_display.cpp
@@ -91,13 +91,13 @@ InteractiveMarkerDisplay::InteractiveMarkerDisplay()
 
 void InteractiveMarkerDisplay::onInitialize()
 {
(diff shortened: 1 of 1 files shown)


v4hn (approved):
This code was copied over from rviz in #342 where @JafarAbdi apparently commented that section to avoid using the rviz node abstraction.
The handle here is only used to select the potential values for the property, so using the rviz handle for that should be fine.

PR #3548, opened 2025-08-10
[Fix][Cross-Platform] Replace <types.h> with <cstddef> and remove explicit const from std::map key for macOS

idesign0:
### Description

This PR fixes a compilation issue on macOS caused by the use of const std::pair<std::string, std::string> as the key type in a std::map.

### Details:

- The original code used const std::pair as the key type:
```cpp
using DistanceMap = std::map<const std::pair<std::string, std::string>, std::vector<DistanceResultsData>>;
```
- This caused assignment errors because `std::map` already treats keys as const internally.
- Explicitly adding `const` leads to conflicts with internal map operations and strict compiler checks.
- The fix removes the explicit const qualifier from the key type:
```cpp
using DistanceMap = std::map<std::pair<std::string, std::string>, std::vector<DistanceResultsData>>;
```
- This change ensures compatibility across platforms, including macOS, and adheres to the C++ standard container usage best practices.

### Description
Replaces the non-standard `<types.h>` header with the standard `<cstddef>` to ensure `size_t` is available on all platforms. This fixes compilation issues on macOS while remaining compatible with Linux and other platforms.

### Changes
```cpp
- #include <types.h>
+ #include <cstddef>  // for size_t, cross-platform


forrest-rm (approved):
Looks simple enough

idesign0:
> Looks simple enough
can you push it further? :)

rhaschke (approved):

PR #3556, opened 2025-08-14
Fix ERROR stream + check only active controllers for multiple chained controllers

thomaspeyrucain:
### Description

Hello Moveit Team,

When trying to control TIAGo Pro using moveit in Rviz I got this error:
```
[move_group-1] [ERROR] [1755075355.583941867] [moveit.plugins.ros_control_interface]: Controller with name %s chains to more than one controller. Chaining to more than one controller is not supported.
[move_group-1] [ERROR] [1755075355.584044598] [moveit_ros.trajectory_execution_manager]: Unable to identify any set of controllers that can actuate the specified joints: [ arm_left_1_joint arm_left_2_joint arm_left_3_joint arm_left_4_joint arm_left_5_joint arm_left_6_joint arm_left_7_joint ]
[move_group-1] [ERROR] [1755075355.584057241] [moveit_ros.trajectory_execution_manager]: Known controllers and their joints:
[move_group-1] 
```
This was due to a controller that is expected to be chained to several other controllers to be able to work, however this controller is not active when I did the test, it is just spawned by default in inactive

My change is to check chained controller only for active controllers and to fix the error message

Would this work or you prefer this to be fixed differently ?


thomaspeyrucain:
Hello @henningkayser @tylerjw ,
Would it be possible to review this MR ?

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -596,11 +596,11 @@ class Ros2ControlManager : public moveit_controller_manager::MoveItControllerMan
     dependency_map_reverse_.clear();
     for (auto& controller : result->controller)
     {
-      if (controller.chain_connections.size() > 1)
+      if (isActive(controller) && controller.chain_connections.size() > 1)
```
MarqRazz, line 599: I don't have a way to verify this so any [test](https://github.com/thomaspeyrucain/moveit2/blob/a62d02d1172dba4bf90cf7da58e0d6af73866a4b/moveit_plugins/moveit_ros_control_interface/test/test_controller_manager_plugin.cpp) you could add would be extremely helpful. 

Example scenario:
```
arm_1_JTC active
arm_2_JTC active
arm_3_chained inactive or active
```

Can you confirm that when `controller = arm_1_JTC`, `controller.chain_connections.size()` == 0
and when `controller = arm_3_chained`, `controller.chain_connections.size()` == 1?

thomaspeyrucain:
Hello,
I could not find a way to perform the test asked as I would need to change how the code is written and change the way it works

I wanted to add a test where 3 controllers A, B and C with C chained to A and B
If you activate A and B it should work but if you activate C it should fail; however the code is made in a way that it is not sure what part failed as this part of the code is not returning True or False

https://github.com/moveit/moveit2/blob/5a8ef4df4ba9bcedd1e65f9042e6fa718e215bf0/moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp#L207-L211

MarqRazz (approved):
Bummer we can't make a test.

MarqRazz:
Looks like you need to run `pre-commit` to fix the Formatting test @thomaspeyrucain

thomaspeyrucain:
> * Waiting on 1 reapproval from someone other than the last pusher. Review from MarqRazz is stale because it was submitted before the most recent code changes

Done, Thanks ^^

MarqRazz (approved):

nbbrooks:
@mergify backport humble

nbbrooks:
@mergify backport jazzy

nbbrooks:
@mergify backport kilted

nbbrooks:
@mergify backport humble

nbbrooks:
@mergify backport jazzy

nbbrooks:
@mergify backport kilted

PR #3608, opened 2025-11-05
Feature/make use of max cartesian speed in MotionPlanRequest with pilz planner

AimanHaidar:
### Description

This pull request make use of  the **`max_cartesian_speed`** field in the `MotionPlanRequest` because I noticed it is not useful for the **Pilz planner**.

Currently, the planner always uses the value from `pilz_cartesian_limits.yaml` (loaded to `planner_limits_`) and does **not** make use of the `max_cartesian_speed` specified in the request. Moreover, setting these limits dynamically with

```bash
ros2 param set /move_group <param>
```

does not update the planner behavior as expected.

This PR introduces a small change to apply the `max_cartesian_speed` value from the **request item** when planning the trajectory, instead of always relying on the static configuration from `planner_limits_`.

As a result, users can now specify Cartesian speed limit directly in their MotionPlanRequest, and seems to be related to moveit issues [issue1](https://github.com/moveit/moveit/issues/3094#issue-1170944639) and [issue2](https://github.com/moveit/moveit/issues/3443#issue-1743869301).

I don't know if my way of coding is good for the planner, since I didn't make use of `cartesian_speed_limited_link` field 

---

#### Example

Below is an example of a sequence request containing three planning items, each with different `max_cartesian_speed` values applied correctly after this PR:
>note: speed inversly proportional to points density

Screenshot from 2025-11-05 15-14-48

---

### Checklist

### DIFF
--- a/moveit_planners/pilz_industrial_motion_planner/src/trajectory_generator_circ.cpp
+++ b/moveit_planners/pilz_industrial_motion_planner/src/trajectory_generator_circ.cpp
@@ -202,9 +202,9 @@ void TrajectoryGeneratorCIRC::plan(const planning_scene::PlanningSceneConstPtr& 
                                    const planning_interface::MotionPlanRequest& req, const MotionPlanInfo& plan_info,
                                    double sampling_time, trajectory_msgs::msg::JointTrajectory& joint_trajectory)
 {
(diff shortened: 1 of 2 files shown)


AimanHaidar:
I will close this because it is now included in #3610

PR #3623, opened 2025-11-25
Fix clang-format pre-commit entry to not use system

EzraBrooks:
Isn't the whole point of pre-commit that the tooling self-installs? :)


shaur-k (approved):

nbbrooks (approved):
I do not know why it was set up this way.

Issue #3612, opened 2025-11-18
findWayPointIndicesForDurationAfterStart will return index which is out of bound


XingxinHE:
### Description

See the code at

https://github.com/moveit/moveit2/blob/ca2bb1ba4581b7213d68094458f2649de17d237e/moveit_core/robot_trajectory/src/robot_trajectory.cpp#L509-L542

If the `duration` is bigger than the duration of the trajectory, the `index` will be `index==num_points` making this line `double before_time = running_duration - duration_from_previous_[index];` invalid and also output a `double blend` not in the `[0, 1]` inclusive according to https://github.com/moveit/moveit2/blob/ca2bb1ba4581b7213d68094458f2649de17d237e/moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.hpp#L312

### ROS Distro

Rolling

### OS and version

Ubuntu 24.04

### Source or binary build?

Source

### If source, which branch?

main, jazzy, humble

### Which RMW are you using?

None

### Steps to Reproduce

Just give a `duration` bigger than the trajectory duration to the `findWayPointIndicesForDurationAfterStart()`.

### Expected behavior

It should be `[0, 1]` and no error.

### Actual behavior

Error and also `blend is not in `[0, 1]`

XingxinHE:
I propose the following fix and I am happy to open a PR to support OSS.

```cpp
void RobotTrajectory::findWayPointIndicesForDurationAfterStart(double duration, int& before, int& after,
                                                               double& blend) const
{
  if (duration < 0.0)
  {
    before = 0;
    after = 0;
    blend = 0;
    return;
  }

  std::size_t index = 0, num_points = waypoints_.size();
  double running_duration = 0.0;
  for (; index < num_points; ++index)
  {
    running_duration += duration_from_previous_[index];
    if (running_duration >= duration)
      break;
  }

  // early return
  if (index >= num_points)
  {
    before = num_points - 1;
    after = num_points - 1;
    blend = 1.0;
    return;
  }

  before = std::max<int>(index - 1, 0);
  after = std::min<int>(index, num_points - 1);


  double before_time = running_duration - duration_from_previous_[index];
  if (after == before)
  {
    blend = 1.0;
  }
  else
  {
    blend = (duration - before_time) / duration_from_previous_[index];
    // clamp for numerical error
   blend = std::max(0.0, std::min(1.0, blend));
  }
}
```

nbbrooks:
[...]

PR #3626, opened 2025-11-29
Fix findWayPointIndicesForDurationAfterStart method when out of bounds

XingxinHE:
### Description

The fix is to address the bug raised in issue #3612 which causes UB. The change has been
- clang-format/tidy with docker image `moveit/moveit2:rolling-ci`
- run unit test with `moveit_core/robot_trajectory/test/test_robot_trajectory.cpp`

### Checklist

### Misc.

Actually, the wording "Create tests, which fail without this PR" is not strictly correct. Without the fix, the unit tests will "pass", but it passes with undefined behavior. Meaning, all the `EXPECT_xx` after the call `findWayPointIndicesForDurationAfterStart` will be omitted and the google test doesn't actually check it. I once plan to use the `EXPECT_EXIT` but UB doesn't always result in a SIGSEGV. 

In my perspective, I prefer the `.at()` over `[index]`. See the [guideline](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#slcon3-avoid-bounds-errors). But I guess using `.at(` might cause a little bit performance degrade in real time control of robot? If we use the `[index]` approach, then the unit test should run with AddressSanitizer or something. Since I am not quite familiar with the CI/CD of moveit2, therefore I defer whether to keep current implementation or change to `.at()` to @nbbrooks .

### DIFF
--- a/moveit_core/robot_model/src/planar_joint_model.cpp
+++ b/moveit_core/robot_model/src/planar_joint_model.cpp
@@ -355,6 +355,10 @@ void PlanarJointModel::computeVariablePositions(const Eigen::Isometry3d& transf,
 
   ASSERT_ISOMETRY(transf)  // unsanitized input, could contain a non-isometry
   Eigen::Quaterniond q(transf.linear());
+  if (q.w() < 0.0)
+  {
+    q.coeffs() *= -1.0;
+  }
   // taken from Bullet
   double s_squared = 1.0 - (q.w() * q.w());
   if (s_squared < 10.0 * std::numeric_limits<double>::epsilon())
--- a/moveit_core/robot_model/test/test.cpp
+++ b/moveit_core/robot_model/test/test.cpp
@@ -172,6 +172,38 @@ TEST(FloatingJointTest, interpolation_test)
   }
 }
 
+TEST(PlanarJointTest, ComputeVariablePositionsNormalizeYaw)
+{
+  // Create a simple planar joint model with some dummy parameters
+  moveit::core::PlanarJointModel pjm("joint", 0, 0);
+
+  // Test various angles for normalization
+  std::vector<double> test_angles = { 0.0, M_PI / 2, -M_PI / 2, M_PI, -M_PI, 3.0, -3.0, 4.1, -4.1 };
+  for (double angle : test_angles)
+  {
+    // Create a transform with the given yaw angle
+    Eigen::Isometry3d transform = Eigen::Isometry3d::Identity();
+    Eigen::AngleAxisd q = Eigen::AngleAxisd(angle, Eigen::Vector3d::UnitZ());
+    transform.linear() = q.toRotationMatrix();
+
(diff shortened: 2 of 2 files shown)


moveit_core/robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -509,7 +509,7 @@ RobotTrajectory& RobotTrajectory::setRobotTrajectoryMsg(const moveit::core::Robo
 void RobotTrajectory::findWayPointIndicesForDurationAfterStart(double duration, int& before, int& after,
                                                                double& blend) const
 {
-  if (duration < 0.0)
+  if (duration < 0.0 || waypoints_.empty())
```
XingxinHE: The change in line 539 is not enough to handle the case that waypoints is empty. i.e. index an empty deque with `[0]`

XingxinHE:
Hey @nbbrooks , this pr is ready for review.

nbbrooks (commented):
This is looking good considering the unclear intended behavior from the original function description and the current [old] error handling patterns in the codebase. As I see it, the expected output with edge cases should be:

```
// If waypoints are empty: before == 0, after == 0, blend == 0 (invalid - empty trajectory)
// If target duration < 0: before == 0, after == 0, blend == 0 (invalid - before trajectory has started)
// If target duration > total duration of waypoints, before == size() - 1, after == size() - 1, blend == 1 (after trajectory has completed)
// If 1 waypoint and target duration == total duration of waypoints, before == 0, after == 0, blend == 1 (at single point trajectory completion)
// Else before == last waypoint index with total duration < target duration, after == before + 1, blend == fraction of target duration between total duration @ before and total duration @ after
//   Therefore, if 2 waypoints and target duration == total duration of waypoints, before == 0, after == 1, blend == 1 (at trajectory completion)
```

Since you are being so thorough, here are the last changes I suggest:

1. Change access to use at() as you noted. I would not expect this to be in a real-time loop or have significant impact.
2. Consider updating your test names to be more descriptive of the corner case they are checking (my suggestions are `RobotTrajectoryFindWayPointIndicesBetweenWaypoints`, `RobotTrajectoryFindWayPointIndicesAtLastOfManyWaypoints`, `RobotTrajectoryFindWayPointIndicesAfterLastWaypoint`, and `RobotTrajectoryFindWayPointIndicesEmptyWaypoints`)
3. Add two more tests to check the remaining cases: `RobotTrajectoryFindWayPointIndicesBeforeFirstWaypoint` and `RobotTrajectoryFindWayPointIndicesAtLastOfSingleWaypoint` or something to that effect.

XingxinHE:
Nice! Then I change the unit test to use `EXPECT_NO_THROW` and future mistakes will be caught. I've updated the tests and docs as well.

nbbrooks (approved):
Thank you!

XingxinHE:
Thank you Nathan for the code review! I learned a lot.

PR #3635, opened 2025-12-09
Populate velocities and accelerations for multidof joint trajectories

stephanie-eng:
### Description
If I were to use `SetRobotTrajectoryMsg`, to set a trajectory message, it appears that only positions are handled for multidof joints, and velocity/acceleration information is discarded. 

This change handles the contents of multidof joint trajectories similarly to done [here](https://github.com/moveit/moveit2/blob/main/moveit_core/robot_trajectory/src/robot_trajectory.cpp#L381-L403) and populates the velocities and accelerations in the robot states. I've also taken the liberty of adding a test to ensure that the multidof joint trajectory positions and accelerations appear as expected.

This currently doesn't support `angular/x` or `angular/y`, for which I've added a comment noting this.

As a note, I tested this on Jazzy. I currently don't have a rolling build going, but I don't think things look that different in this part of the code base anyway.

### DIFF
--- a/moveit_ros/planning/planning_request_adapter_plugins/test/test_check_start_state_bounds.cpp
+++ b/moveit_ros/planning/planning_request_adapter_plugins/test/test_check_start_state_bounds.cpp
@@ -50,7 +50,6 @@ class TestCheckStartStateBounds : public testing::Test
 protected:
   void SetUp() override
   {
-    rclcpp::init(0, nullptr);
     node_ = std::make_shared<rclcpp::Node>("test_check_start_state_bounds_adapter", "");
 
     // Load a robot model and place it in a planning scene.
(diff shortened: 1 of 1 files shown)


stephanie-eng:
CI failure unrelated - same as in https://github.com/moveit/moveit2/pull/3628

EzraBrooks (approved):
doing god's work ensuring multi-DOF joints actually function. nice test

moveit_core/robot_trajectory/test/test_robot_trajectory.cpp
```diff
@@ -38,6 +38,7 @@
 #include <moveit/robot_state/robot_state.hpp>
 #include <moveit/robot_trajectory/robot_trajectory.hpp>
 #include <moveit/utils/robot_model_test_utils.hpp>
+#include "moveit_msgs/msg/robot_trajectory.hpp"
```
stephanie-eng:
```suggestion
#include <moveit_msgs/msg/robot_trajectory.hpp>
```

stephanie-eng:
@EzraBrooks Sorry - I made your review stale since I decided to go muck with angular brackets. This is still blocked on CI anyway...

PR #3636, opened 2025-12-09
Hopefully fix Tutorial Docker images

stephanie-eng:
### Description
The Tutorial Docker Images pipeline appears to be failing because `cache-source` and `cache-target` have been deprecated in favour of `cache-map`. This PR updates "inject ccache into docker" step with the new `cache-map`.

No idea if this will fix things fully, but likely a step in the right direction.


stephanie-eng:
Looks like it got past the troublesome step. Hopefully it's good from here on out, but either way, this will be an improvement.

stephanie-eng:
Lol, it appears some other junk in Tutorial Docker Images appears to be borked. It's unrelated to the contents of this PR. ~I don't think I'm going to continue fixing that pipeline for now.~ Looks like [this may be resolved](https://github.com/PickNikRobotics/ros2_robotiq_gripper/pull/103) in the near future. As for this PR, it can stand on its own for at least fixing one of the issues.

PR #3637, opened 2025-12-10
Attempt to unbork CI

stephanie-eng:
### Description

[This](https://github.com/moveit/moveit2/blob/main/moveit_ros/planning/planning_request_adapter_plugins/test/test_check_start_state_bounds.cpp) lovely set of tests appears to be causing some issues in CI.

Rather than doing `init` and `shutdown` for each test, let's `init` once and `shutdown` once. I'm not a wizard enough to understand why this suddenly started failing, but I think it's a good idea to `init` and `shutdown` like the [other](https://github.com/moveit/moveit2/blob/main/moveit_ros/tests/src/move_group_api_test.cpp#L100-L104) [tests](https://github.com/moveit/moveit2/blob/main/moveit_ros/moveit_servo/tests/test_ros_integration.cpp#L214-L218) [do](https://github.com/moveit/moveit2/blob/main/moveit_ros/planning_interface/test/move_group_ompl_constraints_test.cpp#L182-L187).

### DIFF
--- a/moveit_ros/planning/planning_request_adapter_plugins/test/test_check_start_state_bounds.cpp
+++ b/moveit_ros/planning/planning_request_adapter_plugins/test/test_check_start_state_bounds.cpp
@@ -50,7 +50,6 @@ class TestCheckStartStateBounds : public testing::Test
 protected:
   void SetUp() override
   {
-    rclcpp::init(0, nullptr);
     node_ = std::make_shared<rclcpp::Node>("test_check_start_state_bounds_adapter", "");
 
     // Load a robot model and place it in a planning scene.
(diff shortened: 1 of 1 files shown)


stephanie-eng:
Hoping this will fix CI failures in https://github.com/moveit/moveit2/pull/3628 https://github.com/moveit/moveit2/pull/3620 https://github.com/moveit/moveit2/pull/3635 https://github.com/moveit/moveit2/pull/3636 https://github.com/moveit/moveit2/pull/3626 https://github.com/moveit/moveit2/pull/3631

EzraBrooks (approved):
the global context really do be that way


How did this _ever_ work? Thanks for finding and fixing it!

PR #3640, opened 2025-12-12
Attempt to unbork CI round 2 electric boogaloo

stephanie-eng:
Remove `CLANG_TIDY_ARGS` to support run-clang-tidy
Also pin industrial_ci to a specific commit hash, so that way we can break things when we want to, and not when someone else pushes something.

I haven't tested this at all, so I'm just hoping that this is the sum total of all the things we need to get to green checks again.


EzraBrooks:
I suspect it's going to fail on "upload clang-tidy changes" since I think that is expecting some files to exist that will exist no longer after this change?

Not that anyone ever looked at those clang-tidy changes, I don't think - we can likely remove them.

The PR that merged upstream was actually entirely intended to get rid of that sort of workaround and allow people to use the built-in "fixes file" feature in clang-tidy.

PR #3664, opened 2026-01-17
Fix amend deprecation into moveit_ros folder

mosfet80:
### Description
 amend_target_dependencies is deprecated - move to target_link_libraries()
cleaned moveit_ros folder

### Checklist


urfeex:
@mosfet80 thank you for addressing this! I also started working on that a couple of weeks back but couldn't find the time to complete it. What is your main motivation only fixing deprecations in moveit_ros? Just to get a starting point? I assume that the others might have to be migrated, as well if things would want to get built in the future.

My main motivation to start working on that was to make the Qt6 port easier. Do you also have that on the agenda? Otherwise I would start working on that ontop of your changes.

mosfet80:
> @mosfet80 thank you for addressing this! I also started working on that a couple of weeks back but couldn't find the time to complete it. What is your main motivation only fixing deprecations in moveit_ros? Just to get a starting point? I assume that the others might have to be migrated, as well if things would want to get built in the future.
> 
> My main motivation to start working on that was to make the Qt6 port easier. Do you also have that on the agenda? Otherwise I would start working on that ontop of your changes.

The motivation is to remove deprecated functions. I can't compile moveit with the "rolling" version of ament_cmake_target_dependencies.
https://github.com/ament/ament_cmake/blob/rolling/ament_cmake_target_dependencies/CHANGELOG.rst
Unfortunately, it's not a quick fix.

I actually split the work into two commits: the first to overcome the ament_cmake deprecation and the second to compile with QT6.

As soon as I've completed the changes, I'll share the report containing both changes.

urfeex:
That sounds great, then I'll keep my feed still. 

The deprecated ament functions are not a showstopper for compiling on Rolling in my side (and I think neither in the build farm), but the Qt6 dependencies are. I still think, it's a good thing to combine both, as you'll have to touch a lot of link targets with the Qt6 migration, anyway.

Let me know, if I can be of any help and may it only be helping out in reviewing here.

mosfet80:
#3726 duplicated patch.

PR #3669, opened 2026-01-22
Replace all moveit.ros.org to moveit.ai

L4co77:
The ROS.org team is turning of moveit.ros.org, which redirects to moveit.ai currently. We need to find all instances of moveit.ros.org and make PRs to replace them.

### DIFF
--- a/.docker/README.md
+++ b/.docker/README.md
@@ -1,4 +1,4 @@
 # MoveIt Docker Containers
 
 
-For more information see the pages [Continuous Integration and Docker](http://moveit.ros.org/documentation/contributing/continuous_integration.html) and [Using Docker Containers with MoveIt](https://moveit.ros.org/install/docker/).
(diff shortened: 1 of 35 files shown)


130s:
Just curious, has there been any driving discussion to update URL? Only things I saw were this announcement [discourse.openrobotics.org#37482](https://discourse.openrobotics.org/t/migration-of-moveit-projects-to-a-dedicated-github-org/37482), and https://github.com/moveit/moveit.ros.org/issues/851, but I could've missed others.

davetcoleman:
@gbiggs is pushing to turn off moveit.ros.rog per this post from a year ago 

https://discourse.openrobotics.org/t/move-of-nav2-and-moveit-repositories-at-github/37450

davetcoleman:
Looks like CI is very broken for foxy, which I guess is fine since this is way out of LTS

Issue #3667, opened 2026-01-21
Moveit error code -27


usmanNoor5:
### Description

So basically I am using ros2 humble and have moveit install on it. Working on ros for about 2.5 years so know a little bit of it. Moveit is working perfectly fine and not having problem to plan a path their is a error coming when i tried to collied with the collision object **error code -27** which is not mentioned in the error code list 
int32 val

# overall behavior
int32 SUCCESS=1
int32 FAILURE=99999

int32 PLANNING_FAILED=-1
int32 INVALID_MOTION_PLAN=-2
int32 MOTION_PLAN_INVALIDATED_BY_ENVIRONMENT_CHANGE=-3
int32 CONTROL_FAILED=-4
int32 UNABLE_TO_AQUIRE_SENSOR_DATA=-5
int32 TIMED_OUT=-6
int32 PREEMPTED=-7

# planning & kinematics request errors
int32 START_STATE_IN_COLLISION=-10
int32 START_STATE_VIOLATES_PATH_CONSTRAINTS=-11

int32 GOAL_IN_COLLISION=-12
int32 GOAL_VIOLATES_PATH_CONSTRAINTS=-13
int32 GOAL_CONSTRAINTS_VIOLATED=-14

int32 INVALID_GROUP_NAME=-15
int32 INVALID_GOAL_CONSTRAINTS=-16
int32 INVALID_ROBOT_STATE=-17
int32 INVALID_LINK_NAME=-18
int32 INVALID_OBJECT_NAME=-19

# system errors
int32 FRAME_TRANSFORM_FAILURE=-21
int32 COLLISION_CHECKING_UNAVAILABLE=-22
int32 ROBOT_STATE_STALE=-23
int32 SENSOR_INFO_STALE=-24

# kinematics errors
int32 NO_IK_SOLUTION=-31

Secondly when I change the arm_group name from "arm" to am it has to give me error of -15 but it giving 99999 general failure. 
Can anyone help me to find out where I might be doing wrong 

### ROS Distro

Humble

### OS and version

Ubuntu 22.04

### Source or binary build?

Source

### Which RMW are you using?

CycloneDDS

### Steps to Reproduce

I start the moveit working perfectly fine write a code 
`    #!/usr/bin/env python3

    import rclpy
    from rclpy.node import Node
    from rclpy.action import ActionClient
    from moveit_msgs.action import MoveGroup
[...]

PR #3699, opened 2026-03-07
Preserve planner-specific MoveGroup failure codes in /move_action

eclipse0922:
## Summary

Preserve planner-specific non-success `MoveItErrorCodes` in `/move_action` plan-only execution instead of always collapsing them to generic `FAILURE` (`99999`).

## Root cause

`planning_pipeline->generatePlan(...)` returns `false` for any non-success planner result. In `MoveGroupMoveAction::executeMoveCallbackPlanOnly`, that `false` return currently overwrites `res.error_code.val` with `MoveItErrorCodes::FAILURE` even when the planner already set a more specific non-success code such as:

- `INVALID_GROUP_NAME` (`-15`)
- `NO_IK_SOLUTION` (`-31`)
- other planner-specific request/kinematics failures

That destroys information that downstream action clients need for correct diagnosis and handling.

## Change

Only fall back to generic `FAILURE` when the current code is still `SUCCESS` or `UNDEFINED`. If the planner already populated a specific non-success code, preserve it and log that preservation.

## Validation

Validated downstream against `jazzy` / MoveIt `2.12.4` in a live Isaac ROS cuMotion integration harness.

Observed before this patch:

- direct planner path returned specific failure `-31`
- MoveIt-side planner client received `-31`
- `/move_action` still returned generic `99999`

Observed after this patch:

- `/move_action` preserved the original planner-specific failure code (`-31`)
- repeated requests remained stable
- no stale trajectory reuse
- no success-with-empty-trajectory regressions on the direct planner path

This also matches the user-visible symptom reported in #3667 where invalid group or unreachable goals lose their original error code and appear as generic failure.

## Related

Refs #3667

### DIFF
--- a/moveit_ros/move_group/src/default_capabilities/move_action_capability.cpp
+++ b/moveit_ros/move_group/src/default_capabilities/move_action_capability.cpp
@@ -218,7 +218,17 @@ void MoveGroupMoveAction::executeMoveCallbackPlanOnly(const std::shared_ptr<MGAc
     if (!planning_pipeline->generatePlan(scene, goal->get_goal()->request, res, context_->debug_))
     {
       RCLCPP_ERROR(getLogger(), "Generating a plan with planning pipeline failed.");
-      res.error_code.val = moveit_msgs::msg::MoveItErrorCodes::FAILURE;
+      if (res.error_code.val == moveit_msgs::msg::MoveItErrorCodes::SUCCESS ||
+          res.error_code.val == moveit_msgs::msg::MoveItErrorCodes::UNDEFINED)
+      {
+        res.error_code.val = moveit_msgs::msg::MoveItErrorCodes::FAILURE;
+      }
+      else
+      {
+        RCLCPP_ERROR(
+            getLogger(), "Preserving planner-specific failure code %d for /move_action",
+            res.error_code.val);
+      }
     }
   }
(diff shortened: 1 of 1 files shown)


nbbrooks (approved):
thank you for the contribution!

eclipse0922:
@nbbrooks How do I execute CI?

```
Current runner version: '2.332.0'
Runner Image Provisioner
Operating System
Runner Image
GITHUB_TOKEN Permissions
Secret source: Actions
Prepare workflow directory
Prepare all required actions
Getting action download info
Error: Unable to resolve action sonarsource/sonarcloud-github-c-cpp, repository not found
```

nbbrooks:
I approved the workflows :heavy_check_mark:

eclipse0922:
The `tutorial-source (rolling)` CI failure is unrelated to this PR's changes. It fails during `rosdep install` in the Tutorial Docker Images workflow:

```
E: Unable to locate package ros-rolling-picknik-reset-fault-controller
```

`picknik_reset_fault_controller` is not available in the ROS Rolling apt repository, causing the Docker image build to fail. This is a pre-existing infrastructure issue — the same failure would occur on any PR targeting `main`.

All other checks (Formatting, CI builds for rolling/jazzy/humble, clang-tidy, codecov) are passing.

Issue #3698, opened 2026-03-06
RViz plugin fails to start when MoveIt2 is run across machines with FastDDS security enabled


AravindhDeivaG:
## Problem Description

In a typical setup using MoveIt2, `move_group`, `robot_state_publisher`, and `rviz` are launched on the same machine. 

In our use case we needed to run:

- `move_group` and `robot_state_publisher` on **Machine A**
- `rviz` on **Machine B**

Both machines are connected on the same LAN.

Our routers block multicast flooding by default, so ROS2 discovery using multicast did not work. To solve this we configured **unicast discovery using FastDDS initial peers** via a FastDDS XML profile.

With **FastDDS security disabled**, the system works correctly:
- RViz launches successfully
- MoveIt plugin loads correctly
- markers and planning interface are initialized

However, when **FastDDS security is enabled**, RViz launches but the MoveIt plugin does not initialize properly. Specifically the **markers are not initialized**.

## Root Cause Analysis

The issue originates from a service call from `MoveGroupInterface` inside the MoveIt RViz plugin.

During initialization, the interface queries the `query_planner_interface` service exposed by `move_group` to retrieve the list of available planner plugins.

When FastDDS security is enabled, additional latency is introduced due to:

1. DDS participant authentication
2. Secure endpoint matching

Because of this delay, the service client attempts to call the service before it has been discovered.

Currently there is no check to ensure the service is available before issuing the request. As a result, the request is issued prematurely and the thread blocks indefinitely while waiting for a response.

## Proposed Fix

Ensure the service client waits for the `query_planner_interface` service to become available before sending the request.

Adding a check similar to the following for all the services used in `MoveGroupInterfaceImpl` in it's constructor resolves the issue. I verified that adding this check locally resolves the issue.

```cpp
while(!service_client->wait_for_service())
{
    std::this_thread::sleep_for(std::chrono::milliseconds(100));
}
```

### ROS Distro

Humble

### OS and version

Ubuntu 22.04

### Source or binary build?

Source

### If source, which branch?

Humble

### Which RMW are you using?

FastRTPS

### Steps to Reproduce

Initial peers list fastdds profile xml file for Machine A

```
<?xml version="1.0" encoding="UTF-8" ?>
<profiles xmlns="http://www.eprosima.com/XMLSchemas/fastRTPS_Profiles"> 

    <transport_descriptors>
        <transport_descriptor>
          <transport_id>udp</transport_id>
          <type>UDPv4</type>
[...]

PR #3711, opened 2026-03-21
Fix/service availability race

AravindhDeivaG:
## Summary

Fix a race condition in `MoveGroupInterfaceImpl` initialization where service calls could be made before the corresponding service servers are available.

## Problem

During initialization, `MoveGroupInterfaceImpl` queries services such as `query_planner_interface`. In distributed setups, service discovery can be delayed due to asynchronous DDS initialization.

This can lead to a race condition where the service client attempts to call the service before it is available, resulting in incomplete initialization (e.g. missing markers in RViz) or blocking behavior.

This issue is more likely to appear in multi-machine setups and when additional discovery latency is introduced (e.g. DDS security).

For full reproduction details, see the related issue.

## Solution

Add a guard to ensure that required services are available before issuing service requests during initialization.

This is implemented by waiting for the service using `wait_for_service()` before making the call.

## Testing

- Tested in a multi-machine setup
- Confirmed that initialization is now consistent and no longer exhibits intermittent failures

## Related Issue

#3698

### DIFF
--- a/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
+++ b/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
@@ -96,7 +96,8 @@ class MoveGroupInterface::MoveGroupInterfaceImpl
 
 public:
   MoveGroupInterfaceImpl(const rclcpp::Node::SharedPtr& node, const Options& opt,
-                         const std::shared_ptr<tf2_ros::Buffer>& tf_buffer, const rclcpp::Duration& wait_for_servers)
+                         const std::shared_ptr<tf2_ros::Buffer>& tf_buffer, const rclcpp::Duration& wait_for_servers,
+                         const rclcpp::Duration wait_for_services = rclcpp::Duration(10,0))
     : opt_(opt), node_(node), tf_buffer_(tf_buffer)
   {
     // We have no control on how the passed node is getting executed. To make sure MGI is functional, we're creating
@@ -167,8 +168,8 @@ public:
     execute_action_client_->wait_for_action_server(wait_for_servers.to_chrono<std::chrono::duration<double>>());
     
     // Function to check if the service server is available
-    auto wait_for_service = [this](auto& client, const std::string& service_name) {
-        if (!client->wait_for_service(std::chrono::seconds(10)))
+    auto wait_for_service = [this](auto& client, const std::string& service_name, const rclcpp::Duration& duration) {
+        if (!client->wait_for_service(duration.to_chrono<std::chrono::duration<double>>()))
         {
             throw std::runtime_error("Service server not available for service: " + service_name + 
                                     ". Is move_group running?");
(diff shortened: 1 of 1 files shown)


moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
```diff
@@ -165,6 +165,16 @@ class MoveGroupInterface::MoveGroupInterfaceImpl
     execute_action_client_ = rclcpp_action::create_client<moveit_msgs::action::ExecuteTrajectory>(
         node_, rclcpp::names::append(opt_.move_group_namespace_, move_group::EXECUTE_ACTION_NAME), callback_group_);
     execute_action_client_->wait_for_action_server(wait_for_servers.to_chrono<std::chrono::duration<double>>());
+    
+    // Function to check if the service server is available
+    auto wait_for_service = [this](auto& client, const std::string& service_name) {
+        if (!client->wait_for_service(std::chrono::seconds(10)))
```
riv-mjohnson: We are already seeing problems where MoveIt is not tolerant enough to startup delays on large systems. I think this will cause more startup crashes on our system.

If there is going to be a hard timeout, could it at least be configurable?

AravindhDeivaG: I feel there are two three possible fixes. Either we can make the block indefinite because there is no point in proceeding if rviz plugin can't find the list of planners available. 

Secondly the action above the service has a standard 30 seconds delay which cannot be configured. We can add a similarly large delay.

Third we can declare a parameter which can be passed using yaml file as the plugin uses rviz node for all the communication and user should be able to configure it during launch

riv-mjohnson: My personal preference would be for this to use a configurable timeout, like the `wait_for_servers` timeout used in the action client above.

I'm not sure what the MoveIt maintainers would prefer though.

AravindhDeivaG: Wait for servers uses a constant 30 seconds delay. It is not configurable

riv-mjohnson: I thought `wait_for_servers` was a constructor argument of `MoveGroupInterface`, and so user-configurable?

Maybe we're talking about different waits? Where is the 30 second wait you're talking about defined?

AravindhDeivaG: I am talking in the context of moveit rviz plugin. In moveit rviz plugin the value is set to constant 30 seconds when creating the move group interface object and not left to the user to decide. I was wondering it's similar for services too as I observed this issue only in rviz plugin. In general yeah we can include another argument in constructor. That works too. But it'll mean that the rviz plugin should also be modified to give a larger delay if needed

riv-mjohnson: Ah, yeah, I understand now, thanks for the clarification.

riv-mjohnson: I think it would be useful for the plugin to have configurable waits, but I'm guessing that's out of scope for this PR.

AravindhDeivaG: So I'll add another argument to the constructor call for service wait time,  with default value of 10 seconds. Is that okay?

riv-mjohnson: Sounds sensible to me

AravindhDeivaG: Cool, thanks for the inputs.

Issue #2776, opened 2024-03-30
 MoveIt2 Setup Assistant generates integer values for joint limits when float is expected


JorgePRamos:
### Description

When generating a new MoveIt 2 package the resulting 'config/joint:limits.yaml' presents integer values when double are expected.

### Your environment
* ROS Distro: Humble
* OS Version: Ubuntu 22.04
* Source or Binary build? Binary  2.5.5-1

### Steps to reproduce
Generate a moveit package using the set up assistant.
Once generated and build proceed to launch the 'demo.launch.py '
This will prompt the error regarding the expected double values.

To fix it is enough to locate the file 'config/joint:limits.yaml' in your moveit package and chage the values from '0 or 3' to '0.0 or 3.0'
After building the package launching the 'demo.launch.py ' will work as intended

### Backtrace or Console output
`[move_group-3] terminate called after throwing an instance of 'rclcpp::exceptions::InvalidParameterTypeException'
[move_group-3]   what():  parameter 'robot_description_planning.joint_limits.art1_yaw.max_velocity' has invalid type: expected [double] got [integer]`
![image](https://github.com/ros-planning/moveit2/assets/20907919/0ff6b66d-cf1b-4e50-a4b6-5e0a7b47e621)

JorgePRamos:
I would really like to contribute to its fix. 
The code is daunting but I think that the error may be in the '.joint_limits.yaml' file composition, so maybe the file 'time_optimal_trajectory_generation.cpp'

jaredgonzales:
I'm also having this issue. While it would be nice to accept integers, a different fix could be generating the numbers as doubles from MoveIt Setup Assistant. I believe that is defined in `SRDFConfig::GeneratedJointLimits::writeYaml` inside of the file `moveit_setup_assistant/moveit_setup_framework/src/srdf_config.cpp`

sjahr:
Closed with https://github.com/moveit/moveit2/pull/2803

PR #3716, opened 2026-03-30
Fix joint_limits.yaml generating integer values for double parameters

nkulis:
### Description

When `moveit_setup_assistant` generates `joint_limits.yaml`, whole-number limits like `1.0` are written as `1`. This causes `move_group` to crash on startup with:
```
parameter 'robot_description_planning.joint_limits.joint1.max_velocity' has invalid type: expected [double] got [integer]
```

This was previously reported in #2776 and previous #2803 attempted to fix this with `static_cast<double>`, which doesn't affect yaml-cpp's output.

The underlying issue is that yaml-cpp saves C++'s `double` value like `1.0` as `1`, which is valid per YAML spec. It is then read as integer by  [`launch_param_builder`](https://github.com/PickNikRobotics/launch_param_builder/blob/main/launch_param_builder/utils.py#L74). So `1` would be both a `int` and `float`, it is just that it's matched as `int` first.

This fix is clearly a workaround and yaml-cpp already has a feature to enforce adding trailing zeros (jbeder/yaml-cpp#1407), but yet not available in recently released version.

An alternative solution would be to enfore the type in   `joint_limits.yaml` with `!!float` [tags](https://yaml.org/spec/1.2.2/#10214-floating-point), where `!!float 1` would be matched as `float`.

### Checklist

### DIFF
--- a/moveit_setup_assistant/moveit_setup_framework/src/srdf_config.cpp
+++ b/moveit_setup_assistant/moveit_setup_framework/src/srdf_config.cpp
@@ -35,6 +35,7 @@
 #include <moveit_setup_framework/data_warehouse.hpp>
 #include <moveit_setup_framework/utilities.hpp>
 #include <moveit/rdf_loader/rdf_loader.hpp>
+#include <sstream>
 
 namespace moveit_setup
 {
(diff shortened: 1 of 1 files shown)


moveit_setup_assistant/moveit_setup_framework/src/srdf_config.cpp
```diff
@@ -235,6 +236,17 @@ struct JointModelCompare
 
 bool SRDFConfig::GeneratedJointLimits::writeYaml(YAML::Emitter& emitter)
 {
+  // yaml-cpp serializes whole-number doubles (e.g. 1.0) as integers ("1"), which ROS 2 parameter
+  // loading rejects with "expected [double] got [integer]". Format them explicitly with a decimal point.
```
Timple, line 240: I would add a link to the upstream issue/fix such that it's clear when this can be cleaned up.

nkulis:
@Timple Thanks for taking a look. I've added link to yaml-cpp's fix in comment. Once they publish new release a proper fix can be applied.

PR #3727, opened 2026-05-19
CMake: remove boost system

nim65s:
Hi,

This fix for boost >= 1.89:
```cmake
CMake Error at /nix/store/4qaj1kkf1p3al8z190px12ma6r8jbgqj-boost-1.89.0-dev/lib/cmake/Boost-1.89.0/BoostConfig.cmake:141 (find_package):
  Could not find a package configuration file provided by "boost_system"
  (requested version 1.89.0) with any of the following names:

    boost_systemConfig.cmake
    boost_system-config.cmake

  Add the installation prefix of "boost_system" to CMAKE_PREFIX_PATH or set
  "boost_system_DIR" to a directory containing one of the above files.  If
  "boost_system" provides a separate development package or SDK, be sure it
  has been installed.
  /nix/store/4qaj1kkf1p3al8z190px12ma6r8jbgqj-boost-1.89.0-dev/lib/cmake/Boost-1.89.0/BoostConfig.cmake:262 (boost_find_component)
  /nix/store/v2i1hgv567g3v91im5x4g5bff52143i0-cmake-4.1.2/share/cmake-4.1/Modules/FindBoost.cmake:609 (find_package)
  ConfigExtras.cmake:3 (find_package)
  CMakeLists.txt:43 (include)
```

boost system has been header only since release 1.69 (December 5, 2018), and a useless stub was provided instead.

in 1.89 (August 6, 2025), that stub was removed, so CMake fail if we require it.

ref. https://www.boost.org/doc/libs/latest/libs/system/doc/html/system.html#changes_in_boost_1_89

### DIFF
--- a/moveit_core/ConfigExtras.cmake
+++ b/moveit_core/ConfigExtras.cmake
@@ -10,5 +10,4 @@ find_package(
   program_options
   regex
   serialization
-  system
   thread)
--- a/moveit_kinematics/ConfigExtras.cmake
+++ b/moveit_kinematics/ConfigExtras.cmake
@@ -1,3 +1,3 @@
 # Extras module needed for dependencies to find boost components
 
-find_package(Boost REQUIRED program_options system)
+find_package(Boost REQUIRED program_options)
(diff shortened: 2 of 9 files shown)


nbbrooks (approved):
Thank you for helping with the Resolute migration!

PR #3730, opened 2026-05-20
fix(moveit_cpp): load planning scene monitor params

### LINKED ISSUE #3729: PlanningSceneMonitorOptions:load(): half of the options are ignored
### Description

The PlanningSceneMonitorOptions:load() does load just: name, robot_description and wait_for_initial_state_timeout. The rese of the options joint_state_topic, attached_collision_object_topic, monitored_planning_scene_topic and publish_planning_scene_topic is ignored.

### ROS Distro

Jazzy

### OS and version

Ubuntu 

### Source or binary build?

Source

_No response_

### If source, which branch?

main

### Which RMW are you using?

None

### Steps to Reproduce

Just try to change the joint_state_topic.

### Expected behavior

Expected behaviour is that the options are not ignored.

### Actual behavior

Options are ignored.

_No response_

liborw:
Load all planning_scene_monitor_options values from node parameters so MoveItPy config_dict can override joint_state_topic and related settings. Fixes #3729.

### DIFF
--- a/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.hpp
+++ b/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.hpp
@@ -78,14 +78,21 @@ public:
       const std::string ns = "planning_scene_monitor_options";
       node->get_parameter_or(ns + ".name", name, std::string("planning_scene_monitor"));
       node->get_parameter_or(ns + ".robot_description", robot_description, std::string("robot_description"));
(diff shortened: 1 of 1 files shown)


peci1 (approved):
LGTM (but I haven't run-tested this).

One possible simplification is to remove the initializers of the member vars from the constructor and pass the default values to get_param_or.

galou:
My opinion is that  topic names should not be passed as parameters but topics should be remapped from their default hard-coded value.

Isn't remapping possible? If not, then OK but if yes, parameters just add confusion.

Issue #3744, opened 2026-06-02
MotionPlanning RViz plugin uses Qt5-era QComboBox QString signals, breaking named-state selection under Qt6  Body:


andrearosasco:
The MoveIt MotionPlanning RViz plugin appears to connect `QComboBox` signals using Qt5-era string signatures that are not valid under Qt6. In a Qt6 RViz environment, selecting named Start/Goal states in the MotionPlanning panel does not call the corresponding MoveIt handlers, so named SRDF states such as `home` are displayed in the combo box but are not applied to the query robot state.

This affects the MotionPlanning RViz plugin in `moveit_ros_visualization`.

## Affected component

* Repository: `moveit/moveit2`
* Package: `moveit_ros_visualization`
* Plugin: `moveit_rviz_plugin/MotionPlanning`
* File: `moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp`

## Environment where reproduced

* ROS distro/package channel: RoboStack Lyrical
* `ros-lyrical-rviz2`: `15.2.3`
* `ros-lyrical-moveit-ros-visualization`: `2.14.1`
* Qt runtime used by RViz and the MoveIt RViz plugin: Qt6

Local package metadata shows:

```text
ros-lyrical-rviz2 15.2.3 depends on qt6-main >=6.11.1,<6.12
ros-lyrical-moveit-ros-visualization 2.14.1 depends on ros-lyrical-rviz2
```

`ldd` confirms that both `rviz2` and the MoveIt MotionPlanning RViz plugin resolve Qt6 libraries in this environment:

```text
libQt6Widgets.so.6
libQt6Gui.so.6
libQt6Core.so.6
```

## Symptoms

In RViz, the MoveIt MotionPlanning panel starts and displays named SRDF states, but selecting a named state from the Start State or Goal State combo box does not apply that state.

Observed behavior:

* Selecting `home` in the Goal State dropdown does not move the orange query goal robot to the SRDF `home` state.
* The goal interactive marker then behaves as if it is seeded from the previous/stale query goal state.
* On a 6-axis manipulator this can change the IK branch selected while dragging the marker, for example preferring an elbow-down solution instead of the previous elbow-up/home-biased solution.

## Suspected root cause

`motion_planning_frame.cpp` connects `QComboBox` signals using string signatures that are not valid Qt6 signal names at runtime:

```cpp
[...]

PR #3745, opened 2026-06-02
Fix MotionPlanning combo box signals for Qt6

andrearosasco:
## Summary

Fix the MotionPlanning RViz plugin's named-state combo box connections when built against Qt6.

Qt6 no longer exposes the old text-carrying `QComboBox` signal names used by the plugin's string-based connections:

- `activated(QString)`
- `currentIndexChanged(QString)`

In a Qt6 RViz environment this produces runtime warnings such as:

```text
QObject::connect: No such signal QComboBox::activated(QString)
QObject::connect: No such signal QComboBox::currentIndexChanged(QString)
```

When those connections fail, selecting a named Start State or Goal State in the MotionPlanning panel does not invoke the corresponding handler, so selecting a named goal such as `home` does not update the query goal state.

For more details check #3744 

## Changes

- Use `QComboBox::textActivated` for Start State and Goal State under Qt6.
- Use `QComboBox::currentTextChanged` for Planning Group under Qt6.
- Keep the existing Qt5 string-signal connections for Qt5 builds.

## Testing

Tested downstream in a RoboStack Lyrical Qt6 RViz environment by rebuilding `moveit_ros_visualization` with this patch and verifying that the MotionPlanning panel no longer logs the missing `QComboBox` signal warnings and named goal-state selection works again.


simonschmeisser:
Doesn't the "new" style connect also work for Qt5? It would be good to drop the legacy `SIGNAL` `SLOT` code from Qt4 completely

rhaschke:
Fix via #3823.

Issue #3411, opened 2025-04-02
Moveit2 Servo Always giving "Very close to a singularity, emergency stop"


CanavarB:
### Description

Hi, I have a 4DOF robot which I designed. I want use realtime servo on this robot. I launched moveit servo with my config. For the first try I am publishing some twist commands for servo node but every time it says "Very close to a singularity, emergency stop". I can control my robot with "normal" moveit (approx IK solition checked). I changed many times servo config yaml for singularity thresholds.

### ROS Distro

Jazzy

### OS and version

24.04.2

### Source or binary build?

Source

### If binary, which release version?

From ROS repo

### Which RMW are you using?

None

### Steps to Reproduce

Normal Ubuntu Desktop.

### Expected behavior

Every time I get "Very close to a singularity, emergency stop" warning.

### Actual behavior

.

### Backtrace or Console output

[servo_node-10] [WARN] [1743601979.331145654] [servo_node]: Command type has not been set, cannot accept input
[servo_node-10] [WARN] [1743601979.431078677] [servo_node]: Command type has not been set, cannot accept input
[servo_node-10] [WARN] [1743601979.531184001] [servo_node]: Command type has not been set, cannot accept input
[servo_node-10] [WARN] [1743601979.631153174] [servo_node]: Command type has not been set, cannot accept input
[servo_node-10] [WARN] [1743601979.731071291] [servo_node]: Command type has not been set, cannot accept input
[servo_node-10] [WARN] [1743601979.831163947] [servo_node]: Command type has not been set, cannot accept input
[servo_node-10] [WARN] [1743601979.931170629] [servo_node]: Command type has not been set, cannot accept input
[servo_node-10] [WARN] [1743601980.031155456] [servo_node]: Command type has not been set, cannot accept input
[servo_node-10] [WARN] [1743601980.131110808] [servo_node]: Command type has not been set, cannot accept input
[servo_node-10] [WARN] [1743601980.231149954] [servo_node]: Command type has not been set, cannot accept input
[servo_node-10] [WARN] [1743601980.331077102] [servo_node]: Command type has not been set, cannot accept input
[servo_node-10] [WARN] [1743601980.431096715] [servo_node]: Command type has not been set, cannot accept input
[servo_node-10] [WARN] [1743601980.531079188] [servo_node]: Command type has not been set, cannot accept input
[servo_node-10] [WARN] [1743601986.550210444] [servo_node.servo_node.moveit.ros.servo]: Very close to a singularity, emergency stop
[...]

Issue #2760, opened 2024-03-20
moveit_servo /servo_node/delta_twist_cmds only giving singularity errors 


PreetDhami:
### Description
I'm trying to adapt moveit_servo to a custom robot arm as part of a university capstone project. I have managed to be able to successfully send `/servo_node/delta_joint_cmds` to control the arm, but `/servo_node/delta_twist_cmds` keeps giving singularity errors no matter the input.

### Your environment
* ROS Distro: Humble
* OS Version: e.g. Ubuntu 22.04

### Steps to reproduce
My source code is all located [here](https://github.com/PreetDhami/rover_arm_capstone_2024)
You should just be able to clone the code to a docker container, then build with colcon. To run, use:
```
ros2 launch rover_arm arm.launch.py
```
Then in a second terminal, publish a message such as this one below, but you can also play around with the values for the twist command:

```
ros2 topic pub --rate 10 --print 30 /servo_node/delta_twist_cmds geometry_msgs/msg/TwistStamped "{header: {frame_id: base_link, stamp: now}, twist: {linear: {x: 0.0, y: 0.0, z: -1.0}, angular: {x: 0.0, y: 0.0, z: 0.0}}}"
```

### Expected behaviour
`/servo_node` should publish some non-zero positions and velocitys to `/rover_arm_controller/joint_tragectory` when it receives a message on the `/servo_node/delta_twist_cmds` topic

### Actual behaviour
Any non-zero input for `/servo_node/delta_twist_cmds` leads to `/rover_arm_controller/joint_tragectory` recieving 'nan' values. Also I get the following warning messages:
![image](https://github.com/ros-planning/moveit2/assets/108783637/787b0c13-466e-4b77-8c96-5a2a94eaad06)

Whenever a topic is published to `/servo_node/delta_joint_cmds`, the code behaves as expected which can be checked by subscribing to the `/rover_arm_controller/joint_tragectory` topic. Publish a message such as this one to check:
```
ros2 topic pub --rate 10 --print 30 /servo_node/delta_joint_cmds control_msgs/msg/JointJog "{header: {frame_id: base_joint, stamp: now}, joint_names: [wrist_roll_joint], velocities: [1.0]}"
```

I believe that this is an issue with my config in some way. But I really don't know what the issue is.

Flamethr0wer:
Hey @PreetDhami , did you manage to solve this?

PreetDhami:
[...]

PR #3750, opened 2026-06-12
Fix out-of-bounds SVD indexing in Servo singularity check for robots with fewer than 6 joints

EmirataG:
### Description

`velocityScalingFactorForSingularity()` in `moveit_servo` indexes the SVD of the robot Jacobian with the Cartesian task dimension instead of the number of singular values. The Jacobian is `6 x num_joints`, and the thin SVD (`ComputeThinU`) therefore has `min(6, num_joints)` singular values and U columns, but the code reads `singularValues()(dims - 1)` and `matrixU().col(dims - 1)` with `dims = 6` always. For any move group with fewer than 6 active joints, both accesses are out of bounds. Because release builds define `NDEBUG`, Eigen's bounds assertions are compiled out and the reads silently return garbage memory: the computed "condition number" is sigma_max divided by an arbitrary value, which is almost always far above `hard_stop_singularity_threshold`, so Servo emits `HALT_FOR_SINGULARITY` ("Very close to a singularity, emergency stop") on effectively every TWIST/POSE command, independent of the configured thresholds. The corrupted last-U-column also breaks the moving-toward/away-from-singularity classification that selects the deceleration band.

This explains several reports that were closed without the root cause being found: #3411 (4-DOF arm, constant emergency stops; the reporter worked around it by adding two fake virtual joints to the URDF, which "fixes" the issue only by making the out-of-bounds index legal), #2760, and ROBOTIS-GIT/open_manipulator#192. It is related to but independent of the drift-dimensions discussion in moveit/moveit_msgs#185: even with structurally large condition numbers, users of low-DOF arms can tune thresholds — but only if the condition number is actually computed from their robot's geometry rather than from undefined behavior.

How this was found: a 4-DOF robot emergency-stopped on every POSE/TWIST command under Jazzy. The decisive symptom was that changing `lower_singularity_threshold` / `hard_stop_singularity_threshold` had *no effect whatsoever* on the behavior, which is impossible if the halting decision compares a geometry-derived condition number against those thresholds, and pointed to the comparison itself being broken. Reading `velocityScalingFactorForSingularity()` with the robot's 6x4 Jacobian in mind showed the thin SVD is indexed by the task dimension (6) rather than the number of singular values (4). With the index fixed, the computed condition numbers match an independent numerical model of the arm to four significant figures.

The fix indexes by `singularValues().size() - 1`, which is the position of sigma_min for any number of joints (singular values are sorted in decreasing order), and uses the corresponding last column of thin U as the least-responsive direction. For robots with 6 or more joints in the group this evaluates to the same index 5 as before, so behavior is unchanged for existing 6/7-DOF users.

A unit test is added that builds a 4-DOF chain with `RobotModelBuilder` and calls `velocityScalingFactorForSingularity()` directly, asserting both branches: a well-conditioned bent posture (condition number ~33, verified against an independent numerical model of the same kinematics) must return `NO_WARNING` with scale 1.0, and a near-extended posture (condition number ~979) must return `HALT_FOR_SINGULARITY` with scale 0.0, with at least 2x margin between every condition number and its threshold. The test was verified to fail against the stock, unpatched 2.12.4 release library (it spuriously halts at the well-conditioned posture) — no sanitizer needed; additionally, ASan and debug builds (`eigen_assert`) catch the out-of-bounds read itself deterministically. The existing Panda-based singularity tests pass unchanged.

Verified on real hardware: the same 4-DOF arm that previously emergency-stopped on every pose command now servos correctly with thresholds tuned to its structural condition-number range.

### Checklist
- [x] **Required by CI**: Code is auto formatted using [clang-format](http://contributing/code)
- [x] Extend the tutorials / documentation [reference](http://moveit.ros.org/documentation/contributing/) (not applicable: bug fix, no API or behavior change for supported
configurations)
- [x] Document API changes relevant to the user in the [MIGRATION.md](https://github.com/moveit/moveit2/blob/main/MIGRATION.md) notes (not applicable: no API change)
- [x] Create tests, which fail without this PR [reference](https://moveit.pitests/tests_tutorial.html)
- [x] Include a screenshot if changing a GUI (not applicable)
- [x] While waiting for someone to review your request, please help review [tps://github.com/moveit/moveit2/pulls) to support the maintainers

### DIFF
--- a/moveit_ros/moveit_servo/src/utils/common.cpp
+++ b/moveit_ros/moveit_servo/src/utils/common.cpp
@@ -294,9 +294,6 @@ std::pair<double, StatusCode> velocityScalingFactorForSingularity(const moveit::
   const double hard_stop_singularity_threshold = servo_params.hard_stop_singularity_threshold;
   const double leaving_singularity_threshold_multiplier = servo_params.leaving_singularity_threshold_multiplier;
 
-  // Get size of total controllable dimensions.
-  const size_t dims = target_delta_x.size();
-
   // Get the current Jacobian and compute SVD
   const Eigen::JacobiSVD<Eigen::MatrixXd> current_svd = Eigen::JacobiSVD<Eigen::MatrixXd>(
       robot_state->getJacobian(joint_model_group), Eigen::ComputeThinU | Eigen::ComputeThinV);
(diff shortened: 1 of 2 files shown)


gsokoll:
We hit this independently on a 3-DOF planar group (Jazzy, 2.12.4 release binaries) and can confirm the mechanism: `dims = target_delta_x.size()` is always 6, but the thin SVD of the 6×3 Jacobian has only 3 singular values and a 6×3 U, so `singularValues()(dims - 1)` and `matrixU().col(dims - 1)` read out of bounds. Under NDEBUG that produces garbage condition numbers, and NaN fails both threshold comparisons and falls into the halt branch — so spurious `HALT_FOR_SINGULARITY` can fire regardless of threshold config. Plausibly also the cause of #3411.

PR #3751, opened 2026-06-15
Add distro compatibility policy

nbbrooks:
## Description

Document the in-source patterns and process moveit2 uses to keep `main` building on every active ROS 2 distro, plus the lifecycle for adding and removing distros from the CI matrix. Three topics:

1. **Supporting multiple distros on `main`** — five sub-patterns covering C++ source divergence (`RCLCPP_VERSION_GTE`), `package.xml` conditional deps, launch-loaded file overrides, CMake feature detection, and `MIGRATION.md` updates.
2. **Handling dependencies without a bloom package** — CI source-build via per-distro `.repos` files (preferred), with a CMake-side optional feature-flag fallback.
3. **CI matrix policy** — add / graduate / drop lifecycle.

Each pattern carries a cleanup rule keyed on a distro EOL trigger, so the conditional code paths don't accumulate indefinitely.

### Why now

We're accumulating multi-distro divergences in `main` without a written convention. Recent precedents that motivated this:

- [#3567](https://github.com/moveit/moveit2/pull/3567) uses the `RCLCPP_VERSION_GTE` pattern for C++ source divergence.
- [moveit/moveit_resources#211](https://github.com/moveit/moveit_resources/pull/211) introduces format-3 `condition=` deps and a launch-time YAML override helper.

Capturing these as a single documented policy lets future divergences land consistently and gives reviewers a checklist.

### Checklist

- [x] Code is auto-formatted (documentation-only)
- [ ] Tutorials/documentation — not applicable; this is a contributor doc
- [ ] MIGRATION.md — not applicable; this is the policy that *requires* future MIGRATION.md entries
- [ ] Tests — not applicable
- [ ] GUI screenshots — not applicable

### DIFF
--- a/.github/DISTRO_COMPAT_POLICY.md
+++ b/.github/DISTRO_COMPAT_POLICY.md
@@ -254,3 +254,45 @@ When a distro goes EOL per [REP 2000](https://www.ros.org/reps/rep-2000.html), o
 5. Adds a MIGRATION.md entry noting the support drop (§1.5).
 
 **Where the policy lives**: this document plus the matrix file. Don't fork the support set into a separate "supported distros" list anywhere else — the matrix is the truth.
+
+### Per-distro branches
+
+Each non-EOL ROS distro has a corresponding long-lived branch on `moveit/moveit2` (`humble`, `jazzy`, `kilted`, ...). These branches are **stabilization-only**: bug fix backports from `main`, and nothing else. They are the basis for `catkin_prepare_release` + `bloom-release` for that distro's stream.
+
+**Workflow file presence** — only this minimal set lives on per-distro branches; everything else is `main`-only:
+
+| Workflow                | main | per-distro branch |
+|-------------------------|:---:|:---:|
+| `ci.yaml`               | ✓   | ✓ |
+| `docker.yaml`           | ✓   | ✓ |
+| `docker_lint.yaml`      | ✓   | ✓ |
+| `format.yaml`           | ✓   | ✓ |
+| `prerelease.yaml`       | ✓   | — |
+| `sonar.yaml`            | ✓   | — |
+| `stale.yaml`            | ✓   | — |
+| `tutorial_docker.yaml`  | ✓   | — |
+
+Rationale: per-distro branches don't have the breadth of churn `main` does, so the tutorial/prerelease/sonar/stale machinery is wasted overhead. Anything found by those workflows on a per-distro branch is also found on `main` first.
+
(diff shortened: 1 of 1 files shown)


.github/DISTRO_COMPAT_POLICY.md
```diff
@@ -0,0 +1,256 @@
...
+
+**PR review checklist**: reviewers reject any PR that introduces a distro-conditional change without a MIGRATION.md entry.
+
+**Cleanup**: distro release-note entries stay (they're historical). *Workaround* entries (e.g. "Remove when Resolute syncs osqp_vendor") get retired together with the workaround code.
+
+---
+
+## Topic 2: Handling dependencies without a bloom package
+
+When a moveit2 dependency is **released** for a distro (entry exists in `rolling/distribution.yaml`) but **not yet bloomed** to the apt binary repo, CI fails at `rosdep install`. Two complementary approaches handle this. **Prefer §2.1.**
```
rhaschke, line 187: When a package exists in rosdistro, it was bloomed. However, it might exist in ros-testing only, because it requires a sync. I suggest dropping the mention of bloom (here and below).
```suggestion
When a moveit2 dependency is not yet released for a distro, CI will fail at `rosdep install`. Two complementary approaches handle this. **Prefer §2.1.**
```

.github/DISTRO_COMPAT_POLICY.md
```diff
@@ -0,0 +1,256 @@
...
+
+When a dep is genuinely missing and the consuming code can degrade gracefully, use `QUIET` + a presence check. Always pair the optional `find_package` with a `target_compile_definitions` feature flag so source code can `#ifdef`-guard usages.
+
+```cmake
+# osqp_vendor not yet bloomed for Resolute; degrade gracefully when absent.
+find_package(osqp QUIET)
+if(osqp_FOUND)
+  add_subdirectory(online_signal_smoothing)
+  target_compile_definitions(my_target PRIVATE MOVEIT_HAVE_OSQP)
+endif()
```
rhaschke, line 242: Add an `else` branch to inform users about the omitted feature (as required below).
```suggestion
if(osqp_FOUND)
  add_subdirectory(online_signal_smoothing)
  target_compile_definitions(my_target PRIVATE MOVEIT_HAVE_OSQP)
else()
  message(STATUS "Dropping online_signal_smoothing as osqp wasn't found")
endif()
```

rhaschke (commented):
Thanks for explicitly writing that down. I only have a few minor comments:
- Place the file in doc/

PR #3760, opened 2026-06-22
resolute: GCC 15 / C++20 / rviz API build fixes, and octomap -Werror=cpp suppression

### LINKED ISSUE #3717: Lyrical Luth/Rolling Ament changes and cmakelist incompatibility.
### Description

There's a new ROS2 Rolling update which breaks Moveit2 due to Ament now being obsolete and new switch to target_link_libraries breaking the release build of moveit2.

### ROS Distro

Rolling

### OS and version

Ubuntu 24.04

### Source or binary build?

Source

_No response_

### If source, which branch?

main/release/rolling

### Which RMW are you using?

None

### Steps to Reproduce

Update to latest ROS2 Rolling using apt, then try to rebuild moveit2, almost every single cmakelist breaks due to ament being broken.
Example:
--- stderr: rviz_visual_tools                                    
CMake Error at CMakeLists.txt:243 (target_link_libraries):
  The plain signature for target_link_libraries has already been used with
  the target "rviz_visual_tools_rvt_test".  All uses of target_link_libraries
  with a target must be either all-keyword or all-plain.

  The uses of the plain signature are here:

   * /opt/ros/rolling/share/ament_cmake_gtest/cmake/ament_add_gtest_executable.cmake:53 (target_link_libraries)
   * /opt/ros/rolling/share/ament_cmake_gtest/cmake/ament_add_gtest_executable.cmake:55 (target_link_libraries)
   * /opt/ros/rolling/share/ament_cmake_gtest/cmake/ament_add_gtest_executable.cmake:59 (target_link_libraries)

### Expected behavior

It should build with no cmake errors.

### Actual behavior

It errors out on most moveit2 packages.

_No response_

nbbrooks:
### Description

Five Resolute-only build failures in moveit2 itself, plus one CI suppression. All found by an end-to-end source build in an `ubuntu:resolute` container.

> **Rescoped 2026-08-03.** This PR originally also introduced `moveit2_rolling.repos` and a `CMAKE_POLICY_VERSION_MINIMUM=3.5` override. Both are now obsolete — see *What was dropped* below. The remaining changes are unaffected.

| # | Change | Why |
|---|---|---|
| 1 | `PlanningContextBase<GeneratorT>(...)` → `PlanningContextBase(...)` | Redundant template-id on a constructor. GCC 15 rejects the injected-class-name form. |
| 2 | Replace `[=]` with explicit captures in rviz lambdas | Implicit `this` capture via `[=]` is deprecated in C++20. |
| 3 | Dual-signature `Display::update` for rviz_common 16.0.1+ | rviz changed `(float, float)` → `(std::chrono::nanoseconds, std::chrono::nanoseconds)`. The old override stops being a concrete `Display`. |
| 4 | Install pilz `test_utils` | Installed test binaries can't find it on Resolute otherwise. |
| 5 | Skip trajectory_cache `test_utils` on Resolute | Unblocks the `rolling-resolute` job. |
| 6 | `SUPPRESS_CPP_ERROR: true` on the `rolling-resolute` matrix entry | See below. |

### Why `SUPPRESS_CPP_ERROR` is still needed

System `liboctomap-dev` on Resolute still includes `<ciso646>` at `OcTreeKey.h:40`. That header was removed in C++20, GCC 15 warns on it, and `-Werror=cpp` makes it fatal — so `moveit_core` fails to build.

The upstream fix ([OctoMap/octomap#434](https://github.com/OctoMap/octomap/issues/434)) is **merged and closed**, but Ubuntu hasn't repackaged. Verified against `liboctomap-dev 1.10.0+dfsg-3` on `ubuntu:resolute`, which still ships the unfixed header:

```
/usr/include/octomap/OcTreeKey.h:40:#include <ciso646>
```

Scoped to the `rolling-resolute` entry only — jammy and noble jobs keep `-Werror=cpp`. Remove once Ubuntu picks up the fix.

### What was dropped in the rescope

**`moveit2_rolling.repos`** — `main` now has its own, and it's better: it points at `moveit/osqp_vendor` @ `58a52e8` (OSQP **v1.0.0**) rather than this PR's `tier4/osqp_vendor` @ `0.2.0` (OSQP v0.6.x). That file even carries a note anticipating this conflict. Keeping this PR's version would have been a regression.

**`DOCKER_RUN_OPTS: '-e CMAKE_POLICY_VERSION_MINIMUM=3.5'`** — existed only because OSQP **v0.6.2** declares `cmake_minimum_required(3.2)`, which CMake 4 on Resolute rejects. OSQP v1.0.0 declares 3.16, so the override is dead weight that would mask genuine failures.

Both were consequences of #3806 landing the OSQP v1.0 migration after this PR was written.

### Note on validating this

None of these can be proven green by moveit2 CI today. Every *blocking* job runs on jammy (GCC 11) or noble (GCC 13); GCC 15 exists only on Resolute. And `rolling-resolute (experimental)` currently dies at `install_target_dependencies` — `ros-rolling-warehouse-ros` is not published for Resolute — so it never reaches compilation. The evidence is the container build, not a check mark.

These changes are inert on jammy/noble, so the risk to existing coverage is nil.

### Checklist
- [x] **Required by CI**: Code is auto formatted — pre-commit clean
- [ ] Extend the tutorials / documentation — not applicable
- [ ] Document API changes in [MIGRATION.md](https://github.com/moveit/moveit2/blob/main/MIGRATION.md) — not applicable, no public API change
- [ ] Create tests, which fail without this PR — not applicable; these are build fixes for a platform CI cannot yet reach
- [ ] Include a screenshot if changing a GUI — not applicable

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -36,11 +36,18 @@ jobs:
           # industrial_ci build on ubuntu:resolute, installing Rolling from the ros-testing repo.
           # Setting OS_CODE_NAME=resolute also avoids industrial_ci's default of pinning Rolling to
           # the latest Noble-supported index. Non-blocking until all Resolute packages are released.
+          #
+          # SUPPRESS_CPP_ERROR drops -Werror=cpp so the deprecated <ciso646> include in
+          # system liboctomap-dev (OcTreeKey.h:40) does not fatal-error under GCC 15, which
+          # is what Resolute ships. The upstream fix (OctoMap/octomap#434) is merged, but
+          # Resolute still packages the unfixed header -- verified against
+          # liboctomap-dev 1.10.0+dfsg-3. Remove once Ubuntu repackages it.
           - IMAGE: rolling-resolute
             ROS_DISTRO: rolling
             OS_CODE_NAME: resolute
             ROS_REPO: testing
             SUPPRESS_DEPRECATION_ERROR: true
+            SUPPRESS_CPP_ERROR: true
           - IMAGE: humble-ci
             ROS_DISTRO: humble
           - IMAGE: jazzy-ci
--- a/moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/planning_context_base.hpp
+++ b/moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/planning_context_base.hpp
@@ -56,9 +56,8 @@ template <typename GeneratorT>
 class PlanningContextBase : public planning_interface::PlanningContext
 {
 public:
-  PlanningContextBase<GeneratorT>(const std::string& name, const std::string& group,
-                                  const moveit::core::RobotModelConstPtr& model,
-                                  const pilz_industrial_motion_planner::LimitsContainer& limits)
+  PlanningContextBase(const std::string& name, const std::string& group, const moveit::core::RobotModelConstPtr& model,
+                      const pilz_industrial_motion_planner::LimitsContainer& limits)
     : planning_interface::PlanningContext(name, group)
     , terminated_(false)
     , model_(model)
--- a/moveit_planners/pilz_industrial_motion_planner/test/CMakeLists.txt
+++ b/moveit_planners/pilz_industrial_motion_planner/test/CMakeLists.txt
@@ -33,6 +33,13 @@ target_link_libraries(
   ${orocos_kdl_LIBRARIES}
   Boost::headers)
 
+# Install to lib/ (on AMENT_PREFIX_PATH → LD_LIBRARY_PATH) so the installed
+# unit-test binaries resolve it on Resolute, where they otherwise fail at load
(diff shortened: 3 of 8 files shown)


urfeex:
I can confirm that this builds and works on Ubuntu Resolute with ROS Rolling :-) I don't necessarily feel competent enough to properly review this since this seems to touch a lot of internals that I am not familiar with.

nbbrooks:
Dang, looks like we can't use osqp 1.0 without some non-trivial API migartion. OSQP v1.0 has a real API redesign, and `moveit_core/online_signal_smoothing/acceleration_filter.hpp` uses the OSQP C API directly (OSQPWorkspace*, c_int — both renamed/removed in v1.0).

```
error: cannot convert 'OSQPWorkspace*' to 'OSQPSolver*'
error: 'c_int' was not declared in this scope; did you mean 'u_int'?
```

Will handle in followup PR, this will use a workaround

PR #3776, opened 2026-06-24
Fix colliding `ccache-rolling-ci` keys in CI

nbbrooks:
### Description

Fix long standing issue where both the `rolling + clang-tidy` and `rolling + deprecation` jobs use [read: clobber] the same `ccache-rolling-ci` key.

This splits them into dedicated keys and should significantly speed up rolling job CI times.

### Checklist
- [ ] **Required by CI**: Code is auto formatted using [clang-format](https://moveit.ai/documentation/contributing/pullrequests/)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -138,7 +138,10 @@ jobs:
           restore-keys: ${{ env.CACHE_PREFIX }}
         env:
           GHA_CACHE_SAVE: always
-          CACHE_PREFIX: target_ws${{ matrix.env.CCOV && '-ccov' || '' }}-${{ matrix.env.IMAGE }}-${{ hashFiles('**/CMakeLists.txt', '**/package.xml', '.github/workflows/ci.yaml') }}
+          # The variant token (ccov/tidy/default) keeps caches built with different compilers/flags
+          # in separate key namespaces. Without it, the clang (clang-tidy) and gcc (deprecation/plain)
+          # rolling jobs share one key and restore each other's build tree, forcing a full rebuild.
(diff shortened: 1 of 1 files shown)


nbbrooks:
Benchmarked the cache split. Measuring **colcon build-phase time only** (from the `Summary: N packages finished` lines for upstream + target builds), since the test phase that follows is unaffected by ccache and any wall-clock difference there would be flake-driven noise.

| Job | Baseline (main, old shared key) | 1st run on this PR (new keys, cold migration) | 2nd run on this PR (new keys, fully warm) |
|---|---|---|---|
| `rolling-ci + ccov` | 49m 13s | 47m 16s | **1m 45s** |
| `rolling-ci + clang-tidy (delta)` | 41m 24s | 40m 50s | **1m 35s** |
| `rolling-ci + deprecation check` | 35m 37s | 36m 35s | **1m 08s** |

What each column tells us:

- **1st run vs baseline**: ~equal. The key migration is free — no penalty for switching.
- **2nd run on same SHA**: each job's dedicated cache fully restores its own compile artifacts. Confirms the split keys work in isolation (no warm-up regressions).

A caveat on interpretation: this same-SHA re-run would also be fast under the *old* shared-key setup, but only for whichever job won the write race last time. The actual payoff of the split — that the next PR's 1st CI run can restore all three caches independently instead of seeing two clobbered ones — will only be measurable after this lands on main and a subsequent PR runs against it.

So: a no-cost change that unlocks a hit-rate improvement which becomes observable on future PRs.

---

(Failures on this PR are pre-existing flakes — `_with_move_group` family + osqp_vendor on Resolute + Sonar setup — none caused by the cache change.)

PR #3179, opened 2024-12-24
Update controller_manager_plugin.cpp

wizard-coder:
Fixing the bug where the namespace is not properly applied when using Ros2ControlMultiManager

### Description

When using two ROS2 controls, the namespace is automatically detected, but the namespace is not properly applied when registering the action client.

For more details, please refer to https://github.com/moveit/moveit2/issues/3150

PR #3815, opened 2026-08-13
fix build on rolling and lyrical

christian-rauch:
### Description

Some fixes for missing headers and API changes to make this build on `rolling` and `lyrical`.

This requires the latest unreleased version of `moveit_msgs` due to updates to the message definitions (https://github.com/moveit/moveit_msgs/issues/191).

### Checklist
- [x] **Required by CI**: Code is auto formatted using [clang-format](https://moveit.ai/documentation/contributing/pullrequests/)


moveit_planners/test_configs/prbt_ikfast_manipulator_plugin/src/prbt_manipulator_ikfast_moveit_plugin.cpp
```diff
@@ -54,6 +54,9 @@
 #include <moveit/robot_state/robot_state.hpp>
 #include <moveit_resources_prbt_ikfast_manipulator_plugin/prbt_ikfast_kinematics_parameters.hpp>
 
+// Code generated by IKFast56/61
+#include "prbt_manipulator_ikfast_solver.cpp"
+
```
rhaschke, line 59: Why did you move these lines?

moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
```diff
@@ -120,7 +120,11 @@ bool PointCloudOctomapUpdater::initialize(const rclcpp::Node::SharedPtr& node)
   node_ = node;
   tf_buffer_ = std::make_shared<tf2_ros::Buffer>(node_->get_clock());
   auto create_timer_interface =
+#if RCLCPP_VERSION_GTE(29, 6, 0)
+      std::make_shared<tf2_ros::CreateTimerROS>(*node);
```
rhaschke, line 124: Did you try to use this new API unconditionally?
In another context that worked for me...

rhaschke (commented):
Thanks for looking into this.

moveit_planners/test_configs/prbt_ikfast_manipulator_plugin/src/prbt_manipulator_ikfast_moveit_plugin.cpp
christian-rauch: Because this includes the header within a namespace and causes issues with namespace resolution.

moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
christian-rauch: What do you mean with "unconditionally"? This API changes in-between releases. The build fails on the latest rolling and lyrical. Using this switch makes this build with the new and old API on rolling, lyrical, and humble.

christian-rauch:
@rhaschke Specifically, these fixes should resolve the build failures marked in the CI badges here: https://github.com/moveit/moveit2/pull/3813.

rhaschke:
Note that current CI build doesn't yet include Lyrical. I'm working on an update of the CI and docker stuff in #3816.
Also, `stomp` isn't yet released to Rolling and Lyrical as far as I can see. This breaks the CI builds too.

rhaschke:
Lyrical is missing `osqp_vendor`

christian-rauch:
> Note that current CI build doesn't yet include Lyrical. I'm working on an update of the CI and docker stuff in #3816.

That explains why they only appeared on the buildfarm and locally. I could only verified the build locally.

> Also, `stomp` isn't yet released to Rolling and Lyrical as far as I can see. This breaks the CI builds too.

`moveit_planners_stomp` builds fine on `lyrical`. See the CI badges. Only `pilz_industrial_motion_planner` is failing on `lyrical`.

Tracing the `moveit_planners_stomp` build issues traces to `ros_industrial_cmake_boilerplate` (https://build.ros2.org/job/Rsrc_uR__ros_industrial_cmake_boilerplate__ubuntu_resolute__source/) but I cannot see the actual build failures in the console output as I only get a "503 Service Temporarily Unavailable".

The build works locally so I can only assume it is related to the buildfarm (timeout, ...) and not the project code.

christian-rauch:
> Lyrical is missing `osqp_vendor`

Not any more. This was just released. See CI badges and https://build.ros2.org/job/Lbin_uR64__osqp_vendor__ubuntu_resolute_amd64__binary/.

I am pretty sure the fixes in this PR should unblock all MoveIt packages on the buildfarm.

moveit_planners/test_configs/prbt_ikfast_manipulator_plugin/src/prbt_manipulator_ikfast_moveit_plugin.cpp
rhaschke: I think the generated code was included within the namespace on purpose: As the code generates identically named functions for every IKFast solver, one couldn't import multiple IKFast plugins at the same time with your solution.
Can you elaborate on your issue please?

christian-rauch: I am getting a lot of error message of the kind "[standard function in std::]" is not a member of ‘prbt_manipulator::std’.

E.g.:
```
[WS]/src/moveit2/moveit_planners/test_configs/prbt_ikfast_manipulator_plugin/src/prbt_manipulator_ikfast_moveit_plugin.cpp:1298:14: error: ‘vector’ is not a member of ‘prbt_manipulator::std’
 1298 |         std::vector<double> sol;
      |              ^~~~~~
```

```
[WS]/src/moveit2/moveit_planners/test_configs/prbt_ikfast_manipulator_plugin/src/prbt_manipulator_ikfast_moveit_plugin.cpp:1367:52: error: ‘rand’ is not a member of ‘prbt_manipulator::std’
 1367 |         sampled_joint_vals.push_back(((diff * std::rand()) / (static_cast<double>(RAND_MAX))) + joint_min);
      |                                                    ^~~~
```

The generated header is doing `using namespace std;` inside the `prbt_manipulator` namespace. I have never seen including headers in namespace like this. And doing it properly outside the namespace appears to resolve the issue.

christian-rauch:
I can reproduce this issue and the fix in this PR also with the [`jspricke/ros-deb-builder-action`](https://github.com/jspricke/ros-deb-builder-action/tree/main/) action.

@nbbrooks @rhaschke Can you please review this PR and release a patch release? This should unblock a few packages on rolling.

I still cannot inspect the build logs, e.g. for https://build.ros2.org/job/Lbin_uR64__pilz_industrial_motion_planner__ubuntu_resolute_amd64__binary/, as I get a "503 Service Temporarily Unavailable". But if this builds on all ROS distros using `jspricke/ros-deb-builder-action`, then I think that the remaining issues are only from unreleased dependencies (e.g. https://github.com/moveit/moveit_msgs/issues/191 and https://github.com/ros-industrial/ros_industrial_cmake_boilerplate/issues/108).

moveit_planners/test_configs/prbt_ikfast_manipulator_plugin/src/prbt_manipulator_ikfast_moveit_plugin.cpp
rhaschke: I see those errors in a Rolling build as well: [link](https://github.com/rhaschke/moveit2/actions/runs/31827957400/job/94856522586)
However, the same code builds perfectly with Lyrical: [link](https://github.com/rhaschke/moveit2/actions/runs/31827957400/job/94856522606)
I don't yet understand this. My concerns still hold. The remaining changes are fine.

rhaschke:
Please see https://github.com/moveit/moveit2/pull/3815#discussion_r3786740175 and #3818.

moveit_planners/test_configs/prbt_ikfast_manipulator_plugin/src/prbt_manipulator_ikfast_moveit_plugin.cpp
nbbrooks: I'll leave in my review comment, but the mechanism we'll want to employ here is `IKFAST_NAMESPACE` - we will still want the `IKFAST_NO_MAIN` set (or not set) before this include.

nbbrooks (commented):
I've been throwing Claude at these sorts of regressions to track down the upstream change that causes them. Here is what it discovered which is consistent with the failure modes. I'm testing a change leveraging `IKFAST_NAMESPACE` now.

---

`ikfast.h:36` includes `<list>`, and the generated solver includes `ikfast.h` at its own line 25 — before its own `<cmath>`/`<vector>`/... block.

**[ros2/rclcpp#3179 "Optimize header include"](https://github.com/ros2/rclcpp/pull/3179)** (merged 2026-06-24) removed `#include <list>` from `rclcpp/include/rclcpp/node.hpp`. That landed in rclcpp 33.x, after Lyrical branched at 32.0.2:

| | rclcpp | `<list>` when the namespace opens | result |
|---|---|---|---|
| Kilted | 29.5.10 | present | builds |
| Lyrical | 32.0.2 | present, via `node.hpp` | nested include is a no-op → **builds** |
| Rolling | 33.0.4 | **absent** | fresh `<list>` opens `namespace std` inside `prbt_manipulator` → **breaks** |

Once *any* nested header opens `namespace std`, `prbt_manipulator::std` exists, and from then on every `std::X` inside the namespace resolves to that inner scope first and fails. That's why @christian-rauch's errors scatter across unrelated symbols (`std::vector` at :1298, `std::rand` at :1367) instead of clustering on list usage.

Every other header the solver includes (`<cmath> <vector> <complex> <iostream> <sstream> <limits> <algorithm> <stdexcept>`) is already present on **both** distros, so `<list>` is the sole differentiator. Of the four `rclcpp` headers this plugin includes, only `node.hpp` ever supplied it.

**So Lyrical is passing by accident, and so is Kilted.** The nesting has been latently broken the whole time; #3179 only removed the crutch. The next header-hygiene cleanup anywhere in that chain breaks Lyrical too. I'd treat this as a real bug to fix rather than a Rolling-specific workaround.

rhaschke:
I cherry-picked all commits except 98958ca to #3818. Closing here. 
@christian-rauch: thanks for pushing this forward!

PR #3818, opened 2026-08-14
Fix CI on Ubuntu 26.04

rhaschke:
Eventually, all distros build fine again. ccov fails for unrelated reasons. I don't have time to dig further into this.
Rolling succeeds w/o any deprecation warnings :rocket:
Interestingly, Lyrical issues a deprecation warning for `ament_index_cpp::get_package_share_directory`, which is not shown by Rolling (although Lyrical uses the older version):
```
ros-lyrical-ament-index-cpp                        1.13.3-3resolute.20260728.204744
ros-rolling-ament-index-cpp                        1.14.2-1resolute.20260717.153205
```
Did they revert the deprecation? I didn't handle that deprecation because I don't understand where they are heading to.

**TODOs**:
- [ ] I suggest merging this PR as a **merge commit** to keep the individual commits
- [ ] Decide on the set of CI builds to perform (ROS distro + additional tests)
  - I suggest including `ikfast` in multiple ROS distros
  - I had to move clang stuff from Rolling to Jazzy due to basic clang build issues in Ubuntu 26.04 :-1:
- [ ] Fix generation of coverage report
- [ ] Fix remaining usages of `FindBoost` module
  - [ ] https://github.com/ros-planning/geometric_shapes
  - [ ] https://github.com/ompl/ompl
- [ ] Release missing packages for Rolling:
  - [ ] https://github.com/ros-industrial/ros_industrial_cmake_boilerplate
  - [ ] https://github.com/ros-industrial/stomp
- [ ] Trigger a new MoveIt2 release for Lyrical + Rolling comprising the fixes

## Summary by CodeRabbit

- **Compatibility**
  - Improved support across newer ROS 2 distributions, Qt versions, compilers, and CMake environments.
  - Updated generated IKFast plugins and controller integrations for broader standard-library and compiler compatibility.
  - Preserved compatibility with older dependency versions where applicable.

- **Build & Testing**
  - Improved dependency discovery and workspace setup reliability.
  - Expanded CI coverage for additional distributions and IKFast scenarios.
  - Refreshed pinned dependencies and upstream sources for more reproducible builds.


moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
```diff
@@ -53,7 +53,7 @@
 #include <rclcpp/time.hpp>
 #include <map>
 #include <memory>
-#include <queue>
+#include <unordered_set>
```
nbbrooks, line 56: Permissions on these files changed from 644 to 755 unexpectedly:
- moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
- test/test_controller_manager_plugin.cpp
- moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp

moveit2_rolling.repos
```diff
@@ -18,4 +18,12 @@ repositories:
   osqp_vendor:
     type: git
     url: https://github.com/moveit/osqp_vendor.git
-    version: 58a52e815ad005d1fc971dca0fdc0c0e8100444b  # Update OSQP version to v1.0.0 (#1)
+    version: 1.0.0
```
nbbrooks:
```suggestion
    version: 1df43ba1250bf93347a23a6a1cd481efd1cd2f68  # v1.0.0 tag
```

The convention for our .repos files is to specify a sha, not a branch or a tag. Branches often get deleted, and tags are technically not immutable. sha is immutable and we will get a loud failure if it gets deleted somehow.

I think what would be a better policy is to use a SHA corresponding to a tag  -which should not get deleted, but if for some reason it does, we will get a loud failure and there is no chance of a tag being changed without our knowledge.

This SHA corresponds to the moveit/osqp_vendor v1.0 release.

.github/workflows/ci.yaml
```diff
@@ -17,47 +17,36 @@ jobs:
...
-          - IMAGE: rolling-ci
+          # Running ccov + ikfast tests on lyrical because rolling-ci misses dependencies
+          - IMAGE: lyrical-ci-testing
             CCOV: true
-            ROS_DISTRO: rolling
-            # Ignore deprecation errors from upstream packages on rolling
-            SUPPRESS_DEPRECATION_ERROR: true
+            EXTRA_CXX_FLAGS: -Wno-error=deprecated-declarations
+          - IMAGE: lyrical-ci-testing
+            IKFAST_TEST: true
```
nbbrooks, line 25: I believe the caches for target workspace and ccache both need to take this into account in their `CACHE_PREFIX` to avoid sharing cache

.github/workflows/ci.yaml
```diff
@@ -111,16 +99,18 @@ jobs:
 #         if: github.repository == 'moveit/moveit2'
 #         with:
 #           domain: moveit
+      - uses: rhaschke/docker-run-action@main
+        name: Retrieve ROS_DISTRO from docker image
+        continue-on-error: true
```
nbbrooks, line 105: I'm not familiar with this action, why do we want continue on error here?

.github/workflows/ci.yaml
nbbrooks, line 57:
```suggestion
      # Strip the deprecated <ciso646> include from system liboctomap-dev, which
      # fatal-errors under -Werror=cpp on GCC 15. The upstream fix
      # (OctoMap/octomap#434) is merged but not yet packaged; drop this once
      # Ubuntu repackages. Must run here, not in an AFTER_SETUP_* hook: rosdep
      # installs target dependencies first and would restore the pristine header.
      BEFORE_BUILD_TARGET_WORKSPACE: >-
        test ! -f /usr/include/octomap/OcTreeKey.h ||
        sed -i 's/#include *<ciso646>//' /usr/include/octomap/OcTreeKey.h
```

.github/workflows/ci.yaml
```diff
@@ -17,47 +17,36 @@ jobs:
...
+      DOCKER_IMAGE: moveit/moveit2:${{ matrix.env.IMAGE }}
       UPSTREAM_WORKSPACE: >
         moveit2.repos
         $(f="moveit2_$(sed 's/-.*$//' <<< "${{ matrix.env.IMAGE }}").repos"; test -r $f && echo $f)
       # Pull any updates to the upstream workspace (after restoring it from cache)
-      AFTER_SETUP_UPSTREAM_WORKSPACE: vcs pull $BASEDIR/upstream_ws/src
+      AFTER_SETUP_UPSTREAM_WORKSPACE: |
+        vcs pull $BASEDIR/upstream_ws/src
+        # silent octomap warnings until upstream fix (OctoMap/octomap#434) is released:
+        sed -i 's/#include *<ciso646>//' /usr/include/octomap/OcTreeKey.h
```
nbbrooks, line 50:
```suggestion
      AFTER_SETUP_UPSTREAM_WORKSPACE: vcs pull $BASEDIR/upstream_ws/src
```

Learning more about industrial_ci, it looks like the sequence is

1. BEFORE_/AFTER_SETUP_<NAME>_WORKSPACE
1. BEFORE_/AFTER_INSTALL_<NAME>_DEPENDENCIES
1. BEFORE_/AFTER_BUILD_<NAME>_WORKSPACE

Since octomap is a system dependency, it seems we need it to be in a later stage, perhaps BEFORE_BUILD_TARGET_WORKSPACE

nbbrooks (commented):
Leaving a summary here for other readers/maintainers

## Why

The `lyrical-ci-testing` image (Ubuntu 26.04) surfaces a batch of toolchain and library breakages that don't appear on the older distros: GCC 15's stricter defaults, Qt 6.10 API deprecations, a `tf2_ros` constructor deprecation, a stale `<ciso646>` include in the system `liboctomap-dev`, and Boost's removal of the CMake `FindBoost` module path. This PR gets the 26.04 job green and reshuffles the matrix so the jobs that need extra dependencies run where those dependencies actually exist.

## CI matrix (`.github/workflows/ci.yaml`)

The matrix is restructured around what each image can actually build:

- **ccov and ikfast move from `rolling-ci` to `lyrical-ci-testing`.** The rolling image is missing dependencies these jobs need.
- **clang-tidy moves from `rolling-ci` to `jazzy-ci`.** Ubuntu 26.04 has issues compiling with `clang++`, so the pedantic clang-tidy pass runs on the last distro where clang is healthy.
- **`rolling-ci` keeps a plain build**, and the separate non-blocking `DEPRECATION_CHECK` job is dropped — deprecation tolerance is now expressed per-entry via `EXTRA_CXX_FLAGS` instead of a dedicated job plus `SUPPRESS_DEPRECATION_ERROR`.
- **The bare-OS Resolute entry is removed** along with the `OS_CODE_NAME` plumbing and the conditional `Set prebuilt DOCKER_IMAGE` step. `DOCKER_IMAGE` is now a plain workflow-level `env`, since every remaining job uses a prebuilt `moveit/moveit2` image.
- **`ROS_DISTRO` is no longer duplicated in the matrix.** It's read out of the image itself via `rhaschke/docker-run-action` and fed to the rosdistro-release-date action, so adding an image no longer means keeping two fields in sync.
- **`-Wno-unknown-warning-option` is now applied only to the clang-tidy job.** On GCC it is itself an unrecognized option, which becomes an error under `-Werror` as soon as any other diagnostic fires.
- **`upstream_ws` cache keys drop the `repos_edit_timestamp` segment.** The `hashFiles('moveit2*.repos', ...)` term already covers repos-file changes, so the extra action call was redundant.
- **The `<ciso646>` workaround replaces `SUPPRESS_CPP_ERROR`.** Rather than dropping `-Werror=cpp` wholesale for the whole build, a targeted `sed` strips the deprecated include from `/usr/include/octomap/OcTreeKey.h`. The upstream fix ([OctoMap/octomap#434](https://github.com/OctoMap/octomap/pull/434)) is merged but not yet packaged; this comes out when Ubuntu repackages.
- **`lcov_capture_args` gains `--ignore-errors=inconsistent`**, needed by the newer lcov on 26.04.

## Dependency pins (`moveit2_rolling.repos`)

- `osqp_vendor` moves from SHA `58a52e8` to tag `1.0.0` (two commits ahead — `CHANGELOG.rst` and `package.xml` only), now that upstream has cut a release.
- `ros_industrial_cmake_boilerplate` `0.7.4` and `stomp` (SHA-pinned) are added, since rolling no longer provides them as released packages.

## Source fixes

**Boost: `MODULE` → `CONFIG` (18 files).** Boost's own `BoostConfig.cmake` is now the supported discovery path; CMake's bundled `FindBoost` is gone in the versions shipping on 26.04. Every `find_package(Boost ...)` gains `CONFIG` with component lists preserved verbatim, and the ones that were passing bare component names are normalized to explicit `COMPONENTS`. `BoostConfig.cmake` still exports `Boost_LIBRARIES` / `Boost_INCLUDE_DIRS`, so the handful of call sites that consume those variables are unaffected — this is a discovery-path change only, not a linkage change.

**Qt 6.10 deprecations (`moveit_setup_framework`, `moveit_setup_srdf_plugins`).** `QRegularExpression::match(QStringView)` is replaced by `matchView()` (Qt >= 6.5), and `invalidateFilter()` by the `beginFilterChange()` / `endFilterChange(Direction::Rows)` pair (Qt >= 6.10). Both are behind `QT_VERSION_CHECK` guards so older distros keep the previous code path. `Direction::Rows` is correct here because `SortFilterProxyModel` overrides only `filterAcceptsRow`.

**`tf2_ros::CreateTimerROS` (`pointcloud_octomap_updater`).** The two-interface constructor is `[[deprecated]]` as of rclcpp 29.6.0; the single-`Node&` overload is used above that version, guarded by `RCLCPP_VERSION_GTE`.

**ikfast plugins.** The `#pragma clang diagnostic` block suppressing `-Wvla-cxx-extension` is wrapped in `#ifdef __clang__` — with the ikfast job now running under GCC, `-Wall` enables `-Wunknown-pragmas` and `-Werror` turns the bare pragma into a hard error. The template and the checked-in `prbt` plugin also gain the standard-library includes they were relying on transitively (`<cmath>`, `<list>`, `<vector>`, `<limits>`, `<complex>`), and unused `<algorithm>` / `<iostream>` are dropped from the generated solver.

**Missing include (`moveit_ros_control_interface`).** `<queue>` was included but unused; `<unordered_set>` is what the file actually needs and was being pulled in transitively. Same include added to the test.

moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp
rhaschke: Looks like Christian's commit introduced those.

moveit2_rolling.repos
rhaschke: If you prefer it that way, sure...

.github/workflows/ci.yaml
rhaschke: Well spotted. I will go for `AFTER_INSTALL_TARGET_DEPENDENCIES`.
It worked because the system dependencies were pulled in by the `ci` docker images already.

.github/workflows/ci.yaml
rhaschke: If the action fails, we just don't know `ROS_DISTRO` and thus `rosdistro_release_date` will fail.
I don't think this will harm the overall build.

.github/workflows/ci.yaml
rhaschke: Both caches already consider `CCOV` and `CLANG_TIDY`. `IKFAST_TEST` is not relevant for those caches, is it?

rhaschke:
I merged this now to fix CI as far as possible (only ccov and clang are still failing for reasons unrelated to the actual source).

PR #3819, opened 2026-08-15
Fix the IKFast plugin compilation on Rolling

nbbrooks:
### Problem

The prbt IKFast plugin does not compile on Rolling. The plugin file includes the generated solver inside `namespace prbt_manipulator`. The generated solver includes `ikfast.h` at line 25. The file `ikfast.h` includes `<list>` at line 36.

A nested `#include <list>` opens `namespace std` inside `namespace prbt_manipulator`. This action creates `prbt_manipulator::std`. After this point, every `std::` name inside the namespace resolves to the inner scope and fails. This behaviour explains the errors on unrelated symbols, for example `std::vector` and `std::rand`.

### Cause

The rclcpp pull request [ros2/rclcpp#3179](https://github.com/ros2/rclcpp/pull/3179) deleted `#include <list>` from `rclcpp/node.hpp` on 2026-06-24. Rolling has rclcpp 33.0.4 and does not include `<list>`. Lyrical has rclcpp 32.0.2 and includes `<list>` through `node.hpp`. Kilted has rclcpp 29.5.10 and also includes `<list>`.

An include guard makes a second `#include <list>` have no effect. Therefore Lyrical and Kilted compile the plugin by accident. The defect is old. The rclcpp pull request only removed the accidental protection. A future header change can break Lyrical in the same way.

### Solution

The generated solver contains its own namespace guard at line 80. The guard starts after the standard includes at lines 31 to 40. Therefore CMake can set `IKFAST_NAMESPACE` and keep the generated symbols inside `prbt_manipulator`.

The plugin file now includes the solver at global scope. The standard includes go to global scope. The IKFast symbols stay inside the namespace. Two IKFast plugins can load into one process without a symbol collision.

CMake also sets `IKFAST_NO_MAIN`. The in-file definition must move to CMake, because a definition after the include comes too late. The solver defines `main` at line 20071 if `IKFAST_NO_MAIN` is absent.

The generator template gets the same change, for consistency. The template includes `<rclcpp/rclcpp.hpp>`. This umbrella header still provides `<list>` on Rolling. Therefore a generated plugin compiles today. The template depends on the same accidental protection as the plugin. This change removes that dependency.

The committed prbt plugin includes four separate `rclcpp` headers instead of the umbrella header. This difference explains why the committed plugin fails and a generated plugin does not.

### Test results

The test uses a clean `ubuntu:resolute` container with rclcpp 33.0.4 and moveit_core 2.15.0.

| Variant | Build | Exports `main` | `ComputeIk` in namespace |
|---|---|---|---|
| Committed plugin, current `main` | **fail** in `bits/stl_list.h` | — | — |
| Committed plugin, this branch | **pass** | 0 | 2 |
| Generated plugin, `main` templates | **pass** | 0 | 2 |
| Generated plugin, this branch templates | **pass** | 0 | 2 |

The build failure occurs inside `bits/stl_list.h`. This location confirms `<list>` as the cause.

The last two rows use `create_ikfast_moveit_plugin.py` with the prbt solver file. The generator creates a
complete package and the test builds it. The generator substitutes `IKFAST_NAMESPACE=prbt_manipulator`
correctly.

### Related

This branch is an alternative to the IKFast change in #3815. The three other changes in #3815 are separate and are not part of this branch.


christian-rauch (approved):
My understanding is that this is effectively the same solution as https://github.com/moveit/moveit2/pull/3815/changes/98958ca20fe8eadefc5a531d856a7177ba240441. The remaining change is just replacing the `#define IKFAST_NO_MAIN` by `target_compile_definitions(PRIVATE IKFAST_NO_MAIN)`, which should have the same effect. I am not sure if this `target_compile_definitions(PRIVATE IKFAST_NO_MAIN)` is strictly required, as rolling compile without it. Do we really need this change?

The PR description only mentions the `<list>` header, but there are many more header and symbols that are affected in the same way.

Also, the description says:
> The template includes `<rclcpp/rclcpp.hpp>`. This umbrella header still provides `<list>` on Rolling.

The rolling `rclcpp.hpp` header https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/rclcpp.hpp only includes `<csignal>` and `<memory>` directly (which could be removed). It does not provide `<list>` directly.

In summary, the solution seems fine, although it might contain more diffs than required. But the Claude generated reasoning is somewhat lacking.
But I also cannot provide a better explanation than "never do `using namespace std;`" and specially, "never do `using namespace std;` inside another namespace", to avoid namespace resolution conflicts.

rhaschke:
While Claude's analysis in https://github.com/moveit/moveit2/pull/3815#pullrequestreview-4942005605 is great, the proposed solution here is very poor. I agree with Christian that the proposed solution is effectively the same as his, but adding much more useless noise.
Particularly, the symbols are still exposed in global namespace, which should be avoided. Claude's claim that they are protected by the cmake-defined `IKFAST_NAMESPACE` is plain wrong as that namespace is never actually used!

I pushed an alternative solution to #3818 (https://github.com/moveit/moveit2/pull/3818/commits/7e653e02f49b690b12bf0b6a6e3f13fc986b2e5e) which solves the issue by explicitly including the required headers in global scope.

Closing here. I will cherry-pick the remaining commits from his PR #3815 and close that one too.

PR #3821, opened 2026-08-15
Fix moveit_py joint-group position setter conversion

### LINKED ISSUE #3718: moveit_py: RobotState joint-group setters segfault in Python
### Description

`moveit_py` crashes with a segmentation fault when calling `RobotState.set_joint_group_positions()` or `RobotState.set_joint_group_active_positions()` from Python, even with a valid joint group name and a correctly sized NumPy array.

I was able to reproduce this outside of `MoveItPy` planning/execution. The issue occurs with just `RobotModel + RobotState`, which suggests the problem is in the Python binding path for the joint-group setter methods.

### ROS Distro

Humble

### OS and version

Ubuntu 22.04

### Source or binary build?

Source

_No response_

### If source, which branch?

humble

### Which RMW are you using?

None

### Steps to Reproduce

I can reproduce the crash with a minimal script like this:
``` python
import glob
import numpy as np
from moveit.core.robot_model import RobotModel
from moveit.core.robot_state import RobotState

# Generate or provide a valid URDF path for the robot
urdf_path = sorted(glob.glob("/tmp/m0609_*.urdf"))[-1]
srdf_path = "/home/username/ros2_ws/install/dsr_moveit_config_m0609/share/dsr_moveit_config_m0609/config/dsr.srdf"

model = RobotModel(urdf_xml_path=urdf_path, srdf_xml_path=srdf_path)
state = RobotState(model)
state.update()

state.set_joint_group_positions(
    joint_model_group_name="manipulator",
    position_values=np.array([0.0, 0.0, 1.57, 0.0, 1.57, 0.0], dtype=float),
)
```
I also reproduced the same crash with:
``` python
state.set_joint_group_active_positions(
    "manipulator",
    np.array([0.0, 0.0, 1.57, 0.0, 1.57, 0.0], dtype=float),
)
```
Important notes:

- The group name is valid.
- The input length matches the group DOF.
- Setting individual joints through state.joint_positions = {...} works correctly.
- The crash therefore seems specific to the joint-group setter binding path.
- _Reproduced consistently with Doosan Robotics config (dsr_moveit_config_m0609) on Humble.
I have not yet confirmed whether this reproduces with other robot configs (e.g., Panda), so this may be config/model-dependent or may expose a generic binding issue._

### Expected behavior

The joint group positions should be updated successfully, and the Python process should continue running normally.

### Actual behavior

Python terminates with a segmentation fault.

[INFO] [moveit_robot_model.robot_model]: Loading robot model 'm0609'...
Segmentation fault (core dumped)

trippyogi:
## Summary

- route `set_joint_group_positions` and `set_joint_group_active_positions` through an owned `std::vector<double>` conversion
- preserve accepted 1-D and n-by-1 array shapes while validating dimensions and group-specific value counts
- add focused coverage for Python sequences, NumPy dtype conversion, column arrays, active positions, and invalid inputs

Fixes #3718.

## Context

The issue reports reproducible segmentation faults on ROS 2 Humble when these two `RobotState` setters receive NumPy arrays. @LeeMin-hyeong confirmed that switching the bindings from the `Eigen::VectorXd` overloads to the existing `std::vector<double>` overloads makes the same calls succeed in the affected environment.

This change copies array-like input into owned storage before entering `RobotState`. It intentionally leaves the velocity and acceleration bindings unchanged, and does not assert a specific Eigen or pybind11 lifetime failure without a backtrace.

## Verification

- pre-commit checks pass
- added unit coverage for list input, active-position NumPy input, float32 and integer conversion, n-by-1 arrays, undersized values, and invalid 1-by-n orientation
- tested the reported Doosan M0609 input with this patch; values were set correctly and the process survived

Independent crash reproduction remains inconclusive. I tested unmodified Humble on Ubuntu 22.04 with Python 3.10.12, NumPy 1.21.5, pybind11 2.9.1, and MoveIt 2.5.9 using both Panda and the public Doosan M0609 URDF/SRDF. The original Eigen bindings did not crash locally, including an edge-case sweep and a setters-only ASan run. This PR therefore does not claim an independently reproduced red-to-green SIGSEGV; it implements the conversion path reported to resolve the fault and adds regression coverage around that boundary.

@LeeMin-hyeong, could you test this branch in the environment where the original calls crash and confirm whether the baseline fails and this patch passes?

### DIFF
--- a/moveit_py/src/moveit/moveit_core/robot_state/robot_state.cpp
+++ b/moveit_py/src/moveit/moveit_core/robot_state/robot_state.cpp
@@ -35,6 +35,7 @@
 /* Author: Peter David Fagan */
 
 #include "robot_state.hpp"
+#include <pybind11/numpy.h>
 #include <pybind11/stl.h>
 #include <moveit_py/moveit_py_utils/ros_msg_typecasters.hpp>
 #include <moveit_msgs/msg/robot_state.hpp>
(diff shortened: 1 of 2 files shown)


LeeMin-hyeong:
Thanks for looking into this and for preparing the fix.

Unfortunately, I no longer have access to the original environment where I reproduced the crash, as it was a temporary lab/training environment. Therefore, I’m unable to test the baseline and this PR branch side by side as requested.

What I can confirm from my earlier testing is that the original `Eigen::VectorXd` binding consistently caused the segmentation fault in that environment, while switching the binding to the existing `std::vector<double>` overload allowed the same call to complete successfully.

Sorry that I can't provide the additional verification on the PR branch, and thank you for investigating this further.

PR #3823, opened 2026-08-18
use "text" signals for drop-down text QComboBox

### LINKED ISSUE #3822: RViz MotionPlanning plugin does not allow changing planning group
### Description

Using the `moveit_rviz_plugin/MotionPlanning` in RViz shows a few signals that are not properly connected:
```
qt.core.qobject.connect: QObject::connect: No such signal QComboBox::activated(QString)
qt.core.qobject.connect: QObject::connect:  (sender name:   'start_state_combo_box')
qt.core.qobject.connect: QObject::connect:  (receiver name: 'MotionPlanningUI')
qt.core.qobject.connect: QObject::connect: No such signal QComboBox::activated(QString)
qt.core.qobject.connect: QObject::connect:  (sender name:   'goal_state_combo_box')
qt.core.qobject.connect: QObject::connect:  (receiver name: 'MotionPlanningUI')
qt.core.qobject.connect: QObject::connect: No such signal QComboBox::currentIndexChanged(QString)
qt.core.qobject.connect: QObject::connect:  (sender name:   'planning_group_combo_box')
qt.core.qobject.connect: QObject::connect:  (receiver name: 'MotionPlanningUI')
```

The effect is that some UI elements do not work, e.g. it is not possible to change the planning group via the "Planning" -> "Query" -> "Planning Group:" drop-down list. It is still possible to change the planning group in the Displays tree under "Planning Request" -> "Planning Group".

### Your environment
* ROS Distro: Lyrical
* OS Version: Ubuntu 26.04
* Source or Binary build: binary testing

### Steps to reproduce
Tell us how to reproduce this issue. Attempt to provide a working demo, perhaps using Docker.

Install `moveit_visual_tools` and run:
```
ros2 launch moveit_visual_tools demo_rviz.launch.py
```
and add the `MotionPlanning` plugin. This will show the messages.

You can also run the Franka fake controller and try changing the planning group.

### Expected behaviour
Selecting the planning group name in the drop-down list should change the active planning group.

### Actual behaviour
Nothing happens since the signals are not connected.

Use [gist.github.com](gist.github.com) to copy-paste the console output or segfault backtrace using gdb.

christian-rauch:
### Description

The `QComboBox` in the `moveit_rviz_plugin/MotionPlanning` uses the signals `activated` and `currentIndexChanged`. At least with newer Qt versions on lyrical, they have the signatures `void activated(int index);` and `void currentIndexChanged(int index);`, which are incompatible with the text.

The correct signals would be those with "text": `void textActivated(const QString &);` and `void currentTextChanged(const QString &);`.

Fixes #3822 .

### Checklist
- [ ] **Required by CI**: Code is auto formatted using [clang-format](https://moveit.ai/documentation/contributing/pullrequests/)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved responsiveness when selecting start states, goal states, and planning groups in the motion-planning interface.
  * Improved configuration control interactions so changes are applied more consistently.
  * Enhanced overall consistency of motion-planning UI updates when selecting options or modifying settings.


rhaschke (changes requested):
Note that the `main` branch should support all active ROS distros, thus still Humble with Qt5.
For MoveIt1 we fixed the Qt5/Qt6 compatibility long ago (https://github.com/moveit/moveit/pull/3582) by choosing different signals.
Also, we used the modern function-pointer syntax for signals and slots, which detects mismatching connections at compile time already.
I suggest performing a cherry-pick (which probably needs many manual adaptations).
An AI should be good at this :wink:

christian-rauch:
I replace the SIGNAL/SLOT macros and this is still compiling on humble without any switches.

But `rolling` fails now with `error: ‘void QCheckBox::stateChanged(int)’ is deprecated: Use checkStateChanged() instead [-Werror=deprecated-declarations]`. Is the `deprecated-declarations` check new? This did not fail before.

I think we should not check for `deprecated-declarations` on `rolling` as this can break the CI at any time when packages in `rolling` are updated with new deprecating declarations. IMHO, this is only relevant when preparing for a new LTS release.

rhaschke:
> Is the `deprecated-declarations` check new?

This was part of our `CXXFLAGS` set for a long time, but was disabled recently due to many issues on Rolling.
As we aim for clean code, we should leave the flag in there. This will alert us as early as possible about deprecations.

christian-rauch:
> > Is the `deprecated-declarations` check new?
> 
> This was part of our `CXXFLAGS` set for a long time, but was disabled recently due to many issues on Rolling. As we aim for clean code, we should leave the flag in there. This will alert us as early as possible about deprecations.

I added compile-time version switches for this. It now builds on all LTS distros.

rhaschke (approved):
Looks great. Thanks a lot!

PR #3828, opened 2026-08-25
Fix trajectory execution manager shutdown segfault

MarqRazz:
`TrajectoryExecutionManager` segfaults on destruction whenever a controller manager plugin is loaded.

The controller manager creates action clients on the node it is given in `initialize()` and the `rclcpp_action::Client` is a `Waitable`, so the node's default callback group holds a weak reference to each. Because`controller_manager_loader_` is declared after `controller_mgr_node_`, the plugin and its loader are destroyed before the node causing it to crash on shutdown.

This PR destroys the objects in the correct order so that it can shutdown cleanly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shutdown and cleanup behavior for trajectory execution management.
  * Ensured controller-related resources are released in the correct order.
  * Reduced the risk of lingering resources or incomplete shutdowns when trajectory execution ends.


tylerjw (approved):

rhaschke:
Why do unit tests fail again? Are they flaky? Triggered a new build...

PR #3833, opened 2026-08-26
Fix logger node teardown before RMW shutdown

### LINKED ISSUE #3827: Using moveit with Zenoh causes a crash on process exit
### Description

When using moveit in a process with Zenoh it will cause a crash on exit due to the thread local static rclcpp::Node in the setNodeLoggerName function.
I am on Kilted.

This prints the following crash:

thread '<unnamed>' panicked at /home/buildfarm/.cargo/git/checkouts/zenoh-cc237f2570fab813/2687c51/commons/zenoh-runtime/src/lib.rs:154:21:
The Thread Local Storage inside Tokio is destroyed. This might happen when Zenoh API is called at process exit, e.g. in the atexit handler. Calling the Zenoh API at process exit is not supported and should be avoided.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[ros2run]: Aborted

We have some unit tests which interfaces with moveit, and they have started crashing with this error after switching to Zenoh causing the tests to fail.

### ROS Distro

Rolling

### OS and version

Ubuntu 24.04

### Source or binary build?

Binary

2.14.3-1noble.20260813.090120

### If source, which branch?

_No response_

### Which RMW are you using?

None

### Steps to Reproduce

Here is a MWE, this simple node will crash on exit when ran with zenoh (export RMW_IMPLEMENTATION=rmw_zenoh_cpp)

```
#include <cstdio>
#include <moveit/utils/logger.hpp>
#include <rclcpp/rclcpp.hpp>

int main(int argc, char ** argv)
{
  rclcpp::init(argc, argv);
  moveit::setNodeLoggerName("moveit_crash_package"); // Commenting this line out does not cause any crash
  rclcpp::shutdown();
}
```

### Expected behavior

That running this node should exit cleanly without printing any errors.

### Actual behavior

This crashes on exit and prints the following error:

> thread '<unnamed>' panicked at /home/buildfarm/.cargo/git/checkouts/zenoh-cc237f2570fab813/2687c51/commons/zenoh-runtime/src/lib.rs:154:21:
The Thread Local Storage inside Tokio is destroyed. This might happen when Zenoh API is called at process exit, e.g. in the atexit handler. Calling the Zenoh API at process exit is not supported and should be avoided.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[ros2run]: Aborted

thread '<unnamed>' panicked at /home/buildfarm/.cargo/git/checkouts/zenoh-cc237f2570fab813/2687c51/commons/zenoh-runtime/src/lib.rs:154:21:
The Thread Local Storage inside Tokio is destroyed. This might happen when Zenoh API is called at process exit, e.g. in the atexit handler. Calling the Zenoh API at process exit is not supported and should be avoided.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[ros2run]: Aborted

dlanov:
### Description

Fixes #3827.

MoveIt's logger helpers keep internal `rclcpp::Node` instances alive as function-local statics. Those nodes were destroyed during process teardown, after `rclcpp::shutdown()` had already shut down the RMW context.

With `rmw_zenoh_cpp`, the late node destruction calls back into Zenoh after its runtime has been torn down and aborts the process.

This change registers a pre-shutdown callback on each logger node's own `rclcpp::Context` and resets the node before `rcl_shutdown()` runs.

The callback does not retain the node: it references the caller-owned static node slot and weakly holds the synchronization guard, avoiding a `Context -> callback -> Node -> Context` ownership cycle and keeping static teardown safe when shutdown is not called explicitly.

Existing logger naming and normal `/rosout` behavior before shutdown are preserved. The copied `rclcpp::Logger` remains valid after its node is reset; the node-owned `/rosout` publisher is torn down during pre-shutdown.

Regression coverage verifies:

- logger nodes are destroyed during pre-shutdown;
- the shutdown callback does not retain the logger node or synchronization guard;
- `setNodeLoggerName()` is safe across explicit shutdown;
- the pre-`rclcpp::init()` fallback remains unchanged.

Validation with ROS 2 Jazzy and `rmw_zenoh_cpp` 0.2.9:

- `SetNodeLoggerNameTest.SafeAcrossExplicitShutdown` passes under `rmw_zenoh_cpp`
- all 4 shutdown/lifecycle tests and the isolated pre-init fallback test pass under `rmw_zenoh_cpp`
- existing `test-node_logging` `/rosout` launch test passes under `rmw_zenoh_cpp`
- targeted CTest validation passes for `test_logger`, `test_logger_before_init`, and `test-node_logging`
- `colcon build --packages-select moveit_core --symlink-install --cmake-args -DBUILD_TESTING=ON` passes
- pre-commit passes on all changed files
- `git diff --check` passes

### Checklist
- [x] **Required by CI**: Code is auto formatted using [clang-format](https://moveit.ai/documentation/contributing/pullrequests/)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved logger reliability during application and context shutdown.
  * Prevented logging failures when nodes are destroyed or shutdown unexpectedly.
  * Ensured safe fallback logging before initialization and after shutdown.
  * Preserved logger behavior when node access changes during shutdown.

* **Tests**
  * Added coverage for logger lifecycle, shutdown handling, node destruction, race conditions, and operation before initialization.


moveit_core/utils/src/logger.cpp
```diff
@@ -36,9 +36,11 @@
 
 #include <rclcpp/rclcpp.hpp>
 #include <moveit/utils/logger.hpp>
+#include <mutex>
```
rhaschke: <mutex> is included via logger_detail.hpp
```suggestion
```

moveit_core/utils/src/logger_detail.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+/// Out-of-band flag + mutex used to guard a caller-owned rclcpp::Node::SharedPtr
+/// against a pre-shutdown callback resetting it concurrently, and to record
+/// whether that reset has already happened. Deliberately its own, separately
+/// allocated object -- NOT a wrapper that also stores the Node::SharedPtr
+/// itself. See registerNodeResetOnPreShutdown() for why.
+struct LoggerNodeFlag
+{
+  std::mutex mutex;
+  bool retired = false;
```
rhaschke: The `retired` flag isn't used anywhere except in tests.

moveit_core/utils/src/logger.cpp
```diff
@@ -72,8 +80,17 @@ rclcpp::Logger& getGlobalRootLogger()
...
 {
-  static auto node = std::make_shared<rclcpp::Node>("moveit", name);
-  getGlobalRootLogger() = node->get_logger();
+  static rclcpp::Node::SharedPtr node = std::make_shared<rclcpp::Node>("moveit", name);
+  static std::shared_ptr<detail::LoggerNodeFlag> flag = detail::registerNodeResetOnPreShutdown(node);
+
+  std::lock_guard<std::mutex> lock(flag->mutex);
+  if (node)
+  {
+    getGlobalRootLogger() = node->get_logger();
```
rhaschke: What happens to the logger when its owning node is reset?

rhaschke (commented):
Nice contribution. I have a few remarks and questions though.
And there seems to be a regression on Jazzy with it.

moveit_core/utils/src/logger.cpp
dlanov: Good catch, removed. logger_detail.hpp already provides the required <mutex> include.

moveit_core/utils/src/logger_detail.hpp
dlanov: Removed. I also simplified the helper so the wrapper struct is no longer needed; the returned mutex is the only production state required for synchronization and safe teardown ordering. The tests now verify the node/guard lifetime directly instead of using a test-only retired flag.

moveit_core/utils/src/logger.cpp
dlanov: I checked this against Jazzy rclcpp::Logger. The Logger owns its logger-name state independently and does not retain a reference to the Node, so the copied Logger remains valid after the Node is reset. The Node-owned /rosout publisher is torn down during pre-shutdown, while normal /rosout behavior before shutdown is unchanged. I added regression coverage confirming that the retained logger name remains valid across shutdown.

dlanov:
Thanks for the review. I addressed the three comments and the subsequent race/test-isolation findings. The earlier Jazzy failure was from the clang-tidy delta lane: the new function-local statics needed MoveIt's required s_ prefix. Local Jazzy validation now passes, including the lifecycle tests, isolated pre-init fallback test, existing /rosout test, and repeated runs with rmw_zenoh_cpp. The new GitHub Actions runs are currently waiting for fork-workflow approval.

rhaschke:
> The Jazzy failure was from the clang-tidy delta lane: the new function-local statics needed MoveIt's required s_ prefix.

I don't think that was the issue. test_controller_manager_plugin fails with error code -11:
`run_test.py: return code -11`

dlanov:
Rebased onto current `main` and fixed the cmake-format issue. The targeted `test_logger`, `test_logger_before_init`, and `test-node_logging` tests all pass locally. CI is currently waiting for workflow approval.

rhaschke:
Thanks for iterating on this.

PR #3840, opened 2026-08-28
Fix MoveItPy executor thread lifecycle (#3839)

### LINKED ISSUE #3839: MoveItPy detaches its executor thread, allowing callback/destructor overlap and leaking the thread on constructor failure
### Description

The `MoveItPy` constructor starts and detaches a `SingleThreadedExecutor` thread before constructing `MoveItCpp`. The normal custom deleter calls `executor->cancel()`, global `rclcpp::shutdown()`, and immediately deletes `MoveItCpp`, but it cannot join the executor thread.

There are two failure modes:

1. `cancel()` requests/wakes executor shutdown but does not wait for an already selected callback to finish. Deleting `MoveItCpp` and its components can overlap a callback that uses those components.
2. If `new MoveItCpp(node)` throws after the thread is detached, the shared pointer and its custom deleter are never established. The detached thread retains the node/executor and continues spinning in the process.

Destroying any one instance also calls global `rclcpp::shutdown()`, affecting unrelated ROS nodes and other MoveItPy instances in the same Python process.

### ROS Distro

Jazzy

### OS and version

Ubuntu 24.04

### Source or binary build?

Source

_No response_

### If source, which branch?

_No response_

### Which RMW are you using?

None

### Steps to Reproduce

1. In a long-lived Python process, repeatedly construct `MoveItPy` with an invalid/incomplete robot model configuration that makes `MoveItCpp` construction throw.
2. Catch the Python exception and keep the process alive.
3. Inspect native threads and ROS resources after each attempt. The executor thread was detached before the throwing constructor and has no cleanup owner.

### Expected behavior

Node, executor, thread, and `MoveItCpp` components are managed by one RAII lifetime. Both successful destruction and constructor exceptions stop and join the executor before component resources are released. Instance destruction does not globally shut down unrelated ROS contexts.

### Actual behavior

The executor thread is detached. Normal cleanup cannot wait for it, and constructor failure bypasses cleanup entirely. Global shutdown is tied to deletion of an individual instance.

_No response_

timothyanderson096-ocdealcheck:
### Description

Fixes #3839.

`MoveItPy` currently detaches its executor thread before constructing `MoveItCpp`. That leaves the worker without a lifetime owner: construction failure can leave it running, while normal destruction cannot wait for active callbacks before deleting `MoveItCpp`. The instance deleter also shuts down the process-wide default ROS context.

This change gives the executor worker an RAII owner. Cleanup now:

* requests the worker to stop;
* cancels any bounded `spin_once()` wait;
* joins the worker before deleting `MoveItCpp`;
* performs the same stop-and-join cleanup if `MoveItCpp` construction throws; and
* leaves process-wide ROS shutdown to the existing explicit `shutdown()` method.

The bounded `spin_once()` loop also covers the race where destruction starts before the new thread reaches its first spin call.

### Validation

* Local patch application check against upstream commit `8fcb5d2327767af3c764b1348a69a9b74c5efd18` — passed
* Standalone RAII lifecycle probe compiled with `-Wall -Wextra -Werror -pedantic -fsanitize=address,undefined` — passed
* Probe verified constructor-exception cleanup leaves no worker thread behind
* Probe verified cleanup waits for an active callback before dependent destruction
* Full MoveIt/ROS build, repository test suite, and repository clang-format check — **not run in the current environment**

This is opened as a draft so CI and maintainer feedback can validate the integration and determine the preferred in-repository regression test.

### Checklist

* [ ] **Required by CI**: Code is auto formatted using [[clang-format](https://moveit.ai/documentation/contributing/pullrequests/)](https://moveit.ai/documentation/contributing/pullrequests/)
* [ ] Extend the tutorials/documentation — not applicable
* [ ] Document API changes in `MIGRATION.md` — maintainer decision requested
* [ ] Create tests that fail without this PR — isolated lifecycle proof completed; integrated test still required
* [ ] Include a screenshot if changing a GUI — not applicable
* [ ] Review another open pull request while waiting for review

### AI-assisted verification disclosure

This patch and its evidence package were prepared with AI assistance and reviewed and approved by Tim. The exact base revision, validation performed, and untested boundaries are disclosed above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MoveIt C++ integration shutdown and cleanup behavior.
  * Reduced the risk of lingering background processing after MoveIt components are released.
  * Improved responsiveness during executor shutdown through bounded wait intervals.
  * Prevented potential hangs during cleanup when shutdown is initiated from an active processing callback.

* **Stability**
  * Improved lifecycle management for background processing, supporting more predictable application behavior during startup, runtime, and shutdown.

### DIFF
--- a/moveit_py/src/moveit/moveit_ros/moveit_cpp/moveit_cpp.cpp
+++ b/moveit_py/src/moveit/moveit_ros/moveit_cpp/moveit_cpp.cpp
@@ -73,9 +73,14 @@ public:
   ExecutorThread(const ExecutorThread&) = delete;
   ExecutorThread& operator=(const ExecutorThread&) = delete;
 
+  bool isCurrentThread() const noexcept
+  {
+    return execution_thread_.joinable() && execution_thread_.get_id() == std::this_thread::get_id();
+  }
+
   void stop() noexcept
   {
-    if (!execution_thread_.joinable())
+    if (!execution_thread_.joinable() || isCurrentThread())
       return;
 
     stop_requested_->store(true);
@@ -175,6 +180,18 @@ void initMoveitPy(py::module& m)
              auto executor_thread = std::make_shared<ExecutorThread>(node);
 
              auto custom_deleter = [executor_thread](moveit_cpp::MoveItCpp* moveit_cpp) {
+               if (executor_thread->isCurrentThread())
+               {
+                 // A node callback may release the final MoveItCpp holder. Defer cleanup so the executor thread is
+                 // joined externally and MoveItCpp remains alive until the callback has returned.
+                 std::thread cleanup_thread([executor_thread, moveit_cpp]() {
+                   executor_thread->stop();
+                   delete moveit_cpp;
+                 });
+                 cleanup_thread.detach();
+                 return;
+               }
+
                executor_thread->stop();
                delete moveit_cpp;
              };
--- a/moveit_py/src/moveit/moveit_ros/moveit_cpp/moveit_cpp.cpp
+++ b/moveit_py/src/moveit/moveit_ros/moveit_cpp/moveit_cpp.cpp
@@ -35,9 +35,12 @@
 /* Author: Peter David Fagan */
 
 #include "moveit_cpp.hpp"
+#include <atomic>
+#include <chrono>
 #include <pybind11/pytypes.h>
 #include <moveit/utils/logger.hpp>
 #include <string>
+#include <thread>
 
 namespace moveit_py
 {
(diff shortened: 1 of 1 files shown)


timothyanderson096-ocdealcheck:
Fixed in 9940ffa. `stop()` now refuses a self-join, and the custom
deleter detects final release on the executor thread and transfers
stop/join/destruction to a cleanup thread. Normal external destruction
remains synchronous, while `MoveItCpp` stays alive until an active
callback returns.

Validation:
- the old path reliably aborted with `std::system_error: Resource
deadlock avoided` (exit 134)
- the fixed path passed 200 forced self-destruction cycles
- the normal external-destruction path passed
- AddressSanitizer/UndefinedBehaviorSanitizer and ThreadSanitizer passed

PR #3851, opened 2026-09-09
Give the controller spawners a parameter file

nbbrooks:
### Description

The controller spawners do not receive a parameter file. The `controller_manager` node reads `ros2_controllers.yaml`, but the controller nodes do not read it. Each controller therefore starts with no parameters, and initialization fails:

```
Exception thrown during init stage with message: Invalid value set during initialization
for parameter 'joints': Length of parameter 'joints' is '0' but must be greater than '0'
[ERROR] [controller_manager]: Could not initialize the controller named 'panda_arm_controller'
```

The user can plan a motion in RViz, but the user cannot run the motion. The **Execute** button has no action server to call.

The `spawner` program has a `--param-file` option that loads the file into the controller node before configuration. This pull request gives that option to every spawner that MoveIt starts.

### Changes

**1. `moveit_configs_utils.launches.generate_spawn_controllers_launch()`**

Passes `--param-file` with the `ros2_controllers.yaml` of the configuration package. `generate_demo_launch()` already uses that path convention for the `ros2_control_node`. The option is omitted when the file does not exist, so a configuration package without one keeps its present behavior.

The function also takes an optional `controller_manager_timeout` for callers that start the controller manager in the same launch file.

The callers of this function are generated, not committed: MoveIt Setup Assistant offers a "Spawn Controllers Launch" bundle (`moveit_setup_app_plugins/src/launches.cpp:66`), and `templates/launch/generic.launch.py.template` writes a launch file whose body is `return generate_spawn_controllers_launch(moveit_config)`. Every generated configuration package therefore calls it.

**2. Four `moveit_servo` demo launch files now call that function**

`demo_joint_jog`, `demo_pose`, `demo_ros_api` and `demo_twist` spawn exactly the controllers the function creates, from the same configuration package, so the call is equivalent. They keep their 300 second wait through the new argument.

**3. Eight further launch files receive `--param-file` directly**

The function would change the behavior of these, so they keep their own spawners:

| File | Reason |
|---|---|
| `prbt_moveit_config/launch/demo.launch.py`, `hybrid_planning/test/launch/test_basic_integration.test.py` | spawn `panda_joint_group_position_controller`, not the controller in `moveit_controllers.yaml` |
| `planning_interface/test/launch/move_group_launch_test_common.py`, `tests/launch/move_group_api.test.py` | also spawn `panda_hand_controller` |
| `planning_scene_monitor.test.py`, `servo_{cpp,ros}_integration.test.py`, `servo_utils.test.py` | stage each spawner with a `TimerAction` or an event handler |

That is 18 spawners across 12 files. Every file already holds the correct path in `ros2_controllers_path`.

### Testing

Ubuntu 26.04 (Resolute), ROS 2 Lyrical binary packages, MoveIt 2.15.0 from the build farm.

`ros2 launch moveit_servo demo_ros_api.launch.py` before the change: `Failed loading controller panda_arm_controller`, no active controller.

After:

```
panda_arm_controller     joint_trajectory_controller/JointTrajectoryController  active
joint_state_broadcaster  joint_state_broadcaster/JointStateBroadcaster          active
```

The same correction on `moveit_resources` (moveit/moveit_resources#221) brings all three Panda controllers to `active` and makes `/panda_arm_controller/follow_joint_trajectory` and `/panda_hand_controller/gripper_cmd` available, so trajectory execution works again.

I also confirmed that each changed launch description still builds, and that `ros2_controllers_path` is in scope at all 18 spawners.

### Note on the tests

The CI log of the first run shows a controller failing to initialize while the test still passes:

```
[controller_manager]: Could not initialize the controller named 'panda_arm_controller'
6/6 Test #6: test_best_seen_execution_time_policy_with_move_group ....... Passed
```

The `_with_move_group` tests therefore run today without an active controller. This pull request starts the controllers; a test that fails afterwards was failing silently before.

### Related

`moveit_resources` has the same defect in its hand-written demo launch files: moveit/moveit_resources#221.

### Checklist

- [x] Code is auto formatted using black (pre-commit passes).
- [ ] Extend the tutorials / documentation — not needed.
- [ ] Document API changes relevant to the user in MIGRATION.md — the new argument is optional and the default keeps present behavior.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **Bug Fixes**
  - Controller spawners now load parameters from the configured controller file when available.
  - Controller startup behavior is more consistent across demos and integration tests.
  - Joint-state, arm, and hand controllers now use consistent spawning and configuration behavior across supported launch scenarios.
  - Controller launches support an extended manager startup timeout where needed.
  - The PRBT demo now starts the controller defined by its configured controller mapping.


moveit_configs_utils/moveit_configs_utils/launches.py
```diff
@@ -121,12 +121,21 @@ def generate_spawn_controllers_launch(moveit_config):
         "moveit_simple_controller_manager", {}
     ).get("controller_names", [])
     ld = LaunchDescription()
+    # The controller_manager node reads ros2_controllers.yaml, but the
+    # controller nodes do not read it. Give the file to each spawner so that
+    # every controller gets its parameters before configuration.
```
rhaschke, line 126:
```suggestion
    # Pass the same ros2_controllers.yaml to the controller spawners as to the controller_manager
```

rhaschke (commented):
I don't see any code that actually uses the utility functions in `‎moveit_configs_utils/moveit_configs_utils/launches.py`.
Rather, there exist multiple launch files, manually creating the spawner config, e.g.:
https://github.com/moveit/moveit2/blob/797bdeff95c75158bc3b969330456d131dcf27c1/moveit_ros/planning/planning_scene_monitor/test/launch/planning_scene_monitor.test.py#L46-L51

Hence, I expect more changes are required to actually fix the mentioned issue.

nbbrooks:
Thank you for the review. You are correct that more changes are necessary, and I have extended the pull request.

### The utility function does have callers, but they are generated

MoveIt Setup Assistant offers a **"Spawn Controllers Launch"** bundle:

* `moveit_setup_assistant/moveit_setup_app_plugins/src/launches.cpp:66` registers the bundle with the name `spawn_controllers`.
* `moveit_setup_app_plugins/templates/launch/generic.launch.py.template` writes a launch file whose whole body is `return [FUNCTION_NAME](moveit_config)`.

Every configuration package that a user generates therefore contains a `spawn_controllers.launch.py` that calls `generate_spawn_controllers_launch()`. The callers live in downstream packages, which is why `git grep` finds none in this repository. The change to the utility function is thus necessary — but, as you say, it is not sufficient.

### What the pull request now does

**Four `moveit_servo` demo launch files now call the utility function.** They spawn exactly the controllers it creates (`panda_arm_controller` and `joint_state_broadcaster`) and use the same configuration package, so the call is equivalent. The utility function takes a new optional `controller_manager_timeout`, because these demos start the controller manager themselves and need the 300 second wait that they used before.

**Eight other launch files receive `--param-file` directly.** The utility function would change their behavior:

| File | Why the utility function does not fit |
|---|---|
| `prbt_moveit_config/launch/demo.launch.py`, `hybrid_planning/test/launch/test_basic_integration.test.py` | spawn `panda_joint_group_position_controller`, which is not the controller in `moveit_controllers.yaml` |
| `planning_interface/test/launch/move_group_launch_test_common.py`, `tests/launch/move_group_api.test.py` | also spawn `panda_hand_controller` |
| `planning_scene_monitor.test.py`, `servo_{cpp,ros}_integration.test.py`, `servo_utils.test.py` | stage each spawner with a `TimerAction` or an event handler, which the utility function does not reproduce |

Each of these files already holds the correct path in `ros2_controllers_path`, so the change is one argument per spawner. That is 18 spawners across the 12 files.

### Evidence

On Ubuntu 26.04 with Lyrical binary packages, `ros2 launch moveit_servo demo_ros_api.launch.py` fails today:

```
Exception thrown during init stage with message: Invalid value set during initialization
for parameter 'joints': Length of parameter 'joints' is '0' but must be greater than '0'
[ERROR] [controller_manager]: Could not initialize the controller named 'panda_arm_controller'
[FATAL] [spawner_panda_arm_controller]: Failed loading controller panda_arm_controller
```

With this pull request:

```
[spawner_joint_state_broadcaster]: Configured and activated joint_state_broadcaster
[spawner_panda_arm_controller]: Configured and activated panda_arm_controller

panda_arm_controller     joint_trajectory_controller/JointTrajectoryController  active
joint_state_broadcaster  joint_state_broadcaster/JointStateBroadcaster          active
```

I also confirmed that every changed launch description still builds, and that `ros2_controllers_path` is in scope at all 18 spawners.

### One more thing you may want to know

The CI log of the previous run shows the same defect inside our own tests, and the test still passes:

```
[controller_manager]: Could not initialize the controller named 'panda_arm_controller'
[spawner panda_arm_controller-4]: process has died
...
6/6 Test #6: test_best_seen_execution_time_policy_with_move_group ....... Passed
```

So the `_with_move_group` tests presently run without an active controller. This pull request makes the controllers start; if any test then fails for a real reason, that failure was previously hidden.

### CI

The one red job on the previous run was a timeout of
`planning_scene_monitor_test_launch_planning_scene_monitor.test.py`, which produced no result file. The same job passes on `main`, and this branch changed only Python launch files that, as you noted, nothing in the repository called at that point. I believe it is a flaky fixture of the kind fixed in #3809 and #3777, but please tell me if you would rather I look into it in this pull request.

Your inline suggestion is applied.

moveit_configs_utils/moveit_configs_utils/launches.py
nbbrooks: Applied, thank you.

nbbrooks:
Good catch, now this PR extends to 12 launch files in this repo that were building (also incorrect) spawners by hand.

Beyond these launch files, MSA also leverages this helper script.

rhaschke (approved):

nbbrooks:
Made a minor change after Robert's approval to guard against edge cases involving strings with spaces that are not quote enclosed. Merging so I can cut the release for Lyrical sync.

PR #3853, opened 2026-09-12
fix race on TrajectoryExecutionManager::execution_complete_

### LINKED ISSUE #3852: Data race on TrajectoryExecutionManager::execution_complete_ during stopExecution()
### Description

`TrajectoryExecutionManager::execution_complete_` is a plain `bool` used to communicate completion/stop state between the trajectory worker and the thread calling `stopExecution()`. Some accesses hold `execution_state_mutex_`, while others do not.

In the reviewed commit:

- [`stopExecution()` writes `execution_complete_ = true` while holding `execution_state_mutex_`](https://github.com/moveit/moveit2/blob/797bdeff95c75158bc3b969330456d131dcf27c1/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp#L1211-L1234), before joining the worker.
- [`executeThread()` reads the same flag without that mutex at entry](https://github.com/moveit/moveit2/blob/797bdeff95c75158bc3b969330456d131dcf27c1/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp#L1321-L1330). There are also unlocked reads in `executePart()` (lines 1387, 1550 and 1573).

The worker is a separate `std::thread`, created in `execute()` at lines 1287–1289. Locking only the writer does not synchronize these reads, and the subsequent `join()` does not prevent the preceding conflicting accesses. This constitutes a C++ data race.

Related: #3198 protects trajectory cleanup and `waitForRobotToStop()`, but does not synchronize all accesses to this flag. Those changes are already present in the reviewed code.

### ROS Distro

Jazzy

### OS and version

Ubuntu 24.04

### Source or binary build?

Source

_No response_

### If source, which branch?

_No response_

### Which RMW are you using?

None

### Steps to Reproduce

Source-level triggering sequence (not yet runtime-verified):

1. Push a valid trajectory and call `execute()` so that validation succeeds and the worker starts.
2. Call `stopExecution()` while the worker is active. A minimal conflicting interleaving is the stop-side write at line 1222 overlapping the worker's initial read at line 1325.
3. The writer holds `execution_state_mutex_`, but the worker does not acquire it before this read.

This stop path is used during normal operation: `PlanExecution::executeAndMonitor()` calls the manager's `stopExecution()` when it detects a preemption request or an invalidated path.

### Expected behavior

All concurrent accesses to the completion/stop flag follow a consistent synchronization protocol, using the same mutex or an atomic flag with appropriate coordination of the related state.

### Actual behavior

The stop thread can write a non-atomic flag while the worker reads it without synchronization, resulting in undefined behavior. No specific crash, hang, or failure to stop is claimed as reproduced.

_No response_

dundysm:
### Description

Fixes #3852.

`TrajectoryExecutionManager::execution_complete_` was a plain `bool` used as a stop/completion flag between `stopExecution()` and the execution worker. Some accesses hold `execution_state_mutex_`, while `executeThread()` / `executePart()` (and a few other sites) read it without that mutex. That is a C++ data race: locking only the writer does not synchronize unlocked readers, and a later `join()` does not repair earlier conflicting accesses.

This change makes `execution_complete_` a `std::atomic<bool>`. Existing assignment and boolean checks continue to work; the mutex + `execution_complete_condition_` protocol used by `waitForExecution()` / worker completion is unchanged, and `stopExecution()` still joins the worker the same way.

**Why atomic instead of mutex-only:** #3198 already tightened locking around trajectory cleanup / `waitForRobotToStop()`. Forcing every hot-path poll in `executePart()` under `execution_state_mutex_` would change contention characteristics. `std::atomic<bool>` matches nearby planning code that uses atomics for cross-thread stop/preemption flags (for example `PlanExecution::preempt_`).

All call sites of `execution_complete_` were audited; only the member type needed to change.

### Checklist
- [x] **Required by CI**: Code is auto formatted using [clang-format](https://moveit.ai/documentation/contributing/pullrequests/) (one-line type/include change)

### Test plan
- Audited every `execution_complete_` access in `trajectory_execution_manager.cpp` / header.
- Full `moveit_ros_planning` / TrajectoryExecutionManager package build and gtests were **not** run here (no ROS/MoveIt desktop on the verification host). Existing `trajectory_execution_manager` gtests remain disabled/commented in that package's `CMakeLists.txt`.
- Standalone ThreadSanitizer sanity check of the same locked-write / unlocked-read pattern: plain `bool` reports a data race; `std::atomic<bool>` is clean. This is a pattern-level check, not a TSAN run of the MoveIt translation unit.
- Please rely on MoveIt CI for the real package build.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the reliability of trajectory execution status handling, particularly when updates occur concurrently.


dundysm:
@rhaschke @christian-rauch @elyforl gentle ping on this fix for #3852; execution_complete_ is now std::atomic<bool> so stopExecution and the worker no longer race on a plain bool. happy to adjust if you'd rather keep the bool and take execution_state_mutex_ on every read instead.

PR #3856, opened 2026-09-18
Fix goal joint editing after trajectory execution

### LINKED ISSUE #3848: MotionPlanning Rviz2 plugin: cannot adjust joint angles after executing unless the goal state is re-selected
### Description

If the Rviz2 MotionPlanning panel is used to plan and execute a trajectory for a robot arm, it only works right away on the first run. If the joint angles are modified for subsequent positions, this will have no effect unless the drop-down for the goal state is opened and some state is selected (even if it's the same one that was already set). Only then will the joint angles be reset to the current state and actually change the goal state upon being modified. I do not see what advantage this has over automatically updating the goal state so we can keep modifying the joint angles for further poses without having to do that re-selection. In fact, this does not happen with the interactive markers, which do seem to update the goal state as interacting with them does allow you to then change the angle for each joint manually.

### Your environment
* ROS Distro: Humble
* OS Version: Ubuntu 22.04.05
* Source or Binary build? Binary
* If binary, which release version? 2.5.9

### Steps to reproduce
1) Launch a MoveIt2 robot package and Rviz2
2) Open the MotionPlanning panel (install the plugin if not installed)
3) Change the joint angles and click "Plan and Execute" with the default settings
4) Try to change the joint angles to create the next pose (and notice it doesn't work)

### Expected behaviour
Further combinations of joint angles should be able to create the next pose without having to re-select the goal state to the one that's already selected every time.

### Actual behaviour
For the manually-modified joint angles to be able to create a goal pose, the goal_pose must be re-selected first every time.

Greninja44:
## Summary
Fixes the MotionPlanning RViz2 panel silently editing the wrong robot state (start instead of goal) after "Plan and Execute", requiring the user to re-select the Goal State dropdown to restore correct behavior.

## Problem
After clicking "Plan and Execute" and execution finishes, manually adjusting joint values in the Joints tab has no effect on the next planning goal — until the (already-selected) item in the Goal State dropdown is re-selected. Interactive-marker-driven goal editing is unaffected the whole time.

## Root cause
`MotionPlanningFrameJointsWidget` keeps separate `JMGItemModel` instances for the start and goal state but binds only one to the visible/editable table at a time, via `setActiveModel()`. Its `queryStartStateChanged()`/`queryGoalStateChanged()` slots called `setActiveModel()` unconditionally whenever either query state changed *for any reason* — including the passive, non-user-initiated resync that `MotionPlanningFrame::onFinishedExecution()` performs on the start state after every execution (to keep a `"<current>"` Start State selection mirroring the robot's live pose). That passive resync goes through the exact same signal path as a genuine user selection, so it silently rebinds the Joints tab's table from the goal model to the start model — subsequent joint edits then land in the wrong `RobotState`.

## Solution
Added a `passive_sync` parameter, threaded from `MotionPlanningDisplay::setQueryStartState`/`setQueryGoalState` through to the `queryStartStateChanged(bool)`/`queryGoalStateChanged(bool)` signals and the Joints widget's slots. The widget still always refreshes its cached model data, but only rebinds the visible table (`setActiveModel()`) when `!passive_sync`. The two confirmed passive-resync call sites (`onFinishedExecution()`'s post-execution start refresh, and `MotionPlanningDisplay::updateQueryStates()`'s continuous scene-monitor resync) pass `true`; every other call site (combo box selection, interactive markers, named/random states) relies on the parameter's default of `false`, so existing behavior is unchanged there.

## Testing
Added `motion_planning_rviz_plugin/test/joints_widget_focus_test.cpp` — the first test for this package — exercising `MotionPlanningFrameJointsWidget` directly against a real Panda robot model (`moveit::core::loadTestingRobotModel`), with no RViz/DisplayContext dependency needed. Verified the test fails for the exact expected reason with the fix's gating reverted, and passes with it restored. Built and linked the full package (`colcon build --packages-select moveit_ros_visualization`) against installed ROS 2 Jazzy / MoveIt 2.12.4 binaries; also reproduced the bug live in RViz before the fix. `clang-format`, `codespell`, and `cmake-format` all clean.

Fixes #3848

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented automatic robot-state updates from unexpectedly switching focus between the Start State and Goal State views in the Joints tab.
  * Preserved the user’s active state selection during post-execution and planning-scene synchronization.
  * Explicit user-driven state changes continue to switch the active view as expected.

* **Tests**
  * Added regression coverage for focus preservation during passive state synchronization.


rhaschke (commented):
These are many changes for a rather simple issue.
In MoveIt1, I solved the problem in a much simpler fashion: https://github.com/moveit/moveit/commit/6bd943fcb623761b6bace11f505f91c44986efab

However, always enforcing the goal state might not be desirable as well...
Hint: The joints tab explicitly shows which query state (start or goal) is edited.
Maybe, we should make that state selectable within the widget itself?

rhaschke (commented):
These are many changes for a rather simple issue.
In MoveIt1, I solved the problem in a much simpler fashion: https://github.com/moveit/moveit/commit/6bd943fcb623761b6bace11f505f91c44986efab

However, always enforcing the goal state might not be desirable as well...
Hint: The joints tab explicitly shows which query state (start or goal) is edited.
Maybe, we should make that state selectable within the widget itself?

